PR #1, opened 2015-09-08
shared linux file handling + specinfra config + cleanup

arlimus:
A mixed branch containing:
- Separate Linux file handling (i.e. anything that is sh-based) into its own class; reuse it for local backend and specinfra
- Configure specinfra runtime via --backend or --target (e.g. specinfra+ssh://user@host)
- cleanup (rubocop + gem versions)


chris-rock:
Thanks @arlimus Great work

PR #2, opened 2015-09-08
Package

chris-rock:
- implement package resource with deb, rpm, pacman, brew support
- add specific resource to manage gem, npm and pip
- add be_installed matcher


arlimus:
Huge feature-set, awesome addtion!

PR #9, opened 2015-09-10
Winrm

chris-rock:
- use @conf instance variable
- set default user at a later stage


arlimus:
Thank you for both fixes! :+1:

PR #11, opened 2015-09-14
Concurrent integrationtest

arlimus:
Make sure all container-based integration tests are run fully concurrently without duplicate output in rspec.


chris-rock:
Great! Thanks @arlimus

PR #13, opened 2015-09-14
bugfix: require specinfra backend

chris-rock:


arlimus:
Thank you, will be added to tests later on.

PR #14, opened 2015-09-14
Os detection

chris-rock:
- use backend os detection in resources
- monkey patch specinfra os detection


arlimus:
Thank you for adding it to the backend!

PR #15, opened 2015-09-14
Fix specinfra OS detection

arlimus:


chris-rock:
live saver! thanks @arlimus

PR #16, opened 2015-09-15
Integration tests for the backend runner

arlimus:
Add integration tests for the backend runner, i.e. the layer that executes low-level calls including run_command, file, and os. The current test scope includes the former two.

Test this against real machines by running the local and specinfra+exec backends against vagrant machines using test-kitchen.


chris-rock:
awesome work! thanks @arlimus

PR #19, opened 2015-09-15
Improvements

chris-rock:
- catch nil results in resources
- skip resource instead of fail

### DIFF
--- a/lib/resources/inetd_conf.rb
+++ b/lib/resources/inetd_conf.rb
@@ -29,7 +29,7 @@ class InetdConf < Vulcano.resource(1)
 
   def method_missing(name)
     @params || read_content
-    @params[name.to_s]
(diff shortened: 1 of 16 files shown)


lib/resources/inetd_conf.rb
```diff
@@ -29,7 +29,7 @@ def to_s
 
   def method_missing(name)
     @params || read_content
-    @params[name.to_s]
+    !@params.nil? ? (return @params[name.to_s]) : (return nil)
```
arlimus: how about:

``` ruby
@params.nil? ? nil : @params[name.to_s]
```

chris-rock: yep, lets do it. I change it

chris-rock:
awesome. thanks @arlimus

PR #23, opened 2015-09-18
move integration dependencies to Gemfile

chris-rock:
- move integration dependencies to Gemfile
- add kitchen-vagrant as dendency

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,22 @@
 # encoding: utf-8
 source 'https://rubygems.org'
-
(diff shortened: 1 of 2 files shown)


arlimus:
Great start, let's move the remaining dependencies as well :)

arlimus:
:+1:

chris-rock:
In gemspec we keep dependencies for the gem only. All other dependencies for test, integration etc are kept in Gemfile

PR #26, opened 2015-09-18
bugfix: catch cases, where no service is available

chris-rock:


lib/resources/service.rb
```diff
@@ -311,8 +311,11 @@ def info(service_name)
       return nil
     end
 
-    service['Service']['Status'] == 4 ? (running = true) : (running = false)
-    service['WMI']['StartMode'] == 'Auto' ? (enabled = true) : (enabled = false)
+    # check that we got a response
+    return nil if service.nil? || service['Service'].nil?
+
+    (!service['Service'].nil? && !service['Service']['Status'].nil? && service['Service']['Status'] == 4) ? (running = true) : (running = false)
```
arlimus: Can we skip: 

``` ruby
!service['Service'].nil? && 
```

in this line if it already escapes on LL315?

arlimus: also since the line is quite long, maybe do this in an if/else block instead for styling?

PR #27, opened 2015-09-18
improvement: simplify runner configuration

arlimus:
Don't separate profile_id configuration from the remaining configuration.


chris-rock:
Thanks @arlimus

PR #31, opened 2015-09-21
Exist vs exists

arlimus:
Follow ruby deprecation of `.exists?`


chris-rock:
that is useful. lets stick to standards. Thanks @arlimus

PR #35, opened 2015-09-22
bugfix: backend description for local + docker

arlimus:


chris-rock:
ah that helps users on cli

PR #41, opened 2015-09-22
let travis do dockerized resource tests

arlimus:


arlimus:
It's not perfect, but it should hold for now. Travis and Docker is not as smooth as we might have hoped yet...

chris-rock:
This includes some magic. I like it.

chris-rock:
Thanks @arlimus

PR #43, opened 2015-09-23
Add detect command

arlimus:


chris-rock:
Oh that speeds up some daily tasks. Great stuff @arlimus

PR #46, opened 2015-09-24
Docker runner test

arlimus:
Take the docker runner, throw docker containers against it, see what sticks.

Run via

``` bash
bundle exec rake test:docker_runner
```

Uses the same tests as local_runner.
Includes a micro-provisioner for containers, which it bootstraps on creation-time.


chris-rock:
awesome work. thanks @arlimus

PR #49, opened 2015-09-24
Ssh backend tests

arlimus:


lib/vulcano/backend/specinfra.rb
```diff
@@ -248,7 +251,7 @@ def link_path
 
       def content
         s = Specinfra::Runner.get_file_content(@path).stdout
-        if s.empty? && (directory? or size.nil? or size > 0)
+        if s.empty? && (directory? or (size || 0) > 0 or !exist?)
```
chris-rock: if size is nil, this becomes 0 and is not greater than 0, is this intended?

test/runner/cookbooks/test/recipes/default.rb
```diff
@@ -32,13 +41,32 @@
...
 
-execute 'run tests' do
-  command '/opt/chef/embedded/bin/ruby -I lib test/runner/test.rb test/runner/tests/*_test.rb'
+execute 'run local tests' do
+  command '/opt/chef/embedded/bin/ruby -I lib test/runner/test_local.rb test/runner/tests/*_test.rb'
+  cwd '/tmp/kitchen/data'
+end
+
+execute 'run ssh tests' do
+  command '/opt/chef/embedded/bin/ruby -I lib test/runner/test_ssh.rb test/runner/tests/*_test.rb'
```
chris-rock, line 70: I like that we can run ssh tests like local tests, yeah!

lib/vulcano/backend/specinfra.rb
arlimus: Yes it is, it's a way of not calling size twice, as the result isn't cached and instead called twice... However, it looks confusing, let me take another look.

chris-rock:
Thanks @arlimus

PR #53, opened 2015-09-25
overhaul rule structure

arlimus:
- rename VulcanoBaseRule -> Vulcano::Rule
- initialize rule inside the ProfileContext
- attach all resources to ProfileContext and all rules created within
- rename rule.rb -> dsl.rb, now only containing DSL information
- rename base_rule.rb -> rule.rb, now containing everything for rule


chris-rock:
That was always a little confusing to me. This PR really helps to understand the structure much much better.

PR #56, opened 2015-09-28
OS detection tests

arlimus:
- Add new methods for os family checks: os.linux?, os.unix?, ...
- Move darwin type to BSD
- Change the default mock configuration to be quiet

### DIFF
--- a/lib/vulcano/plugins/backend_os_common.rb
+++ b/lib/vulcano/plugins/backend_os_common.rb
@@ -30,7 +30,7 @@ class Vulcano::Plugins::Backend
 
     OS = {
       'redhat' => %w{
-        redhat oracle centos fedora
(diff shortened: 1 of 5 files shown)


lib/vulcano/plugins/backend_os_common.rb
```diff
@@ -28,6 +28,37 @@ def [](key)
       @platform[key]
     end
 
+    OS = {
+      'redhat' => %w{
+        redhat oracle centos fedora
+      },
+      'debian' => %w{
+        debian ubuntu linuxmint raspbian
```
chris-rock, line 36: is it possible to add scientific and amazon?

chris-rock:
Thanks, this makes handling different OS versions much easier in resources

PR #204, opened 2015-11-03
relax pry version

chris-rock:


arlimus:
This is to prevent conflicts with existing pry installations, where multiple version of pry on the same system may cause require crashes.

PR #206, opened 2015-11-03
Rename gem in test-kitchen example

srenatus:
Obvious fix.

Cheers
Stephan


arlimus:
Good catch, thank you @srenatus !

PR #207, opened 2015-11-03
Fix package resource documentation example.

fnichol:


arlimus:
Thank you @fnichol  :)

PR #214, opened 2015-11-04
Fix contain / should match confusion

zmalone:
If attempting to use "contain" here, you get 

```
     [UNSUPPORTED] `contain` matcher. Please use the following syntax `its('content') { should match('value') }`.
```

it looks like this should say "should match".  This may effect other resources, but I've only tried file.

### DIFF
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -1067,7 +1067,7 @@ The ``content`` matcher tests if contents in the file match the value specified 
 
 .. code-block:: ruby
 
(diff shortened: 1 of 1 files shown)


arlimus:
Great catch, using `match` here is much better.

An alternative, by the way, is to use `include` which is not treated as a regular expression (unlike `match`).

```
describe file(...) do
  its('content') { should include 'hello...' }
end
```

Thank you @zmalone ! :)

PR #218, opened 2015-11-05
Add version to resource declaration

jcreedcmu:
I believe this is necessary, and looking through the git history it seems like it's been necessary for some time.


arlimus:
Good catch, thank you @jcreedcmu ! :+1: 

There was a plan at an early stage to just leave it off and run with v1 if it's nil, but that was never put into practice. A good decision, as this is better practice.

PR #219, opened 2015-11-05
bugfix: correct add_content call to new param structure

arlimus:
this was breaking inspec shell


chris-rock:
Thanks for the fix @arlimus

PR #225, opened 2015-11-07
OS resource requires a Symbol not a String

### LINKED ISSUE #224: OS resource requires requires a Symbol not a String (wrong in docs)
Platform: Windows 10/ChefDK 0.9/InSpec 0.9 + Test Kitchen/AzureRM driver/Windows 2012 R2 Datacenter instance.

As per https://docs.chef.io/release/compliance_1-0/inspec_os.html 

Input:

``` ruby
describe os['family'] do
  it { should eq 'windows' }
end
```

Outcome:

```
  1)  should eq "windows"
     Failure/Error: it { should eq 'windows' }

       expected: "windows"
            got: nil

       (compared using ==)
```

Correct usage appears to be:

``` ruby
describe os[:family] do
  it { should eq 'windows' }
end
```

Submitting a doc PR for this shortly.

stuartpreston:
It appears the OS resource requires a Symbol to be provided rather than a String. Updating the docs to reflect.

Fixes #224

### DIFF
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -2453,7 +2453,7 @@ A ``os`` |inspec resource| block declares the platform to be tested:
 
 .. code-block:: ruby
 
-   describe os['family'] do
+   describe os[:family] do
      it { should eq 'platform' }
    end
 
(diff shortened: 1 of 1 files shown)


chris-rock:
Hi @stuartpreston Thanks for the doc fix!

PR #233, opened 2015-11-12
improvement: rewrite registry_key resource, serverspec compatibility and add integration tests

alexpop:

PR #228, opened 2015-11-09
Upstart with System V fallback

chris-rock:
solves #226 by adding System V fallback for Upstart

PR #234, opened 2015-11-12
lint

arlimus:
These lint errors are currently blocking #228 and #233


chris-rock:
thanks @arlimus for the improvement

PR #237, opened 2015-11-13
api: don't force root on os_env

### LINKED ISSUE #236: os_env not working
I used os_env but i don't get a result from it. I tested it in inspec shell

command('echo $PATH').stdout => "/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl\n"

vs

 os_env('PATH').content => nil

arlimus:
fixes #236

### DIFF
--- a/lib/resources/os_env.rb
+++ b/lib/resources/os_env.rb
@@ -11,15 +11,16 @@
 #   its(:split) { should_not include('.') }
 # end
 
+require 'utils/simpleconfig'
+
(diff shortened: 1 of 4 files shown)


arlimus:
This is also the precursor to reading all ENV params with one go (see osenvparams branch)

jcreedcmu:
:+1:

PR #241, opened 2015-11-16
add port support for centos

### LINKED ISSUE #239: port resource does not work on CentOS
When using port on CentOS, the following is logged

```
The `port` resource is not supported on your OS yet.
```

See https://github.com/chef/inspec/blob/master/lib/resources/port.rb#L26-L38

I think 'centos' should be in the family list for Linux

chris-rock:
- fixes #239
- improves regular expression to parse port data

### DIFF
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -193,36 +193,38 @@ class LinuxPorts < PortsInfo
       port = ip_addr.port
     end
     [host, port]
+  rescue URI::InvalidURIError => e
+    warn "Could not parse #{net_addr}, #{e}"
+    nil
   end
 
   def parse_netstat_line(line)
     # parse each line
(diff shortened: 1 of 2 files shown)


lib/resources/port.rb
```diff
@@ -198,27 +198,31 @@ def parse_net_address(net_addr, protocol)
...
-    }
+      # map data
+      {
+        port: port,
+        address: host,
+        protocol: protocol,
+        process: process,
+        pid: pid,
+      }
+    rescue
```
arlimus: Why the rescue and what errors are caught?

chris-rock: this stabilizes the implementation. In case we cannot parse the line, the whole test will not fail. otherwise a single non-parseable line crashes the whole test

chris-rock:
@arlimus great feedback, I restricted the rescue to `URI::InvalidURIError` error

arlimus:
Great addition and bonus bugfix!!
Thank you so much @chris-rock :+1:

PR #244, opened 2015-11-18
Added links to the different sections.

jjasghar:
- Reading the Doc probably want to use the list as an Table of Contents
  to jump to another section. This does that.
- Group isn't considered Unique in the doc, so the link goes to the same
  page.


chris-rock:
Thanks @jjasghar for this improvement. This removes and annoying piece :+1:

Issue #246, opened 2015-11-19
Multiple computed calls to describe aren't registered


arlimus:
As reported by @schisamo : 

```
%w(
  candle.exe
  heat.exe
  light.exe
).each do |utility|
  describe file("C:/wix/#{utility}") do
    it { should be_file }
  end
```

This will only register one describe call, not all 3.

PR #247, opened 2015-11-19
bugfix: support multiple computed calls to describe

arlimus:
solves #246


chris-rock:
Thanks @schisamo for reporting. Thanks @arlimus for this quick fix

PR #249, opened 2015-11-20
ensure all test directories are on the runner $LOAD_PATH

schisamo:
This change builds on chef/kitchen-inspec#12. All test directories should be on the `$LOAD_PATH` when `Inspec::Runner` executes the test suites with `RSpec::Core::Runner`. This will allow things like `require 'spec_helper'` to work as expected.

/cc @chef/compliance @chef/engineering-services


arlimus:
This is a fantastic compatibility addition!! thank you @schisamo

PR #253, opened 2015-11-23
improvement: update install instructions and add notes for windows builds

chris-rock:
- add description for windows development environments


arlimus:
Good catch and better description, thank you @chris-rock !

PR #255, opened 2015-11-23
Bugfix: make registry_key resource case-insensitive

alexpop:
Add integration test for `NtlmMinClientSec` 

Once https://github.com/chef/inspec/issues/254 is fixed, we should add a test for `NTLMMinServerSec`

### DIFF
--- a/lib/resources/registry_key.rb
+++ b/lib/resources/registry_key.rb
@@ -105,12 +105,14 @@ class RegistryKey < Inspec.resource(1)
     begin
       @registy_cache = JSON.parse(cmd.stdout)
       # convert keys to lower case
-      return @registy_cache = Hash[@registy_cache.map do |key, value|
+      @registy_cache = Hash[@registy_cache.map do |key, value|
         [key.downcase, value]
       end]
     rescue JSON::ParserError => _e
-      return @registy_cache = nil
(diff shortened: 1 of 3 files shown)


chris-rock:
Thanks for adding this @alexpop

chris-rock:
@arlimus added fix to this PR

lib/resources/registry_key.rb
```diff
@@ -87,11 +104,13 @@ def registry_value(path)
     # return nil if cmd.exit_status != 0, try to parse json
     begin
       @registy_cache = JSON.parse(cmd.stdout)
+      # convert keys to lower case
+      return @registy_cache = Hash[@registy_cache.map do |key, value|
+        [key.downcase, value]
+      end]
```
arlimus, line 110: could we:

```
begin
  keys = JSON.parse(cmd.stdout)
  @registry_cache = ... blah ...
rescue JSON::ParseError => _
  @registry_cache = nil
end

@registry_cache
```

This also avoids a duplicate return statement.

chris-rock, line 110: fixed

arlimus:
Thank you @alexpop , this is an awesome addition! Thanks @chris-rock for extending our tests.

PR #258, opened 2015-11-24
bugfix: user resources support for group with whitespace

### LINKED ISSUE #238: User resource doesn't handle group names with spaces
On a Linux server that is joined to a Windows domain, users generaly are member of groups that contain spaces in their names. Eg "domain users". However, this currenly fails.

Test:

```
describe user('jfolmer') do
  its('groups') { should eq ['domain users', 'domain admins', 'denied rodc password replication group'] }
end
```

Output:

```
Failures:

  1) User jfolmer groups should eq ["domain users", "domain admins", "denied rodc password replication group"]

       expected: ["domain users", "domain admins", "denied rodc password replication group"]
            got: ["domain"]
```

chris-rock:
fixes #238

### DIFF
--- a/lib/resources/user.rb
+++ b/lib/resources/user.rb
@@ -193,10 +193,17 @@ class UnixUser < UserInfo
     cmd = inspec.command("id #{username}")
     return nil if cmd.exit_status != 0
 
+    raw = cmd.stdout.chomp
+    data = []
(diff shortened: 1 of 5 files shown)


arlimus:
A non-trivial bug with a great solution! Thank you @chris-rock , :+1: this one is great

Issue #261, opened 2015-11-25
Support the -i switch for key files as per ssh


chris-rock:
Reported by @alexmanly

jcreedcmu:
https://github.com/chef/inspec/pull/262

PR #263, opened 2015-11-25
Add -p alias for --port like ssh

alexmanly:


arlimus:
Great addition, love it :D !
Thank you @alexmanly , also for the `-i` idea in #261

arlimus:
Just to have it referenced: Another shorthand for configuring the port is via target configuration, e.g.:

```
inspec -t user@host:22
```

PR #273, opened 2015-11-27
add kitchen-puppet example with inspec testing

alexpop:


chris-rock:
Thanks for adding this example @alexpop

Issue #280, opened 2015-12-01
Support of linux mint or OS detection via distro families


artem-sidorenko:
Linux Mint is a quite popular (1st place on [distrowatch](http://distrowatch.com/dwres.php?resource=popularity)) linux distro, which is based on Ubuntu LTS.

Obviously linuxmint sells itself in LSB as Mint:)

``` bash
$ cat /etc/lsb-release
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=17.2
DISTRIB_CODENAME=rafaela
DISTRIB_DESCRIPTION="Linux Mint 17.2 Rafaela"
$ cat /etc/upstream-release/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04 LTS"
```

If I try to use inspec on Mint, I get following messages for service and package resources:

``` bash
$ kitchen verify desktop-mint-172-cinnamon
-----> Starting Kitchen (v1.4.2)
.....
-----> Setting up <desktop-mint-172-cinnamon>...
       Finished setting up <desktop-mint-172-cinnamon> (0m0.00s).
-----> Verifying <desktop-mint-172-cinnamon>...
*............*.........*..*

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) System Package crossover The `package` resource is not supported on your OS yet.
     # Not yet implemented
     # /home/artem/.chefdk/gem/ruby/2.1.0/gems/inspec-0.9.5/lib/inspec/runner.rb:114

  2) System Package x11vnc The `package` resource is not supported on your OS yet.
     # Not yet implemented
     # /home/artem/.chefdk/gem/ruby/2.1.0/gems/inspec-0.9.5/lib/inspec/runner.rb:114

  3) Service x11vnc The `service` resource is not supported on your OS yet.
     # Not yet implemented
     # /home/artem/.chefdk/gem/ruby/2.1.0/gems/inspec-0.9.5/lib/inspec/runner.rb:114

  4) System Package owncloud-client The `package` resource is not supported on your OS yet.
     # Not yet implemented
     # /home/artem/.chefdk/gem/ruby/2.1.0/gems/inspec-0.9.5/lib/inspec/runner.rb:114

Finished in 0.45324 seconds (files took 1.97 seconds to load)
27 examples, 0 failures, 4 pending

[...]

PR #281, opened 2015-12-02
Support mint in the integration tests

artem-sidorenko:
As requested in #280, here is the vagrant box

### DIFF
--- a/test/integration/.kitchen.yml
+++ b/test/integration/.kitchen.yml
@@ -34,6 +34,9 @@ platforms:
   - name: ubuntu-12.04-i386
   - name: ubuntu-10.04
   - name: ubuntu-10.04-i386
+  - name: mint-17.2-cinnamon
+    driver_config:
+      box: artem-sidorenko/mint-17.2-cinnamon
 
 suites:
   - name: default


chris-rock:
Thanks for adding this to our test environment @artem-sidorenko !

srenatus:
For future reference: if your `kitchen verify default-mint-172-cinnamon` fails like this,

```
==> default: Waiting for machine to boot. This may take a few minutes...
STDERR: The guest machine entered an invalid state while waiting for it
to boot. Valid states are 'starting, running'. The machine is in the
'poweroff' state. Please verify everything is configured
properly and try again.

If the provider you're using has a GUI that comes with it,
it is often helpful to open that and watch the machine, since the
GUI often has more helpful error messages than Vagrant can retrieve.
For example, if you're using VirtualBox, run `vagrant up` while the
VirtualBox GUI is open.

The primary issue for this error is that the provider you're using
is not properly configured. This is very rarely a Vagrant issue.
---- End output of vagrant up --no-provision --provider virtualbox ----
Ran vagrant up --no-provision --provider virtualbox returned 1]
```

...it might be that VirtualBox is missing its extension pack (`brew cask install virtualbox-extension-pack` for OSX with Brew and Cask)

artem-sidorenko:
@srenatus it might be my fault, probably I didn't remove the usb on the box, so extension pack is needed for it

Issue #284, opened 2015-12-03
Allow for reading options from a file


srenatus:
When I use the inspec CLI, it'd be nice to be able to pass the passwords (ssh for sshpass backend, sudo) by any means different from command line arguments.

Two options come to mind: using _environment variables_ or the process' _stdin_. The former seems less intrusive. ENV vars are passed on to child processes, but this may not be a problem here.

chris-rock:
fixed by #293

PR #285, opened 2015-12-03
allow passing passwords using the environment

srenatus:
This could be done for all options, actually -- opinions?

Also, does the override order "feel right" -- argv overriding env vars?

(Dear github, this is for #284)

### DIFF
--- a/bin/inspec
+++ b/bin/inspec
@@ -21,7 +21,7 @@ class InspecCLI < Thor
     option :user, type: :string, default: nil,
       desc: 'The login user for a remote scan.'
     option :password, type: :string, default: nil,
-      desc: 'Login password for a remote scan, if required.'
+      desc: 'Login password for a remote scan (optional, environment variable INSPEC_SSH_PASSWORD).'
     option :key_files, aliases: :i, type: :array, default: nil,
       desc: 'Login key or certificate file for a remote scan.'
     option :path, type: :string, default: nil,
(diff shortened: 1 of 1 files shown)


alexpop:
Thanks Stephan.
Isn't PASSWORD too generic for an environment variable? It can clash with others.
Same thing for SUDO_PASSWORD.

arlimus:
Password is a bit generic, though also very easy to remember and write ;) then again, what are good alternatives? Maybe `LOGIN_PASSWORD` or similar? What do you guys think?

arlimus:
Btw: overall I really like Stephan's idea of adding passwords in ENV, as it may offer certain runs the option to have it executed, without explicitly appearing in system logs.

srenatus:
I'd go with SSH_PASSWORD, as it's just for the SSH backend...

We could prefix with INSPEC_, too...

alexpop:
WinRM passwords also show in the logs

srenatus:
Updated/prefixed the var names.

srenatus:
@alexpop: you're right, this does not solve that issue, just contribute a little.

chris-rock:
The challenge I see with ENV variables is that those may lead to issues if globally defined. Assume I set the env variable, but forget to kill it for the next inspec run. Inspec would use the env variable again even though this may not be intended by the user.

What about using a json/yaml for describing targets. This would solve multiple use cases:
- all credentials are stored in file
- the target definition can be reused multiple times

``` json
{
  "host" : "example.com",
  "user" : "root",
  "password" : "abc"
}
```

Alternatively we can also think about using the systems keychain if available.

chef-supermarket:

- @srenatus

arlimus:
@chris-rock I like the idea of JSONy targets, we should add that.

System keychains are great for SSH, though they migh also lead to confusion if we pull anything for e.g. sudo password. Even then, unexpectedly pulling from it leads to the same result as having it in the env variable, imho.

Env variables are a great use-case, though with json support, we should check if this feature is actually needed by us or the community. @srenatus If we don't have those right now, how about icing this in favor of json targets?

srenatus:
Sure, let's do that. Closing this.

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

PR #286, opened 2015-12-03
travis workarounds

srenatus:
builds are green again! yay \o/

This is what happened, today travis started routing 50% of the precise-based legacy builds to a new infrastructure (incl. new versions of things). Switched to Trusty since fixing the gem issue brought up an issue with installing lxc in this new infrastructure, so I switched us to that.

References:
- https://github.com/travis-ci/travis-ci/issues/5142#issuecomment-160859891
- https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ cache: bundler
 
 bundler_args: --without integration tools maintenance
 
+before_install:
+  - gem update --system 2.4.5
+  - gem --version
+
 matrix:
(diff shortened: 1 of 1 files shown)


chris-rock:
:100: great fix @srenatus and this makes the whole tests a lot faster. so cool.

solarce:
:heart_eyes:

PR #289, opened 2015-12-03
Adding support for Wind River Linux

adamleff:
WRL is used as the OS on Cisco Nexus devices and acts like a Red
Hat variant. These changes add support for WRL.


chris-rock:
Thanks @adamleff Super cool. Thanks for this great addition.

PR #291, opened 2015-12-04
revert to old find_files interface

### LINKED ISSUE #276: Utils::FindFiles doesn't work
https://gist.github.com/nathenharvey/7303c8dc3baaf2c09717

as repoterd by @nathenharvey , thank you!

srenatus:
Fixes #276.

Included tests are minimal.

### DIFF
--- a/lib/utils/find_files.rb
+++ b/lib/utils/find_files.rb
@@ -16,7 +16,12 @@ module FindFiles
     door: 'D',
   }
 
+  # ignores errors
   def find_files(path, opts = {})
+    find_files_or_error(path, opts) || []
+  end
+
+  def find_files_or_error(path, opts = {})
     depth = opts[:depth]
     type = TYPES[opts[:type].to_sym]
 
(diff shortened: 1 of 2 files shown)


lib/utils/find_files.rb
```diff
@@ -17,6 +17,17 @@ module FindFiles
...
 
   def find_files(path, opts = {})
+    files, exit_status = find_files_with_exitstatus(path, opts)
+
+    if exit_status != 0
+      warn "find_files(): non-zero exit status: #{exit_status}"
+      return []
+    else
+      return files
+    end
```
chris-rock: Maybe we can simplify this:

```
return files if exit_status == 0
# received an error
warn "find_files(): non-zero exit status: #{exit_status}"
return []
```

lib/utils/find_files.rb
```diff
@@ -24,7 +35,7 @@ def find_files(path, opts = {})
     cmd += " -maxdepth #{depth.to_i}" if depth.to_i > 0
     cmd += " -type #{type}" unless type.nil?
 
-    result = inspec.run_command(cmd)
+    result = inspec.command(cmd)
     exit_status = result.exit_status
 
     return [nil, exit_status] unless exit_status == 0
```
chris-rock: Currently we handle the exit status twice. Would could:
- just not return the `exit_status` if not required, only `files` or `[]`
- we return `[]` as default instead of `nil`

This would allow us to simplify the `find_files` method

chris-rock:
@srenatus Thanks for the improvement of the find_files test. This is much better now. I think we can simplify the exit code handling even further.

srenatus:
@chris-rock ok, updated (if I understood your proposal correctly). While the `find_files_or_error` method might seem superfluous, it conveys the intention of https://github.com/chef/inspec/commit/1345c1d017dcb150ecd9c510a53b007866088749 to allow the utility method user to recognize errors.

:warning: That said, the method is still not robust, the "default" handling doesn't work; will add tests and fixes in a minute.

chris-rock:
cool that is so simple now.

srenatus:
Tests should be improved, but I'd rather just see this bug gone now. :)

chris-rock:
Thanks @srenatus for this fix.

PR #293, opened 2015-12-04
read config from file/stdin

### LINKED ISSUE #292: JSON configuration
As discussed in #285, inspec should be able to read its configuration from a JSON file.

This could be another (global) flag, `--json-config=FILE`, with special handling of `FILE=-` for reading from stdin.

(It's both fixing a bug (passes on command-line) and providing an enhancement...)

srenatus:
Fixes #292.

Removing the defaults was pragmatic:
1. `options['something']` is `nil` when the key isn't present anyways (and thus both `default: nil` and `default: false` are not necessary)
2. If the defaults are there, they would override the `options_json` hash in the merge -- and we actually want the CLI parameters to take precedence.

An alternative might have been to declare an explicit "defaults" hashmap, but since the available configurables are supposed to come from train directly (in the future), this might not make too much sense now.

Shall I squash those commits?

### DIFF
--- a/bin/inspec
+++ b/bin/inspec
@@ -157,6 +157,9 @@ class InspecCLI < Thor # rubocop:disable Metrics/ClassLength
     end
 
     JSON.load(config)
+  rescue JSON::ParserError => e
+    puts "Failed to load JSON configuration: #{e}\nConfig was: #{config.inspect}"
+    exit 1
   end
 end
 InspecCLI.start(ARGV)
--- a/bin/inspec
+++ b/bin/inspec
@@ -6,9 +6,13 @@
 
 require 'thor'
 require 'json'
+require 'pp'
 require_relative '../lib/inspec'
 
-class InspecCLI < Thor
+class InspecCLI < Thor # rubocop:disable Metrics/ClassLength
+  class_option :diagnose, type: :boolean,
+    desc: 'Show diagnostics (versions, configurations)'
+
   def self.target_options
     option :target, aliases: :t, type: :string, default: nil,
       desc: 'Simple targeting option using URIs, e.g. ssh://user:pass@host:port'
@@ -36,6 +40,8 @@ class InspecCLI < Thor
       desc: 'Use SSL for transport layer encryption (WinRM).'
     option :self_signed, type: :boolean, default: false,
       desc: 'Allow remote scans with self-signed certificates (WinRM).'
+    option :json_config, type: :string,
+      desc: 'Read configuration from JSON file (`-` reads from stdin).'
   end
 
   desc 'json PATH', 'read all tests in PATH and generate a JSON summary'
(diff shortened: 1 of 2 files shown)


bin/inspec
```diff
@@ -106,5 +122,42 @@ class InspecCLI < Thor
...
+  def options_json
+    conffile = options['json_config']
+    return read_config(conffile) if conffile
+
+    {}
+  end
+
+  def read_config(file)
+    if file == '-'
+      puts 'WARN: reading JSON config from standard input' if STDIN.tty?
```
arlimus, line 153: what advantage do we get from printing a warning here? I'm just thinking about e.g. raw json output on stdout, where this might interfere

arlimus:
Awesome new functionality + diagnose helper!

bin/inspec
```diff
@@ -106,5 +122,42 @@ class InspecCLI < Thor
...
+    return unless opts['diagnose']
+    puts "InSpec version: #{Inspec::VERSION}"
+    puts "Train version: #{Train::VERSION}"
+    puts 'Command line configuration:'
+    pp options
+    puts 'JSON configuration file:'
+    pp options_json
+    puts 'Merged configuration:'
+    pp opts
+  end
```
arlimus, line 139: What do you think about an additional newline?

Here's a use-case where this might be clearer:

```
./bin/inspec detect --diagnose
InSpec version: 0.9.5
Train version: 0.9.1
Command line configuration:
{"diagnose"=>true}
JSON configuration file:
{}
Merged configuration:
{"diagnose"=>true}
{"name":null,"family":"arch","release":"4.2.5-1-ARCH","arch":null}
```

The last line feels like a part of diagnose.

arlimus, line 139: Or am I just using it wrong?

srenatus, line 139: No, it's fine. I'll add a newline.

bin/inspec
srenatus, line 153: @arlimus when a user passes `--json-config=-`, she will just see that nothing happens. With this warning, the user will get an indication what is expected of her at the moment (i.e. provide JSON from the tty). It just affects the case of "reading from stdin when the session is interactive" (as opposed to non-interactive, i.e. `cat config.json | inspec detect --json-config=-`)

srenatus:
I think I missed something with the reading from STDIN. please don't merge this yet :warning:

srenatus:
@arlimus problem solved, ready to go now :)

arlimus:
Ah, didn't see that. Thanks @srenatus , that solves it!

srenatus:
As discussed, I'll give this one a touch-up. Please don't merge. :warning:

srenatus:
ok, done :)

arlimus:
:+1: Thank you @srenatus !!

Issue #290, opened 2015-12-04
Create missing resource unit tests


srenatus:
``` interactive
$ for file in lib/resources/*; do ls test/unit/resources/$(basename $file .rb)_test.rb; done | grep No
ls: test/unit/resources/apache_test.rb: No such file or directory
ls: test/unit/resources/apache_conf_test.rb: No such file or directory
ls: test/unit/resources/command_test.rb: No such file or directory
ls: test/unit/resources/directory_test.rb: No such file or directory
ls: test/unit/resources/file_test.rb: No such file or directory
ls: test/unit/resources/mysql_test.rb: No such file or directory
ls: test/unit/resources/mysql_session_test.rb: No such file or directory
ls: test/unit/resources/os_test.rb: No such file or directory
ls: test/unit/resources/parse_config_test.rb: No such file or directory
ls: test/unit/resources/postgres_test.rb: No such file or directory
ls: test/unit/resources/postgres_conf_test.rb: No such file or directory
ls: test/unit/resources/postgres_session_test.rb: No such file or directory
ls: test/unit/resources/windows_feature_test.rb: No such file or directory
```

arlimus:
Still an issue for the following resources:

```
ls: cannot access 'test/unit/resources/apache_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/command_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/directory_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/groups_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/iis_site_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/mssql_session_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/mysql_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/mysql_session_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/postgres_conf_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/postgres_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/postgres_session_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/ssl_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/users_test.rb': No such file or directory
ls: cannot access 'test/unit/resources/windows_feature_test.rb': No such file or directory
```

clintoncwolfe:
bump

zenspider:
Current:

[...]

PR #298, opened 2015-12-07
[resources/apache_conf]: add tests, fix bug

srenatus:
before, the resource would throw an exception when include_files
returned nil (i.e., [].flatten!)

added basic unit tests capturing the include_files behaviour (addresses #290)


chris-rock:
@srenatus Thanks for this improvement and the extended test coverage

PR #301, opened 2015-12-07
File permission checks should return false unless file exists

adamleff:
Currently, #readable?, #writeable?, and #executable? will incorrectly
return true if the file does not exist.

```
$ cat test.rb
describe file('/tmp/test.txt') do
  it { should be_readable }
  it { should be_writeable }
  it { should be_executable }
end

$ stat /tmp/test.txt
stat: /tmp/test.txt: stat: No such file or directory

$ inspec exec ./test.rb
...

Finished in 0.03351 seconds (files took 0.12235 seconds to load)
3 examples, 0 failures
```

This patch forces these three methods to immediately return false unless the file exists.

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -33,4 +33,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'rspec', '~> 3.3'
   spec.add_dependency 'rspec-its', '~> 1.2'
   spec.add_dependency 'pry', '~> 0'
+
+  spec.add_development_dependency 'mocha', '~> 1.1'
 end
--- a/lib/resources/file.rb
+++ b/lib/resources/file.rb
@@ -19,7 +19,7 @@ module Inspec::Resources
       end
     "
 
-    attr_reader :path
+    attr_reader :file, :path
     def initialize(path)
       @path = path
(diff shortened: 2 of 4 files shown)


chris-rock:
@adamleff great fix. Can you add some tests as well?

adamleff:
@chris-rock sure I can do that. I had trouble locating unit tests for the file resource. Am I missing it or are there no tests for this resource yet?

chris-rock:
@adamleff You are right, we have no unit test for file because those tests are part of train. What we implemented for file is a integration test:
https://github.com/chef/inspec/blob/master/test/integration/test/integration/default/file_spec.rb

I would love to see this implemented here.

adamleff:
@chris-rock I started to write tests for the File resource but found it quite difficult based on the layout of the Class.  I took the opportunity to refactor it to eliminate some repetitiveness in the code, make some more single-purpose methods, and make it far easier to test.  I then included a full unit test for the class.  Please check it out and let me know what you think.

@arlimus if you have feedback too, I'm totally open to it!

chris-rock:
@adamleff awesome improvement. It covers some cleanup that was on my agenda. Great work :100:

PR #303, opened 2015-12-08
port resource: array attributes, resource alternative

### LINKED ISSUE #256: Port resource returns arrays
process, pid and protocol for port all get compared to an array instead of string/number causing weirdness like this

```
Port  22 pid should eq "957"

      expected: "957"
           got: [957]
```

Thanks for reporting Thomas Cate

srenatus:
Fixes #256.

Not sure if there might not be some potential for misunderstanding here, though: practically, a process listening in `0.0.0.0:80` is of course reachable on, say, `192.168.1.1:80` on the same machine -- but asking inspec for the latter will say that it isn't.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ end
 
 describe port(443) do
   it { should be_listening }
-  its('protocol') {should eq 'tcp'}
+  its('protocols') {should include 'tcp'}
 end
 ```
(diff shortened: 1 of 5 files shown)


chris-rock:
Great improvement @srenatus

Issue #308, opened 2015-12-10
Configuration number comparisons


arlimus:
When running against a configuration file with numbers, e.g. ssh_config with:

```
Port 22
```

The test would have to be written:

```
its('Port') { should eq '22' }
```

However users also expect to be able to write:

```
its('Port') { should eq 22 }
```

It would be great to find a way to reconcile this and help users to write the test they expect.

chris-rock:
with #318 it is possible to write:

``` ruby
describe sshd_config do
  its('Port') { should cmp '22' }
  its('Port') { should cmp 22 }
end
```

Issue #307, opened 2015-12-10
auditd_conf parameters should be case insensitive


arlimus:
```
describe auditd_conf do
   its('log_format') { should eq 'raw' } 
   its('log_format') { should eq 'RAW' }
end
```

chris-rock:
+1

chris-rock:
With the latest PR #318 it is possible to write

``` ruby
describe auditd_conf do
   its('log_format') { should cmp 'raw' } 
   its('log_format') { should cmp 'RAW' }
end
```

PR #318, opened 2015-12-11
matcher for less-restrictive comparison

chris-rock:
This matcher implements a less-restrictive comparison. Currently, it was not possible to write integer comparison (#308)

``` ruby
describe sshd_config do
  its('Port') { should eq '22' } # works
  its('Port') { should eq 22 } # does not work
end
```

As highlighted in #307 some resources have case-insensitive values, and I would like tests write

``` ruby
describe auditd_conf do
   its('log_format') { should eq 'raw' } # works
   its('log_format') { should eq 'RAW' } # does not work
end
```

This PR introduces a new comparison matcher `cmp`, that covers those cases:

``` ruby
describe sshd_config do
  its('Port') { should cmp '22' }
  its('Port') { should cmp 22 } # works with integer
  its('Port') { should cmp 22.0 } # also understand float
  its('Port') { should_not cmp 22.1 }

  # case-insensitive
  its('LogLevel') { should cmp 'INFO' } 
  its('LogLevel') { should cmp 'info' }
  its('LogLevel') { should cmp 'InfO' }
end

```

### DIFF
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -275,6 +275,13 @@ This |inspec resource| matches any keyword that is listed in the ``auditd.conf``
 
    its('log_format') { should eq 'raw' }
 
+Since all option names and values are case insensitive for ``auditd_conf``, we recommend to compare values with `cmp` instead of the `eq`.
+
+.. code-block:: ruby
+
+   its('log_format') { should cmp 'raw' }
+   its('max_log_file') { should cmp 6 }
+
 Examples
 -----------------------------------------------------
 The following examples show how to use this InSpec audit resource.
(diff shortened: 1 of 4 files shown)


chris-rock:
fixes #307
fixes #308

docs/resources.rst
```diff
@@ -275,6 +275,13 @@ This |inspec resource| matches any keyword that is listed in the ``auditd.conf``
 
    its('log_format') { should eq 'raw' }
 
+Since all option names and values are case insensitive for ``auditd_conf``, we recommend to compare values with `cmp` instead of the `eq`.
+
+.. code-block:: ruby
+
+   its('log_format') { should cmp 'raw' }
+   its('max_log_file') { should cmp 6 }
+
```
arlimus, line 278: Could we establish cmp as the default matcher instead of eq for auditd_conf? Since the resource would be perfect for that! (i.e. just swap with eq + love the explanation!)

chris-rock, line 278: That is a great idea. I updated the docs.

arlimus:
Really great new functionality, love this from a user's standpoint! Great feature, thank you @chris-rock :santa: :+1:

PR #321, opened 2015-12-15
Fixing issue with security policy always returning nil

jeremymv2:
All of the Security Policies were evaluating to nil.
This fixes the issues and allows it to work correctly.
I imagine this wasn't caught earlier due to a file artifact (win_secpol.cfg) being left around from a manual execution of the secedit export command.

Here is an example of what happens when trying to evaluate any security_policy:

<pre>
Failures:

  1) Security Policy EnableAdminAccount should eq 1
     Failure/Error: its('EnableAdminAccount') { should eq 1 }

       expected: 1
            got: nil

       (compared using ==)
     # ./test/integration/default/guest_account_spec.rb:5:in `block (3 levels) in load'
     # /Users/jmiller/.chefdk/gem/ruby/2.1.0/gems/kitchen-inspec-0.9.6/lib/kitchen/verifier/inspec.rb:55:in `call'

Finished in 0.91339 seconds (files took 0.70664 seconds to load)
8 examples, 1 failure

Failed examples:

rspec  # Security Policy EnableAdminAccount should eq 1
</pre>

### DIFF
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -30,15 +30,19 @@ class SecurityPolicy < Inspec.resource(1)
   # load security content
   def load
     # export the security policy
-    inspec.command('secedit /export /cfg win_secpol.cfg')
-    # store file content
-    command_result ||= inspec.command('type win_secpol.cfg')
-    # delete temp file
-    inspec.command('del win_secpol.cfg')
+    cmd = inspec.command('secedit /export /cfg win_secpol.cfg')
+    return nil if cmd.exit_status.to_i != 0
 
-    @exit_status = command_result.exit_status.to_i
-    @policy = command_result.stdout
+    # store file content
+    cmd = inspec.command('Get-Content win_secpol.cfg')
+    @exit_status = cmd.exit_status.to_i
+    return nil if @exit_status != 0
+    @policy = cmd.stdout
     @loaded = true
+    
+    # delete temp file
+    cmd = inspec.command('Remove-Item win_secpol.cfg')
+    return nil if cmd.exit_status.to_i != 0
 
     # returns self
     self
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -40,12 +40,12 @@ class SecurityPolicy < Inspec.resource(1)
     @policy = cmd.stdout
     @loaded = true
     
-    # delete temp file
-    cmd = inspec.command('Remove-Item win_secpol.cfg')
-    return nil if cmd.exit_status.to_i != 0
-
     # returns self
     self
+
+  ensure
+    # delete temp file
+    inspec.command('Remove-Item win_secpol.cfg').exit_status.to_i
   end
 
   def method_missing(method)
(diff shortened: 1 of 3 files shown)


lib/resources/security_policy.rb
```diff
@@ -30,15 +30,19 @@ def initialize
...
-    @policy = command_result.stdout
+    # store file content
+    cmd = inspec.command('Get-Content win_secpol.cfg')
+    @exit_status = cmd.exit_status.to_i
+    return nil if @exit_status != 0
+    @policy = cmd.stdout
     @loaded = true
+    
+    # delete temp file
+    cmd = inspec.command('Remove-Item win_secpol.cfg')
```
chris-rock: Please ensure that we delete it before we return. Even if we have an error, we should at least try to the file.

chris-rock:
@jeremymv2 thanks for bringing this up. I like to move to Get-Content instead of `type`. Lets just improve it to delete files even if something went wrong. Which OS have you used for testing?

lib/resources/security_policy.rb
jeremymv2: You got it.  I'll get that in.

jeremymv2:
@chris-rock I'm testing with:

<pre>
platforms:
  - name: win-2012r2-standard-amd64-nocm
    driver:
      box: opentable/win-2012r2-standard-amd64-nocm
      box_url: https://atlas.hashicorp.com/opentable/boxes/win-2012r2-standard-amd64-nocm/versions/1.0.0/providers/virtualbox.box
</pre>

jeremymv2:
@chris-rock Added an ensure block to always delete the file.   Tested and works on a windows 2012R2 image.

chris-rock:
@jeremymv2 that is great. Could you do me one favor and add windows platform as a separate PR to our integration tests https://github.com/chef/inspec/blob/master/test/integration/.kitchen.yml?

On top of this PR I would like to have an integration test added as well to ensure we do not miss this again. Let me know if you need help.

chris-rock:
@jeremymv2 One idea that came up in my head: At the point we wrote this resource we had no script resource available in InSpec. Now, it may be easier to just write a simple powershell script. Similar to https://github.com/chef/inspec/blob/f092ba3ac3790d074fc0a458cc4f5e3dc4d4e54d/lib/resources/registry_key.rb#L93-L108 

That would reduce the calls to one request instead of three. What do you think?

jeremymv2:
@chris-rock added a simple integration test for security_policy which should catch this root issue if it occurs again in the future.

 A couple of other things:
1.) Although the idea of embedded powershell scripts sometimes seems appealing, in general I find that it adds too much obfuscation when more than just a one or two-liner.  Powershell is just hard to read and requires the reader to context switch heavily.
2.) There are several things that aren't working integration-wise for windows.  One example is the json_yaml_csv_ini.rb recipe doesn't support the windows platform so it prevents a successful converge.  Many of the of the inspec tests assume the platform is Linux so they run on Windows and fail.  I didn't want to include all those fixes in this pull request.  I will be glad to add those items in other PRs.

chris-rock:
@jeremymv2 Great work and thanks for the improvement. Regarding the integration tests: we started some work in https://github.com/chef/inspec/pull/314 to make this work for Windows. I invite you to contribute more tests and especially to improve the Windows support.

chris-rock:
When travis shows green light, I am going to merge it. Please rebase your improvements on the latest master.

PR #326, opened 2015-12-17
remove format default for `inspec exec`

srenatus:
behaviour is unchanged. (this caused trouble in chef-compliance, using JSON configuration)


chris-rock:
Thanks @srenatus

PR #327, opened 2015-12-18
remove commandline defaults, they break json config

srenatus:
I don't know how that happened, but I had intended this to be part of #293 (even mentioned it in the text...).

It's necessary such that the JSON config (from file or STDIN) has any effect (and is not overridden by the defaults).


arlimus:
Looks good, thank you for fixing it @srenatus

PR #329, opened 2015-12-22
Resource grub conf

thomascate:
Adding a grub_conf resource for parsing boot options on centos/rhel machines. Currently it uses split to break up long kernel lines, putting all kernel entries into an array that can be checked with includes. The example config I've been using is in test/unit/mock/files/grub.conf.


chef-supermarket:

- @thomascate

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

thomascate:
Hi guys, here's a first pass at making a resource for checking grub configs. Currently it only works on rhel/centos/fedora style configs. Ubuntu at least from 14.04 will need it's own parsing logic.

chris-rock:
@thomascate This is a great start and very complete with tests. To find a platform-independent abstraction for grub, lets add a ubuntu example before we merge.

We should also highlight, if we support grub 1 or grub 2 or both.

I am not sure if we should mix the parameters of grub and menu entires. That may lead to confusion.

This is great:

```
describe grub_conf('/etc/grub.conf') do
  its('default') { should_not eq '1' }
  its('timeout') { should eq '5' }
end
```

What I am not sure about is the following since it returns all kernels.

```
describe grub_conf('/etc/grub.conf') do
  its('kernel') { should include '/vmlinuz-2.6.32-573.7.1.el6.x86_64' }
  its('initrd') { should include '/initramfs-2.6.32-573.el6.x86_64.img=1' }
end
```

Maybe we think about a filter like:

```
describe grub_conf('/etc/grub.conf', 'CentOS (2.6.32-573.7.1.el6.x86_64)') do
  its('root') { should include 'root (hd0,0)' }
  its('kernel') { should include '/vmlinuz-2.6.32-573.7.1.el6.x86_64' }
  its('initrd') { should include '/initramfs-2.6.32-573.el6.x86_64.img=1' }
end
```

What do you think?

thomascate:
I'm not sure how we want to handle legacy-grub vs grub2 since they have significantly different configs. Currently this only handles legacy-grub which covers Centos/RHEL <= 6 and should work on Ubuntu <= 9.10. We could either do OS detection in the resource and put both cases, or just split it into 2 resources. Adding grub2 support would cover Ubuntu support, as well Debian and probably several other OSs.

I wasn't super happy with it testing all kernels either. But testing on the exact kernel version is going to get broken often, since kernel updates are pretty regular. How about parsing for the default boot kernel and separating that?

```
describe grub_conf('/etc/grub.conf', 'default') do
  its('root') { should include 'root (hd0,0)' }
  its('kernel') { should include '/vmlinuz-2.6.32-573.7.1.el6.x86_64' }
  its('initrd') { should include '/initramfs-2.6.32-573.el6.x86_64.img=1' }
end
```

Since the default boot kernel entry is almost always the only one the person testing will be concerned about.

thomascate:
I added the ability to check the default kernel as well as filtering by exact kernel version like you described. I'm working on Grub2 support like we discussed yesterday.

chris-rock:
:+1:

thomascate:
FYI, I still have some refactoring and tests to do to get this ready.

thomascate:
Sorry about the long delay between commits, I was pulled away on other work. Currently I have one rubocop error commented out for class length. I made a couple attempts to rewrite this to fix the error with no success. Currently the code works, but if someone wants to be pair with me on a rewrite to clean that up I'd be happy to.

PR #347, opened 2016-01-05
Support old "supports" field in metadata

srenatus:
current compliance-profiles still have

``` yaml
supports:
  - linux
```

and we might want to be a little backwards compatible, too.

:warning: the logging does not work there, needs investigation.

WIP because tests are missing

### DIFF
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -46,10 +46,7 @@ module Inspec
     end
 
     def is_supported(os, entry)
-      try_support = self.class.symbolize_keys(entry)
-      name = try_support[:'os-name'] || try_support[:os]
-      family = try_support[:'os-family']
-      release = try_support[:release]
+      name, family, release = support_fields(entry)
 
       # return true if the backend matches the supported OS's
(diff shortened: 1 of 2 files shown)


chris-rock:
:+1:

lib/inspec/metadata.rb
```diff
@@ -74,6 +71,21 @@ def is_supported(os, entry)
       name_ok && family_ok && release_ok
     end
 
+    def support_fields(entry)
+      @logger.warn("logging")
```
arlimus: one littel debugging call left ^^

srenatus: Ach...

arlimus:
Thank you @srenatus , this fixes our migration issues.

arlimus:
Small clarification: We haven't had an official handling on `supports` until the recent pinning in https://github.com/chef/inspec/pull/344 . Since there was still a backdrop to the old mode, we don't want to just silentry drop it, just because it wasn't specified yet. The old way is still deprecated, but it won't break inspec from running.

srenatus:
@arlimus thanks for fixing that up before the merge. :)

PR #355, opened 2016-01-11
bugfix: ignore supports when generating a profile's json representation

srenatus:
without this, `inspec json PATH` does never contain rules != {}, because
of the usage of the mock backend


arlimus:
Thank you for fixing this @srenatus !

PR #363, opened 2016-01-14
Correctly detect UDP ports on linux

mivok:
netstat on UDP lines doesn't display a port state (e.g. LISTEN), so make this
field optional when parsing the netstat line.


chris-rock:
Great catch @mivok Thank you very much.

PR #364, opened 2016-01-14
bugfix: only skip regkey if required

chris-rock:


lib/resources/registry_key.rb
```diff
@@ -26,7 +26,8 @@ def initialize(name, reg_key = nil)
     reg_key ||= name
     @name = name
     @reg_key = reg_key
-    skip_resource 'The `registry_key` resource is not supported on your OS yet.'
+
+    return skip_resource 'The `registry_key` resource is not supported on your OS yet.' if inspec.os[:family] != 'windows'
```
arlimus: maybe `inspec.os.windows?`

chris-rock: yeah, for sure, thats how you write it. looks like my head is low on sugar :-)

arlimus:
Great catch, cheers @chris-rock !

PR #365, opened 2016-01-15
Fix ssh_config example

tas50:


arlimus:
Sorry for the build failures, we're fixing those linting changes in rubocop: https://github.com/chef/inspec/pull/366
(also pinning rubocop so this type of silent update failure doesnt happen again)

chris-rock:
Thanks @tas50 for finding this

PR #367, opened 2016-01-15
make metadata.rb legacy mode consistent for supports

arlimus:
Before introducing InSpec profiles in https://github.com/chef/inspec/pull/252 we had `metadata.rb` keep all information. This included an undisclosed field called `supports`. However, this field was never actually used in practice. So for legacy profiles, this means that `supports` was ignored. In order to keep old profiles running in exactly the way they were before, ignore this field when reading from metadata.rb

### DIFF
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -35,14 +35,9 @@ module Inspec
     end
 
     def supports(sth, version = nil)
-      params[:supports] ||= []
-      params[:supports].push(
-        # do not use hash syntax for `-` to work properly with ruby 1.9.3 parser
-        {
-          :'os-name' => sth, # rubocop:disable Style/HashSyntax, Lint/UnneededDisable
-          version: version,
-        },
-      )
+      # Ignore supports with metadata.rb. This file is legacy and the way it
+      # it handles `supports` deprecated. A deprecation warning will be printed
+      # already.
     end
 
     def is_supported(os, entry)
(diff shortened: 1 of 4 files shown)


chef-supermarket:

* @arlimus

lib/inspec/metadata.rb
```diff
@@ -35,14 +35,9 @@ def initialize(ref, logger = nil)
     end
 
     def supports(sth, version = nil)
-      params[:supports] ||= []
```
chris-rock, line 38: we share the same implementation for `inspec.yml` and `metadata.rb`. I would assume, this breaks the new feature as well?

arlimus:
This causes an error with compliance profiles, that use the old metadata.rb structure and fill in the field supports. It handles all those targets correctly, where supports specifies the OS by name. However, with setting:

```
supports 'linux'
```

it will fail on all systems (unless you find one where OS-name is 'linux' ;) ).

This takes care of the issue, by ignoring that field completely, since that syntax is deprecated.

lib/inspec/metadata.rb
arlimus, line 38: Actually not quite: We load yaml and then only finalize the params. That function is never called :)

Look:
- https://github.com/chef/inspec/pull/367/files#diff-0bdee087faa99fbed7459cc5302e01efL141
- https://github.com/chef/inspec/pull/367/files#diff-0bdee087faa99fbed7459cc5302e01efL133

chris-rock: Oh sure. Got you. Can we add some inspec.yml tests to increase our confidence :-)

PR #379, opened 2016-01-21
integration tests: stop using alpine iso

srenatus:
`empty.iso` is 348K. Makes integration tests a little faster by not downloading the alpine iso (~80M).


chris-rock:
@srenatus great idea, can you automate the creation of the iso as well?

srenatus:
@chris-rock uhm... we could, it just takes `package 'genisofs'` and an execute, but should we? I'd say it's ok to include this fixture. (We didn't create the alpine image either.)

chris-rock:
Thanks @srenatus Just missed that the iso is included. Awesome improvement

PR #380, opened 2016-01-21
add service overrides for picking specific service managers, add runit_service

### LINKED ISSUE #242: specify process supervision type in service resource
The Inspec service resource determines the process supervision type based on os release version and uses a single process supervision type to check if a process is running, enabled etc. 

Given many chef cookbooks use many different process supervison tools (init, upstart, runit, etc etc) It would be better if you could override the service_mgmt instance variable to specify the specific one for your service.

Is this a design decision, or could this be changed?

srenatus:
The changes to `lib/resource/service.rb` have gotten a little bigger. The handling of `service_ctl` still feels a little awkward, so when you review this, please keep your eyes open for further possible simplifications.

The problem is that when you use `runit_service('name', '/path/to/sv')`, the latter needs to be passed to the underlying `ServiceManager`, and thus both `Service` and `ServiceManager` need the optional argument in their constructors.

I've added integration tests installing runit on centos.

Fixes #242.

TODO: Documentation -- so, currently, `docs/resources.rst` has to be updated manually?


lib/resources/service.rb
```diff
@@ -270,19 +285,14 @@ def info(service_name)
...
 
-    # on debian service is located /usr/sbin/service, on centos it is located here /sbin/service
-    service_cmd = 'service'
-    service_cmd = '/usr/sbin/service' if inspec.os[:family] == 'debian'
-    service_cmd = '/sbin/service' if inspec.os[:family] == 'centos'
-
-    cmd = inspec.command("#{service_cmd} #{service_name} status")
-    cmd.exit_status == 0 ? (running = true) : (running = false)
+    cmd = inspec.command("#{service_ctl} #{service_name} status")
+    running = cmd.exit_status == 0
```
chris-rock, line 295: :+1:

lib/resources/service.rb
```diff
@@ -138,13 +147,13 @@ def info(service_name)
     ).params
 
     # LoadState values eg. loaded, not-found
-    params['LoadState'] == 'loaded' ? (installed = true) : (installed = false)
+    installed = params['LoadState'] == 'loaded'
```
chris-rock, line 150: I like the simplification a lot.

chris-rock:
awesome job @srenatus

PR #386, opened 2016-01-22
bugfix: pin net-ssh 2.9 in gem file

chris-rock:
Unfortunately bundler seems not to be smart enough to remove a version that does not fit to the ruby version. Therefore we pin net-ssh for builds on ruby 1.9.3


srenatus:
Hmm appveyor does take some time...

PR #387, opened 2016-01-22
resources/launchd_service: fix parsing of non-running services

srenatus:
If the service is not running, its pid is given as `"-"`.


chris-rock:
@srenatus I like the improvement for the regular expressions.

PR #392, opened 2016-01-27
basic logging setup for `inspec exec`

### LINKED ISSUE #349: Logging in profile processing
With #347, it turns out that using `inspec check` on a profile (path) that has an inspec.yml with the undocumented old-style

``` yaml
supports:
  - linux
```

[this line](https://github.com/chef/inspec/pull/347/files#diff-0bdee087faa99fbed7459cc5302e01efR81) is not actually output. Somewhere in the call chain, something may be wrong with passing the `logger` argument.

srenatus:
This change tries hard to stay backward compatible: if you ask for JSON
output, you will NOT get any logging.

Fixes #349.

### DIFF
--- a/bin/inspec
+++ b/bin/inspec
@@ -107,7 +107,10 @@ class InspecCLI < Thor # rubocop:disable Metrics/ClassLength
   def exec(*tests)
     diagnose
 
-    runner = Inspec::Runner.new(opts)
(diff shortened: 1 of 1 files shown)


arlimus:
Looks great, thank you Stephan!

PR #401, opened 2016-01-29
runit_service: fix resource, improve integration tests

srenatus:
Turns out using `/usr/bin/yes` to imitate a daemon process is a TERRIBLE idea.


arlimus:
:+1: hehe :D

PR #403, opened 2016-02-01
fix warning in #find_files[_or_error]

srenatus:
This should shed some lights on the errors reported here: https://github.com/chef/kitchen-inspec/issues/37


chris-rock:
Thanks for fixing the warning @srenatus!

PR #418, opened 2016-02-04
ec2 name tag instances for easier cleanup

alexpop:


chris-rock:
thanks @alexpop this is great in addition to the test-kitchen flag

PR #419, opened 2016-02-05
upstart_service: fallback to config files if `show-config` is not available

### LINKED ISSUE #417: be_enabled for upstart service on RHEL 6 not working correctly
I'm running CentOS 6.5 with inspec <x>

I run uwsgi as an upstart service. I had the following inspec test:

```
describe upstart_service('uwsgi') do
    it { should be_installed }
    it { should be_enabled }
    it { should be_running }
end
```

```
Failures:

  1) Service uwsgi should be enabled
       expected that `Service uwsgi` is enabled
     # ./test/integration/default/mytest_spec.rb:11:in `block (2 levels) in load'
     # C:/Users/fletchowns/AppData/Local/chefdk/gem/ruby/2.1.0/gems/kitchen-inspec-0.10.1/lib/kitchen/verifier/inspec.b:43:in `call'
```

Apparently it uses [initctl show-config](https://github.com/chef/inspec/blob/07936e965dccd83d99673538142c59b2e67638cb/lib/resources/service.rb#L244) to do this check, but the RHEL 6 version of upstart doesn't seem to support that:

```
# initctl show-config uwsgi
initctl: invalid command: show-config
Try `initctl --help' for more information.
# initctl version
init (upstart 0.6.5)
```

srenatus:
Fixes #417.

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -227,7 +227,7 @@ class Upstart < ServiceManager
 
     # @see: http://upstart.ubuntu.com/cookbook/#job-states
     # grep for running to indicate the service is there
-    running = !status.stdout[/running/].nil?
+    running = !status.stdout[%r{start/running}].nil?
 
     {
       name: service_name,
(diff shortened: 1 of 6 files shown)


lib/resources/service.rb
```diff
@@ -247,8 +247,12 @@ def info_enabled(status, service_name)
     # $ initctl show-config $job | grep -q "^  start on" && echo enabled || echo disabled
     # Ubuntu 10.04 show-config is not supported
     # @see http://manpages.ubuntu.com/manpages/maverick/man8/initctl.8.html
-    config = inspec.command("#{service_ctl} show-config #{service_name}")
-    enabled = !config.stdout[/^\s*start on/].nil?
+    cmd = inspec.command("#{service_ctl} show-config #{service_name}")
+
+    # if initctl doesn't know `show-config`, try config file as fallback
```
chris-rock: I am not 100% sure, but doesn't show-config also returns a non-zero exit code, when the service is not available. Maybe its better to check for the specific version?

chris-rock:
Thanks @srenatus for this improvement

PR #423, opened 2016-02-05
simplify profile ID handling [stale branch: kept for discussion, don't merge]

srenatus:
This suggests using the profile's metadata as the _single source of truth_ when it is present:
1. `inspec json PROFILE` will use the profile's name as profile ID: `linux`
2. `inspec exec` will use the profile's name if it is given a profile, if not, it depends on the `--id` CLI param.

Opinions? :smile:

### DIFF
--- a/bin/inspec
+++ b/bin/inspec
@@ -16,8 +16,6 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
     desc: 'Show diagnostics (versions, configurations)'
 
   desc 'json PATH', 'read all tests in PATH and generate a JSON summary'
-  option :id, type: :string,
-    desc: 'Attach a profile ID to all test results'
   option :output, aliases: :o, type: :string,
     desc: 'Save the created profile to a path'
   def json(path)
--- a/bin/inspec
+++ b/bin/inspec
@@ -99,8 +99,6 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
   end
 
   desc 'exec PATHS', 'run all test files at the specified PATH.'
-  option :id, type: :string,
-    desc: 'Prefix all example IDs with profile ID'
   target_options
   option :format, type: :string
   def exec(*tests)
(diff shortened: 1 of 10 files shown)


arlimus:
imho from a user's perspective:

```
inspec exec ... --id Bob
```

and results being prefixed with ID Alice feels to me like defying expectations. As a user, this would tremendously confuse me.

I would much rather prefer we get rid of `--id` and the need to set an ID manually altogether. Since it has been started as a mechanism for (now) Chef Compliance, as a stepping stone, we should be able to find a way to fix it on that side.

Which leaves: Unless we find a real use-case for `--id` I'd prefer we remove that.

srenatus:
The only remaining use case is using `inspec exec` on test files, not profiles. Does anyone do that? ;)

arlimus:
@srenatus afaics no, users have not used that feature in this way; I think we are fine with removing it then, and see if users come up with a use-case + issue. Would you agree?

srenatus:
As discussed with @jcreedcmu and @arlimus, we'll get rid of `--id`, and keep the _flat_ example IDs (i.e., `basic-1` instead of `linux/basic-1` or even `base/linux/basic-1`).

srenatus:
:woah: it eventually turned green... lgtm

aaronlippold:
The only use case that I can see is:

User A is working on linux/profile1
User B is working on linux/profile1

Both users upload their version of profile1 to Compliance server. 

In Compliance Server:

UserA/linux/profile1
UserB/linux/proifle1 

Given that we now just have a reference to exec for 'profile1' 

How could I execute both profiles?

This really is a namespace or profile management issue. Given that there will be:

1) General Profiles - open to all 
2) Group / Organization Profiles
3) User Profiles 
- I may be hacking on the 'linux' profile from 1 or 2 to make some adjustments and want to run it from the compliance server

Also, the 'on the ground' use of the compliance server may be that folks break it down like this:
1. General Profiles - open to all
   
   a. approved published - hitting prod systems
2. Group / Organization Profiles
   
   b. in test profiles - hitting prod and test systems
3. User Profiles
   
   c. in development profiles

The 'common name' - i.e. linux - may be overloaded given the 'scope' of the namespace that it was uploaded under.

Just something to keep in mind.

Aaron

adamleff:
Closing this conversation PR, but leaving the branch should it be needed.

PR #435, opened 2016-02-08
semantics: rename CLI plugins registry -> commands

arlimus:
Basically make sure everyone understands these are only subcommands. we might consider adding plugins for options or existing commands instead of new subcommands. this just ensures everyone knows what registry is for

### DIFF
--- a/bin/inspec
+++ b/bin/inspec
@@ -147,7 +147,7 @@ ctl = Inspec::PluginCtl.new
 ctl.list.each { |x| ctl.load(x) }
 
 # load CLI plugins before the Inspec CLI has been started
-Inspec::Plugins::CLI.registry.each { |_subcommand, params|
+Inspec::Plugins::CLI.subcommands.each { |_subcommand, params|
   Inspec::InspecCLI.register(
     params[:klass],
     params[:subcommand_name],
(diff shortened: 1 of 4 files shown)


chef-supermarket:

* @arlimus

arlimus:
@chris-rock not sure if in or out, what do you think?

bin/inspec
```diff
@@ -143,7 +143,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
 end
 
 # load CLI plugins before the Inspec CLI has been started
-Inspec::Plugins::CLI.registry.each { |_subcommand, params|
+Inspec::Plugins::CLI.commands.each { |_subcommand, params|
```
chris-rock: maybe we should use `subcommands` to be very explicit

chris-rock:
+1 i like it, since this solves some confusion

arlimus:
Thank you for your feedback, I like `subcommand` :+1:

test/unit/plugin_test.rb
```diff
@@ -18,11 +18,11 @@
     before do
       # TODO: remove this, once the plugin loading is not based on ruby-load time
       # initialization
-      cli_reg.registry.clear
+      cli_reg.subcommands.clear
```
arlimus, line 20: btw: i don't think these lines go away with ruby-load time initialization (which we changed in https://github.com/chef/inspec/pull/434). e.g. if the second test (which adds subcommand) runs before the first one, the command is still in the registry.

imho: this line must stay in because right now we are dealing with a global singleton registry.

PR #442, opened 2016-02-09
Fix iptables on CentOS6 + more tests for iptables (plus small code improvements)

srenatus:
- unit tests: +1 test case including a comment
- integration tests: haven't been there before

### DIFF
--- a/lib/resources/iptables.rb
+++ b/lib/resources/iptables.rb
@@ -59,7 +59,7 @@ class IpTables < Inspec.resource(1)
     return [] if cmd.exit_status.to_i != 0
 
     # split rules, returns array or rules
-    @iptables_cache = cmd.stdout.chomp.split("\n")
+    @iptables_cache = cmd.stdout.split("\n").map(&:strip)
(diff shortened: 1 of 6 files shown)


lib/resources/iptables.rb
```diff
@@ -43,29 +43,26 @@ def initialize(params = {})
...
-    }
-    found
+    # checks if the rule is part of the ruleset
+    # for now, we expect an exact match
+    retrieve_rules.any? { |line| line.casecmp(rule) == 0 }
   end
 
   def retrieve_rules
-    return @iptables_cache if defined?(@iptables_cache)
+    return @iptables_cache if @iptables_cache
```
arlimus: although ruby implicitly does the right thing, using `defined?` on a variable that has not been defined is better at expressing why this line is used

srenatus: ok reverted :)

arlimus:
Awesome improvements, cheers @srenatus :+1: :) !

PR #452, opened 2016-02-12
Fix two minor logging and config bugs in CLI

srenatus:


arlimus:
Cheers @srenatus :+1:

PR #458, opened 2016-02-16
update targets

chris-rock:
- fix url helper
- update chef compliance and supermarket plugin
- move core to plugins/target


lib/bundles/inspec-compliance/target.rb
```diff
@@ -39,11 +39,11 @@ def resolve(target, opts = {})
...
-    def get_profile_name(uri)
+    def self.get_profile_name(uri)
       uri.host + uri.path
     end
 
-    def build_target_url(target)
+    def self.build_target_url(target)
       owner, profile = target.split('/')
       config = Compliance::Configuration.new
       url = "#{config['server']}/owners/%owner_name%/compliance/%profile_name%/tar"
```
srenatus, line 49: This is a little weird. Why not just this (instead of line 49-52)

`"#{config['server']}/owners/#{owner}/compliance/#{profile}/tar"`?

I know this isn't part of the change set but fixing it here is as good a time as any ;)

arlimus:
I like the separation into a more plugin-aware structure, that's fantastic! The current system (I introduced) is really confusing.

imho DirHelper is not the right base class for UrlHelper. Looking at the code: It hands everything off to zip or tar anyway, i.e. there is no re-use of DirHelper functionality, except for tricky indirection through zip/tar-helpers; the disadvantage is that it forces you to write every other piece that builds on UrlHelper (like Supermarket and Compliance) in a way that requires class-methods (i.e. another bucket of code-changes).
Worse, the indirection of Zip/Tar-helpers is really hard to follow (took me a while to understand what was wrong in the first place).

I'll update this in a bit...

lib/bundles/inspec-compliance/target.rb
arlimus, line 49: +1 great idea for clarity

chris-rock, line 49: like it

chris-rock:
close in favor for #462

Issue #353, opened 2016-01-08
Sharing custom resources?


arothian:
I see that you can add your custom resources to a libraries folder within a profile, but is there a way to include custom resources provided by another gem? It seems like you should be able to build up a library of custom resources, publish as a gem, and then include in a variety of inspec situations.

chris-rock:
@arothian At this moment, InSpec does not support this addition of global resource from outside of the InSpec gem. We are thinking about an extension system. Could you share your use case?

arothian:
@chris-rock thanks for the reply. I realize this is outside the original design/use for inspec, but I'd like to be able to develop resources that interact with AWS infrastructure to do some validation/testing. For example, it would be nice to be able to write controls that target AWS resources such as security groups, IAM roles/policies, VPC routing, etc. These inspec resources would just run with the local executor and interact with the AWS sdk. It would be nice to package these as a 'inspec-aws-resources' gem, since I doubt everyone would find them useful.

I'm thinking that from a testing and audit perspective, some of these resources are just as important as a vm/container.

arothian:
@chris-rock So I dug into the code a little but and discovered a working method. If you define a file in your libraries folder of a profile, you can use `require 'yourresourcesgem'` to load a gem that just has other inspec resources. This seems to work just fine. Not sure if you want to close this out or document the 'feature'.

chris-rock:
@arothian Thanks for sharing the use-case. We are currently thinking about a proper plugin system, that would support your use case. One core design goal is to make make profiles portable. 

Your approach is great, but requires the gem to be available on the running machine beforehand, which may leads to problems. Therefore we cannot support your solution officially yet.

I mark your issue as a feature request. Would love to see your inspec resources!

chris-rock:
The only official solution that we offer at this moment is via InSpec profiles: https://github.com/chef/inspec/blob/master/docs/profiles.rst 

Therefore you can write your own resources as part of InSpec profiles

arothian:
@chris-rock This is very much still a work in progress, but some of my initial resources are up at https://github.com/arothian/inspec-aws

chris-rock:
@arothian that looks really good. we work on a solution that will integrate much better. Stay tuned.

pburkholder:
@chris-rock Any update here on how to do tests against the platform, or how to cleanly include something like `inspec-aws`?

arlimus:
[...]

PR #460, opened 2016-02-17
Add a require_plugin method and the plumbing to support it.

### LINKED ISSUE #459: No way to reload the add resources from test code
Currently the DSL is baked in to an anonymous module at the start of test execution. This means that if you require a gem that adds new resource types, they aren't visible. This makes packaging custom resources strangely difficult given that it is a defining feature of InSpec.

coderanger:
This adds a new `require_plugin` method to the test execution context which allows pulling in custom resources from a gem.

Fixes #459.

### DIFF
--- a/lib/inspec/resource.rb
+++ b/lib/inspec/resource.rb
@@ -21,13 +21,45 @@ module Inspec
     def self.create_dsl(backend)
       # need the local name, to use it in the module creation further down
       my_registry = registry
-      Module.new do
-        my_registry.each do |id, r|
-          define_method id.to_sym do |*args|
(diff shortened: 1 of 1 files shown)


chris-rock:
@coderanger thank you very much for you PR. We see your need to extend InSpec with custom resources as highlighted in #353, too. We will have a look at it and provide you more feedback over the next days.

chris-rock:
Thank you @coderanger for raising the issue for reuse early on. We implemented profile inheritance now, which allows you to share resources and reuse controls. Please let us know if you need anything else.

PR #461, opened 2016-02-17
ensure permissions of inspec-compliance config.json on store

srenatus:
The default umask might allow for reading the authentication token by anyone. It will ensure that it's only readable/writable by the user (`0600`) now.

NB there's no measure to _not read a world-writable config.json_ on `get`.


srenatus:
Ah sh... I'll kill that experiment commit.

chris-rock:
Awesome. Thanks for this fix @srenatus

PR #468, opened 2016-02-18
Move integration tests to test/integration

chris-rock:
- move the test location to its natual location
- fix some mount errors, where we could not create empty.iso in /root


arlimus:
Much simpler, less folder searching! :+1: Thank you @chris-rock :smile:

PR #471, opened 2016-02-19
add shadow resource for /etc/shadow

arlimus:


chef-supermarket:

* @arlimus

lib/resources/shadow.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+  "
+
+  attr_reader :params
+  attr_reader :content
+  attr_reader :lines
+
+  def initialize(path = nil, opts = nil)
+    opts ||= {}
+    @path = path || '/etc/shadow'
```
srenatus, line 41: nitpicking, but how about

``` ruby
def initialize(path = '/etc/shadow', opts = {})
```

lib/resources/shadow.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def filter(hm = {})
+    return self if hm.nil? || hm.empty?
+    res = @params
+    filters = ''
+    hm.each do |attr, condition|
+      condition = condition.to_s if condition.is_a? Integer
+      filters += " #{attr} = #{condition.inspect}"
+      res = res.find_all do |line|
+        case line[attr.to_s]
```
srenatus, line 56: how about

`line[attr.to_s] == condition`

instead of the case stmt?

lib/resources/shadow.rb
```diff
@@ -0,0 +1,130 @@
...
+    filter_by.nil? ? map_data('expiry_date') : filter(expiry_date: filter_by)
+  end
+
+  def to_s
+    f = @filters.empty? ? '' : ' with'+@filters
+    "/etc/shadow#{f}"
+  end
+
+  def count
+    @params.length
```
srenatus: how about `def_delegator` with a 3rd arg (alias)?

`def_delegator :@params, :length, :count`

srenatus:
@arlimus: great resource. Went a little bonkers with non-important code style comments ;)

LGTM :+1:

lib/resources/shadow.rb
```diff
@@ -0,0 +1,130 @@
+# encoding: utf-8
+# copyright: 2015, Chef Software Inc.
```
srenatus: 2016 :tada:

lib/resources/shadow.rb
arlimus, line 56: I need to have the condition check in case of e.g. regex; case is great for those kind of statements

lib/resources/shadow.rb
arlimus: ups :)

lib/resources/shadow.rb
srenatus, line 56: Oh! I wasn't aware of that trick. That's cool. :smiley:

lib/resources/shadow.rb
```diff
@@ -0,0 +1,132 @@
...
+    end
+  "
+
+  attr_reader :params
+  attr_reader :content
+  attr_reader :lines
+
+  def initialize(path = '/etc/shadow', opts = nil)
+    opts ||= {}
+    @path = path || '/etc/shadow'
```
srenatus, line 41: This is redundant now, isn't it?
`@path = path`

chris-rock:
That looks really good. I like to see filter everywhere now.

lib/resources/shadow.rb
arlimus, line 41: Almost: what if `shadow(nil)`. It will not happen intentionally, but may happen by accident. I'm not sure yet how we want to handle this case though. Throwing a ruby fail seems too harsh, maybe improve this in the next iteration. It is true for a lot of resources...

srenatus, line 41: Ok, another edge case I didn't notice. Thanks.

My original comment then doesn't make too much sense then (i.e. the case against `path=nil`)...

chris-rock:
Thanks @arlimus

PR #473, opened 2016-02-22
replace targets with fetcher+reader system

arlimus:
Fully removes the current target system, which typically worked this way:

```
target ---> resolve( ) --->  [ metadata, ruby_tests, ruby_libs ]
```

Although incredibly convenient, this system was convoluted and hard to understand, as it did 2 things at once: Fetch the target you are interested in and then understand it (as either an InSpec profile or flat Ruby files, though there was some early support for Serverspec folders).

The new system instead works like this:

```
#1: get raw contents; e.g. file, folder, zip, tar, url, ...
target ---> fetch( ) ---> Fetcher[target]
                          => files(), read(file)

# optional: find the root folder
Fetcher[target] ---> relative_target() --->  Fetcher[target@root_dir]

#2: understand contents; e.g. inspec, flat, scap, ...
Fetcher[target@root_dir] ---> read_source( ) --->  SourceReader[Fetcher[target]]
                                                   => metadata(), tests(), libraries()
```

What appears to be more complex, now fully exposes all steps and clearly defines them. It provides explicit access to many elements of that chain, that were so far hidden, and led to a lot of code-duplication (e.g. source-paths, especially when chaining fetchers; relative-to-root vs absolute paths; raw contents instead of just converted ruby code).

Mainly targets bugs with InSpec's `json` and `check` commands and anything that uses the profile class to work.


chef-supermarket:

* @arlimus

bin/inspec
```diff
@@ -115,11 +115,13 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
...
 
-    runner = Inspec::Runner.new(opts)
     rel = File.join(File.dirname(__FILE__), *%w{.. lib utils detect.rb})
     detect_util = File.expand_path(rel)
-    runner.add_tests([detect_util])
-    runner.run
+    # exits on execution:
+    runner = Inspec::Runner.new(opts)
+    profile = Inspec::Profile.for_target(detect_util, opts)
+    runner.add_profile(profile)
```
chris-rock, line 123: I like that we pass a profile to runners now. We may discuss if we change the direction here as well, but not as part of this PR. 

```
profile = Inspec::Profile.for_target(detect_util, opts)
profile.run
```

lib/bundles/inspec-compliance/target.rb
```diff
@@ -3,56 +3,43 @@
...
 
 # InSpec Target Helper for Chef Compliance
 # reuses UrlHelper, but it knows the target server and the access token already
 # similar to `inspec exec http://localhost:2134/owners/%base%/compliance/%ssh%/tar --user %token%`
 module Compliance
-  class ChefComplianceHelper < Inspec::Targets::UrlHelper
-    def handles?(target)
+  class Fetcher < Fetchers::Url
+    name 'compliance'
+    priority 500
```
chris-rock, line 15: +1

lib/bundles/inspec-supermarket/target.rb
```diff
@@ -3,30 +3,26 @@
...
-
-    def resolve(target, opts = {})
-      # extract the tool url
+    def self.resolve(target, opts = {})
+      return nil unless URI(target).scheme == 'supermarket'
+      return nil unless Supermarket::API.exist?(target)
       tool_info = Supermarket::API.find(target)
       super(tool_info['tool_source_url'], opts)
+    rescue URI::Error => _e
+      false
```
chris-rock: we should return `nil` here

lib/fetchers/tar.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    def read(file)
+      @contents[file] ||= read_from_tar(file)
+    end
+
+    def read_from_tar(file)
+      return nil unless @files.include?(file)
+      res = nil
+      Gem::Package::TarReader.new(Zlib::GzipReader.open(@target)) do |tar|
+        tar.each do |entry|
```
srenatus, line 39: This might be more explicit like this:

``` ruby
res = tar.find do |entry|
  entry.file? && file == entry.full_name
end

res.read
```

lib/fetchers/zip.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+
+    def read(file)
+      @contents[file] ||= read_from_zip(file)
+    end
+
+    def read_from_zip(file)
+      return nil unless @files.include?(file)
+      res = nil
```
srenatus, line 36: +1 for clarity, but this initialisation is redundant (thanks ruby):

``` interactive
[1] pry(main)> if false
[1] pry(main)*   foo = :bar
[1] pry(main)* end  
=> nil
[2] pry(main)> foo
=> nil
```

lib/inspec/plugins/fetcher.rb
```diff
@@ -0,0 +1,97 @@
...
+      def get_files_prefix(files)
+        return '' if files.empty?
+
+        file = files[0]
+        bn = File.basename(file)
+        # no more prefixes
+        return '' if bn == file
+
+        i = file.rindex(bn)
+        prefix = file[0..i-1]
```
srenatus: This shadows the `attr_reader :prefix`

lib/inspec/plugins/fetcher.rb
```diff
@@ -0,0 +1,97 @@
...
+        pre = files[0] + File::SEPARATOR
+        rest = files[1..-1]
+        if rest.all? { |i| i.start_with? pre }
+          return get_folder_prefix(rest, false)
+        end
+        return get_files_prefix(files) if first_iteration
+        files
+      end
+
+      def get_files_prefix(files)
```
srenatus: This shadows the `attr_reader :files`

test/unit/fetchers/local_test.rb
```diff
@@ -0,0 +1,65 @@
...
+    it 'must only contain this file' do
+      _(res.files).must_equal [__FILE__]
+    end
+
+    it 'must not read if the file doesnt exist' do
+      _(res.read(rand.to_s)).must_be_nil
+    end
+
+    it 'must not read files not covered' do
+      _(File.file?('/proc/cpuinfo')).must_equal true
```
srenatus: Will this work on Windows?

test/unit/fetchers/url_test.rb
```diff
@@ -0,0 +1,150 @@
...
+          @archive = File.new(__FILE__)
+        end
+      end
+    }
+
+    it 'handles a http url' do
+      url = 'http://chef.io.tar.gz'
+      res = fetcher.resolve(url)
+      _(res).must_be_kind_of Fetchers::Local
+      _(res.parent.target).must_equal 'http://chef.io.tar.gz'
```
srenatus: Isn't this actually a bug? I understand that intention, but `URI("http://chef.io.tar.gz").path == ""`

test/unit/profile_test.rb
```diff
@@ -69,14 +69,13 @@
...
       it 'prints loads of warnings' do
-        inspec_yml = "#{home}/mock/profiles/#{profile_id}/inspec.yml"
         logger.expect :info, nil, ["Checking profile in #{home}/mock/profiles/#{profile_id}"]
-        logger.expect :error, nil, ["Missing profile name in #{inspec_yml}"]
-        logger.expect :error, nil, ["Missing profile version in #{inspec_yml}"]
-        logger.expect :warn, nil, ["Missing profile title in #{inspec_yml}"]
-        logger.expect :warn, nil, ["Missing profile summary in #{inspec_yml}"]
-        logger.expect :warn, nil, ["Missing profile maintainer in #{inspec_yml}"]
-        logger.expect :warn, nil, ["Missing profile copyright in #{inspec_yml}"]
+        logger.expect :error, nil, ["Missing profile name in inspec.yml"]
```
srenatus, line 73: Mhm this is much better. Together, line 72 and 73 still give you all the information...

test/unit/fetchers/url_test.rb
chris-rock: This should be `_(res).must_be_kind_of Fetchers::Url`

chris-rock: The URL fetcher needs to return a url with tar.gz, Which is especially interesting if you put in a github url

test/unit/fetchers/local_test.rb
chris-rock: Lets fix that....

lib/fetchers/tar.rb
chris-rock, line 39: Looks like tar reader has no find method: http://ruby-doc.org/stdlib-1.9.3/libdoc/rubygems/rdoc/Gem/Package/TarReader.html

But still we could make it

```
tar.each do |entry|
  # only handle files for now
  next unless entry.file? || file == entry.full_name
  res = entry.read
end
```

lib/fetchers/zip.rb
chris-rock, line 36: yeah, lets remove `res=nil`

lib/fetchers/tar.rb
srenatus, line 39: Oh dear, I thought this was an instance of Enumerable or something similar. Appearances are deceiving...

lib/inspec/plugins/fetcher.rb
chris-rock: please find other local variables, to add clarity. Great catch.

test/unit/fetchers/local_test.rb
srenatus: Done

test/unit/fetchers/url_test.rb
srenatus: @chris-rock hm, curious. Doesn't fail with `Fetchers::Local`, though...

lib/fetchers/tar.rb
srenatus, line 39: Turns out it does [include Enumerable](http://ruby-doc.org/stdlib-2.0.0/libdoc/rubygems/rdoc/Gem/Package/TarReader.html), but only with Ruby 2.x

lib/fetchers/zip.rb
srenatus, line 36: Hmm turns out I was wrong. Not sure how exactly the code is different from my example (well, there're too many ways it differs...), leaving it as-is now.

srenatus:
@chris-rock LGTM.

bin/inspec
arlimus, line 123: Good point, also a curious one. The advantage of passing multiple things to runner is that it can add all things to a run and then execute them. If we have that execution, would we need Profile.run anymore? I guess it's still great for convenience...

lib/bundles/inspec-supermarket/target.rb
arlimus: +1

srenatus: Oops forgot that one.

lib/fetchers/zip.rb
arlimus, line 36: In all these cases: yes, ruby implicitly creates non-existent things as nil. However, I much prefer it this way because it provides clarity. If you don't do it, people might search for a variable name in e.g. attr_readers or method declarations => very confusing!

btw rubocop agrees with this, but only checks it on instance variables (not sure exactly why they don't do it on local vars).

arlimus:
Awesome review, thank you @srenatus  and @chris-rock !
+1

PR #481, opened 2016-02-22
0.14.1

chris-rock:


alexpop:
manual and local `./make upgrade_test` showed no issues for me

PR #499, opened 2016-02-26
add xinetd_conf resource

arlimus:
This clearly shows that we have to unify some of our internals:
- easy read-file with included skip
- easy filtering by conditions
- maybe extract the parser

### DIFF
--- a/lib/inspec/resource.rb
+++ b/lib/inspec/resource.rb
@@ -93,6 +93,7 @@ require 'resources/shadow'
 require 'resources/ssh_conf'
 require 'resources/user'
 require 'resources/windows_feature'
+require 'resources/xinetd'
 require 'resources/yum'
 
 # file formats, depend on json implementation
--- a/lib/resources/xinetd.rb
+++ b/lib/resources/xinetd.rb
@@ -1,0 +1,119 @@
+# encoding: utf-8
+# author: Christoph Hartmann
+# author: Dominik Richter
+
+require 'utils/parser'
+
+class XinetdConf < Inspec.resource(1)
+  name 'xinetd_conf'
+  desc 'Xinetd services configuration.'
+  example "
+    describe xinetd_conf.services('chargen') do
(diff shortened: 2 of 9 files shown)


chef-supermarket:

* @arlimus

srenatus:
Yup, I'm also unhappy with the duplication this introduces... in the small, there's five very similar methods that could be a `define_method` loop or just a `method_missing`... in the large, it's a copy-paste job of  [resources/shadow.rb](https://github.com/chef/inspec/blob/master/lib/resources/shadow.rb)... looking at that one, it could also use some meta programming...

arlimus:
About those 5 methods: I kicked that for now to directly expose what's happening there. I would love to teach source-code inspect to actually show the contents of these methods at some point....

srenatus:
...to auto-generate docs? (Not sure I follow)

arlimus:
So apparently it's just 3 methods, the rest has a tiny variation. Which gets back to the point: let's unify it separately :smile:

arlimus:
@srenatus It's more about pry's included `show-source` method. With dynamically defined methods it just cannot find the right method.

arlimus:
@srenatus of course with your idea we might expose all methods via auto-generated docs at some point? that would completely remove this problem and solve the issue that users cannot yet reliably see what each resource provides... yeah we need that :+1: :100:

srenatus:
@arlimus just tried to make sense of what you told me... `show-source` isn't part of my workflow, so I didn't have it on my radar.

chris-rock:
@arlimus @srenatus While I agree to reduce duplication, I also tend to avoid meta-programming for one reason: readability. It is really hard to understand the code if we use all mechanisms of ruby. Sometimes its better to write some more lines, but make the function obvious. So that does not mean, that we should not use e.g. `define_method`, but especially our resources should be easy readable for users and non ruby-experts.

chris-rock:
Thanks @arlimus

PR #503, opened 2016-02-28
Fix license warning during gem build.

juliandunn:
WARNING:  WARNING: license value 'Apache 2.0' is invalid.  Use a license identifier from
http://spdx.org/licenses or 'Nonstandard' for a nonstandard license.
WARNING:  See http://guides.rubygems.org/specification-reference/ for help


arlimus:
Thank you @juliandunn !

PR #508, opened 2016-03-01
Improve Supermarket CLI

alexpop:
Make PROFILE required and update usage info

### DIFF
--- a/lib/bundles/inspec-supermarket/cli.rb
+++ b/lib/bundles/inspec-supermarket/cli.rb
@@ -13,8 +13,7 @@ module Supermarket
 
       headline('Available profiles:')
       supermarket_profiles.each { |p|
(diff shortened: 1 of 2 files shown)


lib/bundles/inspec-supermarket/cli.rb
```diff
@@ -18,12 +18,12 @@ def profiles
       }
     end
 
-    desc 'exec PROFILE', 'executes a Supermarket profile'
+    desc 'exec PROFILE', 'execute a Supermarket profile'
     option :id, type: :string,
       desc: 'Attach a profile ID to all test results'
     target_options
     option :format, type: :string
-    def exec(*tests)
```
chris-rock, line 26: we need the \* to get all parameters in one array

chris-rock:
This PR improves the following:
- shows an error if the profile does not exists for `inspec supermarket info profilename`
- requires the `profile` parameter for `info` and `exec`

chris-rock:
Thanks @alexpop this was a great catch!

PR #512, opened 2016-03-01
adds a insecure option 

### LINKED ISSUE #511: `compliance` command does not work with self-signed https
The solution is to introduce a `-k/--insecure.` option like curl

chris-rock:
This PR allows `inspec compliance profiles` to work with Chef Compliance servers that use self-signed ssl. During login, you have the `--insecure` or `-k` option.

`inspec compliance login https://yourserver/api --user=admin --password='admin' --insecure`

fixes #511


arlimus:
this will help for quick testing, thank you @chris-rock  :+1:

PR #526, opened 2016-03-08
Placing all resources in the Inspec::Resources namespace

adamleff:
Many of the resources are named as a top-level class with a fairly generic class name, such as "OS". This causes an issue specifically with kitchen-google which depends on a gem which depends on the "os" gem which itself defines an OS class with a different superclass. This prevents users from using TK, Google Compute, and Inspec without this fix.

Since our resource class names aren't unique to just Inspec, I'm moving all resource classes in the `Inspec::Resources` module to hopefully avoid this issue in the future. Since the class name is not used in finding a resource (because of the resource registry), this change should be a no-op to our users.

Some mocked commands had their digest changed as well due to the new indentation, specifically in the User and RegistryKey classes.

I strongly recommend viewing this diff with `git diff --ignore-space-change`
to see the _real_ changes. :)

### DIFF
--- a/lib/resources/apache.rb
+++ b/lib/resources/apache.rb
@@ -4,26 +4,28 @@
 # author: Dominik Richter
 # license: All rights reserved
 
-class Apache < Inspec.resource(1)
-  name 'apache'
+module Inspec::Resources
+  class Apache < Inspec.resource(1)
+    name 'apache'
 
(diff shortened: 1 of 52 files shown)


arlimus:
Thank you so much @adamleff !! This is a stern reminder how much code there's in resources already :D (the diff says it all). Great improvement, kudos :+1:

PR #527, opened 2016-03-09
Rename internal File and OS resource classes

arlimus:
This is a follow-up to https://github.com/chef/inspec/pull/526 by @adamleff . Although it already prevents clashes on the global plane, I still think it's easy enough to accidentally call `File` inside a new resource and instead of getting `::File` you get access to inspec's internal resource - which leads to potentially frustrating debugging. So as an additional layer to the MR ^^, we rename these resources and re-expose simple `File` calls, i.e. detect any problems early on.

### DIFF
--- a/lib/resources/apache.rb
+++ b/lib/resources/apache.rb
@@ -14,12 +14,12 @@ module Inspec::Resources
       when 'ubuntu', 'debian'
         @service = 'apache2'
         @conf_dir = '/etc/apache2/'
-        @conf_path = ::File.join @conf_dir, 'apache2.conf'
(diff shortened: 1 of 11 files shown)


chef-supermarket:

* @arlimus

chris-rock:
This is a great simplification for developers. Thanks @arlimus

adamleff:
I had contemplated doing this but wasn't sure if it would be OK... and thought then maybe we should rename EVERYTHING to be consistent.  Such a slippery slope. :smile: 

Thanks, @arlimus!

PR #529, opened 2016-03-09
add output stream to rspec configuration

vjeffrey:
@arlimus


arlimus:
Thank you @vjeffrey !
Only reachable via internal configuration for now, not yet exposed as an option.

arlimus:
Only used internally for now, but I suspect this will be great to expose!

PR #543, opened 2016-03-16
Add title, description, code, and source_location to example metadata

vjeffrey:
screen shot 2016-03-17 at 9 40 22 am

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -22,6 +22,8 @@ module RSpec::Core::Formatters
         pending_message: example.execution_result.pending_message,
         id: example.metadata[:id],
         impact: example.metadata[:impact],
+        title: example.metadata[:title],
+        desc: example.metadata[:desc],
       }
     end
   end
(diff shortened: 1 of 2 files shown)


lib/inspec/rspec_json_formatter.rb
```diff
@@ -22,6 +22,10 @@ def format_example(example)
         pending_message: example.execution_result.pending_message,
         id: example.metadata[:id],
         impact: example.metadata[:impact],
+        title: example.metadata[:title],
+        descr: example.metadata[:desc],
```
arlimus: Could we use `:desc` instead of `:descr`?

chris-rock:
@vjeffrey +1 I like that implementation. Could we cover the `title`, `desc` in a separate object:

```
{
    ...
   metadata : {
      title: "abc",
      desc: ""
  }
}
```

arlimus:
Great point @chris-rock , let's move to a clean separate JSON format that combines inspec's fields and rspec's data to a proper report structure.

arlimus:
For now: flatten it out until we have another json reporter. Please don't use these fields here, they will go away!

arlimus:
Please not that this has been superseded by https://github.com/chef/inspec/pull/552

PR #553, opened 2016-03-18
rename `script` resource to `powershell` resource

chris-rock:
```
script = <<-EOH
# you powershell script
EOH

# old syntax is using `script` resource
describe script(script) do
  its('stdout') { should eq 'output' }
end

# new syntax is using `powershell` resource
describe powershell(script) do
  its('stdout') { should eq 'output' }
end
```


arlimus:
Much clearer and less confusing, fantastic improvement!
Let's get script deprecated for 1.0 :tada:

PR #560, opened 2016-03-19
add `wmi` resource

chris-rock:
The new WMI resource allows users to query Windows Management Instrumentation (WMI). Under the hood we use Powershell to retrieve the information:

```
# Get-WmiObject -class win32_service
# returns an array of results
describe wmi('win32_service') do
  its(['Path','ClassName']) { should include 'Win32_Service' }
  its('DisplayName') { should include 'Windows Remote Management (WS-Management)'}
end

# Its also possible to add filter to narrow down the result set
# therefore we can use `eq` instead of `include` here
describe wmi('win32_service', {
  filter: "name like '%winrm%'"
}) do
  its(['Path','ClassName']) { should eq 'Win32_Service' }
  its('DisplayName') { should eq 'Windows Remote Management (WS-Management)'}
end
```

### DIFF
--- a/lib/utils/object_traversal.rb
+++ b/lib/utils/object_traversal.rb
@@ -1,0 +1,23 @@
+# encoding: utf-8
+# author: Dominik Richter
+# author: Christoph Hartmann
+module ObjectTraverser
+  def extract_value(keys, value)
+    key = keys.shift
+    return nil if key.nil?
+
+    # if value is an array, iterate over each child
+    if value.is_a?(Array)
+      value = value.map { |i|
+        extract_value([key], i)
+      }
+    else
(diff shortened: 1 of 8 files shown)


lib/utils/object_traversal.rb
```diff
@@ -0,0 +1,23 @@
...
+    if value.is_a?(Array)
+      value = value.map { |i|
+        extract_value([key], i)
+      }
+    else
+      value = value[key.to_s].nil? ? nil : value[key.to_s]
+    end
+
+    # if there are no more keys, just return the value
+    return value if keys.first.nil?
```
arlimus, line 19: Isn't this redundant?

arlimus:
Awesome to get WMI for Windows, thank you @chris-rock !!

PR #576, opened 2016-03-23
inspec-compliance + Compliance 1.0

srenatus:
This change includes the necessary adaptations for having inspec's compliance plugin work with compliance 1.0.

### DIFF
--- a/lib/bundles/inspec-compliance/README.md
+++ b/lib/bundles/inspec-compliance/README.md
@@ -8,7 +8,8 @@ This extensions offers the following features:
 
 To use the CLI, this InSpec add-on adds the following commands:
 
- * `$ inspec compliance login user password` - authentication against Chef Compliance
+ * `$ inspec compliance api_token server --token TOKEN --user USER` - save the Chef Compliance API token for user
+ * `$ inspec compliance login` - authentication of the API token against Chef Compliance
  * `$ inspec compliance profiles` - list all available Chef Compliance profiles
  * `$ inspec compliance exec profile` - runs a Chef Compliance profile
  * `$ inspec compliance upload path/to/local/profile` - uploads a local profile to Chef Compliance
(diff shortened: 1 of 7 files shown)


chris-rock:
We need to ensure that it also works with older versions. This extension does not work with versions pre-1.0

lib/bundles/inspec-compliance/api.rb
```diff
@@ -9,35 +9,58 @@ module Compliance
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
-    # logs into the server, retrieves a token and stores it locally
-    def self.login(server, username, password, insecure)
+    # saves the api token supplied by the user
+    def self.api_token(server, refresh_token, verify, user, insecure)
```
chris-rock: we should still be able to input username and password

lib/bundles/inspec-compliance/README.md
```diff
@@ -8,7 +8,8 @@ This extensions offers the following features:
 
 To use the CLI, this InSpec add-on adds the following commands:
 
- * `$ inspec compliance login user password` - authentication against Chef Compliance
+ * `$ inspec compliance api_token server --token TOKEN --user USER` - save the Chef Compliance API token for user
```
chris-rock, line 11: I think this plugin should offer two methods
- handle the complete oauth flow to retrieve a token. that allows the plugin to get a refresh token as well (requires us to setup a specific client/application for inspec)
- take an api token as input

chris-rock:
At this point of time, this PR misses a couple of things:
- usability for login should be the same, cli should take --token, --refresh-token as option
- proper fallback for legacy auth (version not determined etc)
- simple login flow

chris-rock:
I add additions to this PRs to address this

srenatus:
Thanks @chris-rock for the additions! :+1:

Note that when https://github.com/chef/chef-compliance/pull/767 is merged, we should probably (re)introduce another variant.

PR #583, opened 2016-03-24
Improvements to gordon example and docs

alexpop:


arlimus:
much needed :heart: for these example, cheers @alexpop

PR #610, opened 2016-04-04
Extended gordon_config with more examples

alexpop:
Extended the example Inspec resource based on these two questions:

https://discourse.chef.io/t/inspec-profile-library-classes/8044/3
https://discourse.chef.io/t/inspec-use-returned-results-from-one-testcase-in-another-test-case/8079

Making the ruby code a bit more explicit for users with limited ruby/programming background.


examples/profile/libraries/gordon_config.rb
```diff
@@ -24,20 +25,29 @@ def initialize
...
+  # Example matcher for the number of commas in the file
+  def comma_count
+    text = @file.content
+    return text.count(',')
   end
 
   # Expose all parameters
   def method_missing(name)
-    @params[name.to_s]
+    return @params[name.to_s]
```
chris-rock, line 51: we do not need a return in ruby if its the last statement

chris-rock:
Great addition @alexpop

examples/profile/libraries/gordon_config.rb
alexpop, line 51: I added the returns for users that are not familiar with ruby and are used to see methods `return` something.

PR #625, opened 2016-04-08
Encourage sharing of profiles

nathenharvey:
Add a section about sharing and browsing profiles on the Supermarket.

Perhaps this will encourage more collaboration there.


chris-rock:
Awesome. Thanks Nathen

PR #637, opened 2016-04-13
Add 'static' value as enabled to systemd service enabled check

jmccann:
SystemD can have a service `enabled` not only when the `UnitFileState` is `running` but also when it is `static`.

To back this up: https://www.freedesktop.org/wiki/Software/systemd/dbus/

> UnitFileState encodes the install state of the unit file of FragmentPath. It currently knows the following states: enabled, enabled-runtime, linked, linked-runtime, masked, masked-runtime, static, disabled, invalid. enabled indicates that a unit file is permanently enabled. enable-runtime indicates the unit file is only temporarily enabled, and will no longer be enabled after a reboot (that means, it is enabled via /run symlinks, rather than /etc). linked indicates that a unit is linked into /etc permanently, linked indicates that a unit is linked into /run temporarily (until the next reboot). masked indicates that the unit file is masked permanently, masked-runtime indicates that it is only temporarily masked in /run, until the next reboot. **static indicates that the unit is statically enabled, i.e. always enabled and doesn't need to be enabled explicitly**. invalid indicates that it could not be determined whether the unit file is enabled.

Looks like there is other good info in that article as well ... including more states that could be set for `UnitFileState` such as: enabled, enabled-runtime, linked, linked-runtime, masked, masked-runtime, static, disabled, invalid. enabled

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -214,7 +214,7 @@ module Inspec::Resources
       running = params['SubState'] == 'running'
       # test via systemctl --quiet is-enabled
       # ActiveState values eg.g inactive, active
-      enabled = params['UnitFileState'] == 'enabled'
+      enabled = %w{enabled static}.include? params['UnitFileState']
 
       {
         name: params['Id'],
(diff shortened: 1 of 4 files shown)


chris-rock:
Thanks you very much @jmccann for this addition!

PR #645, opened 2016-04-16
WIP: Provide option to force bash for the command resource

alexpop:
At the moment, the `command` resource, through `train` and `Mixlib::shellout` executes local commands using `/bin/sh`.

Depending on the operating system, this can be `sh`, `dash`, `bash`. 
Some commands require `bash` to succeed, like this one from the Ubuntu 14.04 translated SCAP content:

``` ruby
 describe command("ui=($(echo 0022 -n | fold -w1));sys=($(stat -L --format=\"%a\" /etc/hosts.allow | awk '{printf \"%04d\\n\", $0;}' | fold -w1));for (( i=0; i<4; i++ )); do echo -n $(( ${ui[$i]} & ${sys[$i]})); done;").stdout.to_s.[](/^(\d+)$/, 1) do
    it { should eq "0000" }
  end
```

### DIFF
--- a/lib/resources/command.rb
+++ b/lib/resources/command.rb
@@ -18,13 +18,24 @@ module Inspec::Resources
     "
 
     attr_reader :command
+    attr_reader :options
 
-    def initialize(cmd)
+    def initialize(cmd, opts={})
       @command = cmd
+      if opts.is_a?(String)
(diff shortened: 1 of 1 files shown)


arlimus:
Moved to https://github.com/chef/inspec/pull/659

PR #649, opened 2016-04-16
add usage instructions for inspec container

chris-rock:


arlimus:
Great idea on using `alias` to make it simple! Kudos :+1:

PR #658, opened 2016-04-18
Add omnibus for inspec

chris-rock:
This is the first step to build packages for InSpec. The current PR includes all files required to build deb and rpm packages.


srenatus:
[Inspec is already part of `omnibus-software`](https://github.com/chef/omnibus-software/blob/master/config/software/inspec.rb), did you not use that for a specific reason?

chris-rock:
@srenatus No, that was not left out intentionally. I just figured out that omnibus software uses bundler, which is not required at all. So I harmonize both, so that we can depend on `inspec` defined in `omnibus-software`

omnibus/config/software/inspec.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+# Copyright 2015 Chef Software, Inc.
+#
+# All Rights Reserved.
+#
+
+name 'inspec'
+
+dependency 'ruby'
+dependency 'train'
```
chris-rock: no need to define train, since it is defined as gem dependency

omnibus/config/software/inspec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+license "Apache-2.0"
+license_file "LICENSE"
+
+default_version "v0.18.0"
```
chris-rock: This should point to the version that we've checked out

srenatus: [This is how chef does it](https://github.com/chef/chef/blob/master/omnibus/config/projects/chef.rb#L25-L27), but I suppose we could just `require_relative` lib/inspec/version.rb and use that...

chris-rock: The challenge is, that I want to be able to built exactly the package for the sha I've checked out

srenatus: I think this is addressed here: https://github.com/chef/omnibus-software/blob/master/config/software/chef.rb#L22-L35

omnibus/resources/inspec/deb/conffiles.erb
```diff
@@ -0,0 +1,3 @@
+<% config_files.each do |file| -%>
```
stevendanna: This and the other `deb` and `rpm` resource files appear identical to the default versions that omnibus will use.  Perhaps we can remove them from here?

chris-rock: Go ahead, makes sense from my perspective.

omnibus/config/software/inspec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  # Remove existing built gems in case they exist in the current dir
+  delete "#{name}-*.gem"
+
+  gem "build #{name}.gemspec", env: env
+  gem "install #{name}-*.gem --no-document", env: env
+
+  # Dependecy added to avoid this pry error:
+  # "Sorry, you can't use Pry without Readline or a compatible library."
+  gem "install rb-readline --no-document", env: env
```
stevendanna: If this is required, should we put it in our gemspec instead?

stevendanna:
Ready for Review.  Two questions possible changes we could consider:
- Right now this appbundle's inspec.  We've prefered appbundling for other ruby applications because (1) it can help with load times, (2) it helps ensure that we are running with the dependencies we expect even if someone comes along an messes with our gemset.  However, this brings in bundler which adds about 4MB to the build. For more info about appbundler see: https://github.com/chef/appbundler
- We may want to check in omnibus/Gemfile.lock eventually.  Because of dependency list is small I've left it unlocked since we aren't likely to see a lot of upstream churn that breaks us.

omnibus/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+Then login to the instance and build the project as described in the Usage
+section:
+
+```shell
+$ bundle exec kitchen login ubuntu-1204
+# Now inside the kitchen vm
+$ sudo -i
+$ export PATH=/opt/omnibus-toolchain/bin:/opt/omnibus-toolchain/embedded/bin:$PATH
```
chris-rock: The `/opt/omnibus-toolchain/` is not available in my VM

chris-rock:
@stevendanna Awesome, works as expected!

chris-rock:
I think we should go with appbundler for now and leave the Gemfile.lock out of this PR

ksubrama:
If you're using appbundler, then you need bundler (to generate the Gemfile.lock), negating the need to write the custom inspec software definition.

omnibus/config/software/inspec.rb
```diff
@@ -0,0 +1,49 @@
...
+  # We bundle install to ensure the versions of gems we are going to
+  # appbundle-lock to are definitely installed
+  bundle 'install --without test integration tools maintenance', env: env
+
+  gem "build #{name}.gemspec", env: env
+  gem "install #{name}-*.gem --no-document", env: env
+
+  # Dependecy added to avoid this pry error: "Sorry, you can't use Pry
+  # without Readline or a compatible library."
+  gem 'install rb-readline --no-document', env: env
```
ksubrama, line 47: https://github.com/chef/omnibus-software/blob/master/config/software/rb-readline.rb takes care of that.

chris-rock, line 47: @ksubrama Can you open a PR to add this improvement?

omnibus/config/projects/inspec.rb
```diff
@@ -0,0 +1,37 @@
...
+name 'inspec'
+maintainer 'Chef Software, Inc <maintainers@chef.io>'
+homepage 'https://github.com/chef/inspec'
+
+license 'Apache-2.0'
+license_file '../LICENSE'
+
+# Defaults to C:/inspec on Windows
+# and /opt/inspec on all other platforms
+install_dir "#{default_root}/#{name}"
```
ksubrama, line 28: Can we prefer C:/opscode/inspec?  It is where we land all our other projects and yet another top level directory at root level is going to be a bit annoying.  Also, top level directories in a drive has special pitfalls when it comes to permission/inheritance.

omnibus/config/projects/inspec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+build_version Omnibus::BuildVersion.semver
+build_iteration 1
+
+dependency 'preparation'
+dependency 'inspec'
+
```
ksubrama, line 35: Need an entry for fast_msi and msi upgrade code for the msi packager.  Any random valid GUID should work.

omnibus/config/software/inspec.rb
ksubrama, line 47: Can do - I'm noting down a few other issues I noticed and will open a PR will all of the fixes.

omnibus/config/software/inspec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+dependency 'ruby'
+dependency 'bundler'
+dependency 'appbundler'
+
+license :project_license
+
+default_version "v#{Inspec::VERSION}"
+
+source path: "#{Omnibus::Config.project_root}/../"
```
ksubrama, line 30: Need to exclude bundle/vendor if we end up setting up a Jenkins pipeline - the pipeline tends to do a bundle install in the omnibus directory and we don't want to pick up those files.

PR #659, opened 2016-04-18
configure command execution shells to sh/bash/zsh

arlimus:
Run against docker container:

``` ruby
command('n=($(echo "123")); echo $n').stdout
=> ''
```

Now like this:

``` ruby
command('n=($(echo "123")); echo $n', shell: 'bash').stdout
=> '123\n'
```

Alternative idea for https://github.com/chef/inspec/pull/645
All kudos to @alexpop for raising + original implementation + @srenatus for his input.

### DIFF
--- a/lib/resources/command.rb
+++ b/lib/resources/command.rb
@@ -4,6 +4,8 @@
 # author: Christoph Hartmann
 # license: All rights reserved
 
+require 'shellwords'
+
 module Inspec::Resources
   class Cmd < Inspec.resource(1)
     name 'command'
@@ -17,14 +19,25 @@ module Inspec::Resources
       end
     "
 
+    SHELLS = {
+      'sh'   => ->(x, path = 'sh')   { path + ' -c ' + Shellwords.escape(x) },
+      'bash' => ->(x, path = 'bash') { path + ' -c ' + Shellwords.escape(x) },
+      'zsh'  => ->(x, path = 'zsh')  { path + ' -c ' + Shellwords.escape(x) },
+    }.freeze
+
     attr_reader :command
 
-    def initialize(cmd)
+    def initialize(cmd, opts = {})
       @command = cmd
+      unless opts.is_a?(Hash)
+        skip_resource "Called #{self} with invalid command options. See the resource help for valid examples."
+        opts = {}
+      end
+      @opts = opts
     end
 
     def result
-      @result ||= inspec.backend.run_command(@command)
+      @result ||= inspec.backend.run_command(wrap_cmd)
     end
 
     def stdout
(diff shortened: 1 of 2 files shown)


chef-supermarket:

* @arlimus

arlimus:
(Moved from https://github.com/chef/inspec/pull/645 )
Great discussion! 

This implementation supports a few shells and mainly works via a wrapper; Easy to extend if we wanted to add shell-paths as suggested by @srenatus (though I would separate shell-type and shell-path in that case ;) ):

```
command('echo hello', shell: 'bash', path: '/bin/bash')

sh = ->(cmd) { 'bash -c ' + Shellwords.escape(x) }
command('echo hello', wrapper: sh)
```

### resource

An alternative idea to all this would be to add a `bash(...)` resource for calling this one command.

```
bash('echo world', path: '/bin/bash') ...
```

Though i think it gets us into the domain of writing scripts, not just commands...

### wrapper

Another idea is to initialize the shell once:

```
bash = command.shell('bash')
describe bash(...) ...

bash = command.shell('bash', path: '/bin/bash/or/somewhere/else')

bash = command.wrapper { |cmd| 'bash -c ' + Shellwords.escape(cmd) }
```

arlimus:
@chris-rock ping to our discussion this morning ^^

chris-rock:
I would like to harmonize usability with our `service` resource. Therefore we enable all the options to the command resource (wrapper idea), but combine it with the `bash` resource.

```
class Bash < Cmd
    name 'bash'

    def initialize(cmd, opts = {})
     opts['shell'] = opts['shell'] || 'bash'
     opts['path'] = opts['path'] || '/bin/bash'
     super(cmd, opts)
    end
  end
end
```

I am not sure if we should move so many unix/linux handling into cmd implementation.I would like to keep the `command` as simple as possible and move all the specific os handling to another resource (mabe unix/linux shell). Maybe we could establish the general wrapper logic, and leave the specific parameters to the inherited resources.

arlimus:
Apparently `bash -c` is enough to run streamed scripts. Now this doesn't read sha-bang calls to other interpreters or call options, which is a limitation from expectations...

srenatus:
draft-level nitpicking ;) 

```
  def initialize(cmd, opts = {})
     opts['shell'] ||= 'bash'
     opts['path'] ||= '/bin/bash'
     super # same params are passed by default
  end
```

arlimus:
@chris-rock :+1: on that. In this case, `bash` and others would become the norm. Maybe `command_wrapper` or just without exposing the abstract resource and instead only implementing `bash`/...? This way users can always write their own wrappers:

``` ruby
bash = ->(x) { command("bash -c "+Shellwords.escape(x)) }
```

I really like your idea of moving that logic outside of the `command` resource.

chris-rock:
@arlimus We could use `unix_command` and use the logic as you proposed.

arlimus:
Wouldn't that limit us? I.e. what if there are non-unix systems which we just wanted to wrap?
Would we still go with `bash` resources in that case?

```
unix_command('...', shell: 'bash')
# or
bash('...')
```

chris-rock:
Oh yeah, the wrapper logic should be in `cmd`. As long as we get the shell definition out of the core command implementation, I am fine.

PR #673, opened 2016-04-20
Fix indenting for translated loops and ors

alexpop:
Translated tests like these:

``` ruby
control "uno" do
  shadow.users(/.*/).entries.each do |entry|
    describe entry do
    its(:passwords) { should match /.+/ }
  end
  end
end
```

``` ruby
control "due" do
  describe.one do
    describe xinetd_conf.services("chargen").socket_types("dgram") do
    it { should be disabled  }
  end
    describe package("xinetd") do
    it { should_not be_installed  }
  end
end
```

become after this code change:

``` ruby
control "uno" do
  shadow.users(/.*/).entries.each do |entry|
    describe entry do
      its(:passwords) { should match /.+/ }
    end
  end
end
```

``` ruby
control "due" do
  describe.one do
    describe xinetd_conf.services("chargen").socket_types("dgram") do
      it { should be disabled  }
    end
    describe package("xinetd") do
      it { should_not be_installed  }
    end
  end
end
```

### DIFF
--- a/lib/inspec/objects/each_loop.rb
+++ b/lib/inspec/objects/each_loop.rb
@@ -25,7 +25,7 @@ module Inspec
 
     def to_ruby
       obj = super
-      all_tests = @tests.map(&:to_ruby).join("\n").gsub("\n","\n  ")
+      all_tests = @tests.map(&:to_ruby).join("\n").gsub("\n", "\n  ")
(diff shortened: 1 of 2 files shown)


arlimus:
Kudos @alexpop for fixing this :+1:

PR #682, opened 2016-04-26
Hpux

Anirudh-Gupta:
Added file permission check on hp-ux. The command has to be enclosed in quotes to pickup flags in hp-ux. Not sure about AIX. If its true for AIX as well, we can merge both of them in the same elsif statement.


chris-rock:
Super addition @Anirudh-Gupta

PR #696, opened 2016-04-29
show error if user is not logged in to compliance server

chris-rock:


arlimus:
Awesome ux improvement! kudos

PR #701, opened 2016-04-29
Add basename parameter and add tests

alexpop:
Working on the tests for this PR, I've noticed this train error for Windows:

``` ruby
inspec> f=file('c:\Users\vagrant')
ArgumentError: wrong number of arguments (2 for 3)
from /Users/apop/.chefdk/gem/ruby/2.1.0/gems/r-train-0.11.1/lib/train/extras/file_windows.rb:14:in `initialize'
```


arlimus:
Great catch, thank you @alexpop !

PR #703, opened 2016-04-30
fix: kitchen-inspec cannot use supermarket

### LINKED ISSUE #80: Cannot run supermarket and compliance profiles
kitchen-inspec does not load supermarket urls

```
suites:
  - name: default
    run_list:
      - recipe[prepare]
    verifier:
      inspec_tests:
        - supermarket://hardening/ssh-hardening
```

jeremymv2:
This fixes: https://github.com/chef/kitchen-inspec/issues/80

Without this change one cannot use a Supermarket hosted Profile in Test Kitchen.
The following results:

```
suites:
  - name: default
    run_list:
      - recipe[test-ohai::default]
    attributes:
    verifier:
      inspec_tests:
        - supermarket://hardening/ssh-hardening
...

>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: Failed to complete #verify action: [Could not fetch inspec profile in "supermarket://hardening/ssh-hardening".]
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

This fix simply `include`s the "Cli" library which instantiates a new Plugin controller where the Supermarket bundle lives.  Without this the Supermarket and Compliance plugins are not loaded.

```
# Load all plugins on startup
ctl = Inspec::PluginCtl.new
```

### DIFF
--- a/lib/bundles/inspec-supermarket/target.rb
+++ b/lib/bundles/inspec-supermarket/target.rb
@@ -15,6 +15,7 @@ module Supermarket
     def self.resolve(target, opts = {})
       return nil unless URI(target).scheme == 'supermarket'
       return nil unless Supermarket::API.exist?(target)
+      puts "Target: #{target}"
       tool_info = Supermarket::API.find(target)
       super(tool_info['tool_source_url'], opts)
     rescue URI::Error => _e
--- a/lib/inspec.rb
+++ b/lib/inspec.rb
@@ -19,6 +19,7 @@ require 'inspec/runner'
 require 'inspec/shell'
 
 # all utils that may be required by plugins
+require 'inspec/cli'
 require 'utils/base_cli'
 require 'inspec/fetcher'
 require 'inspec/source_reader'


chris-rock:
Hi @jeremymv2 Thanks for the proposed fix. I am not sure about the way we load it. The reason why we haven't loaded the plugins automatically to separate plugins and cli from the core lib. Therefore we decided that plugins are better loaded in the cli. In our case here, I propose we train kitchen-inspec to load the supermarket and compliance plugin.

This shows some drawbacks in our current draft of the plugin design. It was centered around the cli, which it really shouldn't. As a short term solution, `kitchen-inspec` could load the required files without the cli. For `kitchen-inspec` only the target is relevant:

```
module Compliance
  autoload :Configuration, 'inspec-compliance/configuration'
  autoload :HTTP, 'inspec-compliance/http'
  autoload :Support, 'inspec-compliance/support'
  autoload :API, 'inspec-compliance/api'
end
require 'inspec-compliance/target'
```

Would you like to add a PR in `kitchen-inspec`?

chris-rock:
@jeremymv2 I am going to close your PR https://github.com/chef/inspec/pull/703 in favor for https://github.com/chef/kitchen-inspec/pull/84 The reason is that I do not want to load the plugins by default with the inspec lib. We intend to break out the bundles out of inspec core, once the plugin mechanism is more mature. Now the responsibility is at kitchen-inspec to load the plugin.

chris-rock:
Thanks for your hard work on writing up the PR @jeremymv2 This helped to discuss the implementation approach in our team a lot better.

PR #671, opened 2016-04-20
JSON formatter redesign

arlimus:
The current JSON output formatter follows rspec conventions:

```
{
  "version": "3.5.0.beta3",
  "examples": [
    {
      "description": "should be owned by \"root\"",
      "full_description": "File /bin/sh should be owned by \"root\"",
      "status": "passed",
      "file_path": "examples/profile/controls/meta.rb",
      "line_number": 31,
      "run_time": 0.017937247,
      "pending_message": null,
      "id": "ssh-1",
      "impact": 1
    },
...
  "summary": {
    "duration": 0.022208948,
    "example_count": 5,
    "failure_count": 0,
    "pending_count": 1
  },
  "summary_line": "5 examples, 0 failures, 1 pending"
```

We added `id` and `impact`, but the latter can be inferred from the profile's JSON representation retrieved via `inspec json`. This output is also focused on examples, i.e. tests within control blocks, which didn't help with controls in inspec.

Because there is a difference between integration-test driven by `describe`-blocks and more security-oriented `control`-blocks, we had added the `fulljson` formatter:

``` json
  "examples": [
    {
      "id": "ssh-1",
      "title": "Allow only SSH Protocol 2",
      "desc": "Only SSH protocol version 2 connections should be permitted. The default setting in /etc/ssh/sshd_config is correct, and can be verified by ensuring that the following line appears: Protocol 2",
[...]

PR #704, opened 2016-05-01
temporary: allow profiles with custom resources to load twice

chris-rock:
In case we use the json formatter in combination with a profile that has custom resources, InSpec will crash due to the restriction to load resources only once.

This PR will return the loaded resource instead of throwing an error. It is a temporary fix only, until all custom resources are scoped within a profile.

### DIFF
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -24,6 +24,15 @@ module Inspec
       end
 
       def self.__register(name, obj)
+        # In cases where we use json formatter with profiles, custom resources get loaded twice.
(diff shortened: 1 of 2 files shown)


arlimus:
Great catch! If we have enough time this week, we should get https://github.com/chef/inspec/pull/671 in, as it fixes the underlying problem of duplicate profile-context execution/loading. Unless we need to rush this one instead. The advantage of the other approach is, that it doesn't silently hide duplicate loading calls to the same resource which could be important to debug if it's a valid user error...

chris-rock:
I close it in favor for #671

PR #706, opened 2016-05-03
hpux support for basic port properties

Anirudh-Gupta:
Gives listening ports, its address and protocol


lib/resources/port.rb
```diff
@@ -3,7 +3,7 @@
 # author: Dominik Richter
 
 require 'utils/parser'
-
+require 'pry'
```
chris-rock: Do we need `pry`?

test/unit/mock/cmd/netstat-inet6
```diff
@@ -0,0 +1,11 @@
+Active Internet connections (IPv6, including servers)
```
chris-rock: It may make sense to rename the file to `test/unit/mock/cmd/hpux-netstat-inet6`

chris-rock:
@Anirudh-Gupta Thanks for the addition. I have only small questions. Do we require `pry`? Also, I would like to prefix the test-files with hpux in this case, since `netstat` is available on multiple systems. Do you think that makes any sense?

Anirudh-Gupta:
@chris-rock Yes it does. Thanks for pointing it out.

chris-rock:
Awesome @Anirudh-Gupta

PR #707, opened 2016-05-03
use strings instead of symbols

vjeffrey:
[test-sym-to-string.txt](https://github.com/chef/inspec/files/247608/test-sym-to-string.txt)

resolves: https://github.com/chef/inspec-scap/issues/125

once merged, we will need to do a whole bunch of test fixing in inspec-scap

@arlimus, @alexpop


arlimus:
This looks great, thank you @vjeffrey :smile:

PR #715, opened 2016-05-06
Expose systemd service properties via .info

alexpop:
Systemd has lots of properties that this PR is exposing now. Usage example:

``` bash
describe service('sshd').info[:properties]['Restart'] do
  it { should cmp 'on-failure'  }
end
```

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -229,6 +229,7 @@ module Inspec::Resources
         running: running,
         enabled: enabled,
         type: 'systemd',
+        properties: params,
       }
     end
   end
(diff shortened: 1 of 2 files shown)


chris-rock:
This looks good @alexpop
Be aware, that the test, may crash if `service('sshd').info[:properties]['Restart']` is not available. Therefore `info` is `nil`

alexpop:
For systemd, the service resource will still be instantiated, even for missing services:

``` ruby
inspec> service('BOGUSSSSSSSSS').info[:properties]['Restart']
=> "no"
inspec> service('BOGUSSSSSSSSS').info
=> {:name=>"BOGUSSSSSSSSS.service",
 :description=>"BOGUSSSSSSSSS.service",
 :installed=>false,
 :running=>false,
 :enabled=>false,
 :type=>"systemd",
 :properties=>
  {"Type"=>"",
   "Restart"=>"no",
   "PIDFile"=>"",
   "NotifyAccess"=>"none",
```

arlimus:
@alexpop To Chris' point, if these were to run on e.g. an old Debian system, where we are not based on Systemd, or if any of the calls don't get through, we will have info == nil.

alexpop:
Correct, but we only generate tests like that from SCAP `systemdunitproperty_test` resources.

PR #718, opened 2016-05-06
Add documentation on how to use ruby 

alexpop:


chris-rock:
That is very helpful 💯 @alexpop

PR #720, opened 2016-05-06
inspec detect learns human-readable output

### LINKED ISSUE #300: Optimize InSpec detect
We could enrich the `inspec detect` command with more valuable information. Currently we recieve the following output:

```
$ inspec detect
{"name":"mac_os_x","family":"darwin","release":"10.10.5","arch":null}
```

Especially the `family` does not reflect the power inspec offers with methods like `os.linux?` or `os.redhat?`. Therefore I propose to return an array for family. This returns all tags for an operating system like

``` json
{
    "name":"mac_os_x",
    "release":"10.10.5",
    "arch":null,
    "family":["darwin", "unix"]
}

{
    "name":"centos",
    "release":"7",
    "arch": "x86",
    "family":["linux", "unix", "redhat"]
}
```

We also should think about:
- Users are confused by the json output that may return `null`.
- We should have a json and human-readable output analog to the `--format` for exec (maybe even reuse those formatters), as highlight in a previous iteration of this issue by @arlimus 

A human-readable output may look like

``` prose
Operating System: Windows
Release:           2012 R2
Arch:                 X68
```

chris-rock:
This fixes #300 

```
$ inspec detect 
Operating System Details
------------------------
Name: mac_os_x
Family: darwin
Release: 10.10.5
Arch: 

$ inspec detect --format json
{"name":"mac_os_x","family":"darwin","release":"10.10.5","arch":null}
```

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -113,9 +113,19 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
 
   desc 'detect', 'detect the target OS'
   target_options
+  option :format, type: :string
   def detect
-    options_json[:command] = 'os.params'
-    shell_func
+    o = opts.dup
+    o[:command] = 'os.params'
+    res = run_command(o)
+    if opts['format'] == 'json'
+      puts res.to_json
+    else
+      headline('Operating System Details')
+      %w{name family release arch}.each { |item|
+        puts "#{mark_text(item.to_s.capitalize + ':')} #{res[item.to_sym]}"
+      }
+    end
   end
 
   desc 'shell', 'open an interactive debugging shell'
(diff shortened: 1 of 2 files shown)


arlimus:
Nice!! What do you think about:

``` diff
diff --git a/lib/inspec/cli.rb b/lib/inspec/cli.rb
index 6ec254a..80ea5e1 100644
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -115,9 +115,9 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
   target_options
   option :format, type: :string
   def detect
-    options_json[:command] = 'os.params'
     o = opts.dup
-    res = runner_command(opts, o[:command])
+    o[:command] = 'os.params'
+    res = run_command(o)
     if opts['format'] == 'json'
       puts res.to_json
     else
@@ -141,7 +141,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
       return Inspec::Shell.new(runner).start
     else
-      res = runner_command(opts, o[:command])
+      res = run_command(o)
       jres = res.respond_to?(:to_json) ? res.to_json : JSON.dump(res)
       puts jres
     end
@@ -156,10 +156,10 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength

   private

-  def runner_command(opts, cmd)
+  def run_command(opts)
     opts[:test_collector] = 'mock'
-    runner.create_context.load(cmd)
+    runner.create_context.load(opts[:command])
   end
 end
```

chris-rock:
Would you like to commit it or should I?

arlimus:
Pushed the update, thank you for the quick review!

PR #721, opened 2016-05-06
Provide service params

alexpop:
Replacing the current method of accessing these params:

``` ruby
describe service('avahi-daemon').info[:properties]['UnitFileState'] do
  it { should be 'enabled' }
end
```

with this:

``` ruby
  # This test will succeed if the service is not enabled
  describe service('avahi-daemon').params do
    its('UnitFileState') { should_not eq 'enabled' }
  end

  # This test will fail with expected: "enabled", got nil
  describe service('MissingInActionService').params do
    its('MissingInActionParam') { should eq 'enabled' }
  end
```
1. Not happy with the `hashie` dependency, but wanted to share the vision and start the discussion.
2. Should we make the `info` method private now?
3. `params` looks to be an internal inspec way of handing attributes. Is `properties` a better name for this method? For example:

``` ruby
  describe service('avahi-daemon').properties do
    its('UnitFileState') { should_not eq 'enabled' }
  end
```

### DIFF
--- a/test/unit/resources/service_test.rb
+++ b/test/unit/resources/service_test.rb
@@ -4,6 +4,7 @@
 
 require 'helper'
 require 'inspec/resource'
+require 'hashie'
 
 describe 'Inspec::Resources::Service' do
   let(:runlevels) { {0=>false, 1=>false, 2=>true, 3=>true, 4=>true, 5=>true, 6=>false} }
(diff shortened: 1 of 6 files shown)


arlimus:
+1 in making info private; is there anything in info, that is not exposed via accessors? anything that shouldn't be exposed via accessors but only via info?

alexpop:
``` ruby
inspec> s.info
=> {:name=>"sshd.service",
 :description=>"OpenSSH server daemon",
 :installed=>true,
 :running=>true,
 :enabled=>true,
 :type=>"systemd",
 :params=>
  {"Type"=>"simple",
   "Restart"=>"on-failure",
```

Right now, out of this bunch, only `installed`, `running` and `enabled` are available directly. Missing `name`, `description` and `type`

chris-rock:
I like the `service('avahi-daemon').params` approach because it works in harmony with other resources we have. We could easily have:

```
 describe service('sshd')do
    its('name') { should eq 'sshd.service' }
   its('description') { should eq 'OpenSSH server daemon' }
   its('type') { should eq 'systemd' }
 end
```

alexpop:
Ok, with the latest code, this is working now:

``` ruby
  describe service('avahi-daemon').params do
    its('UnitFileState') { should_not eq 'enabled' }
  end

  describe service('sshd') do
    its('type') { should eq 'systemd' }
    its('name') { should eq 'sshd.service' }
    its('description') { should eq 'OpenSSH server daemon' }
    its('params.UnitFileState') { should eq 'enabled' }
  end
```

test/unit/resources/service_test.rb
```diff
@@ -22,7 +24,9 @@
   it 'verify ubuntu package parsing' do
     resource = MockLoader.new(:ubuntu1404).load_resource('service', 'ssh')
     srv = { name: 'ssh', description: nil, installed: true, running: true, enabled: true, type: 'upstart' }
```
chris-rock: can we remove the `srv`?

test/unit/resources/service_test.rb
```diff
@@ -31,7 +35,9 @@
   it 'verify ubuntu package parsing with default upstart_service' do
     resource = MockLoader.new(:ubuntu1404).load_resource('upstart_service', 'ssh')
     srv = { name: 'ssh', description: nil, installed: true, running: true, enabled: true, type: 'upstart' }
```
chris-rock: we do not need `srv` anymore

test/unit/resources/service_test.rb
```diff
@@ -41,7 +47,9 @@
   it 'verify ubuntu package parsing' do
     resource = MockLoader.new(:ubuntu1504).load_resource('service', 'sshd')
     srv = { name: 'sshd.service', description: 'OpenSSH server daemon', installed: true, running: true, enabled: true, type: 'systemd', :properties=>{"Id"=>"sshd.service", "Names"=>"sshd.service", "Description"=>"OpenSSH server daemon", "LoadState"=>"loaded", "UnitFileState"=>"enabled", "SubState"=>"running"} }
```
chris-rock: also redundant `srv`

test/unit/resources/service_test.rb
```diff
@@ -88,7 +104,9 @@
   it 'verify centos 7 package parsing with systemd_service and service_ctl override' do
     resource = MockLoader.new(:centos7).load_resource('systemd_service', 'sshd', '/path/to/systemctl')
     srv = { name: 'sshd.service', description: 'OpenSSH server daemon', installed: true, running: true, enabled: true, type: 'systemd', :properties=>{"Id"=>"sshd.service", "Names"=>"sshd.service", "Description"=>"OpenSSH server daemon", "LoadState"=>"loaded", "UnitFileState"=>"enabled", "SubState"=>"running"} }
-    _(resource.info).must_equal srv
+    _(resource.type).must_equal 'systemd'
```
chris-rock, line 119: Could we add checks for the properties? e.g. `LoadState`

chris-rock:
@alexpop Implementation is very good. I would like to remove all `srv` code, where we replaced it with specific test. Also I would add the `.params` tests for all systemd tests (those have been covered by the info object before):

```
 describe service('avahi-daemon').params do
    its('UnitFileState') { should_not eq 'enabled' }
  end
```

test/unit/resources/service_test.rb
```diff
@@ -78,26 +85,29 @@
...
+    _(resource.name).must_equal 'sshd.service'
+    _(resource.description).must_equal 'OpenSSH server daemon'
     _(resource.installed?).must_equal true
     _(resource.enabled?).must_equal true
     _(resource.running?).must_equal true
   end
 
   it 'verify centos 7 package parsing with systemd_service and service_ctl override' do
     resource = MockLoader.new(:centos7).load_resource('systemd_service', 'sshd', '/path/to/systemctl')
-    srv = { name: 'sshd.service', description: 'OpenSSH server daemon', installed: true, running: true, enabled: true, type: 'systemd', :properties=>{"Id"=>"sshd.service", "Names"=>"sshd.service", "Description"=>"OpenSSH server daemon", "LoadState"=>"loaded", "UnitFileState"=>"enabled", "SubState"=>"running"} }
```
chris-rock, line 90: Hi Alex, would it make sense to re-add the systemd tests:

```
_(resource.params).must_equal {object}'
```

chris-rock:
Awesome addition @alexpop

Issue #719, opened 2016-05-06
RFC Attributes


arlimus:
InSpec requires attributes to configure:
- per-machine information
- secrets / credentials

At the same time the philosophy here should be to keep attributes to a bare minimum, as more attributes raise complexity and more complex tests will likely lead to false positives/negatives in inspec.

The current idea for the interface from internal testing has been:

``` ruby
user = attribute('user', required: true)
password = attribute('password', required: true)

attribute(
  field_name,
  required: [true]/false
  validate: [nil]/regex/block
  type: [any]/string/int/numeric/boolean/regex/array/hash
)
```

attributes are always scoped to the profile, i.e. `profile/attribute_name` is required. When profiles are renamed/rescoped through dependency resolution, their new name is applied.

This addresses the following issues:
- Fixes https://github.com/chef/inspec/issues/664
- Fixes https://github.com/chef/inspec/issues/626
- Fixes https://github.com/chef/inspec/issues/611
- Fixes https://github.com/chef/inspec/issues/617

It will also lead to:
- https://github.com/chef/inspec/issues/609 by adding a provider for chef-vault

To fully implement this RFC, we should have a reference implementation for:
- All attribures are scoped https://github.com/chef/inspec/issues/1059
- implement [vault](https://www.vaultproject.io/) secrets backend
- implement [chef-vault](https://github.com/chef/chef-vault) secrets backend
- optional: parse attributes from CLI `INSPEC_ATTR_user=john`

chris-rock:
Things we should consider:
- default values for attributes
- description

```
attribute(
  field_name,
  default: 'value'
  description: 'This is a longer-description, why this is required'
  required: [true]/false
  validate: [nil]/regex/block
  type: [any]/string/int/numeric/boolean/regex/array/hash
)
```

Also I propose that we allow different input methods for attributes. This could be done as following:

```
INSPEC_ATTR_user=john INSPEC_ATTR_password=doe inspec exec my-profile
inspec exec my-profile -attr-file=variables.yml variables2.yml
```

We should also think about an interactive-mode for quick testing different results:

```
inspec exec my-profile --interactive
The profile `my-profile` requests 2 attributes. Abort with Ctrl-C.
  `user`:_
  `password`:_

InSpec executes `my-profile` now.
...
```

During my work on #723, a couple of questions came up:
- What happens if required is set to false? Should the profile fail, when the attribute is not set? Wouldn't a default value have the same impact as required?
[...]

PR #723, opened 2016-05-08
Feature: Implementation of RFC Attributes

chris-rock:
This PR implements parts of RFC Attributes #719

To use it:
`$ inspec exec examples/profile-attribute --attrs examples/profile-attribute.yml`

An attributes file looks like:

```
cat examples/profile-attribute.yml 
user: bob
password: secret
```

The usage within a profile is as following:

```
val_user = attribute('user', required: true)

describe val_user do
  it { should eq 'johndoe' }
end
```


examples/profile-attribute/controls/example.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+val_user = attribute('user', default: 'alice', description: 'An identification for the user')
+val_password = attribute('password', description: 'A value for the password')
+
+# that works
+describe 'bob' do
+  it { should eq val_user }
```
arlimus: Inverting feels more readable:

```
describe val_user do
  it { should eq 'bob' }
```

examples/profile-attribute/controls/example.rb
```diff
@@ -0,0 +1,20 @@
...
+end
+
+describe 'secret' do
+  it { should eq val_password }
```
arlimus: ditto

examples/profile-attribute/controls/example.rb
arlimus: ditto vv

arlimus:
Awesome work @chris-rock !!

Adding the ncessary tests should round this up.

examples/profile-attribute/controls/example.rb
chris-rock: Yeah, I like that...

PR #724, opened 2016-05-09
Remove redundant space when missing expectation

alexpop:
This code change removes an extra space that's added when an expectation is not provided. I always noticed that extra space and thought for a second that something is wrong.

Diff for one of the translated tests after this change:

``` ruby
   describe file("/boot/grub2/grub.cfg") do
-    it { should exist  }
+    it { should exist }
   end
   describe file("/boot/grub2/grub.cfg") do
     it { should_not be_executable.by "group" }
   end
```


arlimus:
Nice cleanup @alexpop kudos!

PR #729, opened 2016-05-10
fix: remove non-existent class

chris-rock:
This removes the trail to load a non-existing class


alexpop:
Thanks for the cleanup Chris

PR #732, opened 2016-05-10
run integration tests in docker

chris-rock:
- deactivates all tests that do not work well in docker yet, we use ENV['DOCKER'] for that (Feature #723 would be so great )
- add kitchen-dokken to run the test
- activate kitchen-docken via travis
- enable integration tests for ubuntu 12.04 and 16.04


.travis.yml
```diff
@@ -17,6 +17,17 @@ matrix:
     script: bundle exec rake lint test test:resources config=test/test.yaml
   - rvm: 2.2
     script: N=5 bundle exec rake test:functional test:resources config=test/test-extra.yaml
+  - rvm: 2.2
+    script: N=5 bundle exec rake test:functional test:resources config=test/test-extra.yaml
```
arlimus: please de-dupe

test/integration/default/apache_conf_spec.rb
```diff
@@ -4,8 +4,8 @@
 
   # direct access to params of apache conf
   describe apache_conf do
-    its('LogLevel') { should eq 'warn' }
-    its('MaxKeepAliveRequests') { should eq '100' }
+    its('LogLevel') { should cmp 'warn' }
+    its('MaxKeepAliveRequests') { should cmp '100' }
```
arlimus: `should cmp 100` ? (i.e. as a number)

chris-rock:
I'll add a env variable to deactivate tests that do not work on docker

.travis.yml
chris-rock: Great catch

test/integration/default/apache_conf_spec.rb
chris-rock: that is something we should do

test/integration/default/apache_conf_spec.rb
```diff
@@ -1,15 +1,17 @@
 # encoding: utf-8
 
-if os.linux?
+# TODO: do not run those tests on docker yet
+unless ENV['DOCKER']
+  if os.linux?
```
arlimus: all of these handlers, including the `os.linux?` check are creating huge diffs and unnecessary indentation.

i would like to avoid these in the future. please use:

``` ruby
return if Env['DOCKER']
return unless os.linux?
```

chris-rock: we may think of using controls with only_if

chris-rock: I updated the tests to reflect that idea

arlimus:
Huge change, awesome work @chris-rock :smile: :+1:

PR #735, opened 2016-05-11
Escape os_env command on Windows to handle env variables containing parentheses.

tpcwang:
One of the common environment variable is "ProgramFile(x86)", and the previous implementation does not handle that correctly.


chris-rock:
Thanks @tpcwang for adding this safty-check 💯

PR #736, opened 2016-05-11
Fix contain_match, add none_match

alexpop:
Other changes:
- update matchers doc and add more integration tests
- allow non-string data types and non-arrays

`contain_match` was incorrectly reporting success for a test like this:

``` ruby
describe ['lemon', 'ginger', 'grapes'] do
  it { should contain_match /^ginnnnnnnnnnnn/}
end
```


chris-rock:
I like the addition of the none_matcher

PR #737, opened 2016-05-11
deprecate array matcher

chris-rock:
We decided, that we go full steam backwards regarding the array matchers. Therefore we are going to deprecate all array matchers. #738 is collecting the use cases, that InSpec is not able to cover yet.


arlimus:
Awesome, thank you @chris-rock !! Let's solve this in 738 :grin:

PR #739, opened 2016-05-11
Return empty array instead of nil for port methods

alexpop:
This makes the methods that return arrays be friendly with `each` loops.


chris-rock:
Thanks @alexpop

PR #743, opened 2016-05-12
Make list properties accessible in its

alexpop:
The change is needed in order to access process properties like this:

``` ruby
processes("*").list.each do |entry|
  describe entry do
    its("pid") { should be > 0 }
  end
  describe entry do
    its("command") { should match ".+" }
  end
end
```


arlimus:
Superseded by https://github.com/chef/inspec/pull/744

PR #750, opened 2016-05-13
remove string quotes around regexp (docfix)

lamont-granquist:


lamont-granquist:
i didn't look for other occurrences

chris-rock:
@lamont-granquist Thanks for finding those issues. I added some more occurrences.

PR #751, opened 2016-05-13
another doc tweak.

lamont-granquist:
the docs on this resource matcher are very confusing to someone
with no prior inspec experience.  the meaning of 'item' and 'value' are
totally opaque.   it needs less generic and more relevant examples.

(and i'm still not quite sure i've gotten it right here)


lamont-granquist:
nevermind, just having issues parsing it, syntax still reads odd in my head..

PR #756, opened 2016-05-16
modification in command resource example

Anirudh-Gupta:
Modified the command resource example. The below code fails : 

``` ruby
describe command('ls -al /') do
 it { should exist }
end
```

Instead this should be written :

``` ruby
describe command('ls') do
 it { should exist }
end
```

### DIFF
--- a/lib/resources/command.rb
+++ b/lib/resources/command.rb
@@ -10,10 +10,14 @@ module Inspec::Resources
     desc 'Use the command InSpec audit resource to test an arbitrary command that is run on the system.'
     example "
       describe command('ls -al /') do
-        it { should exist }
         its('stdout') { should match /bin/ }
         its('stderr') { should eq '' }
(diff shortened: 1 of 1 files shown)


chris-rock:
Thanks @Anirudh-Gupta for highlighting this. The reason for that is the checking for command availability:

```
if inspec.os.linux?
  res = inspec.backend.run_command("bash -c 'type \"#{@command}\"'")
elsif inspec.os.windows?
  res = inspec.backend.run_command("where.exe \"#{@command}\"")
elsif inspec.os.unix?
  res = inspec.backend.run_command("type \"#{@command}\"")
else
  warn "`command(#{@command}).exist?` is not suported on your OS: #{inspec.os[:family]}"
  return false
end
res.exit_status.to_i == 0
```

See https://github.com/chef/inspec/blob/master/lib/resources/command.rb#L46-L55

PR #754, opened 2016-05-15
add sudo_command option

jeremymv2:
This adds the sudo_command option implemented in train here: https://github.com/chef/train/pull/107

PR #757, opened 2016-05-16
update train dependency to 0.12.0

chris-rock:
this update is required for #754 to work properly, since those options are exposed by train


arlimus:
Looks great, cheers :+1:

PR #758, opened 2016-05-16
fixed 'it' statements under file_test

Anirudh-Gupta:


chris-rock:
Thanks for finding this @Anirudh-Gupta

PR #774, opened 2016-05-30
add helper methods for os resource

chris-rock:
This PR allows users to

```
describe os.family do
  it { should eq 'redhat' }
end
```

instead of

```
describe os[:family] do
  it { should eq 'redhat' }
end
```

### DIFF
--- a/lib/resources/os.rb
+++ b/lib/resources/os.rb
@@ -7,8 +7,16 @@ module Inspec::Resources
     name 'os'
     desc 'Use the os InSpec audit resource to test the platform on which the system is running.'
     example "
-      describe os[:family] do
+      describe os.family do
         it { should eq 'redhat' }
+      end
+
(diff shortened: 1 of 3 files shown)


lib/resources/os.rb
```diff
@@ -25,6 +33,16 @@ def [](name)
       inspec.backend.os[name]
     end
 
+    # add helper methods for easy access of properties
+    # allows users to use os.name, os.family, os.release, os.arch
+    %w{name family release arch}.each do |property|
+      os = self
```
arlimus: is this line needed ^^ ?

arlimus:
Much simpler, kudos! Could you add the tests for `os` please?

chris-rock:
@arlimus will add a test for os resource

chris-rock:
Added the tests

PR #782, opened 2016-06-06
expose output option for inspec exec

chris-rock:
This allows user to store the result in files:

```
inspec exec examples/profile --format json --output report.json
cat report.json | jq
```


chris-rock:
Okay, we decided to take the `cat` route:

```
inspec exec examples/profile --format json > report.json
cat report.json | jq
```

arlimus:
Imho we should stick to basic shell principles if possible. Also, a great alternative for printing and processing might be:

``` bash
inspec exec examples/profile --format json | tee report.json | jq
```

PR #791, opened 2016-06-15
integer?("0300") should not be true

srenatus:
because if it is, `cmp` can never compare "octal strings", since it doesn't reach https://github.com/chef/inspec/blob/master/lib/matchers/matchers.rb#L262


chris-rock:
@srenatus Awesome. Do you have time to add a test?

arlimus:
@chris-rock just added it ahead :grin: 
kudos @srenatus , great fix :+1:

PR #795, opened 2016-06-16
use train instead of r-train

chris-rock:
Special thanks to @halo for providing the `train` gem name


arlimus:
See https://github.com/chef/train/pull/115
Huge thanks to @halo !!

PR #800, opened 2016-06-18
improve wmi resource

chris-rock:
This improves the wmi resource:

```
# new syntax
describe wmi({
  class: 'RSOP_SecuritySettingNumeric',
  namespace: 'root\\rsop\\computer',
  filter: 'KeyName = \'MinimumPasswordAge\' And precedence=1'
}) do
   its('Setting') { should eq 1 }
end

# deprecated syntax
describe wmi('RSOP_SecuritySettingNumeric', {
  namespace: 'root\\rsop\\computer',
  filter: 'KeyName = \'MinimumPasswordAge\' And precedence=1'
}) do
   its('Setting') { should eq 1 }
   its('setting') { should eq 1 }
end
```

In addition this resource supports `query` now:

```
describe wmi({
  namespace: 'root\rsop\computer',
  query: "SELECT Setting FROM RSOP_SecuritySettingBoolean WHERE KeyName='LSAAnonymousNameLookup' AND Precedence=1"
}) do
  its('Setting') { should eq false }
end
```

### DIFF
--- a/lib/resources/wmi.rb
+++ b/lib/resources/wmi.rb
@@ -13,7 +13,8 @@ module Inspec::Resources
     name 'wmi'
     desc 'request wmi information'
     example "
-      describe wmi('RSOP_SecuritySettingNumeric', {
+      describe wmi({
+        class: 'RSOP_SecuritySettingNumeric',
         namespace: 'root\\rsop\\computer',
         filter: 'KeyName = \'MinimumPasswordAge\' And precedence=1'
       }) do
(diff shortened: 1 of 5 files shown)


lib/resources/wmi.rb
```diff
@@ -24,13 +25,20 @@ class WMI < Inspec.resource(1)
...
+        opts ||= {}
+      end
+
       # verify that this resource is only supported on Windows
       return skip_resource 'The `windows_feature` resource is not supported on your OS.' unless inspec.os.windows?
-
-      @wmiclass = wmiclass
-      @wminamespace = opts[:namespace]
-      @wmifilter = opts[:filter]
+      @options = opts
```
arlimus: Could this be shortened?
Maybe

```
@options = opts || {}
if wmiclass.is_a?(Hash)
  @options.merge!(wmiclass)
else
  @options[:class] = wmiclass
end
```

Should we add a deprecation warning for the old wmiclass mode?

lib/resources/wmi.rb
```diff
@@ -40,36 +48,69 @@ def method_missing(*keys)
...
-    def info
+    def params
       return @content if defined?(@content)
       @content = {}
 
-      # we should abort execution, if wmi class is not given or wmi resource is
-      # executed on a non-windows system
-      return @content if @wmiclass.nil?
+      # abort if no options are available
+      return @content if @options.nil?
```
arlimus: Can `@options` ever be nil? Since the code at the top changed...

chris-rock: lets move the skip to top in initialize. then it will be nil for all non-supported operating systems

lib/resources/wmi.rb
chris-rock: I like the shortened version

lib/resources/wmi.rb
arlimus: Actually it will be undefined :wink: (and throw a warning if we check for value)
Intention understood, lgtm :+1: 

```
return @content unless defined?(@options)
```

arlimus:
Awesome, kudos @chris-rock :+1:

PR #801, opened 2016-06-18
add boolean support for cmp matcher

chris-rock:
cmp matcher learns the comparison of string values with boolean values:

```
describe true do
    it { should cmp 'true' }
    it { should cmp 'True' }
    it { should cmp true }
  end

  describe false do
    it { should cmp 'false' }
    it { should cmp 'False' }
    it { should cmp false }
  end
```


arlimus:
This should get us to feature-completion on `cmp`, awesome addition @chris-rock :+1: 
kudos !!

PR #803, opened 2016-06-21
Add array documentation to yaml / json resource

brentm5:
Added some more documentation for the yaml / json resources because I could not find a way to find items in arrays


chris-rock:
Thanks @bigbam505 for this improvement.

PR #809, opened 2016-06-28
expose resources inside of describe blocks

### LINKED ISSUE #451: `os` resource not accessible within a `describe`
For the purpose of not having to duplicate `describe` items, would it be possible to make the `os` resource available inside a describe?  This way you  can write a concise and elegant test that handles different architectures and releases, all from within a single `describe`. 

For example:

```
describe file('/etc/ssh/sshd_config') do
  its('content') { should match('ChallengeResponseAuthentication no') }
  its('content') { should match('UsePAM yes') }
  its('content') { should match('PasswordAuthentication no') }
  if os.redhat?
    its('content') { should match("Subsystem sftp /usr/libexec/openssh/sftp-server") }
    if os[:release].to_f >= 7
       its('content') { should match("UsePrivilegeSeparation sandbox") }
       its('content') { should match("HostKey /etc/ssh/ssh_host_rsa_key") }
       its('content') { should match("HostKey /etc/ssh/ssh_host_ecdsa_key") }
    end
  end
end
```

arlimus:
Please find the initial user request here: https://github.com/chef/inspec/issues/451

tldr:

``` ruby
describe 'something...' do
  # i can access resources vv
  if os.redhat?
    ...
  end
end
```

In this example ^^ the resource `os` is currently not available inside of a `describe` block.

Fixes https://github.com/chef/inspec/issues/451

Let's pick the discussion back up in here. I'm not sure at this point whether we should actually get this feature into inspec.
# 

RSpec in its pure form did not allow for this. What we are starting to do, is to get our own resource keywords inside of `describe` blocks. This may lead to clashes in the namespace (ie resource names colliding with rspec functions), which is virtually impossible to resolve.

It is possible in its traditional application to define terms outside the block and use them, which also works in inspec:

``` ruby
val = os.redhat?
describe 'my test' do
  if val
    it { should eq true }
  end
end
```

(although obviously a better choice would be to use inspec's `supports` flag ...)

The question now becomes, is it worth diluting the DSL inside the describe block to get this in. i.e. what are the use-cases we are trying to solve?

Please help with this discussion

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -11,7 +11,7 @@ module Inspec
     # Create the transport backend with aggregated resources.
     #
     # @param [Hash] config for the transport backend
-    # @return [TransportBackend] enriched transport instance
+    # @return [TransportBackend] enriched transport module
     def self.create(config)
       conf = Train.target_config(config)
       name = Train.validate_backend(conf)
@@ -25,7 +25,7 @@ module Inspec
         fail "Can't connect to transport backend '#{name}'."
       end
 
-      cls = Class.new do
+      Module.new do
         define_method :backend do
           connection
         end
@@ -36,12 +36,21 @@ module Inspec
         end
       end
 
-      cls.new
-
     rescue Train::ClientError => e
       raise "Client error, can't connect to '#{name}' backend: #{e.message}"
     rescue Train::TransportError => e
       raise "Transport error, can't connect to '#{name}' backend: #{e.message}"
+    end
+
+    # Creates an instance of the transport backend with all resources.
+    # It can be used to invoce methods directly.
+    #
+    # @param [Hash] config of the backend
+    # @param [Any] base optional base class
+    # @return [TransportBackend] backend instance
+    def self.instance(config, base = Object)
+      dsl = create(config)
+      Class.new(base) { include dsl }.new
     end
   end
 end
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -47,8 +47,9 @@ module Inspec
     end
 
     def configure_transport
-      @backend = Inspec::Backend.create(@conf)
-      @test_collector.backend = @backend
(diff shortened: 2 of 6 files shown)


chef-supermarket:

* @arlimus

arlimus:
This change may not be very large, but it took a while to get working. It carries some very heavy implications, i.e. let's treat this with care.

mhedgpeth:
I don't really like it, it seems that it gives people control that they don't need. They should use the supports keyword or an if statement outside of the whole describe block.

When learning the DSL the hardest part is inside of the describe block. We should keep that as simple as possible.

arlimus:
I agree on all the `only_if` cases. Which leads us to the question: Are there other use-cases?

I know that this request has been raised for serverspec migration, which imho leads to an overarching question of use-cases people are coming from. A migration path should be easy enough to understand and feasible to implement.

arlimus:
Also to keep in mind, resource scope in the current implementation is essentially this:

inspec resource layers

mhedgpeth:
Something @anniehedgpeth and I went over with @chris-rock the other day is: how much ruby can be used within the file. Perhaps there is a different workaround that involves writing some ruby code or parameterizing the describe block? Could we keep the block itself in a pure form from a DSL perspective and then also allow for edge cases by allowing for parameterized creation of those blocks?

arlimus:
@mhedgpeth imho it'd be great to think about use-cases first, before covering workarounds. There are different ways to solve problems, but I'd prefer to understand the problem before looking for solutions :wink: 

To answer your question: If we think about Ruby the language, it is a required dependency of the current inspec DSL. It is available at every layer. That being said, I don't believe we should move into deeper layers of code, as it complicates work for non-developers, e.g. auditors.

arlimus:
Closing this PR for a lack of user needs. Please help us make an informed decision by adding use-cases this would impact and helping in the discussion below.

Issue #672, opened 2016-04-20
Windows default path format causes errors with inspec check


grdnrio:
### Description

Inspec check using autocomplete for paths in Powershell causes an error. Deleting the .<directory>\ dot and slashes works.
### InSpec and Platform Version

Inspec version: 0.19.1
Windows 10 x64 1511
### Replication Case

``` powershell
chef exec inspec check .\cis-ubuntu14.04-level1\
C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/gems/inspec-0.19.1/lib/inspec/profile.rb:27:in `resolve_target': Don't understand inspec profile in ".\\cis-ubuntu14.04-level1\\", it doesn't look like a supported profile structure. (RuntimeError)
        from C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/gems/inspec-0.19.1/lib/inspec/profile.rb:34:in `for_target'
        from C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/gems/inspec-0.19.1/lib/inspec/cli.rb:55:in `check'
        from C:/opscode/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from C:/opscode/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from C:/opscode/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from C:/opscode/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/gems/inspec-0.19.1/bin/inspec:9:in `<top (required)>'
        from C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/bin/inspec:22:in `load'
        from C:/Users/jgardiner/AppData/Local/chefdk/gem/ruby/2.1.0/bin/inspec:22:in `<main>'
```
### Possible Solutions

Workaround is to manually remove the backslahes and dot.

chris-rock:
Huh. That is a bumper. We need to fix this. Thanks for reporting @grdnrio

mhedgpeth:
[...]

PR #810, opened 2016-06-30
Attempt at a bug fix to read backslashes as forward slashes

anniehedgpeth:
@mhedgpeth and I attempted to solve the issue for #672. 

We used the `gsub!` method in `self.resolve` to change the backslashes to forward slashes in `local.rb`.

We tested it in `irb` using `File.exist?`, `File.join`, and `File.file?`, and all of those worked once the backslash was changed to a forward slash. 

However, when we tried unit testing using `bundle exec ruby -W -Ilib:test test/unit/fetchers_test.rb` it failed stating:

```
Expected: ["./path/"]
  Actual: [".\\path\\"]
```

So am I correct to assume that we didn't fix the correct module? Or did we write the right test?

### DIFF
--- a/lib/fetchers/local.rb
+++ b/lib/fetchers/local.rb
@@ -10,6 +10,7 @@ module Fetchers
     attr_reader :files
 
     def self.resolve(target)
+      target.gsub!(/\\/, '/')
       return nil unless File.exist?(target)
       new(target)
     end
--- a/test/unit/fetchers_test.rb
+++ b/test/unit/fetchers_test.rb
@@ -20,38 +20,39 @@ describe Inspec::Plugins::RelFetcher do
   let(:src_fetcher) { mock() }
 
   IN_AND_OUT = {
-    []                        => [],
-    %w{file}                  => %w{file},
+    []                           => [],
+    %w{file}                     => %w{file},
     # don't prefix just by filename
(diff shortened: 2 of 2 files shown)


chris-rock:
Hi @anniehedgpeth apologize for the late response, I schedule this for next week.

vjeffrey:
@anniehedgpeth   thank you - this is great!  After some time debugging, I realized the failing unit test was failing because that test never hits that code path -- a test needs to go into test/unit/fetchers/local_test.rb instead...something like:
`describe 'thing' do
    let(:res) { fetcher.resolve('test\unit\fetchers\local_test.rb') }
    it 'must change stuff' do
      _(res.files).must_equal ['test/unit/fetchers/local_test.rb']
    end
  end`

lib/fetchers/local.rb
```diff
@@ -10,6 +10,7 @@ class Local < Inspec.fetcher(1)
     attr_reader :files
 
     def self.resolve(target)
+      target.gsub!(/\\/, '/')
```
vjeffrey, line 13: instead of mutating the string, could we do a target = target.tr('\','/') here?  ...or, alternatively, just a gsub without the bang after?

chris-rock:
This PR is tracked in the rebased version #963

PR #811, opened 2016-07-04
Static keys in all json

arlimus:
**Why? Self-documenting API and data fields.**

I have recently received the question: "Why do we need filenames and paths in those groups field?" to which I responded "They are actually IDs, it's just paths by convenience right now." Our current output format carries 2 inherent assumptions, which users might not understand: controls and groups are actually maps/dicts that connect IDs to their items.

**Changes**

This MR essentially turns:

```
      "groups": {
        "controls/meta.rb": {
          "title": "SSH Server Configuration",
          "controls": [
            "ssh-1"
          ]
        },
        "controls/example.rb": {
...
```

into

```
      "groups": [
        {
          "id": "controls/meta.rb",
          "title": "SSH Server Configuration",
          "controls": [
            "ssh-1"
          ]
        },
        {
          "id": "controls/example.rb",
...
```

and also changes:

```
      "controls": {
        "ssh-1": {
          "title": "Allow only SSH Protocol 2",
...
```

into

```
      "controls": [
        {
          "id": "ssh-1",
          "title": "Allow only SSH Protocol 2",
...
```

**Added bonuses**
- Easier interaction with multiple tools like `jq` and `elasticsearch`
- Easier javascript interaction

**Negative points**
- Controls and groups are not unique by design. (with arrays you may have multiple entries with same IDs)
- Any processing that looks for control and group IDs will need more work to find the right item (it is simpler to access via map/dict). e.g. connecting `inspec exec --format json-min` with `inspec json`

**Please join the discussion below**

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -149,16 +149,22 @@ module Inspec
 
     def info(res = params.dup)
       # add information about the controls
-      controls = res[:controls].map do |id, rule|
+      res[:controls] = res[:controls].map do |id, rule|
         next if id.to_s.empty?
         data = rule.dup
         data.delete(:checks)
         data[:impact] ||= 0.5
         data[:impact] = 1.0 if data[:impact] > 1.0
         data[:impact] = 0.0 if data[:impact] < 0.0
-        [id, data]
+        data[:id] = id
+        data
+      end.compact
+
+      # resolve hash structure in groups
+      res[:groups] = res[:groups].map do |id, group|
+        group[:id] = id
+        group
       end
-      res[:controls] = Hash[controls.compact]
 
       # add information about the required attributes
       res[:attributes] = res[:attributes].map(&:to_hash) unless res[:attributes].nil? || res[:attributes].empty?
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -105,7 +105,7 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
   def start(_notification)
     # Note that the default profile may have no name - therefore
     # the hash may have a valid nil => entry.
-    @profiles_info = Hash[@profiles.map { |x| profile_info(x) }]
+    @profiles_info = @profiles.map(&:info!).map(&:dup)
   end
 
   def dump_one_example(example, control)
@@ -195,11 +195,6 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
 
   private
 
-  def profile_info(profile)
-    info = profile.info!.dup
-    [info[:name], info]
-  end
-
   #
   # TODO(ssd+vj): We should probably solve this by either ensuring the example has
   # the profile_id of the top level profile when it is included as a dependency, or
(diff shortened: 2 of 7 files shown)


chef-supermarket:

* @arlimus

chris-rock:
Thanks @arlimus for bringing up that question. I try to self-answer some questions and feedback is welcome to get the appropriate output.

_What is the purpose of the json format?_

It is a data-exchange format

_Then the next questions comes into my mind: What are the properties of a data-exchange format?_
- easy to consume eg. easy to parse file
- data-integrity

Form file parsing perspective, I see the point that it is easier to iterate over an array of elements, on the other hand the access of single elements is way more difficult. Therefore the pros and cons are on the same level here.

Especially **data-integrity** is difficult to achieve (see https://en.wikipedia.org/wiki/Data_integrity). By using arrays instead of unique keys, users could manipulate the content and add multiple entries with the same id (had a great discussion with @arlimus about that topic). Now the application needs to handle all edge cases, therefore extra logic is required on top of the json parser. Data-integrity is a core property that we added in the first place, to ensure the data is not corrupted. This allows us to reduce the code that ensures the data is correct and leads to simpler and more secure application software. By relying on this safety mechanism of a json parser, its easy to parse the format in all languages. Now only sha sums are required to ensure the data has not been tampered with. Using arrays, would lead to more insecure application code.

Therefore I vote against the change.

alexpop:
@arlimus, `profiles` would also need to be converted to an array if we are to make this change

arlimus:
Great catch Alex, thank you!!

mhedgpeth:
I would vote for data integrity over reportability. I think you keep the core data model as correct as possible and then transform it into easier to read formats for other tools as needed. I agree with @chris-rock against the change.

arlimus:
Let me grab that data-integrity argument for a sec: I don't believe the Array vs Hash discussion helps us in a significant way to improve or deteriorate data integrity. It is indeed true that additional entries can be added to arrays, that essentially lead to duplicate IDs in controls; However this is only a tiny piece of the landscape of integrity. For example, taking the argument that attackers might manipulate via ID's in arrays vs hashes: the scenario would permit manipulation via IDs either way, which could lead to manipulated entries in the Hash, just as much as in the Array (with the only difference being that the array protects against duplication).

If we want to protect data integrity for profiles and results (and we do, this is planned in the future) we should tackle it with measures that address it (e.g. signatures with integrity information).

mhedgpeth:
@arlimus I wasn't thinking in terms of attack protection but in terms of ensuring that you have uniqueness of keys, so data integrity of the keys and no duplication. I think you're right that the signing and protection-oriented features are outside of the scope of this discussion.

chris-rock:
In order to find a solution for the question, we decided to treat our CLI JSON results similar to API endpoints. In order to increase the quality and stability, we played with the latest version of JSON Schema v4. It is very difficult to verify hash objects, but easy to verify array objects.

In addition to the schema verification, dealing with arrays is easier in Javascript and jq.

Therefore I change my opinion from Hash to Array.

arlimus:
## Decision

Decision for 1.0 JSON changes regarding this discussion:

``` json
{
  "version": "0.34.1",
  "profiles": {
    "profile": {
      "controls": {
        "ssh-1": {
~
      "groups": {
        "controls/meta.rb": {
~
        },
~
```

changes to:

``` json
{
  "version": "0.34.1",
  "profiles": [
    {
      "name": "profile",  // only name here
~
      "controls": [
        {
          "id": "control-xyz",
~
      "groups": [
        {
          "id": "controls/meta.rb",
~
        },
~
```

i.e. all hash-of-hashes change to array-of-hashes

vjeffrey:
--format json works great, but running without the format json on this branch resulted in this output for me: 
screen shot 2016-09-19 at 8 27 19 am
@arlimus

chris-rock:
Thanks to all, this was a great discussion. We decided to switch to an array based-solution, because it makes handling in various languages a lot easier.

💯 @arlimus

PR #828, opened 2016-07-13
Fix runner.rb key inconsistency

alexpop:
Fixes https://github.com/chef/inspec/issues/827

### DIFF
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -109,7 +109,7 @@ module Inspec
     end
 
     def create_context(options = {})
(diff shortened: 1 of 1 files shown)


arlimus:
Thank you for catching it @alexpop

arlimus:
This issue has been bugging me for the last 2 days: Why wasn't it caught by our tests? As it turns out, all the values by default (and the way we run functional tests) come in via Thor, which has a smarter variant of `Hash`. It allows the retrieval of fields via symbol and string.

We need an intermediate test layer.

PR #830, opened 2016-07-14
FIX: file(hba_config_file) test in documentation

atomic111:
I got a rubocop issue with `its('content') { should match /local\s.*?all\s.*?all\s.*?md5/ }`, rubocop message: "Ambiguous regexp literal. Parenthesize the method arguments if it's surely a regexp literal, or add a whitespace to the right of the / if it should be a division."

the last two tests were not working because of this /\/host\s._?all\s._?all\s._?::1\/128\s._?md5\//

Signed-off-by: Patrick Münch patrick.muench1111@gmail.com


chris-rock:
Thanks @atomic111

PR #835, opened 2016-07-17
add mssql resource

chrisevett:
Add resource for running tests against MSSQL (sql server).


arlimus:
Huge Kudos @chrisevett !! :+1: 

We are working to get better coverage with integration testing on all these resources, so this will be added later on.

Issue #845, opened 2016-07-25
Address all test related to iptables resource. 


kdshah1:
### Description

Currently there's  no test for iptables is empty and maybe other test to address all the iptables related policies and rules.
### InSpec and Platform Version

Inspec - 0.26.0
Tell us which version of InSpec (`inspec version` or SHA of a commit) and Operating System distribution you are using
 Red Hat Enterprise Linux Workstation release 6.8 (Santiago)
### Replication Case

Hoe do you check if the iptables INPUT chain is empty?

Tell us what steps to take to replicate your problem.  See [How to create a Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve)
for information on how to create a good replication case.
### Possible Solutions

Add is_empty check in the iptables resource.
If you have already ideas how to solve the issue, add them here.
### Stacktrace

Please include the stacktrace output or link to a gist of it, if there is one.

aaronlippold:
👍

chris-rock:
we use iptables (-f table) -S (chain) https://github.com/chef/inspec/blob/master/lib/resources/iptables.rb#L57

```
iptables-restore < ip_rules 
root@vagrant-ubuntu-trusty-64:/vagrant# iptables -L
Chain INPUT (policy DROP)
target     prot opt source               destination         
ACCEPT     all  --  anywhere             anywhere             state RELATED,ESTABLISHED
ACCEPT     icmp --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:ssh
DROP       icmp --  anywhere             anywhere             icmp timestamp-request
DROP       icmp --  anywhere             anywhere             icmp timestamp-reply
REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited

Chain FORWARD (policy DROP)
target     prot opt source               destination         
REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination                  
root@vagrant-ubuntu-trusty-64:/vagrant# iptables -S
-P INPUT DROP
-P FORWARD DROP
-P OUTPUT ACCEPT
-A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
-A INPUT -p icmp -j ACCEPT
-A INPUT -i lo -j ACCEPT
[...]

PR #846, opened 2016-07-27
Handle xinetd config with only one entry

chris-rock:
This PR handles a bug reported by @kdshah1 via gitter:

is this the correct way to check xinetd services are disabled?

```
        describe xinetd_conf.services('telnet') do
         it{ should be_disabled }
       end
```

It fails with following message:

```
 ✖  services-04: System Must Disable telnet Service. (1 failed)
     expected `Xinetd config /etc/xinetd.conf with service == "telnet".disabled?` to return true, got false
```

whereas my telnet file under /etc/xinet.d reads this:

```
 # default: on
# description: The telnet server serves telnet sessions; it uses \
#       unencrypted username/password pairs for authentication.
service telnet
{
        disable = yes
        flags           = REUSE
        socket_type     = stream
        wait            = no
        user            = root
        server          = /usr/sbin/in.telnetd
        log_on_failure  += USERID
}
```

The problem was, that the xinetd resource was expecting multiple entries via `includedir /etc/xinetd.d`. This PR also adds another test to catch that case.

### DIFF
--- a/lib/resources/xinetd.rb
+++ b/lib/resources/xinetd.rb
@@ -66,17 +66,26 @@ module Inspec::Resources
     def read_params
       return {} if read_content.nil?
       flat_params = parse_xinetd(read_content)
+      # we need to map service data in order to use it with filtertable
       params = { 'services' => {} }
 
-      # parse services that were defined:
+      # map services that were defined and map it to the service hash
       flat_params.each do |k, v|
         name = k[/^service (.+)$/, 1]
+        # its not a service, no change required
         if name.nil?
           params[k] = v
+        # handle service entries
         else
+          # store service
           params['services'][name] = v
+
           # add the service identifier to its parameters
-          v.each { |service| service.params['service'] = name }
+          if v.is_a?(Array)
+            v.each { |service| service.params['service'] = name }
+          else
+            v.params['service'] = name
+          end
         end
       end
       params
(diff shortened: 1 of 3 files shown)


chris-rock:
should fix #845

lib/resources/xinetd.rb
```diff
@@ -75,8 +75,13 @@ def read_params
           params[k] = v
         else
           params['services'][name] = v
+
           # add the service identifier to its parameters
-          v.each { |service| service.params['service'] = name }
```
stevendanna, line 79: Minro style point.  You might consider avoidng the type check and doing:

```
Array(v).each { |service| service.params['service'] = name }
```

Array(v) will coerce a scalar into an array but won't touch the arg if it is already an array:

```
Airb(main):001:0> Array(4)
=> [4]
irb(main):002:0> Array([4])
=> [4]
```

stevendanna:
Overall this looks good to me.  I'm wondering if in the future the thing we actually want to store is a merged view of the config for a given service.

lib/resources/xinetd.rb
chris-rock, line 79: Yes, that makes it more save.

chris-rock:
Lets double-check if we should move this logic to the parser

arlimus:
I like the current version and the functionality in here :+1: 
Imho the discussion of splitting it up is a second iteration. Also is there a benefit of moving it to the parser? @stevendanna sounds like a great idea, let's track it as a separate idea

PR #850, opened 2016-08-01
Support either /dev/kcore or /dev/core in the tests

stevendanna:
I think `/dev/core` is the more usual location but `/dev/kcore` is what appears on Travis.  This allows the tests to run for me locally using `Docker for Mac`

### DIFF
--- a/test/resource/file_test.rb
+++ b/test/resource/file_test.rb
@@ -2,6 +2,20 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 
+#
+# On most systems I've used a link to /proc/kcore is placed at
+# /dev/core. However, on TravisCI we also see it at /dev/kcore.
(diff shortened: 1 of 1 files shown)


stevendanna:
Hrm. This seems to be behaving differently for me locally vs in travis. I'll sort it out locally so it behaves the same way it does in travis.

chris-rock:
@stevendanna that looks great. One question: Is `/dev/core` also working on travis? This would allow us to just have one solution that works everywhere.

Update: travis is not linking `/dev/core` to `/proc/kcore`

stevendanna:
@chris-rock Unfortunately that is one of the first things I tried. It doesn't seem like Travis links /dev/core.

chris-rock:
Great work @stevendanna

PR #857, opened 2016-08-02
Update README with correct directions for integration tests

stevendanna:


chris-rock:
Thanks for the update @stevendanna

PR #873, opened 2016-08-05
remote target supported OS also includes hp-ux

Anirudh-Gupta:
Updated Readme.


chris-rock:
Thanks @Anirudh-Gupta Not sure why we missed it!

Issue #866, opened 2016-08-03
parse_config error when searching for fs.suid_dumpable


anniehedgpeth:
### parse_config error

I wanted to use the parse_config resource instead of a file resource to match content for `fs.suid_dumpable` in `/etc/systctl.conf`. So I tried using

```
  describe parse_config('/etc/sysctl.conf') do
    its('fs.suid_dumpable') { should eq '0' }
  end
```

See: 
https://github.com/anniehedgpeth/ubuntu-14-cis-profile/blob/master/controls/4.rb#L22

But it gets this error

```
  ✖  4.1: 4.1 Restrict Core Dumps (Scored) (1 failed)
     undefined method `suid_dumpable' for nil:NilClass
```
### InSpec and Platform Version

InSpec version 0.28.1
MacOs 10.11.6
### Replication Case

Run `kitchen verify` on this [cookbook](https://github.com/anniehedgpeth/ubuntu-14-hardening).
### Possible Solutions

notsomuch
### Stacktrace

```
I, [2016-08-03T12:41:05.019700 #7211]  INFO -- default-bento-ubuntu-1404: -----> Verifying <default-bento-ubuntu-1404>...
I, [2016-08-03T12:41:05.387584 #7211]  INFO -- default-bento-ubuntu-1404: Detected alternative framework tests for `serverspec`
I, [2016-08-03T12:41:05.387788 #7211]  INFO -- default-bento-ubuntu-1404: Use `/Users/anniehedgpeth/chef_repo/cookbooks/ubuntu-14-hardening/test/integration/default/inspec` for testing
E, [2016-08-03T12:41:05.550415 #7211] ERROR -- default-bento-ubuntu-1404: Verify failed on instance <default-bento-ubuntu-1404>.
E, [2016-08-03T12:41:05.550582 #7211] ERROR -- default-bento-ubuntu-1404: ------Exception-------
E, [2016-08-03T12:41:05.550616 #7211] ERROR -- default-bento-ubuntu-1404: Class: Kitchen::ActionFailed
E, [2016-08-03T12:41:05.550636 #7211] ERROR -- default-bento-ubuntu-1404: Message: Inspec Runner returns 1
E, [2016-08-03T12:41:05.550654 #7211] ERROR -- default-bento-ubuntu-1404: ----------------------
E, [2016-08-03T12:41:05.550671 #7211] ERROR -- default-bento-ubuntu-1404: ------Backtrace-------
[...]

PR #876, opened 2016-08-05
add params and content method to parse_config

chris-rock:
This PR allows users to access the content easier:
- `parse_config_file('/etc/sysctl.conf', options).params['kernel.domainname']`
- `parse_config_file('/etc/sysctl.conf', options).params`
- `parse_config_file('/etc/sysctl.conf', options).content`

```

inspec> parse_config_file('/etc/sysctl.conf', options).params['kernel.domainname']
NameError: undefined local variable or method `options' for #<#<Class:0x007fec28acf910>:0x007fec28acf4d8>
from (pry):1:in `load'
inspec> options = {
inspec>   assignment_re: /^\s*([^=]*?)\s*=\s*(.*?)\s*$/  
inspec> }  
=> {:assignment_re=>/^\s*([^=]*?)\s*=\s*(.*?)\s*$/}
inspec> parse_config_file('/etc/sysctl.conf', options).params['kernel.domainname']
=> "example.com"
inspec> parse_config_file('/etc/sysctl.conf', options).params
=> {"kernel.domainname"=>"example.com"}
inspec> parse_config_file('/etc/sysctl.conf', options).content
=> "#\n# /etc/sysctl.conf - Configuration file for setting system variables\n# See /etc/sysctl.d/ for additional system variables.\n# See sysctl.conf (5) for information.\n#\n\nkernel.domainname = example.com\n\n# Uncomment the following to stop low-level messages on console\n#kernel.printk = 3 4 1 3\n\n##############################################################3\n# Functions previously found in netbase\n#\n\n# Uncomment the next two lines to enable Spoof protection (reverse-path filter)\n# Turn on Source Address Verification in all interfaces to\n# prevent some spoofing attacks\n#net.ipv4.conf.default.rp_filter=1\n#net.ipv4.conf.all.rp_filter=1\n\n# Uncomment the next line to enable TCP/IP SYN cookies\n# See http://lwn.net/Articles/277146/\n# Note: This may impact IPv6 TCP sessions too\n#net.ipv4.tcp_syncookies=1\n\n# Uncomment the next line to enable packet forwarding for IPv4\n#net.ipv4.ip_forward=1\n\n# Uncomment the next line to enable packet forwarding for IPv6\n#  Enabling this option disables Stateless Address Autoconfiguration\n#  based on Router Advertisements for this host\n#net.ipv6.conf.all.forwarding=1\n\n\n###################################################################\n# Additional settings - these settings can improve the network\n# security of the host and prevent against some network attacks\n# including spoofing attacks and man in the middle attacks through\n# redirection. Some network environments, however, require that these\n# settings are disabled so review and enable them as needed.\n#\n# Do not accept ICMP redirects (prevent MITM attacks)\n#net.ipv4.conf.all.accept_redirects = 0\n#net.ipv6.conf.all.accept_redirects = 0\n# _or_\n# Accept ICMP redirects only for gateways listed in our default\n# gateway list (enabled by default)\n# net.ipv4.conf.all.secure_redirects = 1\n#\n# Do not send ICMP redirects (we are not a router)\n#net.ipv4.conf.all.send_redirects = 0\n#\n# Do not accept IP source route packets (we are not a router)\n#net.ipv4.conf.all.accept_source_route = 0\n#net.ipv6.conf.all.accept_source_route = 0\n#\n# Log Martian Packets\n#net.ipv4.conf.all.log_martians = 1\n#\n"
inspec> 

```

### DIFF
--- a/lib/resources/parse_config.rb
+++ b/lib/resources/parse_config.rb
@@ -25,20 +25,31 @@ module Inspec::Resources
       end
     "
 
+    attr_reader :content
     def initialize(content = nil, useropts = nil)
       @opts = {}
       @opts = useropts.dup unless useropts.nil?
       @files_contents = {}
-      @params = nil
 
       @content = content
-      read_content if @content.nil?
+      read_params unless @content.nil?
     end
 
     def method_missing(name)
-      @params || read_content
-      @params[name.to_s]
+      read_params[name.to_s]
     end
+
+    def params(*opts)
+      opts.inject(read_params) do |res, nxt|
+        res.respond_to?(:key) ? res[nxt] : nil
+      end
+    end
+
+    def to_s
+      "Parse Config #{@conf_path}"
+    end
+
+    private
 
     def parse_file(conf_path)
       @conf_path = conf_path
@@ -52,21 +63,18 @@ module Inspec::Resources
         return skip_resource "Can't read file \"#{conf_path}\""
       end
 
-      read_content
+      read_params
     end
 
     def read_file(path)
       @files_contents[path] ||= inspec.file(path).content
     end
 
-    def read_content
-      # parse the file
-      @params = SimpleConfig.new(@content, @opts).params
-      @content
-    end
-
-    def to_s
-      "Parse Config #{@conf_path}"
+    def read_params
+      return @params if defined?(@params)
+      return @params = {} if content.nil?
+      @params = SimpleConfig.new(content, @opts).params
+      @params
     end
   end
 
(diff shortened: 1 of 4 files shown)


chris-rock:
fixed #866

arlimus:
Looks great, thank you @chris-rock :+1:

lib/resources/parse_config.rb
```diff
@@ -52,21 +63,18 @@ def parse_file(conf_path)
...
-    def read_content
-      # parse the file
-      @params = SimpleConfig.new(@content, @opts).params
-      @content
-    end
-
-    def to_s
-      "Parse Config #{@conf_path}"
+    def read_params
+      return @params if defined?(@params)
```
stevendanna, line 74: @params.nil? would be the more idiomatic guard here I think.  Instance variables will return nil by default if they haven't previously been defined. In our other ruby code bases I think we would generally follow this style:

```
def read_params
  @params ||= if content.nil?
                {}
              else
                SimpleConfig.new(content, @opts).params
              end
end
```

preferring the early return in cases where the assignment logic is more complex.  Although, I'm not sure what latest rubocop says about this.

chris-rock, line 74: Great feedback, I'll add fix this in another PR.

PR #882, opened 2016-08-05
Explicitly require docker transport to avoid autoload bug

stevendanna:
Ruby's autoload feature is not threadsafe.  We are hoping requiring the
docker plugin early will fix odd failures we have been seeing.

### DIFF
--- a/test/docker_test.rb
+++ b/test/docker_test.rb
@@ -3,6 +3,14 @@
 
 require_relative 'docker_run'
 require_relative '../lib/inspec'
+#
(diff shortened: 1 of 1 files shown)


ksubrama:
Dunno if you need to access them as well...  we'll see.

In https://github.com/test-kitchen/kitchen-ec2/pull/264, we had to touch one of the loaded classes to actually force it to finish autoloading everything before the threaded code could run.

stevendanna:
@ksubrama We'll keep an eye on it to figure out if we need something like that.

PR #890, opened 2016-08-08
Improve windows support of omnibus installer

### LINKED ISSUE #889: Add windows MSI packaging support to omnibus
Supporting windows builds requires a few tweaks to our omnibus setup to include the necessary WiX files and openssl cert files and the like.

This is a follow up to https://github.com/chef/inspec/pull/658 and the comments in there.

ksubrama:


ksubrama:
@stevendanna @chris-rock @arlimus @alexpop

I still need to fix the icons but the rest of it should be ready to look at.

omnibus/.kitchen.yml
```diff
@@ -2,35 +2,89 @@ driver:
...
+        omnibus:
+          build_user:          vagrant
+          build_user_group:    Administrators
+          build_user_password: vagrant
+          install_dir:         /opscode/inspec
+  <% end %>
 
 suites:
-  - name: default
+  - name: inspec
```
chris-rock, line 89: any reason why we renamed that?

omnibus/Gemfile
```diff
@@ -2,11 +2,13 @@
 source 'https://rubygems.org'
 
 # Install omnibus
-gem 'omnibus', '~> 5.3'
+# Sadly due to an ongoing msys2 issue, producing windows builds requires
+# the two branches below.
+gem 'omnibus', github: 'chef/omnibus', branch: 'ksubrama/gcc_investigate'
```
chris-rock, line 7: just curious? any specific reason why they have not been merged into master?

omnibus/config/projects/inspec.rb
```diff
@@ -17,21 +17,52 @@
 #
 
 name 'inspec'
+friendly_name "Inspec"
```
chris-rock: Officially the name is `InSpec` :-)

omnibus/config/projects/inspec.rb
```diff
@@ -17,21 +17,52 @@
...
 license 'Apache-2.0'
 license_file '../LICENSE'
 
-# Defaults to C:/inspec on Windows
-# and /opt/inspec on all other platforms
-install_dir "#{default_root}/#{name}"
+# Defaults to C:/opscode/inspec on Windows
+# and /opt/inspec on all other platforms.
+if windows?
+  install_dir "#{default_root}/opscode/#{name}"
```
chris-rock, line 30: are we still using opscode?

omnibus/config/projects/inspec.rb
```diff
@@ -17,21 +17,52 @@
...
+else
+  install_dir "#{default_root}/#{name}"
+end
 
 build_version Omnibus::BuildVersion.semver
 build_iteration 1
 
 dependency 'preparation'
+dependency 'ruby'
```
chris-rock: I think the `ruby` and `rb-readline` should be a dependency of inspec?

omnibus/config/projects/inspec.rb
```diff
@@ -17,21 +17,52 @@
...
+dependency 'rb-readline'
+
 dependency 'inspec'
 
+dependency 'gem-permissions'
+dependency 'shebang-cleanup'
+dependency 'openssl-customization'
+dependency 'version-manifest'
+dependency 'clean-static-libs'
```
chris-rock, line 49: What are those dependencies doing?

omnibus/config/software/inspec.rb
```diff
@@ -42,8 +43,5 @@
   gem "build #{name}.gemspec", env: env
   gem "install #{name}-*.gem --no-document", env: env
 
-  # Dependecy added to avoid this pry error: "Sorry, you can't use Pry
```
chris-rock, line 45: +1

omnibus/.kitchen.yml
ksubrama, line 89: None in particular - just cosmetic.  If I've a bunch of VMs running, setting the suite name correctly helps identify the VM inside virtual box.  Otherwise everything looks like "kitchen-omnibus-default-windows-2012r2-standard..."

omnibus/Gemfile
ksubrama, line 7: Long story...  has to do with the back-and-forth we've had on https://github.com/chef/omnibus/pull/695 where some of the debug features that I wanted seem like excess baggage (i.e. code to solve problems that shouldn't exist in the first place...  but problems that do exist now...  problems that I don't have time to solve properly).  But mixed in that PR is a bunch of critical fixes too...  so client pool is taking care of extracting the important stuff out and making sure the tests pass and all that good jazz.  Until then, even chef-client and chef-dk are pinned to these branches :(

stevendanna, line 7: Ugh.

omnibus/config/projects/inspec.rb
```diff
@@ -17,21 +17,52 @@
...
 dependency 'preparation'
+dependency 'ruby'
+dependency 'rb-readline'
+
 dependency 'inspec'
 
+dependency 'gem-permissions'
+dependency 'shebang-cleanup'
+dependency 'openssl-customization'
+dependency 'version-manifest'
```
stevendanna: This isn't needed on newer omnibus and is only provided for backcompat.  Omnibus will create the version manifest by defaullt.

omnibus/.kitchen.yml
```diff
@@ -2,35 +2,89 @@ driver:
   name: vagrant
   forward_agent: yes
   customize:
-    cpus: 2
-    memory: 2048
+    cpus: 4
+    memory: 4096
   synced_folders:
     - ['..', '/home/vagrant/inspec']
+    - ['../../omnibus', '/home/vagrant/omnibus']
```
stevendanna, line 9: Unless we have a need to do this, I'd rather keep these out of our default setup.  It is pretty rare to have to build with a custom omnibus/omnibus-software and those who would need to know what to do.  Maybe comment them?

omnibus/config/projects/inspec.rb
ksubrama, line 30: Yep - everything on windows is under opscode.  C:\opscode corresponds to /opt.  C:\chef corresponds to /etc/chef.  Changing that is pretty much impossible at this point between all the chef products we have making that assumptions :(

omnibus/config/projects/inspec.rb
ksubrama, line 49: Clean up steps before the final packaging step.
- gem-permissions sets all the gem files permissions accurately in case one of them has a file that isn't world readable (this came from a customer complaint...).
- shebang-cleanup ensures that all the scripts in embedded/bin (like bundler or gem or whatever else) explicitly ask for the ruby we install rather than env ruby.
- openssl-customization is our set of hacks to inject SSL certs we need to access S3 resources out of the box.
- version-manifest writes an log file with all the various software definition versions so that we can easily verify what it is we built.
- clean-static-libs goes in and deletes any stray .dll.a and .a files in lib directories because we don't use those for anything after compilation is complete on windows.  It's there to help other native gems link to the binaries we compile... but that's not something we expect people to be doing in the context of inspec.  We don't even give them a compatible compiler to link things with.

I should probably just put these in comments...

omnibus/config/projects/inspec.rb
ksubrama: Oh nice! TIL

omnibus/.kitchen.yml
ksubrama, line 9: Can do :).  It doesn't affect me much - I already have to mount my entire code directory for the windows platforms below because mounting just inspec won't work on windows (git hates dealing with a network share as a top-level directory).

stevendanna:
Nice work!

chris-rock:
fixes #889

PR #892, opened 2016-08-09
Support regular expressions for Windows registry key

chris-rock:
This PR aims to bring support for regular expressions to our `registry_key` resource. A sample could look like:

```
control 'regex-test' do
  title "Ensure 'Always install with elevated privileges' is set to 'Disabled'"
  children = registry_key({
      hive: 'HKEY_USERS'
  }).children(/^S-1-5-21-[0-9]+-[0-9]+-[0-9]+-[0-9]{3,}\\Software\\Policies\\Microsoft\\Windows\\Installer/)
  describe children do
    it { should_not eq []}
  end
  children.each { |key|
    describe registry_key(key) do
      its('AlwaysInstallElevated') { should cmp 0 }
    end
  }
end
```
- [x] support hash params as options for registry key
- [x] allow regular expression as input params

Feedback is welcome. 

cc @ksubrama @mhedgpeth


lib/resources/registry_key.rb
```diff
@@ -60,15 +81,15 @@ def has_property_value?(property_name, property_type = nil, value)
...
       registry_property_value(val, meth)
     end
 
     def to_s
-      "Registry Key #{@name}"
+      "Registry Key #{@options[:name]}"
     end
 
-    private
+    # private
```
chris-rock: this should stay private

lib/resources/registry_key.rb
```diff
@@ -20,36 +35,42 @@ class RegistryKey < Inspec.resource(1)
       end
     "
 
-    attr_accessor :reg_key
-
     def initialize(name, reg_key = nil)
+      puts "start init"
```
stevendanna: Probably remove before merge?

chris-rock: yes

arlimus:
Great improvement and feature!! :+1:

PR #896, opened 2016-08-09
Add `inspec env` command to configure shell tab-completion

### LINKED ISSUE #607: tab-completion support in the inspec command line and in the inspec shell :)
Tab-completion ... at least limited support may be good for the release.

I will see if I can grok bash-complete and add it in.

stevendanna:
This adds a new subcommand:

```
inspec env [SHELL]
```

which outputs a shell-appropriate completion script that the user can
source into their shell:

```
eval "$(inspec env SHELL)"
```

Currently, we provide completions for ZSH and Bash. The completion
scripts are generated from the data Thor collects.

If the user doesn't provide SHELL we attempt to detect what the user's
shell may be using a number of methods.

Fixes #607

### DIFF
--- a/lib/inspec/env_printer.rb
+++ b/lib/inspec/env_printer.rb
@@ -1,0 +1,149 @@
+# encoding: utf-8
+require 'inspec/shell_detector'
+require 'erb'
+require 'shellwords'
+
+module Inspec
+  class EnvPrinter
+    def initialize(command_class, shell = nil)
+      if !shell
+        @detected = true
+        @shell = Inspec::ShellDetector.new.shell
+      else
+        @shell = shell
+      end
+      @command_class = command_class
+    end
+
+    def print_and_exit!
+      exit_no_shell if !have_shell?
+      exit_no_completion if !have_shell_completion?
+
+      print_completion_for_shell
+      print_detection_warning($stdout) if @detected
+      print_usage_guidance
+      exit 0
+    end
+
+    private
+
+    def print_completion_for_shell
+      erb = ERB.new(File.read(completion_template_path), nil, '-')
+      puts erb.result(TemplateContext.new(@command_class).get_bindings)
+    end
+
+    def have_shell?
+      ! (@shell.nil? || @shell.empty?)
+    end
+
+    def have_shell_completion?
+      File.exist?(completion_template_path)
+    end
+
+    def completion_dir
+      File.join(File.dirname(__FILE__), 'completions')
+    end
+
+    def completion_template_path
+      File.join(completion_dir, "#{@shell}.sh.erb")
+    end
+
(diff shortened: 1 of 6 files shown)


lib/inspec/shell_detector.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def shell!
+      @shell = detect
+    end
+
+    private
+
+    def detect
+      # Most of our detection code assumes a unix-like environment
+      return nil if (RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/)
```
chris-rock: May it make sense to output a warning? Otherwise the user may be confused that nothing has been executed?

chris-rock:
@stevendanna great work. Bash completion works like a charm. I've some troubles with zsh and get the following errors:

```
eval "$(inspec env zsh)"
zsh: parse error near `>'
```

My generated content is available here: https://gist.github.com/chris-rock/a30592b0b82d13136772c10162dc5752

lib/inspec/env_printer.rb
```diff
@@ -0,0 +1,148 @@
...
+      $stderr.puts "# No completion script for #{@shell}!"
+      print_detection_warning($stderr) if @detected
+      exit 1
+    end
+
+    def exit_no_shell
+      if @detected
+        $stderr.puts '# Unable to automatically detect shell and no shell was provided.'
+      end
+      $stderr.puts <<EOF
```
chris-rock, line 88: cool!

lib/inspec/env_printer.rb
```diff
@@ -0,0 +1,148 @@
...
+# To use this, eval it in your shell
+#
+#    eval "$(inspec env #{@shell})"
+#
+#
+EOF
+    end
+
+    def print_detection_warning(device)
+      device.puts <<EOF
```
chris-rock, line 66: Its great that you can manually override the detection

lib/inspec/shell_detector.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      shellpath = detect_by_ppid
```
chris-rock, line 41: nice to see the different mechanisms! awesome work @stevendanna

chris-rock:
Awesome implementation @stevendanna LGTM once the zsh is working

stevendanna:
@chris-rock I think ZSH should be working now.  The issue was we weren't escaping the description text and you had plugins that had commands with `'` in their description.  Now we use `shellwords` to escape any special characters in the description.  Let me know if it works for you.

chris-rock:
@stevendanna works like a charm now!

stevendanna:
Merging this despite failing tests since TravisCI seems to be having an incident of some sort.

PR #900, opened 2016-08-10
Add readline ignore markers to color escape codes in the shell

stevendanna:
Previously, if you typed more than 20 characters at the prompt and
attempted to press Ctrl+a (readline's "Move to start of line" command),
your prompt would appear at the ~11th character from the start of the
line, unable to go further back.

screen shot 2016-08-10 at 10 00 29 am

This was a result readline counting the terminal escape sequences we use
for color output as part of the line.

Wrapping these sequences in \001 and \002 instructs readline to ignore
them when doing calculations regarding line-length, resolving the
problem.

screen shot 2016-08-10 at 10 07 09 am

### DIFF
--- a/lib/inspec/shell.rb
+++ b/lib/inspec/shell.rb
@@ -32,7 +32,7 @@ module Inspec
 
       # configure pry shell prompt
       Pry.config.prompt_name = 'inspec'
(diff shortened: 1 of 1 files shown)


chris-rock:
Very nice addition @stevendanna +1

PR #901, opened 2016-08-10
Improve detection of postgresql conf dir and data dir

### LINKED ISSUE #494: RHEL postgres data dir is not /var/lib/postgres as coded
```
lib:catchall# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 6.7 (Santiago)
lib:catchall# pwd
/var/lib
lib:catchall# ls -ld postgre*
ls: cannot access postgre*: No such file or directory
lib:catchall# ls -ld pgsql
drwx------ 4 postgres postgres 4096 Nov 21 05:32 pgsql
lib:catchall#
```

See: https://github.com/chef/inspec/blob/master/lib/resources/postgres.rb#L26-L30

stevendanna:
Redhat conf_dir detection was regressed in 57d7275 which inadvertently
removed the setting of @conf_dir. Any attempt to use the postgres
resource on RHEL would rain an exception:

```
inspec> postgres.data_dir
TypeError: no implicit conversion of nil into String
```

Further, the redhat detection code appears to assume that RHEL always
uses versioned data directories. This however, does not appear to be the
case:

```
$ cat /etc/redhat-release
CentOS release 6.7 (Final)
$ sudo ls /var/lib/pgsql/
backups  data  pgstartup.log
```

The code now can handle both versioned and un-versioned directory
formats on RHEL. Further, it provides diagnostic warnings about
uncertainty in the discovered data directories and configuration
directories.

Fixes #494

### DIFF
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -12,29 +12,111 @@ module Inspec::Resources
     def initialize
       os = inspec.os
       if os.debian?
-        @service = 'postgresql'
-        @data_dir = '/var/lib/postgresql'
-        @version = inspec.command('ls /etc/postgresql/').stdout.chomp
-        @conf_dir = "/etc/postgresql/#{@version}/main"
+        #
+        # https://wiki.debian.org/PostgreSql
+        #
+        # Debian allows multiple versions of postgresql to be
+        # installed as well as multiple "clusters" to be configured.
+        #
+        version = version_from_dir('/etc/postgresql')
+        cluster = cluster_from_dir("/etc/postgresql/#{version}")
+        @conf_dir = "/etc/postgresql/#{version}/#{cluster}"
+        @data_dir = "/var/lib/postgresql/#{version}/#{cluster}"
       elsif os.redhat?
-        @service = 'postgresql'
-        @version = inspec.command('ls /var/lib/pgsql/').stdout.chomp
-        @data_dir = "/var/lib/pgsql/#{@version}/data"
+        #
+        # /var/lib/pgsql/data is the default data directory on RHEL6
+        # and RHEL7. However, PR #824 explicitly added version-based
+        # directories. Thus, we call #version_from_dir unless it looks
+        # like we are using unversioned directories.
+        #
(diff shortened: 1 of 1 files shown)


lib/resources/postgres.rb
```diff
@@ -3,6 +3,7 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 # license: All rights reserved
+require 'resources/directory'
```
chris-rock: we do not need a require here, since its available by default via `inspec.directory`

lib/resources/postgres.rb
```diff
@@ -12,29 +13,111 @@ class Postgres < Inspec.resource(1)
...
+        # and RHEL7. However, PR #824 explicitly added version-based
+        # directories. Thus, we call #version_from_dir unless it looks
+        # like we are using unversioned directories.
+        #
+        # TODO(ssd): This has the potential to be noisy because of the
+        # warning in version_from_dir. We should determine which case
+        # is more common and only warn in the less common case.
+        #
+        version = if inspec.directory("/var/lib/pgsql/data").exist?
+                    warn "Found /var/lib/pgsql/data. Assuming postgresql install uses un-versioned directories."
```
chris-rock: we should come up with a better debug logging. This is a good example.

chris-rock:
This is a great addition @stevendanna

lib/resources/postgres.rb
stevendanna: I agree.  I was gonna chat about this when more people are online because the logging style in this project in this project is a bit different than are usual ruby pattern.  Here we seem to create new per-class loggers with the option to pass them in to the constructor.  I like that pattern for some things, but for logging I think it clutters the method signatures and makes it hard to retrofit logging into existing classes/APIs.

In other projects we tend to prefer a global logging singleton object that we configure at the start of the application. So that adding logging to a class is as easy as:

```
require 'inspec/log'
Inspec::Log.warn("foo")
```

I noticed that inspec _does_ have such a logging class already, but it isn't really used anywhere.

Issue #240, opened 2015-11-16
Support writing full tests in inspec shell


arlimus:
At the moment it is not possible to start an `inspec shell` and paste either a `control` or a `describe` block and have it execute and report results.

I.e:

``` bash
inspec shell
```

And then enter:

``` ruby
describe file('/not/here') do
   it { should_not exist }
end
```

which will only result in:

``` ruby
=> [["describe", [File /not/here], #<Proc:0x000000013f3850@(pry):1>]]
```

The desired output should be:
1. A print of the execution using the default cli formatter, without the additonal profile context and version. i.e.:
   
   ``` ruby
     ✔  ssh-1: Allow only SSH Protocol 2
   ```
2. The test should return a result object, which has a result value embedded.

PR #909, opened 2016-08-12
Improve inspec shell by having it evaluate describe/control blocks.

ksubrama:
@arlimus @chris-rock @jbussdieker @alexpop 

This definitely has bugs in it but the is the general strategy I came up with last night after diving head-first into inspec and pry.

### DIFF
--- a/lib/bundles/inspec-compliance/target.rb
+++ b/lib/bundles/inspec-compliance/target.rb
@@ -15,6 +15,7 @@ module Compliance
     priority 500
 
     def self.resolve(target, _opts = {})
+      return nil unless target.is_a?(String)
       # check for local scheme compliance://
       uri = URI(target)
       return nil unless URI(uri).scheme == 'compliance'
--- a/lib/bundles/inspec-supermarket/target.rb
+++ b/lib/bundles/inspec-supermarket/target.rb
@@ -13,11 +13,12 @@ module Supermarket
     priority 500
 
     def self.resolve(target, opts = {})
+      return nil unless target.is_a?(String)
       return nil unless URI(target).scheme == 'supermarket'
       return nil unless Supermarket::API.exist?(target)
       tool_info = Supermarket::API.find(target)
       super(tool_info['tool_source_url'], opts)
-    rescue URI::Error => _e
+    rescue URI::Error
       nil
     end
 
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -159,15 +159,22 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
   def shell_func
     diagnose
     o = opts.dup
-    o[:logger] = Logger.new(STDOUT)
+
+    log_device = opts['format'] == 'json' ? nil : STDOUT
+    o[:logger] = Logger.new(log_device)
     o[:logger].level = get_log_level(o.log_level)
+
     if o[:command].nil?
       runner = Inspec::Runner.new(o)
       return Inspec::Shell.new(runner).start
     else
       res = run_command(o)
-      jres = res.respond_to?(:to_json) ? res.to_json : JSON.dump(res)
-      puts jres
+      if opts['format'] == 'json'
+        jres = res.respond_to?(:to_json) ? res.to_json : JSON.dump(res)
+        puts jres
+      else
+        puts res
+      end
     end
   rescue RuntimeError, Train::UserError => e
     $stderr.puts e.message
--- a/lib/inspec/runner_rspec.rb
+++ b/lib/inspec/runner_rspec.rb
@@ -15,6 +15,10 @@ module Inspec
     def initialize(conf)
       @conf = conf
       @formatter = nil
+      reset
+    end
+
+    def reset
       reset_tests
       configure_output
     end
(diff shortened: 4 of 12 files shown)


ksubrama:
Note - the "-c" arm of the shell already works (I have tested it locally).  I'm trying to figure out how to make the shell work.

lib/bundles/inspec-supermarket/target.rb
```diff
@@ -13,11 +13,11 @@ class Fetcher < Fetchers::Url
     priority 500
 
     def self.resolve(target, opts = {})
-      return nil unless URI(target).scheme == 'supermarket'
+      return nil unless target.is_a?(String)
```
chris-rock, line 16: we still need to ensure that the scheme is `supermarket`

lib/bundles/inspec-compliance/target.rb
```diff
@@ -15,6 +15,7 @@ class Fetcher < Fetchers::Url
     priority 500
 
     def self.resolve(target, _opts = {})
+      return nil unless target.is_a?(String)
```
chris-rock, line 18: 👍

lib/inspec/runner_rspec.rb
```diff
@@ -87,17 +87,17 @@ def report
...
-    private
-
     # Empty the list of registered tests.
     #
     # @return [nil]
     def reset_tests
       @tests = RSpec::Core::World.new
       # resets "pending examples" in reporter
-      RSpec.configuration.reset
+      # RSpec.configuration.reset
```
chris-rock: why is this not executed anymore and moved to `configure_output`?

lib/inspec/cli.rb
```diff
@@ -185,9 +184,9 @@ def version
   private
 
   def run_command(opts)
-    opts[:test_collector] = Inspec::RunnerMock.new
     runner = Inspec::Runner.new(opts)
-    runner.create_context.load(opts[:command])
+    runner.add_target({'in_mem.rb' => opts[:command]}, opts)
```
chris-rock: if we add a name, we should name it `in_memory.rb`

lib/inspec/cli.rb
```diff
@@ -153,19 +153,18 @@ def detect
...
     diagnose
     o = opts.dup
     o[:logger] = Logger.new(STDOUT)
     o[:logger].level = get_log_level(o.log_level)
     if o[:command].nil?
       runner = Inspec::Runner.new(o)
-      return Inspec::Shell.new(runner).start
+      return Inspec::Shell.new(runner).start(o)
     else
-      res = run_command(o)
```
chris-rock, line 168: I am not sure we should remove that. cc @arlimus

lib/inspec/shell.rb
```diff
@@ -3,21 +3,28 @@
...
-    def start
-      # store context to run commands in this context
-      c = { content: 'binding.pry', ref: nil, line: nil }
-      @runner.add_content(c, [])
-      @runner.run
+    def start(opts)
+      # Create an in-memory empty profile so that we can add tests to it later.
+      # This context lasts for the duration of this "start" method call/pry
+      # session.
+      @ctx = @runner.add_target({'shell_context.rb' => ''}, opts)
```
chris-rock: I am having some concerns with adding empty mock files. Investigating alternative solutions ...

ksubrama:
Fix #240

lib/inspec/runner_rspec.rb
ksubrama: I restructured this.  Essentially - I wanted to reset_tests in the test_collector without resetting the output configuration and json formatter stuff.  But I found a more explicit way to do that.

lib/inspec/cli.rb
```diff
@@ -163,11 +163,9 @@ def shell_func
     o[:logger].level = get_log_level(o.log_level)
     if o[:command].nil?
       runner = Inspec::Runner.new(o)
-      return Inspec::Shell.new(runner).start
+      return Inspec::Shell.new(runner, o).start
     else
-      res = run_command(o)
```
chris-rock, line 168: we should reflect the `--format` option here and do the json dump if activated

chris-rock:
This fixes #240

lib/inspec/cli.rb
ksubrama, line 168: run_command should already print the json output.  It's no longer using a mock runner - it's doing the real rspec thing which handles --format on its own.

chris-rock:
Some test cases for that ticket. We should add an automatic test for the shell in another iteration.
```
# simple resources
file('/tmp').exist?
file('tmp/file').exist?

describe file('/tmp') do
  it { should exist}
end

describe file('tmp/file') do
  it { should_not exist}
end

control 'tmp' do
  title 'tmp directory exists'
  describe file('/tmp') do
    it { should exist}
  end
end

control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end

res = describe file('tmp/file') do
  it { should_not exist}
end

shell with json output
b inspec shell -c "passwd.params" --format json
b inspec shell -c "file('/tmp')" --format json


# two controls with same name
control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end
control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end

# result via -c
inspec shell -c 'describe file("/Users/ksubramanian") do it { should exist } end'
```

PR #912, opened 2016-08-12
Windows ports with pid and process name

alexpop:
This PR improves the port resource for Windows. Two main changes went in:
1. Only LISTENING ports are retrieved same as with linux
2. `netstat` is used in favor of `Get-NetTCPConnection` because it gives us the valuable PID and process name behind the listening port.


alexpop:
```
 #<struct  port=49156, address="::", protocol="tcp", process="lsass.exe", pid=484, listening?=nil>,
 #<struct  port=5355, address="0.0.0.0", protocol="udp", process="Dnscache", pid=992, listening?=nil>,
 #<struct  port=137, address="10.0.2.15", protocol="udp", process="System", pid=4, listening?=nil>,
```

arlimus:
This is a great improvement, thank you @alexpop !! :+1:

PR #916, opened 2016-08-15
make integration tests in travis optional

chris-rock:


arlimus:
I would opt against opening this up for all tests.

IMHO: if we have failures on one operating system, which we then cannot resolve, can it be added to the list of accepted failures. The MR should highlight why it is being accepted and have it only for a period of time.

👎

PR #919, opened 2016-08-15
ssh_config parse should be case insensitive

vjeffrey:
https://github.com/chef/inspec/issues/759
screen shot 2016-08-15 at 4 05 33 pm

Please let me know if there's a cleaner way to do this. I looked around and tried to find a different way (one that doesn't involve me messing with the hash), but this was the only solution I could find. Happy to change it up if there's something better out there :)

### DIFF
--- a/test/unit/resources/ssh_conf_test.rb
+++ b/test/unit/resources/ssh_conf_test.rb
@@ -14,6 +14,12 @@ describe 'Inspec::Resources::SshConf' do
       _(resource.Tunnel).must_equal nil
       _(resource.SendEnv).must_equal 'LANG LC_*'
       _(resource.HashKnownHosts).must_equal 'yes'
+      puts resource
+    end
+
+    it 'is case insensitive' do
+      resource = load_resource('ssh_config')
+      _(resource.gssapiauthentication).must_equal 'no'
     end
   end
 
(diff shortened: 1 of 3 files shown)


test/unit/resources/ssh_conf_test.rb
```diff
@@ -14,6 +14,12 @@
       _(resource.Tunnel).must_equal nil
       _(resource.SendEnv).must_equal 'LANG LC_*'
       _(resource.HashKnownHosts).must_equal 'yes'
+      puts resource
+    end
+
+    it 'is case insensitive' do
+      resource = load_resource('ssh_config')
+      _(resource.gssapiauthentication).must_equal 'no'
```
stevendanna, line 21: You might consider adding an assertion here testing other capitalizations.  This test would fail to catch a regression where we only accepted lowercase specifications.

test/unit/resources/ssh_conf_test.rb
```diff
@@ -14,6 +14,12 @@
       _(resource.Tunnel).must_equal nil
       _(resource.SendEnv).must_equal 'LANG LC_*'
       _(resource.HashKnownHosts).must_equal 'yes'
+      puts resource
```
stevendanna: We probably want to clean up this debugging output.

stevendanna:
Overall this looks good to me.  I think this is a good solution for now.  If we wanted to get fancier, rather than storing the configuration in a plain ruby hash, we could create a wrapper data type that intermediated access to the underlying data.  However, until we have another use case for that, I think the one-time transformation is relatively clean.

vjeffrey:
thanks!! just pushed up a commit to test with resource.GSSAPIAuthentication and removed that debugging cruft

PR #921, opened 2016-08-16
Fix errant 0 exit status on SSH errors

### LINKED ISSUE #840: `inspec` binary occasionally exits zero on SSH failures
### Description

If the underlying `ssh` command/transport has an error the `inspec` binary still exits zero. This can prove problematic when CI/CD is driving `inspec` as the job will be marked as green when it actually failed because of underlying config issues.

```
$ inspec exec inspec/package-router-smoke -t ssh://ubuntu@package-router-0-acceptance.es.chef.co -i ~/.ssh/es-infrastructure.pem
ssh: Could not resolve hostname bastion-verify.es.chef.co: nodename nor servname provided, or not known
$ echo $?
0
```

In the scenario above a `ProxyCommand` is configured in `~/.ssh/config` BUT inadvertently pointing at a non-existent bastion/gateway host:

```
Host package-router-*-acceptance.es.chef.co
  ProxyCommand ssh -A ubuntu@bastion-verify.es.chef.co -W %h:%p
```

A similar "raw" `ssh` command does return a non-zero exit code:

```
$ ssh ubuntu@package-router-0-acceptance.es.chef.co
ssh: Could not resolve hostname bastion-verify.es.chef.co: nodename nor servname provided, or not known
ssh_exchange_identification: Connection closed by remote host
$ echo $?
255
```

```
InSpec version: 0.27.1
Train version: 0.15.1
```

stevendanna:
Thor, the command line framework we use, rescues Errno::EPIPE
exceptions and exits 0.  It does this because for many unix-style
command line applications it is a good decision.  For example, if your
command is being used in a pipeline with head(1):

```
my_command | head -10
```

It will get an EPIPE after head exits. You don't want this command to
fail because of the EPIPE as it is what you expected.

However, in our case, the SSH transport will raise EPIPE if the remote
server unexpectedly closed the connection.  This can happen, for
example, if your ssh configuration has a bad ProxyCommand configuration.
In these cases we do not want to exit successfully.

Explicitly rescuing the EPIPE in Inspec::Backend#create solves the issue
reported by users, but it may be better to rescue this exception in
Train itself.

Fixes #840

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -42,6 +42,10 @@ module Inspec
       raise "Client error, can't connect to '#{name}' backend: #{e.message}"
     rescue Train::TransportError => e
       raise "Transport error, can't connect to '#{name}' backend: #{e.message}"
+    rescue Errno::EPIPE => e
+      # The SSH backend may raise an EPIPE if the remote end closes
+      # the connection unexpectedly.
+      raise "Can't connect to '#{name}' backend: #{e.message}"
     end
   end
 end


stevendanna:
I'm looking at perhaps moving this fix lower down in train itself.

chris-rock:
As you already proposed, I think this error should be fixed in train and not InSpec. We already handle a train error in InSpec, and InSpec should not know about anything else. I assume we handle it here: https://github.com/chef/train/blob/master/lib/train/transports/ssh_connection.rb#L111 or https://github.com/chef/train/blob/master/lib/train/transports/ssh_connection.rb#L185

stevendanna:
PR to rescue this in train on connect is here: https://github.com/chef/train/pull/130

chris-rock:
Thanks @stevendanna for the work, I am going to close this in favor of your implementation in train https://github.com/chef/train/pull/130

PR #922, opened 2016-08-16
add kitchen.yml for non-public kitchen boxes

chris-rock:
This PR adds macos, windows and solaris to our kitchen environment to ease integration testing. These boxes are only available for chef employees. Therefore its not merged into `.kitchen.yml` and uses `.kitchen.chef.yml`. To run integration test against those machines run:

```
KITCHEN_YAML=.kitchen.chef.yml bundle exec kitchen test default-windows-2012r2-standard
```

### DIFF
--- a/.kitchen.chef.yml
+++ b/.kitchen.chef.yml
@@ -1,0 +1,67 @@
+---
+driver:
+  name: vagrant
+
+provisioner:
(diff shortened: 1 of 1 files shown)


stevendanna:
LGTM

PR #928, opened 2016-08-17
fix integration tests for Chef Compliance

chris-rock:
As @vjeffrey pointed out, the integration test have not produced the expected results. This PR:
- updates the docs
- names the environment variables for tokens as in https://github.com/chef-cookbooks/audit/tree/master/examples/kitchen
- installs a separate ruby environment for InSpec, to work independently of Chef Compliance ruby environment


vjeffrey:
wooohoooooo it works!!!! 🎉 🌟  thank youuuuu!!!

Issue #888, opened 2016-08-08
RFC Dependencies


arlimus:
Follow up https://github.com/chef/inspec/issues/798 around dependency resolution.
## Design

**(1)** dependencies are written in `inspec.yml`:

``` yaml
depends:
  - name: hello
  [ ... ]
```

**(2)** dependencies are based on semantic versioning: http://semver.org/ . Dependencies do not support alpha/pre-release information ([9](http://semver.org/#spec-item-9), [10](http://semver.org/#spec-item-10)) (see [comment](https://github.com/chef/inspec/issues/888#issuecomment-238934145))
**(3)** dependency version constraints are specified with: 

``` yaml
- version: [op] [version] [...]
```

with `op` being `<`, `>`, `<=`, `>=`, `=`, `!=` or `~>`. Multiple constraints may be specified. Whitespaces are optional. Whitespaces in version numbers are not supported.

**(4)** each dependency has one source, with the default pointing to `supermarket`. sources can be specified via:

``` yaml
- path: ../relative_path/to/profile
- path: /absolute/path
- supermarket: owner/name
- compliance: owner/name
- github: owner/name
- url: http://sth...
```

**(5)** upon resolution, a lockfile is created in `inspec.lock`. If this lockfile exists, dependencies are not resolved but taken from this file intead.

**(6)** dependencies are vendored to a local cache in `${inspec-home}/cache`. Users may specify a custom vendor location.

**(7)** dependencies provide their library functions the profile (without additional specification). For example: Resources that are defined in libraries are available to the profile that requires it.  

**(8)** dependencies may provide controls via `include_controls` (all) or `require_controls` (selective). If none of these are used, dependencies will not execute or report on any controls.

**(9)** dependencies are scoped with their name. scoping applies to all aspects, including resources, controls, and attributes. resources are added to the global space if there is no conflict.

... and discuss :grin: 
### scoping
- All profiles have a simple name. Names are short, to the point (and should not contain spaces). Example: `ssh`, `cis-centos6-lvl1`
- When a profile is pulled in via dependencies, its name may be overwritten. This allows the inclusion of 2 profiles with the same name. Example: my `ssh` becomes `my-ssh` while upstream `ssh` becomes `upstream-ssh`. It's done via the `name` field: 
  
  ```
  depends:
  - name: my-ssh
    url: go.to/my/ssh
  ```
[...]

PR #929, opened 2016-08-17
Fix recursive deps for path-based deps

stevendanna:
Ensure that we always instantiate an Inspec::Requirement and thread through the current working directory of the inspec.yml.

### DIFF
--- a/lib/inspec/dependencies.rb
+++ b/lib/inspec/dependencies.rb
@@ -6,16 +6,17 @@ require 'logger'
 require 'fileutils'
 require 'molinillo'
 require 'inspec/errors'
+require 'inspec/requirement'
 
 module Inspec
   class Resolver
     def self.resolve(requirements, vendor_index, cwd, opts = {})
       reqs = requirements.map do |req|
-        Requirement.from_metadata(req, cwd: cwd) ||
-          fail("Cannot initialize dependency: #{req}")
+        req = Inspec::Requirement.from_metadata(req, cwd: cwd)
+        req || fail("Cannot initialize dependency: #{req}")
       end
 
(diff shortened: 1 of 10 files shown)


lib/inspec/dependencies.rb
```diff
@@ -6,23 +6,25 @@
...
-      new(vendor_index, opts).resolve(reqs)
+      new(vendor_index, opts.merge(cwd: cwd)).resolve(reqs)
     end
 
     def initialize(vendor_index, opts = {})
       @logger = opts[:logger] || Logger.new(nil)
       @debug_mode = false # TODO: hardcoded for now, grab from options
 
       @vendor_index = vendor_index
+      @cwd = opts[:cwd] || './'
```
chris-rock, line 27: I would move that to the cli and set the default there, not as part of the code.

examples/inheritance/controls/example.rb
```diff
@@ -9,3 +9,6 @@
     impact 0.0
   end
 end
+
```
chris-rock: The example is great, I think we should not expose that to our customer examples yet. Could we move those to `test/unit/mock/profiles`?

lib/inspec/dependencies.rb
stevendanna, line 27: Maybe we can reuse the profile path here to avoid adding a new option?  Overall I think we'll need to refactor a lot of this current working directory as it is currently very confusing.  Once we have the fetchers/vendoring more flushed out, I think this level of the code won't need to worry about this.

chris-rock, line 27: 👍

chris-rock:
related to #888

PR #930, opened 2016-08-17
fail gracefully on inspec compliance profiles when bad token is provided

vjeffrey:
just a quick fix to handle inspec compliance profiles when an incorrect token has been previously provided.  
screen shot 2016-08-17 at 12 18 02 pm
@chris-rock

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -70,7 +70,7 @@ Please login using `inspec compliance login https://compliance.test --user admin
 
     # verifies that a profile
     def self.exist?(config, profile)
-      profiles = Compliance::API.profiles(config)
+      _msg, profiles = Compliance::API.profiles(config)
       if !profiles.empty?
         index = profiles.index { |p| "#{p[:org]}/#{p[:name]}" == profile }
         !index.nil? && index >= 0
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -27,16 +27,24 @@ module Compliance
       # TODO, api should not be dependent on .supported?
       response = Compliance::HTTP.get(url, config['token'], config['insecure'], !config.supported?(:oidc))
       data = response.body
-      if !data.nil?
+      response_code = response.code
+      case response_code
+      when '200'
+        msg = 'success'
         profiles = JSON.parse(data)
         # iterate over profiles
-        profiles.map do |owner, ps|
+        mapped_profiles = profiles.map do |owner, ps|
           ps.keys.map do |name|
             { org: owner, name: name }
           end
         end.flatten
+        return msg, mapped_profiles
+      when '401'
+        msg = '401 Unauthorized. Please check your token.'
+        return msg, []
       else
-        []
+        msg = "An unexpected error occurred (HTTP #{response_code}): #{response.message}"
+        return msg, []
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/bundles/inspec-compliance/api.rb
```diff
@@ -36,6 +36,7 @@ def self.profiles(config)
           end
         end.flatten
       else
+        puts '401 Unauthorized. Please check your token.'
```
stevendanna: Are we sure the only case where `data == ''` is a 401?  My first impression when reading this code is that I'm a bit surprised that this isn't a case statement on the http response code. Also, if possible it feels like this is a case where we would want to exit with a non-zero exit status.

vjeffrey: good point, thanks!! ya, i'll rework this a bit later -- a case statement on the http response code sounds like the right thing to do there

chris-rock: agree with @stevendanna we should check the http code

lib/bundles/inspec-compliance/api.rb
```diff
@@ -27,17 +27,22 @@ def self.profiles(config)
...
           ps.keys.map do |name|
             { org: owner, name: name }
           end
         end.flatten
-      else
+      when '401'
         puts '401 Unauthorized. Please check your token.'
         []
+      else
+        puts response_code
```
chris-rock: We should add a message that indicates an error. I assume most users do not know http codes by heart ;-)

vjeffrey:
i'm getting some odd travis failures on ubuntu.... :(

lib/bundles/inspec-compliance/api.rb
```diff
@@ -27,15 +27,21 @@ def self.profiles(config)
...
         profiles.map do |owner, ps|
           ps.keys.map do |name|
             { org: owner, name: name }
           end
         end.flatten
+      when '401'
+        puts '401 Unauthorized. Please check your token.'
+        []
       else
+        puts response_code, 'An error occured'
```
vjeffrey: is this ok? @chris-rock  or were you thinking something else/more specific?

lib/bundles/inspec-compliance/api.rb
vjeffrey: oops, ya! i'll add it :)

lib/bundles/inspec-compliance/api.rb
stevendanna: I'm not familiar with the compliance API, but `response.message` will typically be filled with the error message that the server sent back, so you might try something like:

```
puts "An unexpected error occurred (HTTP #{response_code}): #{response.message}"
```

chris-rock: Cool idea @stevendanna

chris-rock: In addition to that the API class should not do any output. Only the cli class should do that. therefore we need to return the message as the other methods do it.

stevendanna: > Only the cli class should do that. therefore we need to return the message as the other methods do it.

👍 On this.  I was going to potentially recommending creating a custom exception that you could raise in the case of an HTTP error.  The exception can contain the message the cli class should print on rescue.  However, I noticed that this code base stays away from exceptions for the most part.

vjeffrey: oooh nice! thanks! i'll add the message stuff and have the api class the return the message so the cli class can take care of the output :)

chris-rock: @stevendanna the whole API implementation needs a refactor, because I want to merge it with the content in our audit cookbook. But lets not address this in that iteration

PR #932, opened 2016-08-18
return token stored message on login

vjeffrey:
at the moment, on master, we respond with 'successfully authenticated' even when there's an obviously wrong token, so this is just a small change to send back the actual message instead: 
screen shot 2016-08-17 at 8 12 49 pm

### DIFF
--- a/lib/bundles/inspec-compliance/cli.rb
+++ b/lib/bundles/inspec-compliance/cli.rb
@@ -30,26 +30,22 @@ module Compliance
       url = options['server'] + options['apipath']
       if !options['user'].nil? && !options['password'].nil?
         # username / password
-        success, msg = login_legacy(url, options['user'], options['password'], options['insecure'])
+        _success, msg = login_legacy(url, options['user'], options['password'], options['insecure'])
       elsif !options['user'].nil? && !options['token'].nil?
         # access token
-        success, msg = store_access_token(url, options['user'], options['token'], options['insecure'])
+        _success, msg = store_access_token(url, options['user'], options['token'], options['insecure'])
       elsif !options['refresh_token'].nil? && !options['user'].nil?
         # refresh token
-        success, msg = store_refresh_token(url, options['refresh_token'], true, options['user'], options['insecure'])
+        _success, msg = store_refresh_token(url, options['refresh_token'], true, options['user'], options['insecure'])
         # TODO: we should login with the refreshtoken here
       elsif !options['refresh_token'].nil?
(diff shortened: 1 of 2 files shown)


lib/bundles/inspec-compliance/cli.rb
```diff
@@ -30,26 +30,22 @@ def login(server) # rubocop:disable Metrics/AbcSize
       url = options['server'] + options['apipath']
       if !options['user'].nil? && !options['password'].nil?
         # username / password
-        success, msg = login_legacy(url, options['user'], options['password'], options['insecure'])
+        msg = login_legacy(url, options['user'], options['password'], options['insecure'])
       elsif !options['user'].nil? && !options['token'].nil?
         # access token
-        success, msg = store_access_token(url, options['user'], options['token'], options['insecure'])
+        msg = store_access_token(url, options['user'], options['token'], options['insecure'])
```
stevendanna: `store_access_token` returns an array, so now `msg` will be an array (https://github.com/chef/inspec/pull/932/files#diff-e2fd7d8d44ceb659a9c3a3a7984c53a4R237), which will cause the print statement below to print both items:

screen shot 2016-08-18 at 9 31 02 am

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -30,26 +30,22 @@ def login(server) # rubocop:disable Metrics/AbcSize
...
         puts 'Please run `inspec compliance login` with options --token or --refresh_token and --user'
         exit 1
       end
 
-      if success
-        puts '', 'Successfully authenticated'
-      else
-        puts msg
-      end
+      puts '', msg
```
stevendanna, line 48: Why do we want to print this extra blank line?

lib/bundles/inspec-compliance/cli.rb
vjeffrey: oh, oops, sorry. i took that out cause of a lint complaint, I should have just left it in with a leading underscore....late night changes :)  I'll switch it up

lib/bundles/inspec-compliance/cli.rb
vjeffrey, line 48: So, I wasn't sure about doing the extra blank line or not, it's just a user-readability thing.  When you paste that big token in and then it spits out the message, it's kinda hard (imo) to see the message under the block of text, hence the extra line, but I'm not attached to it, so if ppl think it's fine the way it is/if it's not a generally cool thing to spit out an extra line for something like this, I would be happy to remove the extra line.

PR #933, opened 2016-08-18
restructure unit tests

chris-rock:
This PR restructures unit tests, so that the intention of the test is more clear


stevendanna:
Seems reasonable to me. 👍

PR #934, opened 2016-08-18
Ease testing of compliance integration tests

chris-rock:
- use bundler instead of gem
- this help to make local changes and use them directly with your integration tests


vjeffrey:
thank youuuuuuu! this makes it soooo much easier :)  👍

Issue #915, opened 2016-08-15
Transitive dependency loading


chris-rock:
### Description

As discussed in #888, we'll support transitive dependencies (A pulls B pulls C). All urls need to be vendored in a way that they do not conflict on disc. The RFC describes two phases, a resolution phase and an execution phase. This ticket describes the resolution phase, where required parts for execution are vendored locally. Assume we have the following `inspec.yml`

_inspec.yml_

```
name: inheritance
version: 1.0.0
supports:
  - os-family: unix
depends:
  - name: profile
    path: ../profile
  - name: ssh 
    github: dev-sec/tests-ssh-hardening
  - name: os
    supermarket: dev-sec/os-hardening
```

During resolution phase, we resolve all dependencies and built up a dependency graph. This graph is required to implement proper scoping for attributes, controls and resources later on. A sample `inspec.lock` could look like:

_inspec.lock_

The lock file is a yml file as well and includes resolved locations

```
dependencies:
  - name: profile
    path: ../profile
  - name: ssh 
    url: https://github.com/dev-sec/tests-ssh-hardening/archive/c2463115b3f7c254cbab9ddd7a7d215cdb843de1.zip
  - name: os
    git: https://github.com/dev-sec/tests-os-hardening.git
    revision: da3a1b6ce8a845d6818152a824e123c2445c355f

graph:
  inheritance:
    uses:
      profile 
      ssh
        uses:
          otherprofile
      os
```

InSpec should download all dependencies in a local cache directory. The global cache can be used with the `-g` or `--global` option.
- `/current/pwd/.cache/name-*sha256*`
- `~/.inspec/.cache/name-*sha256*`

The following commands can be used to download all dependencies:
- `inspec vendor` - is creating the lock file and vendors all dependencies into a local cache
- `inspec vendor -g` - is creating the lock file and vendors all dependencies into the global cache 

_Dependency resolution mechanism_

Instead of a constraint-resolution mechanism, we use a mechanism to fetch the latest-suitable version. This will require us to allow the loading of multiple profiles with the same name but different version at the same time. Eg. if we have a version dependency `>= 2.3` and the following versions are available: `2.4`, `2.3.1`, `2.2`, we would always fetch `2.4`. If a dependency in that tree depends on a specific version, we would load this as well.

Some open questions:
- should the cache only contain zip/tar files (Would imply that we archive a local path before execution?)
- if I run `inspec exec http://url/to/my/profile`, should we cleanup the cache at the end?
[...]

PR #935, opened 2016-08-18
Fetch deps based on urls

stevendanna:
This extends the dependency feature to include support for url-based
dependencies.  It takes some deviations from the current support for
URLs that we'll likely want to make more consistent.

By default, we store downloaded archives in the cache rather than the
unpacked archive. However, to facilitate debugging, we will prefer the
unpacked archive if we find it in the cache.

This partially addresses #915 by:
- Supporting URL based sources (since most of our sources will resolve to URL based sources, the rest of the sources should be straightforward additions)
- Storing downloaded dependencies in the vendor cache
- Using the cached dependencies when they exist

Still to do:
- Integrate other source types
- Write out a `inspec.lock` file based on the fully resolved dependencies
- Read the lock file rather than inspec.yml when it exists


chris-rock:
Great work @stevendanna I really like the split up into smaller pieces. I would prefer to store all dependency related classes under `lib/inspec/dependencies`

lib/inspec/dependencies.rb
```diff
@@ -1,234 +1,8 @@
 # encoding: utf-8
-# author: Dominik Richter
-# author: Christoph Hartmann
-
-require 'logger'
-require 'fileutils'
-require 'molinillo'
-require 'inspec/errors'
-require 'inspec/requirement'
+require 'inspec/vendor_index'
```
chris-rock: 👍 on the splitup

lib/inspec/requirement.rb
```diff
@@ -16,15 +20,66 @@ def matches_spec?(spec)
...
+                         :local_path
+                       elsif opts[:url]
+                         :url
+                       else
+                         fail "Cannot determine source type from #{opts}"
+                       end
+    end
+
+    def fetcher_class
+      @fetcher_class ||= case source_type
```
chris-rock: Should we reuse our fetcher detection?

```
      fetcher = Inspec::Fetcher.resolve(target)
      if fetcher.nil?
        fail("Could not fetch inspec profile in #{target.inspect}.")
      end
```

chris-rock: see https://github.com/chef/inspec/blob/master/lib/inspec/profile.rb#L20-L26

chris-rock: Let me know if you think we should merge those implementations in another iteration.

chris-rock:
We need to be careful with the cwd:

```
inspec exec test/unit/mock/profiles/dependencies-file/inheritance 
Could not fetch inspec profile in "/Users/chartmann/Development/compliance/profile_a".
```

stevendanna:
@chris-rock Re the current working directory, I was going back and forth about what people expect there.  Currently, as you can see, it is using the current directory, but we could get the base directory of the inspec.yml at the head of the tree if you think that is what people would expect.

test/unit/mock/profiles/dependencies-file/inheritance/inspec.yml
```diff
@@ -10,3 +10,5 @@ depends:
     path: ../profile_a
   - name: profile_b
     path: ../profile_b
+  - name: os-hardening
```
chris-rock, line 13: we should rename the parent directory to `depndencies`, since we are testing more complex content here that files. we may split that up later

lib/inspec/vendor_index.rb
```diff
@@ -0,0 +1,85 @@
+# encoding: utf-8
+require 'digest'
+require 'fileutils'
+
+module Inspec
+  class VendorIndex
```
chris-rock: Lets add some global description what this class should do

lib/inspec/vendor_index.rb
```diff
@@ -0,0 +1,85 @@
...
+    # The `source_url` parameter should be a URI-like string that
+    # fully specifies the source of the exact version we want to pull
+    # down.
+    #
+    # @param [String] name
+    # @param [String] source_url
+    # @return [String]
+    #
+    def key_for(name, source_url)
+      source_hash = Digest::SHA256.hexdigest source_url
```
chris-rock: +1

chris-rock:
I think, users would expect that they are relative to the partent profile location. I just tried to execute it that way as I normally execute profiles

chris-rock:
@stevendanna Could you do me a favor and highlight, which parts of #915 are addressed in that PR?

stevendanna:
@chris-rock I've updated this with:
- Requested renames
- A small fix for the relative paths (your example should now work)
- Basic comments for each new class

I am 👍 on the change you recommended for fetcher but I think we should do that in a follow up PR.

chris-rock:
related to #888

PR #937, opened 2016-08-18
Document awesome where syntax for port

pburkholder:


chris-rock:
Thanks for the improvement @pburkholder

PR #940, opened 2016-08-19
Should be able to check windows service startup mode

Anirudh-Gupta:
Allows one to check windows services start up mode : https://msdn.microsoft.com/en-us/library/aa384896(v=vs.85).aspx
Services in Windows could be installed but may have to be manually started.
For example: 

``` ruby
describe service('AppMgmt') do
  it { should be_installed }
  it { should_not  be_running }
  its('startmode') { should be 'Manual'} 
end
```

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -32,6 +32,10 @@ module Inspec::Resources
       describe host('example.com') do
         it { should be_reachable }
       end
+
+      describe host('example.com', port: '80') do
+        it { should be_reachable }
+      end
     "
 
     def initialize(hostname, params = {})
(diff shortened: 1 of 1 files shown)


Anirudh-Gupta:
@arlimus What other changes do we require for this?

ksubrama:
Thank you for this contribution!  Looks good to me.

lib/resources/service.rb
```diff
@@ -587,6 +594,7 @@ def info(service_name)
         installed: true,
         running: service_running?(service),
         enabled: service_enabled?(service),
+        startmode: !service['WMI']['StartMode'].nil? ? service['WMI']['StartMode'] : nil,
```
ksubrama: Seems like we don't need to check for nil - if it is nil, we assign nil, otherwise we assign whatever non-nil value we get.  It's just a cosmetic change though - i don't think this should hold up merging this PR.

chris-rock:
@Anirudh-Gupta Thanks for the addition. Could you rebase to master, please? I'll merge it then.

chris-rock:
@Anirudh-Gupta let me know if I can help with that. `git rebase -i master` is doing the trick in most cases. Let me know if I can help

Anirudh-Gupta:
@chris-rock opened #968 for the same.

Issue #942, opened 2016-08-19
Implement test cases for inspec shell


chris-rock:
We should add an automatic test for the shell. I can think about the following test cases

```
# simple resources
file('/tmp').exist?
file('tmp/file').exist?

describe file('/tmp') do
  it { should exist}
end

describe file('tmp/file') do
  it { should_not exist}
end

control 'tmp' do
  title 'tmp directory exists'
  describe file('/tmp') do
    it { should exist}
  end
end

control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end

res = describe file('tmp/file') do
  it { should_not exist}
end

shell with json output
b inspec shell -c "passwd.params" --format json
b inspec shell -c "file('/tmp')" --format json


# two controls with same name
control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end
control 'tmp2' do
  title 'tmp/file does not exist'
  describe file('tmp/file') do
    it { should_not exist}
  end
end

# result via -c
inspec shell -c 'describe file("/Users/ksubramanian") do it { should exist } end'
```

chris-rock:
I propose we create separate functional tests. Some are already available here: https://github.com/chef/inspec/blob/master/test/functional/inspec_test.rb#L37-L82. We should split them out into a separate file

PR #943, opened 2016-08-19
Fix command evaluation for inspec shell -c

ksubrama:


ksubrama:
Fixed #942

ksubrama:
@stevendanna @chris-rock @arlimus @vjeffrey Yay tests pass.  I think we have a good set of functional tests.  Clearly there can be more.  I kinda want to get these in before I spend more time over testing this.

lib/inspec/cli.rb
```diff
@@ -162,22 +162,23 @@ def shell_func
     diagnose
     o = opts.dup
 
-    log_device = opts['format'] == 'json' ? nil : STDOUT
+    json_output = ['json', 'json-min'].include?(opts['format'])
```
arlimus, line 165: Great catch!!

arlimus:
These look great, especially the large selection of additional tests. There can always be more, but the tricky part is to select the right amount of tests and I like the selection you've made.

The only slight thing we should keep in mind is this:
https://github.com/chef/inspec/pull/943/files#diff-ca3dd31f0db2b13fca008d2b3a0000b8L138
Since this is a publicly available function that we are removing, it may cause issues in other projects. I do however like that it is being removed, i.e.:

:+1: 

Thank you Kartik, this is a great improvement :smile:

PR #944, opened 2016-08-19
Update platforms in the docs to match the code

tas50:
I was looking at the platform detection and realized some platforms were missing from the docs


chris-rock:
@tas50 thanks for improvement. Let me know when you are ready. I'll merge it then

tas50:
Should be good to go now. I'm still digging into some OS stuff, but that'll be another PR

tas50:
@chris-rock ready

chris-rock:
💯 @tas50

PR #949, opened 2016-08-22
Add prototype of inspec.lock

stevendanna:
This adds a basic prototype of inspec.lock. When the lockfile exists on
disk, the dependencies tree is constructed using the information in the
lock file rather than using the resolver.

### DIFF
--- a/lib/fetchers/tar.rb
+++ b/lib/fetchers/tar.rb
@@ -28,6 +28,14 @@ module Fetchers
       end
     end
 
+    def url
+      if parent
+        parent.url
+      else
+        'file://target'
+      end
+    end
+
     def read(file)
       @contents[file] ||= read_from_tar(file)
     end
--- a/lib/fetchers/tar.rb
+++ b/lib/fetchers/tar.rb
@@ -19,6 +19,10 @@ module Fetchers
       new(target)
     end
 
+    def archive_path
+      target
+    end
+
     def initialize(target)
       @target = target
       @contents = {}
--- a/lib/fetchers/tar.rb
+++ b/lib/fetchers/tar.rb
@@ -32,14 +32,6 @@ module Fetchers
       end
     end
 
-    def url
-      if parent
-        parent.url
-      else
-        'file://target'
-      end
-    end
-
     def read(file)
       @contents[file] ||= read_from_tar(file)
     end
--- a/lib/fetchers/zip.rb
+++ b/lib/fetchers/zip.rb
@@ -29,6 +29,14 @@ module Fetchers
       end
     end
 
+    def url
+      if parent
+        parent.url
+      else
+        'file://target'
+      end
+    end
+
     def read(file)
       @contents[file] ||= read_from_zip(file)
     end
--- a/lib/fetchers/zip.rb
+++ b/lib/fetchers/zip.rb
@@ -29,14 +29,6 @@ module Fetchers
       end
     end
 
-    def url
-      if parent
-        parent.url
-      else
-        'file://target'
-      end
-    end
-
     def read(file)
       @contents[file] ||= read_from_zip(file)
     end
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -93,6 +93,13 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
     exit 1 unless result[:summary][:valid]
   end
 
+  desc 'vendor', 'Download all dependencies and generate a lockfile'
+  def vendor(path = nil)
+    profile = Inspec::Profile.for_target('./', opts)
+    lockfile = profile.generate_lockfile(path)
+    File.write('inspec.lock', lockfile.to_yaml)
+  end
+
   desc 'archive PATH', 'archive a profile to tar.gz (default) or zip'
   profile_options
   option :output, aliases: :o, type: :string,
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -24,12 +24,16 @@ module Inspec
         Inspec::Requirement.from_lock_entry(dep, cwd, vendor_index)
       end
 
-      # Flatten tree because that is all we know how to deal with for
-      # right now.  Last dep seen for a given name wins right now.
       dep_list = flatten_dep_tree(dep_tree)
       new(cwd, vendor_path, dep_list)
     end
 
+    # This is experimental code to test the working of the
+    # dependency loader - perform a proper dependency related search
+    # in the future.
+    #
+    # Flatten tree because that is all we know how to deal with for
+    # right now. Last dep seen for a given name wins right now.
     def self.flatten_dep_tree(dep_tree)
       dep_list = {}
       dep_tree.each do |d|
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require 'inspec/dependencies/vendor_index'
+require 'inspec/dependencies/requirement'
 require 'inspec/dependencies/resolver'
 
 module Inspec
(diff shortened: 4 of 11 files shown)


lib/inspec/profile.rb
```diff
@@ -211,6 +212,32 @@ def locked_dependencies
...
+
+    def lockfile
+      @lockfile ||= if lockfile_exists?
+                      Inspec::Lockfile.from_file(lockfile_path)
+                    else
+                      generate_lockfile
+                    end
+    end
+
+    def generate_lockfile(vendor_path = nil)
```
ksubrama, line 247: Comment: Make this an in-memory operation that doesn't render the lock file to disk.

lib/inspec/dependencies/requirement.rb
```diff
@@ -24,64 +47,81 @@ def matches_spec?(spec)
...
-        else
-          archive = fetcher_class.download_archive(source_url)
-          @vendor_index.add(@name, source_url, archive.path)
-        end
+        @vendor_index.add(@name, source_url, fetcher.archive.path)
+      end
+    end
+
+    def dependencies
+      return @dependencies unless @dependencies.nil?
```
ksubrama: I'm pretty sure the answer is "it's demo code" but I was curious why this wasn't written using the standard pattern of `@dependencies ||= ....`  Can @dependencies be nil even after the first call?  If so, can we document that?

lib/inspec/dependencies/lockfile.rb
```diff
@@ -0,0 +1,86 @@
...
+    def to_yaml
+      {
+        'lockfile_version' => CURRENT_LOCKFILE_VERSION,
+        'depends' => @deps,
+      }.to_yaml
+    end
+
+    private
+
+    def parse_content_hash(lockfile_content_hash)
```
ksubrama, line 79: Comment: Refactor this to be "version-wise" - i.e. make one dispatch function for each version so that even if it duplicates code, it can describe the part of the code that it expects to be different.  Then that dispatch routine can call more well defined methods like "parse_v0_dependencies" or "parse_flat_dependencies" or what not as things generally develop.  It does help people easily set breakpoints/track different entry points of the API.

lib/inspec/dependencies/dependency_set.rb
```diff
@@ -10,17 +11,71 @@ module Inspec
...
+        Inspec::Requirement.from_lock_entry(dep, cwd, vendor_index)
+      end
+
+      # Flatten tree because that is all we know how to deal with for
+      # right now.  Last dep seen for a given name wins right now.
+      dep_list = flatten_dep_tree(dep_tree)
+      new(cwd, vendor_path, dep_list)
+    end
+
+    def self.flatten_dep_tree(dep_tree)
```
ksubrama, line 37: Comment:  This is experimental code to test the working of the dependency loader - perform a proper dependency related search in the future.

Basically the comment from line 27 but on this method because I tend to look at comments describing methods before I look at comments from their call sites.  Up to you though :)

ksubrama:
So far so good.  This makes sense as the "in progress" code that you walked me through this afternoon.  I'm sure a lot of it will be refactored/rewritten soon - hence why most of my comments have a "Comment:" tag in them to denote that I'd simply like you to write a comment to that effect in there in case we don't get around to fixing that bit (you can never be too paranoid). I am :+1: on this.

lib/inspec/profile.rb
stevendanna: I think it is already.  To render it would disk you'd have to do something like we do in the `vendor` function in cli.rb.

lib/inspec/dependencies/requirement.rb
stevendanna: Ah, this is mostly because this function used to be a lot longer so it would have been `@dependencies ||= begin` where the begin block was really big.  I don't think dependencies can be nil after this because 118 either will return a map or raise some NoMethodError.  I suppose it can be nil if someone calls lock_deps(nil), but in that case I feel like you should know what you are doing.

Anyway, I can move this to our more typical style.

lib/fetchers/zip.rb
```diff
@@ -29,6 +29,14 @@ def initialize(target)
       end
     end
 
+    def url
+      if parent
+        parent.url
+      else
+        'file://target'
```
chris-rock: should this be "file://#{target}"?

lib/fetchers/tar.rb
```diff
@@ -28,6 +32,14 @@ def initialize(target)
       end
     end
 
+    def url
```
chris-rock: it feels strange to have url and archive_path and target, since all refer to the same content.

lib/fetchers/tar.rb
```diff
@@ -28,6 +32,14 @@ def initialize(target)
...
+      if parent
+        parent.url
+      else
+        'file://target'
```
chris-rock: should this be "file://#{target}"?

stevendanna: Yeah, it is working right now because these are never really called.

lib/fetchers/tar.rb
stevendanna: Yeah, I think we can remove this for now.  This was part of an earlier organization of the code that I moved away from.  Overall, though, the goal was to eventually have the fetcher be responsible for knowing the fully resolved URL of a given target. So you could do something like `t = Inspec::Fetcher.resolve(target); t.url` and get a URL suitable to stick in a lockfile.

chris-rock:
I think the `inspec.lock` is not generated in the right path. It should be generated within the root profile directory

```
$ b inspec vendor test/unit/mock/profiles/dependencies/inheritance 
$ cat test/unit/mock/profiles/dependencies/inheritance/inspec.lock
cat: test/unit/mock/profiles/dependencies/inheritance/inspec.lock: No such file or directory
$ cat inspec.lock 
---
lockfile_version: 0
depends: []
```

lib/inspec/cli.rb
```diff
@@ -93,6 +93,13 @@ def check(path) # rubocop:disable Metrics/AbcSize
     exit 1 unless result[:summary][:valid]
   end
 
+  desc 'vendor', 'Download all dependencies and generate a lockfile'
+  def vendor(path = nil)
+    profile = Inspec::Profile.for_target('./', opts)
+    lockfile = profile.generate_lockfile(path)
```
chris-rock, line 99: 👍 that this is part of the profile

lib/inspec/cli.rb
```diff
@@ -93,6 +93,13 @@ def check(path) # rubocop:disable Metrics/AbcSize
...
+    File.write('inspec.lock', lockfile.to_yaml)
```
chris-rock, line 100: we should take the `path` into consideration

stevendanna: Currently `path` refers to the path to the vendored dependencies.  If we want we can make it refer to the path to the profile we want to lock and then make the path to the vendored dependencies be something you pass via a flag.

lib/inspec/cli.rb
chris-rock, line 99: maybe should call that profile.dependencies. The lockfile is only a written form of all the dependencies. The word lockfile is confusing on top of profile, because a profile should only know about dependencies but not lockfiles. I would attache the lockfile to DependencySet, instead of the call I would give it a path where the expected lock file should be located

lib/inspec/profile.rb
```diff
@@ -211,6 +212,44 @@ def locked_dependencies
       @locked_dependencies ||= load_dependencies
     end
 
+    def lockfile_exists?
```
chris-rock, line 215: I think we should hand the whole lockfile methods into the Lockfile class and just call it from there. Resolved lockfiles result in Dependencies. I think a profile should not know about lockfiles, only dependencies

stevendanna:
> I think the inspec.lock is not generated in the right path. It should be generated within the root profile directory

See my comment in-line re this.  Currently `vendor` assumes you are in a profile.

lib/inspec/cli.rb
chris-rock: alright, but how would that work if I run:

```
$ inspec exec profile1
$ inspec exec profile2
```

Wouldn't that break the user expectation?

lib/inspec/dependencies/lockfile.rb
```diff
@@ -0,0 +1,94 @@
...
+      lockfile_content = {
+        'lockfile_version' => CURRENT_LOCKFILE_VERSION,
+        'depends' => dep_set.to_array,
+      }
+      new(lockfile_content)
+    end
+
+    def self.from_file(path)
+      parsed_content = YAML.load(File.read(path))
+      version = parsed_content['lockfile_version']
```
chris-rock, line 20: cool to have that built in from the very beginning 👍

lib/inspec/dependencies/lockfile.rb
```diff
@@ -0,0 +1,94 @@
...
+        $stderr.puts <<EOF
+WARNING: This is a version 0 lockfile. Thank you for trying the
+experimental dependency management feature. Please be aware you may
+need to regenerate this lockfile in future versions as the feature is
+currently in development.
+EOF
+      elsif version > CURRENT_LOCKFILE_VERSION
+        fail <<EOF
+This lockfile claims to be version #{version} which is greater than
+the most recent lockfile version(#{CURRENT_LOCKFILE_VERSION}).
```
chris-rock, line 47: not sure I understand this. Is this what your are trying to achieve here?
`This lockfile claims to be version #{version} which is greater than the supported lockfile version(#{CURRENT_LOCKFILE_VERSION})`

lib/inspec/profile.rb
stevendanna: I agree we need to increase the separation between Lockfile and Profile. @ksubrama and I discussed that a bit and I think the work to change how profile loading works will drive some of that out.  However, in terms of the lockfile_exists? method specifically, I'm not sure it is true that we don't want profile knowing about lockfiles at all since the lockfile now lives in the profile directory.

chris-rock:
Thanks @stevendanna this works:

```
$ cd test/unit/mock/profiles/dependencies/inheritance
$ inspec vendor 
$ cat inspec.lock 
---
lockfile_version: 0
depends:
- name: profile_a
  resolved_source: file://../profile_a
  version_constraints: ">= 0"
  dependencies:
  - name: profile_c
    resolved_source: file://../profile_c
    version_constraints: ">= 0"
- name: profile_b
  resolved_source: file://../profile_b
  version_constraints: ">= 0"
- name: os-hardening
  resolved_source: https://github.com/dev-sec/tests-os-hardening/archive/master.tar.gz
  version_constraints: ">= 0"
  content_hash: d6c5299a7cdcce2040c157cc81cb2b21a85d3001493f8f3fb2aba175b9aa046c
```

lib/inspec/profile.rb
chris-rock: I am not fan of introducing side-effect to classes that should work independently. We should at least bundle that into the dependency implementation instead of the profile implementation

stevendanna: It really isn't clear to me how `lockfile_exists?()` is a side-effect?  Whether a lockfile is inside the profile is a property of the profile and depends on information (the path to the profile) that you don't necessarily want to expose to users of the class.

chris-rock:
discussed with @stevendanna, we are going to address my comments in later PRs

stevendanna:
@chris-rock And I had an in-person discussions where we agreed that:
- The high-level coordination of Inspec::Profile, Inspec::Lockfile, and Inspec::DependencySet isn't right yet.  We think that the right path here will get clearer as we work on scoping profile loading.
- The UX around the `vendor` command needs to be finalized. The `path` argument should probably be the path to the profile and then the lockfile should get placed at the same level as `inspec.yml`.  The vendor directory can move to an option.

chris-rock:
related to #888

PR #961, opened 2016-08-23
Replace Molinillo-based resolver

stevendanna:
The Molinillo library is a good library for systems that need a
constraint solver that will solve dependency problems requiring a single
version of each named dependency.

In our case, the eventual goal is to allow libraries to have conflicting
transitive dependencies at runtime. Isolation will be provided by
restricting all calls within a given profile to scope which can only see
that profile's dependencies.

To facilitate working on the isolation feature, I've replaced the
Molinillo-based resolver with a minimal resolver which will allow us to
load multiple versions of the same library.

Since we will likely want a good amount of logging around this feature
in the future, I've added a Inspec::Log singleton-style class, replacing
the previous Inpsec::Log which appeared unused in the code base.

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -34,6 +34,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'rspec-its', '~> 1.2'
   spec.add_dependency 'pry', '~> 0'
   spec.add_dependency 'hashie', '~> 3.4'
-  spec.add_dependency 'molinillo', '~> 0'
+  spec.add_dependency 'mixlib-log'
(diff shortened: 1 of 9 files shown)


vjeffrey:
:+1:

chris-rock:
I am not fan of adding two things at the same time. The logger should have been done in a separate PR. In addition, using the https://github.com/chef/mixlib-log result in additional output on STDOUT, which breaks json output in a computed format. Therefore the default output should be STDERR

chris-rock:
👍 on the dependency resolver

chris-rock:
related to #888

PR #965, opened 2016-08-23
Expand relative paths based on profile location

stevendanna:
Also: Log to STDERR by default

NB: This will result in absolute paths being rendered to lock files. We
think that is OK for now since we are going to build some UX around
path-based dependencies and lock files.  Namely, we are going to tell
people it is a bad idea.


chris-rock:
just had a quick discussion with @arlimus about the logging. To improve compatibility with CI systems we could do the following approach: We only redirect logging to STDERR if `format=json` is activated.

stevendanna:
@chris-rock Updated

PR #968, opened 2016-08-23
can check windows service startup mode now

Anirudh-Gupta:
Allows one to check windows services start up mode : https://msdn.microsoft.com/en-us/library/aa384896(v=vs.85).aspx
For example: 

``` ruby
describe service('AppMgmt') do
  it { should be_installed }
  it { should_not  be_running }
  its('startmode') { should be 'Manual'} 
end
```


ksubrama:
Thank you Anirudh!  I'll merge this in once travis goes green.

PR #969, opened 2016-08-23
Allow JSON 2.x

jkeiser:
As Chef and ChefDK move to Ruby 2.3 (which is happening as of this next release), we're moving away from Ruby 2.1 support. This means that the JSON version can finally be bumped. In the meantime it's best to allow both the old and new version (which allows us to upgrade when ready).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Berksfile.lock
 .librarian
 Puppetfile.lock
 .kitchen.local.yml
+binstubs
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ before_install:
 
 matrix:
   include:
-  - rvm: 1.9.3
-  - rvm: 2.0
-  - rvm: 2.2
+  - rvm: 2.2.5
+  - rvm: 2.3.1
+  - rvm: 2.3.1
     script: bundle exec rake lint test test:resources config=test/test.yaml
-  - rvm: 2.2
+  - rvm: 2.3.1
     script: N=5 bundle exec rake test:functional test:resources config=test/test-extra.yaml
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,8 @@ platform:
 
 environment:
   matrix:
-    - ruby_version: "200-x64"
-    - ruby_version: "21"
+    - ruby_version: "23-x64"
+    - ruby_version: "23"
 
 clone_folder: c:\projects\inspec
 clone_depth: 1
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'r-train', '~> 0.10.4'
   spec.add_dependency 'thor', '~> 0.19'
-  spec.add_dependency 'json', '~> 1.8'
+  spec.add_dependency 'json', '>= 1.8', '< 3.0'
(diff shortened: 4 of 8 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ Berksfile.lock
 .librarian
 Puppetfile.lock
 .kitchen.local.yml
+binstubs
```
chris-rock, line 12: why do we need this here?

.travis.yml
```diff
@@ -17,9 +17,9 @@ before_install:
 
 matrix:
   include:
-  - rvm: 1.9.3
-  - rvm: 2.0
-  - rvm: 2.2
+  - rvm: 2.2.5
```
chris-rock, line 20: At the moment, most workstations use `gem install inspec`, therefore we need to support more versions that chef. Therefore we should not change the old version behavior.

inspec.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'r-train', '~> 0.10.4'
   spec.add_dependency 'thor', '~> 0.19'
-  spec.add_dependency 'json', '~> 1.8'
+  spec.add_dependency 'json', '>= 1.8', '< 3.0'
```
chris-rock, line 29: 👍

appveyor.yml
```diff
@@ -4,8 +4,8 @@ platform:
 
 environment:
   matrix:
-    - ruby_version: "200-x64"
-    - ruby_version: "21"
+    - ruby_version: "23-x64"
```
chris-rock, line 7: as described above, I do not see the need to change it here

chris-rock:
@jkeiser Thanks for adding JSON 2 support 👍 Can we focus the PR on only this one improvement please?

jkeiser:
@chris-rock we can't focus on just json 2.0, because once you allow 2.0 you become 2.1-incompatible.

jkeiser:
The reason to change things to Ruby 2.3 is, the major delivery vehicle for inspec (chef and chef-dk) is on Ruby 2.3. It seems prudent to test against the Ruby you will primarily run on, is all. Happy to change it to 2.2, if you so desire, but I can't change it to 2.1 with the relaxing of the constraint.

.gitignore
jkeiser: It's a typical place where `bundle install` puts binstubs (the main alternative to `bin`, which we use here).

.travis.yml
jkeiser: If we can't abandon 2.1, then we can't do this change at all. Note that Chef and Chef DK have abandoned 2.1 (as of the next release) and in fact will not `gem install` on those versions anymore without a `Gemfile`.

inspec.gemspec
jkeiser: json 2.0 requires ruby version 2.2, and a gem install doesn't take this into account (sadly), so relaxing this constraint will break gem installs on Ruby 2.1 and below.

jkeiser:
@chris-rock in any case, the tests are failing on 2.3 in ways that make me wonder whether it even _works_ on 2.3. It would be prudent to test `inspec` on 2.3 before the next `chef` release, regardless of whether we make the leap to JSON 2.0.

jkeiser:
@chris-rock FYI, I created https://github.com/chef/inspec/pull/971 to add Ruby 2.3 to the test matrix. Let's revisit this once that's all worked out.

PR #970, opened 2016-08-23
Speed up windows package lookup

ksubrama:


ksubrama:
This is supposed to fix https://github.com/chef/inspec/issues/86 but I can't figure out what the unit tests are doing and what and _() function is....  I'll look into this tomorrow.

test/unit/resources/package_test.rb
```diff
@@ -53,8 +53,11 @@
 
   # windows
   it 'verify windows package parsing' do
+    STDERR.puts "FUCK EVERYTHING!"
```
stevendanna: We probably don't want to commit this :D

chris-rock: @ksubrama Its not acceptable to place offensive language in a PR.

chris-rock:
@ksubrama thanks for the new implementation. I added unit and integration tests for that implementation

test/integration/default/package_spec.rb
```diff
@@ -28,3 +28,31 @@
...
+# the following test will iterate over all packages retrieved via WMI and verifies that the
+# optimized package implementation is returning the same results
+if os.windows?
+  # compare optimized version with wmi results
+  packages = powershell("Get-WmiObject -Class Win32_Product |  % { $_.Name }").stdout.strip.split("\n")
+
+  packages.each { |pkg|
+    package_name = pkg.strip
+    # get wmi package information
+    cmd = powershell("Get-WmiObject -Class Win32_Product | Where-Object {$_.Name -eq '#{package_name}'} | Select-Object -Property Name,Version,Vendor,PackageCode,Caption,Description | ConvertTo-Json")
```
stevendanna, line 41: 👍 I like that we are using the old command here to ensure we match the same output we would have gotten with the old code.

stevendanna:
👍 This looks reasonable to me.

chris-rock:
fixes https://github.com/chef/inspec/issues/86

PR #971, opened 2016-08-24
Add Ruby 2.3 to the test matrix, make it the primary test for most suites

jkeiser:
Chef and ChefDK will soon ship with Ruby 2.3, so we should make sure the inspec suite works on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,42 +15,42 @@ before_install:
 matrix:
   include:
   - rvm: 1.9.3
-  - rvm: 2.0
   - rvm: 2.1.8
-  - rvm: 2.2
+  - rvm: 2.2.5
+  - rvm: 2.3.1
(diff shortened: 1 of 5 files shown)


jkeiser:
@chris-rock it looks to be failing on windows ruby 2.3: https://ci.appveyor.com/project/chef/inspec/build/1.0.1258/job/8iwny2t86wi4p7e4

chris-rock:
I can reproduce that behavior on my local Windows instance. The problem seems to be located here: https://github.com/chef/train/blob/master/lib/train.rb#L37-L47 

@ksubrama any ideas?

chris-rock:
Thanks @jkeiser Lets find out, why this is failing on Windows with ruby 2.3.1

ksubrama:
Fixed with https://github.com/chef/inspec/pull/980

Issue #958, opened 2016-08-23
Introduce scoping to the ProfileContext which has a view of all of its dependencies 


arlimus:
Part of https://github.com/chef/inspec/issues/888
- [x] create an up-front profile context for every profile that is pulled in (e.g. as dependencies)
- [x] attach profile contexts to the dependency tree
- [x] runner loads content in the context of profile inside the dep tree

stevendanna:
I believe #994 gets us the first level of scoping we wanted.  We still have potential work to do with namespacing to allow people to differentiate between two resources with the same name in the same scope.

PR #973, opened 2016-08-24
Initial control isolation support

stevendanna:
The goal of this change is to provide an isolated view of the available
profiles when the user calls the include_controls or require_controls
APIs.  Namely,
- A profile should only be able to reference profiles that are part of
  its transitive dependency tree. That is, if the dependency tree for a
  profile looks like the following:
  
  A
  |- B --> C
  |
  |- D --> E
  
  Then profile B should only be able to see profile C and fail if it
  tries to reference A, D, or E.
- The same profile should be include-able at different versions from
  different parts of the tree without conflict.  That is, if the
  dependency tree for a profile looks like the following:
  
  A
  |- B --> C@1.0
  |
  |- D --> C@2.0
  
  Then profile B should see the 1.0 version of C and profile D should
  see the 2.0 profile C with respect to the included controls.

To achieve these goals we:
- Ensure that we construct ProfileContext objects with respect to the
  correct dependencies in Inspec::DSL.
- Provide a method of accessing all transitively defined rules on a
  ProfileContext without pushing all of the rules onto the same global
  namespace.

This does not yet handle attributes or libraries.

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -282,34 +282,34 @@ module Inspec
       params
     end
 
+    #
+    # Returns a new runner for the current profile.
+    #
+    # @params [Symbol] The type of backend to use when constructing a
+    #                  new runner.
+    # @returns [Inspec::Runner]
+    #
+    def runner_for_profile(backend = :mock)
+      opts = @options.dup
+      opts[:ignore_supports] = true
+      r = Runner.new(id: @profile_id,
+                     backend: backend,
+                     test_collector: opts.delete(:test_collector))
+      r.add_profile(self, opts)
+      r
+    end
+
     def load_checks_params(params)
       params[:controls] = controls = {}
       params[:groups] = groups = {}
       prefix = @source_reader.target.prefix || ''
-
-      if @runner_context.nil?
-        # we're checking a profile, we don't care if it runs on the host machine
-        opts = @options.dup
-        opts[:ignore_supports] = true
-        runner = Runner.new(
-          id: @profile_id,
-          backend: :mock,
-          test_collector: opts.delete(:test_collector),
-        )
-        runner.add_profile(self, opts)
-        runner.rules.each do |rule|
-          f = load_rule_filepath(prefix, rule)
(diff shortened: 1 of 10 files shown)


stevendanna:
Partially fixes #958 by ensuring that the profile context that we load during include_control and require control uses the dependencies for the given profile, not for the parent profiles.

lib/inspec/profile_context.rb
```diff
@@ -35,6 +38,16 @@ def reload_dsl
       @profile_context = ctx.new(@backend, @conf, @dependencies, @require_loader)
     end
 
+    def all_rules
+      ret = @rules.values
+      ret += @subcontexts.map(&:all_rules).flatten
+      ret
```
chris-rock, line 44: ah nice :-)

lib/inspec/profile_context.rb
```diff
@@ -10,18 +10,21 @@
 
 module Inspec
   class ProfileContext # rubocop:disable Metrics/ClassLength
-    attr_reader :rules
-    attr_reader :attributes
+    def self.for_profile(profile, backend)
```
chris-rock, line 13: +1

lib/inspec/profile.rb
```diff
@@ -278,34 +282,34 @@ def load_params
...
-        params[:attributes] = runner.attributes
-      else
-        # load from context
-        @runner_context.rules.values.each do |rule|
-          f = load_rule_filepath(prefix, rule)
-          load_rule(rule, f, controls, groups)
-        end
-        params[:attributes] = @runner_context.attributes
+      # Load from the runner_context if it exists
+      runner = @runner_context || runner_for_profile
```
chris-rock, line 307: 👍 great improvement

chris-rock:
This is a great improvement @stevendanna Thanks 💯

chris-rock:
There is a known issue https://github.com/chef-cookbooks/compat_resource, therefore we ignore the errors in our kitchen tests

chris-rock:
related to #888

PR #975, opened 2016-08-24
Fix `bundle install` on Ruby 2.1.9

jkeiser:
`rack` is another gem whose latest version does not work on Ruby 2.1.9.

I also fixed it so `<= 2.1.0` became `< 2.2`, since `2.1.9` was not being covered.

Finally, I added 2.1.9 to the Travis matrix the current `chef` and `chef-dk` stable releases still run 2.1.9, and gave specific versions to all the Rubies we run, since on Travis 2.1 means 2.1.0, which has known bugs and which no one runs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,9 @@ if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('1.9.3')
   gem 'net-ssh', '~> 2.9'
 end
 
-if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('2.1.0')
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.0')
   gem 'json', '~> 1.8'
+  gem 'rack', '< 2.0'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -8,8 +8,9 @@ if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('1.9.3')
   gem 'net-ssh', '~> 2.9'
 end
 
-if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('2.1.0')
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.0')
```
tas50, line 11: needs to be 2.2.2

chris-rock:
@jkeiser awesome fix 💯

chris-rock:
I am going to merge it although not all kitchen test pass. This is a know issue in https://github.com/chef-cookbooks/compat_resource cookbook.

PR #976, opened 2016-08-24
Fix `rake` to work again

jkeiser:
It was searching for `rake generate` because that is the default task, and does not exist.


chris-rock:
Thanks @jkeiser

PR #977, opened 2016-08-25
first pass at collecting command output for demo

vjeffrey:
heya peoples can you take a look and let me know what you think of this??  we would want the files to actually be saved somewhere in the www/ folder that the demo branch will introduce and then pull them in with the html....so i would need to make sure this overwrites too -- but i wanted to make sure something like this would be ok??. thank ya :)


tasks/get_command_output.rb
```diff
@@ -0,0 +1,13 @@
+# encoding: utf-8
+require 'open3'
```
chris-rock: Lets use https://github.com/chef/mixlib-shellout or 

```
backend = Train.create('local', target_config)
# start or reuse a connection
conn = backend.connection
# loop around commands
cmd = conn.run_command('inspec version')
cmd.stdout
conn.close
```

under the hood, train is using mixlib-shellout

tasks/get_command_output.rb
```diff
@@ -0,0 +1,13 @@
+# encoding: utf-8
```
chris-rock: should we name the command like `inspec_simulator`?

tasks/get_command_output.rb
```diff
@@ -0,0 +1,13 @@
...
+require 'open3'
+
+commands = { 'inspec_exec' => 'inspec exec examples/profile/controls/example.rb', 'inspec_version' => 'inspec version' }
+
+commands.each do |keyname, command|
+  stdout, _stdeerr, _status = Open3.capture3(command)
+
+  # filename should be key.txt...and we should be saving it somewhere in www/
+  out_file = File.new("#{keyname}.txt", 'w')
```
chris-rock: use a base path as parameter, but this looks great

Rakefile
```diff
@@ -17,6 +17,13 @@ end
 desc 'Run robocop linter'
 task lint: [:rubocop]
 
+# update command output for demo
+desc 'Running commands and saving TBD'
```
chris-rock: +1

chris-rock:
@vj this looks great, lets just switch the libs, running it via train would allow us build a predefined container that is including the latest inspec later. This would make results consistent across operating systems and CI pipelines.

chris-rock:
Once we have the basic commands covered, LGTM

vjeffrey:
@chris-rock updated to use train

chris-rock:
Thanks @vjeffrey for this great improvement! I'll create a ticket for another iteration :-)

tasks/command_simulator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+commands = { 'inspec_exec' => 'inspec exec examples/profile/controls/example.rb', 'inspec_version' => 'inspec version' }
+
+commands.each do |keyname, command|
+  backend = Train.create('local')
+  conn = backend.connection
+  # loop around commands
+  cmd = conn.run_command(command)
+  cmd.stdout
+  conn.close
```
chris-rock, line 12: move 

```
backend = Train.create('local')
conn = backend.connection
```

and `conn.close` out of the loop

PR #979, opened 2016-08-25
move train connection out of loop for command_simulator

vjeffrey:
follow up to https://github.com/chef/inspec/pull/977

@chris-rock


chris-rock:
Great improvement @vjeffrey

PR #982, opened 2016-08-25
Add windows user SID as 'UID' in user resource. Fix #960

### LINKED ISSUE #960: Support sid for user resource
### Description

The current user resource is not returning the sid for users:

```
describe user("Administrator") do
  it { should exist }
  its('uid') { should_not eq nil }
end
```

0.31.0
### Possible Solutions

We should return the sid as part of the user object: https://github.com/chef/inspec/blob/master/lib/resources/user.rb#L457-L463

This PR is depending on the improvement #86

ksubrama:


ksubrama:
```
PS C:\vagrant\code\inspec> bundle exec inspec shell -c "describe user('Administrator') do its('uid') { should_not eq nil
 } end"
[DEPRECATION] `script(script)` is deprecated.  Please use `powershell(script)` instead.
  ✔  User Administrator uid should not eq nil

Summary: 1 successful, 0 failures, 0 skipped
```

PR #987, opened 2016-08-26
Minor refactor and explanatory comments

stevendanna:
This is a minor refactor that I did while studying our loading code in
preparation for some deeper changes to how content loading works. The
overall goal of the refactor is to remove a few places where we were
passing a generic options hash and then only accessing a single item.

The comment hopefully clarifies to new developers in the code base how
content loading works at a high level.

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -15,9 +15,8 @@ module Inspec
   class Profile # rubocop:disable Metrics/ClassLength
     extend Forwardable
 
-    def self.resolve_target(target, opts)
+    def self.resolve_target(target)
       # Fetchers retrieve file contents
-      opts[:target] = target
       fetcher = Inspec::Fetcher.resolve(target)
       if fetcher.nil?
         fail("Could not fetch inspec profile in #{target.inspect}.")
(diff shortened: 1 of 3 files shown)


lib/inspec/profile.rb
```diff
@@ -15,9 +15,8 @@ module Inspec
   class Profile # rubocop:disable Metrics/ClassLength
     extend Forwardable
 
-    def self.resolve_target(target, opts)
+    def self.resolve_target(target)
```
chris-rock, line 18: 👍

lib/inspec/runner.rb
```diff
@@ -66,83 +80,114 @@ def load_attributes(options)
...
+    #
+    # Once the we've loaded all of the tests files in the profile, we
+    # query the profile for the full list of rules. Those rules are
+    # registered with the @test_collector which is ultimately
+    # responsible for actually running the tests.
+    #
+    # TODO: Deduplicate/clarify the loading code that exists in here,
+    # the ProfileContext, the Profile, and Inspec::DSL
+    #
+    # TODO: Libraries of dependent profiles should be loaded even when
```
chris-rock, line 105: +1 right, that is essential for resource packs aka profiles without any test, but only libraries

PR #988, opened 2016-08-26
Allow functional tests to pass on OSX

stevendanna:
A few minor issues were causing 3 functional test failures on OS X.
These were not program errors but where rather the result of the
profiles under test assuming a linux environment.

Since many of the developers who will work on this project in the future
will be running OS X, let's ensure they can run the functional tests
easily.


chris-rock:
Thanks for fixing that @stevendanna

PR #991, opened 2016-08-29
cmp not treating 0 as integer only as string

jeremymv2:
This change allows both tests to pass, which I believe is the intent of the `cmp` matcher.

```
describe 0 do
  it { should cmp 0 }
  it { should cmp '0' }
end
```

Without the change from this PR, it fails one of the tests:

```
~/Devel/ChefProject/inspec (cmp_treat_0_as_integer $%=)$ inspec exec ../tmp/cmp_test.rb --format=documentation

0
  should cmp 0
  should cmp "0" (FAILED - 1)

Failures:

  1) 0 should cmp "0"

       expected: "0"
            got: 0

       (compared using `cmp` matcher)
     # ../tmp/sshd_test.rb:7:in `block (2 levels) in load'

Finished in 0.01047 seconds (files took 0.79418 seconds to load)
2 examples, 1 failure

Failed examples:

rspec  # 0 should cmp "0"

~/Devel/ChefProject/inspec (cmp_treat_0_as_integer $%=)$
```

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -232,7 +232,7 @@ end
 RSpec::Matchers.define :cmp do |first_expected|
 
   def integer?(value)
-    !(value =~ /\A[1-9]\d*\Z/).nil?
+    !(value =~ /\A[0-9]\d*\Z/).nil?
   end
 
   def float?(value)
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -232,7 +232,7 @@ end
 RSpec::Matchers.define :cmp do |first_expected|
 
   def integer?(value)
-    !(value =~ /\A[0-9]\d*\Z/).nil?
+    !(value =~ /\A0+\Z|\A[1-9]\d*\Z/).nil?
   end
 
   def float?(value)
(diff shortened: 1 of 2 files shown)


lib/matchers/matchers.rb
```diff
@@ -232,7 +232,7 @@
 RSpec::Matchers.define :cmp do |first_expected|
 
   def integer?(value)
-    !(value =~ /\A[1-9]\d*\Z/).nil?
+    !(value =~ /\A[0-9]\d*\Z/).nil?
```
chris-rock: Great catch and really good test-case. We left 0 out, to cover the detection for octal numbers. Therefore 0 is only valid if the length of the string is 1

chris-rock: Great catch and really good test-case. We left 0 out, to cover the detection for octal numbers. Therefore 0 is only valid if the length of the string is 1

jeremymv2:
@chris-rock assuming that last commit passes tests, could you lend 👀 on this please to ensure it's sane for what how you expect 'cmp' to work?

chris-rock:
This is a great fix. Thanks for finding this edge case @jeremymv2

PR #994, opened 2016-08-30
Initial attempt at isolating resources between dependencies

stevendanna:
Previously, all resources were loaded into a single resource registry.
Now, each profile context has a resource registry, when a profile's
library is loaded into the profile context, we update the
profile-context-specific resource registry.  This local registry is
then used to populate the execution context that the rules are
evaluated in.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -70,7 +70,7 @@ module Inspec
       o[:logger].level = get_log_level(o.log_level)
 
       runner = Inspec::Runner.new(o)
-      targets.each { |target| runner.add_target(target, opts) }
+      targets.each { |target| runner.add_target(target) }
       exit runner.run
     rescue RuntimeError, Train::UserError => e
       $stderr.puts e.message
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -60,7 +60,7 @@ module Inspec
         end
 
         def to_s
-          "Profile Context Run #{profile_name}"
+          "Control Evaluation Context (#{profile_name})"
         end
 
         define_method :profile_name do
@@ -73,6 +73,12 @@ module Inspec
           register_control(rule_class.new(id, profile_id, opts, &block))
         end
 
+        #
+        # Describe allows users to write rspec-like bare describe
+        # blocks without declaring an inclosing control. Here, we
+        # generate a control for them automatically and then execute
+        # the describe block in the context of that control.
+        #
         define_method :describe do |*args, &block|
           loc = block_location(block, caller[0])
           id = "(generated from #{loc} #{SecureRandom.hex})"
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -1,0 +1,139 @@
+# encoding: utf-8
+# author: Dominik Richter
+# author: Christoph Hartmann
+require 'inspec/dsl'
+require 'inspec/dsl_shared'
+
+module Inspec
+  #
+  # ControlEvalContext constructs an anonymous class that control
+  # files will be instance_exec'd against.
+  #
+  # The anonymous class includes the given passed resource_dsl as well
+  # as the basic DSL of the control files (describe, control, title,
+  # etc).
+  #
+  class ControlEvalContext
+    # Create the context for controls. This includes all components of the DSL,
+    # including matchers and resources.
+    #
+    # @param [ResourcesDSL] resources_dsl which has all resources to attach
+    # @return [RuleContext] the inner context of rules
+    def self.rule_context(resources_dsl)
+      require 'rspec/core/dsl'
+      Class.new(Inspec::Rule) do
+        include RSpec::Core::DSL
+        include resources_dsl
+      end
+    end
+
+    # Creates the heart of the control eval context:
+    #
+    # An instantiated object which has all resources registered to it
+    # and exposes them to the a test file.
+    #
+    # @param profile_context [Inspec::ProfileContext]
+    # @param outer_dsl [OuterDSLClass]
+    # @return [ProfileContextClass]
+    #
+    # rubocop:disable Lint/NestedMethodDefinition
+    def self.create(profile_context, resources_dsl) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      rule_class = rule_context(resources_dsl)
+      profile_context_owner = profile_context
+      profile_id = profile_context.profile_id
+
+      Class.new do
+        include Inspec::DSL
+        include Inspec::DSL::RequireOverride
+        include resources_dsl
+
+        def initialize(backend, conf, dependencies, require_loader)
+          @backend = backend
+          @conf = conf
+          @dependencies = dependencies
+          @require_loader = require_loader
+          @skip_profile = false
+        end
+
+        define_method :title do |arg|
+          profile_context_owner.set_header(:title, arg)
+        end
+
+        def to_s
+          "Profile Context Run #{profile_name}"
+        end
+
+        define_method :profile_name do
+          profile_id
+        end
+
+        define_method :control do |*args, &block|
+          id = args[0]
+          opts = args[1] || {}
+          register_control(rule_class.new(id, profile_id, opts, &block))
+        end
+
+        define_method :describe do |*args, &block|
+          loc = block_location(block, caller[0])
+          id = "(generated from #{loc} #{SecureRandom.hex})"
+
+          res = nil
+          rule = rule_class.new(id, profile_id, {}) do
+            res = describe(*args, &block)
+          end
+          register_control(rule, &block)
+
+          res
+        end
+
+        define_method :add_resources do |context|
+          self.class.class_eval do
+            include context.to_resources_dsl
+          end
+
+          rule_class.class_eval do
+            include context.to_resources_dsl
+          end
+        end
+
+        define_method :add_subcontext do |context|
+          profile_context_owner.add_subcontext(context)
+        end
+
+        define_method :register_control do |control, &block|
+          ::Inspec::Rule.set_skip_rule(control, true) if @skip_profile
+
+          profile_context_owner.register_rule(control, &block) unless control.nil?
+        end
+
+        # method for attributes; import attribute handling
+        define_method :attribute do |name, options|
+          profile_context_owner.register_attribute(name, options)
+        end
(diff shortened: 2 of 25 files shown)


test/unit/mock/profiles/dependencies/profile_c/libraries/gordon_config.rb
```diff
@@ -0,0 +1,54 @@
...
+    describe gordon_config do
+      its('version') { should eq('1.0') }
+      its('file_size') { should > 1 }
+    end
+  "
+
+  # Load the configuration file on initialization
+  def initialize
+    @params = {}
+    @path = '/tmp/gordon/config.yaml'
```
chris-rock: we may simplify the resource to just return a version number. Otherwise we are relying on external configuration. eg.

```
def version
   1
end
```

test/unit/mock/profiles/dependencies/profile_c/libraries/gordon_config.rb
```diff
@@ -0,0 +1,54 @@
+require 'yaml'
+puts "Loading gordon_config from c"
+
+# Custom resource based on the InSpec resource DSL
+class GordonConfig < Inspec.resource(1)
+  name 'gordon_config'
+
+  desc "
```
chris-rock: we do not need desc and example in this test

test/unit/mock/profiles/dependencies/profile_a/controls/example.rb
```diff
@@ -20,3 +20,9 @@
     it { should be_directory }
   end
 end
+
+control 'profilea-2' do
+  describe gordon_config do
+    its('version') { should eq('1.0') }
```
chris-rock, line 26: lets test for `its('version') { should eq 1 }`

test/unit/mock/profiles/dependencies/profile_b/controls/example.rb
```diff
@@ -14,3 +14,9 @@
     it { should be_directory }
   end
 end
+
+control 'profileb-2' do
+  describe gordon_config do
+    its('version') { should eq('1.0') }
```
chris-rock: lets test for `its('version') { should eq 2 }`

test/unit/mock/profiles/dependencies/inheritance/inspec.yml
```diff
@@ -10,5 +10,3 @@ depends:
     path: ../profile_a
   - name: profile_b
     path: ../profile_b
-  - name: ssh-hardening
```
chris-rock, line 13: any reason why we are removing this?

lib/inspec/profile_context.rb
```diff
@@ -4,17 +4,20 @@
...
   class ProfileContext # rubocop:disable Metrics/ClassLength
     def self.for_profile(profile, backend)
-      new(profile.name, backend, { 'profile' => profile })
+      c = new(profile.name, backend, { 'profile' => profile })
+      c.load_profile_content(profile)
     end
 
-    attr_reader :attributes, :rules, :profile_id
+    attr_reader :attributes, :rules, :profile_id, :resource_registry
+    attr_reader :profile_execution_context
```
chris-rock: may it make sense to just use `execution_context`? since we are in the profile context already

lib/inspec/resource.rb
```diff
@@ -3,17 +3,36 @@
 # license: All rights reserved
 # author: Dominik Richter
 # author: Christoph Hartmann
-
+require 'thread'
 require 'inspec/plugins'
 
 module Inspec
   class Resource
+    REGISTRY_LOCK = Mutex.new
```
chris-rock: Why do we need threading and a mutex?

lib/inspec/profile_context.rb
stevendanna: Based on our discussion I think we'll end up with two context, perhaps:
- `control_eval_context`
- `resource_eval_context`

chris-rock: This makes sense to me

lib/inspec/resource.rb
stevendanna: Summary of discussion:

We can probably replace this by introducing a profile-specific class that all library files are evaluated against.

test/unit/mock/profiles/dependencies/profile_a/controls/example.rb
stevendanna: Simplified these as suggested.

chris-rock: This looks great now

test/unit/dsl/control_test.rb
```diff
@@ -10,7 +10,7 @@ def load(content)
       'inspec.yml' => "name: mock",
       'controls/mock.rb' => "control '1' do\n#{content}\nend\n",
     }
-    opts = { test_collector: Inspec::RunnerMock.new }
+    opts = { test_collector: Inspec::RunnerMock.new, backend: Inspec::Backend.create({}) }
```
chris-rock: wouldn't that require `Inspec::Backend.create({ backend: 'mock' })`?

lib/inspec/runner.rb
```diff
@@ -102,61 +120,15 @@ def load_attributes(options)
...
-    # considered by this call path are:
-    #
-    # - `:ignore_supports`: A boolean that controls whether to ignore
-    #   the profile's metadata regarding supported Inspec versions or
-    #   platforms.
-    #
-    # - `:controls`: An array of controls to run from this
-    #    profile. Any returned rules that are not part of these
-    #    controls will be filtered before being passed to @test_collector.
+    # @params _opts [Hash] Unused, but still here to avoid breaking kitchen-inspec
```
chris-rock, line 124: We should add a ticket in kitchen-inspec as a follow-up to fix that

test/unit/mock/profiles/dependencies/inheritance/inspec.yml
stevendanna: @chris-rock Mostly because the fetching wasn't the main thing we were working so the ssh output was really just cluttering up what we wanted to see.

vjeffrey:
:+1:

lib/inspec/base_cli.rb
```diff
@@ -70,7 +70,7 @@ def run_tests(targets, opts)
       o[:logger].level = get_log_level(o.log_level)
 
       runner = Inspec::Runner.new(o)
-      targets.each { |target| runner.add_target(target, opts) }
+      targets.each { |target| runner.add_target(target) }
```
chris-rock, line 73: 👍

lib/inspec/control_eval_context.rb
```diff
@@ -0,0 +1,139 @@
...
+          @require_loader = require_loader
+          @skip_profile = false
+        end
+
+        define_method :title do |arg|
+          profile_context_owner.set_header(:title, arg)
+        end
+
+        def to_s
+          "Profile Context Run #{profile_name}"
```
chris-rock: Is this really representing a full profile context? Its only representing one section aka file in a profile

lib/inspec/control_eval_context.rb
```diff
@@ -0,0 +1,139 @@
...
+          profile_id
+        end
+
+        define_method :control do |*args, &block|
+          id = args[0]
+          opts = args[1] || {}
+          register_control(rule_class.new(id, profile_id, opts, &block))
+        end
+
+        define_method :describe do |*args, &block|
```
chris-rock, line 82: do you think we should add a comment that this is simulating the rspec describe and implicitly generate a control?

lib/inspec/control_eval_context.rb
```diff
@@ -0,0 +1,139 @@
...
+          res = nil
+          rule = rule_class.new(id, profile_id, {}) do
+            res = describe(*args, &block)
+          end
+          register_control(rule, &block)
+
+          res
+        end
+
+        define_method :add_resources do |context|
```
chris-rock, line 95: the methods `add_resources`, `add_subcontext` and `register_control` should not be available to our DSL right?

lib/inspec/dependencies/dependency_set.rb
```diff
@@ -18,14 +18,22 @@ class DependencySet
     # @param cwd [String] Current working directory for relative path includes
     # @param vendor_path [String] Path to the vendor directory
     #
-    def self.from_lockfile(lockfile, cwd, vendor_path)
+    def self.from_lockfile(lockfile, cwd, vendor_path, backend)
```
chris-rock, line 21: param is not defined in comments

lib/inspec/dependencies/dependency_set.rb
```diff
@@ -18,14 +18,22 @@ class DependencySet
...
       vendor_index = VendorIndex.new(vendor_path)
       dep_tree = lockfile.deps.map do |dep|
-        Inspec::Requirement.from_lock_entry(dep, cwd, vendor_index)
+        Inspec::Requirement.from_lock_entry(dep, cwd, vendor_index, backend)
```
chris-rock, line 24: Somehow it feels strange that the dependency resolution needs a backend? Why is that required?

lib/inspec/dsl_shared.rb
```diff
@@ -0,0 +1,25 @@
+# encoding: utf-8
+module Inspec
+  #
+  # Contains methods we would like in multiple DSL
+  #
+  module DSL
+    module RequireOverride
```
chris-rock, line 7: nice separation

lib/inspec/control_eval_context.rb
stevendanna: We could hide these from the API directly via a layer of indirection.  If we want them to be completely hidden, I think we would need the Inspec::DSL code to change a bit more substantially.  Happy to do either.

lib/inspec/dependencies/dependency_set.rb
stevendanna: Yeah, I agree this is gross.  The reason we started threading this through everything is that we need the backend anywhere we might instantiate a ProfileContext.  Since we instantiate ProfileContext's as we are building Inpec::Profile's, we now need it all over the place.

chris-rock: As discussed, lets improve that in our next iterations

chris-rock:
Great improvements @stevendanna and @vjeffrey

chris-rock:
#958

PR #999, opened 2016-08-30
Promote cmp usage as it provides results with octal mode

alexpop:
The default rspec `eq` matcher outputs the expected mode as decimal:

```
its('mode') { should eq 0644 }

=>

 File /etc/ssh/ssh_config
     ✔  should be mode 420
```

where the inspec `cmp` matcher is outputting it as octal:

```
its('mode') { should cmp '0644' }

=>

  File /etc/ssh/ssh_config
     ✔  mode should cmp "0644"
```


chris-rock:
Thanks @alexpop for this promotion :-)

PR #1003, opened 2016-08-31
Allow service resource to accept Windows service name with spaces

martinheg:
Currently if you want to check a Windows resource using the `service` resource you must use the _non-friendly_ short service name, even though using the _friendlier_ service display name will work as long as it doesn't include any spaces.

For example, this currently works:

```
describe service('AeLookupSvc') do     
  it { should be_installed }
end
```

while this simple change also allows for this:

```
describe service('Application Experience') do # Where 'Application Experience' is the display name for the service 'AeLookupSvc'
  it { should be_installed }
end
```

making it easier to use this resource for Windows services (where the shortname and displayname can often be quite different).

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -575,7 +575,7 @@ module Inspec::Resources
     # - 6: Pause Pending
     # - 7: Paused
     def info(service_name)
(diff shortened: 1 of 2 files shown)


chris-rock:
@martinheg this is a great addition. Could you change https://github.com/chef/inspec/blob/master/test/helper.rb#L162 to

```
"New-Object -Type PSObject | Add-Member -MemberType NoteProperty -Name Service -Value (Get-Service -Name 'dhcp'| Select-Object -Property Name, DisplayName, Status) -PassThru | Add-Member -MemberType NoteProperty -Name WMI -Value (Get-WmiObject -Class Win32_Service | Where-Object {$_.Name -eq 'dhcp' -or $_.DisplayName -eq 'dhcp'} | Select-Object -Property StartMode) -PassThru | ConvertTo-Json" => cmd.call('get-service-dhcp'),
```

?
This would allow our tests to pass the unit tests

martinheg:
@chris-rock Thanks. Tests passing now - sorry for missing that first time round.

chris-rock:
Thanks @martinheg !

PR #1004, opened 2016-09-01
vj/fix cli report

vjeffrey:
screen shot 2016-08-31 at 11 00 23 pm
fixes https://github.com/chef/inspec/issues/984
and one part of 898


vjeffrey:
k, so the one test I was going to add to this was to test the "print exception msg for tests in anonymous describe block" since the test I had to fix covered the status icon commit and there are tests that check that I didn't over-delete the use of 'no tests executed'.....but adding that line to the describe block is messing with almost all the tests....so i'm trying to find a way around it but not sure how much it's worth it for that test?

chris-rock:
Great improvement @vjeffrey

PR #1010, opened 2016-09-02
Add simple GitHub fetcher

stevendanna:
This adds a fetcher for sources that are on github.com. It resolves
branch and tag references using the GitHub API and constructs an https
url to a tar archive that GitHub provides.  This allows us to support
GitHub-based sources without having the git executable on disk.


chris-rock:
I'd like to use oktokit. This PR should also cleanup the existing github fetcher https://github.com/chef/inspec/blob/master/lib/fetchers/url.rb#L29-L58
Once this fetcher understands plain github http urls, we should deny github.com urls in our URL fetcher.

test/unit/fetchers/github_test.rb
```diff
@@ -0,0 +1,55 @@
...
+
+require 'helper'
+require 'uri'
+
+describe Fetchers::Github do
+  let(:fetcher) { Fetchers::Github }
+  let(:ref_for_test) { "0ca497e333a3222e4b517bc3ebe43eba01986c4c" }
+  let(:tag_ref_for_test) { "202d4e0f97b5dc2b88df13d686eae36320a5a53c" }
+  let(:uri_without_spec) { URI.parse("github://chef/inspec") }
+  let(:uri_with_branch) { URI.parse("github://chef/inspec/branch/master") }
```
chris-rock, line 12: I would love to support the existing github fetcher behavior as well, where I am able to just copy the url from the browser. Just for user convenience.

inspec.gemspec
```diff
@@ -36,4 +36,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'hashie', '~> 3.4'
   spec.add_dependency 'mixlib-log'
   spec.add_dependency 'sslshake', '~> 1'
+  spec.add_dependency 'octokit'
```
chris-rock, line 39: nice addition

stevendanna:
I'm trying to figure out if there is a way to make the VCR cassette a bit smaller.

PR #1014, opened 2016-09-02
Fix apache conf

### LINKED ISSUE #1013: apache_conf resource seems to be using incorrect paths to amalgamate apache config (only Centos/RHEL?)
### Description

The Compliance tests I'm writing make heavy use of the `apache_conf` resource to validate `httpd` configuration. As I understand it, the resource builds a hash accessible via `apache_conf.params` by gathering all `.conf` and `.load` files in the files in apache/httpd config directory and combining them. 

The problem is that the resource seems to be using incorrect paths and is therefore failing to find anything but my standard `/etc/httpd/conf/httpd.conf` file. 

Here is an example of a simple test:

``` ruby
control 'WA000-WWA022' do
  impact 0.2 # review
  title 'The KeepAlive directive must be enabled.' 

  describe apache_conf do 
    its('KeepAlive') { should cmp 'On' }
  end
end
```

When I run `$ chef exec kitchen verify`, I can see this output in stdout:

``` bash
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.load -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/ports.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/conf-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/sites-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.load -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/ports.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/conf-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/sites-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.load -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/ports.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/conf-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/sites-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.load -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/mods-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/ports.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/conf-enabled/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/etc/httpd/sites-enabled/*.conf -maxdepth 1 -type f`
```

As you can see, it's failing to find any files, and the find path looks odd: `/etc/httpd/conf/etc/httpd/sites-enabled`. The standard path to `sites-enabled` is `/etc/httpd/sites-enabled`, same with `mods-enabled` etc.

Perplexed, I generated a new dummy cookbook to start from scratch and use Centos 7.2 for good measure. All I changed was installing Apache via `package 'httpd'` in the default recipe, which will grab version 2.4 for Centos 7. Added a similar test using `apache_conf` and got the following output:

```
find_files(): exit 1 from `find /etc/httpd/conf/conf.modules.d/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/conf.d/*.conf -maxdepth 1 -type f`
```

Something seems off, as my default `conf.d` and `conf.modules.d` are in `/etc/httpd`, not `/etc/httpd/conf`.

Still curious, I added Ubuntu 14.04 to the platforms in my dummy cookbook's kitchen config, installed `apache2` and didn't get any failure outputs when running the test.

Tested on:
Windows 7 Enterprise, SP1
ChefDK 0.17.17
kitchen 1.11.1
apache 2.2 installed with community/supermarket apache2 cookbook 
Virtual Box + Centos 6.7 for running tests
### Replication Case
1. `$ chef generate cookbook test`
2.  Add `package 'httpd'` to default recipe
3.  Use `centos-6.7` and/or `centos-7.2` as platforms in `kitchen.yml`
4. Add a test to `test/recipes/default_test.rb` that uses `apache_conf`
5. `$ kitchen verify`
### Stacktrace

See stdout content pasted above.

jeremymv2:
`apache_conf.conf_dir` was resolving incorrectly when OS was NOT Ubuntu || Debian.

It would end up resulting in find_file() exit 1 errors like below:

```
[vagrant@default-centos-67 ~]$ /opt/chef/embedded/bin/inspec shell

inspec> apache_conf
find_files(): exit 1 from `find /etc/httpd/conf/conf.d/*.conf -maxdepth 1 -type f`
=> Apache Config /etc/httpd/conf/httpd.conf
inspec>
```

It should be using a `conf_dir` of `/etc/httpd` so that find_files will search `/etc/httpd/conf.d/*.conf` instead of `/etc/httpd/conf/conf.d/*.conf`.  

This PR corrects the problem and resolves #1013

### DIFF
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -21,7 +21,7 @@ module Inspec::Resources
 
     def initialize(conf_path = nil)
       @conf_path = conf_path || inspec.apache.conf_path
-      @conf_dir = inspec.os.debian? ? File.dirname(@conf_path) : inspec.apache.conf_dir
+      @conf_dir = conf_path ? File.dirname(@conf_path) : inspec.apache.conf_dir
       @files_contents = {}
       @content = nil
       @params = nil
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -21,7 +21,12 @@ module Inspec::Resources
 
     def initialize(conf_path = nil)
       @conf_path = conf_path || inspec.apache.conf_path
-      @conf_dir = File.dirname(@conf_path)
+      case inspec.os[:family]
+      when 'ubuntu', 'debian'
+        @conf_dir = File.dirname(@conf_path)
+      else
+        @conf_dir = inspec.apache.conf_dir
+      end
       @files_contents = {}
       @content = nil
       @params = nil
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -21,12 +21,7 @@ module Inspec::Resources
 
     def initialize(conf_path = nil)
       @conf_path = conf_path || inspec.apache.conf_path
-      case inspec.os[:family]
-      when 'ubuntu', 'debian'
-        @conf_dir = File.dirname(@conf_path)
-      else
-        @conf_dir = inspec.apache.conf_dir
-      end
+      @conf_dir = inspec.os.debian? ? File.dirname(@conf_path) : inspec.apache.conf_dir
       @files_contents = {}
       @content = nil
       @params = nil
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -118,6 +118,8 @@ class MockLoader
       'rootwrap.conf' => mockfile.call('rootwrap.conf'),
       '/etc/apache2/apache2.conf' => mockfile.call('apache2.conf'),
       '/etc/apache2/ports.conf' => mockfile.call('ports.conf'),
+      '/etc/httpd/conf/httpd.conf' => mockfile.call('httpd.conf'),
+      '/etc/httpd/conf.d/ssl.conf' => mockfile.call('ssl.conf'),
       '/etc/apache2/conf-enabled/serve-cgi-bin.conf' => mockfile.call('serve-cgi-bin.conf'),
       '/etc/xinetd.conf' => mockfile.call('xinetd.conf'),
       '/etc/xinetd.d' => mockfile.call('xinetd.d'),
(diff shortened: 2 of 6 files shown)


jeremymv2:
@davidcpell this should fix your issue

lib/resources/apache_conf.rb
```diff
@@ -21,7 +21,7 @@ class ApacheConf < Inspec.resource(1)
 
     def initialize(conf_path = nil)
       @conf_path = conf_path || inspec.apache.conf_path
-      @conf_dir = File.dirname(@conf_path)
+      @conf_dir = inspec.os.debian? ? File.dirname(@conf_path) : inspec.apache.conf_dir
```
chris-rock: @jeremymv2 I am thinking about moving that logic to `inspec.apache`. So we would use `File.dirname(@conf_path)` only if `conf_path` is defined. Otherwise we always read the `conf_dir` from the apache resource. What do you think?

chris-rock:
This is a great improvement of our test coverage @jeremymv2

chris-rock:
Thanks a lot @jeremymv2 💯

PR #1015, opened 2016-09-02
make demo better

vjeffrey:
pr for demo improvements.


www/tutorial/package.json
```diff
@@ -2,12 +2,12 @@
...
   "version": "1.0.0",
   "scripts": {
-    "start": "tsc && concurrently \"npm run tsc:w\" \"npm run lite\" ",
-    "lite": "lite-server",
-    "postinstall": "typings install",
-    "tsc": "tsc",
+    "build": "webpack",
+    "start": "webpack-dev-server  & open http://localhost:8080/",
+    "dev": "webpack-dev-server --inline --hot & open http://localhost:8080/",
+    "try-dev": "tsc && concurrently \"npm run tsc:w\" \"npm run lite\" ",
```
vjeffrey: I don't know why my webpack/dev script isn't rebuilding, so to unblock myself I did the try-dev script with concurrently and lite like it was before. would love to figure out why webpack isn't working, but just making sure it doesn't delay everything by a day :)

vjeffrey:
@chris-rock  if you know what i'm doing wrong with webpack that it's not rebuilding on changes, please let me know!

www/tutorial/package.json
vjeffrey: actually, the script I added (try-dev) doesn't seem to be working.  i think the webpack one was especially not working before cause i had a port conflict -- now the webpack one is working with a refresh....it would be cool if it would just do that on its own, but i can live with it

vjeffrey: so i guess i should prob take out the extra stuff i put back in for that command. i'll do that a bit later

vjeffrey:
thank you thank you thank you @chris-rock :)

PR #1017, opened 2016-09-04
always display error message

### LINKED ISSUE #1016: Error output not informative
### Description

Before https://github.com/chef/inspec/commit/9468007f269ded90a4df0b853af23587d4ce520c the errors have been displayed in the following format:

screen shot 2016-09-04 at 6 37 22 pm

After the PR #946 the same test does not display the full result set:

screen shot 2016-09-04 at 6 37 47 pm

cc @vjeffrey 

This bug has been introduced with InSpec 0.32.0
### Replication Case

```
describe command("echo 'hellp'") do
  its('stdout') { should eq "hello" }
end
```

chris-rock:
fixes #1016

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -386,10 +386,18 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
       control_result = control[:results]
       title = control_result[0][:code_desc].split[0..1].join(' ')
       puts '  ' + title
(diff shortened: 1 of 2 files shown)


stevendanna:
👍 This looks correct, but to be honest this class has quiet a few conditionals and special cases.  I think that the next time we need to look at a bug fix in the reporter we should think about whether there is a refactoring that would slow the rate of conditionals we seem to need to add to the reporter.

chris-rock:
@stevendanna Yeah, I am close to refactor the cli reporter :-)

PR #1018, opened 2016-09-04
use winrm v2

chris-rock:
This PR depends on the following PRs to be merged:
- https://github.com/chef/kitchen-inspec/pull/94
- https://github.com/chef/train/pull/122


test/integration/default/vbscript_spec.rb
```diff
@@ -18,5 +18,5 @@
 
 # ensure that we do not require a newline
```
stevendanna, line 24: This comment implies that the property we were trying to test was explicitly that the user doesn't need to supply the line ending; but the change to the test implies they now do have to provide the line ending.

chris-rock: Great point, let me double-check that

test/unit/resources/powershell_test.rb
```diff
@@ -14,19 +14,17 @@
 
   it 'check if `powershell` for windows is properly generated ' do
     resource = MockLoader.new(:windows).load_resource('powershell', ps1_script)
-    if Gem.loaded_specs['winrm'].version < Gem::Version.new('1.6.1')
-      _(resource.command).must_equal 'powershell -encodedCommand IAAgACAAIAAjACAAYwBhAGwAbAAgAGgAZQBsAHAAIABmAG8AcgAgAGcAZQB0ACAAYwBvAG0AbQBhAG4AZAAKACAAIAAgACAARwBlAHQALQBIAGUAbABwACAARwBlAHQALQBDAG8AbQBtAGEAbgBkAAoA'
-    else
-      _(resource.command).must_equal 'powershell -encodedCommand JABQAHIAbwBnAHIAZQBzAHMAUAByAGUAZgBlAHIAZQBuAGMAZQA9ACcAUwBpAGwAZQBuAHQAbAB5AEMAbwBuAHQAaQBuAHUAZQAnADsAIAAgACAAIAAjACAAYwBhAGwAbAAgAGgAZQBsAHAAIABmAG8AcgAgAGcAZQB0ACAAYwBvAG0AbQBhAG4AZAAKACAAIAAgACAARwBlAHQALQBIAGUAbABwACAARwBlAHQALQBDAG8AbQBtAGEAbgBkAAoA'
-    end
+    # string should be the same
+    hash = Digest::SHA256.hexdigest(ps1_script.to_s)
```
stevendanna: It isn't clear to me why we are hashing these values first.

chris-rock: yeah, you're right, we do not need that :-)

stevendanna:
👍 An alternative to mutating the text that we get back would be to move the "indifferent to linespace ending" behavior into the matcher.

chris-rock:
@stevendanna i tried that first, but other powershell methods like `strip` rely on `response.stdout`. Therefore I decided for that specific approach. Also VBScript somethings treats 2 newlines as one. With this approach we are always sure.

PR #1023, opened 2016-09-06
Remove SourceReader::Inspec#prepare_load_path

stevendanna:
We already monkeypatch require so that it is redirected through the
require_loader.  All of the tests pass with this removal.  We might
cause some breakage with this removal that we aren't testing, but given
that we are mucking with `require` it seems preferable to have one
mechanism by which we do that and solve any bugs with that single path.


chris-rock:
I agree, that we should have only one implementation of `require`. Great finding @stevendanna

PR #1027, opened 2016-09-07
ssl resource fix and speed improvement

alexpop:
- Handled hostname differently for WinRM::Connection. Without this, remote windows scans are skipped with error: `Cannot determine host for SSL test. Please specify it or use a different target.` 
- Parallelize protocol checks to speed up scanning. For each port, five protocols checks are done. Since most ports are not SSL or blocked by firewall, scans can take minutes. Before parallelizing the calls, I scanned a vanilla Windows 2012r2 server in 90 seconds. After applying the change in this PR, it got down to 22 seconds.

### DIFF
--- a/lib/resources/ssl.rb
+++ b/lib/resources/ssl.rb
@@ -6,6 +6,8 @@
 
 require 'sslshake'
 require 'utils/filter'
+require 'uri'
+require 'parallel'
 
 # Custom resource based on the InSpec resource DSL
 class SSL < Inspec.resource(1)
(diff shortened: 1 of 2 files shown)


lib/resources/ssl.rb
```diff
@@ -44,6 +46,11 @@ class SSL < Inspec.resource(1)
   def initialize(opts = {})
     @host = opts[:host] ||
             inspec.backend.instance_variable_get(:@hostname)
+    # FIXME: This can be removed when/if @hostname is available as a property for 'Train::Transports::WinRM::Connection'
+    # Train enhancement request for this here: https://github.com/chef/train/issues/128
+    if @host.nil? && inspec.backend.class.to_s == 'Train::Transports::WinRM::Connection'
```
chris-rock, line 51: It feels like we do not want to add this here, but instead we want to improve train directly. Could you push a PR for train, so that we can depend on that?

arlimus: Please get this merged first. We can remove it in the second cycle and improve train. I don't think we should block this change because of a minor adjustment.

chris-rock:
The FIXME is followed up in https://github.com/chef/train/issues/128

Issue #959, opened 2016-08-23
Vendor Github and Supermarket dependencies


arlimus:
Part of https://github.com/chef/inspec/issues/888

Always pull the latest possible release. The user may specify the branch, which results in the latest commit in that branch. The user may also specify a fixed tag or commit.

(Version resolution via semver over tags is not a part of this ticket) 

User interface:

``` yml
- depends:
  - url: github://owner/repo
    [branch: master]
    [tag: nil]
    [ref: nil]
```

arlimus:
Decisions to be done on the URL pattern for github.

**Option 1**

```
github://owner/repo/branch/branch-name
github://owner/repo/tag/tag-name
github://owner/repo/ref/ref-name
```

Pro: All in one line, easy to copy
Con: New syntax (afaics); wordy and close to just specifying elements separately

**Option 2**

NPM/bower git resolver pattern: https://docs.npmjs.com/files/package.json

```
github://owner/repo#refname
```

Where refname can be a branch name, tag name, or short/full SHA ref. For git refname see: https://git-scm.com/docs/git-show-ref

Pro: All in one line, easy to copy, short, seen in the wild
Con: Refname can be ambiguous!

arlimus:
The additional elements that are contained and specified are:
- `branch` defaults to `master`
- `tag`
- `ref`

In a way, we also have:

**Option 3**

```
github://owner/repo
```

i.e. making it mandatory to specify tag/branch/ref elements manually.

Pro: Explicit, minimal
Con: Needs other attributes, not in one line

Although the last is shaky:

``` yaml
- depends:
  - { url: github://owner/repo, tag: 1.2.3 }
```

arlimus:
It's really hard to decide between these three!

However: Option 3 is the only one which _doesn't introduce a new syntax that people will depend on._

If we can clearly determine a winner between Options 1 and 2 we still have the chance to offer it as a new feature. For now, these two options are not supported and only Option 3 is offered.

CC: @chris-rock @stevendanna would love your input.

stevendanna:
Here are some thoughts:

Personally, I'm in favor of a path where no heuristics are used to determine the source-type.  That is, I'd like to eventually remove the regex-based, "is this a github url?" heuristic we currently have.  This would require some deprecation period most likely.

[...]

PR #1034, opened 2016-09-08
Add GitFetcher and rework Fetchers+SourceReaders

stevendanna:
This adds a new git fetcher. In doing so, it also refactors how the
fetchers work a bit to better support fetchers that need to resolve
user-provided sources to fully specified sources appropriate for a
lockfile.


lib/fetchers/git.rb
```diff
@@ -0,0 +1,119 @@
...
+    def archive_path
+      @repo_directory
+    end
+
+    def resolved_source
+      { git: @remote_url,
+        ref: resolved_ref }
+    end
+
+    def repo_path
```
stevendanna: This is unused, remove

lib/inspec/plugins/fetcher.rb
```diff
@@ -11,96 +11,28 @@ def self.plugin_registry
...
-      def abs_path(file)
-        return nil if file.nil?
-        prefix + file
+      def resolved_source
+        fail "Fetcher #{self} does not implement `resolved_source()`. This is required for terminal fetchers."
       end
 
-      def read(file)
-        @parent.read(abs_path(file))
+      def cacheable?
```
stevendanna: This is unused,remove.  Wether or not a fetcher adds items to the cache is determined by what it does with the argument to `fetch/1`

examples/remote-deps/inspec.yml
```diff
@@ -0,0 +1,15 @@
+name: remote-deps
```
chris-rock: I think we should add a readme for this profile. This should be targeting the built-up of a corporate compliance profile. I think we should execute some rules by default.

lib/bundles/inspec-compliance/target.rb
```diff
@@ -27,7 +26,7 @@ def self.resolve(target, _opts = {})
       # verifies that the target e.g base/ssh exists
       profile = uri.host + uri.path
       Compliance::API.exist?(config, profile)
-      super(target_url(config, profile), config)
+      resolve_next(target_url(config, profile), config)
```
chris-rock: Just to clarify this, we should add the behavior to our fetcher class. We are switching from an inheritance-like approach to full fetcher chaining.

stevendanna:
Test failures look like a difference in Tempfile's api across ruby version, will update this to use a more compatible call in my next update.

lib/bundles/inspec-compliance/target.rb
```diff
@@ -4,13 +4,12 @@
...
 require 'uri'
 require 'inspec/fetcher'
-require 'fetchers/url'
 
 # InSpec Target Helper for Chef Compliance
 # reuses UrlHelper, but it knows the target server and the access token already
 # similar to `inspec exec http://localhost:2134/owners/%base%/compliance/%ssh%/tar --user %token%`
 module Compliance
-  class Fetcher < Fetchers::Url
+  class Fetcher < Inspec.fetcher(1)
```
chris-rock: 👍

lib/bundles/inspec-supermarket/target.rb
```diff
@@ -17,7 +17,7 @@ def self.resolve(target, opts = {})
       return nil unless URI(target).scheme == 'supermarket'
```
chris-rock: should we change the

```
```

to 

```
def self.resolve(opts = {})
```

since we having a hash only. This would allow us to start with `:url` and resolve it from there?

lib/inspec/profile.rb
```diff
@@ -6,36 +6,49 @@
...
+        cache.prefered_entry_for(fetcher.cache_key)
+      else
+        fetcher.fetch(cache.base_path_for(fetcher.cache_key))
+        fetcher.archive_path
+      end
+    end
+
+    def self.for_path(path, opts)
+      file_provider = FileProvider.for_path(path)
+      reader = Inspec::SourceReader.resolve(file_provider.relative_provider)
```
stevendanna, line 42: From zoom discussion: Might be good to merge FileProvider and SourceReader.  We basically always want to work with source_reader's anyway.

lib/bundles/inspec-supermarket/target.rb
stevendanna: @chris-rock I'm 👍 on that overall. Will add it to the TODO list.

lib/fetchers/git.rb
```diff
@@ -0,0 +1,119 @@
...
+    def resolve_ref(ref_name)
+      cmd = shellout("git ls-remote \"#{@remote_url}\" \"#{ref_name}*\"")
+      ref = parse_ls_remote(cmd.stdout, ref_name)
+      if !ref
+        fail "Unable to resolve #{ref_name} to a specific git commit for #{@remote_url}"
+      end
+      ref
+    end
+
+    def parse_ls_remote(output, ref_name)
```
stevendanna, line 112: This is basically copy/pasta from the omnibus git fetcher.  There is a comment there explaining it a bit, we should copy/pasta the comment as well.

chris-rock: Please also add a shout out to the omnibus project for inspiration in our top header comment.

lib/inspec/dependencies/vendor_index.rb
```diff
@@ -23,12 +23,12 @@ def initialize(path = nil)
...
 
-    def add(name, source, path_from)
-      path_to = base_path_for(name, source)
+    def add(key, path_from)
+      path_to = base_path_for(key)
       path_to = if File.directory?(path_to)
                   path_to
                 elsif path_from.end_with?('.zip')
-                  "#{path_to}.tar.gz"
+                  "#{path_to}.zip"
```
chris-rock: great catch

lib/inspec/file_provider.rb
```diff
@@ -0,0 +1,219 @@
+# encoding: utf-8
+require 'rubygems/package'
+require 'zlib'
+require 'zip'
+
+module Inspec
+  class FileProvider
+    def self.for_path(path)
+      if path.is_a?(Hash)
```
chris-rock, line 9: feels like we want to have a resolve method in our provider and just iterate through the instances.

lib/inspec/profile.rb
chris-rock: Reading through the code, we may want to have the separation. Maybe the fileprovider should know about the source reader and return the right instance.

```
reader = file_provider.reader
```

examples/remote-deps/inspec.yml
stevendanna: Done.

lib/inspec/file_provider.rb
stevendanna: I'd like to leave this simple constructor for now since it is unlikely that we are going to support more archive types anytime soon.

stevendanna:
@chris-rock Regarding changing Fetcher.resolve to only take a single hash, I've held off on that for now.  Once I started it I realized that to do that we basically have to have another parsing mechanism somewhere else to translate from string -> hash and then suddenly some fetcher specific stuff lives outside that fetcher, which feels odd since they should be pluggable.  I agree that the resolve() method signature is pretty odd now, definitely in favor of refactoring it, just thinking we might do that in a follow up.

lib/bundles/inspec-compliance/target.rb
stevendanna: Done in a comment in the main fetcher class.

chris-rock:
fixes #959

lib/bundles/inspec-supermarket/api.rb
```diff
@@ -9,18 +9,14 @@ module Supermarket
...
     SUPERMARKET_URL = 'https://supermarket.chef.io'.freeze
 
-    def self.supermarket_url
-      SUPERMARKET_URL
-    end
-
     # displays a list of profiles
-    def self.profiles
-      url = "#{SUPERMARKET_URL}/api/v1/tools-search"
+    def self.profiles(supermarket_url)
```
chris-rock: This breaks our CLI:

```
b inspec supermarket profiles
/Users/chartmann/Development/compliance/inspec/lib/bundles/inspec-supermarket/api.rb:13:in `profiles': wrong number of arguments (0 for 1) (ArgumentError)
    from /Users/chartmann/Development/compliance/inspec/lib/bundles/inspec-supermarket/cli.rb:21:in `profiles'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:115:in `invoke'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor.rb:235:in `block in subcommand'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
    from /Users/chartmann/Development/compliance/inspec/bin/inspec:9:in `<top (required)>'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/bin/inspec:23:in `load'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/bin/inspec:23:in `<main>'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/bin/ruby_executable_hooks:15:in `eval'
    from /Users/chartmann/.rvm/rubies/ruby-2.2.1/lib/ruby/gems/2.2.0/bin/ruby_executable_hooks:15:in `<main>'
```

Looks like we are missing a functional test for supermarket

chris-rock: I propose to set the default to `SUPERMARKET_URL`

lib/fetchers/git.rb
```diff
@@ -0,0 +1,162 @@
...
+require 'inspec/log'
+
+module Fetchers
+  #
+  # The git fetcher uses the git binary to fetch remote git sources.
+  # Git-based sources should be specified with the `git:` key in the
+  # source hash. Additionally, we accept `:branch`, `:ref`, and `:tag`
+  # keys to allow users to pin to a particular revision.
+  #
+  # The code here is was in part borrowed is borrowed from
```
chris-rock: `borrowed is borrowed`  ;-)

PR #1035, opened 2016-09-08
add sys_info resource to get information about the hostname

chris-rock:
Adds the ability to requests the hostname of the current system:

```
describe sys_info do
  its('hostname') { should eq 'example.com' }
end
```


chris-rock:
This is required for some Windows improvements for https://github.com/chef/inspec/pull/990. We could also argue, this belongs into train

chris-rock:
quickly discussed with @arlimus that this does not belong into train

lib/resources/sys_info.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def hostname
+      os = inspec.os
+      if os.linux?
+        inspec.command('hostname').stdout.chomp
+      elsif os.windows?
+        # windows hostname
+        inspec.powershell('$env:computername').stdout.chomp
+      else
+        return skip_resource 'The `system` resource is not supported on your OS yet.'
```
stevendanna: You can elide the explicit return here if you want

lib/resources/sys_info.rb
```diff
@@ -0,0 +1,26 @@
...
+    example "
+      describe sysinfo do
+        its('hostname') { should eq 'example.com' }
+      end
+    "
+
+    def hostname
+      os = inspec.os
+      if os.linux?
+        inspec.command('hostname').stdout.chomp
```
stevendanna, line 18: I'm not familiar with what `$env:computername` returns in powershell.  I know in ohai we do some work to make sure that we only get the "local" part of the hostname (many of systems are configure such that hostname will return the FQDN).  Personally though, I think I like this better (i.e. keeping this info 1-to-1 with the commands of the same name.

stevendanna:
👍 Left some minor, non-critical comments.  It might be worth thinking about whether we want the same `hostname`, `fqdn`, and `machinename` split that ohai has.  The pro of the split is that it makes the semantics of what gets returned a bit more consistent even when run on differently configured systems.  The downside of the ohai method is that you need to understand the difference between those vaguely defined concepts since it doesn't map 1-to-1 with the command output.

Just food for thought.

chris-rock:
@stevendanna lets add those features in another iteration. For now it simply returns what the system is telling us :-)

PR #1036, opened 2016-09-08
Ensure simplecov starts before everything else

stevendanna:
Before this change, simplecov was reporting

```
1864 / 5198 LOC (35.86%) covered
```

After this change it is reporting

```
4131 / 5275 LOC (78.31%) covered.
```

Keeping the require at the top of the file ensure that simplecov is
loaded before any of our application code.


chris-rock:
Thanks @stevendanna for finding that!

PR #1038, opened 2016-09-09
Improve duplicate and cycle detection in resolver

stevendanna:

### DIFF
--- a/lib/inspec/dependencies/resolver.rb
+++ b/lib/inspec/dependencies/resolver.rb
@@ -31,7 +31,23 @@ module Inspec
       new.resolve(reqs)
     end
 
-    def resolve(deps, top_level = true, seen_items = {}, path_string = '')
+    def detect_duplicates(deps, top_level, path_string)
+      seen_items_local = []
+      deps.each do |dep|
+        if seen_items_local.include?(dep.name)
+          problem_cookbook = if top_level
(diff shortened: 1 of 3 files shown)


lib/inspec/dependencies/resolver.rb
```diff
@@ -31,25 +31,43 @@ def self.resolve(dependencies, vendor_index, working_dir, backend)
...
+    def detect_duplicates(deps, top_level, path_string)
+      seen_items_local = []
+      deps.each do |dep|
+        if seen_items_local.include?(dep.name)
+          problem_cookbook = if top_level
+                               'the inspec.yml for this profile.'
+                             else
+                               "the dependency information for #{path_string.split(' ').last}"
+                             end
+          fail Inspec::DuplicateDep, "The dependency #{dep.name} is listed twice in #{problem_cookbook}"
```
chris-rock, line 43: I think it may makes sense to add a test for that behavior?

stevendanna: There is a test that ensures we detect duplicates:

https://github.com/chef/inspec/pull/1038/files#diff-a1477bbe856a2707fce873a5096df630R28

Do you think we should test the content of the error message as well?

chris-rock: Uh missed that. Thanks @stevendanna

PR #1044, opened 2016-09-12
use simple config for security policy resource

chris-rock:
- Uses the simple config parser for `security_policy` resource
- harmonizes the api with `ssh_config`
- `security_policy.content` to retrieve the plain file content
- `security_policy.params` to retrieve the parsed parameters

### DIFF
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -13,72 +13,101 @@
 # All local GPO parameters can be examined via Registry, but not all security
 # parameters. Therefore we need a combination of Registry and secedit output
 
+require 'hashie'
+
 module Inspec::Resources
   class SecurityPolicy < Inspec.resource(1)
     name 'security_policy'
     desc 'Use the security_policy InSpec audit resource to test security policies on the Microsoft Windows platform.'
     example "
       describe security_policy do
-        its('SeNetworkLogonRight') { should eq '*S-1-5-11' }
+        its('SeNetworkLogonRight') { should include 'S-1-5-11' }
       end
     "
-    def initialize
-      @loaded = false
-      @policy = nil
-      @exit_status = nil
+
+    def content
+      read_content
     end
 
-    # load security content
-    def load
+    def params(*opts)
+      opts.inject(read_params) do |res, nxt|
+        res.respond_to?(:key) ? res[nxt] : nil
+      end
+    end
+
+    def method_missing(name)
+      params = read_params
+      return nil if params.nil?
+
+      # deep search for hash key
+      params.extend Hashie::Extensions::DeepFind
+      res = params.deep_find(name.to_s)
+      res
+    end
+
+    def to_s
+      'Security Policy'
+    end
+
+    private
+
+    def read_content
+      return @content if defined?(@content)
+
       # export the security policy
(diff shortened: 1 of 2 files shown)


lib/resources/security_policy.rb
```diff
@@ -13,72 +13,101 @@
...
-    def initialize
-      @loaded = false
-      @policy = nil
-      @exit_status = nil
+
+    def content
+      read_content
+    end
+
+    def params(*opts)
```
stevendanna, line 32: A number of the resources include a `params` function that feels similar.  Might be nice to pull this out into a module where the behavior can be tested once.

It's a bit strange that some of the resources take arguments for their params method while others return a hash.

Fine for now, just figured it was worth pointing out.

lib/resources/security_policy.rb
```diff
@@ -13,72 +13,101 @@
...
 
-    # load security content
-    def load
+    private
+
+    def read_content
+      return @content if defined?(@content)
+
       # export the security policy
       cmd = inspec.command('secedit /export /cfg win_secpol.cfg')
```
stevendanna, line 58: This is existing code, but is there any concern about using a known filename here rather than a temporary file?

lib/resources/security_policy.rb
```diff
@@ -13,72 +13,101 @@
...
-      val
     end
 
-    def to_s
-      'Security Policy'
+    def convert_hash(hash)
+      new_hash = {}
+      hash.each do |k, v|
+        v.is_a?(Hash) ? value = convert_hash(v) : value = extract_value(v)
+        new_hash[k.chomp.strip] = value
```
stevendanna: Minor: I think that the strip makes the chomp redundant.

lib/resources/security_policy.rb
```diff
@@ -13,72 +13,101 @@
...
-      }
+    def read_params
+      return @params if defined?(@params)
+      return @params = {} if read_content.nil?
 
-      # extract variable value
-      result = target.match(/[=]{1}\s*(?<value>.*)/)
+      conf = SimpleConfig.new(
+        @content,
+        assignment_re: /^\s*(.*)=\s*([^\n^\r^]*)\s*$/,
```
stevendanna: Is the trailing carat in this character class intended? `[^\n^\r^]`

chris-rock: otherwise the newlines get part of the content. any other idea?

lib/resources/security_policy.rb
chris-rock: Agree, resources need some harmonization here and it would be nice to extract some basics. This improvement makes sense to me...

lib/resources/security_policy.rb
chris-rock: let me simplify that....

lib/resources/security_policy.rb
chris-rock: Good point, I'd like to do that in another iteration

chris-rock: #1047 covers the requirement to use a temporary generated file

stevendanna:
This appears to be following the conventions of the other resources we have. 👍

PR #1050, opened 2016-09-12
fix webpack start script for tutorial

vjeffrey:
testing this morning I realized npm run dev was working fine but npm run start was not. this should fix it. @chris-rock


vjeffrey:
@chris-rock can you check really quick if everything looks right?  the formatting of the responses was off for me until i re-ran the update_demo task.

chris-rock:
Right, looks much better now.

PR #1061, opened 2016-09-12
update responses

vjeffrey:
noticed the formatting to txt files was off on master.  to test that this worked, I re-ran the rake command, committed, pushed, removed the www/ directory, checked out the branch so it was back to pushed state, npm install and npm run start. everything looks normal now.


stevendanna:
Is this mostly restoring the windows-style newlines that the xterm js library is expecting?  I'm wondering if it is possible to have the js side just accept unix-style line endings.  I think this is going to keep getting broken by git.

vjeffrey:
ooooooh you're probably right that must be what happened.
ugh

vjeffrey:
working theory: this is working on this branch for both steven and I. On the previous commit, the line endings had gone awry.  We think this has something to do with difference between peoples' git configs...  :/

stevendanna:
👍  I think it makes sense to merge this fixup for now.  Then, in a follow-up we can do:

1) Add a .gitattributes file so that git doesn't mangle the output regardless of config (we poked at this a bit but we couldn't coax git into doing the wrong thing locally).

or,

2) Change the javascript side to not care about the line ending style.  One way to to this would be to feed individual lines to xterm rather than the entire output.

PR #1073, opened 2016-09-14
inspec simulator

chris-rock:
To implement the InSpec tutorial, we want to simulate all different backends. This implements a mock for different backends, that can be used to generate files for the tutorial:

```
bundle exec ruby bin/inspec
bundle exec ruby bin/inspec exec path/to/profile -t docker://123456
bundle exec ruby bin/inspec exec path/to/profile -t winrm://user:password@host:22
bundle exec ruby bin/inspec exec path/to/profile -t ssh://user@host:22 -i testkey
```

cc @vjeffrey

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,3 @@ Puppetfile.lock
 profile.tar.gz
 omnibus/.cache
 omnibus/pkg
-www/tutorial/typings
-www/tutorial/node_modules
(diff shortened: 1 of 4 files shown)


vjeffrey:
yayyyyyyyyyyyyyyyyyyyyyyyy! thanks @chris-rock :)

PR #1079, opened 2016-09-15
Process transitive dependencies from lock file

stevendanna:
This is a regression introduced by the changes from string to symbol
keys in v0.34.0. It seems that our test cookbook that had a nested
dependency example wasn't actually wired up to run.

This adds a basic functional test and corrects the typo.


chris-rock:
Thanks for this fix @stevendanna

chris-rock (approved):

PR #1080, opened 2016-09-15
Allow users to reference resources from dependencies

stevendanna:
All resources from deps are added into the control_eval_context used by
the current profile. However, if there is a name conflict, the last
loaded resource wins. The new `require_resource` dsl method allows the
user to do the following:

```
require_resource(profile: 'profile_name',
                 resource: 'other',
                as: 'renamed')

describe renamed do
  ...
end
```

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -92,6 +92,14 @@ module Inspec
           res
         end
 
+        define_method :add_resource do |name, new_res|
+          resources_dsl.module_exec do
+            define_method name.to_sym do |*args|
+              new_res.new(@backend, name.to_s, *args)
+            end
+          end
+        end
+
         define_method :add_resources do |context|
           self.class.class_eval do
             include context.to_resources_dsl
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -72,6 +72,9 @@ module Inspec
       end
 
       # rubocop:enable Lint/NestedMethodDefinition
+      if __resource_registry.key?(name)
+        Inspec::Log.warn("Overwriting resource #{name}. To reference a specific version of #{name} use the resource() method")
+      end
       __resource_registry[name] = cl
     end
   end
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -17,7 +17,7 @@ module Inspec
                                    'attributes' => attributes })
     end
 
-    attr_reader :attributes, :profile_id, :resource_registry
+    attr_reader :attributes, :profile_id, :resource_registry, :backend
     attr_accessor :rules
     def initialize(profile_id, backend, conf)
       if backend.nil?
@@ -28,7 +28,8 @@ module Inspec
       @backend = backend
       @conf = conf.dup
       @rules = {}
-      @subcontexts = []
+      @control_subcontexts = []
+      @lib_subcontexts = []
       @require_loader = ::Inspec::RequireLoader.new
       @attributes = []
       # A local resource registry that only contains resources defined
@@ -46,7 +47,7 @@ module Inspec
     end
 
     def to_resources_dsl
-      Inspec::Resource.create_dsl(@backend, @resource_registry)
+      Inspec::Resource.create_dsl(self)
     end
 
     def control_eval_context
@@ -66,20 +67,34 @@ module Inspec
       @conf['profile'].supports_os?
     end
 
-    def all_rules
+    def all_controls
       ret = @rules.values
-      ret += @subcontexts.map(&:all_rules).flatten
+      ret += @control_subcontexts.map(&:all_rules).flatten
       ret
+    end
+    alias all_rules all_controls
+
+    def subcontext_by_name(name)
+      found = @lib_subcontexts.find { |c| c.profile_id == name }
+      if !found
+        @lib_subcontexts.each do |c|
+          found = c.subcontext_by_name(name)
+          break if found
+        end
+      end
+
+      found
     end
 
     def add_resources(context)
       @resource_registry.merge!(context.resource_registry)
       control_eval_context.add_resources(context)
+      @lib_subcontexts << context
       reload_dsl
     end
 
     def add_subcontext(context)
-      @subcontexts << context
+      @control_subcontexts << context
     end
 
     def load_libraries(libs)
--- a/lib/inspec/resource.rb
+++ b/lib/inspec/resource.rb
@@ -6,6 +6,8 @@
 require 'inspec/plugins'
 
 module Inspec
+  class ProfileNotFound < StandardError; end
+
   class Resource
     def self.default_registry
       @default_registry ||= {}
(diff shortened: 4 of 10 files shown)


lib/inspec/runner.rb
```diff
@@ -182,58 +182,10 @@ def register_rules(ctx)
 
     private
 
-    def block_source_info(block)
```
stevendanna, line 193: This just moved over to the rspec_formatter because I think it makes more sense there.  If we want I can do that in a different pull request.

chris-rock: be aware that `inspec json` should return the same information. Therefore I disagree with the move to the rspec formatter

lib/inspec/plugins/resource.rb
```diff
@@ -72,6 +72,9 @@ def inspec
       end
 
       # rubocop:enable Lint/NestedMethodDefinition
+      if __resource_registry.key?(name)
+        Inspec::Log.warn("Overwriting resource #{name}. To reference a specific version of #{name} use the resource() method")
```
chris-rock, line 76: That is a nice warning

lib/inspec/runner.rb
stevendanna: Minor correction on my original comment: I move this to the rspec_runner not the rspec formatter.

I'm not sure if that difference influences your comment.  My thinking was that the only callsite of this function seems to be called in check_to_example (did I miss one?) whose purpose seems tied precisely to the rspec_runner, so I moved both functions there.

stevendanna:
@arlimus We had a discussion about possible syntaxes for this feature.

# Default:

```
 gordon_config(args)
```

Will continue to work, gives access to the last eval'd resource with that name

# Option 1:

```
some_profile.gordon_config(args)
```

## Pros
- Feels nice to type, minimal extra characters

## Cons
- not all names safe as barewords - / .
- profile and resources that have the same name would still conflict

# Option 2:

```
resource('profile_d', 'gordon_config', args)
```

## Cons
- looks different than normal resource calls

## Pros
- Straightforward implementation (see PR)
- Very explicit

# Option 2b:

```
resource(profile: 'foobar', resource: 'gordon_config', args: ['foo', 'bar'])
```

## Pros
- Extensible
- Can be implemented on top of 2 later

## Cons
- Passing `args` is a bit cumbersome

# Option 3:

```
resource_name = resource(profile_name, resource_name)
resource_name(args)
```

With the potential shortcut (if we can implement it):

```
resource(profile_name, resource_name)
```

## Cons
- Might need some Ruby Magic(TM) to implement

## Pros
- Allows you to locally rebind resources to a new name

# Option 4

```
profile('profilename').resource_name(args)
```

## Pros
- `profile` is possibly good to hang other functions off of in the future

## Cons
- YAGNI?

stevendanna:
My personal vote is for Option 2 for now, moving to 2b if we need more features.

chris-rock:
I'd like options 2 with the possibility to use 2b in combination. Option 4 may become relevant later if we have more that resources to cover in a profile

stevendanna:
I've been looking at 3 a little bit more this morning.  I'm really skeptical that we'd be able to implement that without forcing the user to explicitly `.call` the returned resource.  Maybe I'm just missing the obvious implementation though.

arlimus:
Thank you for listing these options!

Option 1: Scrap due to naming + population of the namespace with profile names.
Option 2/3 are both good. 4 can be shortened to `profile(xyz).resource(abc)`.

Let's think about 2 in how it's being used:

```
describe resource('profile_d', 'gordon_config', 'this_is_my_arg') do
  ...
end
```

This feels very long and clunky. But It'll do for now.

Option 3 is the cleanest solution in my opinion and establishes a good pattern, by separating the point of resource declaration and resource calling. Imho it's a method to extend the context with a pre-defined name:

```
require_resource(profile: 'profile_name', resource: 'resource_name')
require_resource(profile: 'profile_name', resource: 'other', as: 'renamed')

describe resource_name do
   ...
end 

describe renamed do
  ...
end
```

If you can implement it ^^, I'd :+1: this path.

stevendanna:
> 4 can be shortened to profile(xyz).resource(abc).

Hrm, is `abc` here the name of the resource or the arguments to the resource. If the former, where would the arguments go?

The following proposal is pretty different than Option 3, let's call it Option 5:

```
require_resource(profile: 'profile_name', resource: 'resource_name')
require_resource(profile: 'profile_name', resource: 'other', as: 'renamed')

describe resource_name do
   ...
end 

describe renamed do
  ...
end
```

This is possible to implement (the problem with 3 is that I think we will run into issues with how the  ruby parser works). It raises the question of whether, given a `require_resource` DSL method we would want to do _any_ automatic loading of resources.  That horse probably already left the stable because of existing profile code though.

I think I like this the best.

chris-rock:
Since we return the resource, argument handling becomes as natural

```
require_resource(profile: 'profile_name', resource: 'resource_name')
describe resource_name(args) do
   ...
end
```

chris-rock:
We need to verify that once I do `require_resource`, it works in our DSL and in our RSPec context.

chris-rock:
@stevendanna great question if we should load resources by default or not. We need to find a smart way to do that. Right now, we allow single loading for resources, but what you want is more like a package loading eg.

```
require_resource(profile: 'profile_name', resource: '*')
```

Maybe that would be written as 

```
require_resource profile: 'profile_name'
# or 
require_resource profile: 'profile_name', resource: '*'
```

But we could extend that later. For now we can focus on the special case where we have to resources named the same

stevendanna:
> We need to verify that once I do require_resource, it works in our DSL and in our RSPec context.

I've added a test for this case.

lib/inspec/runner.rb
chris-rock: Not sure why we move this to the RSpecRunner. the runner is the right place. By moving this to the RSpec runner, this method is not available to the mock runner anymore.

lib/inspec/control_eval_context.rb
```diff
@@ -92,6 +92,14 @@ def to_s
           res
         end
 
+        define_method :add_resource do |name, new_res|
```
chris-rock, line 95: Do you think we should make those methods private?

lib/inspec/profile_context.rb
```diff
@@ -67,18 +68,31 @@ def profile_supports_os?
 
     def all_rules
```
chris-rock: we should rename this to `all_controls` and provide an alias for backwards-compatibility

lib/inspec/resource.rb
```diff
@@ -25,9 +27,21 @@ def self.new_registry
...
+      my_registry = profile_context.resource_registry
+
       Module.new do
+        define_method :resource_class do |profile_name, resource_name|
+          inner_context = profile_context.subcontext_by_name(profile_name)
+          fail ProfileNotFound, "Cannot find profile named: #{profile_name}" if inner_context.nil?
+          inner_context.resource_registry[resource_name]
+        end
+
+        define_method :resource do |profile_name, resource_name, *args|
```
chris-rock: Do we need that methods still?

lib/inspec/control_eval_context.rb
stevendanna: We can, but I'm not sure that will actually do what we want.  Since currently control code gets executed via instance_exec on this class, It'll still have access to private methods.

chris-rock: Alright, lets keep them as is

lib/inspec/runner.rb
stevendanna: I'll move it back.  The general reasons why I moved it was because with it here it makes are distinction between rspec runner a bit academic since this is tied directly to the rspec runner.  But this isn't particularly a hill I care about.

lib/inspec/resource.rb
stevendanna: That depends.  Do we want a means of accessing specific resources without mutating the current namespace or not?  If no, then we can certainly remove this.

test/unit/mock/profiles/dependencies/resource-namespace/controls/example.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe gordy_config do
+    its('version') { should eq('1.0') }
+  end
+
+  describe gordon_config do
+    its('version') { should eq('2.0') }
+  end
+
+  describe gordy_config do
+    its('version') { should eq(gordy_config.version) }
```
chris-rock, line 20: 👍

PR #1084, opened 2016-09-16
ssl resource to use inspec.backend.hostname and require train 0.19.1

alexpop:
Uses this change in train: https://github.com/chef/train/pull/150


arlimus:
Awesome Alex, thank you :) 👍

PR #1090, opened 2016-09-16
fix inspec shell handling

vjeffrey:
addresses inspec shell handling as mentioned in https://github.com/chef/inspec/issues/1089


arlimus:
Thank you for fixing this!

PR #1098, opened 2016-09-19
use Gem::Version instead of a regular expression for a test version bump

chris-rock:
We bumped to `1.0.0-beta1`, that version was not expected by our unit tests ;-)


stevendanna:
👍 Nice solution

PR #1099, opened 2016-09-19
Build with master of omnibus

yzl:
cc @arlimus 

Using `omnibus` master enables package builds on the linuxes and OS X.


chris-rock:
Thanks a lot @yzl

Issue #1100, opened 2016-09-19
Package Resource isn't searching 64-bit Registry Hives


seththoenen:
### Description

The package resource should search additional registry hives if the system has a 64-bit architecture. This seems to not be functioning as intended.
### InSpec and Platform Version

0.35 - Windows
### Replication Case

64-bit packages that are installed and have registry keys under the following hives do not show as installed during Inspec tests:

`HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*`
`HKCU:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*`
### Possible Solutions

In order to load the additional 64-bit hives, there is an if statement that checks the architecture of the system to see if it matches `x86_64`. If so, it will add the additional registry hives to search over. However, the method that is supposed to return the architecture value (`inspec.os.arch`) returns a value of '64-bit' instead of `x86_64` when on 64 bit editions of Windows Server

Changing the following line worked for my Inspec tests on Windows Server 2008 R2 and Windows Server 2012 R2:

/inspec/lib/resource/package.rb line 197

Current code: `if inspec.os.arch == 'x86_64'`
Proposed code: `if inspec.os.arch == '64-bit'`
### Stacktrace

No stack trace is available for this issue.

jerryaldrichiii:
@seththoenen, looks like the `64-bit` value is coming from `wmic os get * /format:list` command located [here](https://github.com/chef/train/blob/master/lib/train/extras/os_detect_windows.rb#L40).

Since that value is coming from Windows and the section of InSpec code you are referring to applies to Windows packages, I think your solution is valid.

chris-rock:
This has been solved with train 0.20.0 https://github.com/chef/train/releases/tag/v0.20.0 via https://github.com/chef/train/pull/151

PR #1102, opened 2016-09-20
Fix `inspec.os.arch` check in `WindowsPkg#info`

jerryaldrichiii:
This fixes issue #1100

### DIFF
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -194,7 +194,7 @@ module Inspec::Resources
       ]
 
       # add 64 bit search paths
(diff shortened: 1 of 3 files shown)


chris-rock:
@jerryaldrichiii Thanks for the fix. What do you think of fixing this at the root in train? I would prefer to convert the original values into our InSpec values here: https://github.com/chef/train/blob/master/lib/train/extras/os_detect_windows.rb#L53

chris-rock:
Thanks @jerryaldrichiii for this PR, I am going to close this PR in favor of your PR in train https://github.com/chef/train/pull/151

PR #1103, opened 2016-09-20
Fail if a remote source content doesn't match lockfile

stevendanna:
If a URL based source does not match the shasum recorded in the
lockfile, it likely means a new version has been pushed to the remote
source. In this case, we fail to help ensure that when using a lockfile
we always run the same code as when the lockfile was created.

### DIFF
--- a/lib/fetchers/local.rb
+++ b/lib/fetchers/local.rb
@@ -55,8 +55,19 @@ module Fetchers
       @target
     end
 
+    def cache_key
+      archive_shasum.to_s
+    end
+
+    def archive_shasum
+      return nil if File.directory?(@target)
+      @archive_shasum ||= Digest::SHA256.hexdigest File.read(@target)
+    end
+
     def resolved_source
-      { path: @target }
+      h = { path: @target }
+      h[:shasum] = archive_shasum if archive_shasum
+      h
     end
   end
 end
--- a/lib/fetchers/local.rb
+++ b/lib/fetchers/local.rb
@@ -56,17 +56,17 @@ module Fetchers
     end
 
     def cache_key
-      archive_shasum.to_s
+      sha256.to_s
     end
 
-    def archive_shasum
+    def sha256
       return nil if File.directory?(@target)
       @archive_shasum ||= Digest::SHA256.hexdigest File.read(@target)
     end
 
     def resolved_source
       h = { path: @target }
-      h[:shasum] = archive_shasum if archive_shasum
+      h[:sha256] = sha256 if sha256
       h
     end
   end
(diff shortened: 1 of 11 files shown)


lib/fetchers/url.rb
```diff
@@ -81,34 +82,61 @@ def initialize(url, opts)
     end
 
     def fetch(path)
-      Inspec::Log.debug("Fetching URL: #{@target}")
-      @archive_path = download_archive(path)
+      @archive_path ||= download_archive(path)
+    end
+
+    def shasum
```
chris-rock: maybe simply `sha256`?

lib/fetchers/local.rb
```diff
@@ -55,8 +55,19 @@ def archive_path
       @target
     end
 
+    def cache_key
+      archive_shasum.to_s
+    end
+
+    def archive_shasum
```
chris-rock: maybe just `sha256`?

lib/fetchers/url.rb
```diff
@@ -81,34 +82,61 @@ def initialize(url, opts)
...
+                  File.read(@archive_path)
+                else
+                  remote_archive_content
+                end
+      Digest::SHA256.hexdigest content
     end
 
     def resolved_source
-      { url: @target }
+      @resolved_source ||= { url: @target, shasum256: shasum }
```
chris-rock: `shasum256` What do you think about `sha256`?

lib/fetchers/url.rb
```diff
@@ -81,34 +82,61 @@ def initialize(url, opts)
...
+    def open_target
+      Inspec::Log.debug("Fetching URL: #{@target}")
       http_opts = {}
       http_opts['ssl_verify_mode'.to_sym] = OpenSSL::SSL::VERIFY_NONE if @insecure
       http_opts['Authorization'] = "Bearer #{@token}" if @token
+      open(@target, http_opts)
+    end
 
-      remote = open(@target, http_opts)
+    def remote_archive_content
```
chris-rock: maybe just `content`?

stevendanna:
Suggested renames pushed

lib/inspec/fetcher.rb
```diff
@@ -16,7 +16,7 @@ def resolve(target)
       end
     end
 
-    NON_FETCHER_KEYS = [:name, :version_constraint, :cwd, :backend, :cache].freeze
+    NON_FETCHER_KEYS = [:name, :version_constraint, :cwd, :backend, :cache, :shasum256].freeze
```
chris-rock: should this be `sha256`?

chris-rock:
fixes #1066

chris-rock:
Thanks @stevendanna

PR #1111, opened 2016-09-21
fixes debian package manager and some of the code examples

### LINKED ISSUE #1092: uninstalled package shows as installed
### Description

Some systems do not fully uninstall packages, they leave some files used in case of reinstall, and this leads to false positives. Especially when dealing with packages under Ubuntu and is in 'rc' stage, that is, it was not removed with --purge.

In general it shows what package manager shows, and it is true, but I think it is rather misleading.

inspec 0.32.0
Ubuntu 15.10, tests executed via ssh on Ubuntu 14.04.
### Replication Case

Test file:

``` ruby
title 'htop package'
control 'package-htop' do
  impact 1.0
  title 'htop package should be installed'
  desc 'htop package should be installed'
  describe package('htop') do
    it { should be_installed }
  end
end
```

Steps:
- install htop package
- run inspec test file, see the output, package is installed
- uninstall htop package
- run inspec test file, see the output, package is _still_ installed

see `dpkg -l |grep -i htop` showing the package is in `rc` state (removed, configured) so dpkg still shows package exists but actually it is in incomplete state.
### Possible Solutions

'installed' package should actually be installed, fix parsing of package manager's output.
Current test implementation for deb packages just checks exit code on `dpkg -s <package>`, which is not enough.
### Stacktrace

No stack trace.

Anirudh-Gupta:
fixes #1092

### DIFF
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -95,7 +95,7 @@ module Inspec::Resources
       # If the package is purged cmd fails with non-zero exit status
       {
         name: params['Package'],
-        installed: params['Status'].split(' ').first == 'install' ? true : false,
+        installed: params['Status'].split(' ').first == 'install',
         version: params['Version'],
         type: 'deb',
       }
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -91,9 +91,11 @@ module Inspec::Resources
         assignment_re: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
         multiple_values: false,
       ).params
+      # If the package is removed and not purged, Status is "deinstall ok config-files" with exit_status 0
+      # If the package is purged cmd fails with non-zero exit status
       {
         name: params['Package'],
-        installed: true,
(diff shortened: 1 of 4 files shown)


lib/resources/package.rb
```diff
@@ -91,9 +91,11 @@ def info(package_name)
         assignment_re: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
         multiple_values: false,
       ).params
+      # If the package is removed and not purged, Status is "deinstall ok config-files" with exit_status 0
+      # If the package is purged cmd fails with non-zero exit status
       {
         name: params['Package'],
-        installed: true,
+        installed: params['Status'].split(' ').first == 'install' ? true : false,
```
srenatus: This is equivalent to

```
installed: params['Status'].split(' ').first == 'install',
```

Anirudh-Gupta: Agree.

chris-rock (approved):

chris-rock:
Awesome, thanks @Anirudh-Gupta

PR #1112, opened 2016-09-21
fix double-space in pip to_s resource

### LINKED ISSUE #1043: 1.10.2 has an extra space in pip package output
With kitchen 1.10.2 as installed by ChefDK, I'm seeing another display error, this time when checking if Pip Packages are installed. There's an extra space between package and the name of the package as shown below:

```
✔  Pip Package  awscli should be installed
```

should be

```
✔  Pip Package awscli should be installed
```

This is with the following test stanza:

```
describe pip 'awscli' do
  it { should be_installed }
end
```

It also has the same extra space in failing pip package tests.

chris-rock:
fixes #1043

### DIFF
--- a/lib/resources/pip.rb
+++ b/lib/resources/pip.rb
@@ -49,7 +49,7 @@ module Inspec::Resources
     end
 
     def to_s
-      "Pip Package  #{@package_name}"
+      "Pip Package #{@package_name}"
     end
 
     private


stevendanna:
I'm not sure man, are you sure we don't need some regression tests here?

Only kidding, 👍

stevendanna (approved):

chris-rock:
@stevendanna You're right :-)

PR #1113, opened 2016-09-21
Add `--cache` option to `inspec exec`

stevendanna:
This allows users to run:

  inspec exec ./ --cache PATH

which will use `PATH` as the dir to retrieve and store remote
dependencies.  The hope is that this can eventually be used with
`inspec vendor PATH` to package up a profile for offline use.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -58,6 +58,8 @@ module Inspec
         desc: 'Use colors in output.'
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
+      option :cache, type: :string,
+        desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
     end
 
     private
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -1,14 +1,10 @@
 # encoding: utf-8
-require 'inspec/dependencies/vendor_index'
 require 'inspec/dependencies/requirement'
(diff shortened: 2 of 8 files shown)


stevendanna:
We still have a lot of inconsistency around whether we want to call it the `vendor_index` or the `cache`.  I'm leaning towards `cache` in all my edits since I think that more accurately conveys what it actually does.

chris-rock:
@stevendanna I like the wording `cache`

lib/inspec/dependencies/dependency_set.rb
```diff
@@ -1,5 +1,4 @@
 # encoding: utf-8
-require 'inspec/dependencies/vendor_index'
```
chris-rock, line 2: should we rename vendor_index to cache then?

stevendanna: Good call. I think I've completed the rename in my latest commit.

lib/inspec/base_cli.rb
```diff
@@ -58,6 +58,8 @@ def self.exec_options
         desc: 'Use colors in output.'
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
+      option :cache, type: :string, default: File.expand_path('~/.inspec/cache'),
```
stevendanna: Per discussion, let's nuke this default, the Cache class already has one.

chris-rock: +1

chris-rock (approved):

PR #1116, opened 2016-09-21
Always write lockfiles for local top-level profiles 

stevendanna:
This commit threads through some state related to whether or not a
profile is "local", that is whether it is a directory on disk.  If it
is, we then write out the lockfile to disk.

### DIFF
--- a/lib/fetchers/local.rb
+++ b/lib/fetchers/local.rb
@@ -55,6 +55,10 @@ module Fetchers
       @target
     end
 
+    def writable?
+      File.directory?(@target)
+    end
+
     def cache_key
       sha256.to_s
     end
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -60,8 +60,8 @@ module Inspec
         desc: 'Load attributes file (experimental)'
       option :cache, type: :string,
         desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
-      option :no_write_lockfile, type: :boolean, default: false,
-        desc: 'Do not write out a lockfile based on this execution'
+      option :create_lockfile, type: :boolean, default: true,
+        desc: 'Write out a lockfile based on this execution (unless one already exists)'
     end
 
     private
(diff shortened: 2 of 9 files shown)


stevendanna:
This is based on #1113

lib/inspec/base_cli.rb
```diff
@@ -60,6 +60,8 @@ def self.exec_options
         desc: 'Load attributes file (experimental)'
       option :cache, type: :string,
         desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
+      option :no_write_lockfile, type: :boolean, default: false,
```
chris-rock: that is leading to a strange cli output:
`[--no-write-lockfile], [--no-no-write-lockfile]  # Do not write out a lockfile based on this execution`
I think we need to switch to `write-lockfile` with default `true`

lib/fetchers/local.rb
```diff
@@ -55,6 +55,10 @@ def archive_path
       @target
     end
 
+    def writable?
+      File.directory?(@target)
```
chris-rock, line 59: Not all directories are writable

lib/inspec/base_cli.rb
chris-rock: maybe `create-lockfile`, since otherwise I am thinking its overwriting the old lockfile

lib/inspec/runner.rb
```diff
@@ -94,6 +96,18 @@ def run(with = nil)
       run_tests(with)
     end
 
+    def write_lockfile(profile)
+      return false if !profile.writable?
+
+      if profile.lockfile_exists?
+        Inspec::Log.debug "Using existing lockfile #{profile.lockfile_path}"
+      else
+        Inspec::Log.debug "Writing lockfile: #{profile.lockfile_path}"
```
chris-rock: `Writing` == `Creating`?

lib/fetchers/local.rb
stevendanna: True, in this context by writable? we simply mean "is this the type of source that I actually want to write new files into on disk".  We may still fail with EACCESS or something when we go to write.

chris-rock:
Thanks @stevendanna

PR #1124, opened 2016-09-22
Avoid spurious downloads during dependency management

stevendanna:
Before, a URL based source might be downloaded multiple times during the
dependency fetching and lockfile creation. This commit tries to avoid
this by:

1) Memoizing data about the archive to avoid re-fetching the archive

2) Adding a CachedFetcher wrapper around the fetcher class to help
ensure that callers always consult the cache before fetching.

### DIFF
--- a/lib/inspec/dependencies/requirement.rb
+++ b/lib/inspec/dependencies/requirement.rb
@@ -1,5 +1,5 @@
 # encoding: utf-8
-require 'inspec/fetcher'
+require 'inspec/cached_fetcher'
 require 'inspec/dependencies/dependency_set'
 require 'digest'
 
@@ -82,9 +82,7 @@ module Inspec
     end
 
     def fetcher
-      @fetcher ||= Inspec::Fetcher.resolve(opts)
-      fail "No fetcher for #{name} (options: #{opts})" if @fetcher.nil?
-      @fetcher
+      @fetcher ||= Inspec::CachedFetcher.new(opts, @cache)
     end
 
     def dependencies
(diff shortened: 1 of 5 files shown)


lib/inspec/dependencies/requirement.rb
```diff
@@ -94,17 +92,18 @@ def dependencies
...
+      return @profile if ! @profile.nil?
+
       opts = @opts.dup
-      opts[:cache] = @cache
       opts[:backend] = @backend
       if !@dependencies.nil?
         opts[:dependencies] = Inspec::DependencySet.from_array(@dependencies, @cwd, @cache, @backend)
       end
-      @profile ||= Inspec::Profile.for_target(opts, opts)
+      @profile = Inspec::Profile.for_fetcher(fetcher, opts)
```
chris-rock, line 106: I was thinking about this for a while, since it now deviated from our CLI code. We are using a separate mechanism for `inspec json` now.

stevendanna: Can you rephrase this comment? I'm don't understand what you are trying to indicate.

stevendanna: Discussion: We are going to leave for_fetcher() but change it so that for_target calls for_fetcher so that it is hard to introduce drifting behavior between the two, this will require the fetcher to expose the original target via an attr_accessor.

stevendanna: This is done now.

lib/inspec/profile.rb
```diff
@@ -72,9 +35,14 @@ def self.for_path(path, opts)
       new(reader, opts)
     end
 
+    def self.for_fetcher(fetcher, opts)
+      path, writable = fetcher.fetch
+      for_path(path, opts.merge(target: fetcher.target, writable: writable))
```
chris-rock, line 40: wouldn't it be better to attach the fetcher to the profile and just read the target then from the fetcher?

stevendanna: @chris-rock My plan was to do soemthing like that in a follow up PR since I think that is ultimately what we need to do to remove the Requirement class.

chris-rock: Alright, lets go for that approach

chris-rock:
Thanks @stevendanna for this improvement.

PR #1126, opened 2016-09-22
Improve error messages from compliance fetcher

stevendanna:

### DIFF
--- a/lib/bundles/inspec-compliance/target.rb
+++ b/lib/bundles/inspec-compliance/target.rb
@@ -4,6 +4,7 @@
 
 require 'uri'
 require 'inspec/fetcher'
+require 'inspec/errors'
 
 # InSpec Target Helper for Chef Compliance
 # reuses UrlHelper, but it knows the target server and the access token already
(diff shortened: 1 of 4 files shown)


chris-rock:
I am not convinced that a fetcher should exit the application. It may throw an exception, but it should never exit the application. If used as a library (as in audit cookbook), the fetcher is not executed as cli application. From my perspective only the CLI application should decide if an exit code should be thrown or not

lib/inspec/ui.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+module Inspec
+  module UI
+    def self.exit_with_error(message)
+      $stderr.puts message
+      exit(1)
```
chris-rock: I strongly advice against to exit application from an internal method.

lib/bundles/inspec-compliance/target.rb
```diff
@@ -24,11 +25,23 @@ def self.resolve(target)
 
       # check if we have a compliance token
       config = Compliance::Configuration.new
-      return nil if config['token'].nil?
+      if config['token'].nil?
+        Inspec::UI.exit_with_error <<EOF
```
chris-rock: Its a great idea to add better information about an error, but that should be encapsulated in exceptions

chris-rock (changes requested):

stevendanna:
@chris-rock Yeah, I am not thrilled with exiting from here either.  One of the issues with the exception is not dumping the stack trace which obscures the error message.

stevendanna:
Updated per code review

chris-rock (approved):

PR #1135, opened 2016-09-22
implement filter table for group/groups resource

chris-rock:
- implements filtertable for the new `groups` resource
- uses `groups` behind the scenes
- removes the domain support for windows (temporary)
- no WMI 👍


arlimus:
Awesome! :grin: :+1:

PR #1148, opened 2016-09-25
Extend example for parse_config.rb

nvtkaszpir:
Extend example section.


chris-rock:
Thanks for this great example @nvtkaszpir

PR #1143, opened 2016-09-23
Truncate long filename. Temporary fix

stevendanna:
This is a temporary fix that somehow needs to be integrated into the
command generator

Signed-off-by: Steven Danna steve@chef.io

PR #1159, opened 2016-09-26
tutorial response filename length

chris-rock:
Followup on #1143 to prevent long filenames for Windows environments


arlimus:
After some investigation, I found the underlying issue: Some of our systems generate CRLFs in these git commits, while @vjeffrey 's last update cleared it to LF line endings. IMHO this is the way we should choose instead of CRLF.

Github doesn't provide insight to this in their diff view.
Git on CLI can help, if you don't ignore CRLF diffs. They will show up as `^M` characters at all line endings.

See http://stackoverflow.com/questions/1822849/what-are-these-ms-that-keep-showing-up-in-my-files-in-emacs

---

At this point I think we should avoid having all of these responses checked into the repo. More investigation...

arlimus:
Limited to the change without response updates for now. Kudos @chris-rock for atomic commits!!

vjeffrey:
I pushed a commit to regenerate the responses...

arlimus:
Thank you @chris-rock for fixing this!!
Kudos @vjeffrey for the updated responses.

PR #1160, opened 2016-09-26
Fix require_controls DSL method

stevendanna:
Previous, require_controls was including all controls from the named
profile, despite the documented behavior being that it only includes
controls explicitly pulled in by the user.  The cause was two-fold:

1) A previous refactor meant that we weren't removing the rule from the
correct context, and

2) We weren't descending down the dependency tree when filtering rules.

This commit fixes the require_controls DSL method and adds a test to
help prevent future regressions.

### DIFF
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -53,12 +53,15 @@ EOF
   end
 
   def self.filter_included_controls(context, profile, &block)
-    include_ctx = profile.runner_context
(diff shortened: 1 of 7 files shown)


chris-rock (approved):

alexpop:
Thanks Steven for the quick turnaround!

Tested both `include_controls` and `require_controls` and they work as expected with this change!

PR #1170, opened 2016-09-26
print profile info before test results (inherited profiles)

### LINKED ISSUE #1071: Confusing reporter output with inherited profiles
### Description

When using a profile that inherit all of their controls, the CLI test reporter confusingly places the profile information at the end of the output rather than the begging:

```
Profile: Meta Compliance Profile (meta-profile)
Version: 0.2.0
Target:  docker://d7f7ab8d3263af087ad5860b8276a2607b2ddea92585e52aeb2b807e733cc524


Summary: 60 successful, 39 failures, 68 skipped
```

0.34.0
### Replication Case

In the inspec source code repository:

```
cd examples/inheritance
inspec exec ./
```

vjeffrey:
fixes https://github.com/chef/inspec/issues/1071

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -279,6 +279,7 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
     @missing_controls = []
     @anonymous_tests = []
     @control_tests = []
+    @profile_printed = false
     super(*args)
   end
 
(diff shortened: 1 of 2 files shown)


vjeffrey:
now:
screen shot 2016-09-26 at 1 09 51 pm

before: 
screen shot 2016-09-26 at 1 11 28 pm

stevendanna (approved):
👍 I think that the observed behavior with this patch is an improvement.  @vjeffrey and I discussed it a bit and there is still some unintended behavior when multiple profiles are passed on the command line, but we think that fixing that puts us past the tipping point where we actually need to just fix how controls and profiles are re-associated in the reporter.  

That is a long way to say, let's merge this and do that work in another commit.

Issue #995, opened 2016-08-30
Process user should eq ["longusername"]: usernames get truncated with a '+' at the end


TD-4242:
### Description

When testing for a username running a command if it is more than 8 characters it gets truncated as per the default ps command.
### InSpec and Platform Version

inspec 0.29.0 from chefdk 0.17.17
### Replication Case

```
describe processes('process') do
  its('users') { should eq ['longusername'] }
end

  1) Processes users should eq ["longusername"]
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }

       expected: ["longusername"]
            got: ["longuse+"]
```
### Possible Solutions

http://askubuntu.com/questions/523673/ps-aux-for-long-charactered-usernames-shows-a-plus-sign
### Current workaround

shorten test to first 7 letters and a plus(+) sign:

```
describe processes('process') do
  its('users') { should eq ['longuse+'] }
end
```

chris-rock:
@TD-4242 Thanks for reporting. Are you aware of a solution that works without a fixed character length?

TD-4242:
I don't, but I can dig a little further, I'm sure there is something.

chris-rock:
That would be awesome.

TD-4242:
`ps -C <process_name> -o user=`

might be the trick?

alexpop:
Hey Bill, can you try out the code in this PR?
https://github.com/chef/inspec/pull/1225

Here's how you can try it out:

```
git clone https://github.com/alexpop/inspec
cd inspec
git checkout ag-ap/processes-user-comm
bundle install
bundle exec inspec exec ...
```

alexpop:
I'm going to merge #1225.
Bill, @Anirudh-Gupta and everyone involved with this issue. Your help is greatly appreciated!

PR #1173, opened 2016-09-26
fix processes resource for os's where username is long to avoid truncation

Anirudh-Gupta:
Proposed fix for #995 
Moves the username at the end of the field list to avoid getting truncated in OS's like Ubuntu. The last field does not get truncated, however long it may be.

### DIFF
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -26,7 +26,7 @@ module Inspec::Resources
         grep = '(/[^/]*)*'+grep if grep[0] != '/'
         grep = Regexp.new('^' + grep + '(\s|$)')
       end
-      all_cmds = ps_aux
+      all_cmds = ps_axo
       @list = all_cmds.find_all do |hm|
         hm[:command] =~ grep
       end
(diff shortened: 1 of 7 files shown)


chris-rock:
Just discussed with @srenatus about this. @Anirudh-Gupta this is a great idea. Will this lead to a truncated process name?

Anirudh-Gupta:
@chris-rock 
Just took out one of the entry from ps aux command and the corresponding from ps axo. Have a look.
`libvirt+  1822  0.0  0.0  28204   976 ?        S    Sep26   0:00 /usr/sbin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/default.conf`

`1822  0.0  0.0  28204   976 ?        S    23:02:33 00:00:00 dnsmasq         libvirt-dnsmasq`

Process name does not get truncated but instead of using a full command, it uses a shorter name.

vjeffrey:
@Anirudh-Gupta tested this morning using inspec shell to make sure we can parse correctly.  something seems to be off:  running inspec shell normally and testing a user for a process is resolving correctly but running from this branch with a bundle exec inspec shell is giving me output such as: 
screen shot 2016-09-30 at 12 25 24 pm

Anirudh-Gupta:
@vjeffrey Strange. Works fine for me. Not able to reproduce/replicate it.

screen shot 2016-10-03 at 12 54 24 pm

vjeffrey:
hmm. i'll look into it some more today then. thanks!

sorry, I was not being very coherent on Friday :)  

my monday- er, tuesday - brain seems to be doing a little better.  the problem is longer process names get truncated, which means that if you test for, say 'upstart-udev-bridge', you end up with that output i pasted above, because the process name is then set to 'upstart-udev-br'.  (testing for 'upstart-udev-br' gives the correct output in this case)

so, the question becomes 1) is there a way to do this without truncating long usernames or processes?  2) if not - which one do we default to?

@Anirudh-Gupta @arlimus    what do ppl think?

Anirudh-Gupta:
Just to add to it - user field is limited by the kernel to 8 bytes(7 characters + terminating NUL byte), whereas comm field is limited to 16 bytes(15 characters + terminating NUL byte).

If we are going with 2) - I feel usernames could be **almost** similar(like 'anirudh' or 'anirud' or 'aniruddh') but process names won't be(plus process names have extra bytes to be differentiated).
1) is undoubtedly the  best. But could not find any solution to it yet.

arlimus:
Suggestion: How about running the `ps` command twice in the same execution i.e. `command('ps ... && ps ...')`. It will still have the issue where the process might have gone away and we need merging of both;

The only other option we have (afaics) is: run the process to unly return user IDs (not names) because they are not truncated, and use them together with the users resource to map userIDs to names. i.e. also include both in the process resource (since the API is stable we still want the user name to be in here)

Priority 1: Get untruncated results for both user and command (and everything else!)
Priority 2: Optimize for speed e.g. only run 1 command

alexpop:
The command used in the PR is not showing the process arguments for Linux:

```
[root@ap-cc6 ~]# ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,comm,user | grep '1470'
system_u:system_r:init_t:s0      1470  0.0  3.2 415332 50284 ?        Ssl  22:51:19 00:00:11 core            chef-compliance
```

We need to use the `command` option, but this is getting truncated as well if it's not the last option:

```
[root@ap-cc6 ~]# ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,command,user | grep '1470'
system_u:system_r:init_t:s0      1470  0.0  3.2 415332 50284 ?        Ssl  22:51:19 00:00:11 /opt/chef-compliance/embedd chef-compliance
```

I found a solution that gives us what we need in one go. We keep `command` as the last option, and increase the column size for username, for example:

```
[root@ap-cc6 ~]# ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command | grep '1470'
system_u:system_r:init_t:s0      1470  0.0  3.2 415332 50284 ?        Ssl  22:51:19 00:00:11 chef-compliance                  /opt/chef-compliance/embedded/service/core/bin/core run --mode=setup --oidc --oidc-issuer-url https://ap-cc6.opschef.tv --oidc-client-id JZik-QcuP1LK2lOmL8l1qQwyazeI6XoUhvYnOb7oFz0=@ap-cc6.opschef.tv --oidc-redirect-url https://ap-cc6.opschef.tv/api/callback
```

^ tested on CentOS 6.6 and Ubuntu 14.04

chris-rock:
@Anirudh-Gupta anything we can help you with?

alexpop:
Hi Anirudh, I'm going to fork your change and update it with the command I mentioned here: https://github.com/chef/inspec/pull/1173#issuecomment-252898743

I'll send over the new PR when it's ready for review.
Thanks for raising this issue and working together with us on a solution.

Anirudh-Gupta:
#1225 fixes this. Hence closing this. Thanks to @alexpop

PR #1175, opened 2016-09-27
Remove default parameter in `updateInstructions()`

### LINKED ISSUE #1165: Demo at `http://inspec.io/` stuck on `Loading`
### Description

Both the `Try the Demo` and the `Start the Demo` button redirect to a black page that just says `Loading`.

screen shot 2016-09-26 at 10 22 10 am
### Platforms Tested

Kudos to @seththoenen for testing on Windows 10 for me

Fails:
- Chrome 47.0.2526.106 on Mac OS X 10.11.5
- Safari 9.1.1 (11601.6.17) on Mac OS X 10.11.5
- Internet Explorer 11.0.9600 on Windows Server 2012 R2
- Internet Explorer 11.545.10586.0 on Windows 10
- Edge 25.10586.0.0 on Windows 10

Passes:
- Firefox 47.0.1 on Mac OS X 10.11.5
- Firefox 45.0.01 on Windows 10
- Chrome 53.0.2785.116 on Windows 10
### More Info

When using Chrome's `inspect` feature a `Uncaught SyntaxError: Unexpected Token =` error is displayed and references `inspec_tutorial.js` line `2165`. The line is:

```
updateInstructions(step = null) {
```

jerryaldrichiii:
Some browsers [do not support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Browser_compatibility) the use of default parameters. 

Moving the logic that sets `step` to null if undefined into the function body allows those browsers to render the tutorial correctly.

This fixes #1165

### DIFF
--- a/www/tutorial/app/app.component.ts
+++ b/www/tutorial/app/app.component.ts
@@ -68,7 +68,7 @@ export class AppComponent implements OnInit {
   }
 
   // called when tutorial commands need to be displayed
-  updateInstructions(step = null) {
+  updateInstructions(step?) {
     // if step is given, we calculate the new index
     let totalSteps = this.instructionsArray.length - 1;
     switch(step) {


www/tutorial/app/app.component.ts
```diff
@@ -68,7 +68,8 @@ export class AppComponent implements OnInit {
   }
 
   // called when tutorial commands need to be displayed
-  updateInstructions(step = null) {
```
chris-rock, line 71: since we check the step value anyway, we do not have to set the default value, lets just use `updateInstructions(step)`

www/tutorial/app/app.component.ts
```diff
@@ -68,7 +68,8 @@ export class AppComponent implements OnInit {
...
+  updateInstructions(step?) {
+    if (typeof step === 'undefined') { step = null; }
```
chris-rock: I think we do not need this, since the `switch` statement should cover us

jerryaldrichiii:
Agreed @chris-rock. PR has been updated 😄

Leaving `updateInstructions(step?)` instead of changing to `updateInstructions(step)` though because we call this method with no arguments [at line 216](https://github.com/chef/inspec/blob/master/www/tutorial/app/app.component.ts#L216).

vjeffrey:
thanks @jerryaldrichiii !!!  not sure what's up with the random travis test failing, and appveyor has an odd error too, since the change shouldn't have affected any of that. trying to look into it/see if i have permissions to rerun those

jerryaldrichiii:
Thanks @vjeffrey. Looks like a Ruby issue. 

Specifically:

```
Gem::InstallError: term-ansicolor requires Ruby version >= 2.0
```

It looks like Ruby 1.9.3 is on those nodes:
https://ci.appveyor.com/project/chef/inspec/build/1.0.1918/job/dvp7ixv27wa0jho2#L312
https://travis-ci.org/chef/inspec/jobs/163080067#L354

vjeffrey:
we discovered it was a problem with term-ansicolor, which requires ruby 2.0+ now, which is a dependency of inquirer.  plan is to move that from the test group to a new group called exclude that group when bundle installing in travis

resolved with https://github.com/chef/inspec/pull/1178

once this pr is rebased on master we should be g2g @jerryaldrichiii

jerryaldrichiii:
I rebased my branch to include your fix @vjeffrey.

Yay teamwork! ✋

PR #1178, opened 2016-09-27
move inquirer to deploy

vjeffrey:
this will unblock us. at the moment, we're failing out all 1.9.3 tests on appveyor and travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache: bundler
 dist: trusty
 services:
 - docker
-bundler_args: "--without integration tools maintenance"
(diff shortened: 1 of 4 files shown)


vjeffrey:
i'm not actually certain what we were using inquirer for (in tasks/www.rb) ...I don't see any references to it outside of that first require statement. I didn't feel comfortable yanking it out, since I'm unsure about that, but I did comment it out at the moment, so that we can move forward in our pipeline

arlimus (approved):
Thank you for loosening the binding here @vjeffrey !

PR #1180, opened 2016-09-27
include control section in instructions

vjeffrey:
noticed we managed to leave out the controls part...

### DIFF
--- a/www/tutorial/app/app.component.ts
+++ b/www/tutorial/app/app.component.ts
@@ -70,7 +70,7 @@ export class AppComponent implements OnInit {
   // called when tutorial commands need to be displayed
   updateInstructions(step?) {
     // if step is given, we calculate the new index
-    let totalSteps = this.instructionsArray.length - 1;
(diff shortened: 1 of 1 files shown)


vjeffrey:
um, also, i had merged a change for the site earlier, and some changes to the main website hadn't been updated yet, so i just did the whole build to gh pages thing. confirmed all was working and good.....kinda maybe did it from this branch to get this fix in too i'm sorry i know that was prob not best form but i felt silly waiting sorrrryyyyyyy :) :) :)

www/tutorial/app/app.component.ts
```diff
@@ -73,7 +73,7 @@ export class AppComponent implements OnInit {
     let totalSteps = this.instructionsArray.length - 1;
```
chris-rock: may it makes sense to just change it to `let totalSteps = this.instructionsArray.length`?

vjeffrey: yup! just pushed up a change to that. :) nice catch!

PR #1199, opened 2016-10-05
Missing registry keys should not exist

alexpop:
Fixes https://github.com/chef/inspec/issues/1196


arlimus (approved):
Kudos @kenmacleod and @alexpop for finding and fixing this issue!

PR #1200, opened 2016-10-05
bugfix: always use the mock backend for inspec archive

arlimus:
Otherwise it will run commands against the local stack.


alexpop:
nice catch Dom, thank you

PR #1204, opened 2016-10-05
cmp code_desc missing operation and expected valid

alexpop:
Having the following spec file:

``` ruby
describe 6 do
  it { should cmp 6 }
end

describe 7 do
  it { should cmp '7' }
end

describe 8 do
  it { should cmp <= '10' }
end

describe 9 do
  it { should cmp == 9 }
end
```

I get the following exec output with inspec 1.1.0(latest):

``` ruby
$ be inspec exec ~/tmp/cmp.rb
Target:  local://

  6 should
     ✔  cmp 6
  7 should
     ✔  cmp "7"
  8 should
     ✔  cmp
  9 should
     ✔  cmp
```

^ see the missing operator and value for 8 & 9

The code change in this PR produces the following exec output:

``` ruby
$ be inspec exec ~/tmp/cmp.rb
Target:  local://

  6 should
     ✔  cmp == 6
  7 should
     ✔  cmp == "7"
  8 should
     ✔  cmp <= "10"
  9 should
     ✔  cmp == 9
```

8 & 9 have operator and value 🎉

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -313,6 +313,10 @@ RSpec::Matchers.define :cmp do |first_expected|
     actual = '0' + actual.to_s(8) if octal?(@expected)
     "\n" + format_expectation(true) + "\n     got: #{actual}\n\n(compared using `cmp` matcher)\n"
   end
+
+  description do
+    "cmp #{@operation} #{@expected.inspect}"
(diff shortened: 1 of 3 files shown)


arlimus:
Awesome bugfix, kudos @alexpop :+1:

Issue #783, opened 2016-06-06
File resource permissions for windows


mhedgpeth:
### Description

Right now permissions can't be tested with inspec on windows.

This won't work:

``` ruby
control 'chef-installation-should-be-readable' do
  impact 1
  title 'Chef Installation should be readable by vagrant user'
  desc 'To make sure we can run chef, the install should be readable'
  describe directory('C:/opscode/chef') do
    it { should be_readable.by('vagrant') }
  end
end
```

It gives the error:

```
 1) File C:/opscode/chef should be readable by vagrant
    Failure/Error: fail 'Checking file permissions is not supported on your os'

    RuntimeError:
      Checking file permissions is not supported on your os
    # ./test/integration/default/inspec/image.rb:27:in `block (3 levels) in load'
```
### InSpec and Platform Version

0.24
### Replication Case

Create a cookbook with inspec as the verifier with `ncr/win2008r2` as the virtualization box for kitchen. Run `kitchen verify`.
### Possible Solutions

It's simply not coded yet in the File resource. I assume this is because file permissions on the command line are difficult; it's probably easier on Powershell but I'm not sure how you do that.

chris-rock:
@mhedgpeth Thanks for reporting. We need to break this down into two cases:
-  `it { should be_readable.by('owner') }` will not be supported on Windows since this is requesting unix file permissions. I agree that this is hard to read. We added this to be compatible with Serverspec http://serverspec.org/resource_types.html#file
- `it { should be_owned_by 'vagrant' }` should be supported on Windows. The implementation is not trivial, but should be part of InSpec.

```
control 'chef-installation-should-be-readable' do
  impact 1
  title 'Chef Installation should be readable by vagrant user'
  desc 'To make sure we can run chef, the install should be readable'
  describe directory('C:/opscode/chef') do
    # verifies specific users
    it { should be_owned_by 'vagrant' }
    it { should be_readable.by_user('vagrant') }
    it { should be_writable.by_user('vagrant') }

    # only applicable to unix group system
    # it { should be_readable.by('owner') }
    # it { should be_readable.by('group') }
    # it { should be_readable.by('others') }
  end
end
```

chris-rock:
The following powershell commands are useful to get this running:

```
# File permissions for user
  (Get-Acl .\test).Access | ConvertTo-Json
  [
    {
        "FileSystemRights":  2032127,
        "AccessControlType":  0,
        "IdentityReference":  {
[...]

PR #1215, opened 2016-10-12
adding by_user permissions support for windows

jeremymv2:
This is a first stab at adding mvp support for checking read/write/execute file permissions, for a specific user.

```
  # verifies if a specific user has access to a file
    describe file("C:/users/vagrant/setup.bat") do
      it { should exist }
      it { should_not be_readable.by_user('vagrant-2012-r2\jdoe') }
      it { should be_readable.by_user('vagrant-2012-r2\vagrant') }
      it { should_not be_executable.by_user('vagrant-2012-r2\jdoe') }
      it { should be_executable.by_user('vagrant-2012-r2\vagrant') }
      it { should_not be_writable.by_user('vagrant-2012-r2\jdoe') }
      it { should be_writable.by_user('vagrant-2012-r2\vagrant') }
      it { should be_readable }
      it { should be_writeable }
      it { should be_executable }
    end
```

The above results in the following when `vagrant-2012-r2\jdoe` was denied access to the file:

screen shot 2016-10-11 at 9 59 33 pm

I believe it addresses #783 

It will fail unit tests since they will have to be refactored.  If this passes muster I'll refactor the tests.

Signed-off-by: Jeremy J. Miller jm@chef.io

### DIFF
--- a/lib/resources/file.rb
+++ b/lib/resources/file.rb
@@ -7,18 +7,19 @@
 require 'shellwords'
 
 module Inspec::Resources
-  module FilePermsSelector
+  module FilePermissionsSelector
     def select_file_perms_style(os)
       if os.unix?
-        @perms_provider = UnixPermission.new(inspec)
+        @perms_provider = UnixFilePermissions.new(inspec)
       elsif os.windows?
-        @perms_provider = WindowsPermission.new(inspec)
+        @perms_provider = WindowsFilePermissions.new(inspec)
       end
     end
   end
 
   class FileResource < Inspec.resource(1)
-    include FilePermsSelector
+    include FilePermissionsSelector
+    include MountParser
 
     name 'file'
     desc 'Use the file InSpec audit resource to test all system file types, including files, directories, symbolic links, named pipes, sockets, character devices, block devices, and doors.'
@@ -32,7 +33,6 @@ module Inspec::Resources
         its('mode') { should cmp '0644' }
       end
     "
-    include MountParser
 
     attr_reader :file, :mount_options
     def initialize(path)
@@ -153,14 +153,14 @@ module Inspec::Resources
     end
   end
 
-  class FilePermission
+  class FilePermissions
     attr_reader :inspec
     def initialize(inspec)
       @inspec = inspec
     end
   end
 
-  class UnixPermission < FilePermission
+  class UnixFilePermissions < FilePermissions
     def check_file_permission_by_user(user, access_type, path)
       flag = case access_type
              when 'read'
@@ -187,7 +187,7 @@ module Inspec::Resources
     end
   end
 
-  class WindowsPermission < FilePermission
+  class WindowsFilePermissions < FilePermissions
     def check_file_permission_by_user(user, access_type, path)
       access_rule = case access_type
                     when 'read'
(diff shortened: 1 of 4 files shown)


jeremymv2:
@chris-rock @arlimus thoughts?

lib/resources/file.rb
```diff
@@ -20,11 +33,13 @@ class FileResource < Inspec.resource(1) # rubocop:disable Metrics/ClassLength
...
       end
     "
-    include MountParser
 
     attr_reader :file, :mount_options
     def initialize(path)
+      # select permissions style
+      @perms_provider = select_file_perms_style(inspec.os)
       @file = inspec.backend.file(path)
+      return skip_resource 'The `file` resource is not supported on your OS yet.' if @perms_provider.nil?
```
chris-rock: I think we should not skip the resource if the no perm_provider is available. Most of the file implementation is done in train anyway. I would prefer to move this skip return to the individual functions that cannot be executed aka `readable?` etc

lib/resources/file.rb
```diff
@@ -52,19 +67,19 @@ def contain(*_)
     def readable?(by_usergroup, by_specific_user)
       return false unless exist?
```
chris-rock, line 68: This is the place where we should return the skip the resource

lib/resources/file.rb
```diff
@@ -109,16 +124,14 @@ def to_s
 
     private
 
-    def file_permission_granted?(flag, by_usergroup, by_specific_user)
-      unless inspec.os.unix?
-        fail 'Checking file permissions is not supported on your os'
-      end
-
+    def file_permission_granted?(access, by_usergroup, by_specific_user)
+      return nil if @perms_provider.nil?
```
chris-rock: we should fail here, once we have the skip as the first statement of eg. in `readable?` this method should never be called if perms_provider is not defined.

lib/resources/file.rb
```diff
@@ -129,31 +142,63 @@ def check_file_permission_by_mask(usergroup, flag)
...
+  class UnixFilePermissions < FilePermissions
+    def check_file_permission_by_user(user, access_type, path)
+      flag = case access_type
+             when 'read'
+               'r'
+             when 'write'
+               'w'
+             when 'execute'
+               'x'
+             end
```
chris-rock, line 177: we should return an error if the access_type is unknown

chris-rock:
Awesome work @jeremymv2 I had some minor comments. Nothing difficult to improve. On top, I would prefer if we could add some integration tests as well in https://github.com/chef/inspec/blob/master/test/integration/default/file_spec.rb#L173-L200

lib/resources/file.rb
```diff
@@ -51,20 +65,23 @@ def contain(*_)
 
     def readable?(by_usergroup, by_specific_user)
       return false unless exist?
+      return skip_resource '`readable?` is not supported on your OS yet.' if @perms_provider.nil?
```
chris-rock, line 69: 👍

chris-rock:
Awesome work @jeremymv2, lets rebase, fix the tests and I'll merge it. Let me know if I can help with fixing the tests.

jeremymv2:
@chris-rock unit and integration tests added to the PR.
I heavily refactored the unit tests - let me know what you think.

chris-rock:
Awesome @jeremymv2

PR #1220, opened 2016-10-14
extend the attributes object with helper methods

chris-rock:
- add helper methods
- allow the attribute object to generate ruby code

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -21,11 +21,32 @@ module Inspec
       @opts[:default]
     end
 
+    def title
+      @opts[:title]
+    end
+
+    def description
+      @opts[:description]
+    end
+
+    def ruby_var_identifier
+      'attr_' + @name.downcase.strip.gsub(/\s+/, '-').gsub(/[^\w-]/, '')
+    end
+
     def to_hash
       {
         name: @name,
         options: @opts,
       }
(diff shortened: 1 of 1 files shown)


lib/inspec/objects/attribute.rb
```diff
@@ -21,13 +21,34 @@ def default
...
+    def title
+      @opts[:title]
+    end
+
+    def description
+      @opts[:description]
+    end
+
+    def ruby_var_identifier
+      'attr_' + @name.downcase.strip.tr(' ', '-').gsub(/[^\w-]/, '')
```
alexpop: PR looks good. What you think of updating this slug maker from:

``` ruby
[10] pry> " hello\tdear  spaces\n".downcase.strip.tr(' ', '-').gsub(/[^\w-]/, '')
=> "hellodear--spaces"
```

To:

``` ruby
[11] pry> " hello\tdear  spaces\n".downcase.strip.gsub(/\s+/, '-').gsub(/[^\w-]/, '')
=> "hello-dear-spaces"
```

chris-rock: good point, lets use your approach

alexpop:
Looks good, merging. Thanks Chris!

aaronlippold:
Should this also have the `required` boolean as well as part of the object?

PR #1221, opened 2016-10-15
update train to fix empty target URIs

arlimus:
See train's 0.20.1 release https://github.com/chef/train/blob/master/CHANGELOG.md#0201-2016-10-15


chris-rock:
+1

Issue #966, opened 2016-08-23
Signed compliance profiles


arlimus:
### Description

Support profile signing so that users are certain that the profiles they use and share are not manipulated and produce the expected results. Especially important for all organization putting high trust in this type of compliance automation, anyone working in the federal space, and inheritance from existing profiles.

chris-rock:
As a reference, we should take a look how its done for container images: https://coreos.com/rkt/docs/latest/signing-and-verification-guide.html

chris-rock:
The habitat format is also looking great:
https://github.com/habitat-sh/habitat/blob/master/components/core/src/crypto/mod.rs#L123-L190

chris-rock:
Very good comparison for JSON signing https://medium.facilelogin.com/json-message-signing-alternatives-897f90d411c#.rkh1114mz

kekaichinose:
__What’s happening? Why was this issue closed?__
This issue was closed due to some much needed review of legacy issues or issues that were spawned in older versions of InSpec, i.e. < v3.

__Why do I care?__
You would care about this if this was an issue that you are still seeing and/or feel needs to be addressed in the current version of InSpec.

__What do I need to do?__
If this issue is no longer important, no further action is necessary. However, if you think this is something that should be addressed, please open a new issue and refer to the original issue in the description.

PR #1228, opened 2016-10-18
Implements profile signing and verification [Experimental]

bookshelfdave:
This PR is a prototype/work in progress of profile signing and verification.
It's most definitely NOT complete and has many TODOs and ideas that need to be worked out.

Demo:
- 0) Generate a pair of 2048 bit RSA keys to use to sign and verify an artifact:
  - `bundle exec inspec artifact generate --keyname foo`
- 1) Sign an existing profile:
  - `bundle exec inspec artifact sign-profile --profile ./examples/profile-attribute/ --keyname foo
    `
- 2) Verify the `.iaf` file:
  - `bundle exec inspec artifact verify --infile profile-attribute-0.1.0.iaf --outfile bar.tar.gz
    `
## Notes:

Implementation notes here: https://gist.github.com/metadave/a11745db15fc1bc33a44719c5e3a026e
### Installing artifacts

  The current implementation allows for a .iaf file to be extracted to a .tar.gz
  file with verification. We could extend the command to install a profile
  to a given location w/ verification. (via lib/inspec/file_provider.rb)
### Generate keys

  The initial implementation uses 2048 bit RSA key pairs (public + private).
  Public keys must be available for a customer to install and verify an artifact.
  Private keys should be stored in a secure location and NOT be distributed.
    (They're only for creating artifacts).
### .IAF file format

  .iaf = "Inspec Artifact File", easy to rename if you'd like something more appropriate.
  The iaf file wraps a binary artifact with some metadata. The first implementation
  looks like this:

```
INSPEC-PROFILE-1
name_of_signing_key
algorithm
signature
<empty line>
binary-blob
<eof>
```

Let's look at each line:

`INSPEC-PROFILE-1`:

  This is the artifact version descriptor. It should't change unless the
  format of the archive changes.

`name_of_signing_key`

  The name of the public key that can be used to verify an artifact

`algorithm`

  The digest used to sign, I picked SHA512 to start with.
  If we support multiple digests, we'll need to have the verify() method
  support each digest.

`signature`

  The result of passing the binary artifact through the digest algorithm above.
  Result is base64 encoded.

`<empty line>`

  We use an empty line to separate artifact header from artifact body (binary blob).
  The artifact body can be anything you like.

`binary-blob`

  A binary blob, most likely a .tar.gz or tar.xz file. We'll need to pick one and
  stick with it as part of the "INSPEC-1" artifact version. If we change block
  format, the artifact version descriptor must be incremented, and the sign()
  and verify() methods must be updated to support a newer version.
### Key revocation

  This implementation doesn't support key revocation. However, a customer
  can remove the public cert file before installation, and artifacts will then
  fail verification.
### Key locations

  This implementation uses the current working directory to find public and
  private keys. We should establish a common key directory (similar to /hab/cache/keys
  or ~/.hab/cache/keys in Habitat).
### Extracting artifacts outside of Inspec

  As in Habitat, the artifact format for Inspec allows the use of common
  Unix tools to read the header and body of an artifact.
To extract the header from a .iaf:
       `sed '/^$/q' foo.iaf`
To extract the raw content from a .iaf:
       `sed '1,/^$/d' foo.iaf`

---
- Ruby OpenSSL docs
  - http://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL.html
- The IAF file format is heavily influenced by the Habitat .hart file format:
  - https://github.com/chef/inspec/blob/725e0a1bc50008c8ad252bd8d5c833a7af20c606/lib/inspec/file_provider.rb#L122-L135

Here's a mini-demo of this feature:

1__tmux_and_cli_rb______oc_inspec_lib_bundles_inspec-artifact__-_vim

NOTE: this screenshot is out of date, however the general idea is still the same


chris-rock:
cc @aaronlippold @mhedgpeth

chris-rock:
implements #966

chris-rock:
@metadave Awesome work. I was thinking about https://github.com/chef/inspec/issues/967 as well. Do we use the same file extensions for profiles and reports? Or we use specific extensions like: InSpec Signed Profile `.isp` and InSpec Signed Report `.inr`

I like the current scope, but lets the feature mark as experimental. Distribution of valid keys is not part of this PR.

bookshelfdave:
@chris-rock I would suggest keeping signed reports and profiles separate, I like your suggestion: `.isp` and `.isr`.

This might change the headers to:

```
INSPEC-PROFILE-1
name_of_signing_key
algorithm
signature
<empty line>
binary-blob
<eof>
```

```
INSPEC-REPORT-1
name_of_signing_key
algorithm
signature
<empty line>
binary-blob
<eof>
```

aaronlippold:
👍

bookshelfdave:
~~I still have to write the profile install bits.~~

chris-rock:
Awesome work @metadave This is a great step forward to secure the execution and reporting of compliance profiles

PR #1231, opened 2016-10-20
Headers and list elements that include more than one `_` character we…

nathenharvey:
Headers and list elements that include more than one `_` character were being rendered incorrectly.  (Perhaps it is only when there are an even number but in any case I only found words with either one or two `_`s.)

The html rendered from markdown would look like:

`<h3>be<em>grouped</em>into</h3>`

When there is only one `_` the html rendered correctly:

`<h3>be_directory</h3>`

This change escapes the `_` when there are multiple.
- New markdown:  `### be\_grouped\_into`
- New html:      `<h3>be_grouped_into</h3>`

resource matchers fixed:
- be_block_device
- be_character_device
- be_grouped_into
- be_linked_to
- be_owned_by
- have_app_pool
- have_property_value

resource fixed:
- parse_config_file

Signed-off-by: Nathen Harvey nharvey@chef.io

### DIFF
--- a/docs/resources/file.md.erb
+++ b/docs/resources/file.md.erb
@@ -29,13 +29,13 @@ This InSpec audit resource has the following matchers:
 
 <%= partial "/shared/matcher_be" %>
 
-### be_block_device
+### be\_block\_device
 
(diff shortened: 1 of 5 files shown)


chris-rock:
Thanks @nathenharvey for this fix!

PR #1233, opened 2016-10-20
Fix broken link in README.md

swalberg:
Link pointed to the right doc but the wrong extension.

Obvious fix.

Signed-off-by: Sean Walberg sean@ertw.com


chris-rock:
Thanks @swalberg for fixing

PR #1242, opened 2016-10-25
Accessing nested mappings in a yam file

chriswessells:
I spent hours following the example in the documentation.  When trying to access the nested mappings I received the error "unknown method <name of nested key>".  When I read the code from lib/utils/object_traversal.rb I found where keys.shift was used to traverse the object.  So I changed from the dotted notation to an array and voila I accessed the nested mapping and could easily verify my tests.  I use rubocop for linting so it complained about the [] notation.  I updated it to the %w() notation and now pass rubocop linting and my tests were failing correctly but now passing accurately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ end
 
 ```ruby
 describe yaml('.kitchen.yml') do
-  its('driver.name') { should eq('vagrant') }
+  its(%w(driver name)) { should eq('vagrant') }
 end
 ```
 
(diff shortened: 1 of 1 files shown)


chris-rock:
This is related to https://github.com/chef/inspec/issues/875. Thanks for adapting the readme until we've fixed the its implementation.

chris-rock:
Please sign-off your work as described in https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

chris-rock:
Thanks @chriswessells for this docs improvement!

PR #1243, opened 2016-10-25
Fix issue with registry_key example

seththoenen:
There are two issues with this `registry_key` resource example.

1) The hive in the example isn't wrapped in single-quotes. Ruby attempts to evaluate this as an object when it needs to be string data. Surrounding `HKEY_USERS` with single-quotes resolves this issue.

2) The `.each` method expects the beginning of the provided array to be on the same line. Moving `{ |key|` to directly follow `.each` allows the method to properly execute.

### DIFF
--- a/docs/resources/registry_key.md.erb
+++ b/docs/resources/registry_key.md.erb
@@ -73,9 +73,8 @@ For example, to get all child items for a registry key:
 The following example shows how find a property that may exist against multiple registry keys, and then test that property for every registry key in which that property is located:
 
     describe registry_key({
-      hive: HKEY_USERS
-    }).children(/^S-1-5-21-[0-9]+-[0-9]+-[0-9]+-[0-9]{3,}\\Software\\Policies\\Microsoft\\Windows\\Installer/).each
-      { |key|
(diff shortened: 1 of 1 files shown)


chris-rock:
@seththoenen Awesome thanks a lot! Just run `git commit --amend -s` to sign-off your work. We recently switched from signed CLAs to simple git commit signoffs. See https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

Also great work on the InSpec tutorial http://seththoenen.com/inspec/ Would you like to add this to https://github.com/chef/inspec/blob/master/www/source/tutorials.html.slim?

seththoenen:
@chris-rock, I've signed-off on this commit. Please let me know if you need anything else from me.

I would be most honored if you would like to add the InSpec tutorials to the official tutorials page. Right now they are designed for people who don't have any experience with InSpec and are using Windows. I'm planning to expand these to more advanced InSpec topics in the future.

Thanks!

Seth

chris-rock:
Thanks @seththoenen for the fix for the example! Please add another PR for the tutorial.

PR #1245, opened 2016-10-25
Ignore RVM files

artem-sidorenko:
For the case somebody is using RVM for inspec development
with different environments (rubies/gemsets)


chris-rock:
Thanks @artem-sidorenko

PR #1247, opened 2016-10-25
Code-block directive is not needed here

artem-sidorenko:
and breaks the layout: http://inspec.io/docs/reference/resources/file/


chris-rock:
Nice catch @artem-sidorenko

PR #1248, opened 2016-10-25
Use include instead of match in the error message

artem-sidorenko:
as include behaves like contain (exact matching instead of regexp)

[Related comment](https://github.com/chef/inspec/pull/214#issuecomment-154132333)

Are they any reasons why contain is not mapped to include with a warning? 
It might simplify the migration from serverspec, where deprecated contain matcher is used quite often

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -219,7 +219,7 @@ end
 # unsupported
 RSpec::Matchers.define :contain do |_rule|
   match do |_resource|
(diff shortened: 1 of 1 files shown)


chris-rock:
@artem-sidorenko I think it makes sense to map it to `include` but still keep the deprecation warning.

artem-sidorenko:
@chris-rock Ok. I will have a look to rspec and how to do that, I'm not familiar with rspec on this level yet.

Maybe it makes sense to merge this and I will provide another PR with mapping to `include`

chris-rock:
@artem-sidorenko Yes it does make sense. Thanks @artem-sidorenko

PR #1254, opened 2016-10-26
Supermarket tools get and filter by tool_type

alexpop:
With InSpec in master:

``` bash
$ inspec supermarket profiles | grep '^ * ' | wc -l
       8
```

With the change in this PR:

``` bash
$ bundle exec inspec supermarket profiles | grep '^ * ' | wc -l
      11
```

It fixes the [profiles list issue](https://github.com/chef/inspec/issues/1219) by querying the most recent 100(out of 126 at the moment) tools and filtering by tool_type. The API returns max 100 items. 
The oldest compliance profile is number 25 on that list. This gives us a few months until the supermarket search API will have a type filter.


chris-rock:
Thanks @alexpop for this intermediate fix!

PR #1256, opened 2016-10-26
Fix nil timeout and retries

alexpop:
`@timeout` and `@retries` were always `nil` when calling `SSLShake.hello`

Bumped into this while troubleshooting: https://github.com/chef/inspec/issues/1091


chris-rock:
Thanks @alexpop for this fix!

PR #1262, opened 2016-10-28
Use Slack Badge instead of Gitter badge

chris-rock:


alexpop:
Looks good, thanks Chris!

PR #1271, opened 2016-11-02
Enable the json resource to accept command output or JSON content

jerearista:


lib/resources/json.rb
```diff
@@ -21,19 +30,28 @@ class JsonConfig < Inspec.resource(1)
 
     def initialize(path)
```
chris-rock: should we rename `path` to `opts` then? Its only a path if used as a string then

chris-rock:
Great improvement @jerearista

lib/resources/json.rb
jerearista: Agreed.   Changed.

chris-rock:
Thanks @jerearista

Issue #1269, opened 2016-11-01
Windows Service State incorrectly interpretted for use in describe service. 


hj70ft:
### Description

When running an InSpec test against a Windows service and attempting to utilize the be_enabled option the code as written will only consider 'Auto' as a valid enabled state for a service, but this is incorrect as 'Manual' is also a valid state for a Windows server to be considered enabled. 

The code effectively treats the enabled state as a binary with Auto being enabled and Manual, Disabled as the disabled states. Manual should be considered a valid enabled state for a service instead of a valid disabled state. 

### InSpec and Platform Version

Using InSpec version 0.26.0 on a Windows 7 laptop and running code against either the laptop or a Windows 2012 vagrant VM. 

### Replication Case

describe service test case for any service that you can manually change the state from Automatic, Manual, Disabled. 

Use a test in the case for should 'be_enabled' and you will see that...
Automatic = passed
Manual = failed
Disabled = failed

Tweaking the service state between each execution you should see that Manual is treated identical to the disabled state instead of the automatic state. 

### Possible Solutions

Include code at line 616 that tests for either Auto or Manual as you have already exposed the state with WMI and just need to include both cases as valid. 

### Stacktrace

N/A

chris-rock:
@hj70ft Thanks for highlighting this issue. A PR is very welcome to fix it!

hj70ft:
Working on the details now. Code fix checks out on my end and seems to be working after tests. I'll get a PR done as soon as I figure out how to do all this stuff for the first time. I'm a bit of a noob when it comes to github.

chris-rock:
@hj70ft No rush, let us know if we can help

chris-rock:
Solved with #1274

PR #1274, opened 2016-11-03
Windows Service Check for service enabled state. 

hj70ft:
Updated Windows Service check for Enabled state to allow both Auto and Manual to be considered valid states for a service to be enabled.

Reference Issue #1269


chris-rock:
Thanks for this improvement @hj70ft

PR #1276, opened 2016-11-03
Contain matcher maps to include matcher with warning

artem-sidorenko:
to allow easier migration from serverspec, where contain
matcher is used often

Signed-off-by: Artem Sidorenko <artem@posteo.de>


artem-sidorenko:
As agreed in https://github.com/chef/inspec/pull/1248

chris-rock:
Awesome, thanks @artem-sidorenko

Issue #1283, opened 2016-11-07
Vendor dependent profiles in archive


chris-rock:
### Description

Lets assume we have a meta-profile like https://github.com/chris-rock/acme-inspec-profile

```
acme-inspec-profile
├── LICENSE
├── README.md
├── controls
│   ├── hardening.rb
│   ├── patch.rb
│   └── ssl.rb
└── inspec.yml
```

That depends on a couple of other profiles:

```
name: acme-inspec-profile
...
depends:
  - name: linux-patch-benchmark
    git: https://github.com/dev-sec/linux-patch-benchmark.git
  - name: windows-patch-benchmark
    git: https://github.com/dev-sec/windows-patch-benchmark.git
  - name: os-hardening
    git: https://github.com/dev-sec/tests-os-hardening.git
  - name: ssh-hardening
    git: https://github.com/dev-sec/tests-ssh-hardening.git
  - name: ssl-benchmark
    git: https://github.com/dev-sec/ssl-benchmark.git
```

Now we can use InSpec to archive the profile:

```
inspec archive /Users/chartmann/Development/Demo/InSpec-1.0-Webinar/acme-inspec-profile
I, [2016-11-07T11:12:42.366206 #52646]  INFO -- : Checking profile in /Users/chartmann/Development/Demo/InSpec-1.0-Webinar/acme-inspec-profile
I, [2016-11-07T11:12:42.366280 #52646]  INFO -- : Metadata OK.
`command(ssh).exist?` is not suported on your OS: 
I, [2016-11-07T11:12:43.383553 #52646]  INFO -- : Found 120 controls.
W, [2016-11-07T11:12:43.383748 #52646]  WARN -- : Control verify-kb has no description
W, [2016-11-07T11:12:43.383777 #52646]  WARN -- : Control important-count has no description
W, [2016-11-07T11:12:43.383789 #52646]  WARN -- : Control important-patches has no description
W, [2016-11-07T11:12:43.383799 #52646]  WARN -- : Control important-patches has no tests defined
W, [2016-11-07T11:12:43.383819 #52646]  WARN -- : Control optional-count has no description
W, [2016-11-07T11:12:43.383829 #52646]  WARN -- : Control optional-patches has no description
W, [2016-11-07T11:12:43.383837 #52646]  WARN -- : Control optional-patches has no tests defined
W, [2016-11-07T11:12:43.383846 #52646]  WARN -- : Control verify-patches has no description
[...]

PR #1291, opened 2016-11-09
copy vendored dependencies into cache

chris-rock:
This implements the second part of #1283. It copies vendored content into the global cache before the profile is executed


arlimus:
Thank you @chris-rock :) :+1:

PR #1297, opened 2016-11-16
enable inspec compliance cli support automate

### LINKED ISSUE #1295: inspec compliance profiles should support automate as a backend
### Description

The inspec subcommand `compliance` should be able to work with the automate backend. We should configure everything during `inspec compliance login`. All other subcommands should then work as expected eg. `inspec compliance profiles` and `inspec exec compliance://namespace/name`

1.x

vjeffrey:
fixes https://github.com/chef/inspec/issues/1295

`b inspec compliance login_automate http://delivery --user admin --dctoken $DC_TOKEN --ent $AUTOMATE_ENT `

`b inspec compliance login_automate http://delivery --user $AUTOMATE_USER --usertoken $AUTOMATE_TOKEN --ent $AUTOMATE_ENT `

`b inspec compliance profiles`

`b inspec compliance upload ~/linux.tar.gz`

`b inspec compliance logout` <-- just destroys config when automate

`b inspec compliance version` <-- returns msg saying that's not available with automate

`b inspec compliance exec admin/linux`

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -8,12 +8,12 @@ require 'uri'
 module Compliance
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
-  class API
+  class API # rubocop:disable Metrics/ClassLength
     # return all compliance profiles available for the user
     def self.profiles(config)
-      url = "#{config['server']}/user/compliance"
+      config['automate'][0] ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/user/compliance"
       # TODO, api should not be dependent on .supported?
-      response = Compliance::HTTP.get(url, config['token'], config['insecure'], !config.supported?(:oidc))
+      response = Compliance::HTTP.get(url, config['token'], config['insecure'], config['user'], !config.supported?(:oidc), config['automate'], config['ent'])
       data = response.body
       response_code = response.code
       case response_code
@@ -21,11 +21,17 @@ module Compliance
         msg = 'success'
         profiles = JSON.parse(data)
         # iterate over profiles
-        mapped_profiles = profiles.map do |owner, ps|
-          ps.keys.map do |name|
-            { org: owner, name: name }
-          end
-        end.flatten
+        if config['automate'][0]
+          mapped_profiles = profiles.map do |owner, ps|
+            { org: ps['owner_id'], name: owner }
+          end.flatten
+        else
+          mapped_profiles = profiles.map do |owner, ps|
+            ps.keys.map do |name|
+              { org: owner, name: name }
+            end
+          end.flatten
+        end
         return msg, mapped_profiles
       when '401'
         msg = '401 Unauthorized. Please check your token.'
@@ -69,8 +75,8 @@ Please login using `inspec compliance login https://compliance.test --user admin
 
     def self.upload(config, owner, profile_name, archive_path)
       # upload the tar to Chef Compliance
-      url = "#{config['server']}/owners/#{owner}/compliance/#{profile_name}/tar"
-      res = Compliance::HTTP.post_file(url, config['token'], archive_path, config['insecure'], !config.supported?(:oidc))
+      config['automate'][0] ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/owners/#{owner}/compliance/#{profile_name}/tar"
+      res = Compliance::HTTP.post_file(url, config['token'], config['user'], archive_path, config['insecure'], !config.supported?(:oidc), config['automate'], config['ent'])
       [res.is_a?(Net::HTTPSuccess), res.body]
     end
 
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -12,8 +12,8 @@ module Compliance
     # return all compliance profiles available for the user
     def self.profiles(config)
       config['server_type'] == 'automate' ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/user/compliance"
-      # TODO, api should not be dependent on .supported?
-      response = Compliance::HTTP.get(url, config['token'], config['insecure'], config['user'], !config.supported?(:oidc), config['automate'], config['server_type'])
+      headers = get_headers(config)
+      response = Compliance::HTTP.get(url, headers, config['insecure'])
       data = response.body
       response_code = response.code
       case response_code
@@ -76,7 +76,8 @@ Please login using `inspec compliance login https://compliance.test --user admin
     def self.upload(config, owner, profile_name, archive_path)
       # upload the tar to Chef Compliance
       config['server_type'] == 'automate' ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/owners/#{owner}/compliance/#{profile_name}/tar"
-      res = Compliance::HTTP.post_file(url, config['token'], config['user'], archive_path, config['insecure'], !config.supported?(:oidc), config['automate'], config['server_type'])
+      headers = get_headers(config)
+      res = Compliance::HTTP.post_file(url, headers, archive_path, config['insecure'])
       [res.is_a?(Net::HTTPSuccess), res.body]
     end
 
@@ -126,6 +127,21 @@ Please login using `inspec compliance login https://compliance.test --user admin
       end
 
       [success, msg, access_token]
+    end
+
+    def self.get_headers(config)
+      if config['server_type'] == 'automate'
+        headers = { 'chef-delivery-enterprise' => config['automate']['ent'] }
+        if config['automate']['token_type'] == 'dctoken'
+          headers['x-data-collector-token'] = config['token']
+        else
+          headers['chef-delivery-user'] = config['user']
+          headers['chef-delivery-token'] = config['token']
+        end
+      else
+        headers = { 'Authorization' => "Bearer #{config['token']}" }
+      end
+      headers
     end
   end
 end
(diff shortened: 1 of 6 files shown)


vjeffrey:
@alexpop could you test this out please?

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -56,6 +57,32 @@ def login(server) # rubocop:disable Metrics/AbcSize
       puts '', msg
     end
 
+    desc "automate SERVER --user='USER' --dctoken='DATA_COLLECTOR_TOKEN' or --usertoken='AUTOMATE_TOKEN' --ent='ENT'", 'Log in to an Automate SERVER'
```
chris-rock: should we call it `login-automate`?

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -79,10 +106,8 @@ def profiles
     def exec(*tests)
       config = Compliance::Configuration.new
       return if !loggedin(config)
-
       # iterate over tests and add compliance scheme
-      tests = tests.map { |t| 'compliance://' + t }
-
+      config['automate'][0] ? tests = tests.map { |t| 'automate://' + t } : tests = tests.map { |t| 'compliance://' + t }
```
chris-rock: I would prefer if we could stick to `compliance://` no matter if the profiles are stored in Chef Compliance or Chef Automate

lib/bundles/inspec-compliance/target.rb
```diff
@@ -13,10 +13,11 @@ module Compliance
   class Fetcher < Fetchers::Url
     name 'compliance'
     priority 500
-
-    def self.resolve(target)
+    def self.resolve(target) # rubocop:disable PerceivedComplexity, Metrics/CyclomaticComplexity
       uri = if target.is_a?(String) && URI(target).scheme == 'compliance'
               URI(target)
+            elsif target.is_a?(String) && URI(target).scheme == 'automate'
```
chris-rock: I think we do not need the automate scheme

lib/bundles/inspec-compliance/cli.rb
chris-rock: I am not 100% sure on this. What would be a better way to name it? Another option could be to just stick with `inspec compliance login` and just offer more options there. On the other hand this may become more complex for a user to understand which config is for which server? What are you thinking?

chris-rock (commented):
Great work @vjeffrey

lib/bundles/inspec-compliance/cli.rb
vjeffrey: I had started out with just making it an extension of inspec compliance login, but then i couldn't display the desc well with the different options...which is why I went this route.  I'm up for a rename to login-automate, i think that makes sense.

lib/bundles/inspec-compliance/target.rb
vjeffrey: oh ya, you're right - i'll change that

vjeffrey:
conversation with hannah:

inspec compliance

`inspec compliance login > automate http://delivery --user=admin --ent=cd`

`Please specify a token using --dctoken='DATA_COLLECTOR_TOKEN' or usertoken='AUTOMATE_TOKEN'`

`--dctoken='fnjknefkje'`

hannah-radish:
@vjeffrey Does this make sense to you?
automate-login-cli
automate-login-cli2

vjeffrey:
thanks hannah!!!!!
coloring and prompt will be taken care of in separate issues, referenced above (1310 and 1311)

vjeffrey:
i'll update this change to address comments later today/wednesday morning at latest

hannah-radish:
screen shot 2016-11-22 at 4 09 40 pm

Updated the color pallet here as well: https://github.com/chef/inspec/pull/1313/files

vjeffrey:
updated to name it login_automate and removed the unnecessary automate scheme logic

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -169,24 +196,27 @@ def upload(path) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Percei
...
     desc 'logout', 'user logout from Chef Compliance'
     def logout
       config = Compliance::Configuration.new
-      unless config.supported?(:oidc) || config['token'].nil?
+      unless config.supported?(:oidc) || config['token'].nil? || config['automate']
         config = Compliance::Configuration.new
         url = "#{config['server']}/logout"
         Compliance::API.post(url, config['token'], config['insecure'], !config.supported?(:oidc))
       end
-
```
chris-rock, line 189: I think we should clear the token information of the config

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -198,6 +228,29 @@ def logout
 
     private
 
+    def login_automate_config(url, user, dctoken, usertoken, ent)
+      config = Compliance::Configuration.new
+      config['server'] = url
```
chris-rock: we should have a `config['server_type']` to indicate the type of server. At the moment that is `compliance` and `automate`

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -198,6 +228,29 @@ def logout
...
+      config['ent'] = ent
```
chris-rock: We should store all automate specific config under the section `config['automate']`

lib/bundles/inspec-compliance/api.rb
```diff
@@ -8,24 +8,30 @@
 module Compliance
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
-  class API
+  class API # rubocop:disable Metrics/ClassLength
     # return all compliance profiles available for the user
     def self.profiles(config)
-      url = "#{config['server']}/user/compliance"
+      config['automate'][0] ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/user/compliance"
```
chris-rock: I am not fan of `config['automate'][0]`. We should ask directly for the information we need: `config['automate']['ent']`. For this case I propose config['server_type']

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -56,6 +57,33 @@ def login(server) # rubocop:disable Metrics/AbcSize
       puts '', msg
     end
 
+    desc "login_automate SERVER --user='USER' --ent='ENT' --dctoken or --usertoken='TOKEN'", 'Log in to an Automate SERVER'
```
chris-rock, line 60: I like the `login_automate` approach

lib/bundles/inspec-compliance/target.rb
```diff
@@ -26,14 +25,21 @@ def self.resolve(target)
       # check if we have a compliance token
       config = Compliance::Configuration.new
       if config['token'].nil?
+        if config['automate'][0]
+          server = 'automate'
+          msg = 'inspec compliance automate https://your_automate_server --user USER --ent ENT --dctoken DCTOKEN or --usertoken USERTOKEN'
```
chris-rock: This should be `inspec compliance login_automate ...`

lib/bundles/inspec-compliance/target.rb
```diff
@@ -26,14 +25,21 @@ def self.resolve(target)
       # check if we have a compliance token
       config = Compliance::Configuration.new
       if config['token'].nil?
+        if config['automate'][0]
+          server = 'automate'
```
chris-rock, line 29: Lets use this in `config['server_type']`

lib/fetchers/url.rb
```diff
@@ -126,7 +126,19 @@ def download_archive_to_temp
       Inspec::Log.debug("Fetching URL: #{@target}")
       http_opts = {}
       http_opts['ssl_verify_mode'.to_sym] = OpenSSL::SSL::VERIFY_NONE if @insecure
-      http_opts['Authorization'] = "Bearer #{@token}" if @token
+      if @config
```
chris-rock, line 129: If we have the `server_type`, lets do a simple if then clause

lib/bundles/inspec-compliance/cli.rb
chris-rock: Also we should remove the the server information from the config. This was missing even before

vjeffrey: :+1:

lib/fetchers/url.rb
vjeffrey: :+1:

lib/bundles/inspec-compliance/api.rb
```diff
@@ -8,24 +8,30 @@
...
       when '200'
         msg = 'success'
         profiles = JSON.parse(data)
         # iterate over profiles
-        mapped_profiles = profiles.map do |owner, ps|
-          ps.keys.map do |name|
-            { org: owner, name: name }
-          end
-        end.flatten
+        if config['server_type'] == 'automate'
```
chris-rock, line 24: as discussed with @vj we talk with @alexpop to find a way to harmonize this part

lib/bundles/inspec-compliance/http.rb
```diff
@@ -9,13 +9,21 @@ module Compliance
   # implements a simple http abstraction on top of Net::HTTP
   class HTTP
     # generic get requires
-    def self.get(url, token, insecure, basic_auth = false)
+    def self.get(url, token, insecure, user, basic_auth = false, automate = nil, server_type) # rubocop:disable Metrics/ParameterLists
```
chris-rock: As discussed with @vjeffrey we think we can abstract that even further:

`def self.get(url, headers, insecure)`. This would allow us to pass in the required headers from outside:
```
headers = {
  'x-data-collector-token': token
}

headers = {
  'chef-delivery-user': user,
  'chef-delivery-token': token,
}

headers = {
  'Authorization', "Bearer #{token}",
}
```

vjeffrey:
@chris-rock pushed up a commit with updated headers stuffs

chris-rock:
Awesome. Thank you @vjeffrey

PR #1303, opened 2016-11-20
Fixes resources in the docs

burtlo:
* Fixes the npm package example to state 'npm' vs 'gem'
* Fixes powershell resource to specify the resource instead of 'script'
* Updates the `vbscript` example to rename script variable to 'script'

> While ruby would allow the local variable and the presence of the InSpec method at the same time I think that it is bad form. Other resource examples also use 'script'.

* Changes pip to show generic example

> Other package like resources show a generic example in the default.


chris-rock:
@burtlo Awesome. Thank you very much. This includes great improvements

PR #1305, opened 2016-11-20
ensure metadata release entry is a string

chris-rock:
As reported by Shone Jacob in https://chefcommunity.slack.com/archives/inspec/p1479483621000703 InSpec is not handling the version numbers from Ubuntu properly.

```
$ inspec exec ...
This OS/platform (ubuntu) is not supported by this profile.
This is the ubuntu version:
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.1 LTS
Release:    16.04
Codename:   xenial
```

inspec.yml:

```
supports:
 - os-name: ubuntu
   release: 16.04
```

Internally, YAML parses `16.04` as a number instead as string. This PR fixes that behavior and ensurs that versions always compared as strings.

### DIFF
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -68,6 +68,7 @@ module Inspec
                     os.method(family_check).call
                   )
 
+      # ensure we do have a string if we have a non-nil value eg. 16.06
       release_ok = release.nil? || os[:release] == release
 
       # we want to make sure that all matchers are true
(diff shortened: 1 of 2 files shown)


lib/inspec/metadata.rb
```diff
@@ -68,7 +68,8 @@ def is_supported?(os, entry)
                     os.method(family_check).call
                   )
 
-      release_ok = release.nil? || os[:release] == release
+      # ensure we do have a string if we have a non-nil value eg. 16.06
+      release_ok = release.nil? || os[:release] == release.to_s
```
arlimus: Could we move this to `finalize` ?

arlimus: Sorry, `finalize_supports` my bad!

arlimus:
Thank you @chris-rock :+1: :taco:

Issue #1214, opened 2016-10-11
Resource for windows scheduled task


modille:
A resource for `windows_scheduled_task` would be helpful.

Even just having an `exists` matcher, similar to what ServerSpec has:

https://github.com/mizzy/serverspec/blob/master/lib/serverspec/type/windows_scheduled_task.rb#L3

The Powershell query used there is

``` powershell
(schtasks /query /v /fo csv /TN 'foo' | ConvertFrom-CSV).TaskName -eq '\foo'
```

See:
- https://github.com/mizzy/specinfra/blob/c8a97ef154684452ba059beee4cddbc8e62258c2/lib/specinfra/command/windows/base/scheduled_task.rb#L6
- https://github.com/mizzy/specinfra/blob/c8a97ef154684452ba059beee4cddbc8e62258c2/lib/specinfra/backend/powershell/support/find_scheduled_task.ps1#L5

chris-rock:
@modille Great idea. Are you going to write a PR for that?

modille:
@chris-rock unfortunately I don't have time at the moment, so I just used the command resource as a workaround

username-is-already-taken2:
Hello @modille My name is Gary

My colleague and I are learning ruby and your suggestion might be something we could get our teeth stuck into, can't promise it would get commited into core but we might be able to get something running that you could import into a compliance profile as a library and use.

Could you give me your thoughts on the spec.

We may have to get a stear later with resource namespace but thats easy to change
We would look to invoke it like this? 

`windows_scheduled_task(<taskname>)`

and look to have at least these matches

`it { should exist }`
where we will look for (case-insensitive) match of the taskname? 
`it { should be_enabled }`
where we  will look to match status to "Ready" or "Running"?
`it { should be_disabled }`
where we will look to match status to "Disabled"
`it { should be_scheduled }`
where we will look to match the "Next Run Time" to not match "N/A"

I think that should get us something we could use?

Let me know what you think,

modille:
Thanks for being willing to contribute this, @username-is-already-taken2.

I'm not entirely sure about the criteria around enabled/disabled/scheduled -- for example, I think other possible states are Queued and Unknown, so you'd have to account for those.

Otherwise, those matchers would be very helpful in my opinion.

username-is-already-taken2:
Hi 

Together @cdbeard2016 and I have made some good progress with this and we have something ready for testing

changed a couple of things, the resource is now `windows_task` (much quicker to type :)
[...]

PR #1306, opened 2016-11-20
Adding windows_task resource

username-is-already-taken2:
Hi 

Both Chris ( @cdbeard2016 ) and myself would like to put forward this PR to add the `windows_task` resource to inspec

We got the idea to create it from the issue #1214 

We tried to meet all the contribution requirements, let us know if we fell short anywhere

Best Regards

Gary


chris-rock:
Awesome work @username-is-already-taken2 and @cdbeard2016

chris-rock:
Integration tests work smoothly and the unit test are great.

PR #1307, opened 2016-11-21
also push docker latest tag with each release

chris-rock:
This PR ensures that all new pushed inspec docker images, also get tagged as latests


alexpop:
👍

PR #1321, opened 2016-11-24
inspec check and json to use vendored dependencies

alexpop:
Fixes: https://github.com/chef/inspec/issues/1316 by providing a cache directory. Otherwise `copy_deps_into_cache` is not called.

### DIFF
--- a/test/functional/inspec_vendor_test.rb
+++ b/test/functional/inspec_vendor_test.rb
@@ -21,6 +21,10 @@ describe 'example inheritance profile' do
   end
 
   it 'can vendor profile dependencies from the profile path' do
+    # clean existing vendor directory
+    FileUtils.rm_r ("#{inheritance_path}/vendor")
+
+    # vendor all dependencies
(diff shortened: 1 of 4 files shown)


test/functional/inspec_vendor_test.rb
```diff
@@ -53,10 +61,51 @@
     out.stderr.must_equal ''
   end
 
+  it 'ensure json command is not fetching remote profiles if vendored' do
```
alexpop, line 64: A good test for this would be to check that the vendored profile(ex: 01414bd307ea2f7d4dc8cd141085ba7ad61d4c3b2606d57b2dae987c1c3954cb.tar.gz was copied in the cache) instead of a remote one

alexpop: As in length is 1 and that profile is `01414bd307ea2f7d4dc8cd141085ba7ad61d4c3b2606d57b2dae987c1c3954cb.tar.gz`

chris-rock:
Awesome. Thank you @alexpop

PR #1323, opened 2016-11-24
Provide inspec.lock for archives as well

alexpop:
Fixes https://github.com/chef/inspec/issues/1322 by providing the loading the inspec.lock from both unpacked and archived profiles.


chris-rock (approved):
So cool. Thank you @alexpop

PR #1328, opened 2016-11-28
fix docker release script

chris-rock:
With #1307 I introduced a bug into the docker release script. This PR fixes that issue.


vjeffrey:
👍

PR #1329, opened 2016-11-28
Show process name during inspec output

jcastillocano:
Testing a _process_ resource I got this output:

```
  Processes list.length
     ✔  should eq 1
```

when I expected something like _Processes mysqld_. Using _grep_ variable to verify the output:

```
  Processes mysqld
     ✔  users should cmp == "mysql"
```

or, when we use RegExp:

```
  Processes /mysqld_safe/
     ✔  users should cmp == "root"
```

I also added some unit tests to verify this behavior.

### DIFF
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -21,9 +21,10 @@ module Inspec::Resources
                 :states
 
     def initialize(grep)
+      @grep = grep
       # turn into a regexp if it isn't one yet
       if grep.class == String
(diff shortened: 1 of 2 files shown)


chris-rock:
@jcastillocano great addition. Could you squash your commits and rebase it to master? I propose you fetch latest master and call `git rebase -i master` from your branch. Let me know if you need any help.

chris-rock:
Thank you @jcastillocano for this contribution!

PR #1315, opened 2016-11-23
WIP: Vj/clean up formatter

burtlo:
These are my changes on top of the changes you are proposing @vjeffrey.

THIS IS A WORK-IN-PROGRESS. Still committing work.

PR #1331, opened 2016-11-29
burtlo/cleaner cli formatter

burtlo:
Updates RSpec CLI Formater to print profiles correctly

The profiles will display  the controls with their results and then display the examples not associated with any control but within the profile.

```
Profile: InSpec Profile (bananas_in_pajamas)
Version: 0.1.0
Target:  local://

  ✖  ssh (b in p): A human-readable title (1 failed)
     ✔  File sandia/bananas_in_pajamas should exist
     ✖  System Package ssh should be installed
     expected that `System Package ssh` is installed
  ✔  Not Running a Web Server (b in p): A human-readable title
     ✔  File /tmp should be directory
     ✔  System Package httpd should not be installed
     ✔  Service httpd should not be running
  ✔  sftp (b in p): A human-readable title
     ✔  System Package sftp should not be installed

  File file_in_profile_between_controls
     ✖  should exist
     expected File file_in_profile_between_controls to exist
     ✔  should not be owned by "root"
  File file_in_profile_outside_of_control
     ✔  should not exist

Profile: Broken Arrow Profile (broken_arrow)
Version: 1.0.0
Target:  local://

  ✖  CHEF FILES (broken_arrow): Chef files are important to have in place (expected File broken_arrow to exist)
     ✖  File broken_arrow should exist
     expected File broken_arrow to exist

  Service broken_arrow
     ✔  should not be running

Profile: InSpec Profile (my_blank)
Version: 0.1.0
Target:  local://


  File profile_with_a_control_no_examples
     ✔  should not exist

Target:  local://

  ✖  Control (no profile): Hello World (expected nil to match "Hello World")
     ✖  File control_outside.txt content should match "Hello World"
     expected nil to match "Hello World"


Profile Summary: 4 successful, 4 failures, 0 skipped
Test Summary: 17 successful, 6 failures, 0 skipped
```

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -107,35 +107,21 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
 
   attr_reader :profiles
 
-  def profiles_info
-    @profiles_info ||= profiles.map(&:info!).map(&:dup)
-  end
-
   # Called by the runner during example collection.
   def add_profile(profile)
     profiles.push(profile)
   end
 
-  def dump_one_example(example, control)
-    control[:results] ||= []
-    example.delete(:id)
-    example.delete(:profile_id)
-    control[:results].push(example)
-  end
-
   def stop(notification)
     super(notification)
-    examples = @output_hash.delete(:controls)
-    missing = []
 
-    examples.each do |example|
-      control = example2control(example, profiles_info)
-      next missing.push(example) if control.nil?
-      dump_one_example(example, control)
-    end
-
+    @output_hash[:other_checks] = examples_without_controls
     @output_hash[:profiles] = profiles_info
-    @output_hash[:other_checks] = missing
+
+    examples_with_controls.each do |example|
+      control = example2control(example)
+      move_example_into_control(example, control)
+    end
   end
 
   def controls_summary
@@ -203,14 +189,30 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
 
   private
 
-  #
-  # TODO(ssd+vj): We should probably solve this by either ensuring the example has
-  # the profile_id of the top level profile when it is included as a dependency, or
-  # by registrying all dependent profiles with the formatter. The we could remove
-  # this heuristic matching.
-  #
-  def example2profile(example, profiles)
-    profiles.find { |p| profile_contains_example?(p, example) }
+  def examples
+    @examples ||= @output_hash.delete(:controls)
+  end
+
+  def examples_without_controls
+    examples.find_all { |example| example2control(example).nil? }
+  end
+
+  def examples_with_controls
+    (examples - examples_without_controls)
+  end
+
+  def profiles_info
+    @profiles_info ||= profiles.map(&:info!).map(&:dup)
+  end
+
+  def example2control(example)
+    profile = profile_from_example(example)
+    return nil unless profile && profile[:controls]
+    profile[:controls].find { |x| x[:id] == example[:id] }
+  end
+
+  def profile_from_example(example)
+    profiles_info.find { |p| profile_contains_example?(p, example) }
   end
 
   def profile_contains_example?(profile, example)
@@ -226,10 +228,11 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
     end
   end
 
-  def example2control(example, profiles)
-    profile = example2profile(example, profiles)
-    return nil unless profile && profile[:controls]
-    profile[:controls].find { |x| x[:id] == example[:id] }
+  def move_example_into_control(example, control)
+    control[:results] ||= []
+    example.delete(:id)
+    example.delete(:profile_id)
+    control[:results].push(example)
   end
 
   def format_example(example)
@@ -286,7 +289,6 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
   end
 
   def close(_notification) # rubocop:disable Metrics/AbcSize
-    require 'pry' ; binding.pry
     flush_current_control(@current_control)
     output.puts('') unless @current_control.nil?
     print_tests(@anonymous_tests)
@@ -597,12 +599,12 @@ class InspecRspecJUnit < RSpecJUnitFormatter
 
   def format_example(example)
     data = super(example)
-    control = example2control(data, profiles_info) || {}
+    control = example2control(data) || {}
     control[:id] = data[:id]
     control[:profile_id] = data[:profile_id]
 
     data[:status_type] = status_type(data, control)
-    dump_one_example(data, control)
+    move_example_into_control(data, control)
 
   def initialize(*args)
     super(*args)
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -107,7 +107,7 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
 
   # Called by the runner during example collection.
   def add_profile(profile)
-    profiles.push(profile)
+    @profiles.push(profile)
   end
 
   def stop(notification)
@@ -123,8 +123,6 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
   end
 
   private
-
-  attr_reader :profiles
 
   def all_unique_controls
     Array(@all_controls).uniq
@@ -205,7 +203,7 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
   end
 
   def profiles_info
-    @profiles_info ||= profiles.map(&:info!).map(&:dup)
+    @profiles_info ||= @profiles.map(&:info!).map(&:dup)
   end
 
   def example2control(example)
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -589,14 +589,6 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
   class Control
     include Comparable
 
-    def <=>(other)
-      if id == other.id && profile_id == other.profile_id
-        0
-      else
-        -1
-      end
-    end
-
     STATUS_TYPES = {
       'unknown'  => -3,
       'passed'   => -2,
@@ -610,10 +602,7 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
     def initialize(control, profile)
       @control = control
       @profile = profile
-      @summary_status = STATUS_TYPES['unknown']
-      @skips = []
-      @fails = []
-      @passes = []
+      summary_calculation_is_needed
     end
 
     attr_reader :control, :profile
@@ -626,8 +615,23 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
       control[:id]
     end
 
-    #
-    # Adds an example to the control. This example
+    def anonymous?
+      control[:id].to_s.start_with? '(generated from '
+    end
+
+    def profile_id
+      control[:profile_id]
+    end
+
+    def examples
+      control[:results]
+    end
+
+    def summary_indicator
+      calculate_summary! if summary_calculation_needed?
+      STATUS_TYPES.key(@summary_status)
+    end
+
     def add_example(example)
       control[:id] = example[:id]
       control[:profile_id] = example[:profile_id]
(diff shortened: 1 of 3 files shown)


burtlo:
I would like to also break these formatters out into separate files inside a formatters directory. But let's start with this large file and this huge change.

burtlo:
I tried to clean up the entire CLI formatter by addressing what was happening in `#format_example` and `#close`. You will find the logic there is easier to understand as I opted for long method names that describe what I am trying to convey.

From there I wrap the control data into a class I call `Control` which allows me to make it behave a little more sanely throughout the rest of the methods.

This is still pretty tricky code because you have:

* When one example fails for an example you want the control to report as failed which means you can't simply print them right away and instead have to wait until `#close` or looking at a new example for another control
* Anonymous examples (examples not in a control inside of a profile) are collected up and displayed at the end of a profile which means you have to collect them and then display them before switching profiles or at `#close`

lib/inspec/rspec_json_formatter.rb
```diff
@@ -63,7 +63,7 @@ def stop(notification)
   private
 
   def format_example(example)
-    if example.metadata[:description_args].length > 0 && !example.metadata[:skip].nil?
+    if !example.metadata[:description_args].empty? && example.metadata[:skip]
```
burtlo, line 68: The first statement was a suggestion by Rubocop. The second is the same way of expressing the same thing.

lib/inspec/rspec_json_formatter.rb
```diff
@@ -92,7 +92,7 @@ def format_example(example)
 end
 
 class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLength
-  RSpec::Core::Formatters.register self, :start, :stop, :dump_summary
+  RSpec::Core::Formatters.register self, :stop, :dump_summary
```
burtlo, line 98: I refactored the @profiles_info to a memozied helper method so I no longer need to use `#start`.

lib/inspec/rspec_json_formatter.rb
```diff
@@ -278,272 +268,437 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
...
-        color:      test_color,
-        indicator:  @indicators['small'] + indicator,
-        summary:    format_lines(msg, @indicators['empty']),
-        id: nil, profile: nil
-      )
+  #
+  # This class wraps a control hash object to provide a useful inteface for
+  # maintaining the associated profile, ids, results, title, etc.
+  #
+  class Control
```
burtlo: Don't love that I used the name `Control` here but it is one within this context. I wouldn't want to confuse anyone here.

chris-rock:
@burtlo This is changing a lot. Thank you for the hard work. Some comments I have:
- lets rebase to latest master (that would resolve the conflict)
- does this mean #1315 is obsolete now?
- lets add functional tests to reflect that behavior (reuse profiles from https://github.com/chef/inspec/tree/master/test/unit/mock/profiles or add new ones)

burtlo:
@vjeffrey did the initial work during our hack day and then I jumped on board. When I pushed my latest changes on my branch I failed to find that pull request again and get that integrated. So yes this appends more changes on top of her work.

@chris-rock I'll add the tests today and continue to rebase this branch.

chris-rock:
@burtlo Just rebase it on latest master, this enables us to see your individual contributions.

burtlo:
This is the current state of `rake test:functional` on master:
```
130 runs, 275 assertions, 56 failures, 0 errors, 0 skips
```

You may need to disregard my results here. I noticed that a lot of the tests are failing because I have some noisy gems.
```
-""
+"WARN: Unresolved specs during Gem::Specification.reset:
+      ffi (>= 1.0.1)
+      multi_json (~> 1.10)
+WARN: Clearing out unresolved specs.
+Please report a bug if this causes problems.
+"
```

I need to run `bundle exec rake test:functional`.

I found a number of issues that I have addressed. Thank goodness for test suites.

burtlo:
Given a control defined like this:
```ruby
control "tmp-1.0" do                        # A unique ID for this control
  impact 0.7                                # The criticality, if this control fails.
  title "Create /tmp directory"             # A human-readable title
  desc "An optional description..."         # Describe why this is needed
  tag data: "temp data"                     # A tag allows you to associate key information
  tag "security"                            # to the test
  ref "Document A-12", url: 'http://...'    # Additional references

  describe file('/tmp') do                  # The actual test
    it { should be_directory }
    it { should_not be_directory }
  end
end
```

The results were ordered: fails + skips + passes. Generating the following:

```
Profile: InSpec example simple inheritance (simple inheritance)
Version: 1.0.0
Target:  local://

  ×  tmp-1.0: Create /tmp directory (1 failed)
     ×  File /tmp should not be directory
     expected `File /tmp.directory?` to return false, got true
     ✔  File /tmp should be directory
```

My refactored formatter orders based on the examples position within the control. Generating the following:

```
Profile: InSpec example simple inheritance (simple inheritance)
Version: 1.0.0
Target:  local://

  ×  tmp-1.0: Create /tmp directory (1 failed)
     ✔  File /tmp should be directory
     ×  File /tmp should not be directory
     expected `File /tmp.directory?` to return false, got true
```

Should I leave it as positioning or have them re-ordered based on previous ordering?

chris-rock:
@burtlo I like the new ordering, it is more natural for the user

burtlo:
This is getting trickier by the minute. What should the output look like when a profile includes another profile like the one defined in `simple_inheritance` which has a control file defined as:

```
include_controls 'failures'

describe file('/tmp') do
  it { should be_directory }
end
```

Because of the way the examples are processed and the way that the profiles are treated the output looks like the following at the moment:

```
Profile: InSpec example simple inheritance (simple inheritance)
Version: 1.0.0
Target:  local://


  File /tmp
     ✔  should be directory

  ×  tmp-1.0: File /tmp should not be directory (expected `File /tmp.directory?` to return false, got true)
     ×  File /tmp should not be directory
     expected `File /tmp.directory?` to return false, got true
  ✔  cmp-1.0: 7 should cmp == "7"
     ✔  7 should cmp == "7"
```

The anonymous example is processed first and belongs to the `simple_ineritance` profile. When the included controls are brought in no profile is added through `#add_profile` so there is `nil` for the profile. Which means we are switching profiles which means we need to display all of the anonymous examples and then display all of the new examples without a profile.

I think the best course of action is to see if I can get the profile added to the list of profiles known to the formatter when the `include_controls` happens.

I assume the desired output would be:
```
Profile: InSpec example simple inheritance (simple inheritance)
Version: 1.0.0
Target:  local://


  File /tmp
     ✔  should be directory

Profile: failures
Version: 0.1.0
Target: local://

  ×  tmp-1.0: File /tmp should not be directory (expected `File /tmp.directory?` to return false, got true)
     ×  File /tmp should not be directory
     expected `File /tmp.directory?` to return false, got true
  ✔  cmp-1.0: 7 should cmp == "7"
     ✔  7 should cmp == "7"
```

Which means I need to update `lib/inspec/runner.rb` to collect all the profiles from the requirements:
```ruby
    def load
      all_controls = []

      @target_profiles.each do |profile|
        @test_collector.add_profile(profile)
        write_lockfile(profile) if @create_lockfile
        profile.locked_dependencies
        # store the profile_context generated ...
        profile_context = profile.load_libraries
        # so that it can be used here to review the requirements
        profile_context.dependencies.list.values.each do |requirement|
          @test_collector.add_profile(requirement.profile)
        end

        @attributes |= profile.runner_context.attributes
        all_controls += profile.collect_tests
      end

      all_controls.each do |rule|
        register_rule(rule) unless rule.nil?
      end
    end
```

burtlo:
I'm getting intermittent failures when I run `bundle exec rake test:functional`. Most of the time everything passes. Occasionally an error or failure. 

```
$ bundle exec rake test:functional
/Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/requirement.rb:3: warning: loading in progress, circular require considered harmful - /Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/dependency_set.rb
# ...
Run options: --seed 63158

# Running:

.................................................................................F................................................

Finished in 76.951614s, 1.6894 runs/s, 6.6665 assertions/s.

  1) Failure:
example inheritance profile#test_0008_use lockfile in tarball [/Users/franklinwebber/inspec/inspec/test/functional/inspec_vendor_test.rb:135]:
Expected: 0
  Actual: 1

130 runs, 513 assertions, 1 failures, 0 errors, 0 skips
```

```
$ bundle exec rake test:functional
/Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/requirement.rb:3: warning: loading in progress, circular require considered harmful - /Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/dependency_set.rb
# ...
Run options: --seed 65266

# Running:

..................................................................................................................................

Finished in 79.132946s, 1.6428 runs/s, 6.5207 assertions/s.

130 runs, 516 assertions, 0 failures, 0 errors, 0 skips
```

```
$ bundle exec rake test:functional
/Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/requirement.rb:3: warning: loading in progress, circular require considered harmful - /Users/franklinwebber/inspec/inspec/lib/inspec/dependencies/dependency_set.rb
# ...
Run options: --seed 57062

# Running:

.............E....................................................................................................................

Finished in 96.696932s, 1.3444 runs/s, 5.3363 assertions/s.

  1) Error:
inspec exec#test_0001_can generate keys:
Errno::ENOENT: No such file or directory @ unlink_internal - a30878ac-8e99-4b9c-a340-5bf5f13e82a8.pem.pub
```

lib/inspec/runner.rb
```diff
@@ -80,7 +80,12 @@ def load
         @test_collector.add_profile(profile)
         write_lockfile(profile) if @create_lockfile
         profile.locked_dependencies
-        profile.load_libraries
+        profile_context = profile.load_libraries
+
+        profile_context.dependencies.list.values.each do |requirement|
+          @test_collector.add_profile(requirement.profile)
```
burtlo, line 86: To properly display the dependent profiles I needed to view the ProfileContext and the profiles within it. Is there a better way to grab the profiles that are required?

test/functional/inspec_exec_test.rb
```diff
@@ -197,7 +197,7 @@
     let(:out) { inspec('exec ' + simple_inheritance) }
 
     it 'should print the profile information and then the test results' do
-      out.stdout.force_encoding(Encoding::UTF_8).must_include "local://\n\n\n\e[38;5;9m  ×  tmp-1.0: Create /tmp directory (1 failed)\e[0m\n\e[38;5;9m     ×  File /tmp should not be directory\n"
+      out.stdout.force_encoding(Encoding::UTF_8).must_include "\e[38;5;9m  ×  tmp-1.0: Create /tmp directory (1 failed)\e[0m\n\e[38;5;41m     ✔  File /tmp should be directory\e[0m\n\e[38;5;9m     ×  File /tmp should not be directory\n"
```
burtlo, line 200: This test was expecting errors + passes + skips. Now the format of the output is in the natural order of the control.

lib/inspec/rspec_json_formatter.rb
```diff
@@ -75,6 +75,7 @@ def format_example(example)
 
     res = {
       id: example.metadata[:id],
+      profile_id: example.metadata[:profile_id],
```
burtlo, line 78: This makes it easier to get to the id of the profile instead of the previous process of trying to find it.

lib/inspec/rspec_json_formatter.rb
```diff
@@ -465,106 +582,160 @@ def format_lines(lines, indentation)
...
-      if x[:message]
-        msg = x[:code_desc] + "\n" + x[:message]
+  #
+  # This class wraps a control hash object to provide a useful inteface for
+  # maintaining the associated profile, ids, results, title, etc.
+  #
+  class Control
+    include Comparable
+
+    def <=>(other)
```
burtlo: Comparable is implemented here not for sorting but for better comparison between controls to ensure that similar named controls in different profiles are considered and displayed separately.

burtlo:
Other than those intermittent errors and failures I feel like this content is tested with the existing tests and ready for some more review.

burtlo:
I've seen this issue locally and it was fixed when I ran it again. https://travis-ci.org/chef/inspec/jobs/180257685#L631

burtlo:
Rebasing off of master this morning to keep it current.

burtlo:
Rebasing off of master this morning to keep it current.

This is a new error that seems related to the output of the formatter: https://travis-ci.org/chef/inspec/jobs/183295983#L686

I will take a look and address the issue.

burtlo:
say-it-again

I've re-run this several times on my local machine, with the seed that showed the issue `rake test:functional TESTOPTS="--seed=15875"` and other random seeds. They all pass.

Would someone run these tests on their system with this seed or other random seeds and ensure that you see it all working.

BTW I was getting serious errors when initially running the test because of some gems that needed cleaning that were gumming up the very brittle output that we were verifying against:

```
$ bin/inspec --help
    ffi (some version)
    gem_name (some version)
```

So I needed to do some gem cleaning: http://stackoverflow.com/questions/17936340/unresolved-specs-during-gemspecification-reset

lib/inspec/rspec_json_formatter.rb
arlimus: Agreed on the first. For the second my personal opinion has always been to prefer clarity. It does require one more step in thinking. overall ok :)

lib/inspec/rspec_json_formatter.rb
```diff
@@ -105,50 +106,36 @@ def initialize(*args)
...
     skipped = 0
     passed = 0
     critical = 0
     major = 0
     minor = 0
 
-    @control_tests.each do |control|
+    all_unique_controls = Array(@all_controls).uniq
+
+    all_unique_controls.each do |control|
```
arlimus, line 139: just consistency: your next change further down doesn't assign the variable
```
Array(@all_controls).uniq.each do |control|
```
pick one 😁 ; variables are great for clarity and re-use, but may not be needed in this case

lib/inspec/rspec_json_formatter.rb
```diff
@@ -205,35 +191,43 @@ def tests_summary
...
-    else
-      false
-    end
+  def examples_without_controls
+    examples.find_all { |example| example2control(example).nil? }
+  end
+
+  def examples_with_controls
+    (examples - examples_without_controls)
+  end
```
arlimus, line 203: Nice section overall, gz 👍

lib/inspec/rspec_json_formatter.rb
```diff
@@ -105,50 +106,36 @@ def initialize(*args)
     @backend = nil
   end
 
+  attr_reader :profiles
```
arlimus: Do we need to expose profiles? If not imho it's great to leave it private. What do you think?

lib/inspec/rspec_json_formatter.rb
```diff
@@ -205,35 +191,43 @@ def tests_summary
...
-  #
-  # TODO(ssd+vj): We should probably solve this by either ensuring the example has
-  # the profile_id of the top level profile when it is included as a dependency, or
-  # by registrying all dependent profiles with the formatter. The we could remove
-  # this heuristic matching.
-  #
-  def example2profile(example, profiles)
-    profiles.find { |p| profile_contains_example?(p, example) }
+  def examples
+    @examples ||= @output_hash.delete(:controls)
```
arlimus: I really like caching personally. In this example I'm not sure about when it is called. What do you think about assigning examples when they naturally need to be in the code flow?

arlimus (commented):
Overall I love these huge improvements. There is a lot in here and some fantastic changes to readability and flow. Let's get a few comments covered. Huge kudos @burtlo !!

lib/inspec/rspec_json_formatter.rb
burtlo: You're right it doesn't need to be public. The methods associated with the interface for formatter and `#add_profile` are likely the only things that need to be public. Will definitely make the change and make more things `private`.

lib/inspec/rspec_json_formatter.rb
burtlo: So true. 👍

lib/inspec/rspec_json_formatter.rb
burtlo: It is set through in `InspecRspecJson#stop` when it is called through  `InspecRspecJson#examples_without_controls` / `InspecRspecJson#examples_with_controls`.  I've didn't like the `@output_hash.delete(:controls)` so I've changed it to retrieve the value from the hash, no longer storing the value. So the method remains as a way to alias of `@output_hash[:controls]` to `examples`.

burtlo:
Rebasing off current chef/inspec:master

arlimus:
Thank you so much for the huge improvement @burtlo and for all the continuous fixes you have added to get this merged!! 🎉 👍 😀

PR #1333, opened 2016-11-29
fix bug: RegExp in processes resource can't match long-run process #1332

### LINKED ISSUE #1332: RegExp in processes resource can't match long-run process
### Description

It use ```([^ ]+)``` to match each ps columns in file: lib/resources/processes.rb.
But as man page of ps says, start format may be mmm dd or HH:MM:SS.
Because ```mmm dd``` comtains space, it can't be match with the RegExp. 

man of ps
```
start      STARTED  time the command started. If the process was started less than 24 hours ago, 
              the output format is  "HH:MM:SS", else it is "  mmm dd" 
              (where mmm is a three-letter month name). 
              See also lstart, bsdstart, start_time, and stime.
```

Tested against 1.5 and 1.6

### Replication Case

```ruby
# assume httpd processes are running more than 24 hours
describe processes('httpd') do
  its('list.length') { should be >= 1 }
end
```

It will output
```
 ×  httpd process: Process httpd should be running (expected: >= 1
          got:    0)
     ×  Processes list.length should be >= 1
     expected: >= 1
          got:    0
```

### Possible Solutions

Change 9th ```([^ ]+)```  to ```(\w{3} \d{2}|\d{2}:\d{2}:\d{2})```

Wing924:
Fix #1332 by change RegExp


chris-rock:
@Wing924 awesome addition. Lets add a test to our unit tests to ensure future changes will not break this behavior: https://github.com/chef/inspec/blob/master/test/unit/resources/processes_test.rb

Wing924:
@chris-rock Thank you for your comment. I added the test cases.
The travis test fails randomly. Can you help me with this problem?

chris-rock:
Great work @Wing924. Thank you for your contribution. I take care of the failing tests, since they are not related to your PR.

PR #1345, opened 2016-12-06
Fix variable name. Add test.

makotots:
Fix a reference to a variable which does not exist.
Also add a test case to demonstrate this issue and fix.


makotots:
Will resubmit after fixing the test.

PR #1348, opened 2016-12-08
use cached profile for compliance dependencies if vendored

chris-rock:
This PR prevents the case, where vendored profiles still try to fetch profiles from Chef Compliance

```
inspec check my-cis-profile

Cannot fetch compliance://cis/cis-amazonlinux-2014.09-2015.03-level1 because your compliance token has not been
configured.

Please login using

    inspec compliance login https://your_compliance_server --user admin --insecure --token 'PASTE TOKEN HERE' 
```


chris-rock:
Integration tests will follow, once the Automate tests are included

alexpop (approved):
Thanks Chris 💯

PR #1349, opened 2016-12-08
Adding support for virtuzzo os to service resource

jaxxstorm:
Signed-off-by: Lee Briggs <lbriggs@apptio.com>

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -134,9 +134,9 @@ module Inspec::Resources
         else
           SysV.new(inspec, service_ctl || '/usr/sbin/service')
         end
-      elsif %w{redhat fedora centos}.include?(platform)
+      elsif %w{redhat fedora centos parallels virtuozzo}.include?(platform)
         version = os[:release].to_i
-        if (%w{ redhat centos }.include?(platform) && version >= 7) || (platform == 'fedora' && version >= 15)
+        if (%w{ redhat centos parallels virtuozzo }.include?(platform) && version >= 7) || (platform == 'fedora' && version >= 15)
           Systemd.new(inspec, service_ctl)
         else
           SysV.new(inspec, service_ctl || '/sbin/service')


jaxxstorm:
Can't really seem to figure out why the functional tests are failing here, any ideas?

lib/resources/service.rb
```diff
@@ -134,9 +134,9 @@ def select_service_mgmt # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCom
         else
           SysV.new(inspec, service_ctl || '/usr/sbin/service')
         end
-      elsif %w{redhat fedora centos}.include?(platform)
+      elsif %w{redhat fedora centos parallels virtuozzo}.include?(platform)
```
chris-rock, line 137: maybe we should add `parallels` and `virtuozzo` platforms to https://github.com/chef/train/blob/master/lib/train/extras/os_detect_linux.rb#L17 and use os.redhat? + `fedora` as defined in https://github.com/chef/train/blob/master/lib/train/extras/os_common.rb#L45 in here

chris-rock:
@jaxxstorm I apologize for the late response. Great improvement. I think we should fix this in train, since this will then help coverage with other resources as well

lib/resources/service.rb
jaxxstorm: Yes I think that's a better idea, I already submitted the fix to train: https://github.com/chef/train/pull/166

But this was explicitly set as well so fixed it in both places. I can rebase and switch to use your suggestion if needed

chris-rock: I think the service resource was written before we had `os.redhat?` :-)

chris-rock: Train PR is merged, will be released on Friday with train 0.23.0

jaxxstorm:
@chris-rock how do you want me to go ahead here? I can fix up the failing tests or we can reconsider using `os.redhat` if you let me know what exactly you'd like to replace

chris-rock:
@jaxxstorm I would go forward with `os.redhat?`

chris-rock:
@jaxxstorm any updates on this?

jaxxstorm:
need to find some time to get back to this, sorry. Been dragged onto something else

adamleff:
This PR has been stale for a while, so we will be closing it in an effort to keep our work queue clean. 😄  However, we're more than happy to re-open it when you are ready to move forward.

PR #1362, opened 2016-12-13
new inspec.io frontpage

hannah-radish:
Base SCSS, JS, fonts, images, HTML ported over to old inspec.io
tempting engine.
—To do:
Media queries, Header, footer, docs, tutorials


chris-rock:
Thank you @hannah-radish for this awesome work!

Issue #1358, opened 2016-12-12
JSON resource's error message is missing filename when file is not found


makotots:
Hello!

### Description

In an example below, I expect filename is printed in its error message:

```
% inspec exec demo.rb
(snip)
  Json /nonexistent.json
     ↺  Can't find file ""    <---- here

% cat demo.rb
describe json('/nonexistent.json') do
  its('name') { should eq 'hello' }
end
%
```

### InSpec and Platform Version

InSpec version : 1.7.1
Platform Version: macOS Sierra 10.12.1

### Replication Case

See demo.rb above - run a test with a non-existent json filename.

### Possible Solutions

According to resources/json.rb, the error message refers to @conf_path, which is undefined.

```
        @path = opts
        @file = inspec.file(@opts)
        @file_content = @file.content

        # check if file is available
        if !@file.file?
          skip_resource "Can't find file \"#{@conf_path}\""
          return @params = {}
        end
```

I believe the pull request #1345 is almost correct to handle this (successful unit test included in the PR) but I couldn't figure out why one of the functional test failed in Travis CI. I ran the same test using Travis CI several times, which did not reproduce the same test failure.

### Stacktrace

No stacktrace is produced.

Thank you for your help in advance.

chris-rock:
@makotots Great finding. Are you pushing up your PR again? I am happy to help with the unit tests.

makotots:
@chris-rock Can you take a look at this PR (1363, not 1345) and help with the failed functional test?  Thanks!

PR #1363, opened 2016-12-15
WIP: JSON resource's error message is missing filename when file is not found

makotots:
See #1358


chris-rock:
@makotots You can run the unit tests locally as well with: `bundle exec ruby -W -Ilib:test test/unit/resources/json_test.rb`

chris-rock:
@makotots All good. It works as expected. Could you remove all unnecessary commits via `git rebase -i master` ? I think we only require `b7b7d28`

makotots:
I am withdrawing this and would like to continue at #1366.

PR #1364, opened 2016-12-15
extending cmp to support better version ops

jeremymv2:
This change hopes to extend the functionality of `cmp` matcher to allow for better comparisons when checking version strings.  It adds `Gem::Version` as a type for operand comparison.


chris-rock:
This is a great addition @jeremymv2

PR #1365, opened 2016-12-15
display if inspec version is outdated

chris-rock:
```
$ inspec version
1.7.0

Your version of InSpec is out of date! The latest version is 1.7.2.
```

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -9,6 +9,7 @@ require 'thor'
 require 'json'
 require 'pp'
 require 'utils/json_log'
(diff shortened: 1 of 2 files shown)


arlimus:
Nice feature overall! Reminds users if there is something new already or how far they are behind latest.

I think we should be careful here, since `inspec version` should be very highly failure resistant. More handling around timeouts to kill this would be great.

chris-rock:
@arlimus Great idea. I agree we should be really careful. I added a timeout of 500 milliseconds

arlimus:
Kudos!

PR #1366, opened 2016-12-15
JSON resource's error message is missing filename when file is not found

### LINKED ISSUE #1358: JSON resource's error message is missing filename when file is not found

makotots:
fixes #1358
Signed-off-by: Makoto Nozaki <makoto.nozaki@twosigma.com>

### DIFF
--- a/lib/resources/json.rb
+++ b/lib/resources/json.rb
@@ -44,13 +44,13 @@ module Inspec::Resources
 
         # check if file is available
         if !@file.file?
-          skip_resource "Can't find file \"#{@conf_path}\""
+          skip_resource "Can't find file \"#{@path}\""
           return @params = {}
         end
 
         # check if file is readable
         if @file_content.empty? && @file.size > 0
-          skip_resource "Can't read file \"#{@conf_path}\""
+          skip_resource "Can't read file \"#{@path}\""
           return @params = {}
         end
       end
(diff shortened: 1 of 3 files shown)


chris-rock:
Awesome. Thank you @makotots  for this fix!

PR #1368, opened 2016-12-15
add guidance for Serverspec migration

### LINKED ISSUE #804: Document Serverspec Migration
### Description

InSpec inherits a lot of similarities to Serverspec, especially the Syntax is very similar and most Serverspec resources are available in InSpec as well https://github.com/chef/inspec/issues/514

Nevertheless, InSpec and Serverspec have some differences
- InSpec is promoting a flat test structure instead of nested describe blocks
- No clear guide is available that helps users to easily migrate: https://github.com/chef/inspec/issues/514
- We should add a warning, if we detect Serverspec syntax that we do not support (e.g. nested describes)

InSpec 0.26.0

chris-rock:
fixes #804 

This is the first step to provide a migration guide for Serverspec

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1,0 +1,232 @@
+---
+title: InSpec Migration Guide
+---
+
+# Migrate from Serverspec to InSpec
+
+## How is InSpec different from Serverspec
+
+We've written a complete blog post about that topic: [The Road to InSpec](https://blog.chef.io/2015/11/04/the-road-to-inspec/)
+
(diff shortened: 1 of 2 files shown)


chris-rock:
cc @tas50 @jeremymv2 @grdnrio

arlimus:
Awesome work, thank you so much @chris-rock

tas50:
This will be super handy for new users. It might be worth touching on the differences in the OS detection logic. That was something that caused me a lot of grief when I started moving over from Serverspec

PR #1369, opened 2016-12-15
call ssh cookbook from prepare cookbook

chris-rock:
make dependency to ssh-hardening cookbook visible and prevent it from running on windows


arlimus:
Thank you!!

PR #1373, opened 2016-12-17
control and lib_eval_context unit tests

jeremymv2:
I found that these unit tests were not being executed due to their file names ending in `.._tests.rb` instead of `.._test.rb`.  Not sure if that is intentional or not.  I did rename both and I had to modify `control_eval_context_test.rb` because it seemed to be based on an outdated implementation?


chris-rock:
@jeremymv2 Can you please rebase on latest master? Everything should be green by then.

jeremymv2:
@chris-rock sorry for the delay.  I was fully checked out on PTO - back now!

chris-rock:
@jeremymv2 Cool. No worries. Thank you for adding the tests

PR #1384, opened 2017-01-02
drop ruby 1.9.3

chris-rock:
This PR:
- removes ruby 1.9.3 (its is becoming too difficult to maintain 1.9.3 with pinning all dependencies in Gemfile)
- update functional tests to meet the updated profile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,48 +9,47 @@ services:
 - docker
 bundler_args: "--without integration tools maintenance deploy"
 before_install:
-- gem install bundler
-- gem update --system 2.4.5
+- gem update --system
+- gem update bundler
 - gem --version
 matrix:
   include:
-  - rvm: 1.9.3
-  - rvm: 2.1.8
-  - rvm: 2.2.5
-  - rvm: 2.3.1
+  - rvm: 2.1.10
+  - rvm: 2.2.6
+  - rvm: 2.3.3
     script: bundle exec rake $SUITE
     env: SUITE="lint test test:functional"
-  - rvm: 2.3.1
+  - rvm: 2.3.3
     script: bundle exec rake $SUITE
     env: SUITE="test:resources config=test/test.yaml" N=2
-  - rvm: 2.3.1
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -9,14 +9,14 @@ services:
...
+- gem update bundler
 - gem --version
 matrix:
   include:
-  - rvm: 1.9.3
-  - rvm: 2.1.8
-  - rvm: 2.2.5
+  - rvm: 2.1.10
+  - rvm: 2.2.6
+  - rvm: 2.3.3
```
arlimus, line 19: instead of adding `2.3.3` here, what do think about changing all the `2.3.1` runs below to this release instead?

appveyor.yml
```diff
@@ -6,7 +6,6 @@ environment:
   bundler_url: https://rubygems.org/downloads/bundler-1.9.9.gem
 
   matrix:
-    - ruby_version: "193"
```
arlimus, line 9: imho we should either add ruby_version 21 here or remove it from bundler above; what do you think?

arlimus (changes requested):
Thank you! Ruby 1.9.3 officially ended [Feb 23rd 2015](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/), however, we had continue to support it because it is still in supported distros like RHEL6.

We will continue to support a broad range of runtimes, but need to reduce 1.9 at this point, because it is blocking the selection of newer gem releases in our dependencies.

.travis.yml
chris-rock: Sounds like a good plan?

appveyor.yml
chris-rock: Not sure what you are requesting?

arlimus: let's add ruby 2.1

chris-rock: done

arlimus:
Thank you!

PR #1387, opened 2017-01-02
Unit test the 'package' resource for OSX (with brew)

jvrplmlmn:
See https://github.com/chef/inspec/issues/1386 for reference.


arlimus:
Thank you @jvrplmlmn !! 👍 

Merging with failures; For the fix of the underlying parser see https://github.com/chef/inspec/pull/1389

PR #1402, opened 2017-01-05
Download InSpec profiles from Chef Compliance

chris-rock:
```
$ inspec compliance download base/linux
Downloading `base/linux`
Profile stored to linux.tar.gz
$ inspec compliance download base/linux2
Profile base/linux2 is not available in Chef Compliance.
$ echo $?
1
```


arlimus:
Thank you @chris-rock 👍 !!

PR #1403, opened 2017-01-05
Add an http test method

### LINKED ISSUE #336: HTTP request resource
As `kitchen-inspec` runs remotely (unlike `busser-serverspec`) any tests using an HTTP client (e.g. `Net::HTTP`) to test HTTP responses on localhost will no longer work. This should be reworked into some kind of `http_request` resource that works in both local and remote execution.

guilhem:
fix #336


lib/resources/http.rb
```diff
@@ -0,0 +1,51 @@
...
+# license: All rights reserved
+
+require 'requests'
+
+module Inspec::Resources
+  class Http < Inspec.resource(1)
+    name 'http'
+    desc 'Use the http InSpec audit resource to test http call.'
+    example "
+      describe http(http://localhost:8080/ping', auth: ['user, 'test'], params: {format: 'html'}) do
```
chris-rock: maybe better:

```
describe http('http://localhost:8080/ping', auth: ['user, 'test'], params: {format: 'html'}) do
        its('body') { should cmp 'pong' }
        its('Content-Type') { should include('text/html') }
end
```

inspec.gemspec
```diff
@@ -38,4 +38,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'sslshake', '~> 1'
   spec.add_dependency 'parallel', '~> 1.9'
   spec.add_dependency 'rspec_junit_formatter', '~> 0.2.3'
+  spec.add_dependency 'requests', '~> 1.0.0'
```
chris-rock: Do we really need the requests module? It is not maintained for a couple of years anymore. See https://github.com/cyx/requests.

lib/resources/http.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    # rubocop:disable ParameterLists
+    def initialize(url, method: 'GET', params: nil, auth: nil, headers: {}, data: nil)
+      @method = method
+      @url = url || 'http://localhost/'
+      @params = params
+      @auth = auth
+      @headers = headers
+      @data = data
+      @request = Requests.request(@method, @url,
```
chris-rock: I think we should try to use Net::HTTP. We could extract some helpers and place them into inspec utils from https://github.com/chef/inspec/blob/master/lib/bundles/inspec-compliance/http.rb

chris-rock (changes requested):
@guilhem This is a great addition to our suite. Lets try to use native ruby classes instead of the requests gem. I propose to reuse http helper already used for the compliance plugin. We could move the helper to inspec/lib/utils https://github.com/chef/inspec/tree/master/lib/utils

inspec.gemspec
guilhem: @chris-rock yes it isn't well maintained but it works like a charm.
I try to use native Net::HTTP but it finally was quite the same as `requests`... with less stability.
I also try to find another lib without success.

`requests` wasn't a wonderful choice but a pragmatic one. :/

lib/resources/http.rb
guilhem: as said in other comment, doing stuff in `Net::HTTP` is really painful.
_https_, _headers_, _path_, _params_, _method_ everything is in different class without good layering.
URI / Request / ...
`Requests` already do the stuff. If 1 day requests stop working (with a ruby X.X) it would be time to vendor it inside our code.

For the moment it seems quite overkill to refactor inside our code a lib without any other dependency.

lib/resources/http.rb
guilhem: Done :)

inspec.gemspec
guilhem: @chris-rock or I just find http://www.rubydoc.info/gems/http/HTTP/Request https://github.com/httprb/http who is:
* quite maintained :)
* some more work to use it "as it" (basic auth etc) :|
* has some dependencies :(

It's up to you :)

lib/resources/http.rb
chris-rock: I am open for another gem. Unfortunately we cannot go with the current one. Since, once merged, this PR it is part of a maintained project, we need to be sure we get all PRs in a way that we can maintain them going forward. I am open for alternative solutions.

guilhem: @chris-rock link to https://github.com/chef/inspec/pull/1403#discussion_r94944994
is https://github.com/httprb/http ok for you?

guilhem:
@chris-rock I changed to use `http` lib :)

chris-rock:
@guilhem I apologize for the delayed response. I'd like your change. Can you please add documentation as well in here https://github.com/chef/inspec/tree/master/docs/resources?

chris-rock:
@guilhem Anything I can help you with to get this merged this week?

guilhem:
@chris-rock sorry I'm in holiday this week, but you can push anything you want in my branch :)

chris-rock:
@guilhem Enjoy our holiday, lets bring this in next week then.

guilhem:
@chris-rock should be ok

lib/resources/http.rb
```diff
@@ -0,0 +1,49 @@
+# encoding: utf-8
+# copyright: 2017, Criteo
+# author: Guilhem Lettron
+# license: All rights reserved
```
chris-rock: We do not allow `All rights reserved`. Every contribution needs to be Apache 2 licensed

lib/resources/http.rb
```diff
@@ -0,0 +1,49 @@
...
+    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil)
+      @url = url
+      @method = method
+      @params = params
+      @auth = auth
+      @headers = headers
+      @data = data
+      http = HTTP.headers(@headers)
+      http = http.basic_auth(@auth) unless @auth.empty?
+      @response = http.request(@method, @url, { body: @data, params: @params })
```
chris-rock: We should do a lazy response request, otherwise the request will be done during resource instantiation. Lets have a private method `request` that is called by all other functions and it should also cache the result in `@response`

chris-rock (changes requested):
Awesome work @guilhem I just have some minor comments. Once those are fixed, we are ready to get the resource in

lib/resources/http.rb
guilhem: ok for me, I just copy/pasta other resources with same lines :)

chris-rock:
@guilhem Can you rebase on latest master? functional tests are fixed there. I apologise for the inconvenience.

guilhem:
@chris-rock should be ok :)

chris-rock:
Awesome work @guilhem

chris-rock (approved):

Issue #1401, opened 2017-01-05
apache_conf does not read symlinked configuration files


carldjohnston:
### Description

Symlinked configuration files are not read by the apache_conf resource. 

Using the apache2 cookbook to configure apache will create links to configuration files from `conf-enabled/configfile.conf` to `conf-available/configfile.conf`; the `conf-enabled/*.conf` directory is then included through an `IncludeOptional` apache directive. I believe this is the default behavior on Ubuntu also (without the apache2 cookbook).

### InSpec and Platform Version

InSpec 1.7.1 on Oracle Linux 7.3

### Replication Case

Apply the apache2::default recipe to a system, and check for directives that are included by default in `/etc/httpd/conf-enabled/security.conf` such as `ServerTokens Prod` with:

```
describe apache_conf do
  its('ServerTokens') { should cmp 'Prod' }
end
```

### Possible Solutions

Perhaps adding the following as line 110 to apache_conf.rb might solve the issue:
`files += find_files(id, depth: 1, type: 'link')`

https://github.com/chef/inspec/blob/master/lib/resources/apache_conf.rb#L109

chris-rock:
Are you opening up a PR to fix that?

adamleff:
Believe this is fixed in #1406, so I'm going to close this issue. Please reopen if that's not the case. Thanks!

Issue #1394, opened 2017-01-04
apache_conf resource does not include files from *.d directories when using conf_path parameter


carldjohnston:
### Description

apache_conf resource does not include files from *.d directories when using conf_path parameter.

### InSpec and Platform Version

InSpec version 1.7.1 on Oracle Linux 7.3

### Replication Case

Install httpd on an EL7 flavor of linux.  Then we'll test for something that's included into the `/etc/httpd/conf.d/autoindex.conf` file in a default install.

From within inspec shell (on an EL7.3 system) showing normal behavior (no conf_path):
```
inspec> describe apache_conf do
inspec>   its('DefaultIcon') { should cmp '/icons/unknown.gif' }
inspec> end

Profile: inspec-shell
Version: unknown


  Apache Config
     ✔  /etc/httpd/conf/httpd.conf DefaultIcon should cmp == "/icons/unknown.gif"

Test Summary: 1 successful, 0 failures, 0 skipped
```
Again from within inspec shell (on an EL7.3 system) showing broken behavior (with conf_path pointed to default location):
```
inspec> describe apache_conf('/etc/httpd/conf/httpd.conf') do
inspec>   its('DefaultIcon') { should cmp '/icons/unknown.gif' }
inspec> end
find_files(): exit 1 from `find /etc/httpd/conf/conf.modules.d/*.conf -maxdepth 1 -type f`
find_files(): exit 1 from `find /etc/httpd/conf/conf.d/*.conf -maxdepth 1 -type f`

Profile: inspec-shell
Version: unknown


  Apache Config
     ∅  /etc/httpd/conf/httpd.conf DefaultIcon should cmp == "/icons/unknown.gif"

     expected: "/icons/unknown.gif"
          got:

     (compared using `cmp` matcher)


Test Summary: 0 successful, 1 failures, 0 skipped
```
### Possible Solutions

It appears that line 24 of the apache_conf resource (https://github.com/chef/inspec/blob/master/lib/resources/apache_conf.rb#L24) might need to trim the `conf/` folder on EL7 to end up with `conf_dir = '/etc/httpd'`

chris-rock:
@carldjohnston Are you saying we need to update the directories here: https://github.com/chef/inspec/blob/master/lib/resources/apache.rb#L12-L24?

carldjohnston:
Hi @chris-rock, I think that code in apache.rb only gets called when we don't specify a conf_path and it does get the correct conf_dir.
[...]

PR #1406, opened 2017-01-07
Allow apache_conf to include symlinked configuration files

carldjohnston:
Fix for #1401 

Signed-off-by: Carl Johnston <carldjohnston@gmail.com>


chris-rock:
@carldjohnston Great work. I would like to see an integration test for this PR to verify everything works as expected.

carldjohnston:
Hi @chris-rock - Hoping these tests are what you were after.

I've basically mocked the "find ..-type l" commands that the original commit generates and pointed it at "security.conf"; then we're looking for "ServerSignature Off" within that file.

I had to poke through and learn how it's all put together - thanks for teaching me something new :)

adamleff:
These are great additional unit tests, @carldjohnston - thank you!  However, @chris-rock was asking for some integration tests where we actually run InSpec on a real filesystem/configs to check to see that the resource functions correctly.

Here's a suggestion on how you might do that:
 * Modify the `os_prepare::apache` recipe (https://github.com/chef/inspec/blob/master/test/cookbooks/os_prepare/recipes/apache.rb) to write out:
    * an apache config that will be `Include`'d 
    * a symlink to that apache config
    * a new top-level apache config that `Include`s the symlink
 * Modify the apache_conf integration spec (https://github.com/chef/inspec/blob/master/test/integration/default/apache_conf_spec.rb) to test that you can read the the new top-level apache config and read values from the symlink'd include conf.

Hopefully that makes sense - if not, let me know!

carldjohnston:
Thanks @adamleff, that makes perfect sense.

I'm working through this and have a working config, just the inspec side isn't working as expected.

By default (without parameters) using the apache_conf resource defaults to looking at /etc/httpd/conf/httpd.conf, however the os_prepare cookbook makes use of the httpd cookbook, which is putting configuration in /etc/httpd-default/conf/httpd.conf, and that's not being tested correctly as it's not including files correctly.. See Issue #1394.

I can rework this to use the apache2 cookbook, which replaces configuration in the default location and would make it easier to test for linked files, or we could work through the above issue - what do you think?

adamleff:
@carldjohnston so sorry for the delay in my response!

I think there might be a middle-ground we can pursue until a fix for #1394 is committed. In the os_prepare cookbook, in the apache recipe, we can just use a combination of Chef `directory` and either `template` or `file` resource to write out a file in the default location `/etc/httpd/conf/httpd.conf` for purposes of this test. We don't have to use the `httpd` helper cookbook to do so. I think that would be enough to lay down the test fixtures we need to properly test this.

Does that make sense? What do you think of that approach?

carldjohnston:
Hi @adamleff, I'm happy to go with that approach, probably makes understanding the os_prepare cookbook a little easier too; I'll put it together in the next couple of days.

carldjohnston:
Hi @adamleff, I've added those modifications, let me know if you need anything else.

Thanks!

adamleff:
Haven't forgotten about you @carldjohnston! This is on my list for tomorrow if not later today!

adamleff (approved):
Superb work, @carldjohnston. Thanks for being agreeable to enhance the tests, as well!

chris-rock (approved):
@carldjohnston Thank you for this improvement

adamleff:
Merging! Thanks again, @carldjohnston!

PR #1412, opened 2017-01-13
Add ability to get a new token from a refresh token set in the config

brentm5:
With this refresh tokens can now be used as intended, instead of logging in again with all the required you can refresh the token on disk with a simple command.

Honestly this is a stopgap instead of fixing the issue which is the token never gets updated if there is an issue with login and a refresh token is added.  This would allow automated systems a way to keep updating their authentication credentials with refresh tokens.

### DIFF
--- a/lib/bundles/inspec-compliance/cli.rb
+++ b/lib/bundles/inspec-compliance/cli.rb
@@ -57,6 +57,16 @@ module Compliance
       puts '', msg
     end
 
+    desc 'login_refresh', 'Refresh compliance login token from a stored refresh token'
+    def login_refresh
+      config = Compliance::Configuration.new
+      return if !loggedin(config)
+
+      _success, msg = store_refresh_token(config['server'], config['refresh_token'], true, config['user'], config['insecure'])
+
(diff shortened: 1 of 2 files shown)


brentm5:
The test that failed in travis is something I did not even touch not sure if this is a known issue

```
example inheritance profile#test_0002_can vendor profile dependencies from the profile path [/home/travis/build/chef/inspec/test/functional/inspec_vendor_test.rb:40]:
```

chris-rock:
@brentm5 I think we should do this automatically if users stored a refresh token already. Therefore we just exchange the refresh_token for an access_token. For that case we should:

- check if access token has expired
- if expired, use refresh token to retrieve a new access token

brentm5:
Yeah I think that is a better approach.  Just wasn't 100% sure of all the touch points.  I will go ahead and close this and try and submit something to solve it this week hopefully.

chris-rock:
Awesome, thank you @brentm5 Ping me on Slack if you have questions.

brentm5:
Created #1414 to track this

PR #1413, opened 2017-01-16
Docs examples: use double quotes to prevent escaping backslash in the expected string

techraf:
Using single quotes produced:

```
  ×  echo-01: Test echo (
     expected: "hello\\n"
          got: "hello\n"
```

Using double quotes fixes the problem.


chris-rock:
@techraf Thank you for pointing this out.

This is an obvious fix according to our contribution guidelines. https://github.com/chef/inspec/blob/master/CONTRIBUTING.md Therefore we do not need the DCO. Please sign-off your commits next time.

Issue #1414, opened 2017-01-16
Compliance server actions do not auth correctly with refresh token


brentm5:
### Description

If you login to a compliance server with a refresh token the token will only be valid for the max age before a token expires.  This happens because the login action simply takes the refresh token and generates a token which gets saved with the configuration on disk.

### InSpec and Platform Version

All Versions of inspec

### Replication Case

run the following with valid values `inspec compliance login inspec compliance login <compliance address> --user='<user>' --refresh_token='<refresh_token>'`
Wait an hour or so
Try to run `inspec compliance profiles` or any other command that requires auth to compliance

### Possible Solutions

The solution is to actually implement how the refresh_tokens get used instead of  just using them to generate a token.

chris-rock:
Thank you @brentm5 Indeed, this is annoying.

jerryaldrichiii:
@brentm5, is this still an issue?

clintoncwolfe:
As Compliance is no longer a current product, and the behavior described appears to be as intended, we're going to close this for now.

PR #1416, opened 2017-01-18
Updated compliance api requests to actually use refresh token correctly

brentm5:
We do not store a token in the config file but rather generate one on each command called.  This also does not cache any tokens but generates them every time.  Not sure if this is ideal or not but this a first pass and needs some work.  Also there are no tests added yet.  At this point just looking for feedback on the approach.  This is to help solve #1414

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -81,7 +81,7 @@ Please login using `inspec compliance login https://compliance.test --user admin
       [res.is_a?(Net::HTTPSuccess), res.body]
     end
 
-    # Use username and refresh_toke to get an API access token
+    # Use username and refresh_token to get an API access token
     def self.get_token_via_refresh_token(url, refresh_token, insecure)
       uri = URI.parse("#{url}/login")
       req = Net::HTTP::Post.new(uri.path)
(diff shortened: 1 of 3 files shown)


lib/bundles/inspec-compliance/target.rb
```diff
@@ -55,6 +55,9 @@ def self.resolve(target) # rubocop:disable PerceivedComplexity, Metrics/Cyclomat
         end
         profile_fetch_url = Compliance::API.target_url(config, profile)
       end
+      # We need to pass the token to the fetcher
+      config['token'] = Compliance::API.get_token(config)
```
brentm5, line 59: Looking at this it might make sense to just follow this convention everywhere.  Just on each request we set the config['token'] so it lives the entire life of the command but then do not save this to disk.

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -179,7 +179,7 @@ def upload(path) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Percei
       end
 
       # determine user information
-      if config['token'].nil? || config['user'].nil?
+      if (config['token'].nil? && config['refresh_token'].nil?) || config['user'].nil?
```
brentm5, line 182: Having one or the other is now a valid use case, if a token is still in the config it will default to using that (which is the legacy behavior

lib/bundles/inspec-compliance/api.rb
```diff
@@ -130,20 +130,29 @@ def self.get_token_via_password(url, username, password, insecure)
...
       else
-        headers = { 'Authorization' => "Bearer #{config['token']}" }
+        headers = { 'Authorization' => "Bearer #{token}" }
       end
       headers
     end
 
+    def self.get_token(config)
+      return config['token'] unless config['refresh_token']
+
```
chris-rock: we do not need the spaces here

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -344,11 +343,10 @@ def store_refresh_token(url, refresh_token, verify, user, insecure)
         success = true
         msg = 'API refresh token stored'
       else
-        success, msg, access_token = Compliance::API.get_token_via_refresh_token(url, refresh_token, insecure)
+        success, msg = Compliance::API.get_token_via_refresh_token(url, refresh_token, insecure)
```
chris-rock: I think we need the `_access_token` here as well

lib/bundles/inspec-compliance/target.rb
chris-rock: Good idea. In general we want to avoid having the configuration in our API. Right now, storing it on disk is the only way to share it with the fetcher, since they run independently.

brentm5: So currently this does not save to disk so if that is a requirement this approach might need to be changed.

brentm5:
@chris-rock are the functional tests known to have these failures or is that something I broke?

https://travis-ci.org/chef/inspec/jobs/193035548

chris-rock:
I am taking care of the functional tests right now.

chris-rock:
@brentm5 Can you rebase on latest master? functional tests are fixed there. I apologise for the inconvenience.

brentm5:
@chris-rock Done

chris-rock:
Very nice improvement @brentm5

PR #1423, opened 2017-01-20
Docs: fix resource name ('processes' instead of 'process')

techraf:
Currently `processes` resource is listed as `process` on the [InSpec Resources Reference](http://inspec.io/docs/reference/resources/) page.


chris-rock:
Thank you @techraf This is an obvious fix. Therefore we do not require a DCO.

PR #1424, opened 2017-01-21
Add '--overwrite' argument suggestion to 'Vendoring dependencies'

techraf:
Running `inspec vendor` in when adding or updating dependencies always produces the error `Profile is already vendored. Use --overwrite.`.

The only case when it does not produce the error is when creating a profile from scratch - but this clause specifically addresses "adding or updating".

So I suggest adding the parameter to the documentation.


chris-rock:
@techraf could you please sign-off your work? See https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

chris-rock:
@arlimus This raises a good question. Since `inspec vendor` is intended to update the dependencies, we may make overwrite a default? What do you think?

adamleff:
This PR has been stale for quite a while. Therefore, I'm closing it for now. We will be happy to reopen it when the work has a valid DCO. Thank you for your contribution :)

PR #1432, opened 2017-01-26
improve http header handling

chris-rock:
- removes method_missing
- exposes headers as a hash
- improves output of `to_s` method


chris-rock:
cc @guilhem

guilhem:
LGTM

I wasn't a huge fan of `method_missing`.
But it lack doc change on headers ;)

chris-rock:
@guilhem :-) Good catch

Issue #993, opened 2016-08-30
kernel_module should support checking module version loaded


nvtkaszpir:
### Description

Currently inspec in 0.32.0 supports checking if kernel module is loaded.
Would be nice to extend it with checking if module version is valid.
### InSpec and Platform Version

inspec 0.32.0
### Possible Solutions

Extend kernel_module with additional methods.

Currently this can be done using other modules, for example command(), and checking its output, like this:

```
control 'aws-sriovNetSupport-001' do
  impact 1.0
  title 'kernel module ixgbevf should be minimum version is 2.14.2'
  desc  'http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sriov-networking.html'
  describe command('modinfo -F version ixgbevf') do
    its('stdout') { should match (/^2.14.2.*/) }
  end
end

```

This can be tested on AWS EC2 instance launched in VPC using AMI with HVM support (for example Amazon Linux AMI).
No sure if possible to test it within docker.
### Stacktrace

No stacktrace.

chris-rock:
That is a great idea @nvtkaszpir Do you want to add the version method to https://github.com/chef/inspec/blob/master/lib/resources/kernel_module.rb#L37?

nvtkaszpir:
sorry but my knowledge of writing in ruby is not sufficient to fulfill this task, and learning how to do it is currently loosing a battle with amount of daily work assigned to me :(

chris-rock:
@nvtkaszpir would the following syntax work for you?

```
describe kernel_module('ixgbevf') do
  it { should be_loaded }
  its('version') { should match (/^2.14.2.*/) }
end
```

nvtkaszpir:
Yeha this looks good.
I was wondering if there is an option to specify minimum version matching, like in Gemfile with '>=2.14.2', because I think this would be the most versatile option (I just don't know if match or eq support this at all).

aladmit:
I will add the version method. And I think we should add **have_version** matcher with the following syntax for compare versions:

```
describe kernel_module('ixgbevf') do
    it { should be_loaded }
    it { should have_version '>= 2.14' }
end
```

chris-rock:
@postgred Do we really need the have_version matcher or would cmp matcher work? cmp already understands version comparison.

```
describe kernel_module('ixgbevf') do
  it { should be_loaded }
  its('version') { should cmp >= 2.14 }
end
```

aladmit:
@chris-rock You are right, cmp will work.

[...]

PR #1435, opened 2017-01-27
Version method for kernel_module

aladmit:
Hi! This version method for kernel_module.

issue: #993


lib/resources/kernel_module.rb
```diff
@@ -36,6 +36,16 @@ def loaded?
       !found.nil?
     end
 
+    def version
+      # default modinfo command
+      modinfo_cmd = "modinfo -F version #{@module}"
+      # command for CentOS 5 and sudo
+      modinfo_cmd = "/sbin/modinfo -F version #{@module}" if inspec.os[:name] == 'centos' && inspec.os[:release].to_i == 5
```
chris-rock: maybe:

```
if inspec.os.name == 'centos' && inspec.os.release.to_i == 5
   ...
else
   modinfo_cmd = "modinfo -F version #{@module}"
end

```

lib/resources/kernel_module.rb
```diff
@@ -36,6 +36,16 @@ def loaded?
...
+
+      cmd = inspec.command(modinfo_cmd)
+      cmd.exit_status.zero? ? cmd.stdout.gsub("\n", '') : false
```
chris-rock: It may be better to return `nil`, since we could not determine a version

chris-rock (commented):
Thank you @postgred for this great addition. I added ideas for minor improvements.

Right now, rubocop is failing with:
```
Offenses:
lib/resources/kernel_module.rb:46:42: C: Use delete instead of gsub.
      cmd.exit_status.zero? ? cmd.stdout.gsub("\n", '') : false
                                         ^^^^^^^^^^^^^^
```

Could you please sign-off your work, too. Instructions are available here: https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

chris-rock:
@postgred Thank you for the quick improvements. The implementation looks great. I was just thinking, we should add this to our integration tests as well, to ensure it works across linux systems. The test is located here: https://github.com/chef/inspec/blob/master/test/integration/default/kernel_module_spec.rb#L21-L23

This will be executed with every travis run. See https://github.com/chef/inspec/blob/master/.travis.yml#L28-L55

aladmit:
@chris-rock We can't add this to integration tests because **modinfo** can't work in docker. Example:

```
root@580f55841391:/# modinfo video
libkmod: ERROR ../libkmod/libkmod.c:556 kmod_search_moddep: could not open moddep file '/lib/modules/4.9.5-200.fc25.x86_64/modules.dep.bin'
modinfo: ERROR: Module alias video not found.
```

chris-rock:
@postgred Agreed. Thank you.

PR #1444, opened 2017-01-31
Fix xinetd parsing of services from the same file. Expose resource.protocols

alexpop:
When services are in the same file, with the `tcp` and `udp` protocol configurations, only the second one was being exposed by the resource. This PR fixes that.

It also adds another mocked xinetd file, same way they are by default in Ubuntu:
```bash
[root@vagrant-local-ubuntu1404]# ls /etc/xinetd.d/echo*
/etc/xinetd.d/echo
```

where the existing `chargen` mocked configs follow the split RedHat format:
```bash
[root@ap-cc6 ~]# ls /etc/xinetd.d/echo*
/etc/xinetd.d/echo-dgram  /etc/xinetd.d/echo-stream
```

### DIFF
--- a/lib/resources/xinetd.rb
+++ b/lib/resources/xinetd.rb
@@ -41,6 +41,7 @@ module Inspec::Resources
           .add(:ids,          field: 'id')
           .add(:socket_types, field: 'socket_type')
           .add(:types,        field: 'type')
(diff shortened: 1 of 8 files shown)


chris-rock (approved):

chris-rock:
Thank you @alexpop

PR #1445, opened 2017-01-31
Restrict to net-ssh < 4

cnunciato:
The current stable version of Chef (12.18.19) restricts net-ssh to 3.2, but Inspec (via train), allows for version 4, which breaks chef-client runs. This change applies the same < 4 restriction.

Signed-off-by: Christian Nunciato <cnunciato@chef.io>

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
+  spec.add_dependency 'net-ssh', '>= 2.9", "< 4.0'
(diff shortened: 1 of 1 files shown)


tduffield (changes requested):
This will break ChefDK

tduffield:
So, clarification. This will **not** break ChefDK 1.2 but it **will** break ChefDK 1.3. If we downgrade net-ssh to <4 we will need to re-upgrade it when we release chef-client 12.19 in a week or so otherwise it will break the current channel of ChefDK

chris-rock:
@cnunciato InSpec does not have a dependency on net-ssh. If we need a fix, that needs to be done in train. See https://github.com/chef/train/blob/master/train.gemspec#L31

PR #1448, opened 2017-02-01
derive xinetd protocol from socket_type when not defined in the config file

alexpop:
> SocketType will sometimes implicitly indicate which ProtocolType will be used within an AddressFamily. For example when the SocketType is Dgram, the ProtocolType is always Udp.When the SocketType is Stream, the ProtocolType is always Tcp.

https://msdn.microsoft.com/en-us/library/system.net.sockets.sockettype(v=vs.110).aspx

From my research, the most common `socket_type` values are `stream`(TCP) and `dgram`(UDP). There's no implicit protocol for `raw` and `seqpacket`.

### DIFF
--- a/test/unit/resources/xinetd_test.rb
+++ b/test/unit/resources/xinetd_test.rb
@@ -34,10 +34,22 @@ describe 'Inspec::Resources::XinetdConf' do
       _(one.ids).must_equal %w{chargen-dgram}
     end
 
-    it 'get all protocols' do
+    it 'get all protocols for echo' do
       one = resource.services('echo')
       _(one.protocols).must_equal %w{tcp udp}
       _(one.ids).must_equal %w{echo-stream echo-dgram}
+    end
+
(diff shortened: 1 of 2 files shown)


test/unit/resources/xinetd_test.rb
```diff
@@ -34,12 +34,24 @@
...
+    it 'get all protocols for chargen, including derived from socket_type' do
+      one = resource.services('chargen')
+      _(one.protocols).must_equal %w{tcp udp}
+      _(one.ids).must_equal %w{chargen-stream chargen-dgram}
+    end
+
+    it 'params has only the protocols parsed from the config files' do
+      one = resource.params['services']['chargen'].map{|x| x.params['protocol']}
+      # in this example(CentOS), protocol is not defined in the config
+      _(one).must_equal [nil, nil]
```
chris-rock, line 52: I like this separation! Thank you @alexpop

PR #1451, opened 2017-02-02
Add FilterTable support to processes resource

alexpop:
Looking for a quick review before writing tests for the new properties.
Ok with the `.list` deprecation for the commonly used `entries`?

### DIFF
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -4,6 +4,8 @@
 # author: Christoph Hartmann
 # license: All rights reserved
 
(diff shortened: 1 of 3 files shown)


arlimus:
Hey @alexpop , awesome to see this move to FilterTable!! Could you please also change the current tests from `list` to `entries` to reflect the deprecation? Maybe keep one call to `list` in there to make sure we are checking the deprecated command until we remove it.

arlimus:
Awesome work, thank you @alexpop ! 👍

PR #1452, opened 2017-02-02
switch to faraday as http backend

chris-rock:
This PR switches to faraday as http backend lib. It has less dependencies and does not require `ruby-dev`.


arlimus:
Awesome change, much appreciated @chris-rock !!!

PR #1458, opened 2017-02-03
add "packages" resource

jtimberman:
This pull request adds a packages resource so that we can check for pattern matches against all the packages on a system. This initially implements only dpkg support for debian-based platforms so we can cover this use case:

```ruby
describe packages(/^xserver-xorg.*/) do
  its("list") { should be_empty }
end
```

This uses FilterTable so we can supply additional queries, too.

```ruby
describe packages(/vi.+/).where { status != 'installed' } do
  its('statuses') { should be_empty }
end
```

Users can specify the name as a string or a regular expression. If it is a string, we will escape it and convert it to a regular expression to use in matching against the full returned list of packages. If it is a regular expression, we take that as is and use it to filter the results.

While some package management systems such as `dpkg` can take a shell glob argument to filter their results, we eschew this and require a regular expression to match multiple package names because we will need this to work across other platforms in the future. This means that the following:

```ruby
packages("vim")
```

Will return *all* the "vim" packages on the system. The `packages` resource will take `"vim"`, turn it into `/vim/`, and greedily match anything with "vim" in the name. To match only a single package named `vim`, it needs to be an anchored regular expression.

```ruby
packages(/^vim$/)
```


alexpop:
I added a few commits to use the more standard `entries`, add a few more tests, fix lint.

`list` as an attribute only works on the main resource(i.e. `packages("xserver-xorg*").list`) where `entries` works on the main resource + on where filter(i.e. `packages("xserver-xorg*").where { status == 'installed' }.entries`)

lib/resources/packages.rb
```diff
@@ -0,0 +1,86 @@
...
+          .add(:versions,  field: 'version')
+          .connect(self, :filtered_packages)
+
+    private
+
+    # dpkg takes a shell glob, like 'vim*', but we're filtering using
+    # a regular expression. let's make sure that the pattern gets
+    # converted to a regular expression
+    def pattern_regexp(p)
+      shmap = { '*'=>'.*', '?' => '.', '[' => '[', ']' => ']' }
```
alexpop: What's the deal with the "[" and "]" replaces? They don't seem to do much:
```ruby
[9] pry(main)> "vim][".gsub(/(.)/) { |w| shmap[w] || w }
=> "vim]["
```

lib/resources/packages.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    Package = Struct.new(:status, :name, :version)
+
+    def build_package_list(command)
+      cmd = inspec.command(command)
+      all = cmd.stdout.split("\n")[1..-1]
+      return [] if all.nil?
+      all.map do |m|
+        a = m.split
+        a[0] = 'installed' if a[0] =~ /^ii/
```
alexpop: So only `installed` status is expanded, what about these combinations?

```
 Desired action:
    u = Unknown
    i = Install
    h = Hold
    r = Remove
    p = Purge

  Package status:
    n = Not-installed
    c = Config-files
    H = Half-installed
    U = Unpacked
    F = Half-configured
    W = Triggers-awaiting
    t = Triggers-pending
    i = Installed
```

I've added in the output two packages what have the two status characters `rc`

arlimus: Ignore the desired action for now. Let's focus on the actual state of the package. We will keep them separate in the future. I.e.:

```
match(/^([uihrp])([ncHUFWti])\s/)
=> res[1] we will ignore for now, it is part of a future field for desired state
=> res[2] is the status, we support installed right now
```

We will establish `installed` as a shared status across all OS. The others are ok to be os-specific for now. imho implement as lowercase with spaces, e.g. `not installed`. Do not export the other status types for now, just stick to installed + open the others in a RFC. Sync with the package resource.

test/unit/mock/cmd/dpkg-query-W
```diff
@@ -0,0 +1,570 @@
...
+ii  xserver-xorg-video-nouveau 1:1.0.12-1build2
+ii  xserver-xorg-video-qxl 0.1.4-3ubuntu3
+ii  xserver-xorg-video-radeon 1:7.7.0-1
+ii  xserver-xorg-video-vesa 1:2.3.4-1build2
+ii  xserver-xorg-video-vmware 1:13.1.0-2ubuntu3
+ii  xterm 322-1ubuntu1
+ii  xz-utils 5.1.1alpha+20120614-2ubuntu2
+ii  zerofree 1.0.3-1
+ii  zlib1g 1:1.2.8.dfsg-2ubuntu4
+ii  zlib1g-dev 1:1.2.8.dfsg-2ubuntu4
```
arlimus: Let's shorten this just to the necessary lines

arlimus (commented):
Awesome work thank you so much!! 😄 Just 2 tiny comments

arlimus:
Just one more thing: I think we should keep the string as a non-regex only check. i.e. no glob expression. Keep regex'y behavior to real regex input for the `packages` initialization. Kudos again!

jtimberman:
Don't merge yet - I'm going to rebase and rewrite the original commit message to match reality.

jtimberman:
This is ready for review/merge.

alexpop:
👍  for the current version. Thanks for the great work Joshua!

arlimus:
Awesome work and huge kudos for adding it Joshua!!
👍

PR #1477, opened 2017-02-08
Use Chefstyle rather than generic Rubocop

tduffield:
To keep consistency in Chef Ruby projects, use chefstyle for linting.

This has the added benefit of preventing rubocop conflicts between inspec and other projects that depend on it.

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -1,7 +1,6 @@
-# encoding: utf-8
-source 'https://supermarket.chef.io'
+source "https://supermarket.chef.io"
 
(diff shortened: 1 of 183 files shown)


arlimus:
👎 for now. We need to take a long and hard look at the details before this is merged.
Let's have a discussion.

arlimus (changes requested):
I love some of the updates (updating the encoding header wherever not needed anymore on ruby 2+). As noted in the comment above, let's have a discussion about the implications and direction first.

PR #1478, opened 2017-02-08
Use RuboCop 0.39.0 (same as chefstyle)

tduffield:
Avoid a gem conflict with Chef/ChefDK and use Rubocop 0.39.0


arlimus (approved):
Thank you for the update to Rubocop, Kudos @tduffield 👍 😃

PR #1482, opened 2017-02-09
Adding new crontab resource

adamleff:
The crontab resource parses a particular user's crontab file into
individual entries and allows the user to assert information about
each entry as needed.


lib/resources/crontab.rb
```diff
@@ -0,0 +1,81 @@
...
+module Inspec::Resources
+  class Crontab < Inspec.resource(1)
+    name 'crontab'
+    desc 'Use the crontab InSpec audit resource to test the contents of the crontab for a given user which contains information about scheduled tasks owned by that user.'
+    example "
+      describe crontab('root') do
+        its('commands') { should include '/path/to/some/script' }
+      end
+
+      describe crontab('myuser').command('/home/myuser/build.sh') do
```
arlimus: Not sure that the resource supports the singular right now. It works with `commands` but not `command` out of the box. That's a whole other discussion, but lovely to solve :)

arlimus (commented):
Huge contribution, kudos @adamleff !! Includes tests, docs, everything I can ask for.

Just one tiny comment vv (also same situation in docs)

adamleff:
@arlimus this should be good for another review! Fixed the example/docs, and also added a `skip_resource` for non-unix OSs.

arlimus:
Awesome @adamleff  thank you 😄 
You are setting a high bar here

PR #1484, opened 2017-02-10
Skip packages resource for unsupported OS

alexpop:
`raise "packages resource is not yet supported on #{os.name}"` was being hit when `inspec check` was executed on a non debian OS.

The test will be skipped on unsupported OSes if written like this:
```ruby
  describe packages(/bash/) do
    its("names") { should be_empty }
  end
```

But succeed when written like this, hence the `warn` in `filtered_packages`:
```ruby
  describe packages(/bash/).names do
    it { should be_empty }
  end
```

More about this limitation here: https://github.com/chef/inspec/issues/1205#issuecomment-252236255

### DIFF
--- a/lib/resources/packages.rb
+++ b/lib/resources/packages.rb
@@ -23,10 +23,14 @@ module Inspec::Resources
     "
 
     def initialize(pattern)
+      @command = packages_command
+
+      return skip_resource "The packages resource is not yet supported on OS #{inspec.os.name}" unless @command
+
       @pattern = pattern_regexp(pattern)
-      all_pkgs = package_list
+      all_pkgs = build_package_list(@command)
       @list = all_pkgs.find_all do |hm|
-        hm[:name] =~ pattern_regexp(pattern)
+        hm[:name] =~ @pattern
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/resources/packages.rb
```diff
@@ -44,31 +48,29 @@ def to_s
 
     private
 
+    def packages_command
+      os = inspec.os
+      if os.debian?
+        command = "dpkg-query -W -f='${db:Status-Abbrev} ${Package} ${Version}\\n'"
+      end
+      command
```
arlimus, line 56: Could we please add `skip_resource` for all those operating systems where this is not supported?

arlimus (changes requested):
Awesome bugfix Alex, just one more tiny addition ;)

lib/resources/packages.rb
alexpop: I've done it in `initialize`:
`return skip_resource "The packages resource is not yet supported on OS #{inspec.os.name}" unless @command`

arlimus: Ah perfect, I missed that! Thank you 👍

arlimus (approved):

arlimus:
Kudos Alex!!

alexpop:
Code shipped in inspec `1.14.1`

PR #1487, opened 2017-02-13
Packages resource support for RedHat

alexpop:
`processes` support for RedHat is needed to support the new CIS SCAP profiles.

Also fixes dpkg trimming of first line.


test/helper.rb
```diff
@@ -250,8 +250,10 @@ def md.directory?
       'pkginfo -l SUNWzfsr' => cmd.call('pkginfo-l-SUNWzfsr'),
       # solaris 11 package manager
       'pkg info system/file-system/zfs' => cmd.call('pkg-info-system-file-system-zfs'),
-      # dpkg-query package list
+      # dpkg-query all packages
       "dpkg-query -W -f='${db:Status-Abbrev} ${Package} ${Version}\\n'" => cmd.call('dpkg-query-W'),
+      # rpm query all packages
+      "rpm -qa --queryformat '%{NAME}   %{VERSION}-%{RELEASE}\\n'" => cmd.call('rpm-qa-queryformat'),
```
chris-rock: Do we need multiple spaces between name and version? better `'%{NAME} %{VERSION}-%{RELEASE}\\n'`

lib/resources/packages.rb
```diff
@@ -23,12 +23,17 @@ class Packages < Inspec.resource(1)
     "
 
     def initialize(pattern)
-      @command = packages_command
-
-      return skip_resource "The packages resource is not yet supported on OS #{inspec.os.name}" unless @command
+      os = inspec.os
+      if os.debian?
+        @pkgs = Debs.new(inspec)
+      elsif %w{redhat suse amazon fedora}.include?(os[:family])
```
chris-rock: can we use `os.redhat?`? Otherwise this will not work on scientific or oracle linux

chris-rock (changes requested):
Looks great @alexpop Just some minor requests

test/helper.rb
adamleff: I can see why he's doing it... but is there another character that is guaranteed to not be part of a RPM package name and version string that we can use as the delimiter instead? If not, I'm fine with the two spaces since that character combo should never be in a name or version.

adamleff (changes requested):
Looks fine to me. I echo the feedback from @chris-rock as well.

lib/resources/packages.rb
alexpop: I used the same checks as the existing `package` resource:
```ruby
      elsif %w{redhat suse amazon fedora}.include?(os[:family])
        @pkgman = Rpm.new(inspec)
```

Noticed that `os.redhat?` is not matching Fedora:
```ruby
inspec> os.redhat?
=> false
inspec> os.params
=> {:name=>"fedora", :family=>"fedora", :release=>"25", :arch=>"x86_64"}
```

If we go with `os.redhat?`, will need something like:
```
lib/resources/grub_conf.rb
35:    if os.redhat? || os[:name] == 'fedora'
```

test/helper.rb
alexpop: Exactly, I played the paranoid here to avoid splitting on a space in package name or version. I'm fine with two spaces as well

lib/resources/packages.rb
chris-rock: Thank you @alexpop yes `os.redhat? || os[:name] == 'fedora'` sounds right and we should apply this to the `package` resource as well.

test/helper.rb
chris-rock: I am okay with taking that approach right now. We should document it in code to make other developers aware of the intended use of the whitespace

lib/resources/packages.rb
alexpop: What about these?
```ruby
inspec> os.redhat?
=> false
inspec> os.params
=> {:name=>"amazon", :family=>"amazon", :release=>"2016.09", :arch=>"x86_64"}
```

```ruby
inspec> os.redhat?
=> false
inspec> os.params
=> {:name=>"suse", :family=>"suse", :release=>"11.4", :arch=>"x86_64"}
```

chris-rock: As far as I know, Suse is using YAST and zypper. Lets add support for those systems seperately.

alexpop: Works for amazon linux:
```bash
[root@ip-172-31-12-168 ~]# rpm -qa --queryformat '%{NAME}  %{VERSION}-%{RELEASE}\n'
sysvinit  2.87-6.dsf.15.amzn1
kbd-misc  1.15-11.4.amzn1
rubygem20-json  1.8.3-1.51.amzn1
glibc-common  2.17-106.168.amzn1
```

and SUSE Linux Enterprise Server 12:
```bash
ip-172-31-10-176:~ # rpm -qa --queryformat '%{NAME}  %{VERSION}-%{RELEASE}\n'
nfs-client  1.3.0-26.3
sle-module-legacy-release  12-6.2
python-oauthlib  0.7.2-2.3
libwbclient0  4.4.2-29.4
procps  3.3.9-7.1
python-colorama  0.2.7-1.9
```

chris-rock: @alexpop I agree with you. Lets add support for those systems, too

alexpop: in this case, do I make the condition:
```ruby
if os.redhat? || %w{suse amazon fedora}.include?(os[:family])
```

chris-rock: Sounds resonable

adamleff: I'd love it if we could align a bit more with Chef/Ohai in terms of platform / os / platform_family so it's not confusing for our users. Not something to address in this PR, but perhaps something we can consider for a future enhancement. Curious as to your thoughts.

alexpop:
Gents, updated and rebased based on the discussions above.

chris-rock (approved):

adamleff (approved):

PR #1489, opened 2017-02-13
Doc fix for SourceReaders::InspecReader

adamleff:
The inline docs for SourceReaders::InspecReader#new state that it takes
a SourceReader object for the target... but we're trying to create the
SourceReader object! It actually takes a FileProvider object that is
capabile of listing files for the given profile and reading them.


chris-rock (approved):
Thank you @adamleff

adamleff:
Going to merge this since it's a doc-only change.

PR #1501, opened 2017-02-21
Add FreeBSD support for ZFS datasets and pools

jbenden:
The following new resources have been added; however, they presently only support FreeBSD and similar.

* `zfs_dataset`: tests if a named ZFS dataset is present and/or has certain properties.
* `zfs_pool`: tests if a named ZFS pool is present and/or has certain properties.

The `mount` resource has been refactored to include support for FreeBSD too.


lib/resources/mount.rb
```diff
@@ -49,12 +49,38 @@ def method_missing(name)
...
+
+  class MountsInfo
+    attr_reader :inspec
+    def initialize(inspec)
+      @inspec = inspec
+    end
+  end
+
+  class LinuxMounts < MountsInfo
+    include MountParser
```
adamleff: This is probably a good time to rename this to `LinuxMountParser` to avoid any confusion.

lib/resources/mount.rb
```diff
@@ -49,12 +49,38 @@ def method_missing(name)
...
+      end
+    end
+  end
+
+  class MountsInfo
+    attr_reader :inspec
+    def initialize(inspec)
+      @inspec = inspec
+    end
+  end
```
adamleff: Should we add a default `#parse_mount_options` method here that `raise`s an exception to make it clear for future users that the method is required?

lib/resources/zfs_dataset.rb
```diff
@@ -0,0 +1,62 @@
...
+    "
+    example "
+      describe zfs_dataset('tank/tmp') do
+        its('exec') { should eq('off') }
+        its('setuid') { should eq('off') }
+      end
+    "
+
+    def initialize(zfs_dataset)
+      return skip_resource 'The `zfs_dataset` resource is not supported on your OS yet.' if !inspec.os.bsd?
```
adamleff, line 19: We don't need to account for it in this PR, but I suspect this resource would work OK as-is on Solaris and any Linux that is using ZFS-on-Linux too, with the small exception of the actual ZFS binary paths may be different. This is worth some investigation.

adamleff (changes requested):
@jbenden Thanks for your contribution! I'm a huge fan of ZFS... used it a lot on OmniOS at my last job... so adding ZFS support fills me with glee. :)

I added a few comments, and we should add some unit tests to this with some mocked output for all the items changed/added (i.e. BSD mounts, ZFS pools, and ZFS datasets). I'd be happy to help you navigate that process if you need any assistance.

jbenden:
Hello Adam,

I have made all requested modifications; with the exception of supporting Solaris and Linux ZFS. If you have access to either of these operating systems (especially Solaris) I simply need the binary path locations. I am positive this would work with Solaris, but am not sure where the required binaries are... I did see that Linux places them in `/usr/sbin`, but do not know if the output is exactly the same...

lib/resources/zfs_dataset.rb
chris-rock: Would love to see the Solaris support for this as well!

adamleff (approved):

adamleff:
@chris-rock I think we should accept this as-is and open an enhancement issue for the backlog to extend this to Solaris-variants and even perhaps Linux-on-ZFS users. But since the contributor doesn't have access to a Solaris-based machine, it seems reasonable to accept as-is. Thoughts?

chris-rock:
I agree with you @adamleff

chris-rock (approved):

adamleff:
@jbenden thanks again for this contribution!

Issue #1508, opened 2017-02-24
Colours and symbols broken on Windows


trickyearlobe:
### Description

* Passing inspec tests display white on a red backround instead of green on black.
* Failing inspec tests display white on black instead of red on black
* Fancy UTF-8 checkmarks show as ? or box character (depends which terminal used)

### InSpec and Platform Version

* Inspec 1.11.0, 1.14.1
* Windows 2012R2, Windows 2008R2, Windows 10
* Powershell, ConEmu

### Replication Case

Make some sample tests with passing and failing cases
Run ```inspec exec <path to tests>```

### Possible Solutions

Pull request will follow which detects OS and uses:-

* Basic ANSI colors instead of extended colors
* Standard [PASS] [FAIL] messages instead of fancy UTF8 checkmarks.

adamleff:
Closing, fixed in #1510.

PR #1509, opened 2017-02-24
Fix formatting and colors on Windows

trickyearlobe:
Fixes issue #1508
 
* Windows terminals don't support extended ANSI colours. Use basic + intensity
* Windows terminals don't support UTF-8 well so don't use special characters

Other OS'es get what they had before.

Signed-off-by: Richard Nixon \<richard.nixon@btinternet.com\>


trickyearlobe:
Messed up the signed off by

PR #1510, opened 2017-02-24
Fix formatting and colors on Windows

trickyearlobe:
Fixes issue #1508

* Windows terminals don't support extended ANSI colours. Use basic + intensity
* Windows terminals don't support UTF-8 well so don't use special characters

Other OS'es get what they had before.

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -238,27 +238,60 @@ end
 class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
   RSpec::Core::Formatters.register self, :close
 
-  COLORS = {
-    'critical' => "\033[38;5;9m",
-    'major'    => "\033[38;5;208m",
-    'minor'    => "\033[0;36m",
-    'failed'   => "\033[38;5;9m",
-    'passed'   => "\033[38;5;41m",
-    'skipped'  => "\033[38;5;247m",
-    'reset'    => "\033[0m",
-  }.freeze
+  case RUBY_PLATFORM
+  when /windows|mswin|msys|mingw|cygwin/
 
-  INDICATORS = {
-    'critical' => '  ×  ',
-    'major'    => '  ∅  ',
-    'minor'    => '  ⊚  ',
-    'failed'   => '  ×  ',
-    'skipped'  => '  ↺  ',
-    'passed'   => '  ✔  ',
-    'unknown'  => '  ?  ',
-    'empty'    => '     ',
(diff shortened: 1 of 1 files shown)


lib/inspec/rspec_json_formatter.rb
```diff
@@ -238,27 +238,60 @@ def format_example(example)
...
-    'minor'    => '  ⊚  ',
-    'failed'   => '  ×  ',
-    'skipped'  => '  ↺  ',
-    'passed'   => '  ✔  ',
-    'unknown'  => '  ?  ',
-    'empty'    => '     ',
-    'small'    => '   ',
-  }.freeze
+  case RUBY_PLATFORM
+  when /windows|mswin|msys|mingw|cygwin/
```
chris-rock, line 242: We should find a solution that is not dependent on OS detection, since there are terminal available for windows that support UTF8 like http://cmder.net/
I am okay to merge it anyway since this is solving the problem for most Windows users.

chris-rock (approved):
Great finding @trickyearlobe Thank you for raising the issue and proposing a fix.

lib/inspec/rspec_json_formatter.rb
trickyearlobe: My customer prefers to use Powershell. CMDER does support UTF8 but the symbols don't render and it still has the extended colors problem. The symbols may just be down to selected font or codepage (have to execute ```chcp 65001``` to get UTF8 on most machines)

trickyearlobe: BTW... I love CMDER... its my go-to terminal when I have to use Windows :)

chris-rock: Thank you @trickyearlobe for the explanation.

trickyearlobe: Doing a bit more digging, it seems that for the UTF8 symbols to display, you need to switch to UTF8 with ```chcp 65001``` and then change the terminal font to one that includes emojis (such as Segoe UI Symbol).

I also did a bit of digging around autodetection of terminal type or color support. ANSI does define some queries but most of the Windows terminals ignore them (despite the sequences being documented on MSDN)

See this link https://en.wikipedia.org/wiki/Emoji#Microsoft_Windows

chris-rock: @trickyearlobe Thank you for the deeper investigation. I agree with you that the os detection is a safe bet.

arlimus (approved):
Great work, kudos for all the investigation and fixing it Richard!!

Issue #1507, opened 2017-02-23
kernel_module doesn't work in CentOS 7.2


aladmit:
### Description

Kernel_module doesn't work in CentOS 7.2

### InSpec and Platform Version

InSpec 1.11.0
ChefDK 1.2.22
My OS: macOS Sierra 10.12.3
Server's OS: CentOS 7.2

### Replication Case

```
describe kernel_module('ip_tables') do
  it { should be_loaded }
end
```

### Possible Solutions

`command('lsmod').stdout` return empty string, but `command('/sbin/lsmod')` return normal output. So, I think we should remove condition about [`inspec.os[:release]`](https://github.com/chef/inspec/blob/master/lib/resources/kernel_module.rb#L27) from kernel_module.

### Stacktrace

```
inspec> describe kernel_module('ip_tables') do
inspec>   it { should be_loaded }
inspec> end

Profile: inspec-shell
Version: unknown

  Kernel Module
     ∅  ip_tables should be loaded
     expected `Kernel Module ip_tables.loaded?` to return true, got false
```

aladmit:
Yeah, my solution is work, I'll send PR.

adamleff:
Fixed in #1513, and there's a pending PR #1798 that cleans this up even further.

PR #1513, opened 2017-02-24
Fix kernel_module for centos/redhat

aladmit:
Hi! I fixed kernel_module resource for centos/redhat.

issue: #1507

### DIFF
--- a/lib/resources/kernel_module.rb
+++ b/lib/resources/kernel_module.rb
@@ -21,7 +21,7 @@ module Inspec::Resources
     end
 
     def loaded?
-      if inspec.os[:family] == 'redhat'
+      if inspec.os.family == 'redhat'
         lsmod_cmd = '/sbin/lsmod'
       else
         lsmod_cmd = 'lsmod'
@@ -38,7 +38,7 @@ module Inspec::Resources
     end
 
     def version
-      if inspec.os[:family] == 'redhat'
+      if inspec.os.family == 'redhat'
         modinfo_cmd = "/sbin/modinfo -F version #{@module}"
       else
         modinfo_cmd = "modinfo -F version #{@module}"
--- a/lib/resources/kernel_module.rb
+++ b/lib/resources/kernel_module.rb
@@ -21,10 +21,11 @@ module Inspec::Resources
     end
 
     def loaded?
-      # default lsmod command
-      lsmod_cmd = 'lsmod'
-      # special care for CentOS 5 and sudo
-      lsmod_cmd = '/sbin/lsmod' if inspec.os[:name] == 'centos' && inspec.os[:release].to_i == 5
+      if inspec.os[:family] == 'redhat'
+        lsmod_cmd = '/sbin/lsmod'
+      else
+        lsmod_cmd = 'lsmod'
+      end
 
       # get list of all modules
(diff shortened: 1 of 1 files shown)


lib/resources/kernel_module.rb
```diff
@@ -21,10 +21,11 @@ def initialize(modulename = nil)
     end
 
     def loaded?
-      # default lsmod command
-      lsmod_cmd = 'lsmod'
-      # special care for CentOS 5 and sudo
-      lsmod_cmd = '/sbin/lsmod' if inspec.os[:name] == 'centos' && inspec.os[:release].to_i == 5
+      if inspec.os[:family] == 'redhat'
```
chris-rock: We should try to avoid using `inspec.os[:family] `. A better approach is to use `inspec.os.redhat?`

chris-rock (changes requested):
Thank you @postgred for bringing up the change! Have you tested this on RedHat 6 and RedHat 7?

lib/resources/kernel_module.rb
```diff
@@ -37,7 +38,7 @@ def loaded?
     end
 
     def version
-      if inspec.os[:name] == 'centos' && inspec.os[:release].to_i == 5
+      if inspec.os[:family] == 'redhat'
```
Fodoj: This would resolve to `true` for any version of any rhel-based distro, so high chance it will either break for some of existing versions or it will break for some of future versions. I would suggest to keep this check, but only for distro versions in trouble.

chris-rock: @Fodoj thank you for highlighting this issue. I assume this applies to RedHat 5 as well, therefore we may need to ask for
```
if inspec.os.redhat? && inspec.os[:release].to_i == 5
```

aladmit:
@chris-rock Yeah, I have tested this on CentOS5-7 and RedHat6-7

lib/resources/kernel_module.rb
```diff
@@ -37,7 +38,7 @@ def loaded?
     end
 
     def version
-      if inspec.os[:name] == 'centos' && inspec.os[:release].to_i == 5
+      if inspec.os.family == 'redhat'
```
chris-rock: @postgred it is better to use `inspec.os.redhat?`. we try to avoid `inspec.os.family`. This is inherited from Serverspec. We are going to prefer the name or the helper method.

chris-rock (changes requested):

lib/resources/kernel_module.rb
aladmit: Yes, If you don't want use `os.family`, we can use 'os.redhat? || os.name == 'fedora'` (I have found the problem on fedora too).

chris-rock: @postgred Yes, that would be appreciated.

aladmit:
@chris-rock I have removed os.family. Should I do something else?

chris-rock (approved):
Thank you @postgred

chris-rock:
@Fodoj do you see any issues left?

Fodoj:
lgtm

chris-rock:
Thank you @postgred

adamleff:
Looks great. Thanks @postgred!

PR #1516, opened 2017-02-26
Fix problem with `kitchen-inspec` link on website

grv87:
Link to `kitchen-inspec` from header navigation bar on website doesn't work. Link from the left sidebar works.
I think this PR should fix it.


adamleff:
@grv87 thank you for your contribution! However, I'm afraid I'm having trouble reproducing or understanding the issue you're trying to correct.

Could you elaborate a bit more, and perhaps provide a screenshot or two indicating where the problem is?

chris-rock:
I was able to reproduce the issue with the burger menu on small screen sizes screen shot 2017-02-28 at 12 23 32 pm

grv87:
Maybe it's actually the burger menu. With scripts turned off I see it as header navigation bar.
This link doesn't work (note link's target in the left bottom corner):
image 148

Compare it to other menu and other link - all these work now:
image 149
image 150
image 151

chris-rock:
cc @hannah-radish

chris-rock:
@grv87 I'like the change to `docs/plugin_kitchen_inspec.md`. In this case we need to rename https://github.com/chef/inspec/blob/master/www/data/docs_sidebar.yml#L27-L28 as well

grv87:
@chris-rock I see that this is not necessary. Other source files are `*.md` but listed in `docs_sidebar.yml` as `*.html`. Jekyll (or other generator whatever you use) substitutes one for another.

adamleff (approved):
This is WAY overdue, but @grv87 thank you for your contribution! This looks to be a fine fix and I'll get this merged.

PR #1531, opened 2017-03-02
Updating .gitignore for Habitat and direnv

adamleff:
For direnv users, excluding the .direnv and .envrc entries will
help avoid any unnecessary local environment settings from getting
committed.

Excluding the results directory will avoid any Habitat artifacts
from getting unnecessarily committed.


chris-rock (approved):
Thank you @adamleff

PR #1533, opened 2017-03-02
Write Habitat-driven InSpec output to svc_var directory

adamleff:
When running a InSpec profile built with Habitat, we now
write the formatter/reporter data to a JSON file in the
pkg.svc_var_path rather than STDOUT. This will allow for
programmatic collection of this data and future enhancements
to allow this data to be passed around a Habitat ring.

This also corrects an issue creating a Habitat profile if the
profile had never been in the local InSpec cache. By setting a
mock Backend when creating the profile object, similarly to what
the archivers do, this issue is avoided.

### DIFF
--- a/lib/bundles/inspec-habitat/profile.rb
+++ b/lib/bundles/inspec-habitat/profile.rb
@@ -79,7 +79,7 @@ module Habitat
     private
 
     def create_profile_object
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
This is a great improvement @adamleff

PR #1536, opened 2017-03-03
Fixing a typo

tescalada:
changing paht to be path.

### DIFF
--- a/docs/resources/json.md.erb
+++ b/docs/resources/json.md.erb
@@ -23,7 +23,7 @@ A `json` resource block declares the data to be tested. Assume the following JSO
 
 This file can be queried using:
 
-    describe json('/paht/to/name.json') do
+    describe json('/path/to/name.json') do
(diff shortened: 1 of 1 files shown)


adamleff:
@tescalada Thank you so much for this fix!

Our contribution policy requires a DCO sign-off on all contributions (see https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco for more information). However, a change like this would easily fall under the "obvious fix" rule if you do not wish to sign-off on your commit. I simply need you to invoke the Obvious Fix rule by saying so in your commit message.

Thank you!

adamleff (approved):
Since the DCO does not apply to the chef-web-docs repo, and this is only a documentation update, I am going to accept this PR as-is. I will also be putting forth a Chef RFC/procedure change exempting in-repo docs-only changes (such as what we do in the chef/inspec repo) from DCO requirements.

PR #1538, opened 2017-03-05
Avoid connection timeout of "inspec version"

makotots:
Fixes https://github.com/chef/inspec/issues/1537 
Signed-off-by: Makoto Nozaki <makoto.nozaki@twosigma.com>

### DIFF
--- a/lib/utils/latest_version.rb
+++ b/lib/utils/latest_version.rb
@@ -9,8 +9,8 @@ class LatestInSpecVersion
   def latest
     uri = URI('https://rubygems.org/api/v1/gems/inspec.json')
     res = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https',
-                          open_timeout: 0.5) {|http|
-      http.read_timeout = 0.5
+                          open_timeout: 0.5, read_timeout: 0.5
+                         ) {|http|
       http.get(uri.path)
     }
     inspec_info = JSON.parse(res.body)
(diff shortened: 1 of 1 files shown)


makotots:
Note: The next line, `http.read_timeout = 0.5` could be moved up to the same level for consistency.

lib/utils/latest_version.rb
```diff
@@ -8,7 +8,8 @@ class LatestInSpecVersion
   # fetches the latest version from rubygems server
   def latest
     uri = URI('https://rubygems.org/api/v1/gems/inspec.json')
-    res = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') {|http|
+    res = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https',
+                          open_timeout: 0.5) {|http|
       http.read_timeout = 0.5
```
adamleff: I agree with your comment that this should be moved up to the `#start` opts for consistency. Thank you for suggesting it!

adamleff (changes requested):
@makotots thank you so much for this PR! I agree that keeping the definition of the options at the same level for consistency is the right thing to do. Would you mind pushing up a new commit with that change? Then I'll be happy to get this approved.

chris-rock (approved):
Thank you @makotots and @adamleff This is a great improvement.

adamleff (approved):

adamleff:
Great work, @makotots! Thank you for your contribution!

PR #1539, opened 2017-03-06
Fix omnibus build after new JUnit formatter

adamleff:
The new JUnit formatter requires nokogiri, so we need
to build nokogiri via omnibus to ensure liblzma, etc.
is built as part of the omnibus package instead of
`gem` picking up a system liblzma, such as on Mac OS X.

Also bumping ruby to 2.3.1 to match ChefDK.


chris-rock (approved):
Thank you @adamleff

PR #1544, opened 2017-03-06
Provide a method-based accessor for SimpleConfig hashes

### LINKED ISSUE #1541: inspec outputs full hash when testing ini style file option with [header]
### Description

When testing config in a mysql conf file with a `[mysqld]` header using the `mysql_conf` resource, the entire hash is returned in the test result instead of just the config option being tested.  This could affect other INI style files too.

Instead of seeing:

```
  {"log_bin"=>"/var/log/mysql/mariadb-bin", "sync_binlog"=>"0"}
     ✔  ["sync_binlog"] should eq "0"
```

I would expect to see this (as you would without the ini style section header):

```
  MySQL Configuration
     ✔  sync_binlog should eq nil
```

inspec 1.16.1 on CentOS Linux release 7.3.1611 (Core)

### Replication Case

Create a test file (I used `/tmp/my.cnf`):

```
[mysqld]
log_bin = /var/log/mysql/mariadb-bin
sync_binlog = 0
```

Using InSpec shell:

```
inspec> describe mysql_conf('/tmp/my.cnf').mysqld do its(['sync_binlog']) { should eq '0' } end

Profile: inspec-shell
Version: (not specified)

  {"log_bin"=>"/var/log/mysql/mariadb-bin", "sync_binlog"=>"0"}
     ✔  ["sync_binlog"] should eq "0"

```

adamleff:
When SimpleConfig parses a config file that has sections, such as a mysqld
config file, the values within that section are returned via a Hash. However,
we do not provide an easy way to write tests for those deep hash values:

```
describe mysql_conf('/tmp/my.cnf') do
  its('mysqld.expire_logs_days') { should cmp 10 }
end

  MySQL Configuration
     ∅  undefined method `expire_logs_days' for #<Hash:0x007fe463795a00>
```

This change provides a method-based accessor for Hashes that are built via
SimpleConfig.

```
describe mysql_conf('/tmp/my.cnf') do
  its('mysqld.expire_logs_days') { should cmp 10 }
end

  MySQL Configuration
     ✔  mysqld.expire_logs_days should cmp == 10
```

Fixes #1541 by changing the way the attributes are fetched.

### DIFF
--- a/lib/utils/simpleconfig.rb
+++ b/lib/utils/simpleconfig.rb
@@ -73,7 +73,13 @@ class SimpleConfig
     m = opts[:group_re].match(line)
     return nil if m.nil?
     @groups.push(m[1])
-    @vals = @params[m[1]] = {}
+
+    # We use a Hashie::Mash to provide method syntax for retrieving
+    # values. For configs that have keys whose values may be hashes
+    # (such as a mysql config that has [sections]), providing
+    # a hash whose values are accessible via methods provide
+    # a better user experience with `its` blocks.
+    @vals = @params[m[1]] = Hashie::Mash.new
   end
 
   def parse_implicit_assignment_line(line, opts)
--- a/test/unit/utils/simpleconfig_test.rb
+++ b/test/unit/utils/simpleconfig_test.rb
@@ -77,4 +77,11 @@ describe 'SimpleConfig Default Parser' do
     cur.params.must_equal(res)
     cur.groups.must_equal(['g', 'k'])
   end
+
+  it 'provides methods to access returned hashes' do
+    cur = SimpleConfig.new("[section1]\nkey1 = value1\n\n[section2]\nkey2 = value2\n")
+    cur.params['section1'].key1.must_equal('value1')
+    cur.params['section2'].key2.must_equal('value2')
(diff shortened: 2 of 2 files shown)


lib/utils/simpleconfig.rb
```diff
@@ -7,6 +7,16 @@
 require 'utils/parser'
 
 class SimpleConfig
+  # for configs that have keys whose values may be hashes
+  # (such as a mysql config that has [sections]), provide
+  # a hash whose values are accessible via methods to provide
+  # a good user experience with `its` blocks.
+  class ConfigHash < Hash
```
chris-rock: This is definitely a valid solution. We use [hashie](https://github.com/intridea/hashie#mash) in other areas like https://github.com/chef/inspec/blob/546486ff6afc371103943501e9d7e15bcc6941ff/lib/resources/service.rb#L180-L183. Should we stick to the same solution?

adamleff: Hashie works for me too - updating now...

adamleff:
@chris-rock do you approve of this one now that I switched to using `hashie` instead?

chris-rock (approved):
Thank you for this improvement @adamleff

PR #1551, opened 2017-03-08
WIP: support postgres 9.4 service on rhel

chris-rock:
Addresses an error detected by @atomic111 in https://github.com/dev-sec/postgres-baseline/issues/17 where `postgres.service` is not returning the right service name for RHEL systems in combination with Postgres 9.4

### DIFF
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -10,6 +10,8 @@ module Inspec::Resources
 
     attr_reader :service, :data_dir, :conf_dir, :conf_path
     def initialize
+      @service = 'postgresql'
+
       os = inspec.os
       if os.debian?
         #
@@ -41,6 +43,7 @@ module Inspec::Resources
                   end
 
         @data_dir = File.join('/var/lib/pgsql/', version.to_s, 'data')
+        @service = 'postgresql-' + version if version == '9.4'
       elsif os[:name] == 'arch'
         #
         # https://wiki.archlinux.org/index.php/PostgreSQL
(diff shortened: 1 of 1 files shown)


lib/resources/postgres.rb
```diff
@@ -41,6 +43,7 @@ def initialize
                   end
 
         @data_dir = File.join('/var/lib/pgsql/', version.to_s, 'data')
+        @service = 'postgresql-' + version if version == '9.4'
```
adamleff, line 46: Does this only happen with 9.4, or does it happen with 9.5 and 9.6 as well? Do we perhaps need to change this to:

```
@service = 'postgresql-' + version if version.to_f >= 9.4
```

adamleff (commented):
Left a question regarding version treatment. Also, I think we should add a test for this ensuring that we properly set the service. Since `service` is already an `attr_reader`, this should be pretty easy.

lib/resources/postgres.rb
chris-rock: Good point @adamleff I try to find that out with https://github.com/sous-chefs

aaronlippold: I believe @adamleff  is correct

sharmay:
Changes for service name with version number look correct way to get the service name.

aaronlippold:
@chris-rock This PR can be closed, it was merged into #1758

aaronlippold:
@adamleff I think this can be closed.

PR #1559, opened 2017-03-14
Add additional example for matching crontab commands

### LINKED ISSUE #1526: Crontab regex matching
### Description

The crontab resource is unable to apply Regex matching via the `to match` matcher. It would be nice if this were a feature, since I prefer to match a part of the command instead of the whole command in my test. This is especially useful if there are parts of a command that don't care too much about to verify working behaviour, such as paths to a command that are dynamically created by factors including programming language versions.

For example, suppose that I am adding a `cron` in my cookbook for running a custom rake task nightly. The rake task could be something like `custom_task`, so I just want my test to ensure that there is a cron command that matches to the regex `/rake custom_task/`.

```ruby
describe corntab do
  its ('commands') { should match /rake custom_task/ }
end
```

Unfortunately, it looks like I can only do an equality check via `include` instead of a regex check with `match`. So I have to add some hardcoded paths to my test code, which makes me cringe slightly.

```ruby
describe corntab do
  its ('commands') { should include '/home/deploy/ruby-2.3.1/bin/rake custom_task' }
end
```

Inspec version 1.15.0
MacOS 10.12.3 (host machine where I am running tests from)

### Replication Case

See *Description*

### Possible Solutions

My workaround is to bite my tongue and just use the `include` matcher. I do not think I know the best way to write a PR for this. But I imagine that new code should allow for `match` matcher support in `lib/resources/crontab.rb`.

adamleff:
As raised in #1526, adding an additional example showing how
a user can use the `where` accessor to find commands matching
a pattern and write a test using the results.

Fixes #1526

### DIFF
--- a/docs/resources/crontab.md.erb
+++ b/docs/resources/crontab.md.erb
@@ -60,3 +60,9 @@ The following examples show how to use this InSpec audit resource.
     describe crontab.where({'hour' => '*', 'minute' => '*'}) do
       its('entries.length') { should cmp '0' }
     end
+
+### Test that the logged-in user's crontab contains a single command that matches a mattern
+
+    describe crontab.where { command =~ /a partial command string/ } do
+      its('entries.length') { should cmp 1 }
+    end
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @adamleff

PR #1561, opened 2017-03-14
Fix ObjectTraverser when accessing array values

### LINKED ISSUE #1560: json resource array index access not working
### Description

Currently trying to use the json resource. Array access via index is not working as described in the documentation.

I got the following json in test.json
```
{
  "hash":{
    "subhash":{
      "key": "value"
    }
  }
  "array": [
    "zero",
    "one",
    "three"
  ]
}
```
and the following control:
```
control 'test' do
  impact 1.0
  title 'test'
  desc '
   Test
  '

  describe file('/root/test.json') do
    it { should be_file }
  end

  describe json('/root/test.json') do
          its(['hash','subhash','key']) {should eq "value"}
          its(['array']) { should include "zero" }
          its(['array',0]) { should eq "zero" }
  end
end
```
And i get the following output:
```
×  test: Test (1 failed)
     ✔  File /root/test.json should be file
     ✔  Json /root/test.json ["hash", "subhash", "key"] should eq "value"
     ✔  Json /root/test.json ["array"] should include "zero"
     ×  Json /root/test.json ["array", 0] should eq "zero"

     expected: "zero"
          got: [nil, nil, nil]

     (compared using ==)

     Diff:
     @@ -1,2 +1,2 @@
     -"zero"
     +[nil, nil, nil]
```
Hash access and access to the array node itself is working as expected but indexed access to the array is not working, only returning nil objects.

inspec version 1.16.1

adamleff:
When attempting to access array values via the `json` resource:

```
describe json('/tmp/test.json') do
      its(['array',0]) { should eq "zero" }
end
```

... the resulting data would be an array of the size of the original array
with all the values replaced with nils:

```
     expected: "zero"
          got: [nil, nil, nil]
```

This was due to a bug in the ObjectTraverser mixin that mapped array values
back through `extract_value` rather than properly handling the passed-in
key(s). This worked fine for the specific data format created by the `csv`
resource but did not work `json` or any other resource that subclassed the
`JsonConfig` resource.

This change fixes the logic when dealing with an array when it's encountered,
and fixes up the `csv` resource with its own `value` method.

This change also adds tests for ObjectTraverser.

Fixes #1560

### DIFF
--- a/test/unit/resources/csv_test.rb
+++ b/test/unit/resources/csv_test.rb
@@ -17,7 +17,7 @@ describe 'Inspec::Resources::CSV' do
     end
 
     it 'gets all value lines' do
-      _(resource.params.length).must_equal 3
+      _(resource.params.length).must_equal 4
     end
 
     it 'captures a hashmap of entries of a line' do
@@ -25,11 +25,15 @@ describe 'Inspec::Resources::CSV' do
     end
 
     it 'gets params by header fields' do
-      _(resource.params[0]['addressable']).must_equal 'ast'
+      _(resource.params[0]['name']).must_equal 'addressable'
     end
 
     it 'retrieves nil if a param is missing' do
       _(resource.params[0]['missing']).must_be_nil
+    end
+
+    it 'returns an array of values by column name' do
+      _(resource.value(['name'])).must_equal([ 'addressable', 'ast', 'astrolabe', 'berkshelf' ])
     end
   end
 end
--- a/lib/resources/csv.rb
+++ b/lib/resources/csv.rb
@@ -15,17 +15,34 @@ module Inspec::Resources
       end
     "
 
-    # override file load and parse hash from csv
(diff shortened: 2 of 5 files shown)


test/unit/resources/csv_test.rb
```diff
@@ -17,19 +17,23 @@
...
+      _(resource.params.length).must_equal 4
     end
 
     it 'captures a hashmap of entries of a line' do
       _(resource.params[0]).must_be_kind_of Hash
     end
 
     it 'gets params by header fields' do
-      _(resource.params[0]['addressable']).must_equal 'ast'
+      _(resource.params[0]['name']).must_equal 'addressable'
```
chris-rock, line 28: man, who wrote this in the first place :-)

chris-rock (approved):
Thank you @adamleff

chris-rock:
giphy 3

PR #1563, opened 2017-03-14
Hide Event Feature on Homepage

hannah-radish:
Darn that snowstorm. Event is cancelled!

Signed-off-by: Hannah Maddy <hmaddy@chef.io>


adamleff (approved):
Thank you, @hannah-radish!

PR #1564, opened 2017-03-14
implement JSON schema for `inspec exec` json outputs

### LINKED ISSUE #884: Add JSON Schema validation
### Description

It is critical that the `inspec exec /path/to/profile` returns stable output, since its a API that is consumed in other applications. The current test coverage can be improved.

0.28.0
### Possible Solutions

Develop a JSON Schema for each operation and validate output against it

arlimus:
This is a current draft to implement JSON schema verification for InSpec `json` and `json-min` formatters. 

It would also have prevented this https://github.com/chef/inspec/issues/884#issuecomment-283024142 😄 

Fixes #884 

This is a low-priority item, no rush.


chris-rock (commented):
This is a huge improvement. Thank you @arlimus. I really like that we have a clear specification now. The only thing that I was confused with is the location of the files in `lib/utils/schema`. Do we intent do use those in core inspec as well or are they for testing purposes only? If the latter, should we move the directory to `test/unit/schema`?

lib/utils/schema/schema.rb
```diff
@@ -0,0 +1,165 @@
...
+  'additionalProperties' => false,
+  'properties' => {
+    'profiles' => {
+      'type' => 'array',
+      'items' => profile,
+    },
+    'statistics' => statistics,
+    'version' => { 'type' => 'string' },
+
+    # DEPRECATED PROPERTIES!! These will be removed with the next major version bump
```
chris-rock: What would be a good way do communicate this?

arlimus: We could put it into the output of the generator

arlimus:
I don't like the location either; `lib/utils` felt like the least objectionable; I'd prefer not to go into `test/*`, as this is a verification layer that will be used outside of inspec tests. Traditionally i'd put it into an `extra` folder somewhere, but didn't want to bloat our file structure yet... (but it might be a better idea)

chris-rock:
Just discussed with @arlimus offline and we came to the conclusion that the current location is the best one, since we want to use the schema to verify reports too in future.

chris-rock (approved):

lib/utils/schema/inspec.exec.full.json
```diff
@@ -0,0 +1 @@
...id":{"type":"string"},"title":{"type":["string","null"]},"desc":{"type":["string","null"]},"impact":{"type":"number"},"refs":{"type":"array","items":{"type":"object","additionalProperties":false,"properties":{"ref":{"type":"string"},"uri":{"type":"string","optional":true},"url":{"type":"string","optional":true}}}},"tags":{"type":"object"},"code":{"type":"string"},"source_location":{"type":"object","properties":{"ref":{"type":"string"},"line":{"type":"number"}}},"results":{"type":"array","items":{"type":"object","additionalProperties":false,"properties":{"status":{"type":"string"},"code_desc":{"type":"string"},"run_time":{"type":"number"},"start_time":{"type":"string"},"skip_message":{"type":"string","optional":true},"resource":{"type":"string","optional":true}}}}}}},"groups":{"type":"array","items":{"type":"object","additionalProperties":false,"properties":{"id":{"type":"string"},"title":{"type":"string","optional":true},"controls":{"type":"array","items":{"type":"string"}}}}},"attributes":{"type":"array"}}}},"statistics":{"type":"object","additionalProperties":false,"properties":{"duration":{"type":"number"}}},"version":{"type":"string"},"controls":"array","other_checks":"array"}}
```
adamleff: Is there any harm in prettyifying this to make it easier for people to see? I realize the output out of the tool isn't "pretty" (and perhaps we should change that) but for purposes of comprehension, it'd be good to see this formatted nicely for humans.

lib/utils/schema/inspec.exec.min.json
```diff
@@ -0,0 +1 @@
+{"type":"object","additionalProperties":false,"properties":{"statistics":{"type":"object","additionalProperties":false,"properties":{"duration":{"type":"number"}}},"version":{"type":"string"},"controls":{"type":"array","items":{"type":"object","additionalProperties":false,"properties":{"id":{"type":"string"},"profile_id":{"type":["string","null"]},"status":{"type":"string"},"code_desc":{"type":"string"},"skip_message":{"type":"string","optional":true},"resource":{"type":"string","optional":true}}}}}}
```
adamleff: Same comment as inspec.exec.full.json - would love to see this structured so humans can grok it.

lib/utils/schema/schema.rb
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env ruby
```
adamleff: For safety/paranoia purposes, should we wrap these methods in a module? I'd hate for something to accidentally `require` this and have some methods overridden.

adamleff (commented):
Left a few comments, and it looks like you have some lint errors to contend with.

lib/utils/schema/schema.rb
arlimus: Sure, good point! Also valid since it's still in `lib/utils` instead of some `extras` folder, so I fully 👍 this argument

lib/utils/schema/inspec.exec.full.json
arlimus: We certainly can :) I left it compressed initially to make sure this is just the minified schema. Reasoning should be done via the `schema.rb` file, changes too; These schemas get complex rather quicly where the `schema.rb` is easy to understand and follow. By making it more human-readable we give more access to it, but may also invite people to try to change the json file directly.

What do you think?

arlimus:
We had 2 more suggestions to make this part of the inspec CLI. While I hesitate to bloat the CLI calls, Adam also mentioned to make it a hidden CLI command. I like this approach, as it is only needed in m2m cases and should not bother regular users. That way the location of `lib/utils` is justified and won't cause any trouble either. I'll rewrite this.

arlimus:
@chris-rock @adamleff updated to `inspec schema NAME`

adamleff (approved):

chris-rock:
Thank you @arlimus and @adamleff

PR #1566, opened 2017-03-15
Docker resource

### LINKED ISSUE #71: docker resources
```
    describe docker.container('busybox') do
      it { should exist }
      it { should be_running }
    end

    describe docker.image('busybox:latest') do
      it { should exist }
    end
```

chris-rock:
This PR ships 3 new docker resources:

**docker_container**
```
describe docker_container('an-echo-server') do
  it { should exist }
  it { should be_running }
  its('id') { should_not eq '' }
  its('image') { should eq 'busybox:latest' }
  its('repo') { should eq 'busybox' }
  its('tag') { should eq 'latest' }
  its('ports') { should eq [] }
  its('command') { should eq 'nc -ll -p 1234 -e /bin/cat' }
end

describe docker_container(id: 'e2c52a183358') do
  it { should exist }
  it { should be_running }
end
```

**docker_image**

```
describe docker_image('alpine:latest') do
  it { should exist }
  its('id') { should_not eq '' }
  its('image') { should eq 'alpine:latest' }
  its('repo') { should eq 'alpine' }
  its('tag') { should eq 'latest' }
end

describe docker_image('alpine:latest') do
  it { should exist }
end

describe docker_image(id: '4a415e366388') do
  it { should exist }
end
```

**docker**
```
describe docker.containers do
  its('images') { should_not include 'u12:latest' }
end

describe docker.images do
  its('repositories') { should_not include 'inssecure_image' }
end

describe docker.version do
  its('Server.Version') { should cmp >= '1.12'}
  its('Client.Version') { should cmp >= '1.12'}
end

describe docker.object(id) do
  its('Configuration.Path') { should eq 'value' }
end

docker.containers.ids.each do |id|
  # call docker inspect for a specific container id
  describe docker.object(id) do
    its(%w(HostConfig Privileged)) { should cmp false }
    its(%w(HostConfig Privileged)) { should_not cmp true }
  end
end

```

This resource started as a port from https://github.com/dev-sec/cis-docker-benchmark and was refactored to cover more use cases.

Fixes #71


docs/resources/docker.md.erb
```diff
@@ -0,0 +1,90 @@
+---
+title: About the docker Resource
+---
+
+# docker
+
+Use the `docker` InSpec audit resource to test configuration data for docker daemon.
```
adamleff: I suppose it's not just "configuration data"... maybe "configuration and runtime data"?

docs/resources/docker.md.erb
```diff
@@ -0,0 +1,90 @@
...
+
+## Syntax
+
+A `docker` resource block declares the configuration data to be tested:
+
+    describe csv('file') do
```
adamleff: Just a reminder note to actually flesh this out with real/accurate docs before we merge. Lots of file/csv references in here, though I totally realize this is a WIP.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+        .add(:names,          field: 'names')
+        .add(:networks,       field: 'networks')
+        .add(:ports,          field: 'ports')
+        .add(:running_for,    field: 'runningfor')
+        .add(:sizes,          field: 'size')
+        .add(:status,         field: 'status')
+        .add(:exists?) { |x| !x.entries.empty? }
+  filter.connect(self, :containers)
+
+  attr_reader(:containers)
```
adamleff: Can we remove the `( )` here? It's customary to not include them when declaring attributes (i.e. `attr_reader :containers`)

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+        .add(:repositories,     field: 'repository')
+        .add(:tags,             field: 'tag')
+        .add(:sizes,            field: 'size')
+        .add(:digests,          field: 'digest')
+        .add(:created,          field: 'createdat')
+        .add(:created_since,    field: 'createdsize')
+        .add(:exists?) { |x| !x.entries.empty? }
+  filter.connect(self, :images)
+
+  attr_reader(:images)
```
adamleff: Same comment as above - can we remove the `( )`?

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+        .add(:labels,         field: 'labels')
+        .add(:local_volumes,  field: 'localvolumes')
+        .add(:mounts,         field: 'mounts')
+        .add(:names,          field: 'names')
+        .add(:networks,       field: 'networks')
+        .add(:ports,          field: 'ports')
+        .add(:running_for,    field: 'runningfor')
+        .add(:sizes,          field: 'size')
+        .add(:status,         field: 'status')
+        .add(:exists?) { |x| !x.entries.empty? }
```
adamleff: This feels like we need a default `exists?` accessor in `FilterTable` that can be added with `.add_accessor`. Obviously don't address this as part of this PR, but I've seen this multiple times so it feels ripe for consolidation.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+      DockerImageFilter.new(parse_images)
+    end
+
+    # returns the path of the docker service
+    # only works for Debian 8+, Ubuntu 16.04+ and CentOS 7+ where systemd is installed
+    def path
+      # TODO: use exception, once https://github.com/chef/inspec/issues/1205 is fixed
+      return nil unless is_systemd_system
+
+      cmd = inspec.command('systemctl show -p FragmentPath docker.service')
```
adamleff: Since we're doing something similar both here, and in `#socket`, maybe we should extract this to a `#systemd_value` helper method:

```
def systemd_value(unit_name, key)
  # return nil if not a systemd system
  # do the inspec.command call to systemctl, passing in #{unit_name}
  # return nil if nonzero
  # parse the stuff
  # return parsed_stuff[key]
end
```

That would simplify both of these to:

```
def path
   systemd_value('docker.service', 'FragmentPath')
end

def socket
  systemd_value('docker.socket', 'FragmentPath')
end
```

... and makes it easier to test.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+      params['FragmentPath']
+    end
+
+    def to_s
+      'Docker Host'
+    end
+
+    # returns the json output of `docker inspect`
+    def inspect(id)
+      return @inspect if defined?(@inspect)
```
adamleff: Since `docker` is the resource, do we run the risk of preventing a user from calling `inspect` twice in a profile/control with different IDs to inspect?

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+      data = JSON.parse(inspec.command('docker version --format \'{{ json . }}\'').stdout)
+      @version = Hashie::Mash.new(data)
+    end
+
+    private
+
+    def parse_containers
+      raw_containers = inspec.command('docker ps -a --no-trunc --format \'{{ json . }}\'').stdout
+      ps = []
+      # since docker is not outputting valid json, we need to parse each row
```
adamleff, line 137: LOL - thanks for nothing, docker! Just realized they output individual lines of valid JSON, but not a full JSON structure. Sigh.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+      c_images = []
+      raw_images.each_line { |entry|
+        c_images.push(JSON.parse(entry))
+      }
+      c_images
+    end
+
+    # checks if the system is Debian 8+, Ubuntu 16.04+ and RHEL 7+
+    def is_systemd_system
+      return true if inspec.os.redhat? && inspec.os.release.to_i >= 7
```
adamleff: This kinda makes me wish init system detection was part of the `os` resource because we're now defining what makes a systemd system here and in the `service` resource.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+        it { should be_running }
+        its('command') { should eq '/bin/sh' }
+      end
+    "
+
+    # TODO: we need to define if a simple string is a ID or NAME
+    # TODO: handle short ids
+    def initialize(opts = {})
+      # fallback for simple strings
+      if opts.is_a?(String)
```
adamleff: I'm confused about this section... aren't all IDs and names strings? I don't think there's anything that not a string...

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+    def to_s
+      'Docker Container'
+    end
+
+    private
+
+    def container_info
+      return @info if defined?(@info)
+      opts = @opts
+      @info = inspec.docker.containers.where { id == opts[:id] || names == opts[:name] }
```
adamleff: Should this be `name == opts[:id]` instead of `names`? I think we use the field name in the `#where` block rather than the accessor name.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+
+    # TODO: handle short image ids
+    def initialize(opts = {})
+      @opts = opts
+
+      # do sanitizion of input values
+      @opts = { image: opts } if opts.is_a?(String)
+
+      unless @opts[:image].nil?
+        unless @opts[:image].index(':').nil?
```
adamleff: Would you be OK changing this to an `if !... else` instead of `unless ... else`? I think most Rubyists try to shy away from `unless ... else` flows.

lib/resources/docker.rb
```diff
@@ -0,0 +1,341 @@
...
+    "
+
+    # TODO: handle short image ids
+    def initialize(opts = {})
+      @opts = opts
+
+      # do sanitizion of input values
+      @opts = { image: opts } if opts.is_a?(String)
+
+      unless @opts[:image].nil?
```
adamleff: I'd love to see this broken out into some helper methods for easier comprehension and easier testing...

```
@opts[:repo] = parse_repo_from_image(@opts[:image])
@opts[:tag] = parse_tag_from_image(@opts[:image])
# ... etc ...
```

adamleff (commented):
This is so cool. I can't wait to see this firmed up and ready-to-go.

lib/resources/docker.rb
chris-rock: Yes, that is not finalized

lib/resources/docker.rb
chris-rock: difficult, since init system is not necessarily tight to a specific operating system. Chef is the best example and uses runit. I know we need to extend this in future. Maybe we kick this out of the resource, let me think about it

lib/resources/docker.rb
chris-rock: No, since docker only exposes `names`, see https://github.com/chef/inspec/pull/1566/files#diff-a405d35eca4d46fa85c19029d6f3b79bR23. Technically multiple names could be attached, but I have not seen that in real world ...

lib/resources/docker.rb
chris-rock: Great idea!

lib/resources/docker.rb
chris-rock: I am okay with that...

lib/resources/docker.rb
aaronlippold: agree, and another example is cases where I have seen folks put in systemd on older distros, so perhaps detecting it in a more generic way is the way to go.

lib/resources/docker.rb
adamleff: Yup, you're right - I was totally looking at something different.

lib/resources/docker.rb
adamleff: @chris-rock Oh sure, totally. The Chef example is a good example, but it's also an example of where an installed application is providing its own process supervision vs. what the OS default is. It's pretty hard for someone to take a CentOS 7 box and convert it to be upstart natively, so it still feels like there's a pretty strong connection to the OS and the expected init system.

We're at least making that assumption now in two resources... Not sure the right way to solve it, but duplicating it feels wrong. 🙂  Shouldn't block this, but we should probably add an issue to solve init detection if more resources are going to rely on it.

lib/resources/docker.rb
chris-rock: I am going to remove that feature, since this is very systemd heavy

lib/resources/docker.rb
chris-rock: I am going to remove that for now, until I have a better idea. I do not think we need that at the moment, especially if used with our filter table

lib/resources/docker.rb
chris-rock: @adamleff Yes indeed, that is why we assume the default in our resources which is the only sane thing. If customers have a custom one, they can use `runit_service` etc. Detection of the init system is not easy, especially since they have fallbacks implemented, too. I am going to remove that feature for now, since I am not considering it as stable

adamleff (approved):
Superb work, @chris-rock. This is gonna be great.

PR #1572, opened 2017-03-17
Extend `gem` to take an optional `gem_binary`

nvwls:
This allows one to test if gems are being installed into an omnibus environment.

Implement `chef_gem` and `chef_server_gem` subclasses.


lib/resources/gem.rb
```diff
@@ -48,4 +52,48 @@ def to_s
...
+      super(package_name, gem_binary)
+    end
+
+    def to_s
+      "chef_gem #{@package_name}"
+    end
+  end
+
+  class ChefServerGemPackage < GemPackage
+    name 'chef_server_gem'
```
adamleff, line 81: I totally support the changes to the `gem` resource to support alternate gem binaries, and totally support the `chef_gem` resource as an easy shortcut.

However, I wonder if this is a slippery slope and `chef_server_gem` is the top of that slope. 🙂  Given how often InSpec is run via or alongside Chef, `chef_gem` has a lot of use cases. I just wonder if `chef_server_gem` will be used too infrequently and therefore should just use the new extension of `gem` directly.

lib/resources/gem.rb
```diff
@@ -13,16 +14,19 @@ class GemPackage < Inspec.resource(1)
       end
```
adamleff, line 14: Can you please add an additional example here using an alternate gem binary?

adamleff (changes requested):
@nvwls thank you for this contribution! I have an inquiry that I'd love to be discussed with the other InSpec maintainers.

Also, before we merge this, are you willing to put together the docs pages for `chef_gem` and, if agreed to be added, `chef_server_gem`?

PR #1575, opened 2017-03-18
Support YAML attributes files ending in .yaml

### LINKED ISSUE #1569: YAML file extensions support only (.yml)
### Description

Ignoring loading of attributes files

```
inspec exec ./test.rb --attrs ./test.yaml 
example.com should
     ∅  eq "example2.com"

     expected: "example2.com"
          got: "example.com"

     (compared using ==)
```

os: CentOS 7
inspec: 1.16.1

### Possible Solutions

here: https://github.com/chef/inspec/blob/725e0a1bc50008c8ad252bd8d5c833a7af20c606/lib/inspec/secrets/yaml.rb

You use filter for file name ```target.end_with?('.yml')```

But in wiki for YAML file we have both extensions https://en.wikipedia.org/wiki/YAML

```
Filename extension	.yaml, .yml
```

It make me confused, maybe it good for you to remove this filter for file name, because people can use other name.

mr-exz:
Fixes #1569

### DIFF
--- a/lib/inspec/secrets/yaml.rb
+++ b/lib/inspec/secrets/yaml.rb
@@ -9,7 +9,7 @@ module Secrets
     attr_reader :attributes
 
     def self.resolve(target)
-      unless target.is_a?(String) && File.file?(target) && target.end_with?('.yml')
+      unless target.is_a?(String) && File.file?(target)
         return nil
       end
       new(target)
@@ -17,7 +17,11 @@ module Secrets
 
     # array of yaml file paths
     def initialize(target)
-      @attributes = ::YAML.load_file(target)
+      begin
+        @attributes = ::YAML.load_file(target)
+      rescue => e
+        raise "Error reading Inspec attributes: #{e}"
+      end
     end
   end
 end
--- a/lib/inspec/secrets/yaml.rb
+++ b/lib/inspec/secrets/yaml.rb
@@ -9,7 +9,7 @@ module Secrets
     attr_reader :attributes
 
     def self.resolve(target)
-      unless target.is_a?(String) && File.file?(target) && [ '.yml', '.yaml' ].include?(File.extname(target).downcase)
+      unless target.is_a?(String) && File.file?(target) && ['.yml','.yaml'].include?(File.extname(target).downcase)
         return nil
       end
       new(target)
@@ -17,11 +17,9 @@ module Secrets
 
     # array of yaml file paths
     def initialize(target)
-      begin
         @attributes = ::YAML.load_file(target)
       rescue => e
         raise "Error reading Inspec attributes: #{e}"
-      end
     end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/inspec/secrets/yaml.rb
```diff
@@ -9,15 +9,19 @@ class YAML < Inspec.secrets(1)
     attr_reader :attributes
 
     def self.resolve(target)
-      unless target.is_a?(String) && File.file?(target) && target.end_with?('.yml')
```
adamleff, line 12: This class is part of our plugin system. Since there can be multiple plugins for a particular area of the InSpec codebase (in this case, the "secrets" area, or rather, the "attributes" area), we need a way to determine whether a plugin is appropriate for the target/file in question.  Therefore, for each plugin, there needs to be a `resolve` method that can be called to determine if this plugin is appropriate for the file.

If we remove the file extension check, this YAML plugin will get called for every file, even if it's not a YAML file. This doesn't matter right now since we only have one "secrets" plugin, but we need to fix it regardless.

My recommendation would be to change this line to something like this:

```
unless target.is_a?(String) && File.file?(target) && [ '.yml', '.yaml' ].include?(File.extname(target).downcase)
```

lib/inspec/secrets/yaml.rb
```diff
@@ -9,15 +9,19 @@ class YAML < Inspec.secrets(1)
...
+      unless target.is_a?(String) && File.file?(target)
         return nil
       end
       new(target)
     end
 
     # array of yaml file paths
     def initialize(target)
-      @attributes = ::YAML.load_file(target)
+      begin
```
adamleff: Travis is currently red because of a lint issue here. You've done excellent here catching a potential exception of loading in invalid YAML, however, you don't need to do a `begin` when you're in a method definition if you're essentially catching the entire method body. This can be shortened to:

```
def initialize(target)
  @attributes = ::YAML.load_file(target)
rescue => e
  raise "Error reading Inspec attributes: #{e}"
end
```

That should clear up the failed travis tests.

adamleff (changes requested):
@mr-exz Thank you very much for your contribution and taking the initiative to fix the issue you raised!

I left a few comments that we need to address to make the tests green and fix potential future issues with additional `secrets` plugins.

We also need you to ensure each of your commits has a signed-off-by line in order to meet the requirements of the DCO for licensing purposes. Please see https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco for more information.

I look forward to merging your contribution soon!

lib/inspec/secrets/yaml.rb
mr-exz: thx for explain, i'll fix it

mr-exz (commented):
done

mr-exz (commented):
ok

mr-exz (commented):
sorry for so many commits, please review again

mr-exz (commented):
done

PR #1579, opened 2017-03-20
Fix omnibus configuration

adamleff:
The latest omnibus master expects certain settings to be enabled for
S3 caching to work, and our omnibus.rb wasn't set up properly. Also
aligning some of our settings to chef/chef's settings which we know
to work properly.


chris-rock (approved):
Thank you @adamleff

PR #1587, opened 2017-03-21
Fix Habitat plan for nokogiri support

adamleff:
Nokogiri is failing to build in the habitat artifact due to the lack
of libxml2 and libxslt. This brings them in as dependencies and also
properly configures bundler to use them.


chris-rock (approved):
Thank you @adamleff

PR #1590, opened 2017-03-22
add tag object

chris-rock:
This PR adds tag support to InSpec objects.

### DIFF
--- a/lib/inspec/objects/tag.rb
+++ b/lib/inspec/objects/tag.rb
@@ -1,0 +1,27 @@
+# encoding:utf-8
+
+module Inspec
+  class Tag
+    attr_accessor :key, :value
+
+    def initialize(key, value)
+      @key = key
+      @value = value
+    end
+
+    def to_hash
+      {
+        name: @name,
+        options: @opts,
+      }
+    end
+
+    def to_ruby
+      "tag #{key.inspect}: #{value.inspect}"
+    end
+
+    def to_s
+      "Tag #{@key} with #{@value}"
+    end
+  end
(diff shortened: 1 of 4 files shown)


lib/inspec/objects/tag.rb
```diff
@@ -0,0 +1,27 @@
...
+module Inspec
+  class Tag
+    attr_accessor :key, :value
+
+    def initialize(key, value)
+      @key = key
+      @value = value
+    end
+
+    def to_hash
```
adamleff, line 12: I don't see `@name` and `@opts` used here. Should this be `@key` and `@value`?

lib/inspec/objects/tag.rb
```diff
@@ -0,0 +1,27 @@
...
+        options: @opts,
+      }
+    end
+
+    def to_ruby
+      "tag #{key.inspect}: #{value.inspect}"
+    end
+
+    def to_s
+      "Tag #{@key} with #{@value}"
```
adamleff: Just a Ruby styling thing, but I try to avoid using instance variables if they're set up as attr_readers or attr_accessors, just for consistency.

adamleff (changes requested):
This looks good. Just one question on the `to_hash` method of `Inspec::Tag`.

lib/inspec/objects/tag.rb
chris-rock: That is fixed now :-) I also added a test for `to_hash`

adamleff (approved):

PR #1585, opened 2017-03-21
Move simulator gems to www

adamleff:
The simulator gems are not necessary for building InSpec, only the website.
This is currently causing issues in the Jenkins infrastructure due to
a downstream dependency of github-markup.

Signed-off-by: Adam Leff <adam@leff.co>

PR #1591, opened 2017-03-22
Fixing www/docs rake tasks

adamleff:
This is a follow-up to #1585. The website-related Rake tasks have been
moved to a www-specific Rakefile which can be found in the www/
directory. Any web-release-related gems have been moved to the www
Gemfile as well.

This also included modifying the docs rake tasks to be path-
independent as they currently expect that Rake is being run from the
root of the repo.


chris-rock (approved):
Thank you @adamleff

PR #1594, opened 2017-03-24
Support vendored profiles in Habitat-packaged profiles

adamleff:
This change adds support in Habitat-packaged profiles for
profiles that depend on other profiles. When `inspec habitat
profile create` or `inspec habitat profile upload` is run,
it will see if the profile's dependencies have been vendored
yet, and if not, it will vendor them before creating the
habitat artifact.

For the git and URL fetchers, more explicit creation of the
target directories for the vendored profiles is done. This
is implicitly done via normal CLI interactions a user may
go through, but in our case, we want to ensure those directories
are there before the fetchers try to write out content.

By adding this support, we also fix a bug experienced in Habitat
where a profile that was packaged before an `inspec exec` was run
for the profile would cause a failure in Habitat. This is caused
by `inspec exec` doing a vendor of the dependencies if necessary
and generating the inspec.lock file. In Habitat, the package dir
is not writable by the hab user and InSpec would fail to run due
to an inability to write out an inspec.lock.

### DIFF
--- a/lib/bundles/inspec-habitat/profile.rb
+++ b/lib/bundles/inspec-habitat/profile.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 # author: Adam Leff
 
+require 'inspec/profile_vendor'
 require 'mixlib/shellout'
 require 'toml'
 
@@ -83,17 +84,8 @@ module Habitat
     def create_profile_object
       @profile = Inspec::Profile.for_target(
         path,
-        cache: Inspec::Cache.new(cache_path.to_s),
         backend: Inspec::Backend.create(target: 'mock://'),
       )
-    end
-
-    def cache_path
-      File.join(path, 'vendor')
-    end
-
-    def inspec_lockfile
-      File.join(path, 'inspec.lock')
     end
 
     def verify_profile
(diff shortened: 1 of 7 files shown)


lib/bundles/inspec-habitat/profile.rb
```diff
@@ -92,6 +106,20 @@ def verify_profile
       Habitat::Log.info('Profile is valid.')
     end
 
+    def vendor_profile_dependencies
```
chris-rock, line 101: It feels like we're duplicating code here: https://github.com/chef/inspec/blob/master/lib/inspec/base_cli.rb#L149 Can we extract one method to meet the needs of `base_cli` and this implementation?

This `vendor_deps` is also used https://github.com/chef/inspec/blob/master/lib/bundles/inspec-compliance/cli.rb#L160

lib/fetchers/url.rb
```diff
@@ -153,7 +153,9 @@ def download_archive_to_temp
     def download_archive(path)
       download_archive_to_temp
       final_path = "#{path}#{@archive_type}"
+      FileUtils.mkdir_p(File.dirname(final_path))
       FileUtils.mv(temp_archive_path, final_path)
+      FileUtils.chmod(0644, final_path)
```
chris-rock: Shouldn't but has this change any impact on on-habitat environments?

lib/bundles/inspec-habitat/profile.rb
adamleff: While there's not much to it, I will create a `Inspec::ProfileVendor` class right now that does will do the right thing and expose the necessary methods for the two implementations to get what they need.

I specifically can't reuse this implementation line-for-line because I need the vendored content in the actual profile path itself, not the working directory, etc. But I can work with it.

lib/fetchers/url.rb
adamleff: it shouldn't, but out of abundance of caution, I'll move it out of the fetcher and into the habitat bundle instead.

adamleff:
PR updated based on feedback from @chris-rock. @arlimus and @alexpop -- if either of you have time, I'd appreciate a set of eyes on this one.

arlimus (approved):
Love the new functionality, thank you @adamleff !!
Also great on the updated re-use + kudos for the review @chris-rock 

lgtm 😁

PR #1595, opened 2017-03-24
Feature/fix ability to pass in supermarket url

rylarson:
# Enable customization of supermarket_url

It looks like this was originally supposed to work, but at some point
the default value was put in the method body rather than in the method
parameters.

This change allows you to configure the supermarket_url in test kitchen
like so:

```
verifier:
  inspec_tests:
  - name: linux-hardening
    supermarket: som3guy/apache-disa-stig
    supermarket_url: https://my.supermarket.com
```

I added tests for all of the methods on `Supermarket::API` to make sure we don't have any regressions going forward.

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -40,4 +40,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'nokogiri', '~> 1.6'
   spec.add_dependency 'faraday', '>=0.9.0'
   spec.add_dependency 'toml', '~> 0.1'
+  spec.add_dependency 'addressable', '~> 2.5'
 end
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @rylarson for finding and fixing that annoying bug

adamleff (changes requested):
@rylarson thanks for the fix! Can you please add `addressable` to the gemspec for this change? Right now, we're bringing in `addressable` via other dependencies. Since we're now depending on it directly, we should declare that for future-proofing purposes.

rylarson:
@adamleff The gemspec has been updated with addressable. Sorry for missing that.

adamleff:
No apologies necessary, @rylarson! I'll review this shortly.

adamleff (approved):

adamleff:
@rylarson this looks fantastic, and major props for adding tests for the `Supermarket::API` class, that's awesome.

We greatly appreciate your contribution!

PR #1596, opened 2017-03-25
Extend `gem` to take an optional `gem_binary`

nvwls:
Allows one to test whether gems are installed into an omnibus environment.


chris-rock (approved):
@nvwls I like this simpler design. Thank you for improving the resources

adamleff (approved):
@nvwls this is a great contribution, thank you! I definitely like this version.

PR #1599, opened 2017-03-27
Require Ruby 2.1

### LINKED ISSUE #1597: inspec fails to install on centos 7
### Description

Briefly describe the issue and the expected behavior
```
# gem install inspec   
```
above command fails on centos 7 with
```
ERROR:  Error installing inspec:
	nokogiri requires Ruby version >= 2.1.0.
```

```
Linux localhost.localdomain 3.10.0-514.10.2.el7.x86_64 #1 SMP Fri Mar 3 00:04:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
Ruby version on centos 7 is 
```
ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
```
centos 7 instance is up-to-date

adamleff:
InSpec now depends on nokogiri which has a dependency of Ruby 2.1 or
later. If someone tries to `gem install inspec` using Ruby 2.0, it
will fail on the nokogiri dependency rather than InSpec itself.

Additionally, Ruby 2.0 is already EOL.

Fixes #1597

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -24,6 +24,8 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
+  spec.required_ruby_version = '>= 2.1'
+
   spec.add_dependency 'train', '>=0.22.0', '<1.0'
   spec.add_dependency 'thor', '~> 0.19'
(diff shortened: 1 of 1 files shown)


arlimus:
Makes a lot of sense and also it's good to finally have it inside of gemspec. Thank you for this 👍

arlimus (approved):

PR #1601, opened 2017-03-28
Remove errant puts in inspec habitat CLI command

adamleff:
Left a `puts` behind during some testing that made its way into
a PR. Removing it!


arlimus (approved):
Kudos for the quick fix!

PR #1607, opened 2017-03-29
[www] Resurrect global message, promote ChefConf and NYC Meetup

adamleff:
The global message was removed during the redesign. Adding it back
to promote ChefConf. Also adding the upcoming NYC meetup to the
front screen.


hannah-radish:
Nice Job! 👍

PR #1608, opened 2017-03-30
Add OSX support for host resource

RyanJarv:
- Ping on OSX uses -W instead of -w
- Can use nc for testing TCP connections. There is also a UDP mode for this but doesn't seem to work (and would be unreliable anyways) so following the behavior on windows here.

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -145,7 +145,7 @@ module Inspec::Resources
   # @see http://blogs.technet.com/b/josebda/archive/2015/04/18/windows-powershell-equivalents-for-common-networking-commands-ipconfig-ping-nslookup.aspx
   # @see http://blogs.technet.com/b/heyscriptingguy/archive/2014/03/19/creating-a-port-scanner-with-windows-powershell.aspx
   class WindowsHostProvider < HostProvider
-    def ping(hostname, port = nil, proto = nil)
+    def ping(hostname, port = nil, _proto = nil)
       # ICMP: Test-NetConnection www.microsoft.com
       # TCP and port: Test-NetConnection -ComputerName www.microsoft.com -RemotePort 80
       request = "Test-NetConnection -ComputerName #{hostname}"
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -48,6 +48,8 @@ module Inspec::Resources
         @host_provider = LinuxHostProvider.new(inspec)
       elsif inspec.os.windows?
         @host_provider = WindowsHostProvider.new(inspec)
+      elsif inspec.os.darwin?
+        @host_provider = DarwinHostProvider.new(inspec)
       else
         return skip_resource 'The `host` resource is not supported on your OS yet.'
       end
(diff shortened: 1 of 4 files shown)


lib/resources/host.rb
```diff
@@ -93,6 +95,30 @@ def initialize(inspec)
     end
   end
 
+  class DarwinHostProvider < HostProvider
+    def ping(hostname, _port = nil, _proto = nil)
```
adamleff: Prefixing a parameter name with an underscore is generally used to indicate that the parameter is not going to be used. Now that we're using it, we should change `_proto` to `proto`, and `_port` to `port`. This is a Rubocop check, and appears to be the reason why the Travis tests are failing for your change, as well. 🙂

adamleff (changes requested):
Thanks for your submission, @RyanJarv! This looks like a reasonable approach to me, and the TCP check is something we need to do for Linux as well.

We need to get your tests green, and it appears it should hopefully be as easy as fixing the parameter named for your `#ping` method.

lib/resources/host.rb
```diff
@@ -48,6 +48,8 @@ def initialize(hostname, params = {})
         @host_provider = LinuxHostProvider.new(inspec)
       elsif inspec.os.windows?
         @host_provider = WindowsHostProvider.new(inspec)
+      elsif inspec.os.darwin?
```
chris-rock, line 53: We should add a check for the `proto` and skip the resource if udp is used, since we do not support if for darwin. Same should be true for Linux but this is not related to this PR.

chris-rock (commented):
@RyanJarv Thank you for this great addition! I added a minor improvement request.

lib/resources/host.rb
RyanJarv: I just copied this behavior from the windows resource, but yeah I see what you're saying.. Going to put a `skip_resource` in the beginning of the initialize for UDP since this isn't supported on any platform if that makes sense.

RyanJarv: Also I kinda feel like you shouldn't really be checking for UDP remotely.. just going to lead to confusing results half the time. But that's kinda a different issue.

RyanJarv:
Thanks! Pushed a change for the variable names (guess I was looking at `rake rubocop` before which isn't used maybe?). Also went ahead and moved the check for UDP out to the initializer since none of the checks support it currently. Windows was just returning nil in that case rather then skipping the resource, I imagine there wasn't a good reason for that but I'm not to familiar with Inspec so correct me if I'm wrong.

Edit: nvm `rake rubocop` is all green now.. I must be confused

RyanJarv:
Looks like the travis build fail is unrelated? Not getting that running the tests locally.

adamleff:
@RyanJarv Yes, likely unrelated. The functional tests are a bit fragile at the moment. I've rerun the Travis job.

adamleff (approved):

chris-rock (approved):

adamleff:
Thanks for your contribution @RyanJarv!

PR #1611, opened 2017-03-30
[www] Fixing rake www:release task

adamleff:
After moving the www tasks into the www directory, the release
task needed to be updated to run certain commands from the repo
root such that the newly-built web content was in the repo root
of the gh-pages branch, rather than the www/ directory.

Also updating the www Gemfile.lock with the new inspec release.


arlimus (approved):
Great update and fixes, thank you @adamleff 😄 👍

PR #1454, opened 2017-02-03
Functional JUnit reporter

jkerry:
resolves #1438 by removing the gem based junit reporter and manually rolling up JUnit formatted XML based on the JSON reporter's @output_hash.

Some minor ancillary changes:

- The testsuite is no longer expected to have an errors count
- nokogiri gem elevated to a core dependency

PR #1621, opened 2017-04-03
Replace Nokogiri with REXML in the JUnit formatter

adamleff:
In #1454, we welcomed a newly-revamped JUnit formatter which has
a dependency on Nokogiri. Unfortunately, this had led us to problems
getting InSpec included in Chef omnibus builds (see chef/chef#5937)
because Chef is using Ruby 2.4.1 and the Nokogiri maintainers have
not yet released a windows binary gem that supports Ruby 2.4.x.
This has led to breaking builds in Chef's CI platform and would
block the acceptance of chef/chef#5937.

This change replaces Nokogiri use with REXML instead. While REXML
can be slower than Nokogiri, it does not require native extensions
and is supported on all Chef platforms.

/cc @jkerry


inspec.gemspec
```diff
@@ -39,7 +39,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'mixlib-log'
   spec.add_dependency 'sslshake', '~> 1'
   spec.add_dependency 'parallel', '~> 1.9'
-  spec.add_dependency 'nokogiri', '~> 1.6'
```
chris-rock, line 42: do we need `REXML` as dependency then?

test/functional/inspec_exec_junit_test.rb
```diff
@@ -11,55 +11,55 @@
     out = inspec('exec ' + example_control + ' --format junit --no-create-lockfile')
     out.stderr.must_equal ''
     out.exit_status.must_equal 0
-    doc = Nokogiri::XML(out.stdout)
-    doc.errors.length.must_equal 0
+    doc = REXML::Document.new(out.stdout)
+    doc.has_elements?.must_equal true
```
chris-rock, line 15: Isn't that a different test? before we checked the errors in doc but now we just check that doc has elements?

test/functional/inspec_exec_junit_test.rb
```diff
@@ -11,55 +11,55 @@
...
   end
 
   it 'can execute the profile with the junit formatter' do
     out = inspec('exec ' + example_profile + ' --format junit --no-create-lockfile')
     out.stderr.must_equal ''
     out.exit_status.must_equal 0
-    doc = Nokogiri::XML(out.stdout)
-    doc.errors.length.must_equal 0
+    doc = REXML::Document.new(out.stdout)
+    doc.has_elements?.must_equal true
```
chris-rock, line 23: same as above?

chris-rock (commented):
Thank you @adamleff for this improvement. I've added some comments. Can REXML be used in conjunction with nokogiri?

inspec.gemspec
adamleff: No, `REXML` is in the Ruby stdlib.

chris-rock: Oh, stupid me.

test/functional/inspec_exec_junit_test.rb
adamleff: No. Nokogiri will happily try and parse anything you throw at it and return a `Nokogiri::XML::Document` object. It will not raise an exception unless you enable strict parsing.

So the way to determine if Nokogiri parsed the XML successfully is to call `#errors` on it, or set string parsing and catch the exception.

REXML is somewhat similar... it will take any string or `IO` object and try to return a `REXML::Document` object. You can tell if it's valid XML if it has any `REXML::Element` objects contained within it.

test/functional/inspec_exec_junit_test.rb
adamleff: No. Nokogiri will happily try and parse anything you throw at it and return a `Nokogiri::XML::Document` object. It will not raise an exception unless you enable strict parsing.

So the way to determine if Nokogiri parsed the XML successfully is to call `#errors` on it, or set string parsing and catch the exception.

REXML is somewhat similar... it will take any string or `IO` object and try to return a `REXML::Document` object. You can tell if it's valid XML if it has any `REXML::Element` objects contained within it.

adamleff:
@chris-rock there is nothing blocking someone from using Nokogiri and REXML in the same library/application. REXML is in the stdlib and all its classes are namespaced accordingly.

chris-rock (approved):
Awesome. Thank you for the clarification @adamleff

PR #1623, opened 2017-04-03
[www] Add warning to the http resource documentation

adamleff:
Currently, the `http` resource only operates on the node on which
`inspec exec` is run, even if the user uses the `--target` flag.
This has caused some confusion in the community.

Until this can be fixed, this change adds a warning to the documentation.


arlimus (approved):
Love the note in the docs on this Adam, thank you for quickly adding it!! 👍

PR #1624, opened 2017-04-04
Adding SNI utilization to ssl resource

supergicko:
Hi

updated the ssl resource to utilize the new sni feature [1] in the underlying sslshake lib.
its now possible to scan servers serving multiple domains over ssl/tls

[1] https://github.com/arlimus/sslshake/pull/5

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -37,7 +37,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'pry', '~> 0'
   spec.add_dependency 'hashie', '~> 3.4'
   spec.add_dependency 'mixlib-log'
-  spec.add_dependency 'sslshake', '~> 1'
+  spec.add_dependency 'sslshake', '~> 1.1'
   spec.add_dependency 'parallel', '~> 1.9'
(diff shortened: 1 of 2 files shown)


adamleff (changes requested):
Thank you for your contribution @supergicko!

We should probably update the Gemspec to require the newer version of `sslshake`.

It's currently: 

```
```
... which, I believe, is the same thing as saying `>= 1.0.0`.

I'd propose we change this to:

```
spec.add_dependency 'sslshake', '~> 1.1'
```

... so we pick up at least v1.1.0 which has the new SNI support but don't pick up v2 if it's ever released without us deliberately opting in.

Otherwise, I'm fine with this change. :)

supergicko:
@adamleff good point, pinned it.

supergicko:
@adamleff i guess the CI fail is unrelated to the change (failing also in other PRs)

adamleff:
Yeah, the functional tests are a bit fragile in Travis. I'll re-run and it should clear up.

adamleff (approved):

chris-rock (approved):
Thank you @supergicko

PR #1627, opened 2017-04-04
Loosen addressable version pin

adamleff:
The version of addressable that was pinned in the gemspec was
too new and conflicted with some of chef's dependencies. Loosening
it will allow us to include InSpec in the chef omnibus packages.


chris-rock (commented):
Thank you @adamleff

PR #1636, opened 2017-04-07
Add type conversion routines to `SimpleConfig`

nvwls:
Allows one to pass in `:conversion` hash to convert value into a number or boolean.  This allows for more complex verification, i.e. `its('gpgcheck') { should eq true }`

### DIFF
--- a/docs/resources/parse_config.md.erb
+++ b/docs/resources/parse_config.md.erb
@@ -117,6 +117,29 @@ Use `standalone_comments: false`, to parse the following:
     'key = value # comment'
     params['key'] = 'value'
 
+### conversion
+
+Use `conversion` to specify a hash of type conversion routines.
+
+    config = <<EOF
+    uid = 636
+    gid = 389
+    EOF
+
+    options = {
+      conversion: {
+        'uid' => :decimal,
+        'gid' => :decimal,
+      }
+    }
+
+    describe parse_config(config, options) do
+      its('uid') { should be >= 500 }
+      its('gid') { should be < 500 }
+    end
+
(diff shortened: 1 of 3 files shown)


chris-rock:
@nvwls Thank you for your contribution.

We had the question about type conversion early on in our design decision for InSpec. Serverspec is doing decisions on current values. We decided to move conversion logic out of the resources in order to make the results fully deterministic. Therefore we've implemented most of the `smart` conversion in our `cmp` matcher: https://github.com/chef/inspec/blob/master/lib/matchers/matchers.rb#L233-L330

It may be still useful to have a conversion in our `parse_config ` parser. Before I am able to make my mind, I need to understand the use-case and the reason why you have not used the `cmp` matcher?

```
# use cmp instead of eq should do the trick:
its('gpgcheck') { should cmp true }
```

nvwls:
One use case was fuzzy booleans, i.e. true => ['y', 'yes', 't', 'true', '1'] and false => ['n', 'no', 'f', 'false', '0']

chris-rock:
@nvwls Got your idea. Can you provide a specific configuration file where you had this issue?

aaronlippold:
Just a thought, I would want to be careful with how we go forward with this given that we want to ensure that we have some standardization. I.e. that we maintain some semblance of "Subject/Verb" agreement in our sentence structure if that analogy makes sense. I agree that convince methods are useful but we also need to ensure that how the tests are written across resources is consistent.

pwelch:
Closing this pull request because there has been no traction on it for a while. If this change is still relevant, please reopen. Thanks!

PR #1640, opened 2017-04-08
Yaml file warning and stop go forward

mr-exz:
Add notification about file with wrong expansion was ignored.
Exit if file not loaded because if we pass it, we run tests without variables and it hard to troubleshoot.


lib/inspec/secrets/yaml.rb
```diff
@@ -10,7 +10,8 @@ class YAML < Inspec.secrets(1)
 
     def self.resolve(target)
       unless target.is_a?(String) && File.file?(target) && ['.yml', '.yaml'].include?(File.extname(target).downcase)
-        return nil
+        puts "Warning file: #{target} was ignored, please use to .yml or .yaml"
```
chris-rock: We cannot exit here. This code will also be used if another secrets backend is used.

chris-rock (changes requested):
@mr-exz I fully understand your desire to get clear output if the config is not loaded. Unfortunately we cannot just exit in our secrets backend. Our code in https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L121-L135 asks all registered secret backends to check if the support the input.

Instead of exiting here, we should implement the logic in https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L121-L135. If a given secrets input does not match a backend, we should throw an exception here: https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L127. Errors should only be handled in CLI. Backend implementation should never exit. Once the exception is raised properly, it can be handled in https://github.com/chef/inspec/blob/master/lib/inspec/base_cli.rb#L75-L87

mr-exz:
i remove 'exit'

chris-rock:
@mr-exz As I mentioned earlier, the warning belongs into https://github.com/chef/inspec/blob/master/lib/inspec/runner.rb#L127 Whenever we got a target and we could not find a suitable backend we should stop the execution.

Therefore we should:
- `load_attributes` is raising an exception if a target argument cannot be mapped to an implementation. Therefore `secrets` in load_attributes will be `nil`
- that exception will be propagated to https://github.com/chef/inspec/blob/master/lib/inspec/base_cli.rb#L82 and we need to display an error message here: https://github.com/chef/inspec/blob/master/lib/inspec/base_cli.rb#L84

Would that make sense?

mr-exz:
I should investigate a bit more, before, i misunderstand.
Thx for explain, give me a time.

mr-exz:
please help with error

adamleff:
The Travis error is likely unrelated. I will re-run the job.

adamleff:
@mr-exz thanks for your submission!

It appears there's still a disconnect between @chris-rock's recommendation on how to implement and your proposed solution. I'd like to offer to adopt your PR and help push it across the finish line. I'll make sure you are still credited with the initial work.

Are you OK with that?

mr-exz:
@adamleff i'm ok, no problem, thx

adamleff:
@mr-exz thanks again for your contribution and your desire to make InSpec easier to use. I have opened new PR #1775 that is inspired by your work and will be closing this PR accordingly.

Thank you!

PR #1646, opened 2017-04-11
pass options hash sans target key

### LINKED ISSUE #1631: cannot upload meta profile to Automate
### Description
Uploading a meta profile to Automate does not work with inspec 1.19.1, however it does work with inspec 1.11.0 from chefdk 1.2.22.

1.19.1

### Replication Case

```
~/Devel/compliance-profiles.sav (test_upload *%)$ /opt/chefdk/bin/inspec version
1.11.0

Your version of InSpec is out of date! The latest version is 1.19.1.
~/Devel/compliance-profiles.sav (test_upload *%)$ /opt/chefdk/bin/inspec compliance upload meta-profile.tar.gz --overwrite
I, [2017-04-05T08:47:35.826510 #6236]  INFO -- : Checking profile in meta-profile.tar.gz
I, [2017-04-05T08:47:35.826627 #6236]  INFO -- : Metadata OK.
I, [2017-04-05T08:47:38.983669 #6236]  INFO -- : Found 79 controls.
W, [2017-04-05T08:47:38.983775 #6236]  WARN -- : Control tls1.2 has no description
W, [2017-04-05T08:47:38.983792 #6236]  WARN -- : Control tls1.2 has no tests defined
W, [2017-04-05T08:47:38.983802 #6236]  WARN -- : Control ssl2 has no description
W, [2017-04-05T08:47:38.983811 #6236]  WARN -- : Control ssl2 has no tests defined
W, [2017-04-05T08:47:38.983854 #6236]  WARN -- : Control ssl3 has no description
W, [2017-04-05T08:47:38.983862 #6236]  WARN -- : Control ssl3 has no tests defined
W, [2017-04-05T08:47:38.983870 #6236]  WARN -- : Control tls1.0 has no description
W, [2017-04-05T08:47:38.983877 #6236]  WARN -- : Control tls1.0 has no tests defined
W, [2017-04-05T08:47:38.983884 #6236]  WARN -- : Control tls1.1 has no description
W, [2017-04-05T08:47:38.983891 #6236]  WARN -- : Control tls1.1 has no tests defined
W, [2017-04-05T08:47:38.983898 #6236]  WARN -- : Control rc4 has no description
W, [2017-04-05T08:47:38.983906 #6236]  WARN -- : Control rc4 has no tests defined
W, [2017-04-05T08:47:38.983913 #6236]  WARN -- : Control verify-kb has no description
W, [2017-04-05T08:47:38.983923 #6236]  WARN -- : Control important-count has no description
W, [2017-04-05T08:47:38.983931 #6236]  WARN -- : Control important-patches has no description
W, [2017-04-05T08:47:38.983938 #6236]  WARN -- : Control important-patches has no tests defined
W, [2017-04-05T08:47:38.983946 #6236]  WARN -- : Control optional-count has no description
W, [2017-04-05T08:47:38.983954 #6236]  WARN -- : Control optional-patches has no description
W, [2017-04-05T08:47:38.984033 #6236]  WARN -- : Control optional-patches has no tests defined
Profile is valid
Start upload to admin/meta-profile
Uploading to Chef Automate
Successfully uploaded profile
~/Devel/compliance-profiles.sav (test_upload *%)$ inspec version
      net-ssh (< 5.0, >= 2.6.5, >= 2.9)
      ffi (>= 1.0.1)
      builder (>= 2.1.2)
      diff-lcs (< 2.0, >= 1.2.0)
1.19.1
~/Devel/compliance-profiles.sav (test_upload *%)$ inspec compliance upload meta-profile.tar.gz --overwrite
      net-ssh (< 5.0, >= 2.6.5, >= 2.9)
      ffi (>= 1.0.1)
      builder (>= 2.1.2)
      diff-lcs (< 2.0, >= 1.2.0)
/opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/train-0.22.1/lib/train.rb:122:in `validate_backend': Cannot determine backend from target configuration "meta-profile.tar.gz". Valid example: ssh://192.168.0.1. (Train::UserError)
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/backend.rb:17:in `create'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/profile.rb:92:in `initialize'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/profile.rb:63:in `new'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/profile.rb:63:in `for_path'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/profile.rb:69:in `for_fetcher'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/inspec/profile.rb:75:in `for_target'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/lib/bundles/inspec-compliance/cli.rb:165:in `upload'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:115:in `invoke'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:235:in `block in subcommand'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /opt/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.19.1/bin/inspec:12:in `<main>'
~/Devel/compliance-profiles.sav (test_upload *%)$ chef -v
Chef Development Kit Version: 1.2.22
chef-client version: 12.18.31
delivery version: master (0b746cafed65a9ea1a79de3cc546e7922de9187c)
berks version: 2017-04-05T08:49:35.714968 6305] 2017-04-05T08:49:35.715121 6305] 2017-04-05T08:49:35.715230 6305] 2017-04-05T08:49:35.715331 6305] 2017-04-05T08:49:35.745557 6305] 2017-04-05T08:49:35.745687 6305] 5.6.0
kitchen version: 1.15.0
~/Devel/compliance-profiles.sav (test_upload *%)$
```

### Possible Solutions

### Stacktrace

jeremymv2:
This PR fixes https://github.com/chef/inspec/issues/1631

```
validate_backend': Cannot determine backend from target configuration "meta-profile.tar.gz". Valid example: ssh://192.168.0.1. (Train::UserError)
```

Originally, the `:target` key was deleted from the `options` hash in the `initialize` method of `Inspec::Profile`, before being passed to `Inspec::Backend.create()`  Recently, in the commit below, the `:target` delete method was removed so that the key still existed in the hash:  https://github.com/chef/inspec/commit/d0bc085412fb68890f6d6ac8a042381240385f2c#diff-d1eede4085683a605d9b8892a198f642L85

This change introduced an unintended behavior: when train goes off to validate in the `validate_backend` method, it's raising an error since `:target` is not `nil`: https://github.com/chef/train/blob/master/lib/train.rb#L121-L124

This PR simply reverts the original behavior by passing the `options` hash to `Inspec::Backend.create()` without the `:target` key in it.  With this, the functionality works as before and preserves the desired effect from the commit above.

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -89,7 +89,7 @@ module Inspec
       @writable = options[:writable] || false
       @profile_id = options[:id]
       @cache = options[:cache] || Cache.new
-      @backend = options[:backend] || Inspec::Backend.create(options)
+      @backend = options[:backend] || Inspec::Backend.create(options.select { |k, _| k != 'target' })
       @attr_values = options[:attributes]
       @source_reader = source_reader
       @tests_collected = false


lib/inspec/profile.rb
```diff
@@ -89,7 +89,7 @@ def initialize(source_reader, options = {})
       @writable = options[:writable] || false
       @profile_id = options[:id]
       @cache = options[:cache] || Cache.new
-      @backend = options[:backend] || Inspec::Backend.create(options)
+      @backend = options[:backend] || Inspec::Backend.create(options.select { |k, _| k != 'target' })
```
adamleff, line 92: The fact that we hit (read: I created) this bug in the first place is because we blindly pass a hash of options that is created by the caller rather than curating a hash of options we actually need. Rather than deleting an option we don't need, I'd rather we only pass the options we **do** need. Otherwise, people like me will see calls to `#delete` without knowing why, there won't be a comment as to why, and then bugs are created. 🙂 

It's possible that we need to be actually creating the hash we need in `Inspec::Backend` rather than here.

@chris-rock / @arlimus - thoughts?

jeremymv2: To me this area of the code is a bit opaque (regarding passing of a `:target`) without knowing what the message spec from Profile -> Backend should look like under different circumstances.  I agree that we should determine what the message(s) should contain and be able to enforce that on the receiving end.

chris-rock: I think we need to make the separation more clear. The problem with `the options we do need` is that the backend is getting most of the cli options, since all the ssh or winrm options are just passed through. I think we should create an issue to solve that in another case. The real issue I am seeing here is that we need to run integration tests with Chef Automate for every PR, since a change should break the build.

So we need to do two things:
- Integration testing with Chef Compliance / Chef Automate
- Make the interface between inspec / train more clear

chris-rock (approved):
I confirmed that issue. LGTM. Thank you for spotting that issue @jeremymv2

lib/inspec/profile.rb
adamleff: Agreed that additional testing is always a good thing. I don't agree that that's the "real issue." We have a situation in the codebase where `A` is creating an options hash, `B` is passing it to `C`, but not all the options are needed by `C` and, when they are all included, we break. So even if "all the ssh and winrm options are just passed through," we're not doing a good job with the separation. Testing will catch it, but the lack of testing is not the cause here.

adamleff (approved):
Approved with #1652 and #1653 to address the root issues.

PR #1650, opened 2017-04-12
Fix gem resource on Windows

### LINKED ISSUE #1645: gem resource :chef symbol isn't valid on Windows
### Description

On Windows platforms, `:chef` does not resolve to the correct `gem_binary` location causing InSpec to fail to find gems installed to Chef Omnibus embedded Ruby. 

1.19.1
Windows 7 x64

### Replication Case

```
  describe gem('chef-vault', :chef) do
    it { should be_installed }
  end
```
fails; however this works:
```
  describe gem('chef-vault', 'C:\opscode\chef\embedded\bin\gem.cmd') do
    it { should be_installed }
  end
```

### Possible Solutions

Perhaps https://github.com/chef/inspec/blob/master/lib/resources/gem.rb#L26 just needs `.cmd` appended?

@nvwls Thoughts?

adamleff:
RubyGems on windows comes with a batch file that wraps the `gem` command
so it executes correctly. This change uses that batch file for windows
for our `gem` resource, and also properly handles when we receive no output
from the command.

Fixes #1645

### DIFF
--- a/lib/resources/gem.rb
+++ b/lib/resources/gem.rb
@@ -23,7 +23,7 @@ module Inspec::Resources
                       'gem'
                     when :chef
                       if inspec.os.windows?
-                        'c:\opscode\chef\embedded\bin\gem'
+                        'c:\opscode\chef\embedded\bin\gem.bat'
                       else
                         '/opt/chef/embedded/bin/gem'
                       end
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @adamleff

PR #1651, opened 2017-04-12
[www] Update community page

adamleff:
Adding additional text and links to discourse, blog, and events. The current
community page only has a link to Slack.

### DIFF
--- a/www/source/community.html.slim
+++ b/www/source/community.html.slim
@@ -1,6 +1,6 @@
 ---
 title: InSpec - Community
-description: This is where you interact with the InSpec open source community - contribute to the project, provide feedback and ask questions of other community members.
+description: Interact with the InSpec community - contribute to the project, provide feedback, and ask questions of other community members.
 ---
 
 .row.margin-both-offset
@@ -13,9 +13,9 @@ description: This is where you interact with the InSpec open source community - 
       | Let’s build together.
     hr.strict-left.margin-under-xs
     h3.clear
-      | Join us on Slack and get answers fast  
-    p 
-      | Is something unclear or are you just feeling stuck?  We're here to help keep you moving. Chat with us on Slack so we can help you keep moving.
+      | Join us on Slack and get answers fast
+    p
+      | Best practices, live discussions, and help from the community.
     p
       a.btn.slack href="http://community-slack.chef.io/" Join us on Slack
     p.margin-under-xs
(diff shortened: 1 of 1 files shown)


www/source/community.html.slim
```diff
@@ -11,16 +11,34 @@ description: This is where you interact with the InSpec open source community -
...
+    p
+      | Post questions, help out fellow community members, and learn about upcoming events. The <a href="https://discourse.chef.io/c/chef-release" target="_blank">release announcements</a> category is a great place to stay in-the-know about new Chef software, too!
+    p.margin-under-xs
+      a.btn.forums href="https://discourse.chef.io/c/inspec" Visit the Forums
+    h3
+      | Upcoming Events
+    p
+      | Find an event in your area to learn about InSpec, try it out for yourself, and meet fellow InSpec community members.
+    p.margin-under-xs
+      a.btn.events href="https://discourse.chef.io/c/inspec" Meet at Events
```
chris-rock: That is not referring to the events page

www/source/community.html.slim
```diff
@@ -11,16 +11,34 @@ description: This is where you interact with the InSpec open source community -
...
+    h3
+      | The InSpec Blog
+    p
+      | Recaps of past events, quick how-to guides on new functionality, and lots more can be found on the blog.
+    p.margin-under-xs
+      a.btn.blog href="https://blog.chef.io/category/inspec/" Read Blog Posts
```
chris-rock, line 40: Do we distinguish between InSpec and Compliance categories? I would like to include the compliance blogs as well

chris-rock (changes requested):

www/source/community.html.slim
adamleff: Thanks for the catch.

www/source/community.html.slim
adamleff: We'd need to either do two buttons (which I think looks weird on this page) or ask Marketing to create a top-level category that brings InSpec and Compliance in... we can't create a wordpress URL that links to a two-category combination.

I'd recommend we move forward with just the InSpec category now and we can ask for Marketing's help in handling this on the wordpress side.

chris-rock: Agreed

chris-rock (approved):

PR #1656, opened 2017-04-13
Habitat packages should run as root

adamleff:
Many InSpec resources require root access to properly scan. Let's
default the run user to root until we need to accommodate different
options.


chris-rock (approved):
Thank you @adamleff

PR #1662, opened 2017-04-14
Move Habitat sleep time to config file

adamleff:
Because the sleep_time is not written to a config file but instead
only rendered into the run hook, hab-sup doesn't restart the running
process upon any config updates. This change moves the sleep_time to
a settings config file which is read in by the run hook. This will
allow Habitat to restart the InSpec process whenever a user changes
the sleep time.

I also cleaned up the non-zero exit error message to give the user
a better indication as to why the run may have "failed."

### DIFF
--- a/lib/bundles/inspec-habitat/profile.rb
+++ b/lib/bundles/inspec-habitat/profile.rb
@@ -43,6 +43,7 @@ module Habitat
       copy_profile_to_work_dir
       create_plan
       create_run_hook
+      create_settings_file
       create_default_config
 
       # returns the path to the .hart file in the work directory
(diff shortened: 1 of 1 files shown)


lib/bundles/inspec-habitat/profile.rb
```diff
@@ -185,6 +187,12 @@ def create_run_hook
       File.write(run_hook_file, run_hook_contents)
     end
 
+    def create_settings_file
```
chris-rock, line 190: Is this a habitat settings file? Are we potentially overwriting other habitat configs?

adamleff: This is a settings/config file that is specific to our habitat package and contains only the `sleep_time` environment variable we've created to govern how often InSpec should run. It has no affect on other Habitat settings.

As I described in the PR description, this is necessary so that Habitat Supervisor is controlling a config (vs. when I had it only in the run hook) so that a config value change at run-time will cause Habitat to restart the service. If it doesn't have a config file, it will think no restart is necessary.

chris-rock: Thank you for clarification @adamleff I never questioned the config value 👍

chris-rock (approved):

Issue #1649, opened 2017-04-12
Improve `http` resource


chris-rock:
### Description

Right now, we cover basic tests with `http` only:

```
describe http('http://localhost:8080/ping', auth: {user: 'user', pass: 'test'}, params: {format: 'html'}) do
  its('status') { should cmp 200 }
  its('body') { should cmp 'pong' }
  its('headers.Content-Type') { should cmp 'text/html' }
end

describe http('http://example.com/ping').headers do
  its('Content-Length') { should cmp 258 }
  its('Content-Type') { should cmp 'text/html; charset=UTF-8' }
end
```

It would be nice if we could make testing via `http` easier.

### InSpec and Platform Version

1.19.2

### Possible Solutions

**Write a resource around faraday and reuse its interface**

To highlight the reuse of faraday, I am going to use `faraday` as the resource name instead of `http`

```
describe faraday(:url => 'http://sushi.com').get '/nigiri/sake.json' do
  its('body') { should cmp 'pong' }
end      
```

```
describe faraday(:url => 'http://sushi.com').get '/nigiri', { name: 'Maguro' }
  its('status') { should cmp 'pong' }
  its('body') { should cmp 'pong' }
end
```

```
describe faraday(:url => 'http://sushi.com').post {url: '/nigiri', headers: { 'Content-Type' => 'application/json' }, body: '{ "name": "Unagi" }' }
  its('body.name') { should eq "Unagi" }
end
```

**Support for custom ca_path**

Sometimes we have custom `ca_path`. See https://github.com/lostisland/faraday/wiki/Setting-up-SSL-certificates for full description. We would follow the faraday pattern too:

```
cert_path = "~/.chef/trusted_certs"

describe faraday(url: 'https://encrypted.google.com', ssl: { ca_path: "/usr/lib/ssl/certs" }).get '/nigiri/sake.json' do
  its('body') { should cmp 'pong' }
end   
```

chris-rock:
Just to be clear. Although I used `faraday` as the resource name, this was just used to highlight our reuse of faraday gem. The resource name will stay `http`

wswheeler:
I started using the http resource as part of running tests against webservers that I was building up with Chef. The feature that I think would be most useful is if there was a way to lie about DNS to validate SSL certificates as well as well configurations. For example, I'd like to go to the full host name with https and validate that it connected and additionally got the expected content. Right now I'm faking this with the following type of describe. 

```
  describe http('https://127.0.0.1/',
[...]

PR #1663, opened 2017-04-14
ER-508 Extended http resource to support no ssl verification

ElizabethU:
We needed to be able to run inspec against endpoints with self signed certificates and this was the quickest way for us to get there.

We originally started working on a [Faraday resource](https://github.com/chef/inspec/issues/1649), but had a lot of implementation questions, and felt this was the simplest way to have something up for review. We'd love advice on how to test this if you think this is a fine change, or advice on how to change it, if not.

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -25,13 +25,14 @@ module Inspec::Resources
     "
 
     # rubocop:disable ParameterLists
-    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil)
+    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil, ssl_verify: true)
       @url = url
       @method = method
       @params = params
       @auth = auth
       @headers = headers
       @data = data
+      @ssl_verify = ssl_verify
     end
 
     def status
@@ -53,7 +54,7 @@ module Inspec::Resources
     private
 
     def response
-      conn = Faraday.new url: @url, headers: @headers, params: @params
+      conn = Faraday.new url: @url, headers: @headers, params: @params, ssl: { verify: @ssl_verify }
 
       # set basic authentication
       conn.basic_auth @auth[:user], @auth[:pass] unless @auth.empty?


lib/resources/http.rb
```diff
@@ -25,13 +25,14 @@ class Http < Inspec.resource(1)
...
     # rubocop:disable ParameterLists
-    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil)
+    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil, ssl_verify: true)
       @url = url
       @method = method
       @params = params
       @auth = auth
       @headers = headers
       @data = data
+      @verify = ssl_verify
```
adamleff: Can we set the instance variable to `@ssl_verify` as well? `@verify` is fairly ambiguous, and all the other instance variables match the incoming arguments.

lib/resources/http.rb
```diff
@@ -53,7 +54,7 @@ def to_s
     private
 
     def response
-      conn = Faraday.new url: @url, headers: @headers, params: @params
+      conn = Faraday.new url: @url, headers: @headers, params: @params, ssl: {verify: @verify}
```
adamleff: The lack of spaces inside the `{ }` is causing you fail the lint check.

adamleff (changes requested):
I'm totally fine with this approach. Left a request for a variable name change. We also need to get the Travis and DCO bot tests to go green. Let me know if you need help with that.

I don't think any additional tests are necessary here. Any integration/functional tests would be testing Faraday rather than InSpec, and a unit test to check to make sure we're passing the same parameter value to `Faraday#new` feels overkill to me.

adamleff (approved):

lib/resources/http.rb
```diff
@@ -25,13 +25,14 @@ class Http < Inspec.resource(1)
     "
 
     # rubocop:disable ParameterLists
-    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil)
+    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil, ssl_verify: true)
```
chris-rock, line 28: I would prefer to use `ssl: { verify: true }` as input, since we do not need to do a lot of translation between our own config and faraday config. Also this would make the path to implement the improvements proposed in #1649


I propose we hand over a ssl hash entry and pass this. Right now, we support verify only.

chris-rock (changes requested):

adamleff:
@chris-rock we need to transform this resource to support running HTTP checks on the remote node eventually, and I think it's unlikely we'll be able to use Faraday for that. Instead, we'll likely have to try to use curl, wget, python, etc. Having the user pass in something that's Faraday-specific doesn't make sense in that case, and it also requires the user to know more about the implementation of the resource itself. I'm all for using Faraday locally, though.

I suggest we leave this as is and continue to expose the tunables that our users need but can also be support simultaneously between local and remote targets depending on how we implement them.

lib/resources/http.rb
adamleff: @chris-rock we need to transform this resource to support running HTTP checks on the remote node eventually, and I think it's unlikely we'll be able to use Faraday for that. Instead, we'll likely have to try to use curl, wget, python, etc. Having the user pass in something that's Faraday-specific doesn't make sense in that case, and it also requires the user to know more about the implementation of the resource itself. I'm all for using Faraday locally, though.

I suggest we leave this as is and continue to expose the tunables that our users need but can also be support simultaneously between local and remote targets depending on how we implement them.

chris-rock: Even if we are not using Faraday in future, I like the structure for the config `ssl: { verify: true }`. We already had a discussion about using `cert_path = "~/.chef/trusted_certs"` in the past. This `ssl` hash would allow us to easily add this support via `ssl: { verify: true , ca_path: "/usr/lib/ssl/certs" }`.

chris-rock: Is the default value `verify: true` also working with http connections?

adamleff: This feels a bit like YAGNI since we don't know exactly what the future looks like for this resource, but I'll defer to @chris-rock's guidance on whether to accept a hash or a boolean here.

chris-rock: @ElizabethU are you going to need the `ca_path` for your implementation?

chris-rock: Talked with @adamleff We are agreed that we go with the current implementation. Thank you @ElizabethU and @adamleff

chris-rock (approved):

PR #1669, opened 2017-04-18
fix web reference url

chris-rock:


adamleff (approved):
Thanks for beating me to the punch, @chris-rock :)

PR #1673, opened 2017-04-18
Update postgres.rb

kidbrax:
adding a check for /var/lib/postgresql/data which is used in their docker images

### DIFF
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -36,6 +36,9 @@ module Inspec::Resources
         version = if inspec.directory('/var/lib/pgsql/data').exist?
                     warn 'Found /var/lib/pgsql/data. Assuming postgresql install uses un-versioned directories.'
                     nil
+                  elsif inspec.directory('/var/lib/postgresql/data').exist?
+                    warn 'Found /var/lib/postgresql/data. Assuming postgresql install uses un-versioned directories. (This path is used by their docker images)'
+                    nil
                   else
                     version_from_dir('/var/lib/pgsql/')
                   end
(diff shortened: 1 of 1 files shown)


kidbrax:
Signed-off-by: Braxton Beyer <braxton@braxtonbeyer.com>

lib/resources/postgres.rb
```diff
@@ -36,6 +36,9 @@ def initialize
         version = if inspec.directory('/var/lib/pgsql/data').exist?
                     warn 'Found /var/lib/pgsql/data. Assuming postgresql install uses un-versioned directories.'
                     nil
+                  elsif inspec.directory('/var/lib/postgresql/data').exist?
```
chris-rock, line 39: Just to be sure, is that docker specific or os specific?

chris-rock (commented):
Thank you for your contribution @kidbrax  Can you share a little bit more feedback? Which operating system is involved?

lib/resources/postgres.rb
kidbrax: I'm not sure.  Now that you mention it, I assume it is OS specific.  It is Alpine-based.
https://github.com/docker-library/postgres/blob/3d4e5e9f64124b72aa80f80e2635aff0545988c6/9.4/alpine/Dockerfile

aaronlippold: @kidbrax @chris-rock @adamleff I folded this into PR #1758 so we can close both, it makes more sense to push the change there given the updates we made to the resource in general.

adamleff:
Closing this issue, superseded by #1758

PR #1681, opened 2017-04-20
fetch user groups while building user object

Happycoil:
Signed-off-by: Tor Magnus Rakvåg <tm@intility.no>

Fixes https://github.com/chef/inspec/issues/1400

### DIFF
--- a/test/integration/default/user_spec.rb
+++ b/test/integration/default/user_spec.rb
@@ -91,7 +91,7 @@ if os.windows?
     it { should exist }
     # should return the SID of the user
     its('uid') { should_not eq nil}
-    its('groups') { should include 'Administrators' }
+    its('groups') { should include userinfo[:groups] }
   end
 
   # also support simple username for local users without domain
@@ -99,7 +99,7 @@ if os.windows?
     it { should exist }
     # should return the SID of the user
     its('uid') { should_not eq nil}
-    its('groups') { should include 'Administrators' }
+    its('groups') { should include userinfo[:groups] }
   end
 else
   # test single `user` resource
--- a/test/integration/default/user_spec.rb
+++ b/test/integration/default/user_spec.rb
@@ -30,7 +30,7 @@ elsif os.windows?
     groupname: nil,
     uid: nil,
     gid: nil,
-    groups: nil,
+    groups: "Administrators",
     home: nil,
     shell: nil,
   }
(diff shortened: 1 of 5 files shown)


adamleff (approved):
@Happycoil Thank you for your contribution! I'm not a big PowerShell user but this looks OK to me.

Also, thanks for cleaning up some whitespace consistency issues. In the future, it's incredibly helpful to do all the "code cleanup" work in a separate commit on the same PR so it's easier to separate the "style" changes from the material "code" changes. We can take care of squashing those commits if deemed necessary before merge. But stuff like that is nicer as separate commits.

Thanks again for submitting this PR. Looking to @chris-rock or @arlimus for additional feedback as they have more experience with this resource and PowerShell than I do.

test/integration/default/user_spec.rb
```diff
@@ -91,13 +91,15 @@
     it { should exist }
     # should return the SID of the user
     its('uid') { should_not eq nil}
+    its('groups') { should include 'Administrators' }
```
chris-rock: this should be
```
its('groups') { should include userinfo[:username] }

```

test/integration/default/user_spec.rb
```diff
@@ -91,13 +91,15 @@
...
   end
 
   # also support simple username for local users without domain
   describe user('Administrator') do
     it { should exist }
     # should return the SID of the user
     its('uid') { should_not eq nil}
+    its('groups') { should include 'Administrators' }
```
chris-rock: Same as above

test/unit/resources/user_test.rb
```diff
@@ -103,7 +103,7 @@
     _(resource.uid).wont_be_nil
     _(resource.exists?).must_equal true
     _(resource.group).must_equal nil
-    _(resource.groups).must_equal nil
+    _(resource.groups).must_equal ['Administrators', 'Users']
```
chris-rock, line 106: We should add a test for `Guest` user to see that we receive `Users` only

chris-rock (commented):
This is a great addition @Happycoil Thank you! I left some minor improvements.

test/integration/default/user_spec.rb
Happycoil: Fixed.

test/integration/default/user_spec.rb
Happycoil: Fixed.

test/unit/resources/user_test.rb
Happycoil: Fixed.

chris-rock (approved):
Thank you @Happycoil

PR #1683, opened 2017-04-20
[www] Fix docs pages for x509_certificate and key_rsa

adamleff:
These files were not named correctly and the docs rake task
was not picking them up.


chris-rock (approved):
Thank you @adamleff

PR #1689, opened 2017-04-21
showing how to shellout in docs

rshade:


docs/ruby_usage.md
```diff
@@ -152,3 +152,31 @@ describe command('ls -al /') do
   its('exit_status') { should eq 0 }
 end
 ```
+
+## Shelling out in tests
+
+When writing tests you can not use standard ruby methods to shellout as it tries to execute those commands locally.
+However, the `command` resource has a `.stdout` method that will allow you to manipulate the results.
+Using the above example, you could check the writes on several subdirectories.
```
chris-rock, line 160: Maybe we should add an example for a test file, so that users can copy paste.

chris-rock (commented):
Great work @rshade This helps to make it more clear.

chris-rock (approved):
Looks like a great addition to me

adamleff (approved):
LGTM! 👍 Thanks for your contribution @rshade!

PR #1694, opened 2017-04-22
bugfix: solve warn on uninitialized

arlimus:
`@expectation` was analyzed without initializeation, leading to ruby warnings. fix it


chris-rock (approved):
Thank you @arlimus

Issue #1439, opened 2017-01-27
The `be_reachable` matcher in the Host resource should use the port/protocol provided


cullenmcdermott:
### Description

The `reachable?` method for the Host resource should do more than check if a host is reachable via ping. If ping is blocked then the test will fail.

### InSpec and Platform Version

Inspec 1.10.0
Ubuntu 16.04

### Replication Case

1) Setup a firewall rule that blocks icmp traffic from Host A to Host B
2) Create an simple test that checks reachability on Host A and run it:
```
describe host("#{Host_A}", port: 22, proto: 'tcp') do
    it { should be_reachable }
  end
```

### Expected Outcome

If I am specifying a port and protocol I expect that to be what is tested as sometimes icmp traffic is blocked by a firewall. 

### Possible Solutions

A possible solution would be to execute nc on the host running the test. However not all distributions come with nc installed so that could pose an issue.

A better solution might be to open a tcp(or udp) socket with bash.

The best solution would probably be to use [socket](https://ruby-doc.org/stdlib-2.3.1/libdoc/socket/rdoc/Socket.html) since that should be platform agnostic and work even if bash isn't available.

MooreDerek:
I'd be interested in helping to fix this, from the proposed solutions there are some interesting issues when looking at Windows and *nix support of available commands.

**Proposed Solution - \*nix**

Use, in order, *nc, wget, telnet, bash* then *ping*.  Use *exists* to determine if they are installed and report in the output which command was used.  

The main issue that I've come across in testing is the timeout on both *telnet* and *bash* when testing a port with no listener; *telnet* timeouts after 60 seconds and *bash* can take over 10 minutes.

*wget* and *telnet* do not provide udp support, but as it's not guaranteed delivery, not sure that's a big issue.

**Proposed Solution - Windows**

Use, in order, *Test-NetConnection, telnet* and then *ping*, using exists to verify installation and report in the output.

Main issue is *Test-NetConnection* is only available in Windows 8.1 and upwards, and *telnet* tends to be removed on a lot of enterprise Windows installs. Again timeout on *telnet* is a consideration.

Thoughts?

adamleff:
@MooreDerek I'm super excited you're willing to contribute to this, thank you!

Some of my thoughts:

[...]

PR #1698, opened 2017-04-23
feat: Adds port/ protocol support for host

walker233:
This update extends the host resource to use correct
port and protocol for the redhat and debian families.
It uses a feature from bash and the timeout command
from the coreutils packages.

When port is specified but protocol is omitted, it
assumes tcp is what was intended. When neither port
nor protcol is specified it falls back to ping the
host.

Tries to address issue #1439 to some degree.

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -121,11 +121,7 @@ module Inspec::Resources
   end
 
   class LinuxHostProvider < HostProvider
-    def ping(hostname, port = nil, proto = nil)
-      # assume tcp if port is given without corresponding protocol
-      if port && !proto
-        proto = 'tcp'
-      end
+    def ping(hostname, port = nil, proto = 'tcp')
       # use bash builtin capability for checking connectivity
       # for those os's with timeout from coreutils packages
       # TODO: Verify whether this works on other linux versions
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -121,12 +121,20 @@ module Inspec::Resources
   end
 
   class LinuxHostProvider < HostProvider
-    # ping is difficult to achieve, since we are not sure
-    def ping(hostname, _port = nil, _proto = nil)
-      # fall back to ping, but we can only test ICMP packages with ping
-      # therefore we have to skip the test, if we do not have everything on the node to run the test
-      ping = inspec.command("ping -w 1 -c 1 #{hostname}")
-      ping.exit_status.to_i != 0 ? false : true
+    def ping(hostname, port = nil, proto = nil)
+      #assume tcp if port is given without corresponding protocol
+      if port && !proto
+        proto = 'tcp'
+      end
+      #use bash builtin capability for checking connectivity
+      #for those os's with timeout from coreutils packages
+      # TODO: Verify whether this works on other linux versions
+      if port && ( inspec.os.redhat? || inspec.os.debian? )
+        resp = inspec.command("timeout 1 bash -c 'cat </dev/null >/dev/#{proto}/#{hostname}/#{port}'")
+      else
+        resp = inspec.command("ping -W 1 -c 1 #{hostname}")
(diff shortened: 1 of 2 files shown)


walker233:
I'm pretty sure that the one failed test job in Travis CI is from the modification. The error and failure should be investigated to that other do no fail in the same way.

lib/resources/host.rb
```diff
@@ -121,12 +121,20 @@ def resolve(hostname)
...
+    def ping(hostname, port = nil, proto = nil)
+      # assume tcp if port is given without corresponding protocol
+      if port && !proto
+        proto = 'tcp'
+      end
+      # use bash builtin capability for checking connectivity
+      # for those os's with timeout from coreutils packages
+      # TODO: Verify whether this works on other linux versions
+      if port && (inspec.os.redhat? || inspec.os.debian?)
+        resp = inspec.command("timeout 1 bash -c 'cat </dev/null >/dev/#{proto}/#{hostname}/#{port}'")
```
chris-rock, line 129: Should we check if bash is available? eg. via `command('bash').exist?`

lib/resources/host.rb
```diff
@@ -121,12 +121,20 @@ def resolve(hostname)
...
   class LinuxHostProvider < HostProvider
-    # ping is difficult to achieve, since we are not sure
-    def ping(hostname, _port = nil, _proto = nil)
-      # fall back to ping, but we can only test ICMP packages with ping
-      # therefore we have to skip the test, if we do not have everything on the node to run the test
-      ping = inspec.command("ping -w 1 -c 1 #{hostname}")
-      ping.exit_status.to_i != 0 ? false : true
+    def ping(hostname, port = nil, proto = nil)
+      # assume tcp if port is given without corresponding protocol
+      if port && !proto
```
chris-rock: should we just set the default to `tcp`?
```
def ping(hostname, port = nil, proto = 'tcp')
```

chris-rock (commented):
Thank you @walker233 for this improvement!

lib/resources/host.rb
walker233: I agree. That is a better way to do that.

lib/resources/host.rb
walker233: That would probably be safe, but I thought for some reason it was nearly universal in inclusion. At least on redhat and debian families. Is that not the case. By the same token we are going to be safe with bash, we should probably do the same thing for the timeout command, which is in both redhat coreutils and debian coreutils packages. And were you suggesting we'd modify as:
~~~
if port && (inspect.os.redhat? || inspec.os.debian?) && 
   command('bash').exist? && command('timeout').exist?
~~~
Or were you thinking of a different modification.

walker233:
@chris-rock I've been trying to get useful test cases but there is something I'm missing. I'm expecting a failure for 4 of the test cases when trying to connect to a tcp port on the host example.com, but I'm only getting positive indicators. I tried ports 8080 and now 9080 which all are returning true, when they are returning false in my environment. Any thoughts on why the host test come back as reachable?

chris-rock:
@walker233 I tried the following quickly:
```
inspec> cmd = command("bash -c 'cat </dev/null >/dev/tcp/localhost/80'") 
=> Command bash -c 'cat </dev/null >/dev/tcp/localhost/80'
inspec> cmd.stdout
=> ""
inspec> cmd.stderr
=> "bash: connect: Connection refused\nbash: /dev/tcp/localhost/80: Connection refused\n"
inspec> cmd.exit_status
=> 1
```
I would expect to get an error back for any unassigned port. Isn't that working on your side?

adamleff:
I haven't seen any movement on this PR, and we recently added support for using `nc` for testing TCP ports in the `host` resource in #1915. Therefore, I'm going to close this PR. Please let me know if we need to reopen this. Thank you!

PR #1705, opened 2017-04-25
inspec control.to_ruby to use newlines instead of `\n`

arlimus:
I.e. instead of printing them as:

```
desc "hello\nworld"
```

it would instead do:

```
desc "hello
world"
```


arlimus:
The next step will be to indent the description.
For that, however, I need to clean up a 2 year old bug... 😁

chris-rock (approved):
Thank you @arlimus This makes it so much nicer to read if the control text is generated!

arlimus:
No a lot prettier with: https://github.com/chef/inspec/pull/1711

Issue #1205, opened 2016-10-05
resource fail (exception) aborts all tests


chris-rock:
### Description

The ssl resource fails if it cannot connect to an external host https://github.com/chef/inspec/blob/master/lib/resources/ssl.rb#L55. This results in a strange behavior, if this is only one tests in a profile, since the full profile execution is aborted:

```
 b inspec exec https://github.com/chris-rock/acme-inspec-profile -t docker://67bcf1bf4019
Cannot determine host for SSL test. Please specify it or use a different target.
```
### InSpec and Platform Version

InSpec 1.0
### Possible Solutions

We should not use fail in the resource or handle resource failures as a control failure.

alexpop:
Chris, agree that a failure for this resource should not stop the entire run.

As an alternative to this:
  https://github.com/chef/inspec/blob/v1.1.0/lib/resources/ssl.rb#L55

I looked into using the `skip_resource` option we have:

``` ruby
  return skip_resource 'Cannot determine host for SSL test. Please specify it or use a different target.'
```

With that ^, targeting a docker container with this spec:

``` ruby
describe ssl(port: 443) do
  it { should be_enabled }
end

describe 5 do
  it { should be_positive }
end
```

results in:

``` ruby
Target:  docker://92222ab36457157f7be6ebefbd775860918ec131e61a1c0bd2c34e829039e647

  SSL/TLS on
     ○  Cannot determine host for SSL test. Please specify it or use a different target.
  5 should
     ✔  be positive

Test Summary: 1 successful, 0 failures, 1 skipped
```

This is because resources calling `skip_resource` in initialize will be skipped here:
  https://github.com/chef/inspec/blob/v1.1.0/lib/inspec/runner.rb#L221-L223

Problem is when we are calling a method of the ssl resource, like here:

``` ruby
describe ssl(port: 443).mymethod do
  it { should_not be_enabled }
end
```

`arg[0]` [here](https://github.com/chef/inspec/blob/v1.1.0/lib/inspec/runner.rb#L222) is the instance returned by `mymethod`

This prevents the `resource_skipped` attribute, set for the ssl instance, to be available to the instance retuned by `mymethod`.

A workaround for this in the ssl resource would be to define the method like:

``` ruby
def mymethod
  return self if resource_skipped
  'value'
end
```

This way `arg[0]` will be the ssl instance and have the `resource_skipped` attribute set.

[...]

PR #1706, opened 2017-04-25
handle json parse errors in docker resource

chris-rock:
Until #1205 we need to continue to handle errors in our resources. In this case, the mock backend was throwing an exception for inspec check if the following test was used:

```
  describe docker.version do
    its('Client.Version') { should cmp >= '17.03' }
    its('Server.Version') { should cmp >= '17.03' }
  end
```


arlimus (approved):
Great to see these errors being handled @chris-rock kudos!

PR #1710, opened 2017-04-25
document reference to other tools

arlimus:
* add link in readme, so users can CTRL-F for Serverspec quickly
* add the link to the podcast with more information in the website docs

collected from feedback here: https://github.com/chef/inspec/issues/229


chris-rock (approved):
Thank you very much @arlimus

PR #1713, opened 2017-04-25
describe the value ranges of `impact`

arlimus:
Fixes https://github.com/chef/inspec/issues/358

### DIFF
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -44,7 +44,10 @@ where
 
 * `'sshd-8'` is the name of the control
 * `impact`, `title`, and `desc` define metadata that fully describes the importance of the control, its purpose, with a succinct and complete description
-* `impact` is an float that measures the importance of the compliance results and must be a value between `0.0` and `1.0`.
+* `impact` is an float that measures the importance of the compliance results and must be a value between `0.0` and `1.0`. The value ranges are:
+    * `0.0 to <0.4` these are controls with minor criticality
+    * `0.4 to <0.7` these are controls with major criticality
+    * `0.7 to 1.0` these are critical controls
 * `tag` is optional meta-information with with key or key-value pairs
 * `ref` is a reference to an external document
 * `describe` is a block that contains at least one test. A `control` block must contain at least one `describe` block, but may contain as many as required


chris-rock (approved):
Great addition @arlimus `inspec check` will also return warning for the misuse.

aaronlippold (commented):
We may also want to note that they should be whole number values - an impact of 0.45 is not allowed. 

Also, this brings up another point, should we also allow a mapping so impact: minor, major or critical is acceptable, would this lead to the more general case to say that we could allow them to map there own verbs via a defined hash. Like:  

``` 
impact = 
{ critical=>high,
major=>medium,
minor=low,
1.0=>critical,
0.0=>N/A }
```

The keys are the internal inspec static values for the mapped ranges of impact and the values are the user 'remapped' values.

aaronlippold:
I would also suggest that we reserve 1.0 and 0.0 as special cases basically as 1.0 life stops and you do something now, 0.0 is totally ignore. In my use cases there is a large difference between a high and a critical - critical is reserved for truly pressing issues that will and without a doubt cause harm. 

I believe it is important to outline these to cases in the documentation as well. i.e. that our reporting will always put 1.0 issues on the top and 0.0 issues will be on the bottom for sorting, alerting etc.

aaronlippold:
I was suggesting that we put the ability to pass in a "mapping" object - like a hash perhaps. We are proposing a default mapping of 0-4 => minor 5-7 => major 7-10 => critical in our current discussion. My suggestion is that the default mapping be - i.e. these are the hard coded numbers and words in inspec - i.e. per our doc:  

```
0 => N/A
.1-.3=> minor
.4 - .6 => major
.7 - .9 => significant
1 => critical```

The reason I like this scale is that it natural supports sorting, i.e. for each of the three middle elements - there is a high, center and low - so you can sort each of the minor, major and significant elements above and below the others.

Mapping 0 to N/A supports our concept of "non-selected" controls. (https://github.com/chef/inspec/issues/1346)  and our concept of the control type that is a policy/paperwork/guns-gates-guards type control - i.e. the traditional 'non-technical' control.

Mapping 1 to critical allows us to override default sorting and reporting to ensure those issue are always "on top and visible.

The use case for this is:
1. I want to use the words 'high, medium and low' rather than, 'significant, major and minor'
2. I want to only use 0, low, high and 1 in my reporting https://www.first.org/cvss or just have 0, minor, major and 1. 

I propose that we focus on the first use case as we have seen it come up in more than one user use case and have been asked about it. If I am thinking about it correctly, we could perhaps just have a defined hash named: 'inspec_control_mappings' that could be put into the `inspec.yml` file.

PR #1717, opened 2017-04-26
print and prettyprint the inspec backend class

arlimus:
This is always bothersome when debugging code and drilling down objects, since it will just a return a two-layer anonymous class with no help at all:

```ruby
inspec> os.inspec
=> #<#<Class:0x000000036fc9f0>:0x0000000371e2d0>
```

"What is this? ^^" is the repeated question and "Where does it come from?".

Instead print a nice name and even give a bit of information on pretty-printing (which pry does naturally):

```ruby
inspec> os.inspec
=> Inspec::Backend::Class @transport=Train::Transports::Local::Connection
inspec> puts os.inspec
Inspec::Backend::Class
```

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -26,9 +26,20 @@ module Inspec
       end
 
       cls = Class.new do
+        # Ruby internal for printing a nice name for this class
+        def to_s
+          'Inspec::Backend::Class'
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @arlimus

PR #1718, opened 2017-04-26
provide the `inspec` keyword

arlimus:
Instead of my favorite shortcut of `os.inspec` just finally add it as a global keyword.

Preparation for https://github.com/chef/inspec/issues/1396


chris-rock (approved):
I think it is very handy and we should discourage the use of `os.inspec`. I think we should deprecate that. Lets cover that in another PR.

PR #1719, opened 2017-04-26
provide `inspec.version` information

arlimus:
as a friendly shortcut and a native call within profiles (i.e. instead of `Inspec::VERSION`)

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -36,6 +36,13 @@ module Inspec
           "Inspec::Backend::Class @transport=#{backend.class}"
         end
 
+        # Provide a shorthand to retrieve the inspec version from within a profile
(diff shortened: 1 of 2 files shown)


lib/inspec/backend.rb
```diff
@@ -36,6 +36,13 @@ def inspect
           "Inspec::Backend::Class @transport=#{backend.class}"
         end
 
+        # Provide a shorthand to retrieve the inspec version from within a profile
+        #
+        # @return [String] inspec version
+        def version
```
chris-rock, line 42: Great idea, this reminds me that we should make the meta-data accessible via `inspec.profile`. But lets do that later.

chris-rock (approved):
Great idea @arlimus

PR #1720, opened 2017-04-26
add `inspec.profile.file(...)` for profile files

arlimus:
Heavily depends on the outcome of the RFC discussion here: Fixes https://github.com/chef/inspec/issues/1396

WIP:

* [x] add docs

Fixes https://github.com/chef/inspec/issues/1396

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -17,6 +17,8 @@ A profile should have the following structure::
     │   └── control_etc.rb
     ├── libraries
     │   └── extension.rb
+    |── files
+    │   └── extras.conf
     └── inspec.yml
 
 where:
@@ -24,6 +26,7 @@ where:
 * `inspec.yml` includes the profile description (required)
 * `controls` is the directory in which all tests are located (required)
 * `libraries` is the directory in which all InSpec resource extensions are located (optional)
+* `files` is the directory with additional files that a profile can access (optional)
 * `README.md` should be used to explain the profile, its scope, and usage
 
 See a complete example profile in the InSpec open source repository: https://github.com/chef/inspec/tree/master/examples/profile
@@ -301,3 +304,39 @@ The following command runs the tests and applies the secrets specified in `profi
     $ inspec exec examples/profile-attribute --attrs examples/profile-attribute.yml
 
 See the full example in the InSpec open source repository: https://github.com/chef/inspec/tree/master/examples/profile-attribute
+
+# Profile files
+
+An InSpec profile may contain additional files that can be accessed during tests. This covers use-cases where e.g. a list of ports is provided to be tested.
+
+To access these files, they must be stored in the `files` directory at the root of a profile. They are accessed by their name relative to this folder with `inspec.profile.file(...)`.
+
+Here is an example for reading and testing a list of ports. The folder structure is:
+
+    examples/profile
+    ├── controls
+    │   ├── example.rb
+    |── files
+    │   └── services.yml
+    └── inspec.yml
+
+With `services.yml` containing:
+
+    - service_name: httpd-alpha
+      port: 80
+    - service_name: httpd-beta
+      port: 8080
+
+The tests in `example.rb` can now access this file:
+
+    my_services = yaml(content: inspec.profile.file('services.yml')).params
+
+    my_services.each do |s|
+      describe service(s['name']) do
+        it { should be_running }
+      end
+
+      describe port(s['port']) do
+        it { should be_listening }
+      end
+    end
--- a/lib/inspec/runtime_profile.rb
+++ b/lib/inspec/runtime_profile.rb
@@ -1,0 +1,26 @@
+# encoding: utf-8
+# Copyright 2017 Chef Software
+# author: Dominik Richter
+# author: Christoph Hartmann
+
+module Inspec
+  class RuntimeProfile
+    attr_reader :files
+
+    def initialize(profile)
+      @src = profile.source_reader
+      @files = @src.data_files.keys.map do |k|
+        k.sub('files' + File::SEPARATOR, '')
+      end
+    end
+
+    # Retrieve a profile file's contents
+    #
+    # @param name [String] the name of the file
+    # @return [String] contents of the file of RuntimeError if missing
+    def file(name)
+      @src.data_files[File.join('files', name)] ||
+        raise("Cannot find file #{name} in profile.")
+    end
+  end
+end
--- a/lib/source_readers/flat.rb
+++ b/lib/source_readers/flat.rb
@@ -20,13 +20,14 @@ module SourceReaders
       new(target, files)
     end
 
-    attr_reader :metadata, :tests, :libraries
+    attr_reader :metadata, :tests, :libraries, :data_files
 
     def initialize(target, files)
       @target = target
       @metadata = ::Inspec::Metadata.new(nil)
       @tests = load_tests(files)
       @libraries = {}
+      @data_files = {}
     end
 
     private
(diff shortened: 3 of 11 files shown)


docs/profiles.md
```diff
@@ -301,3 +304,35 @@ The following command runs the tests and applies the secrets specified in `profi
     $ inspec exec examples/profile-attribute --attrs examples/profile-attribute.yml
 
 See the full example in the InSpec open source repository: https://github.com/chef/inspec/tree/master/examples/profile-attribute
+
+# Profile files
+
+An InSpec profile may contain additional files that can be accessed during tests. This coveres use-cases where e.g. a list of ports is provided to be tested.
```
adamleff: `s/coveres/covers`

adamleff: This feels like a semi-bad example of why you'd want to use a file because something like a port list is easy-enough to put in the control itself.

Maybe something like "This allows a profile to include a file that contains many different fields and values that could be used through the profile in across mulitple control files without including that data in each control file."

docs/profiles.md
```diff
@@ -301,3 +304,35 @@ The following command runs the tests and applies the secrets specified in `profi
...
+
+To access these files, they must be stored in the `files` directory at the root of a profile. They are accessed by their name relative to this folder with `inspec.profile.file(...)`.
+
+Here is an example for reading and testing a list of ports. The folder structure is:
```
adamleff: If you decide to change the wording, maybe the example could be a YAML file like so:

```yaml
  web_alpha:
    service_name: httpd-alpha
    port: 80
  web_beta:
    service_name: httpd-beta
    port: 8080
```

... and the test could be:

```ruby
services = YAML.parse(inspec.profile.file('services.yml'))

services.each do |svc|
  describe service(svc['name']) do
    it { should be_running }
  end
  describe port(svc['port']) do
    it { should be_listening }
  end
end
```

Sorry if I'm being pedantic, but I wouldn't want to encourage profiles becoming too abstracted and difficult to follow. I'd want the new `file` functionality to be used when it's really truly necessary.

lib/source_readers/flat.rb
```diff
@@ -20,13 +20,14 @@ def self.resolve(target)
       new(target, files)
     end
 
-    attr_reader :metadata, :tests, :libraries
+    attr_reader :metadata, :tests, :libraries, :files
```
adamleff: I think it's going to get super-confusing for newcomers to this codebase that we call this "files" but it only includes files in the `files` directory, when the controls are stored in files also.

Should we call them "supplemental_files" or "data_files" or something of that nature?

lib/inspec/runtime_profile.rb
```diff
@@ -0,0 +1,31 @@
...
+# Copyright 2017 Chef Software
+# author: Dominik Richter
+# author: Christoph Hartmann
+
+module Inspec
+  class RuntimeProfile
+    def initialize(profile)
+      org = profile.source_reader.files
+      @files = org.each_with_object({}) do |(k, v), acc|
+        acc[k.sub('files' + File::SEPARATOR, '')] = v
```
adamleff: Do we need to prepopulate this hash with the files AND their contents? The SourceReader is already doing that, and then we do it again here just to have different keys... feels like we're consuming 2x the memory we need for files that may not even get used by the controls.

It feels like we should let the SourceReader provide the list of files and provide a method to return the contents, and all this class has to do is take a parameter, prefix it with "files/" and then pass it to the SourceReader to get us the content... and to list the files, it can strip "files/" before it hands it back to the caller.

arlimus: 👍 will implement this thank you!

lib/source_readers/flat.rb
arlimus: 👍 yeah had a similar feeling, just not a good name. i'll go with `data_files`, can't thing of anything else and it looks good

docs/profiles.md
arlimus: Funny enough, I have taken the ports exactly from a customer example 😁 I also see why they want to use it that way and it's ok from my perspective.

I also agree with your premise! I wanted to avoid parsing a file to keep it really simple, but it might not hurt that much. I'll take your example, thank you for sharing.

adamleff: Totally your call. I'd be pretty frustrated if I came across a profile that had values like that in a separate file... just feels like unnecessary abstraction.

If those values were used in 3+ different control files, then sign me up, it sounds great. If it's just in one, I don't understand the benefit.

docs/profiles.md
```diff
@@ -301,3 +304,39 @@ The following command runs the tests and applies the secrets specified in `profi
...
+
+    examples/profile
+    ├── controls
+    │   ├── example.rb
+    |── files
+    │   └── services.yml
+    └── inspec.yml
+
+With `ports.conf` containing:
```
chris-rock: I assume this is `services.yml` now?

docs/profiles.md
```diff
@@ -301,3 +304,39 @@ The following command runs the tests and applies the secrets specified in `profi
...
+
+    - service_name: httpd-alpha
+      port: 80
+    - service_name: httpd-beta
+      port: 8080
+
+The tests in `example.rb` can now access this file:
+
+    my_services = YAML.load(inspec.profile.file('services.yml'))
```
chris-rock: should we use:
```
my_services = yaml(content: inspec.profile.file('services.yml'))
```

chris-rock (commented):
How is that working with inherited profiles? Can I access files from inherited profiles?

Will a control still work, if we use it in combination with `include_controls`?

docs/profiles.md
arlimus: ah fail... thank you!

arlimus:
Good points on inheritance @chris-rock . Short answer: It should work, down the line. It doesn't work, right now. The reason is the source-reader and how it handles dependencies (at least in a couple of corner-cases I tested). Love to address this separately.

chris-rock:
Thank you @arlimus for clarification. This should not block the feature, we just need to be aware of the limitation.

chris-rock (approved):
Thank you for the quick turnaround @arlimus

docs/profiles.md
arlimus: great point!

aaronlippold:
This would also cover use cases where we need to specify things like : approved_accounts or approved_x data structure for login accounts or debase roles, valid or invalid connection types, etc. In the end, basically any data structure you could want to test, or file to compare etc. Good add.

aaronlippold:
Given https://github.com/CrunchyData/postgresql-baseline/issues/9 as a use case, one thing that we may want to support with this is the ability to also pass in a file on the command line or during a run. 

I can also see this location being a variable that may come from chef or automate or even gitlab.

Simple use:
1. I have an {mydb}-approved_account_exceptions.yml file that I pass to the general DB profile to use
2. {mydb} changes as do the approved exceptions from DB to DB

Advanced Use:
In Automate or in the larger CI/CD pipeline - from env to env - the required support files may have a standard format but - using @arlimus examples - the services file and ports may change between environments.

aaronlippold:
@arlimus @chris-rock @adamleff I love this new option, one question, what happens when I want to create a file in the files directory called 'approved users and connections' - the YAML file has a standard structure and I want users to fill in this file so that the generalized control I wrote in the profile will run with respect to their target system. I.e. the file name will stay the same, the structure, only the data changes relative to the system it is being run on, would you say then that we would use attributes to over-ride that data or would you want to pass in the new YAML file?

arlimus:
We had a small chat about ^^ on the phone: Some aspects may be handled via profile files, usually those that are universal and shared with the profile itself. Things that are less universal are recommended to be handled via attributes.

PR #1723, opened 2017-04-26
rename SimpleConfig / parse_config / parse_config_file options

arlimus:
See https://github.com/chef/inspec/issues/1709
Fixes https://github.com/chef/inspec/issues/1709


adamleff (approved):
Small changes like this make a world of difference for new users willing to become contributors. Thank you so much, @arlimus!

Let's get a brand new issue created in the repo, attached to the 2.0 milestone, to remove the old configuration keys. We can use that milestone as a TODO list for this type of deprecation.

PR #1726, opened 2017-04-26
use cmp in port docs instead of eq

arlimus:
Fixes https://github.com/chef/inspec/issues/438


adamleff (approved):
LGTM! Good to merge once tests go green. Merging pre-tests-going-green since these are just docs updates.

PR #1727, opened 2017-04-27
Update postgresql conf resource to accept include_dir as a string as well as an array

elliott-davis:
Even though I couldn't find any docs around include_dir accepting anything other than a string I left the existing functionality alone.
This forces include_dir to check multiple directories as well as single string directories for additional conf files.

Signed-off-by: Elliott Davis <edavis@chef.io>

### DIFF
--- a/lib/resources/postgres_conf.rb
+++ b/lib/resources/postgres_conf.rb
@@ -93,7 +93,7 @@ module Inspec::Resources
     def include_files(params)
       include_files = params['include'] || []
       include_files += params['include_if_exists'] || []
-      dirs = params['include_dir'] || []
+      dirs = Array(params['include_dir']) || []
       dirs.each do |dir|
         dir = File.join(@conf_dir, dir) if dir[0] != '/'
         include_files += find_files(dir, depth: 1, type: 'file')


lib/resources/postgres_conf.rb
```diff
@@ -93,7 +93,8 @@ def read_content
     def include_files(params)
       include_files = params['include'] || []
       include_files += params['include_if_exists'] || []
-      dirs = params['include_dir'] || []
+      dirs = []
+      dirs.push(params['include_dir']) unless params['include_dir'].nil?
```
adamleff: Another option is to leave the original `||` syntax and do this:

```ruby
dirs = Array(params['include_dir']) || []
```

If `params['include_dir']` is an array, it will leave it. If it's a string, it will become a single-element array. If it's nil, we get an empty array.

schisamo (approved):
Any chance we could get some test coverage that ensures this fixes the issue you uncovered?

lib/resources/postgres_conf.rb
elliott-davis: TIL - I'll update it

adamleff (approved):

PR #1730, opened 2017-04-27
Rakefile updates to support appbundle-updater

adamleff:
InSpec is "appbundled" into the ChefDK which means users can't
simply run `chef gem install inspec` and get an updated InSpec.
`appbundle-updater` exists to combat this and, when run, will
pull the source and install the gem using `rake install`.

In the appbundle-updater config, we'll want to exclude unnecessary
gems from getting installed, such as the "tests" group. Our Rakefile
doesn't work without the "tests" group being installed.

This change makes the Rakefile usuable when the "tests" gems aren't
available.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,19 +4,26 @@
 require 'bundler'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
-require 'rubocop/rake_task'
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @adamleff

PR #1735, opened 2017-04-28
Habitat profile bug fixes and improvements

rhass:
* Add `core/hab` to `pkg_deps` in hab plan.
  This fixes a bug in which tests needing to execute things like `hab pkg path myorigin/mypath` in the can profile/test can successfully execute the command. This was caused by the runtime of `hab pkg exec chef/inspec` which changes the path for the inspec runtime to match that of the inspec hab package.

* Fixed bug with install step where profile would include the .hart files from previous builds.
* Added ruby to pkg_deps rather than adding it to the path via the runhook. This ensures `hab pkg exec chef/inspec` behaves in the same way as the run hook.
* Updated the generated plan to support plan.sh syntax changes in habitat 0.21.0 and later by removing the `pkg_source` and the `do_download`, `do_verify`, and `do_unpack` overrides.
* Updated the generate run hook to leverage habitat to perform most of the origin, package name, and path variable interpolations.

/cc @schisamo @elliott-davis

### DIFF
--- a/lib/bundles/inspec-habitat/profile.rb
+++ b/lib/bundles/inspec-habitat/profile.rb
@@ -347,9 +347,8 @@ mkdir -p {{pkg.svc_var_path}}/inspec_results
 while true; do
   echo "Executing InSpec for ${PROFILE_IDENT}"
   hab pkg exec chef/inspec inspec exec {{pkg.path}}/dist --format=json > ${RESULTS_FILE} 2>${ERROR_FILE}
-  RC=$?
 
-  if [ "x${RC}" == "x0" ]; then
+  if [ $? -eq 0 ]; then
     echo "InSpec run completed successfully."
   elsif [ -s ${ERROR_FILE} ]
     echo "InSpec run did NOT complete successfully. Error:"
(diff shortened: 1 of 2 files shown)


lib/bundles/inspec-habitat/profile.rb
```diff
@@ -336,23 +332,21 @@ def run_hook_contents
...
 #!/bin/sh
 
-export PATH=${PATH}:$(hab pkg path core/ruby)/bin
-
 # InSpec will try to create a .cache directory in the user's home directory
 # so this needs to be someplace writeable by the hab user
 export HOME={{pkg.svc_var_path}}
 
-PROFILE_IDENT="#{habitat_origin}/#{package_name}"
+PROFILE_IDENT="{{pkg.origin}/{{pkg.name}}"
```
adamleff: Missing a `}` for `{{pkg.origin}}`... without it, the sup fails with:

```
Failed to load hook: hab-sup(ER)[src/error.rs:355:8]: TemplateError(TemplateError { reason: InvalidSyntax, template_name: Some("hook"), line_no: Some(7), column_no: Some(28) })
```

... and then the sup fails because the run hook never got copied to `/hab/svc`

```
Service start failed: hab-sup(ER)[src/error.rs:379:8]: No such file or directory (os error 2)
```

adamleff (changes requested):
@rhass Love it. Need to fix one breakage though and then we should be good.

adamleff (approved):

schisamo (approved):

chris-rock (approved):
Thank you @rhass

arlimus:
Nice kudos @rhass and everyone else for the quick review!! 😁

PR #1736, opened 2017-04-28
Improve "Which Serverspec resources are available in InSpec?"

shortdudey123:
- Switch serverspec and inspec resource compare to a table and fix the links
- Add links for serverspec resources not in inspec
- update inspec.io links to HTTPS in migration doc

@adamleff this is the update to the migration doc page that i talked to you about a couple weeks ago on slack

### DIFF
--- a/docs/reference/migration/index.html
+++ b/docs/reference/migration/index.html
@@ -12,11 +12,125 @@
 
 <p>The following resources are available in InSpec:</p>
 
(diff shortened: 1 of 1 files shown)


chris-rock:
@shortdudey123 Thank you for this addition. I think it is a great idea to use a table here. Unfortunately you have edited the html files which is auto-generated. Would it be possible to transfer the same to our markdown file? This is located here: https://github.com/chef/inspec/blob/master/docs/migration.md

Is there anything to prevent this in future? Eg. document where the docs are located?

adamleff (changes requested):
@shortdudey123 Thanks, Grant! So glad to have you as a contributor here. 🙂 

Unfortunately, @arlimus is correct. The `gh-pages` branch is auto-generated via a Rake task we run. The link he provided is the Markdown doc that needs to be modified which will then get translated to HTML during our release process.

I think the right thing to do for now is to close this PR (even though you can change the target branch) and allow you the opportunity to make the changes in Markdown instead. Are you OK with that approach?

shortdudey123:
Oh! i completely missed that, sorry!  let me close this and i will copy the info into the md files.

PR #1739, opened 2017-04-28
[www] Add rake task for flushing Fastly cache

adamleff:
inspec.io is now hosted by Fastly which operates as a caching reverse
proxy. This change adds a Rake task that will flush the cache at Fastly
if the API key and Service ID are set in the appropriate environment
variables.

Flushing the cache is not required, so this will not error out if the
environment variables aren't provided; the cache will eventually clear
on its own. This will simply speed along the process as needed.


arlimus (approved):
Great to have this as a nice task! Thank you

PR #1741, opened 2017-04-28
Error and exit when using --sudo locally

### LINKED ISSUE #1690: Warn when using --sudo locally
The `--sudo` flag should only be used with remote scan targets... when scanning locally, native Ruby methods are used and escalating privileges is difficult if not impossible.

The CLI should warn a user when they use `--sudo` without `--target`.

adamleff:
Running `inspec exec` with --sudo locally produces unintended results
given that we cannot escalate local Ruby methods after we're already
running. --sudo is meant to only be used with remote targets. We do
not currently enforce that.

This change will print an error for the user if they attempt to use
--sudo with a local exec and exit non-zero.

Fixes #1690

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -155,6 +155,13 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
     configure_logger(opts)
     o = opts.dup
 
+    # print error if user passed --sudo but with no --target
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Great catch and improvement, thank you @adamleff 👍 😄

chris-rock (approved):
Thank you @adamleff

1moere1:
any reason the option `--host` option is ignored here?

PR #1747, opened 2017-05-01
Fix YAML resource documentation

adamleff:
The `yaml` resource documentation was missing many clues that users
need to successfully use the resource.


arlimus (approved):
Great catch @adamleff thank you!!

PR #1749, opened 2017-05-02
Make the --no-color flag work for inspec exec

### LINKED ISSUE #1024: --no-color switch not working when invoked with "inspec exec"
### Description

the --color/--no-color switch is not working when invoked with inspec exec.
It outputs in color anyway.

CentOS Linux release 7.2.1511 (Core)

Inspec versions 0.32.0, 0.33.0
### Replication Case

Run:

```
inspec exec test.rb --no-color
```

Result: the printed output in terminal is still colored.
### Possible Solutions

None.
### Stacktrace

None.

adamleff:
The CLI formatter is not currently honoring the --no-color flag
when outputting CLI output. This change cleans up how we format
with color and properly support the flag for use cases where
color-encoding characters make the output difficult to use
(i.e. when someone redirects CLI output to a text file for
sharing with others).

I also added the 'm' gem to the `test` group in the Gemfile, a
super-handy gem for running the minitest tests in a single file,
or a single test in a file.

Fixes #1024

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ group :test do
   gem 'ruby-progressbar', '~> 1.8'
   gem 'webmock', '~> 2.3.2'
   gem 'jsonschema', '~> 2.0.2'
+  gem 'm'
 end
 
 group :integration do
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -531,7 +531,7 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
   #
   def print_control(control)
     print_line(
-      color:      COLORS[control.summary_indicator] || '',
+      color:      control.summary_indicator,
       indicator:  INDICATORS[control.summary_indicator] || INDICATORS['unknown'],
       summary:    format_lines(control.summary, INDICATORS['empty']),
       id:         "#{control.id}: ",
@@ -541,7 +541,6 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
 
   def print_result(result)
     test_status = result[:status_type]
-    test_color = COLORS[test_status]
     indicator = INDICATORS[result[:status]]
     indicator = INDICATORS['empty'] if indicator.nil?
     if result[:message]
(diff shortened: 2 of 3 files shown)


lib/inspec/rspec_json_formatter.rb
```diff
@@ -586,43 +585,54 @@ def print_anonymous_control(control)
...
+    output.puts(format_with_color(fields[:color], format_line(fields)))
   end
 
   # Helps formatting summary lines (called from within print_line arguments)
   def format_lines(lines, indentation)
     lines.gsub(/\n/, "\n" + indentation)
   end
 
+  def format_with_color(color_name, text)
+    return text unless RSpec.configuration.color
```
chris-rock, line 630: Is there any reason we are going to support this?

Gemfile
```diff
@@ -20,6 +20,7 @@ group :test do
   gem 'ruby-progressbar', '~> 1.8'
   gem 'webmock', '~> 2.3.2'
   gem 'jsonschema', '~> 2.0.2'
+  gem 'm'
```
chris-rock, line 23: any reason why we need that?

lib/inspec/rspec_json_formatter.rb
adamleff: This is currently what passing `--no-color` modifies to indicate the user's intent to not use color. There is also no other way to pass data (other than profiles) all the way down to the formatter, so I think this is a reasonable approach for this iteration.

Gemfile
adamleff: I explained my intention in the PR message. This is a super easy way to run a single minitest test file, or even a single test in a file. If you don't want it in there, that's fine, but it makes for a better dev experience.

lib/inspec/rspec_json_formatter.rb
chris-rock: Okay, makes sense. Its time to make reporting independent or RSpec. Nothing we can solve in that PR though.

Gemfile
chris-rock: Those who are able to read are at an advantage indeed :-)

chris-rock: Can we add docs in our readme for that? Next time, I would prefer to have such things in separate PRs. No big deal though.

chris-rock (approved):
Quick turnaround. Awesome work @adamleff

PR #1750, opened 2017-05-02
Add support for Windows auth in mssql_resource

nsdavidson:
This adds supports for connecting to MS SQL instances using Window
authentication rather than SQL authentication.  By leaving either the
user or password parameters blank causes the sqlcmd to leave off the -U
and -P params.  This will cause sqlcmd to authenticate as the current
Windows user.

### DIFF
--- a/docs/resources/mssql_session.md.erb
+++ b/docs/resources/mssql_session.md.erb
@@ -10,13 +10,13 @@ Use the `mssql_session` InSpec audit resource to test SQL commands run against a
 
 A `mssql_session` resource block declares the username and password to use for the session, and then the command to be run:
 
-    describe mysql_session('username', 'password').query('QUERY') do
+    describe mssql_session(user: 'username', pass: 'password').query('QUERY') do
       its('output') { should eq('') }
     end
 
 where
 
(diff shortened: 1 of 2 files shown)


nsdavidson:
@arlimus

aaronlippold (changes requested):
there is a small typo in the ##Syntax section. `describe mysql_session(` should be `describe mssql_session(` correct? 

Also, same issue below that we ran into with the postgres and mysql session resources. 

```Open a Command Prompt window, and type sqlcmd -S myServer\instanceName. Replace myServer\instanceName with the name of the computer and the instance of SQL Server that you want to connect to.``` https://technet.microsoft.com/en-us/library/ms166559(v=sql.105).aspx.

I would suggest that if sqlcmd allows the definition of a host an 'instance'/'tenant' - we should provide the ability for the user to specify that but default them to either nil or localhost if that is the 80%-90% use case but we shouldn't assume they will also be the case.

nsdavidson:
Thanks for the feedback @aaronlippold!  I switched over to an option hash and made the instance and host names configurable.

arlimus:
Thank you for adding support for MSSQL @nsdavidson !!

I'll close this MR here and cleaned it up and update here: https://github.com/chef/inspec/pull/1786
Could you have a look if this is ok for you?

Kudos again :)

PR #1751, opened 2017-05-02
Add an oracledb_session resource

nsdavidson:
This adds an oracle_session resource similar to the existing resource
for MySQL and MSSQL.  It assumes the sqlplus tool is installed and in
the path of the user InSpec connects as.


nsdavidson:
@arlimus

aaronlippold (commented):
You may want to consider adding:
```def initialize(user = nil, pass = nil, host = localhost, service = nil)``` and `... | sqlplus -s #{@user}/#{@pass}@{@host}/#{@service}")`. This will default the value to localhost but as we found in the postgres resource, it is needed and useful sometime to be able to not force the host value on the backend to just be localhost, especially in the context of a remote test.

**Review by aaronlippold — COMMENTED** [review]
And also, don't forget to document that you can pass in the host and session values in your examples, but that they default to 'localhost' and to 'nil'.

**Review by aaronlippold — CHANGES_REQUESTED** [review]
Sorry, forgot to mark this request changes.

File: lib/resources/oracle_session.rb [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+      @user = user
+      @pass = pass
+      @service = service
+      return skip_resource("Can't run Oracle checks without authentication") if user.nil? or pass.nil?
+    end
+
+    def query(q)
+      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"').gsub(/\$/, '\\$')
+
+      cmd = inspec.command("echo \"#{escaped_query}\" | sqlplus -s #{@user}/#{@pass}@localhost/#{@service}")
```
chris-rock: we need to be able to parametrize the sqlplus bin. e.g. in a oracle docker container https://store.docker.com/images/oracle-database-enterprise-edition this is located in `/u01/app/oracle/product/12.1.0/dbhome_1/bin/sqlplus`

lib/resources/oracle_session.rb
```diff
@@ -0,0 +1,39 @@
...
+    name 'oracle_session'
+    desc 'Use the oracle_session InSpec resource to test commands against an Oracle database'
+    example "
+      sql = oracle_session('my_user','password')
+      describe sql.query('SELECT NAME FROM v$database;') do
+        its('stdout') { should_not match(/test/) }
+      end
+    "
+
+    def initialize(user = nil, pass = nil, service = nil)
```
chris-rock: Do we need to support DB_SID?

chris-rock: Not 100% sure yet, but should we use hash options here like:
```
oracle_session(user: 'my_user', password: 'password')
```

chris-rock (commented):
This is a great addition to inspec @nsdavidson
Do you think we can have some kind of integration testing? At least we should add some unit testing similar to https://github.com/chef/inspec/blob/master/test/unit/resources/mysql_conf_test.rb

lib/resources/oracle_session.rb
aaronlippold: Yes you need to support DB_SID

I agree, a hash makes sense.

lib/resources/oracle_session.rb
aaronlippold: If we are going to parametrize, let's parametrize! :) 👍 

```def query(q)
+      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"').gsub(/\$/, '\\$')
+
+      cmd = inspec.command("echo \"#{escaped_query}\" | {@sqlplus_bin} {@sqlplus_opts} #{@user}/#{@pass}@{@host}/#{@service}")
```

nsdavidson:
Thanks for the feedback @aaronlippold and @chris-rock.  I *believe* I addressed most of your suggestions :).

arlimus (approved):
This is a great start, thank you @nsdavidson !!

Also kudos to the reviews from @chris-rock and @aaronlippold .

PR #1764, opened 2017-05-04
Update changelog generator to work from PRs only

adamleff:
To clean up our Changelog, this change modifies the use of
github-changelog-generator to only use PRs. This ensures
that our Changelog only reflects actual changes and makes
it easier for our users to see the actual changes made to
the codebase.

docs and www changes will be excluded from the Changelog as well.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@
 require 'bundler'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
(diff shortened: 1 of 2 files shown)


adamleff:
Hold off on merging this one - it's still not 100% the way I want it... I'll have a change pushed up shortly after I do the v1.23.0 release.

PR #1773, opened 2017-05-04
Ignore web acceptance travis errors

adamleff:
The website acceptance job in Travis will fail for non-chef-github
users because we use encrypted environment variables. This job
should never hold up an accepted PR, so I'm moving it to the
allowed_failures section.


chris-rock (approved):
Thank you @adamleff

PR #1775, opened 2017-05-05
Handle parse errors for attrs/secrets

adamleff:
Inspired by #1640, this change cleans up the logic used when
reading in secrets files, provides clearer warnings when the
secrets files can't be parsed, and adds tests for those methods.


arlimus (approved):
Great to have these exceptions handled! Kudos @adamleff

chris-rock (approved):
That looks awesome. Thank you @adamleff and @mr-exz

PR #1785, opened 2017-05-09
bugfix: dev-sec/ssl-benchmark -> ssl-baseline

arlimus:
It was renamed upstream, we missed to update it.


chris-rock (approved):
Thank you @arlimus

PR #1786, opened 2017-05-09
Add support for Windows auth in mssql_resourcet

arlimus:
This adds supports for connecting to MS SQL instances using Window
authentication rather than SQL authentication.  By leaving either the
user or password parameters blank causes the sqlcmd to leave off the -U
and -P params.  This will cause sqlcmd to authenticate as the current
Windows user.

rebased from https://github.com/chef/inspec/pull/1750

### DIFF
--- a/docs/resources/mssql_session.md.erb
+++ b/docs/resources/mssql_session.md.erb
@@ -1,0 +1,79 @@
+---
+title: About the mssql_session Resource
+---
+
+# mssql_session
+
+Use the `mssql_session` InSpec audit resource to test SQL commands run against a Microsoft SQL database.
+
+## Syntax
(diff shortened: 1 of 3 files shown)


chris-rock (changes requested):
@nsdavidson Great addition. Can we add a unit test as well?

chris-rock:
Going forward I see us using https://hub.docker.com/r/microsoft/mssql-server-windows-express/ for integration testing

arlimus:
I'll add the unit tests, since this is a hard rebase. Good point!

chris-rock (approved):
Thank you @nsdavidson and @arlimus

PR #1791, opened 2017-05-10
Update hab exporter to use inspec in path over calling to hab sup

elliott-davis:
Signed-off-by: Elliott Davis <edavis@chef.io>

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -14,7 +14,6 @@ pkg_deps=(
   core/libxslt
   core/ruby
   core/net-tools
-  core/hab
 )
 pkg_build_deps=(
   core/bundler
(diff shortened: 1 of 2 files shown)


habitat/plan.sh
```diff
@@ -14,7 +14,6 @@ pkg_deps=(
   core/libxslt
   core/ruby
   core/net-tools
-  core/hab
```
adamleff, line 17: Yeah, I can totally see how this is going to cause problems if we leave it in here, and profiles depend on it... especially depending on the workstation that build the InSpec package itself. I'm :+1: removing this and immediately rebuilding the InSpec package on the depot.

adamleff (approved):

Issue #1005, opened 2016-09-01
inspect detect should additionally output virtualization platform


nvtkaszpir:
### Description

As user when using inspec detec command on remote host I'd like to be able to see virtualization platform.
### InSpec and Platform Version

inspec 0.30.0
### Replication Case

In LXC:

``` bash
17:08:50 $ inspec detect -t ssh://root@10.0.3.251 --password root

== Operating System Details

Name:      debian
Family:    debian
Release:   7.9
Arch:      x86_64
```

Expected output:

``` bash
17:08:50 $ inspec detect -t ssh://root@10.0.3.251 --password root

== Operating System Details

Name:      debian
Family:    debian
Release:   7.9
Arch:      x86_64
Virtual:   lxc
```
### Possible Solutions

It could leverage using virt-what http://git.annexia.org/?p=virt-what.git;a=summary

Maybe extend inspec os module with more detailed info https://docs.chef.io/inspec_os.html ? 
But I think it should be as separate module.
### Stacktrace

No stacktrace.

chris-rock:
@nvtkaszpir That is a great idea. Once we add this, we should thank about the detection of multiple virtual platforms. I would propose we implement this in two steps:
- add the InSpec resource to detect the virtual platform (this would allow us to use that information in InSpec tests as well)
- include this resource in InSpec detect
  A PR is welcome to add this feature to InSpec.

lamont-granquist:
There's going to be a lot of duplication with ohai here.  Highly encourage you to look at that code, since detecting virtualization and cloud platforms is something that we've been fighting with there for years now.

chris-rock:
Awesome @lamont-granquist Lets see how we can reuse that.

aaronlippold:
Hi, I second this as it would apply to the generalization and implementation of profiles for all profiles - such as the STIG and CIS - that will have different control implications given they are inside a container or not. 

With respect to @lamont-granquist comment - the code in ohai is here : https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/virtualization.rb#L163

kekaichinose:
__What’s happening? Why was this issue closed?__
This issue was closed due to some much needed review of legacy issues or issues that were spawned in older versions of InSpec, i.e. < v3.

__Why do I care?__
You would care about this if this was an issue that you are still seeing and/or feel needs to be addressed in the current version of InSpec.

__What do I need to do?__
[...]

PR #1803, opened 2017-05-11
Add support for virtualization resource

tkak:
This PR adds support for virtualization resource to detect the virtualization platform. (#1005)

Examples:

```
case virtualization[:system]
when 'docker'
  describe ...
    ...
  end
when 'vbox'
  describe ...
    ...
  end
end
```


lib/resources/virtualization.rb
```diff
@@ -0,0 +1,228 @@
...
+      end
+    "
+
+    def initialize()
+      unless inspec.os.linux?
+        skip_resource 'The `virtualization` resource is not supported on your OS yet.'
+      end
+    end
+
+    def [](name)
```
chris-rock: I think we should use `role`, `system` and `systems` as proper methods, since this is more similar to other resources

chris-rock (commented):
@tkak This is a great addition. Thank you @tkak Can you add documentation to this PR as well? The docs files are located in https://github.com/chef/inspec/tree/master/docs/resources

tkak:
@chris-rock Thank you for your feedback! I've fixed my PR.

docs/resources/virtualization.md.erb
```diff
@@ -0,0 +1,71 @@
...
+<%= partial "/shared/matcher_match" %>
+
+## Examples
+
+The following examples show how to use this InSpec audit resource.
+
+### Test for Docker
+
+    describe virtualization.system do
+      it { should eq 'docker' }
```
adamleff: I'd recommend using `its` here like the example below if not just to be consistent and to ensure the user gets good CLI formatter output. Otherwise, it will say something like "expected docker to == vbox" rather than being clear that the `virtualization` resource is the item under test.

docs/resources/virtualization.md.erb
```diff
@@ -0,0 +1,71 @@
...
+
+    describe virtualization do
+      its('system') { should eq 'vbox' }
+      its('role') { should eq 'guest' }
+    end
+
+### Detect the virtualization platform
+
+    if virtualization.system == 'vbox'
+      describe package('name')do
```
adamleff: Missing a space between `)` and `do`

lib/resources/virtualization.rb
adamleff: Since you have proper methods defined below around like 45/46, can we remove this method now?

lib/resources/virtualization.rb
```diff
@@ -0,0 +1,246 @@
...
+    def docker_exists?
+      inspec.command('docker').exist?
+    end
+
+    def nova_exists?
+      inspec.command('nova').exist?
+    end
+
+    def collect_data_linux # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/AbcSize
+      virtualization = Hashie::Mash.new unless virtualization
```
adamleff: `virtualization` appears to be a local variable in this method, is that true? If so, I'm unsure when the `unless` would ever get hit.

adamleff: Since `collect_data_linux` is used by each of the three defined accessor methods, it could be called multiple times in a single test (i.e. the user had multiple `its` statements). Should we cache the data in an instance variable and then return it if it's defined? That would eliminate multiple runs through this method.

lib/resources/virtualization.rb
```diff
@@ -0,0 +1,246 @@
...
+
+    def docker_exists?
+      inspec.command('docker').exist?
+    end
+
+    def nova_exists?
+      inspec.command('nova').exist?
+    end
+
+    def collect_data_linux # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/AbcSize
```
adamleff: Wow, one heck of a method!

My main concern is that it tests for EVERY virtualization possibility, even if one has already matched. For example, if we detect the host is a Xen host, we'll still test for vbox, openstack, etc.

I wonder if it would make more sense to break apart each of the detections into smaller methods that return true if it matched, false if it didn't... example:

```
def detect_xen
  return false unless inspec.file('/proc/xen').exist?
  @virtualization_data[:system] = 'xen'
  # etc...
  true
end
```

Then `collect_data_linux` could be written like so:

```
def collect_data_linux
  # cache data in an instance var to avoid doing multiple detections for a single test
  @virtualization_data ||= Hashie::Mash.new
  return unless @virtualization_data.empty?

  # each detect method will return true if it matched and was successfully
  # able to populate @virtualization_data with stuff.
  return if detect_xen
  return if detect_openstack
  return if detect_foo
end
```

I think this would make the code a bit more readable/manageable, and will also make it easier to test with each system detection being its own method.

What do you think, @tkak?

adamleff (changes requested):
@tkak thank you so much for your contribution! This appears to be your first time contributing to InSpec as well, so welcome to the community!

I left some feedback that I think will tidy up this new resource and make it easier to understand, add unit tests, etc. Please let me know what you think and how we can help.

Again, thanks again for your contribution!

lib/resources/virtualization.rb
tkak: @adamleff : I agree with you. I referred the [ohai code](https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/virtualization.rb). I will try to fix the method to be smaller. Thank you for your review!

adamleff: @tkak ah ha! That makes sense. I think the ohai implementation could be improved too, but when running in InSpec against a remote system, it's even more critical to address this so we don't remotely check file status more than we need to, etc.

Looking forward to your update!

tkak:
@adamleff Thank you for your review. I made some changes. Would you be able to check them?

test/integration/default/virtualization_spec.rb
```diff
@@ -0,0 +1,13 @@
+# encoding: utf-8
+if ENV['DOCKER']
```
adamleff: We could simplify this to:

```
if ENV['DOCKER'] && os.linux
   # rest of code here
end
```

... and eliminate one indentation level. Not critical to fix here, just something for the future :)

lib/resources/virtualization.rb
```diff
@@ -0,0 +1,275 @@
...
+        return false
+      end
+      true
+    end
+
+    def collect_data_linux # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
+      # cache data in an instance var to avoid doing multiple detections for a single test
+      @virtualization_data ||= Hashie::Mash.new
+      return unless @virtualization_data.empty?
+      @virtualization_data[:systems] ||= Hashie::Mash.new
```
adamleff: In the original Ohai implementation, I think the `@virtualization_data[:systems]` hash was intended to collect all the virtualization data in case a host fell into more than one category. I don't think that's the case, and this likely will confuse users. Can I suggest that we remove this and all entries in this Hash? I think the primary use cases here are best served with the `system` and `role` keys that exist at the top-level. What do you think?

adamleff (changes requested):
@tkak thanks for your patience! Last week for ChefConf for us, so we're still catching up.

I think this is super-close. I have one comment/question/suggestion about `@virtualization_data[:systems]` and I think that should do it!

lib/resources/virtualization.rb
```diff
@@ -0,0 +1,275 @@
...
+
+      describe virtualization do
+        its('role') { should eq 'guest' }
+      end
+
+      describe virtualization.systems[:docker] do
+        it { should eq 'guest' }
+      end
+
+      if virtualization.system == 'docker'
```
chris-rock: I would prefer to have a control block with only_if here:

```
control 'test' do
  describe file('/var/tmp/foo') do
    it { should be_file }
  end
  only_if { virtualization.system == 'docker' }
end

```

aaronlippold: @chris-rock @tkak Thanks this closes a long standing request of mine when we started developing the os profiles. Great work! :) 👍 

Agree, it is also a great way to show an example of good use of `only_if` as it is intended. Can't wait to start using this resource in the OS security profiles.

lib/resources/virtualization.rb
tkak: @adamleff OK, I'll remove `@virtualization_data[:systems]` hash.  :)

adamleff:
@tkak well done! This is a great addition. Thanks for working with us to make this PR great.

adamleff (approved):
💯 

**NOTE TO NEXT APPROVER:** this should be squashed via GitHub UI instead of merged, please.

arlimus (approved):
Awesome contribution, thank you so much @tkak !! 😄 

Kudos @adamleff and @chris-rock for your reviews

PR #1806, opened 2017-05-11
minor ui fix

echohack:
Fixing a minor problem with this error message. Copy+paste of the error will work better with this output.


echohack:
I'm an idiot. Resubmitting with dco later

PR #1809, opened 2017-05-11
Fix version string in changelog generator

adamleff:
When I made the change to pull the version from ENV (as is used
during the bump_version rake task), I neglected to prefix it with
a "v", which causes the "full changelog" links to be incorrect.


chris-rock (approved):
Thank you @adamleff

PR #1812, opened 2017-05-12
Remove controls from JSON formatter that have no results

### LINKED ISSUE #1810: Inherited controls listed in multiple profiles in JSON formatter output
### Description

When a profile `include_controls` from another profile, the included controls are listed in both the original and the including profile in the JSON formatter output.

InSpec 1.24.0, CentOS 7 and macOS

### Replication Case

Given this profile:

```ruby
control 'inherited-1.0' do
  title 'an inherited control'
  desc 'checking to see if this shows up more than once'
  impact 0.5

  describe file('/tmp/foo.yaml') do
    it { should exist }
  end
end
```

... which is `depend`ed on as "inherited" in a profile with this control file:

```ruby
include_controls 'inherited'
```

... `inherited-1.0` is listed in both the inheriting and inherited profiles output in the JSON formatter, but is missing results in the profile output for the profile that called `include_controls`:

https://gist.github.com/adamleff/236866527b7e8fd221f9cfc372f34cdd

### Possible Solutions

The CLI formatter keeps the inherited controls listed in the inherited profile output, not the profile that called `include_controls`. While this is likely not ideal for when a user overrides an aspect of the original control (or even skips it), we should keep the output the same and refactor the reporters at a later date.

adamleff:
When profile dependencies/inheritance is used, controls from
depended-on profiles are tracked in the "including profile" run
context as well as the "depended-on profile." When you ask the
"including profile" for its info, it returns its controls as well
as controls from the "depended-on profile."

In the JSON formatter, this causes these depended-on controls to
be output twice, once with results and once without. Systems that
consume this JSON can incorrectly show these controls as skipped.

Until we dig into how best to separate profile run contexts, I'm
proposing removing any controls from profile output that have no
results. This will better align it with the CLI formatter and
avoid downstream systems from seeing the same control twice.

Fixes #1810

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -123,6 +123,19 @@ class InspecRspecJson < InspecRspecMiniJson # rubocop:disable Metrics/ClassLengt
       control = example2control(example)
       move_example_into_control(example, control)
     end
+
+    # Until we revamp the reporting infrastructure and properly
+    # handle controls from inherited profiles correctly, we
+    # need to remove any controls without results from profiles
+    # to avoid the JSON reporter outputting duplicate controls.
+    # This will make it match the output shown in the CLI
+    # formatter.
+    #
+    # A longer-term fix will involve looking into profile run
+    # contexts and ensure we don't duplicate the tracking of
+    # controls between profiles when inheritance/dependencies
+    # are used.
+    delete_controls_without_results_from_profile
   end
 
   private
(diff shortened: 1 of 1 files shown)


elliott-davis (approved):
tenor-45349855

adamleff:
@chris-rock @arlimus pinging you again for review on this one.

chris-rock:
I agree that we need to find a solution for this problem, but I am not sure this is the right approach. I'd like to find a consensus of the `right` approach first

adamleff:
Because this would uncover edge cases of controls with no tests getting dropped from results, I'm going to close this PR for now and we can revisit another time. Issue #1810 will be left open to track the original reported issue.

Issue #1580, opened 2017-03-21
inspec json/exec missing control code data


alexpop:
### Description

The output of `inspec json` or `inspec exec --format json` is missing the code details for controls when using an archived profile.

### InSpec and Platform Version

```
$ bundle exec inspec version
1.16.1
$ bundle exec inspec shell -c os.params
{:name=>"mac_os_x", :family=>"darwin", :release=>"10.11.6", :arch=>"x86_64"}
```

### Replication Case

```
[05:28:40 /tmp/profile]$ wget https://github.com/nathenharvey/tmp_compliance_profile/archive/master.zip

[05:28:51 /tmp/profile]$ unzip master.zip

[05:28:56 /tmp/profile]$ ll
-rw-r--r--   1 apop  wheel  6443 Mar 21 17:28 master.zip
drwxr-xr-x   7 apop  wheel   238 Jan 14  2016 tmp_compliance_profile-master

[05:29:24 ~/git/inspec {master}]$ bundle exec inspec json /tmp/profile/master.zip
{"name":"tmp_compliance_profile","title":"/tmp Compliance Profile","summary":"An Example Compliance Profile","version":"0.1.1","maintainer":"Nathen Harvey <nharvey@chef.io>","license":"Apache 2.0 License","copyright":"Nathen Harvey <nharvey@chef.io>","supports":[],"controls":[{"title":"A /tmp directory must exist","desc":"A /tmp directory must exist","impact":0.3,"refs":[],"tags":{},"code":"","source_location":{"ref":"tmp_compliance_profile-master/controls/tmp.rb","line":3},"id":"tmp-1.0"},{"title":"/tmp directory is owned by the root user","desc":"The /tmp directory must be owned by the root user","impact":0.3,"refs":[],"tags":{},"code":"","source_location":{"ref":"tmp_compliance_profile-master/controls/tmp.rb","line":12},"id":"tmp-1.1"}],"groups":[{"title":"/tmp Compliance Profile","controls":["tmp-1.0","tmp-1.1"],"id":"controls/tmp.rb"}],"attributes":[]}

[05:29:44 ~/git/inspec {master}]$ bundle exec inspec json /tmp/profile/tmp_compliance_profile-master/
[...]

PR #1816, opened 2017-05-16
read source code if profile is in tgz/zip

arlimus:
To note: the source location between archived files and an uncompressed profiel is still different. When the block gets processed the location for uncompressed files will have the full path on the system, while archives still only carry their mocked paths. This can be improved in a separate PR. This PR respects this distinction and only targets the empty code fields.

This issue has been bugging me since InSpec's original inception 😁

### DIFF
--- a/lib/inspec/method_source.rb
+++ b/lib/inspec/method_source.rb
@@ -1,0 +1,24 @@
+# encoding: utf-8
+# author: Dominik Richter
+# author: Christoph Hartmann
+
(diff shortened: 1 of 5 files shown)


chris-rock (approved):
Great improvement @arlimus

chris-rock:
Fixes #1580

arlimus:
Working on the windows issue, know where it is...

PR #1819, opened 2017-05-16
support new automate compliance backend

chris-rock:


lib/bundles/inspec-compliance/api.rb
```diff
@@ -21,15 +28,21 @@ def self.profiles(config)
...
           mapped_profiles = []
           profiles.values.each { |org|
             mapped_profiles += org.values
           }
+        # Chef Automate pre 0.8.0
+        elsif config['server_type'] == 'automate' && config['version'].empty?
+          mapped_profiles = profiles.values.flatten
+        else
+          owner_id = config['user']
+          mapped_profiles = profiles.map { |e|
```
adamleff: Stylistically, we should ensure multiline blocks be of the `do...end` variety to remain consistent.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -21,15 +28,21 @@ def self.profiles(config)
         msg = 'success'
         profiles = JSON.parse(data)
         # iterate over profiles
-        if config['server_type'] == 'automate'
-          mapped_profiles = profiles.values.to_a.flatten
-        else
+        if config['server_type'] == 'compliance'
```
adamleff: We're doing this `if... elsif...else` expressions a few times throughout this change. Can I propose some helper methods?

```
def is_compliance_server?(config)
  config['server_type'] == 'compliance'
end

def is_automate_server_pre_080?(config)
  config['server_type'] == 'automate' && config['version'].empty?
end

def is_automate_server_080_and_later?(config)
  config['server_type'] == 'automate' && !config['version'].empty?
end

def is_automate_server?(config)
  config['server_type'] == 'automate'
end
```

That way we're only defining the logic in one place and it's easier to add unit tests.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -43,14 +56,19 @@ def self.profiles(config)
...
-    def self.version(url, insecure)
+    def self.version(config)
+      url = config['server']
+      insecure = config['insecure']
+
       if url.nil?
         puts "
 Server configuration information is missing.
 Please login using `inspec compliance login https://compliance.test --user admin --insecure --token 'PASTE TOKEN HERE' `
 "
```
adamleff: I know this line isn't changed in your PR, but shouldn't we be exiting here if we don't have the config? We seem to continue on, and it looks like `data` will never be an assigned variable and we'll likely throw an exception.

```
if url.nil?
  puts 'warning message'
  exit 1
end
# reach out to compliance API as needed
```

lib/bundles/inspec-compliance/api.rb
```diff
@@ -72,8 +90,17 @@ def self.exist?(config, profile)
     end
 
     def self.upload(config, owner, profile_name, archive_path)
-      # upload the tar to Chef Compliance
-      config['server_type'] == 'automate' ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/owners/#{owner}/compliance/#{profile_name}/tar"
+      # Chef Compliance
+      if config['server_type'] == 'compliance'
```
adamleff: See earlier comment about using helper methods here instead of repeating the `config[]` logic.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -151,12 +178,23 @@ def self.get_token(config)
 
     def self.target_url(config, profile)
       if config['server_type'] == 'automate'
-        target = "#{config['server']}/#{profile}/tar"
+        owner, id = profile.split('/')
```
adamleff: The splitting of the profile into owner and ID is now redundantly defined. We could eliminate the redundancy:

```
owner, id = profile.split('/')
if is_automate_server?(config)
   # etc...
```

lib/bundles/inspec-compliance/api.rb
```diff
@@ -151,12 +178,23 @@ def self.get_token(config)
...
+        target = "#{config['server']}/profiles/#{owner}/#{id}/tar"
       else
         owner, id = profile.split('/')
         target = "#{config['server']}/owners/#{owner}/compliance/#{id}/tar"
       end
       target
     end
+
+    # returns a parsed url for `admin/profile` or `compliance://admin/profile`
+    def self.sanitize_profile_name(profile)
```
adamleff, line 192: I'm sure I'm missing something with this method... The end state is to return `'admin/profile'` regardless of whether it was supplied with a URI scheme or not, correct? It seems like we don't need to prepend a non-schemed URI with a scheme just to remove the scheme from the string. In fact, that's kinda what was being done in an earlier version of this method is removed later in this PR. 

Could this method be simplified to:

```
uri = URI(profile)
"#{uri.host}#{uri.path}"
```


```
irb(main):021:0* a = URI('compliance://admin/profile')
=> #<URI::Generic:0x007fc51b0b1980 URL:compliance://admin/profile>
irb(main):022:0> b = URI('admin/profile')
=> #<URI::Generic:0x007fc51b0a8420 URL:admin/profile>
irb(main):023:0> "#{a.host}#{a.path}"
=> "admin/profile"
irb(main):024:0> "#{b.host}#{b.path}"
=> "admin/profile"
```

Like I said, I'm sure I'm missing something here so feel free to educate me!

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -237,16 +237,13 @@ def upload(path) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, Percei
     desc 'version', 'displays the version of the Chef Compliance server'
     def version
       config = Compliance::Configuration.new
-      if config['server_type'] == 'automate'
-        puts 'Version not available when logged in with Automate.'
+      info = Compliance::API.version(config)
+      if !info.nil? && info['version']
```
adamleff, line 244: It might be nice to have `Compliance::API.version` return an empty hash instead of `nil` to avoid having to do this extra check.

adamleff (commented):
I don't think any of my comments are blocking, but my opinion is that they'd make the code cleaner and easier to test if implemented.

lib/bundles/inspec-compliance/api.rb
chris-rock: We can definitely do that. Even after my fix, this module needs massive refactoring. It started with 1 server version, and we have essentially 2 versions Chef Compliance and 2 versions Chef Automate now. That needs to be abstracted.

lib/bundles/inspec-compliance/api.rb
chris-rock: API implementation may throw an error but will never exit, this can only be done by CLI

lib/bundles/inspec-compliance/api.rb
chris-rock: I just learned yesterday that Automate is using fancy usernames like `adam@leff.com`, which breaks the current parsing since `adam@` is not part of host. I agree that we need way more testing around this

lib/bundles/inspec-compliance/cli.rb
chris-rock: I like that

lib/bundles/inspec-compliance/api.rb
```diff
@@ -6,12 +6,22 @@
...
     # return all compliance profiles available for the user
     def self.profiles(config)
-      config['server_type'] == 'automate' ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/user/compliance"
+      # Chef Compliance
+      if is_compliance_server?(config)
+        url = "#{config['server']}/user/compliance"
+      # Chef Automate
+      elsif is_automate_server?(config)
+        url = "#{config['server']}/profiles/#{config['user']}"
+      end
```
arlimus: Let's catch the third case i.e. `else raise ...` becuse there is no url instead of getting the nil error further down

arlimus (approved):
Thank you Chris for adding all of this!
Ditto Adam for the review, I agree this needs more work.
My comment isn't critical for the merge, but needs to land at some point (afaics).

chris-rock:
@arlimus will add that quickly...

lib/bundles/inspec-compliance/api.rb
chris-rock: yes

lib/bundles/inspec-compliance/api.rb
```diff
@@ -6,12 +6,24 @@
...
     def self.profiles(config)
-      config['server_type'] == 'automate' ? url = "#{config['server']}/#{config['user']}" : url = "#{config['server']}/user/compliance"
+      # Chef Compliance
+      if is_compliance_server?(config)
+        url = "#{config['server']}/user/compliance"
+      # Chef Automate
+      elsif is_automate_server?(config)
+        url = "#{config['server']}/profiles/#{config['user']}"
+      else
+        raise ServerConfigurationMissing
```
adamleff, line 24: I'd just add a string message to this to make it complete and give the troubleshooter an additional clue.

```
raise ServerConfigurationMissing, 'No `server_type` entry in the config, unable to determine server type.'

File: lib/bundles/inspec-compliance/api.rb [hunk 10]
```diff
@@ -43,16 +61,17 @@ def self.profiles(config)
...
-"
-      else
-        response = Compliance::HTTP.get(url+'/version', nil, insecure)
-        data = response.body
-      end
+    def self.version(config)
+      url = config['server']
+      insecure = config['insecure']
+
+      raise ServerConfigurationMissing if url.nil?
```
adamleff, line 68: Same here...

```
raise ServerConfigurationMissing, 'No `server` entry in config, unable to determine server URL.' if url.nil?
```

lib/bundles/inspec-compliance/api.rb
adamleff: Ah, I understand. I'd love for us to use URI smarter than just deleting the scheme from the string, but we can tackle that another time.

adamleff (approved):
I'd obviously love to see some unit tests for this bundle but understand that time is not on our side. At least factoring out some of the logic into helper methods will make it easier to do so in the future.

Thanks for fixing this all up, @chris-rock.

adamleff:
@chris-rock don't forget to fix up the DCO issue with your third commit before we merge.

chris-rock:
Once @alexpop approves, I squash the commits

alexpop (approved):
Works great, thanks Chris!

PR #1820, opened 2017-05-17
Improvements to Habitat plan

### LINKED ISSUE #1582: Inspec installed via Habitat errors on SSL
### Description

When running `hab pkg exec chef/inspec inspec exec https://github.com/dev-sec/ssl-baseline` I get the error `/hab/pkgs/core/ruby/2.3.1/20161214031900/lib/ruby/2.3.0/net/http.rb:933:in `connect_nonblock': SSL_connect returned=1 errno=0 state=error: certificate verify failed (OpenSSL::SSL::SSLError)`

[root@automate ~]# hab pkg exec chef/inspec inspec --version
1.16.1

[root@automate ~]# cat /etc/centos-release
CentOS Linux release 7.3.1611 (Core) 

### Replication Case
```
  2  curl -LO https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh && sudo bash ./install.sh
    3  hab package install chef/inspec
    4  hab pkg exec chef/inspec inspec --version
    5  hab pkg exec chef/inspec inspec exec https://github.com/dev-sec/ssl-baseline
```

### Possible Solutions

### Stacktrace
https://gist.github.com/stephenlauck/b9ccbf0fbcccaa09efd78d94aa754e9c

smith:
These are kind of all over the place, but should improve things:

* Use the new `pkg_version` mechanism to set the version, and fail if
  the VERSION file is not present
* Use inspec.io for the upstream url
* Remove pkg_source and it's associated callbacks; they aren't required
  any more
* Alphabetize the deps list
* Remove duplicate coreutils from build deps
* Move environment variable setting to `do_prepare`
* Delete all binstubs in bin that aren't inspec
* Put the generated Gemfile in $CACHE_PATH so it doesn't stomp on the
  developer's Gemfile
* Insert the SSL_CERT_FILE env var in the binstub (Fixes #1582)
* Use install instead of cp to drop off Gemfile.lock

Works for me on Habitat 0.23.

Since all the "building" is done now in `do_install`, it would be
possible to define a `do_check` that runs `inspec exec` on profiles to
verify inspec is working by running inspec.

Signed-off-by: Nathan L Smith <smith@chef.io>

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -1,68 +1,85 @@
 pkg_name=inspec
 pkg_origin=chef
-pkg_version=$(cat "VERSION")
 pkg_description="InSpec is an open-source testing framework for infrastructure
   with a human- and machine-readable language for specifying compliance,
   security and policy requirements."
-pkg_upstream_url=https://github.com/chef/inspec
+pkg_upstream_url=https://www.inspec.io/
 pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
 pkg_license=('Apache-2.0')
-pkg_source=false
 pkg_deps=(
+  core/busybox-static
+  core/cacerts
   core/coreutils
   core/libxml2
   core/libxslt
-  core/ruby
   core/net-tools
-  core/busybox-static
+  core/ruby
 )
 pkg_build_deps=(
   core/bundler
-  core/coreutils
   core/gcc
   core/make
   core/readline
+  core/sed
 )
 pkg_bin_dirs=(bin)
 
-do_download() {
-  return 0
+pkg_version() {
+  cat VERSION
 }
 
-do_verify() {
-  return 0
-}
-
-do_unpack() {
-  return 0
+do_before() {
+  if [[ ! -f VERSION ]]; then
+    exit_with "habitat/VERSION not found. Run 'rake release_habitat' to ensure it exists" 1
+  fi
+  update_pkg_version
 }
 
 do_prepare() {
-  # Create a Gemfile with what we need
-  cat > Gemfile <<GEMFILE
+  # If we use the Gemfile in the project ($SRC_PATH/Gemfile), and run `bundle
+  # install` (which we do in `do_build` below), Bundler will write out a
+  # .bundle/config into the $SRC_PATH. If this happens, if you try to use
+  # InSpec outside of the build process it will fail, since the settings used
+  # in this plan will be saved in the .bundle/config.
+  #
+  # Instead, we build a minimal Gemfile in the $CACHE_PATH and bundle using that.
+  cat > "$CACHE_PATH/Gemfile" <<GEMFILE
 source 'https://rubygems.org'
-gem '$pkg_name', '= $pkg_version'
-gem 'rake'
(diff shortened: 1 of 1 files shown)


adamleff:
Can't currently verify due to ongoing public Depot issues.  Will revisit this later. This seems pretty good to me at first glance though.

habitat/plan.sh
```diff
@@ -1,68 +1,76 @@
...
+  if [[ ! -f VERSION ]]; then
+    exit_with "habitat/VERSION not found. Run 'rake release_habitat' to ensure it exists" 1
+  fi
+  update_pkg_version
 }
 
 do_prepare() {
   # Create a Gemfile with what we need
-  cat > Gemfile <<GEMFILE
+  cat > "$CACHE_PATH/Gemfile" <<GEMFILE
```
elliott-davis: I've always been confused by this step. We have a gemfile in the main repo, why recreate one here? I think this means we will always be building the version in rubygems and never in master.

adamleff: I think that's intended currently. Ultimately I'd rather this be part of a CI pipeline that cut a release anytime we promoted to rubygems, but seeing as how these tasks can sometimes be async, this feels like the safest way for now.

adamleff (approved):
Thanks for the contribution @smith!

adamleff:
Going to hold on merging this until after 1.25 is released, will get this in next week's 1.26 release.

habitat/plan.sh
smith: @elliott-davis the main reason for this is that there is now way to bundle install without having bundler write out a .bundle/config to the same directory where the Gemfile is located, so if you build using the "real" Gemfile, then you won't be able to bundle anything on your local machine, because bundler will save those settings.

I'm pushing up a commit where we use `path: $SRC_PATH` instead of `= $pkg_version`. This will make it use the current version on disk rather than pulling from rubygems.

smith: Updated in d030d74.

habitat/plan.sh
```diff
@@ -1,68 +1,85 @@
...
+  if [[ ! -f VERSION ]]; then
+    exit_with "habitat/VERSION not found. Run 'rake release_habitat' to ensure it exists" 1
+  fi
+  update_pkg_version
 }
 
 do_prepare() {
-  # Create a Gemfile with what we need
-  cat > Gemfile <<GEMFILE
+  # If we use the Gemfile in the project ($SRC_PATH/Gemfile), and run `bundle
```
elliott-davis, line 39: I think this is avoidable: https://github.com/habitat-sh/core-plans/blob/master/scaffolding-ruby/lib/scaffolding.sh#L162-L192 but it's fine for this. It'd be cool to get the ruby scaffolding working for ruby bins

elliott-davis (approved):

habitat/plan.sh
smith: I guess that's another solution, but I wouldn't necessarily call it a better one :)

chris-rock (approved):
Thank you @smith for this improvement

PR #1826, opened 2017-05-17
Habitat Profiles: redirect stderr to stdout

adamleff:
Due to habitat-sh/habitat#2395, we shouldn't try to log stderr output
to a file for now. While this makes for a less-than-awesome UX, it's
better than a process locking up due to a buffer filling up!

This change redirects stderr from InSpec to stdout and adds some
helpful troubleshooting messages. Should InSpec be able to generate
unique exit codes for when controls fail (vs. a Ruby eval failure)
then we can fix this up some more, too.

### DIFF
--- a/lib/bundles/inspec-habitat/profile.rb
+++ b/lib/bundles/inspec-habitat/profile.rb
@@ -332,6 +332,12 @@ do_install() {
       <<-EOL
 #!/bin/sh
 
+# redirect stderr to stdout
+# ultimately, we'd like to log this somewhere useful, but due to
+# https://github.com/habitat-sh/habitat/issues/2395, we need to
+# avoid doing that for now.
+exec 2>&1
+
 # InSpec will try to create a .cache directory in the user's home directory
 # so this needs to be someplace writeable by the hab user
 export HOME={{pkg.svc_var_path}}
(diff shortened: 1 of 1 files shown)


lib/bundles/inspec-habitat/profile.rb
```diff
@@ -346,16 +352,15 @@ def run_hook_contents
 
 while true; do
   echo "Executing InSpec for ${PROFILE_IDENT}"
-  inspec exec {{pkg.path}}/dist --format=json > ${RESULTS_FILE} 2>${ERROR_FILE}
+  inspec exec {{pkg.path}}/dist --format=json > ${RESULTS_FILE}
```
elliott-davis, line 354: Now that we're deleting this can you also remove https://github.com/chef/inspec/pull/1826/files#diff-93d8a0fc1b64e361ffb0725d8ed10a37L342

adamleff: Done.

elliott-davis (approved):
tenor-222272103

chris-rock (approved):
While I would prefer to have stdout and stderr in separate stream, I understand that we need this combined at this point

PR #1832, opened 2017-05-18
throw an error during inspec check if the version is not correct

chris-rock:
This ensures that inspec check fails if a non-semver version is used in inspec.yml

### DIFF
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -6,6 +6,7 @@
 require 'logger'
 require 'rubygems/version'
 require 'rubygems/requirement'
+require 'semverse'
 
 module Inspec
   # Extract metadata.rb information
@@ -109,6 +110,12 @@ module Inspec
         next unless params[field.to_sym].nil?
         errors.push("Missing profile #{field} in #{ref}")
       end
+
+      # if version is set, ensure it is correct
+      if !params[:version].nil? && !valid_version?(params[:version])
+        errors.push('Version needs to be in SemVer format')
+      end
+
       %w{ title summary maintainer copyright }.each do |field|
         next unless params[field.to_sym].nil?
         warnings.push("Missing profile #{field} in #{ref}")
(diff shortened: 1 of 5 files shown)


lib/inspec/metadata.rb
```diff
@@ -109,6 +110,12 @@ def valid
         next unless params[field.to_sym].nil?
         errors.push("Missing profile #{field} in #{ref}")
       end
+
+      # if version is set, ensure it is correct
+      if !params[:version].nil? && !valid_version?(params[:version])
+        errors.push("Version need to be in SemVer format in #{ref}")
```
alexpop: `needs` here?

lib/inspec/metadata.rb
```diff
@@ -109,6 +110,12 @@ def valid
...
+      end
+
       %w{ title summary maintainer copyright }.each do |field|
         next unless params[field.to_sym].nil?
         warnings.push("Missing profile #{field} in #{ref}")
```
adamleff, line 121: Since the `inspec check` output will provide the filename that generated the warning:

```
✖  /Users/aleff/profiles/test1/inspec.yml:0:0: Version need to be in SemVer format in inspec.yml
```

... I propose can just change this to "Version needs to be in SemVer format" to avoid the duplicate "inspec.yml" output

lib/inspec/metadata.rb
```diff
@@ -123,6 +130,13 @@ def valid?
       errors.empty? && unsupported.empty?
     end
 
+    def valid_version?(value)
+      Semverse::Version.new(value)
+      true
+    rescue Semverse::InvalidVersionFormat => _ex
```
adamleff: Since we're not using the exception object, we can remove the `=> _ex` from here.

adamleff (changes requested):
Two quick/small changes and then I'm good. :)

adamleff (approved):

PR #1835, opened 2017-05-19
Bump default timeouts for `http` resource

schisamo:
This changes the default read and open timeouts to be 60 seconds which matches the defaults for `Net::HTTP` backend which Faraday uses by default:
https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#read_timeout-attribute-method
https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#open_timeout-attribute-method

The current timeout values are too small which causes tests to be flakey.

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -24,15 +24,16 @@ module Inspec::Resources
       end
     "
 
-    # rubocop:disable ParameterLists
-    def initialize(url, method: 'GET', params: nil, auth: {}, headers: {}, data: nil, ssl_verify: true)
+    def initialize(url, opts = {})
       @url = url
-      @method = method
-      @params = params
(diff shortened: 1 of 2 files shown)


adamleff:
I imagine there might be some people counting on those smaller timeouts. I'm totally cool with changing the defaults, but I wonder if now's the time we expose a `read_timeout` and `open_timeout` parameter to the resource that defaults to your newly-propose options.

adamleff (approved):
Thanks @schisamo! Sorry you were experiencing flaky tests, but this will help you and existing/future users too. I appreciate it!

chris-rock (approved):
Great improvement @schisamo

PR #1838, opened 2017-05-19
Allow docs Rake task to be run from outside www dir

adamleff:
The docs Rake task requires classes defined in the "shared"
Rake tasks file. However, only the www Rakefile includes the
"shared" tasks file.

Since the "docs" Rake task is what needs it, I'm adding a require
there to ensure it works from outside the www directory.


chris-rock (approved):
Thank you @adamleff

PR #1851, opened 2017-05-25
Amended the processes resource to skip on windows

### LINKED ISSUE #1571: Give error when using Process Resource on Windows
### Description

The Process Resource is not supported on Windows.  Using the resource will simply cause a failed test case rather than an error or warning that it is not supported on the OS.  It would be nice to either have it implemented in Windows (powershell supports the ps command it is using for linux) or give an error message saying that it is not supported.

### Replication Case
The following code will fail but there is no message that this isn't supported on Windows.

```
describe processes('explorer') do
    its('entries.length') { should eq 1 }
    its('users') { should eq ['username'] }
end
```

username-is-already-taken2:
Fixes #1571 

This is more of a patch then a fix as I haven't added support for windows but I have amended the resource to skip if windows is detected.

### DIFF
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -32,6 +32,8 @@ module Inspec::Resources
       @list = all_cmds.find_all do |hm|
         hm[:command] =~ grep
       end
+
+      return skip_resource 'The `processes` resource is not supported on your OS yet.' if inspec.os.windows?
     end
 
     def to_s


chris-rock (approved):
Thank you @username-is-already-taken2 for adding more clarity!

arlimus (approved):
Thank you!!

PR #1856, opened 2017-05-29
Fix parameters to `find` commands

### LINKED ISSUE #1828: "-type f -maxdepth 1" should be "-maxdepth 1 -type f"
this is a bug: 
instead of: 

-type f -maxdepth 1

in /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.20.0/lib/resources/interface.rb
https://github.com/chef/inspec/blob/master/lib/resources/interface.rb

it should be: 

-maxdepth 1 -type f

because of: 
find: warning: you have specified the -maxdepth option after a non-option argument -type, but options are not positional (-maxdepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

Just try the find command on any unix-like/linux machine, it will give the error if the command order is not ok!

Thanks!

chris-rock:
fixes #1828

### DIFF
--- a/lib/resources/interface.rb
+++ b/lib/resources/interface.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 # author: Christoph Hartmann
 # author: Dominik Richter
+# author: Aaron Lippold
 
 require 'utils/convert'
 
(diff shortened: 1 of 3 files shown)


chris-rock:
This PR includes the commits from #1854 and fixes the same issue in other places, too

adamleff (approved):

adamleff (approved):
Thanks @chris-rock, and thanks @aaronlippold for the initial PR.

PR #1857, opened 2017-05-29
oracle_session and mssql_session improvement

### LINKED ISSUE #1821: Inspec oracledb_session resource not connecting to the database.
### Description

I am using oracledb_session resource in test cases, its not connecting to the database since its using wrong connection string.

[oracle@xx~]$ inspec version
      rake (>= 0)
1.24.0

### Replication Case

describe oracledb_session(user: 'system', pass: 'test123',sid: 'HELLODB', sqlplus_bin: '/u00/app/oracle/product/11.2.0.4/database/bin/sqlplus').query('SELECT NAME FROM v\$database;') do
  its('stdout') { should eq('HELLODB') }
  end

Output :+1: 

     ×  Command echo "SELECT NAME FROM v\$database;" | /u00/app/oracle/product/11.2.0.4/database/bin/sqlplus -s system/test123@localhost/HELLODB stdout should eq "HELLODB"

     expected: "HELLODB"
          got: "ERROR:\nORA-12541: TNS:no listener\n\n\nSP2-0306: Invalid option.\nUsage: CONN[ECT] [{logon|/|proxy}...nnect_identifier>]\n      <proxy> ::= <proxyuser>[<username>][/<password>][@<connect_identifier>]\n"

     (compared using ==)

     Diff:
     @@ -1,2 +1,9 @@
     -HELLODB
     +ERROR:
     +ORA-12541: TNS:no listener
     +
     +
     +SP2-0306: Invalid option.
     +Usage: CONN[ECT] [{logon|/|proxy} [AS {SYSDBA|SYSOPER|SYSASM}] [edition=value]]
     +where <logon> ::= <username>[/<password>][@<connect_identifier>]
     +      <proxy> ::= <proxyuser>[<username>][/<password>][@<connect_identifier>]

### Possible Solutions

Its actually trying to connect "localhost/HELLODB" connection string instead of "HELLODB"

### Stacktrace

chris-rock:
- fixes #1821
- fix: SID does not work with sqlplus, we need to use service names
- use `password` instead of `pass` as argument
- parse result table and make it accessible
- support oracle sqlci tool
- tested against oracle 12c and mysql server 2016

TODO:
- [x] update docs
- [x] unit tests


aaronlippold:
no love for postgres_session? :)

aaronlippold:
@chris-rock did you cover the case where the SID and SERVICE_NAME change? With one you need the port/host with the other you don't.

I don't see a service_name param in the current push.

```
    def db_conn
      if @sid
        "#{@user}/#{@password}@#{@host}:#{@port}/#{@sid}"
      else # using a 'service_name'
        "#{@user}/#{@password}@/#{@service_name}"
      end
    end
```
With a call like:
```
 escaped_query = Shellwords.escape(@query)
      cmd = inspec.command("echo -ne \"set heading off\nset echo off\n#{escaped_query}\" | #{@sqlplus_bin} -s #{db_conn_string}")
```
code:
```
require 'shellwords'

module Inspec::Resources
  class OracledbSession < Inspec.resource(1)
    name 'oracledb_session'
    desc 'Use the oracledb_session InSpec resource to test commands against an Oracle database'
    example "
      describe oracledb_session(user: 'my_user',
                                password: 'password',
                                host: '192.168.0.10',
                                service_name: 'TESTDB',
                                query: 'SELECT NAME FROM v$database;') do
        it { should be_successful }
        its('output') { should_not match /test/ }
      end

      Required values:
       - user
       - pass
       - sid or service_name
       - query

      Optional values:
       - host (default: localhost)
       - port (default: 1526)
    "

    attr_reader :user, :password, :host, :sid, :sqlplus_bin, :service_name, :port

    def initialize(opts = {})
      @user = opts[:user]
      @host = opts[:host] || 'localhost'
      @port = opts[:port] || '1526'
      @sid = opts[:sid]
      @query = opts[:query]
      @service_name = opts[:service_name]
      @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'

      # legacy option was "pass" but we prefer "password" for clarity
      if opts[:pass]
        @password = opts[:pass]
      else
        @password = opts[:password]
      end

      return skip_resource 'The `oracledb_session` resource does not support Windows yet' if inspec.os.windows?
      return skip_resource 'Username and password are required' if @user.nil? || @password.nil?
      return skip_resource 'You must provide at least an SID and/or a Service Name for the session' if @sid.nil? && @service_name.nil?
    end

    def output
      run_query
      @output
    end

    def errors
      run_query
      @errors
    end

    def successful?
      run_query
      @errors.nil? || @errors.empty?
    end

    def query(q)
      warn '[DEPRECATED] pass query in as an option when creating oracledb_session resource'
      @query = q
      run_query
      self
    end

    def run_query
      return if @query_already_run

      db_conn_string = db_conn
      # escaped_query = @query.gsub(/\\/, '\\\\').gsub(/"/, '\\"')
      escaped_query = Shellwords.escape(@query)
      cmd = inspec.command("echo -ne \"set heading off\nset echo off\n#{escaped_query}\" | #{@sqlplus_bin} -s #{db_conn_string}")
      @output = cmd.stdout
      @errors = cmd.stderr
      @query_already_run = true
    end

    def stdout
      warn '[DEPRECATED] use the `output` method to get the output of the oracle query'
      @output
    end

    def stderr
      warn '[DEPRECATED] use the `errors` method to get the errors from the oracle query'
      @errors
    end

    def db_conn
      if @sid
        "#{@user}/#{@password}@#{@host}:#{@port}/#{@sid}"
      else # using a 'service_name'
        "#{@user}/#{@password}@/#{@service_name}"
      end
    end

    def to_s
      'Oracle Session'
    end
  end
end
```

chris-rock:
@aaronlippold postgres is another iteration

chris-rock:
@aaronlippold we always use a service name, no sid

aaronlippold:
Hi

In the case of Oracle shouldn't we support both side service name?

On May 30, 2017 13:39, "Christoph Hartmann" <notifications@github.com>
wrote:

@aaronlippold <https://github.com/aaronlippold> we always use a service
name, no sid

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/chef/inspec/pull/1857#issuecomment-304953216>, or mute
the thread

Gemfile
```diff
@@ -8,6 +8,7 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.2')
 end
 
 gem 'ffi', '>= 1.9.14'
+gem 'htmlentities'
```
adamleff: This should go in the gemspec since the InSpec core (and therefore users who install the gem) is relying on it.

lib/resources/mssql_session.rb
```diff
@@ -24,31 +30,59 @@ class MssqlSession < Inspec.resource(1)
...
       end
-      puts cmd_string
       cmd = inspec.command(cmd_string)
       out = cmd.stdout + "\n" + cmd.stderr
       if out =~ /Sqlcmd: Error/
-        skip_resource("Can't connect to the MS SQL Server.")
+        return skip_resource("Can't connect to the MS SQL Server.")
       end
-      cmd
+      SQLQueryResult.new(cmd, parse_result(cmd))
```
adamleff: `parse_result` doesn't exist... did you mean `parse_csv_result` which is below in line 68?

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
-      @sid = opts[:sid]
+      @port = opts[:port] || '1521'
+      @service = opts[:service]
+
+      # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
+      @sqlcl_bin = 'sql'
       @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'
-      return skip_resource("Can't run Oracle checks without authentication") if @user.nil? or @pass.nil?
+
+      return skip_resource "Can't run Oracle checks without authentication" if @user.nil? or @password.nil?
```
adamleff: Use `||` instead of `or`

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
+      return skip_resource 'You must provide at least an SID and/or a Service Name for the session' if @service.nil?
     end
 
     def query(q)
       escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"')
-      cmd = inspec.command("echo \"#{escaped_query}\" | #{@sqlplus_bin} -s #{@user}/#{@pass}@#{@host}/#{@sid}")
+      # escape tables with $
+      escaped_query = escaped_query.gsub('$', "\\$")
```
adamleff: Can we use `Shellwords.shellescape` instead of these gsubs?

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
+
+      p = nil
+      # check if sqlcl is available and prefer that
+      if inspec.command(@sqlcl_bin).exist?
+        bin = @sqlcl_bin
+        opts = "set sqlformat csv\nSET FEEDBACK OFF"
+        p = 'parse_csv_result'
```
adamleff: Let's use a better variable name here, such as `parse_method`, and since we're already going to convert it to a symbols, let's set it as a symbol here and avoid the indirection.

```
parse_method = :parse_csv_result
```

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
+      elsif inspec.command(@sqlplus_bin).exist?
+        bin = @sqlplus_bin
+        opts = "SET MARKUP HTML ON\nSET FEEDBACK OFF"
+        p = 'parse_html_result'
```
adamleff: Same comment here as on line 50.

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
+        headers = rows[0].elements.to_a('/th').map { |entry| entry.text.strip }
+        rows.delete(0)
+
+        # iterate over each row, first row is header
+        hash = rows.map { |row|
+          res = {}
+          entries = row.elements.to_a('/td')
+
+          headers.each_with_index { |header, index|
+            # TODO: once nokogiri is back, we can remove html entities
```
adamleff: Are there plans to bring `nokogiri` back in? It has only caused us pain on multiple projects, especially since we support Windows. Can we just remove this TODO and continue to use `htmlentities` if we need it?

I don't suppose there's a non-HTML markup we can use with Oracle to avoid this?

lib/resources/oracledb_session.rb
```diff
@@ -13,30 +16,110 @@ class OracledbSession < Inspec.resource(1)
...
+            require 'htmlentities'
```
adamleff: Can we move this to the top of the file and treat it as first class? I'd rather throw a compile-time error on this when the resource is loaded/eval'd rather than at runtime.

lib/utils/database_helpers.rb
```diff
@@ -0,0 +1,66 @@
...
+    def initialize(cmd, results)
+      @cmd = cmd
+      @results = results
+    end
+
+    def empty?
+      @results.empty?
+    end
+
+    def stdout
```
adamleff: I'm nitpicking, but this and `stderr` really bug me because it feels like we're exposing implementation-level details to the user. They shouldn't care that we're making `command` calls underneath to get the data.

Can we call these `output` and `errors`? That way if we ever change the implementation, the method names will still be truthful.

adamleff (commented):
@chris-rock I know this is a WIP, but I have some feedback I'm hoping you'll take into consideration during your next iteration.

YarNhoj:
Since all we are doing is shelling out to sqlplus does this resource require that the stand alone sqlplus client be installed, or are we assuming that the user running the scans has all of the environment variables set up that sqlplus requires to run correctly?

As a non Oracle DB user I would have to set LD_LIBRARY_PATH, ORACLE_HOME, and a couple of other env variables before the sqlplus client would function correctly.

Here's a link to the sqlplus documentation from Oracle http://docs.oracle.com/database/121/SQPUG/ch_two.htm#SQPUG012

In my current use case I have an audit user that checks that seed data was populated correctly after install. This user isn't a DB user and won't have all of the env variables set.

chris-rock:
@YarNhoj Right now, this expects that the sqlplus client is installed and configured. Do you see any way around this?

lib/utils/database_helpers.rb
chris-rock: Good point. I agree on that. I am going to remove that since there is no need to expose this right now

lib/resources/oracledb_session.rb
chris-rock: Yes

lib/resources/oracledb_session.rb
chris-rock: I still try to find a way to circumvent `htmlentities`

lib/resources/oracledb_session.rb
chris-rock: +1

Gemfile
chris-rock: My bad. Thank you for the reminder!

lib/resources/oracledb_session.rb
chris-rock: Yes, I like that

chris-rock:
As discussed with @adamleff in https://github.com/chef/inspec/pull/1857#discussion_r119446961 we are going to remove stdout. Therefore this is a breaking change now.

adamleff:
It doesn't have to be... why don't we deprecate stdout with a warning now, and we can add another item to our 2.0 depredations issue to remove it then. What do you think?

chris-rock:
@adamleff great idea

lib/resources/mssql_session.rb
```diff
@@ -1,54 +1,103 @@
...
+# author: Nolan Davidson
 # author: Christoph Hartmann
 # author: Dominik Richter
 
+require 'hashie/mash'
+require 'utils/database_helpers'
+
 module Inspec::Resources
+  # STABILITY: Experimental
+  # This resouce needs further testing and refinement
```
adamleff: `s/resouce/resource`

lib/resources/mssql_session.rb
```diff
@@ -1,54 +1,103 @@
...
       end
-      puts cmd_string
       cmd = inspec.command(cmd_string)
       out = cmd.stdout + "\n" + cmd.stderr
-      if out =~ /Sqlcmd: Error/
-        skip_resource("Can't connect to the MS SQL Server.")
+      if cmd.exit_status != 0 || out =~ /Sqlcmd: Error/
+        # TODO: we need to throw an exception here
+        # change once https://github.com/chef/inspec/issues/1205 is in
+        # require "pry"; binding.pry
```
adamleff: Should remove this line.

lib/resources/oracledb_session.rb
```diff
@@ -1,41 +1,139 @@
...
+
+require 'hashie/mash'
+require 'utils/database_helpers'
+require 'htmlentities'
+require 'rexml/document'
+require 'csv'
 
 module Inspec::Resources
+  # STABILITY: Experimental
+  # This resouce needs further testing and refinement
```
adamleff: `s/resouce/resource`

lib/resources/oracledb_session.rb
```diff
@@ -1,41 +1,139 @@
...
+      @port = opts[:port] || '1521'
+      @service = opts[:service]
+
+      # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
+      @sqlcl_bin = 'sql'
       @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'
-      return skip_resource("Can't run Oracle checks without authentication") if @user.nil? or @pass.nil?
+
+      return skip_resource "Can't run Oracle checks without authentication" if @user.nil? || @password.nil?
+      return skip_resource 'You must provide at least an SID and/or a Service Name for the session' if @service.nil?
```
adamleff: We should eliminate the reference to "SID" from this skip message since we only support service name.

adamleff (changes requested):
Just a few last cleanup things that I'm happy to do myself.

adamleff (approved):
Sweet update, @chris-rock!

PR #1859, opened 2017-05-30
Make it a bit easier to actually use the junit mode

coderanger:
I can't find a way to actually test this since even loading the `runner_rspec.rb` file entirely breaks minitest (it shows 0 tests found no matter what). One option would be add this to BaseCLI too and test it via an integration test but since I don't use the inspec CLI, I defer to y'all on that front.

### DIFF
--- a/lib/inspec/runner_rspec.rb
+++ b/lib/inspec/runner_rspec.rb
@@ -120,6 +120,11 @@ module Inspec
       @formatter = RSpec.configuration.add_formatter(format)
       RSpec.configuration.color = @conf['color']
 
+      if @conf['junit']
+        junit_path = @conf['junit'].is_a?(String) ? @conf['junit'] : 'junit.xml'
+        RSpec.configuration.add_formatter('InspecRspecJUnit', junit_path)
+      end
+
       setup_reporting if @conf['report']
     end
 


chris-rock:
@coderanger Is the idea output the content to a file?

coderanger:
Yes

chris-rock:
I think we should make a global `-o` for all report types, similar to https://github.com/chef/kitchen-inspec/blob/master/lib/kitchen/verifier/inspec.rb#L187

adamleff:
I'm all for an `--output` flag, but I'm not sure that's exactly what @coderanger is after here. I think he wants to be able to have the JUnit formatter output to a file while still outputting the CLI formatter output to STDOUT. Am I getting that right, @coderanger?

coderanger:
Yep, put the junit data somewhere Jenkins can see it while leaving the human readable build output for debugging. Rspec itself supports this by allowing multiple formatters on the command line, with separate output paths for each.

mhedgpeth:
I'm confused because my colleague @jkerry added this specifically for jenkins support, and got it working (I thought). John, how did you integrate your CLI call with Jenkins?

coderanger:
You can (I think) make it work today with something like `inspec --format junit ... > junit.xml` but then you can't see any output in the build log, which makes debugging harder as a human. And with kitchen-inspec (which is my real target for this) it's just totally not a thing.

jkerry:
The approach recently has been to build formatters off of the json formatter. Junit was refactored into this and there has been some discussion of other modes also being moved into a position to build off of the json formatter. This has the known limitation of not providing any in-process output. Any fix to that limitation doesn't conflict with previous work

coderanger:
Yep, this uses the same junit formatter, just takes advantage of the fact that RSpec can have multiple formatters active at once.

mhedgpeth:
Very good, thanks for the clarification. I'm looking forward to this, then!

chris-rock:
We are going to move the complete report handling from audit cookbook to inspec core. Once this is in, you will be able to use CLI + File output at the same point in time. So I see two issues here:

- `-o` option to save the report to a file
- support reporting to multiple locations (eg. CLI, file etc)

Is that reflecting our understanding correctly?

adamleff:
I think that's accurate. I would want to expose a flag for each of the reporter format options though, not just `junit`, with the following behavior characteristics...

 * If you pass a `--format` flag with an option, that is the ONLY reporter that will run, and its content will be displayed to STDOUT (current behavior)
 * If you pass a `--FORMATTER` flag (where `FORMATTER` is the name of an available formatter`), the CLI formatter output will be displayed to STDOUT and the FORMATTER's output will be saved to a file (either the default file name per formatter, or the flag will accept the path to a file to which the formatter content will be written).
 * `--format` and `--FORMATTER` cannot be used simultaneously
 * Multiple `--FORMATTER` flags are permitted.

This will allow us to maintain the current backward compatibility while adding the functionality @coderanger is requesting (which I think is a great idea, for the record).

adamleff:
I also think this PR from @coderanger is part of the way there to what I'd love to see, and there's a direct use case for supporting JUnit output to a separate file, so my vote would be to allow this PR to be merged and we log a separate issue on reporter output overhaul.

coderanger:
@adamleff I would probably copy the behavior of the `rspec` CLI where you can pass `--format` multiple times and `--out` (or `--output` in this case for consistent naming?) only affects the previous formatter. For example in RSpec land we use `rspec --format doc --format yarjuf --out junit.xml` in CI. If no `--format` is specified, you get the default.

Also the thing I care about is `kitchen-inspec` so improving the `inspec` CLI is cool, but doesn't actually get me what I need anyway (though presumably if this is improved in core it can be in a way that Kitchen can hook in too).

adamleff:
I personally dislike that particular behavior of the rspec CLI with regards to the `--format` flag, but we can hash out the implementation details later. I'm more concerned right now with backward compatibility since many users are expecting the `--format` flag to work the way it does now... adding the `--junit` flag means that if we go to your `--format` suggestion would require deprecation of that flag pretty soon thereafter, etc., etc., etc....

coderanger:
Notably there is no `--junit` here :) I just added an internal option that `kitchen-inspec` is setting.

adamleff:
Yeah, understood... but that's something I'd want fixed before approving. This is a useful-enough feature to more than just Test Kitchen users so it's worth hashing this out now. But I want @chris-rock, @arlimus, and @hannah-radish to weigh in on the UX aspect of this before moving forward.

coderanger:
Any progress here?

chris-rock:
@adamleff and I discussed the reporting handling inside of inspec and @adamleff is working on a major overhaul that should make reporting to a file a lot easier.

jquick:
@coderanger - Thanks for opening this and getting discussion going about these issues. With inspec 2.0 we have completed the major overhaul for the reporter framework. This allows you to specify multiple reporters and output locations. As such I am going to close this PR. Please let us know if you run into any issues. For more information please visit: https://www.inspec.io/docs/reference/reporters/

PR #1860, opened 2017-05-30
Commenting the `contain_duplicates` deprecation until we have a good alternative

alexpop:


lib/matchers/matchers.rb
```diff
@@ -88,7 +88,7 @@
 
 RSpec::Matchers.define :contain_duplicates do
   match do |arr|
-    warn '[DEPRECATION] `contain_duplicates` is deprecated and will be removed in the next major version. See https://github.com/chef/inspec/issues/738 for more details'
+    # warn '[DEPRECATION] `contain_duplicates` is deprecated and will be removed in the next major version. See https://github.com/chef/inspec/issues/738 for more details'
```
chris-rock: I think we should just remove the comment completely

adamleff: Agreed.

adamleff: I'm going to adopt this change now and remove the line.

adamleff (approved):

alexpop:
Thank you Adam!

PR #1863, opened 2017-05-30
reject `nil` as a command input

arlimus:
It is not valid and the errors get very cryptic once it passes train and goes to the various backends. Let's just catch it early on. It is a user error that should be avoided.

### DIFF
--- a/lib/resources/command.rb
+++ b/lib/resources/command.rb
@@ -24,6 +24,9 @@ module Inspec::Resources
     attr_reader :command
 
     def initialize(cmd)
+      if cmd.nil?
+        raise 'InSpec `command` was called with `nil` as the argument. This is not supported. Please provide a valid command instead.'
+      end
       @command = cmd
     end
 
--- a/test/unit/resources/command_test.rb
+++ b/test/unit/resources/command_test.rb
@@ -1,0 +1,28 @@
+# encoding: utf-8
+# author: Christoph Hartmann
+# author: Dominik Richter
+
+require 'helper'
+require 'inspec/resource'
+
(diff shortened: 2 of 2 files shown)


lib/resources/command.rb
```diff
@@ -28,6 +28,9 @@ def initialize(cmd)
     end
 
     def result
+      if @command.nil?
+        raise 'InSpec `command` was called with `nil` as the argument. This is not supported. Please provide a valid command instead.'
```
adamleff: This feels like it belongs in `#initialize` since no method in this resource should work / will work with a `nil` command. Is there a technical reason you put it here instead of in the constructor?

adamleff (commented):
Love the change @arlimus but wonder if we should place it elsewhere.

lib/resources/command.rb
chris-rock: Lets make it skip the command resource for now until we have proper exception handling

arlimus: Actually, let's not skip. Skip points to unsupported use on a platform, at least in the current context. This however is a full user-error and should never be done this way. It's like adding nil to a number ;) We need the exception handling piece but it needs to fail afaics.
I can move it to initialize as well, will update this pr...

chris-rock: @arlimus this will stop inspec profile execution, even if you have multiple profiles in execution

adamleff (approved):
Even though we try to avoid calling `raise` because it will unwind the whole run, I feel like an issue like this is worthy of it as it points to an implementation issue vs. an error on the target. This will obviously get better once we implement better exception handling.

PR #1864, opened 2017-05-30
Add docs for http resource options

smith:
`auth` and `data` were incorrect; the timeouts and `ssl_verify` were missing.


adamleff (approved):
Thanks @smith. The red Travis tests appear unrelated to this (possibly a bad version of the `openssl` cookbook was out there during this run), but they appear to be going green now. Rerunning...

PR #1870, opened 2017-05-30
sudo-detection for target execution

arlimus:
When running `inspec exec` without the `target` option but against remote endpoints OR when executing it with the `localhost://` target AND having `--sudo` active it would abort the execution. `--target` is a helper to set the Train parameters for `backend`, `host`, `user`, `port`, and potentially `password`. The detection would fail on providing any of these separately without specifying `--target`. The same holds true for the `localhost` train backend or just `localhost://` target.

This type of detection has since moved to Train. The driving reason was to have this very useful check for localhost vs sudo run for any type of inspec (or for that matter: train) execution.

This PR depends on https://github.com/chef/train/pull/179 and the next release of train.

Strongly influenced by the original bugfix: https://github.com/chef/inspec/pull/1741

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.1'
 
-  spec.add_dependency 'train', '>=0.22.0', '<1.0'
+  spec.add_dependency 'train', '>=0.24.0', '<1.0'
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
That is a great improvement @arlimus

chris-rock:
Travis is having some issues today. I tested debian real quick locally.
screen shot 2017-05-30 at 6 18 49 pm

PR #1878, opened 2017-05-31
Add windows support to the `processes` resource

username-is-already-taken2:
Hi 

This PR is my attempt to really fix #1851 rather than patch it, by adding windows support.

I've also added an `exists` method as I found it useful. 

- [x] Add support for Windows
- [x] Add Exists Method
- [x] Create unit tests to include windows runner
- [x] Update the docs to include the an example for Exists
- [x] Update the docs to provide a windows example

  **Exists Method example...**
```
You are currently running on:

    OS platform:  ☺☻Windows Server 2012 R2 Standard☺☻
    OS family:  ☺☻windows☺☻
    OS release: ☺☻6.3.9600☺☻
inspec> describe processes('svchost') do
inspec>   it { should exist }
inspec> end

Profile: inspec-shell
Version: (not specified)

  Processes svchost
     [PASS]  should exist

```

Any comments or feedback is appreciated.

**Question regarding converting strings**
Don't know if anyone has an opinion but I thought it was better to convert the cpu and memory stats to numbers to make it easier to use is something like is it greater than a number rather then use a regex to parse a string?  (Hope that makes sense :) I saw that the pid,vsz,rss where already converted.

but thought it could be a breaking change which is why they are only converted on windows.  Happy to strip out.

Well let me know what you think, happy to start work on the tests (something else to learn :)

Gary

### DIFF
--- a/docs/resources/processes.md.erb
+++ b/docs/resources/processes.md.erb
@@ -60,16 +60,31 @@ The following examples show how to use this InSpec audit resource.
       its('list.length') { should eq 1 }
     end
 
-### Test if the init process is owned by the root user
+### Test if the process is owned by a specifc user
 
     describe processes('init') do
       its('users') { should eq ['root'] }
     end
 
+    describe processes('winlogon') do
+      its('users') { should cmp "NT AUTHORITY\\SYSTEM" }
+    end
+
+
 ### Test if a high-priority process is running
 
-    describe processes('some_process') do
+    describe processes('linux_process') do
       its('states') { should eq ['R<'] }
+    end
+
+    describe processes('windows_process') do
+      its('labels') { should cmp "High" }
+    end
+
+### Test if a process exists on the system
+
+    describe processes('some_process') do
+      it { should exist }
     end
 
 ### Test for a process using a specific Regexp
@@ -81,3 +96,28 @@ needed.
     describe processes(Regexp.new("/usr/local/bin/swap -d")) do
       its('list.length') { should eq 1 }
     end
+
+### Notes for auditing Windows systems
+
+Sometimes with system properties there isn't a direct comparison between different operating systems.
+Most of the `property_name`'s do align between the different OS's.  
+
+There are however some exception's, for example, within linux `states` offers multiple properties. 
+Windows doesn't have direct comparison that is a single property so instead `states` is mapped to the property of `Responding`, This is a boolean true/false flag to help determine if the process is hung.
+
+Below is a mapping table to help you understand what property the unix field maps to the windows `Get-Process` Property
+
+| *unix ps field* | *windows PowerShell Property* |
+|:---------------:|:-----------------------------:|
+|labels           |PriorityClass|
+|pids             |Id|
+|cpus             |CPU|
+|mem              |PM|
+|vsz              |VirtualMemorySize|
+|rss              |NPM|
+|tty              |SessionId|
+|states           |Responding|
+|start            |StartTime|
+|time             |TotalProcessorTime|
+|users            |UserName|
+|commands         |Path|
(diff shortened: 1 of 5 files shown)


adamleff:
@username-is-already-taken2 really loving your contributions - keep 'em coming!

Regarding casting the CPU/memory to numbers - you are correct that this could potentially be backward incompatible depending on how users have written their profiles.  Hopefully, it wouldn't be (calling `to_i` on an `Integer` still returns an `Integer)... but you never know.

Our recommendation to users would be use the `cmp` matcher which will automatically try to convert the value to a number so a comparison will work. So I'd suggest simply leaving those conversions out and letting the `cmp` do its magic.

alexpop:
Thanks for the contribution Gary!
Regarding the command mocking, take a look at this example:

https://github.com/chef/inspec/blob/v1.26.0/test/helper.rb#L186

https://github.com/chef/inspec/blob/v1.26.0/test/unit/mock/cmd/get-package-firefox

username-is-already-taken2:
Thanks very much @alexpop 

Appreciate your help.

I'm building a list of things I want to add to  `README.md` file, I'll include that.

I've added a some unit tests for windows to the `processes_test.rb` file using this MockLoader
```resource = MockLoader.new(:windows).load_resource('processes', 'winlogon.exe')``` and while the powershell command is now mocked my `resource` object is `nil` and so the following tests fail
```
  it 'retrieves the users and states as arrays on windows os' do
    resource = MockLoader.new(:windows).load_resource('processes', 'winlogon.exe')
    _(resource.users.sort).must_equal ['NT AUTHORITY\\SYSTEM']
  end

  it 'process should exist' do
    resource = MockLoader.new(:windows).load_resource('processes', 'winlogon.exe')
    _(resource.exists?).must_equal true
  end
```
Yet I know the resource does work :(
```
inspec> describe processes('winlogon.exe') do
inspec>   its ('users') { should cmp 'NT AUTHORITY\\SYSTEM' }
inspec>   it { should exist }
inspec> end
 
Profile: inspec-shell
Version: (not specified)

  Processes winlogon.exe
     [PASS]  should exist
     [PASS]  users should cmp == "NT AUTHORITY\\SYSTEM"

Test Summary: 2 successful, 0 failures, 0 skipped
```

If I'm honest, I'm a little out of my depth with unit testing, I don't know if its my test or my resource :(

Anyone willing to help me out ?

**Comment by username-is-already-taken2** [comment]
I think its something to do with the way the command get mocked, when the cmd is called and it reads in my sample data from `get-process_processes` its converting the control chars from `\r\n` to `\n` which throws the function as its expecting `\r\n` on windows

humm..... back to the drawing board.

```
[1] pry(#<MockLoader>)> cmd.call('get-process_processes')
=> #<struct Train::Transports::Mock::Connection::Command
 stdout=
  "PriorityClass,Id,CPU,PM,VirtualMemorySize,NPM,SessionId,Responding,StartTime,TotalProcessorTime,UserName,Path\nNormal,2456,0.296875,4808704,118202368,14576,1,True,5/31/2017 9:13:17 AM,00:00:00.2968750,WINVAGR-QQQNHPN\\Administrator,C:\\Windows\\system32\\mmc.exe\nHigh,396,0.15625,1323008,53710848,7776,1,True,5/31/2017 9:12:56 AM,00:00:00.1562500,NT AUTHORITY\\SYSTEM,C:\\Windows\\system32\\winlogon.exe\n",
```

adamleff:
@username-is-already-taken2 this is coming along quite nicely! How can we help you at this point?

username-is-already-taken2:
Well @adamleff I think I'm there.  Been a bit of a labour of love and such a learning experience :)

I think this is now ready for review.

alexpop:
windows_logo

Thank you Gary, great addition!

adamleff (approved):
Another great contribution, @username-is-already-taken2!

arlimus (approved):
Amazing contribution, thank you Gary @username-is-already-taken2 !! 🎆 

Also kudos @adamleff and @alexpop for your reviews and support 😄

PR #1879, opened 2017-05-31
Fix release_habitat rake task

adamleff:
A change made to how we generate the Gemfile during the Habitat build process means we cannot have the PLAN_CONTEXT be the "habitat" directory but instead need it to be the repo root itself.

Also changed to the preferred `hab pkg build` command instead of the original `hab studio build` command.


chris-rock (approved):
Thank you @adamleff

PR #1883, opened 2017-06-01
Use RubyGems version for habitat plan

smith:
In #1820 we made it so inspec would install the checked out source
version rather than the version from RubyGems.

This actually didn't work (though it wasn't apparent in a development
environment) because it used a relative path to bin/inspec that pointed
at /src/bin/inspec, which only exists if you're in a Habitat studio
started from the InSpec repo.

Revert back to getting the gem from RubyGems to avoid this problem and
have a working package.

tenor-36056381


adamleff (approved):
Thanks for the fix, @smith!

PR #1887, opened 2017-06-03
support FIPS 140-2 compliant digest calls

arlimus:
Calling the `digest` library directly unfortunately causes issues in FIPS 140-2 mode:

    sha512.c(81): OpenSSL internal error, assertion failed:
    Low level API call to digest SHA512 forbidden in FIPS mode!

Switching to `OpenSSL` as the caller resolve these issues

### DIFF
--- a/lib/fetchers/local.rb
+++ b/lib/fetchers/local.rb
@@ -2,6 +2,8 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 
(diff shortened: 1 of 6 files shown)


arlimus:
Raised internally at https://chefio.atlassian.net/browse/CMP-182

chris-rock (approved):
Thank you @arlimus

PR #1888, opened 2017-06-03
add the Nginx parser

arlimus:
The resource will be handled i a separate PR.

### DIFF
--- a/lib/utils/nginx_parser.rb
+++ b/lib/utils/nginx_parser.rb
@@ -1,0 +1,74 @@
+# encoding: utf-8
+# author: Dominik Richter
+# author: Christoph Hartmann
+
+require 'parslet'
+
+class NginxParser < Parslet::Parser
+  root :outermost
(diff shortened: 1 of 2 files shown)


aaronlippold:
👍 looking forward to the next PR :)

adamleff (approved):
I think I'd rather see the tests show real examples of pulling values out of the parsed configuration rather than the exact representation of the parselet output (i.e. `:section=>{:identifier=>\"group\"@0}`) as it feels like we're testing parselet more than our ability to use parselet to get values out of a configuration.

Otherwise approved though :)

arlimus:
@adamleff Since I'm testing both the lexer and the transformer and both can have errors I prefer these tests to quickly tell me where problems are. I fully agree that they aren't as nice, especially with the positional arguments, but it was the only way I found so far to get the interface between those components tested.

The last layer will come through the nginx config itself, which are the testable end restults i.e. the nginx to params mapping.

PR #1896, opened 2017-06-06
remove duplicate message in describe.one blocks

arlimus:
Generated duplicate messages due to the way that examples are aggregated in RSpec. Make sure we never show any duplicate test result messages, as they offer not value to any user.


arlimus:
Solves internal issue: https://chefio.atlassian.net/browse/CMP-177

adamleff (approved):
Elegant fix for an annoying problem. :) Thanks, @arlimus!

PR #1904, opened 2017-06-07
Add warning for potential empty password option

### LINKED ISSUE #1901: Unexpected `nil` authentication with `inspec exec -t` and WinRM
### Description

When running `inspec exec foo -t winrm://user@some.host.com --password` and not specifying a password, the user is not warned that a one is required. Instead, InSpec performs an auth with a `nil` password thus causing the account to become locked out.

### Possible Solutions

Prompt the user with a warning to prevent unexpected lock outs.

adamleff:
If a user supplies --password on the command line but fails to provide a value, Thor will set the value of the `password` option to the word "password". Users may expect to be prompted for a password if leaving this blank, but InSpec will happily try to log in to the target system with the password of "password".

Unfortunately, there is no way to alter Thor's behavior for this, and users may actually have a password of "password".

Proposing we add a warning if the password is "password" in an attempt to catch those users that accidentally provide an empty password. It may also serve as a warning to users that legitimately have a password of "password" they should perhaps use a different phrase. :)

Fixes #1901

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -155,6 +155,20 @@ class Inspec::InspecCLI < Inspec::BaseCLI # rubocop:disable Metrics/ClassLength
     configure_logger(opts)
     o = opts.dup
 
+    # Thor is silly. If you supply a flag like --password but don't provide a value,
+    # it uses a value matching the name of the options key. In this case: {'password' => 'password'}
+    # But some users may pass in --password expecting it to prompt for a password or
+    # otherwise. This means there's really no way for us to detect if a user is
+    # trying to login with a password of "password" or if they supplied --password
+    # without a value.
+    #
+    # This is a helpful warning that perhaps InSpec is going to try and log in with a
+    # password they didn't mean to.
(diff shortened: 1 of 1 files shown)


adamleff:
@arlimus / @chris-rock: this is a result of issue #1901. Thor isn't too friendly when it comes to detecting this particular situation, and given that users may actually have a password of "password" it adds some challenges. 🙂 

I'm totally fine with saying we simply don't merge this, but it's an attempt at catching bad user behavior early.

adamleff:
To wit, if you run `inspec exec profileA --password` without providing a password, this is what the options are:

```
[1] pry(#<Inspec::InspecCLI>)> o
=> {"color"=>true, "create_lockfile"=>true, "password"=>"password"}
```

Damn you, Thor.

arlimus:
Thank you for digging deep into this one! It's indeed a tricky thing to solve...
Is this potentially an alternative: https://github.com/chef/inspec/pull/1918 ?

adamleff:
Addressed in #1918

PR #1905, opened 2017-06-07
Adds support for iis_app InSpec testing

EasyAsABC123:
Signed-off-by: Justin Schuhmann <jmschu02@gmail.com>


lib/resources/iis_app.rb
```diff
@@ -0,0 +1,125 @@
...
+    def has_path?(path)
+      iis_app[:path] == path
+    end
+
+    def has_physical_path?(physical_path)
+      iis_app[:physical_path] == physical_path
+    end
+
+    def has_protocol?(protocol)
+      (iis_app[:protocols].include? protocol)
```
adamleff: This syntax is a bit strange. Can we change this to:

```
iis_app[:protocols].include?(protocol)
```

lib/resources/iis_app.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    def to_s
+      "iis_app '#{@site_name}#{@path}'"
+    end
+
+    def iis_app
+      return @cache unless @cache.nil?
+      @cache = @app_provider.iis_app(@path, @site_name) unless @app_provider.nil?
```
adamleff: We usually introduce a new "provider" class like this whenever we have a situation where we may need to interact differently with a target depending on the OS type/version, etc. In this case, since this will only be supported on windows, we can probably get rid of `AppProvider` and move the contents of `AppProvider#iis_app` into this `iis_app` method. That will eliminate an unnecessary layer of abstraction.

adamleff (changes requested):
@EasyAsABC123 well done! I'm so green when it comes to Windows knowledge, so it brings me great joy to see PRs like this come in.

I have two things I'd love to see changed before we move forward on this. Otherwise, it's a very clean implementation and I can't wait to get it into InSpec.

EasyAsABC123:
@adamleff let me know if this is what you meant

adamleff:
@EasyAsABC123 yes, exactly!

Now you're failing tests unfortunately. It looks like indentation got messed up which has led to you missing an `end`.  Lines 17 onward should be indented two spaces in under the `class` declaration, and then it becomes clearer that an `end` got dropped.

Ping me once you've got that fixed up and I'll give it another review.

EasyAsABC123:
@adamleff good catch, totally didn't notice.  It is now **_fixed_**
...locally it didn't break because i can't have module Inspec::Resources at the top current with how i was testing.

adamleff (approved):
@EasyAsABC123 well done! Thanks for your contribution!

lib/resources/iis_app.rb
```diff
@@ -0,0 +1,111 @@
...
+
+    def has_protocol?(protocol)
+      iis_app[:protocols].include?(protocol)
+    end
+
+    def to_s
+      "iis_app '#{@site_name}#{@path}'"
+    end
+
+    def iis_app
```
chris-rock: This should be a private function

lib/resources/iis_app.rb
```diff
@@ -0,0 +1,111 @@
+# encoding: utf-8
+# frozen_string_literal: true
+# check for web applications in IIS
+# Usage:
+# describe iis_app('/myapp', 'Website') do
```
chris-rock: I think we do not need the usage comments here, since it is part of the inline `examples`

test/cookbooks/os_prepare/recipes/default.rb
```diff
@@ -16,7 +16,7 @@
 include_recipe('os_prepare::service')
 include_recipe('os_prepare::package')
 include_recipe('os_prepare::registry_key')
-include_recipe('os_prepare::iis_site')
+include_recipe('os_prepare::iis')
```
chris-rock, line 19: This is a good move. We should add this resource to our integration tests then too

chris-rock (changes requested):
Great job @EasyAsABC123 I left some comments about minor improvements. Once they are addressed, LGTM

EasyAsABC123:
@chris-rock Thanks, good recommendations.  Added them, although i implemented New-WebApplication as a powershell script for testing purposes.

lib/resources/iis_app.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    def to_s
+      "iis_app '#{@site_name}#{@path}'"
+    end
+    
+    private
+    
+    def iis_app
+      return @cache unless @cache.nil?
+      command = "Import-Module WebAdministration; Get-WebApplication -Name '#{@path}' -Site '#{@site_name}' | Select-Object * | ConvertTo-Json"
```
chris-rock: Missed that before: What are the preconditions for the Powershell module? Can we expect this is deployed on IIS servers?

chris-rock:
Wow, that was quick :-)

lib/resources/iis_app.rb
EasyAsABC123: None that i'm aware of this is just a built in module for 2012r2 as far as i know, i am avoid xWebAdministration for this reason. 

```
For Windows 7 to 10 you will need to install the feature IIS Management Scripts and tools under Internet Information Services >> Web Management Tools
```

So it should be done simply from the IIS install.

EasyAsABC123:
yeah, can't sleep...always a bummer tomorrow is going to be a coffee fueled haze

lib/resources/iis_app.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    def to_s
+      "iis_app '#{@site_name}#{@path}'"
+    end
+
+    private
+
+    def iis_app
+      return @cache unless @cache.nil?
+      command = "Import-Module WebAdministration; Get-WebApplication -Name '#{@path}' -Site '#{@site_name}' | Select-Object * | ConvertTo-Json"
```
chris-rock, line 82: One last question: What is the requirement for the WebAdministration module. Will it be on all IIS servers by default or do users need to install it?

EasyAsABC123: @chris-rock 
None that i'm aware of this is just a built in module for 2012r2 as far as i know, i am avoid xWebAdministration for this reason. 

```
For Windows 7 to 10 you will need to install the feature IIS Management Scripts and tools under Internet Information Services >> Web Management Tools
```

So it should be done simply from the IIS install.

chris-rock: Perfect, thank you.

chris-rock (approved):

lib/resources/iis_app.rb
EasyAsABC123: @chris-rock never an issue, i'll take these recommendations to the other inspec resources i've written in the `iis-cookbook`.

chris-rock: Do you want to bring more of those resources into core inspec?

EasyAsABC123: @chris-rock of course! i'll be bringing them over one at a time to keep the reviews small.

chris-rock: Very cool. Thank you for doing that!

PR #1906, opened 2017-06-08
reading binary profile data on windows

arlimus:
On Windows, we ran into the problem that the execution of inherited profiles would (sometimes) not work. This was due to the use of `File.read` and `File.write` and handling inside the file provider, which works in most cases (especially *nix systems), but does not behave as expected on Windows. A better and more correct way of treating these files is via binary read and write mode, which changes the underlying encoding of both strings that are passed along.

### DIFF
--- a/lib/inspec/file_provider.rb
+++ b/lib/inspec/file_provider.rb
@@ -24,9 +24,20 @@ module Inspec
     def initialize(_path)
     end
 
+    # Read the contents of a file. This will typically refer to a text
+    # file reading a string.
+    #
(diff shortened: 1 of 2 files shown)


arlimus:
@chris-rock this is the one you have been hunting for 😁 
i got one more...

chris-rock (approved):
@arlimus Great finding! Thank you!

arlimus:
Likely caused: https://github.com/chef/inspec/issues/1790

PR #1907, opened 2017-06-08
inspec archive for tgz files on windows

arlimus:
On Windows, `inspec archive` would sometimes produce incorrect archive files. These would look fine, as tgz files, but would not execute correctly. This would lead to bewildering error messages like this one:

```
Unable to parse inspec.yml: line 1, control characters are not allowed
```

Fix it by treating the files as binary before writing them to get around any encoding issues, since the stream handler is a raw io object anyway.

Closely related to https://github.com/chef/inspec/pull/1906


chris-rock (approved):
Nice, thank you @arlimus

PR #1910, opened 2017-06-08
Updated unit tests for the groups resource

username-is-already-taken2:
Hi 

I noticed a number of DEPRECATED messages when rake testing last week, I thought I would attempt to resolve them, So this PR is to start to remove the Minitest DEPRECATED messages.  I've also tried to improve the coverage or this resource/
 example.....
`DEPRECATED: Use assert_nil if expecting nil from C:/git-repo/chef/inspec/test/unit/shell_detector_test.rb:75. This will fail in Minitest 6.`

Let me know if this if OK and I'll chip away at these whilst my other half watches soaps :)

Best Regards

Gary

### DIFF
--- a/test/unit/resources/group_test.rb
+++ b/test/unit/resources/group_test.rb
@@ -26,7 +26,7 @@ describe 'Inspec::Resources::Group' do
   it 'verify group on ubuntu' do
     resource = MockLoader.new(:ubuntu1404).load_resource('group', 'nogroup')
     _(resource.exists?).must_equal false
-    _(resource.gid).must_equal nil
+    _(resource.gid).must_be_nil
     _(resource.has_gid?(0)).must_equal false
   end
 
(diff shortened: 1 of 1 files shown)


adamleff (approved):
@username-is-already-taken2 you can't fool me, I know you're watching the soaps too. 🙂 

Thanks for the PR! looks great. I also put in PR #1916 to fix the `must_be_nil` assertions across the board, so I beat you to it!

arlimus (approved):
Thank you @username-is-already-taken2 for quality-control 👍

username-is-already-taken2:
Thx @arlimus, annoyingly I was shooting for 100% of coverage but I couldn't work out a way to stub the MockLoader so that I could trigger some of the exceptions.

Do you know of any examples I could review or technique that I could research where I might learn how to do it?

I thought of trying to override some of the values?

groups-gid

groups_groupinfo

groups_json

adamleff:
@username-is-already-taken2 `MockLoader` isn't too helpful when it comes to being able to stub out different returns or different exit codes.

The potential workaround is to use different parameters in each test scenario so that the command that would get executed would be different and then you could return different output. For example, I could test the user `userhasgroups` for a user that has groups, and then test user `userhasnogroups` for a user that no groups. That would generate two different PowerShell commands which would require two different mocks and therefore could have two different sets of output.

PR #1912, opened 2017-06-09
Fix code block CSS.

phiggins:
Hi,

While looking at the Inspec docs recently, I noticed what looked like some extra whitespace in the code samples. It turns out that it's actually some left padding that makes it look like the first line of a code block look like it has an extra space in front of it. The attached CSS fixes things for me.

| Before | After |
| --- | --- |
| Before Resource DSL | After Resource DSL | 
| Before service resource | After service resource
 |

### DIFF
--- a/www/source/stylesheets/_code.scss
+++ b/www/source/stylesheets/_code.scss
@@ -7,6 +7,10 @@ code {
 	border-color: $color_lt_blue;
 }
 
+.highlight code {
+  padding-left: 0;
+}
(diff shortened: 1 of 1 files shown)


hannah-radish (approved):
Thank you for taking care of this! Once Travis is green, good to go! 🎉

adamleff (approved):
Looks great, @phiggins - thanks for the fix!

PR #1913, opened 2017-06-09
Remove relative path from Gemfile.

phiggins:
Removes a relative path from the `Gemfile.lock`. Without this change, running `bundle install` modifies the `Gemfile.lock` with a path specific to my machine, making it a hassle to change between branches.


adamleff (approved):
Thanks, @phiggins!

PR #1916, opened 2017-06-09
Update minitest nil assertions

adamleff:
`must_equal nil` will fail in MiniTest 6. Changing those to `must_be_nil`
quiets down all the warnings we currently see and preps us for Minitest 6.


phiggins:
Nice, that was pretty noisy! 👍

arlimus (approved):
Awesome yes please!!

PR #1917, opened 2017-06-09
Reduce warnings

phiggins:
I noticed yesterday the test output was pretty noisy. I fixed some of the low hanging fruit and knocked just over 100 lines off the output:

```
Petes-MBP:inspec pete$ wc -l test_output_before
     510 test_output_before
Petes-MBP:inspec pete$ wc -l test_output_after
     401 test_output_after
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,9 +4,6 @@
 require 'bundler'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
-
-PROJECT_DIR = File.expand_path(".")
-
 require_relative 'tasks/changelog'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,9 @@
 require 'bundler'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
+
+PROJECT_DIR = File.expand_path(".")
+
 require_relative 'tasks/changelog'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -58,7 +58,7 @@ module Inspec
         end
 
         def resource_skipped
-          @resource_skipped
+          @resource_skipped if defined?(@resource_skipped)
         end
 
         def skip_resource(message)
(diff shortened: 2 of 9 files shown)


lib/inspec/plugins/resource.rb
```diff
@@ -58,7 +58,7 @@ def self.example(example = nil)
         end
 
         def resource_skipped
-          @resource_skipped
+          @resource_skipped if defined?(@resource_skipped)
```
arlimus, line 61: Ah fantastic catch!!

lib/resources/auditd_rules.rb
```diff
@@ -15,6 +15,7 @@ def initialize(content)
         assignment_regex: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
         multiple_values: true,
       }
+      @legacy = nil
```
arlimus: It looks like this class doesn't use `@legacy`. I see it in AuditDaemonRules below, but not in here.

arlimus (changes requested):
Fantastic update, thank you so much @phiggins !! I only found one tiny thing I'd either change or I might not fully understand in the `auditd_rules` (see comment below). Would love to get this merged :)

lib/resources/auditd_rules.rb
phiggins: Good catch, I did put that in the wrong place. I moved that to the other initialize method.

Rakefile
```diff
@@ -4,6 +4,9 @@
 require 'bundler'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
+
+PROJECT_DIR = File.expand_path(".")
```
adamleff: We need to add this to `www/Rakefile` as well since that is called independently of the project's Rakefile, but the www Rakefile calls the docs task which needs this constant as well.

adamleff (changes requested):
You rule, @phiggins :) I have one thing that needs to be addressed and then I'm good.

Rakefile
phiggins: Hmm, I wonder if there's a better way to do this then? I can probably get to this towards the end of the day and will think about it in the meantime.

adamleff: There's some history as to why the Rakefiles are separate, so I'm fine with this as a temporary hack until we have a reason to diverge.

phiggins: I removed the references to `PROJECT_DIR` in order to reduce the coupling between rake files. Let me know what you think.

arlimus (approved):
Thank you for all the improvements and follow-up @phiggins !! 😁

adamleff (approved):
Thanks again, @phiggins :)

PR #1918, opened 2017-06-11
enforce option values where needed

arlimus:
Due to limitations in Thor it is not possible to set an argument to be both optional and its value to be mandatory. E.g. the user supplying the --password argument is optional and not always required, but whenever it is used, it requires a value. Handle options that were defined with mandatory values in a way that fails with an `ArgumentError` if the value is missing, i.e.:

```
> inspec exec examples/profile --password
ArgumentError: Please provide a value for --password. For example: --password=hello.
```

It works without `--password` or with `--password=arg`. Also handled for `--sudo-password`.

Fixes: https://github.com/chef/inspec/issues/1901
An alternative approach: https://github.com/chef/inspec/pull/1904

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -19,7 +19,7 @@ module Inspec
         desc: 'Specify the login port for a remote scan.'
       option :user, type: :string,
         desc: 'The login user for a remote scan.'
-      option :password, type: :string,
+      option :password, type: :string, lazy_default: -1,
         desc: 'Login password for a remote scan, if required.'
       option :key_files, aliases: :i, type: :array,
         desc: 'Login key or certificate file for a remote scan.'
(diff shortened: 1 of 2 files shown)


lib/inspec/base_cli.rb
```diff
@@ -100,6 +100,23 @@ def diagnose
...
+      # Due to limitations in Thor it is not possible to set an argument to be
+      # both optional and its value to be mandatory. E.g. the user supplying
+      # the --password argument is optional and not always required, but
+      # whenever it is used, it requires a value. Handle options that were
+      # defined above and require a value here:
+      %w{password sudo_password}.each do |v|
+        if o[v.to_sym] == -1
+          raise ArgumentError, "Please provide a value for --#{v}. For example: --#{v}=hello."
+        end
+      end
```
arlimus: The trick here is that due to the type constraint these values will never reach `-1`, even if called with `--password=-1` as it's turned into `"-1"`. We can't use `nil` because it just says the default is not defined (or really it's empty, since it cannot check for "not defined"), so this looked like a nice alternative. We could also take the golang approach and pass a `{}`, but since it's a full hash in ruby unlike Golang's empty interface I just picked the next best thing...

adamleff (approved):
This is a pretty crafty way of handling this. And TIL about `lazy_default` in Thor. Thanks, @arlimus!

arlimus:
😁 kudos Adam!!

PR #1920, opened 2017-06-12
bugfix: reading tgz files with binread

arlimus:
This is currently failing because both `alias` and `alias_method` create method imprints in the FileProvider. This leads to a failure where the TarProvider assumes it doesn't implement `binread`, which in fact it does, since it just calls `read`.

Easily visible with:
```
> inspec json meta-profile-0.2.0.tar.gz
RuntimeError: #<Inspec::TarProvider:0x00000003b83280> does not implement `read(...)`. This is required.
```

### DIFF
--- a/lib/inspec/file_provider.rb
+++ b/lib/inspec/file_provider.rb
@@ -24,6 +24,13 @@ module Inspec
     def initialize(_path)
     end
 
+    # List all files that are offered.
+    #
+    # @return [Array[String]] list of file paths that are included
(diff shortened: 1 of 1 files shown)


arlimus:
Updates to the test that is catching this is in: https://github.com/chef/inspec/pull/1919
Left it there to prevent merge-conflicts.

chris-rock (approved):
Thank you @arlimus

PR #1922, opened 2017-06-12
Fix detection of Automate pre-0.8.x in Compliance::API

### LINKED ISSUE #1921: Error when listing compliance profiles against Automate pre 0.8
### Description

Ruby stacktrace when running `inspec compliance profiles` with a pre-0.8.x Automate Server

InSpec v1.27.0

### Replication Case

 * Set up Automate 0.7.151
 * Install InSpec 1.27.0
 * Run `inspec compliance login_automate`
 * Run `inspec compliance profiles` (which will fail)

### Possible Solutions

Unknown

### Stacktrace

```
C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/api.rb:47:in `[]=': no implicit conversion of String into Integer (TypeError)
        from C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/api.rb:47:in `block in profiles'
        from C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/api.rb:46:in `each'
        from C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/api.rb:46:in `map'
        from C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/api.rb:46:in `profiles'
        from C:/Users/admin-hr/AppData/Local/chefdk/gem/ruby/2.3.0/gems/inspec-1.27.0/lib/bundles/inspec-compliance/cli.rb:101:in `profiles'
        from C:/opscode/chefdk/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from C:/opscode/chefdk/em
```

adamleff:
The is_automate_server_pre_080? and is_automate_server_080_and_later?
methods needed some fixing. The Compliance configuration could have
a "version" key that was not nil but was an empty hash, indicating
that it came from a pre-0.8.x Automate server. What we really need
to look for is config['version']['version'] being nil?.

Fixes #1921

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -204,11 +204,22 @@ module Compliance
     end
 
     def self.is_automate_server_pre_080?(config)
-      config['server_type'] == 'automate' && config['version'].nil?
+      # Automate versions before 0.8.x may have a "version" key in the config.
+      # Unless it's a hash that also contains a "version" key, it came from
+      # an Automate server that is pre-0.8.x.
+      return false unless config['server_type'] == 'automate'
+      return true unless config.key?('version')
+      return true unless config['version'].is_a?(Hash)
+      config['version']['version'].nil?
     end
 
     def self.is_automate_server_080_and_later?(config)
-      config['server_type'] == 'automate' && !config['version'].nil?
+      # Automate versions 0.8.x and later will have a "version" key in the config
+      # that looks like: "version":{"api":"compliance","version":"0.8.24"}
+      return false unless config['server_type'] == 'automate'
+      return false unless config.key?('version')
+      return false unless config['version'].is_a?(Hash)
+      !config['version']['version'].nil?
     end
 
     def self.is_automate_server?(config)
--- a/lib/bundles/inspec-compliance/configuration.rb
+++ b/lib/bundles/inspec-compliance/configuration.rb
@@ -28,6 +28,10 @@ module Compliance
       @config[key] = value
     end
 
+    def key?(key)
+      @config.key?(key)
+    end
+
     def clean
       @config = {}
     end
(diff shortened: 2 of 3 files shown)


lib/bundles/inspec-compliance/api.rb
```diff
@@ -204,11 +204,22 @@ def self.is_compliance_server?(config)
...
+      # Unless it's a hash that also contains a "version" key, it came from
+      # an Automate server that is pre-0.8.x.
+      return false unless config['server_type'] == 'automate'
+      return true unless config.key?('version')
+      return true unless config['version'].is_a?(Hash)
+      config['version']['version'].nil?
     end
 
     def self.is_automate_server_080_and_later?(config)
-      config['server_type'] == 'automate' && !config['version'].nil?
```
chris-rock, line 211: should we go and use `!self.is_automate_server_pre_080?`

adamleff: I don't think that's necessary. I'd rather these methods be able to stand on their own rather than depending on themselves. It also makes it clearer to the next person that stumbles upon this part of the code.

adamleff: Technically, a compliance server would test "true" for `!is_automate_server_pre_080?` too, right? I think that's another reason to leave this as proposed but I'd like to understand your position better.

arlimus: makes sense, it returns differently for the two middle statements. lgtm :)

chris-rock (approved):
Thank you @adamleff

PR #1924, opened 2017-06-13
Adding toml resource

nsdavidson:
This adds a `toml` resource that inherits from the json resource and
behaves the same way as the JSON and YAML resources.


lib/resources/toml.rb
```diff
@@ -0,0 +1,33 @@
+# encoding: utf-8
+# author: Nolan Davidson
+
+require 'toml'
+
+# Parses a TOML document
+# Usage
+# describe toml('default.toml') do
```
chris-rock: We do not need the comments, since the example is inline as part of the `example`

chris-rock (commented):
Great addition @nsdavidson Lets just remove the additional comments and we are good to go :-)

nsdavidson:
@chris-rock Thanks!  Comments removed.

adamleff (approved):
Nicely done, @nsdavidson!

PR #1928, opened 2017-06-15
Add support for CoreOS to the service resource

rarenerd:
This PR adds CoreOS to the list of platforms in the service resource.

As a separate commit I've added a text change in `test/unit/resources/service_test.rb`. All tests here mentioned 'package' which seemed wrong to me :)


chris-rock (approved):
Works like a charm with my local coreos. Awesome @rarenerd

adamleff (approved):
Thanks as always, @rarenerd 🙂

PR #1931, opened 2017-06-15
Extract Compliance::API version parsing to separate method

adamleff:
For cleanliness and ease of testing, I've moved the logic that parses the server version from the compliance config to a separate method.


chris-rock (approved):
Awesome. Thank you @adamleff

PR #1933, opened 2017-06-16
Allow mysql resource to accept socket path

rshade:
We can not use this by default against mysql installations via the mysql cookbook.


adamleff:
@rshade thanks for your PR!

Unfortunately, we cannot accept this change as written. The `-S/--socket` parameter overrides the `-h/--host` parameter, and we do have users that connect to MySQL instances that are not on the local machine that's being scanned (i.e. testing DB connectivity from a web server, for example).

I verified this by creating a Vagrant machine, installing MySQL, configuring it to listen on 13306 (and forwarded that port), and then tried to connect to it using both `-h` and `-S`:

```
mysql -uvagrant -p -S /tmp/nonexistent.sock -h localhost --port 13306
Enter password:
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/tmp/nonexistent.sock' (2)
```

One potential solution would be to set the default `socket` parameter to nil, and then have this kind of logic:

```
command = "mysql .... -u username .... "
if !socket.nil?
  command += "-S #{socket}"
else
  command += "-h #{host} ...."
```

That would still allow you to define a socket in your tests without breaking existing users.

adamleff (changes requested):
Unable to accept as-written, would break existing known use.

adamleff (approved):
This looks pretty great to me, @rshade, thanks!

If you have time to level-up this change, I would break out the code the builds the CLI command string into a separate method (called `mysql_command` or something similar), and then that will allow us to test that logic easily and independently. If you don't have time for that, lemme know and I can follow-up with another PR to add that as well.

rshade:
I will try and level up a little beginning of next week. Would it be cool if I open issues on a few of the items and we can work from there?

adamleff:
That works for me!

lib/resources/mysql_session.rb
```diff
@@ -16,10 +16,12 @@ class MysqlSession < Inspec.resource(1)
       end
     "
 
-    def initialize(user = nil, pass = nil, host = 'localhost')
+    def initialize(user = nil, pass = nil, host = 'localhost', port = nil, socket = nil)
```
chris-rock, line 19: In the future we probably want to use opts here to avoid passing nil values

chris-rock (approved):
Thank you @rshade for this improvement

lib/resources/mysql_session.rb
```diff
@@ -31,7 +32,16 @@ def query(q, db = '')
...
       # run the query
-      cmd = inspec.command("mysql -u#{@user} -p#{@pass} -h #{@host} #{db} -s -S #{@socket} -e \"#{escaped_query}\"")
+      command = "mysql -u#{@user} -p#{@pass}"
+      if !socket.nil?
+        command += " -S #{@socket}"
+      else
+        command += " -h #{@host}"
+      end
+      command += " --port #{@port}" unless @port.nil?
+      command += " #{db} -s -S #{@socket} -e \"#{escaped_query}\""
```
n-rodriguez, line 42: The `-S #{@socket}` argument is redundant with `-h localhost` and leads to MySQL error

stefanandres: Yep, the current version breaks for all queries without socket connections because of this

stefanandres: Ah, this is fixed in newer versions.

PR #1941, opened 2017-06-19
attributes-related note added to docs

doertedev:
Hello friends.

I just received two separate stacktraces and spent half an hour trying to figure out why my attribute()-calls didn't work. The Slack channel then notified me that these attribute()-calls must live on the top-level of the file. Thought I'd prepare a PR to save others this fight :)

Greetings from southern germany.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -286,12 +286,20 @@ For example, a control:
     val_user = attribute('user', default: 'alice', description: 'An identification for the user')
     val_password = attribute('password', description: 'A value for the password')
 
-    describe val_user do
-      it { should eq 'bob' }
-    end
+    control 'system-users' do
+      impact 0.8
+      desc '
+        This test assures that the user "Bob" has a user installed on the system, along with a
+        specified password.
+      '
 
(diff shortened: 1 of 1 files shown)


adamleff (approved):
Thanks @doertedev, and congrats on submitting your first PR to InSpec!

This is good with me. What might make it even clearer is to make one of the example `describe` blocks below the attributes a full-fledged `control` block. That way a user sees this is clearly something that happens outside of the `control` block, vs. possibly within a `control` block as one might assume by looking at this example.

Would you be willing to make that change?

doertedev:
Done. There's a problem though (I think). Looking at the profiles code example beneath, technically we do not test anything on the system here but just test the passed vars. Maybe this could cause confusion as well? I mean, why not check if the speicified system user has the specified vars and remove the hardcoded strings from the control?

adamleff:
Yeah, that's true. I think it's fine as a demonstration of what attributes are for but all docs can always be better. I think it's fine as-is.

chris-rock (approved):
Thank you @doertedev

PR #1948, opened 2017-06-21
Don't send HTTP headers that have nil values

adamleff:
Net::HTTP does not gracefully handle HTTP options/headers that have nil values. This updates Fetchers::Url to verify that all headers we attempt to configure have non-nil, non-empty values.

This originally surfaced via the audit cookbook with the chef-automate fetcher in use without the data_collector token being set.

### DIFF
--- a/lib/fetchers/url.rb
+++ b/lib/fetchers/url.rb
@@ -143,21 +143,6 @@ module Fetchers
     def download_archive_to_temp
       return @temp_archive_path if ! @temp_archive_path.nil?
       Inspec::Log.debug("Fetching URL: #{@target}")
-      http_opts = {}
-      http_opts['ssl_verify_mode'.to_sym] = OpenSSL::SSL::VERIFY_NONE if @insecure
-      if @config
-        if @config['server_type'] == 'automate'
-          http_opts['chef-delivery-enterprise'] = @config['automate']['ent']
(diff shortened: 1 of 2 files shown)


lib/fetchers/url.rb
```diff
@@ -178,5 +163,42 @@ def download_archive(path)
       @temp_archive_path = nil
       final_path
     end
+
+    def http_opts
+      opts = {}
+      opts[:ssl_verify_mode] = OpenSSL::SSL::VERIFY_NONE if @insecure
+
+      if @config['server_type'] == 'automate'
```
chris-rock, line 171: I think we should deprecate all delivery specific handling in this class and move all the required logic into audit cookbook `files/default/vendor/chef-automate/fetcher.rb`. I understand that we do not want to make this in one move. But I'd like to see a follow-up on that PR to cover that. This allows us to have the chef-* specific parts in one place.

chris-rock (approved):
Thank you @adamleff

lib/fetchers/url.rb
adamleff: Yeah, that makes sense to me. I didn't want to tackle that in this PR but rather just add the error checking. Some better tests need to be written when we do that, and probably before, also.

PR #1950, opened 2017-06-22
Fix directory resource output and exists check

### LINKED ISSUE #1944: Directory resource says it's file
### Description

Documentation about `directory` resource says that
> This is equivalent to using the file resource and the be_directory matcher

However, test output says about `File` and nothing about `be_directory` matcher.

inspec version: 1.28.1
host os: Windows 7 SP 1
tested os: Ubuntu Server 16.04.2

### Replication Case
```
    describe directory('/home/ubuntu/.gems') do
      it { should exist }
    end
```
Output:
```
     [PASS]  File /home/ubuntu/.gems should exist
```

I wonder whether `inspec` really checks that specified file is actually a directory.

adamleff:
The `to_s` method on the `directory` resource is not defined in the correct class, leading `directory` resources to be printed as the parent resource (`file`) instead.

Also, the `exist?` method just reuses the `file` resource's `exist?` method which can lead to perceived false results when a user calls `it {should exist}` on a file with the `directory` method thinking it's actually checking to see if the target is a directory as well. This change ensures that the `exist?` method also checks to make sure the target is a directory.

Fixes #1944

### DIFF
--- a/lib/resources/directory.rb
+++ b/lib/resources/directory.rb
@@ -13,9 +13,9 @@ module Inspec::Resources
         it { should be_directory }
       end
     "
-  end
 
-  def to_s
-    "Directory #{@path}"
+    def to_s
+      "Directory #{source_path}"
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @adamleff

PR #1960, opened 2017-06-23
Add `rpm_dbpath` support to the package resource

### LINKED ISSUE #1959: Add support for `rpm_dbpath` to the `package` resource
### Description

As a consultant, I have encountered scenarios where a custom RPMDB is used to store RPM info. Currently, the InSpec `package` resource does not support modifying the RPM query command to support the `--dbpath` command line argument.

### Possible Solutions

Modify the `package` resource to support `rpm_dbpath`.

Example:

```
describe package('my_package', rpm_dbpath: '/var/lib/my_rpmdb') do
  it { should be_installed }
end
```

jerryaldrichiii:
This adds support for querying a specific RPMDB

Example:

```
describe package('my_package', rpm_dbpath: '/var/lib/my_rpmdb') do
  it { should be_installed }
end
```

Fixes #1959

### DIFF
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -19,7 +19,7 @@ module Inspec::Resources
       end
     "
 
-    def initialize(package_name = nil) # rubocop:disable Metrics/AbcSize
+    def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize
       @package_name = package_name
       @name = @package_name
       @cache = nil
@@ -30,7 +30,7 @@ module Inspec::Resources
       if os.debian?
         @pkgman = Deb.new(inspec)
       elsif os.redhat? || %w{suse amazon fedora}.include?(os[:family])
-        @pkgman = Rpm.new(inspec)
+        @pkgman = Rpm.new(inspec, opts)
       elsif ['arch'].include?(os[:family])
         @pkgman = Pacman.new(inspec)
       elsif ['darwin'].include?(os[:family])
@@ -46,6 +46,8 @@ module Inspec::Resources
       else
         return skip_resource 'The `package` resource is not supported on your OS yet.'
       end
+
+      evaluate_missing_requirements
     end
 
     # returns true if the package is installed
@@ -71,12 +73,26 @@ module Inspec::Resources
     def to_s
       "System Package #{@package_name}"
     end
+
+    private
+
+    def evaluate_missing_requirements
+      missing_requirements_string = @pkgman.missing_requirements.uniq.join(', ')
+      return if missing_requirements_string.empty?
+      skip_resource "The following requirements are not met for this resource: #{missing_requirements_string}"
+    end
   end
 
   class PkgManagement
     attr_reader :inspec
     def initialize(inspec)
       @inspec = inspec
+    end
+
+    def missing_requirements
+      # Each provider can provide an Array of missing requirements that will be
+      # combined into a `skip_resource` message
+      []
     end
   end
 
@@ -104,8 +120,25 @@ module Inspec::Resources
 
   # RHEL family
   class Rpm < PkgManagement
+    def initialize(inspec, opts)
+      super(inspec)
+
+      @dbpath = opts.fetch(:rpm_dbpath, nil)
+    end
+
+    def missing_requirements
+      missing_requirements = []
+
+      unless @dbpath.nil? || inspec.directory(@dbpath).directory?
+        missing_requirements << "RPMDB #{@dbpath} does not exist"
+      end
+
+      missing_requirements
+    end
+
     def info(package_name)
-      cmd = inspec.command("rpm -qia #{package_name}")
+      rpm_cmd = rpm_command(package_name)
+      cmd = inspec.command(rpm_cmd)
       # CentOS does not return an error code if the package is not installed,
       # therefore we need to check for emptyness
       return nil if cmd.exit_status.to_i != 0 || cmd.stdout.chomp.empty?
@@ -132,6 +165,17 @@ module Inspec::Resources
         version: "#{v}-#{r}",
         type: 'rpm',
       }
+    end
+
+    private
+
+    def rpm_command(package_name)
+      cmd = ''
+      cmd += 'rpm -qia'
+      cmd += " --dbpath #{@dbpath}" if @dbpath
+      cmd += ' ' + package_name
+
+      cmd
     end
   end
 
--- a/docs/resources/package.md.erb
+++ b/docs/resources/package.md.erb
@@ -96,6 +96,12 @@ The following examples show how to use this InSpec audit resource.
       it { should_not be_running }
     end
 
+### Verify if some_package is installed according to my_rpmdb
+
+   describe package('some_package', rpm_dbpath: '/var/lib/my_rpmdb') do
+     it { should be_installed }
+   end
+
 ### Verify if Memcached is installed, enabled, and running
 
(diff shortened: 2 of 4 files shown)


lib/resources/package.rb
```diff
@@ -104,8 +104,24 @@ def info(package_name)
 
   # RHEL family
   class Rpm < PkgManagement
+    def initialize(inspec, opts)
+      super(inspec)
+
+      @dbpath = opts.fetch(:rpm_dbpath, nil)
+    end
+
+    def rpm_command(package_name)
```
chris-rock: that should be a private function of that class

chris-rock (commented):
This looks really great @jerryaldrichiii I just added a suggestion for a minor improvement

lib/resources/package.rb
jerryaldrichiii: Agreed. I'll make the change now!

chris-rock (dismissed):
Awesome, very quick turn-around @jerryaldrichiii

adamleff (changes requested):
@jerryaldrichiii this is a great contribution, thank you!

My main concern is the one that I raised with you directly regarding returning `false` if the RPM DB cannot be found. We've already seen similar situations confuse our users (i.e. returning nil file contents simply because the InSpec user didn't have permissions to read, but not giving them that feedback). In this case, we'd return `false` but if they log in manually and poke around, they'd see it's installed. Ideally, we'd be able to raise an exception and provide the user with more helpful error messaging, but until we implement #1205, our best option is to `skip_resource`.

Since you can't `skip_resource` in an object that's not an `Inspec::Resource`, you need to handle this in the top-level of the resource. Take a look at the `missing_requirements` pattern in the `host` resource:

https://github.com/chef/inspec/blob/master/lib/resources/host.rb#L68-L71
https://github.com/chef/inspec/blob/master/lib/resources/host.rb#L136-L140
https://github.com/chef/inspec/blob/master/lib/resources/host.rb#L182-L190

I'd love to see something like that implemented for this before we merge. Then, after #1205 is implemented, we can change that `skip_resource` into an exception that either skips or fails the test as necessary. I'm happy to pair with you if needed.

jerryaldrichiii:
Great feedback @adamleff and thank you for the reference material!

I added something similar by doing the following:
  - Create a list for `missing_requirements` so `Inspec::Resources::Package` can support multiple reasons for skipping a test
  - Add a private method to `Inspec::Resources::Package` to evaluate missing requirements (allowing for multiple reasons to skip the resource)
  - Add a specific evaluator for `@pkgman` missing requirements (this sets a pattern for others going forward)
  - Add a public `missing_requirements` method to `Inspec::Resources::Rpm` which allows for more than one argument to be validated going forward
  - Add a private `validate_arguments` method `Inspec::Resoruces::Rpm` to append missing requirements to the `missing_requirements` list for command arguments
  - Add a private `validate_rpm_dbpath` method `Inspec::Resoruces::Rpm` to append missing requirements to the `missing_requirements` list specific to the `--dbpath` command line argument
  - Modify the Unit tests to match Minitest patterns (no `context` available like RSpec). NOTE: Technically this `describe` block should move to another file since it is separate  from the `Inspec::Resources::Package` module

jerryaldrichiii:
@adamleff, I took your offline feedback and made some modifications.

Now the `PkgManagement` class has a `missing_requirements` method that the inheriting classes can use. This removes that `defined?` logic.

lib/resources/package.rb
```diff
@@ -44,8 +44,10 @@ def initialize(package_name = nil) # rubocop:disable Metrics/AbcSize
       elsif ['hpux'].include?(os[:family])
         @pkgman = HpuxPkg.new(inspec)
       else
-        return skip_resource 'The `package` resource is not supported on your OS yet.'
+        skip_resource 'The `package` resource is not supported on your OS yet.'
```
adamleff: You probably want to leave the `return` here - there's no reason to continue to evaluating missing requirements if the resource isn't supported on the target OS.

lib/resources/package.rb
```diff
@@ -44,8 +44,10 @@ def initialize(package_name = nil) # rubocop:disable Metrics/AbcSize
...
       end
+
+      evaluate_missing_requirements(@pkgman.missing_requirements) if @pkgman
```
adamleff: If you leave the `return` above, you'll always have a populated `@pkgman` variable, so I'd just remove `if @pkgman`

adamleff: Also, since the pkg manager is stored in an instance variable, you probably don't need to pass it. Just let `evaluate_missing_requirements` take zero arguments and it can refer to `@pkgman` by itself.

lib/resources/package.rb
```diff
@@ -71,13 +73,28 @@ def version
     def to_s
       "System Package #{@package_name}"
     end
+
+    private
+
+    def evaluate_missing_requirements(missing_requirements)
```
adamleff: You want to bail out of a method as quickly as possible if you can. In this case, you build up a `skip_msg` that you may never use because you don't check for emptiness of the missing requirements until the very last step.

I'd rewrite this method like this:

```ruby
def evaluate_missing_requirements
  missing_requirements = @pkgman.missing_requirements
  return if missing_requirements.empty?

  skip_resource "The following requirements are not met for this resource: #{missing_requirements.join(', ')}"
end
```

I don't know that you need to compact and uniq unless you're doing something in your code that you think would require that... in which case I'd encourage you to solve that problem since you control the whole kit and kaboodle. 🙂

lib/resources/package.rb
```diff
@@ -133,6 +164,30 @@ def info(package_name)
         type: 'rpm',
       }
     end
+
+    private
+
+    def validate_arguments
```
adamleff: It feels like this method and the one below it are unnecessary layers of abstraction. Any reason not just to fold these into `missing_requirements` in the `Rpm` class? If we collect additional requirements later, we can break those down until separate methods but it feels like a bunch of hopping around between methods to get to a simple "if this parameter isn't nil, check to make sure it's actually a file on disk" check.

adamleff (changes requested):
@jerryaldrichiii this is much closer to what I'd like to see! I think there's a couple of small clean-up things we can do and then this will be good to go.

lib/resources/package.rb
jerryaldrichiii: Agreed, will fix!

lib/resources/package.rb
jerryaldrichiii: Agreed, will fix!

lib/resources/package.rb
jerryaldrichiii: Agreed, will fix!

lib/resources/package.rb
jerryaldrichiii: Agreed, will fix!

jerryaldrichiii:
@adamleff I pushed up my recent changes based on your feedback.

Keep it coming!

adamleff (approved):

adamleff:
Looks great, @jerryaldrichiii - thanks!

lib/resources/package.rb
```diff
@@ -71,13 +73,27 @@ def version
     def to_s
       "System Package #{@package_name}"
     end
+
+    private
+
+    def evaluate_missing_requirements
```
chris-rock, line 79: While I agree that we need to find a pattern for checking missing options, the current approach feels bloated and very complicated. I'd like to replace this with a version that is more lightweight and more flexible to be used in all resources. I do not have an idea right now, therefore this is not a PR blocker.

chris-rock (approved):
Great addition @jerryaldrichiii

PR #1961, opened 2017-06-23
make postgres resource working in mock runner (for inspec check)

chris-rock:
This helps to use the postgres resources in mock environments with empty data better. Also, this is required to get inspec check for https://github.com/dev-sec/postgres-baseline green again

### DIFF
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -11,6 +11,9 @@ module Inspec::Resources
 
     attr_reader :service, :data_dir, :conf_dir, :conf_path, :version, :cluster
     def initialize
+      @conf_dir = ''
+      @data_dir = ''
+
       os = inspec.os
       if os.debian?
         #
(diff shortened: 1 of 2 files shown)


lib/resources/postgres.rb
```diff
@@ -11,6 +11,9 @@ class Postgres < Inspec.resource(1)
 
     attr_reader :service, :data_dir, :conf_dir, :conf_path, :version, :cluster
     def initialize
+      @conf_dir = ''
```
adamleff: Why empty strings and not nil?

adamleff (dismissed):
The reason for the change makes sense, but why did you use empty strings as defaults instead of `nil`? It makes it inconsistent with the way we treat non-set default variables in other resources.

Also, you've got some rubocop complaints to deal with.

lib/resources/postgres.rb
chris-rock: Agree on that one :-)

arlimus (approved):
Great update, kudos Chris.
Also thank you Adam for your initial review.

PR #1968, opened 2017-06-26
Update iptables docs

swalberg:
* Correct nomenclature and be a bit more specific for existing exampls
* Provide an example of allowing a specified port in.


adamleff (approved):
Thanks for the doc enhancement, @swalberg!

chris-rock (approved):
Thank you @swalberg

PR #1986, opened 2017-07-05
Fix host resolution on Darwin, use dig wherever possible

adamleff:
The `host` and `dig` commands do not return non-zero if a query returns NXDOMAIN or NOERROR, but the DarwinHostProvider was expecting it when deciding whether to fall back to IPv4 if a IPv6 query failed. Therefore, the `host` resource would not function properly when resolving hostnames on Darwin. The logic has been changed to use `dig` short output and query for both v6 and v4 addresses.

Additionally, the LinuxHostProvider has been modified to prefer `dig` if it's available to keep behavior similar between Darwin and Linux whenever possible. This has the added benefit of providing v6 and v4 resolution if possible where `getent hosts` only returns v6 if v6 records exist.

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -24,6 +24,8 @@
 #   it { should be_resolvable.by('dns') }
 # end
 
+require 'resolv'
+
 module Inspec::Resources
   class Host < Inspec.resource(1)
     name 'host'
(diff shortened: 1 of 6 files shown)


lib/resources/host.rb
```diff
@@ -138,6 +140,26 @@ def missing_requirements(_protocol)
       # be enumerated in a skip_resource message
       []
     end
+
+    def resolve_with_dig(hostname)
```
chris-rock: It feels like we need a new `UnixHostProvider` class to map those functions.

adamleff: Yeah I can totally do that. You thinking DarwinHostProvider and LinuxHostProvider would subclass from it, and we'd move both `resolve_with_bind` and `resolve_with_getent` into it?

chris-rock: exactly, that was my thinking.

chris-rock (approved):
Very nice addition. Awesome work @adamleff

PR #1989, opened 2017-07-05
clarify PowerShell requirement

Happycoil:
Fixes https://github.com/chef/inspec/issues/1982


adamleff (approved):
LGTM @Happycoil - thanks! In addition, once we implement the exception stuff for #1205, we can do better pre-requisite checking to see the version of PowerShell in use and then alert the user in a more meaningful manner so they don't have to read the README.

Thanks again!

PR #1993, opened 2017-07-05
Add support for ncat in host resource for CoreOS

### LINKED ISSUE #1980: Support `ncat` and `nc` (for CoreOS)
### Description

CoreOS does not ship with `nc` and only has `ncat`. There is also no reasonable way to install `nc` on CoreOS, since it doesn't have a package manager.

* Client:
```
OS X 10.12.5

# inspec version
1.29.0
```

* Host:
```
# cat /etc/os-release
cat /etc/os-release
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1409.2.0
VERSION_ID=1409.2.0
BUILD_ID=2017-06-19-2321
PRETTY_NAME="Container Linux by CoreOS 1409.2.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```

### Replication Case

Try to do something like this on CoreOS:

```
control "ports" do
  impact 0.6
  title 'ports should be exposed.'
  desc 'ssh port should be reachable.'
  tag 'ports','ssh'
    describe host('localhost', port: 22, protocol: 'tcp') do
      it { should be_reachable }
    end
end
```

### Possible Solutions

Add checks for `ncat` in addition to `nc` and utilize whichever one is preferred and available.

### Stacktrace

N/A

adamleff:
CoreOS is considered a member of the Linux family, and the `host` resource tries to use `nc` on Linux hosts to test TCP reachability. Unfortunately, `nc` is not available on CoreOS, but `ncat` is.

This change attempts to use `nc` first, then `ncat` if it's available.

Fixes #1980

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -210,8 +210,8 @@ module Inspec::Resources
     def missing_requirements(protocol)
       missing = []
 
-      if protocol == 'tcp'
-        missing << 'netcat must be installed' unless inspec.command('nc').exist?
+      if protocol == 'tcp' && (!inspec.command('nc').exist? || !inspec.command('ncat').exist?)
+        missing << 'netcat must be installed'
       end
 
       missing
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @adamleff

PR #1995, opened 2017-07-06
host resource: fix netcat detection

adamleff:
The logic used to determine whether a viable netcat binary exists is wrong and prevents Linux hosts from doing TCP reachability checks.


chris-rock (approved):
Thank you @adamleff

PR #1999, opened 2017-07-07
Promote artifact to stable during Habitat upload

adamleff:
`pkg` commands in Habitat 0.25 now use channels and prefer the `stable` channel by default. However, artifacts uploaded with hab pkg upload go to `unstable` by default (as it should).

This change ensures that `chef/inspec` artifacts land in `stable` during our release process.


arlimus (approved):
Awesome thank you @adamleff !!

PR #2002, opened 2017-07-07
Change host resource to use getent ahosts on Linux

### LINKED ISSUE #2001: host ipaddress no longer honors /etc/hosts entries as of Inspec 1.31
### Description

The host resource no longer honors /etc/hosts entries when verifying the ipaddress of a host, as of version 1.31.0. This worked in Inspec 1.30.0.

I am using kitchen-inspec to verify an Oracle Linux 7.2 VM from my Mac running Mac OS 10.11.6.

Working Inspec version:

```
$ inspec --version
1.30.0
```

No longer working Inspec version:

```
$ inspec --version
1.31.1
```

### Replication Case

Add an entry to /etc/hosts for a host name that is also resolvable in DNS, and use Inspec to assert that the host has an IP address listed in /etc/hosts.

```
$ grep google.com /etc/hosts
127.0.0.2 google.com
```

Inspec resource:

```ruby
describe host('google.com') do
  it {should be_resolvable}
  its('ipaddress') {should include '127.0.0.2'}
end
```

Failure with Inspec 1.31:

```
  Host google.com
     ✔  should be resolvable
     ∅  ipaddress should include "127.0.0.2"
     expected ["216.58.208.142"] to include "127.0.0.2"
```

Success with Inspec 1.30:

```
  Host google.com
     ✔  should be resolvable
     ✔  ipaddress should include "127.0.0.2"
```

### Possible Solutions

This behavior changed in version 1.31.0, so I'm guessing the difference has something to do with the switch from `host` to `dig` for the ipaddress matcher, but that's just a guess.

### Stacktrace

n/a

adamleff:
In InSpec 1.31, we changed the `host` resource to use `dig` instead of `getent hosts` for name resolution because `getent hosts` does not return all entries (only the first v6 entry if it exists, then the first v4 entry) and we wanted to keep the Darwin and Linux implementation as close as possible. Unfortunately, this affected users' ability to do resolution checks for entried stored in their /etc/hosts file.

This change goes back to using `getent` for Linux and changes to `getent ahosts` which returns both v4 and v6 records. Additionally, the Darwin provider's dig implementation was reordered to return v4 addresses before v6 addresses to be consistent with how `getent ahosts` returns records.

Fixes #2001


arlimus (approved):
Wow, thank you for catching all of the tricky details here! 😄

PR #2003, opened 2017-07-07
Source reader should not hand back files with nil contents

adamleff:
If a profile has a data files directory that looks like this:

```
files/platforms/one/data.json
files/platforms/two/data.json
files/platforms/three/data.json
```

... the source reader will return the directories in the list of files but with
nil contents. This causes an issue when Inspec::Profile tries to create a sha256
checksum of the profile contents only to try to cast nil to a string when
building the null-delimited profile contents string.

Files that are empty will have an empty string as its contents, so it's safe to
assume that file entries with nil contents are actually a directory and have no
affect on the profile's checksum. Therefore, this change will eliminate any file
entries in responses from the source readers where the contents are nil.

### DIFF
--- a/lib/source_readers/inspec.rb
+++ b/lib/source_readers/inspec.rb
@@ -54,21 +54,21 @@ module SourceReaders
       tests = @target.files.find_all do |path|
         path.start_with?('controls') && path.end_with?('.rb')
       end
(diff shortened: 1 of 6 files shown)


arlimus (approved):
Great catch @adamleff thank you!!

PR #2008, opened 2017-07-10
Read info from XCCDF source

NF07:
This change would allow someone to write rules move easily, because they could specify the id and XCCDF source and would not have to type all the information in manually.

### DIFF
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -11,6 +11,7 @@ require 'inspec/profile'
 require 'inspec/metadata'
 require 'inspec/secrets'
 require 'inspec/dependencies/cache'
+require 'rexml/document'
 # spec requirements
 
 module Inspec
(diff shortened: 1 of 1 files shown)


NF07:
When writing a rule this allows you to use the tags to make writing rules easier. The format is below.

   tag xccdf_id 'id'
   tag xccdf_source: '/path/to/xccdf/file.xml'

NF07:
I will be heading back to college in a week or two. I will not have time to work on it, so if there are any changes that you want me to make please reach out. Thanks.

trevor-vaughan:
@aaronlippold This is what I was referencing the other day and it would **really** simplify the work in the collective profiles.

adamleff:
Hello, @NF07, and thanks for submitting this PR! We had a number of internal discussions about the impact and direction this has for InSpec.

While I completely understand the motivation for this pull request, it will lead us into a direction with the core InSpec project we don't want to go. The use-case you present is valid as is the need to have solution for it. However, we believe this type of functionality belongs in extensions to InSpec rather than in the core. InSpec profiles are meant to be self-contained without requiring additional data from outside the construct of the profile (unless passed in as an InSpec Attribute). Your change would go beyond this design decision.

A better pattern would be to publish a tool that consumes an XCCDF file and outputs InSpec controls. We have a number of classes that have `#to_ruby` methods that, once instantiated and populated, would output a well-formed InSpec control. In fact, we at Chef already have a tool (`inspec-scap`) that does something similar that we make available to customers who have purchased our Chef Automate offering.

Again, I want to thank you for the contribution and I apologize that we will not be able to accept it at this time.

Thanks,
~Adam

trevor-vaughan:
@adamleff I would like to note that this PR solves a specific problem that I've had with using InSpec. This problem is that the actual InSpec test is lost in all the generated prose that is available in an external source.

When I'm looking at tests, I want to be able to find the test easily and, like in most systems, cleanly separate my data from my logic. This also prevents situations in the future where a slight change to the wording of a policy will need to be propagated across numerous InSpec control repositories.

Targeting this paradigm would allow a MUCH greater breadth of control coverage with a LOT less work on the part of InSpec test developers and I would like to push for the inclusion.

It is optional, and users do not have to use it, so it becomes something that some projects can use and others can ignore.

It is unfortunate that it interferes with a paid offering, but this should not preclude the inclusion of a problem solving solution into the codebase and I, as a user, would VERY MUCH like to see it become part of InSpec so that InSpec stands a chance of becoming a NIST approved scanner (this is the first half).

adamleff:
@trevor-vaughan The fact that we have a paid offering that does conversion-to-InSpec-syntax is not the reason for this exclusion. One of InSpec's primary focus and strengths is its easy-to-understand language. Supporting additional external data formats muddies the waters.

I'd like to suggest that you, me, @arlimus, and @chris-rock get on a call to discuss this further to see how we can address your use case and see if we can't support this as a plugin or similar. But I and the other maintainers still believe that supporting a data format other than the InSpec language causes us a variety of issues moving forward.

trevor-vaughan:
@adamleff I'd definitely like to get on a call soon if we can.

In terms of an easy to understand language, linking over to a formally specified NIST document is pretty darn understandable and doesn't hit me with a wave of information that I don't really care about for the test.

What makes it easy for me to use is 1) A descriptive title 2) A good test. If I want to know the justification as to why something happened, I can either read the output report or go look at the source material. Honestly, I was *super* excited when I saw this come in because it made things easier to understand when you were looking at a test.

Having this be a plug-in would be fine as long as that plug-in could extend the language without needing some sort of weird intermediary step that makes things difficult to read.

Honestly, I think that formally adopting this pattern as an extension point would be a HUGE boon to the success of InSpec overall.

The *same* test could easily reference XCCDF docs from multiple sources and would increase code reuse and the ability to generate compliance output for multiple bodies simultaneously.

adamleff:
@trevor-vaughan thanks for your feedback - we'll reach out soon.

chris-rock:
Thank you @trevor-vaughan for your feedback, lets have a chat early next week to discuss this matter.

PR #2010, opened 2017-07-11
add Inspec::Describe for abstract describe state

arlimus:
Unlike `Inspec::Test` this supports having multiple tests within one block that describes a resource. This has now been seen as an optimization problem where a resource may be computed once and tested multiple times with `it` and `its` within the body.

If successful, it requires a follow-up to deprecated Inspec::Test and remove it for 2.0 completely with a recommendation to use Inspec::Describe.


chris-rock (approved):
Thank you @arlimus for this improvement

jquick (approved):
This is really cool and adds lots of options to our testing! Looks good to me.

PR #2023, opened 2017-07-19
Website: Disable particle animation if users prefer reduced motion

k-nut:
The website uses animations in quite a lot of places. To some people those animations can be a distraction from the text.

Safari introduced the new `prefers-reduced-motion` media query which lets users define that they prefer things not to move so much.

This simple change will check if the users made this setting (in `Accessibility > Display > Reduced Motion`) and if so only show a static version of the particles.

Would you be interested in merging something like this?

More information can be found here: https://webkit.org/blog/7551/responsive-design-for-motion/

out


adamleff:
@k-nut thank you for your PR, and please accept our welcome to the InSpec community!

I think PRs like this are wonderful... enhancing the user experience for EVERYONE is a good thing. I'd like our UX extraordinaire @hannah-radish to chime in on this one since she has more knowledge of the website design than I do.

www/source/javascripts/app.js
```diff
@@ -4,6 +4,8 @@
 
 /* To load a config file (particles.json) you need to host this demo (MAMP/WAMP/local)... */
 
+var animate = !matchMedia('(prefers-reduced-motion)').matches;
+
```
chris-rock: I would prefer if we could use `window.matchMedia('(prefers-reduced-motion)').matches`

chris-rock (changes requested):
Thank you for your improvement @k-nut I added a small request

k-nut:
@chris-rock I added the `window` as proposed (which I agree is a good idea) and pushed a rebased version.

adamleff:
@k-nut thanks again for this PR! Could you please sign-off on your commit per the [DCO requirements](https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) so that we can move this forward? thank you!

k-nut:
@adamleff 📝✅

adamleff (approved):
Thanks @k-nut!

adamleff:
I loaded this up in the acceptance environment and it works as described, as well.

chris-rock (approved):

PR #2027, opened 2017-07-24
bugfix: empty file strings from archive readers

arlimus:
Empty files in archives are sometimes possible (we just ran into this with TGZ), but is never a valid file to extract. So remove it and discount it altogether. Changed structure to support testing of these global calls.


chris-rock (approved):
Thank you @arlimus

PR #2043, opened 2017-08-02
catch newline issues in xinet.d

arlimus:
The fix is already provided here: https://github.com/chef/inspec/pull/2040

This PR only adds a unit test to catch it

As discovered and fixed by @kareiva


chris-rock (approved):
Thank you @arlimus

PR #2046, opened 2017-08-03
Prep for 1.32.0 release

adamleff:
Version bump, changelog updates for period between when expeditor was enabled and when v1.31.1 was released


VERSION
```diff
@@ -1 +1 @@
-1.31.8
\ No newline at end of file
+1.32.0
```
tduffield, line 1: Just for clarity, Expeditor will automatically bump this to 1.32.1 (unless you skip the version bump)

adamleff: Yup, I'm OK with that. :) Gives me something to talk to in the upcoming release announcement for people to understand that patch version numbers will likely be higher from now on.

PR #2048, opened 2017-08-03
Remove changelog generator stuff

adamleff:
Version bumping and changelog generation has been moved to Chef's expeditor tool. The github_changelog_generator bits and the Rake tasks that use them are no longer necessary.

Also, the old historical changelog has been merged back into the main CHANGELOG.md file since Expeditor does not alter any existing changelog content like github_changelog_generator does, so it's safe to bring this stuff back.


arlimus (approved):
Thank you Adam, makes sense :)

PR #2050, opened 2017-08-03
Add label to skip all Expeditor tasks

adamleff:
In the event we have a docs-only change or similar which does not necessitate a version bump, changelog update, or omnibus build, a single label named `Expeditor: Skip All` can be used rather than adding individual labels.


arlimus (approved):
Nice thank you  👍

PR #2052, opened 2017-08-03
Fix docker_container.tag to use last element of image

### LINKED ISSUE #2051: docker_container.tag should use last element of split
### Description

A docker image may contain multiple colons if the repository has a port specified. The `tag` attribute apparently only does a split on `image` and takes the second element. This will result in the port and image name instead of the tag.

1.32.1

### Replication Case

Add a port number to a repository and pull an image. `docker ps` will show you something like:
```
CONTAINER ID        IMAGE                                                                 COMMAND              CREATED             STATUS              PORTS               NAMES
8aad75a0a0c4       mydocker.example.com:5000/some/image:1.2.3   "/start.sh"          24 hours ago        Up 24 hours                             mycontainer
```

In your InSpec test:
```
describe docker_container('mycontainer') do
  its('tag') { should eq '1.2.3' }
end
```

And running it will yield:
```
  Docker Container
     ∅  mycontainer tag should eq "1.2.3"
     
     expected: "1.2.3"
          got: "5000/some/image"
     
     (compared using ==)
```

### Possible Solutions

Change https://github.com/chef/inspec/blob/master/lib/resources/docker_container.rb#L81 to use `-1` index.

mattlqx:
Fixes #2051

Images with repos containing port numbers will have multiple colons.

Signed-off-by: Matt Kulka <mkulka@parchment.com>


arlimus (approved):
Great bugfix thank you @mattlqx !!

lib/resources/docker_container.rb
```diff
@@ -78,7 +78,7 @@ def repo
     end
 
     def tag
-      image.split(':')[1] unless image.nil?
+      image.split(':')[-1] unless image.nil? || !image.index(':')
```
arlimus: @jerryaldrichiii added the suggestion to change this from `index` to `include?` :)

adamleff: Splitting on the colon like this could potentially introduce a problem if a non-default repohost was present but no tag was present. For example, if we received `docker ps` output that included:  `repo.example.com:5000/ubuntu`, the tag we'd return would be `5000/ubuntu`

I'd rather see us treat the non-default repo host as a standalone unit and split on `/`, and we need to fix both `#repo` and `#tag`... perhaps something like this:

```ruby
def repo
  return if image_name_from_image.nil?
  image_name_from_image.split(':')[0]
end

def tag
  return if image_name_from_image.nil?
  image_name_from_image.split(':')[1]
end

def image_name_from_image
  return if image.nil?
  # possible image names include:
  #   alpine
  #   ubuntu:14.04
  #   repo.example.com:5000/ubuntu
  #   repo.example.com:5000/ubuntu:1404
  image.include?('/') ? image.split('/')[1] : image
end

```

What do you think, @mattlqx?

adamleff (changes requested):
Hi @mattlqx and thanks for your PR!

Thanks for finding this bug and proposing a fix. I have a small concern about splitting on a colon for a particular edge case, and I've left a comment accordingly. Please let me know what you think.

mattlqx:
Thanks for bringing up this additional case. Your suggestion works well, so let's go with that. I've added an additional test case for it as well.

adamleff (approved):
Thanks for your first PR to InSpec, @mattlqx! This LGTM. 👍

chris-rock (approved):
Great fix @mattlqx

PR #2056, opened 2017-08-10
fix command.exists for mock environments

chris-rock:
…since inspec.os.name was “” instead of unknown. I changed to nil to catch that case.


chris-rock:
Rebased, all green now

adamleff (approved):

PR #2069, opened 2017-08-14
Add missing command mocks to fix tests after train 0.26.1 upgrade

adamleff:
Train 0.26.1 fixes a bug where missing mocked commands would return a zero exit code. This resulted in some unit tests failing due to missing mocked commands.


chris-rock (approved):
Thank you @adamleff

PR #2070, opened 2017-08-14
add mock support for os_env resource

chris-rock:
This PR prevents issues when the `os_env` resource is called via `inspec check`. This issues appears now, since we doing the return handling right in train for mock requests https://github.com/chef/train/pull/187

```
/Users/chartmann/Development/compliance/inspec/lib/resources/os_env.rb:61:in `value_for': undefined local variable or method `os' for Environment variable DOCKER_CONTENT_TRUST:#<Class:0x007fd1a4cdd700> (NameError)
	from /Users/chartmann/Development/compliance/inspec/lib/resources/os_env.rb:29:in `initialize'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/plugins/resource.rb:47:in `initialize'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/resource.rb:47:in `new'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/resource.rb:47:in `block (3 levels) in create_dsl'
	from /Users/chartmann/Development/hardening/cis-docker-benchmark/controls/container_images.rb:111:in `block in load_with_context'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/rule.rb:49:in `instance_eval'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/rule.rb:49:in `initialize'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/control_eval_context.rb:71:in `new'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/control_eval_context.rb:71:in `block (2 levels) in create'
	from /Users/chartmann/Development/hardening/cis-docker-benchmark/controls/container_images.rb:99:in `load_with_context'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile_context.rb:146:in `instance_eval'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile_context.rb:146:in `load_with_context'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile_context.rb:130:in `load_control_file'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile.rb:151:in `block in collect_tests'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile.rb:148:in `each'
	from /Users/chartmann/Development/compliance/inspec/lib/inspec/profile.rb:148:in `collect_tests'
[...]

PR #2072, opened 2017-08-15
Moves logic from os_env from initialize phase to runtime phase

chris-rock:
While #2070 addresses an issue where `os_env` is called during `inspec check` the solution was not great since we need to add special mock handling. This PR is the second iteration that moves the logic to runtime instead of object creation time.


adamleff (approved):
Thanks for this cleaner implementation, @chris-rock.

Issue #2041, opened 2017-08-01
Errors packaging and validating wrapper profiles having dependencies


tpetchel:
### Description

While specifying dependencies from a wrapper profile, I can use `url`, but not the other fields, to package and validate my profile.

* 👍 `url`
* 🙅‍♂️ `git`
* 🙅‍♂️ `github`
* 🙅‍♂️ `compliance`

### InSpec and Platform Version

```
$ inspec version
1.31.1
```

macOS 10.12.4
Chef Automate 1.5.46  (using Chef Automate Pilot)

### Replication Case

* Run `inspec init profile` to create a profile.

```
$ inspec init profile issue-2041
Create new profile at /Users/thomaspetchel/scratch/profiles/issue-2041
 * Create directory controls
 * Create file controls/example.rb
 * Create file inspec.yml
 * Create directory libraries
 * Create file README.md
 * Create file libraries/.gitkeep
```

* Modify `inspec.yml` like each of the examples that follow.

### `url`

```
name: issue-2041
title: InSpec Profile
maintainer: The Authors
copyright: The Authors
copyright_email: you@example.com
license: Apache-2.0
summary: An InSpec Compliance Profile
version: 0.1.0
depends:
  - name: apache-baseline
    url: https://github.com/dev-sec/apache-baseline/archive/master.tar.gz
    #git: https://github.com/dev-sec/apache-baseline.git
    #github: dev-sec/apache-baseline
    #compliance: admin/apache-baseline
```

`inspec vendor` passes.

```
$ inspec vendor issue-2041 --overwrite
Dependencies for profile issue-2041 successfully vendored to issue-2041/vendor
```

`inspec check` also succeeds.

```
$ inspec check issue-2041
`command(httpd).exist?` is not suported on your OS:
Location:    issue-2041
Profile:     issue-2041
Controls:    2
Timestamp:   2017-08-01T13:24:10-04:00
Valid:       true

No errors or warnings
```

### `git`

```
name: issue-2041
title: InSpec Profile
maintainer: The Authors
copyright: The Authors
copyright_email: you@example.com
license: Apache-2.0
summary: An InSpec Compliance Profile
version: 0.1.0
depends:
  - name: apache-baseline
    #url: https://github.com/dev-sec/apache-baseline/archive/master.tar.gz
    git: https://github.com/dev-sec/apache-baseline.git
    #github: dev-sec/apache-baseline
    #compliance: admin/apache-baseline
```

`inspec vendor` succeeds.

```
[...]

PR #2075, opened 2017-08-15
Remove github fetcher from docs

adamleff:
The github fetcher does not exist. The git fetcher or URL fetcher is what users should use.

Related to #2041


chris-rock (approved):
Thank you for that cleanup @adamleff

PR #2082, opened 2017-08-17
Remove www Gemfile.lock

adamleff:
No reason to keep the Gemfile.lock in the www directory under source control. It adds extra steps to the www release process after an InSpec release because the release process will fail if the git working tree is not clean (and since the www Rake tasks rely on InSpec stuff, the Gemfile.lock will be perpetually out of date).


arlimus (approved):
Yes please thank you!!

PR #2084, opened 2017-08-17
Publish rubygem via expeditor

adamleff:
When InSpec is promoted to the `stable` channel, Expeditor will take care of the gem publishing!

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -34,5 +34,5 @@ merge_actions:
 
 artifact_actions:
   promoted_to_stable:
-    built_in:publish_rubygems:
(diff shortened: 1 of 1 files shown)


tduffield (changes requested):
So this action only supports gems that are available via artifactory. Based on what I can see, I don't see any gems in artifactory for inspec. Can you confirm that this is the case. 

Also, Expeditor now supports array syntax:

```
artifact_actions:
  promoted_to_stable:
    - built_in:publish_rubygems
    - built_in:rollover_changelog
```

tduffield (approved):

PR #2088, opened 2017-08-18
OVAL Results Output

NF07:
Does not work properly.
I am going back to school so I just wanted to put my code out there in case anyone wants to pick it up.
In order to have a valid OVAL Results Output the following have to be fixed.
Wrong namespace for schemaLocation it is supposed to be xsi: not xmlns:.
Incorrect time format.
The system characteristics are missing some aspects and does not perform well in general.

Signed-off-by: Nick Francino <nicksf07@gmail.com>


NF07:
This is the other half of my first PR #2008.

PR #2093, opened 2017-08-22
File size lines

MarkGibbons:
Add a file resource matcher to check the size of a file in lines.


lib/resources/file.rb
```diff
@@ -44,7 +44,7 @@ def initialize(path)
     %w{
       type exist? file? block_device? character_device? socket? directory?
       symlink? pipe? mode mode? owner owned_by? group grouped_into?
-      link_path linked_to? mtime size selinux_label immutable?
+      link_path linked_to? mtime size size_lines selinux_label immutable?
```
chris-rock, line 47: There is no need to add this here, since the implementation is done in this file anyway

chris-rock (commented):
Is there a reason why we need this function?

```
describe file('/proc/cpuinfo') do
  its('size_lines') { should be 0 }
end
```
can also be written as:
```
describe file('/proc/cpuinfo') do
  its('content.lines.size') { should cmp 0 }
end
```

lib/resources/file.rb
MarkGibbons: Thank you. I didn't know you could do that with content or other matchers inside its.  Even afterwards searching using google I can't find any similar examples.  

It's a minor edge case but is there anyway to put code like this "content ? content.lines.count : 0"  inside the its().  I'd like to return size 0 in the case where a file does not exist.

MarkGibbons:
Not needed.

adamleff:
@MarkGibbons re: your question about defaulting to `0` instead of `nil`:

You wouldn't be able to use the `its` syntax for that without presetting some variables beforehand, and that will likely make the CLI output not very helpful should it fail.

You could try using the `expect`-style syntax instead:

```ruby
describe file('/path/to/file') do
  expect(subject.content.nil? ? 0 : content.lines).to be > 100
end
```

... or something like that.

PR #2094, opened 2017-08-22
Refine the profile/test summary output of the CLI formatter

### LINKED ISSUE #1752: Improve 'Test Summary" Coloring; especially "0 Failures" being in Red
I suspect the following behavior is inherited but here's a complaint anyway...I'll be happy to post an issue elsewhere if you can give me a hint as to where.

`Test Summary: 1 successful, 0 failures, 0 skipped`

The above output, when in color, is alarming. The 0 failures show up as a nice bright red (using bash shell on Windows 10 presently).

Might I suggest that "0 failures" be output in grey instead. In a similar fashion a non-zero skipped might warrant something like a cyan output to flag its non-zero presence while leaving the existing grey for '0 skipped'. This later item being less bothersome than seeing red on a test output and having to read the result to realize that it isn't actually a problem.

Running via Chef Development Kit Version: 1.3.43

adamleff:
* The "Profile Summary" is misleading as it's not a summary of profile success/failure but rather the controls within the profile(s). Altered the output to be clear. I still like calling it the "profile summary" but wanted to add clarity that the numbers are about the controls.

* Made the colorized output dynamic. The success/failure will only be green/red if there are controls/tests that fall into that category. That way we are not printing red failure text when there are no actual failures. Fixes #1752.

* Cleaned up some grammar issues. ("1 failure" vs "1 failures")

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -613,10 +613,18 @@ class InspecRspecCli < InspecRspecJson # rubocop:disable Metrics/ClassLength
     summary = profile_summary
     return unless summary['total'] > 0
 
+    success_str = summary['passed'] == 1 ? '1 successful control' : "#{summary['passed']} successful controls"
+    failed_str  = summary['failed']['total'] == 1 ? '1 control failure' : "#{summary['failed']['total']} control failures"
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Fantastic improvement and I really enjoy the updated output!!

PR #2095, opened 2017-08-22
[WIP] docs: add documentation for file resource sgid, sticky, and suid matchers

### LINKED ISSUE #2032: The sticky bit function does not seem to be documented in the file resource

clintoncwolfe:
Fixes #2032 

Doc entries for sgid, sticky, and suid matchers.  Unit tests appear to already exist.  Docs may be a bit chatty, I'm happy to make them more concise if needed.

Update: While there are methods on the File resource for these three, there are not actually matchers for them.


clintoncwolfe:
So, in putting together a simple PR for this, I assumed that that sticky (and sgid and suid, which were also undocumented) were simple boolean matchers on File - so you could say 
```
it { should be_sticky } 
```
Which reads OK to me, and is similar to the be_readable matcher, which is also looking at permissions.  

But in fact, it's implemented as a non-predicate, so you have to access it as an attribute of the subject, like this:
```
its('sticky') { should be true } 
```

That reads odd to me - it's not like the other permissions predicates; and it would read better as `its('sticky_bit') ...` anyway.

I propose it be implemented as a real matcher (and while we're at it, rename sgid to setgid and setuid, which sound better to my ears).  Then you'd have:

```
file('/foo') do
  it { should be_sticky }
  it { should be_setgid }
  it { should be_setuid }
end
```

adamleff:
@clintoncwolfe I apologize, but a miscommunication internally caused us to accidentally push some commits to your branch, and the work was not rebased so now your PR includes a bunch of stuff it shouldn't. I'm very sorry about that.

Would you mind terribly either force-pushing from your local copy, or closing this PR and resubmitting so we get a clean trail?

Also, I completely agree with you that the setuid/setgid stuff is worthy of having regular matchers like `be_setgid` and I would totally welcome that change.

clintoncwolfe:
No worries!  I have to step out for my kid's school orientation, but I'll open a new one later, with be_setgid, etc.

adamleff:
Thanks @clintoncwolfe - looking forward to it!

PR #2097, opened 2017-08-24
pip resource: support non-default pip locations, such as virtualenvs

### LINKED ISSUE #516: describe pip not working for packages in virtualenv?
Hi,

I am testing 

```
describe pip('s3cmd') do
  it { should be_installed }
end
```

while installing it via 

```
python_package service do
  version '1.6'
  virtualenv "/home/#{service}/.env"
end
```

works fine, the test fails:

```
Failures:

  1) Pip Package  s3cmd should be installed
     Failure/Error: it { should be_installed }
       expected that `Pip Package  s3cmd` is installed
     # ./test/integration/default/default_spec.rb:8:in `block (2 levels) in load'

Finished in 0.48711 seconds (files took 2 minutes 20.6 seconds to load)
2 examples, 1 failure

Failed examples:

rspec  # Pip Package  s3cmd should be installed
```

Is this just not supported or am I doing sth wrong?

Cheers
Jannis

tonybaloney:
Fixes #516

Implements optional path to pip executable (like when virtualenv is used) instead of using system PATH.
Tested by installing newer version of Django in virtualenv and testing for it
```ruby
describe pip('django', '/Users/anthonyshaw/repo/wayfinder/wayfinder/bin/pip') do
  it { should be_installed }
  its('version') { should eq('1.11.4')}
end
describe port(80) do
  it { should_not be_listening }
end
```

Before:

```
somepersonslaptop:inspec anthonyshaw$ bundle exec bin/inspec exec test.rb

Profile: tests from test.rb
Version: (not specified)
Target:  local://


  Pip Package
     ∅  django version should eq "1.11.4"

     expected: "1.11.4"
          got: nil

     (compared using ==)

  Port 80
     ✔  should not be listening

Test Summary: 1 successful, 2 failures, 0 skipped
```

After 
```
somepersonslaptop:inspec anthonyshaw$ bundle exec bin/inspec exec test.rb

Profile: tests from test.rb
Version: (not specified)
Target:  local://


  Pip Package
     ✔  django should be installed
     ✔  django version should eq "1.11.4"
  Port 80
     ✔  should not be listening

Test Summary: 3 successful, 0 failures, 0 skipped
```


tonybaloney:
sorry. Forgot DCO. I'll amend the commit

lib/resources/pip.rb
```diff
@@ -15,10 +15,16 @@ class PipPackage < Inspec.resource(1)
...
       end
+
+      describe pip('django', '/path/to/virtualenv/bin/pip') do
+        it { should be_installed }
+        its('version') { should eq('1.11.4')}
+      end
     "
 
-    def initialize(package_name)
+    def initialize(package_name, virtualenv_path=nil)
```
srenatus: [nit] I think `virtualenv_path` is a bit confusing here: From looking at this, I'd expect that I was to put `/path/to/virtualenv` in there, not `/path/to/virtualenv/bin/pip`. No idea what's the best approach, though.

lib/resources/pip.rb
```diff
@@ -75,7 +81,7 @@ def pip_cmd
           return nil
         end
       end
-      pipcmd || 'pip'
+      pipcmd = @virtualenv_path.nil? ? 'pip' : @virtualenv_path
```
srenatus: [nit] I think to keep the logic that was there before -- this would maybe better be something like this:
```ruby
pipcmd || @virtualenv_path || 'pip'
```
So, the order of precedence would be: if the `pip` is found (windows), take that. Otherwise, take `@virtualenv_path`; fall back to 'pip'.

Makes me wonder if this shouldn't be like this, then:
```ruby
@virtualenv_path || pipcmd || 'pip'
```

srenatus (commented):
Looks good. I've got some inline comments, also, an added test case would be awesome, too (https://github.com/chef/inspec/blob/master/test/unit/resources/pip_test.rb)

lib/resources/pip.rb
tonybaloney: agreed. `pip_path` would make sense. OR having it test for the existence of `@virtualenv_path` + `/bin/pip` before executing.

lib/resources/pip.rb
tonybaloney: @srenatus I updated it. let me know if the new one makes more sense? I'm new to this so feel free to [nit]

lib/resources/pip.rb
```diff
@@ -75,7 +81,7 @@ def pip_cmd
           return nil
         end
       end
-      pipcmd || 'pip'
+      @virtualenv_path.nil? ? pipcmd || 'pip' : @virtualenv_path
```
srenatus: So, if `@virtualenv_path` is _nil_,  `@virtualenv_path.nil?` is _true_. However, `@virtualenv_path` will also be understood as "falsy" by ruby, so that line is equivalent to
```ruby
@virtualenv_path || pipcmd || 'pip'
```

tonybaloney:
@srenatus decided to solve both problems. it now works with a directory (and adds /bin/pip to the path) and then the flow through logic is cleaner as well

```ruby
describe pip('django') do
  it { should be_installed }
  its('version') { should eq('1.10.5')}
end
describe pip('django', '/Users/anthonyshaw/repo/wayfinder/wayfinder/') do
  it { should be_installed }
  its('version') { should eq('1.11.4')}
end
describe pip('django', '/Users/anthonyshaw/repo/wayfinder/wayfinder/bin/pip') do
  it { should be_installed }
  its('version') { should eq('1.11.4')}
end
```

Gives
```
somepersonslaptop:inspec anthonyshaw$ bundle exec bin/inspec exec test.rb

Profile: tests from test.rb
Version: (not specified)
Target:  local://


  Pip Package
     ✔  django should be installed
     ✔  django version should eq "1.10.5"
  Pip Package
     ✔  django should be installed
     ✔  django version should eq "1.11.4"
  Pip Package
     ✔  django should be installed
     ✔  django version should eq "1.11.4"

Test Summary: 6 successful, 0 failures, 0 skipped
```

srenatus:
@tonybaloney I'm not one of @chef/inspec-maintainers, but I'm sure they'd appreciate an added testcase or two? 😉

lib/resources/pip.rb
```diff
@@ -15,10 +16,16 @@ class PipPackage < Inspec.resource(1)
...
       end
+
+      describe pip('django', '/path/to/virtualenv/bin/pip') do
+        it { should be_installed }
+        its('version') { should eq('1.11.4')}
+      end
     "
 
-    def initialize(package_name)
+    def initialize(package_name, virtualenv_path = nil)
```
adamleff: Let's change this variable to `pip_path` since you can technically install pip in a "non-PATH" location without using a virtualenv, and the examples given are actually paths to an actual `pip` command, not the top-level directory of the virtualenv.

lib/resources/pip.rb
```diff
@@ -15,10 +16,16 @@ class PipPackage < Inspec.resource(1)
...
       @package_name = package_name
+      @virtualenv_path = virtualenv_path
```
adamleff: See above for my comment re: `virtualenv_path`. Instead, I think we should set an instance variable that contains the validated path to `pip`, and change the `#pip_cmd` method name to be clearer... maybe `resolve_pip_path`?

So, assuming you take my advice to change the argument name to `pip_path` above, it would look something like this:

```ruby
@pip_cmd = resolve_pip_path(pip_path)
return skip_resource "pip not found" if @pip_cmd.nil?
```

... and then `resolve_pip_path` would change a bit to support that. Does that make sense?

lib/resources/pip.rb
```diff
@@ -57,7 +64,14 @@ def to_s
     def pip_cmd
       # Pip is not on the default path for Windows, therefore we do some logic
       # to find the binary on Windows
-      if inspec.os.windows?
+      if !@virtualenv_path.nil?
```
adamleff: This is a real good first attempt, @tonybaloney! I think it's a better user experience to just assume the user is going to give us a path to `pip` directory rather than adding logic to test if the path is a dir and then assume it's a virtualenv.

How about logic like this:

```
 - if user gave us a pip path, verify it exists, return nil if it doesn't, otherwise return what they gave us
 - if on windows, run the powershell to find the full pip path. Return nil if we can't find anything
 - Otherwise, assume it's just "pip", check to see if it exists with command('pip').exist?... return 'pip' if it does, return nil if it doesn't
```

Then, in `initialize`, we can call `skip_resource` if pip_cmd returns nil with helpful error messaging.

adamleff (changes requested):
@tonybaloney welcome to the InSpec community and congratulations on your first contribution!

I've left you some suggestions in an effort to streamline the resource and provide better error messaging to our users.

As @srenatus mentioned, we should definitely add a test for when a user provides a custom pip path. The process would look something like this:
 * Add a test that invokes a resource with a custom pip path - maybe test for django this time instead of jinja2 currently tested for? It would go around here: https://github.com/chef/inspec/blob/master/test/unit/resources/pip_test.rb#L14
 * Add a mock for the non-standard pip location (see https://github.com/chef/inspec/blob/master/test/helper.rb#L205 for an example)
 * Add the mock data for the non-standard pip call. Make it different than the existing pip mock so we're 100% certain the right thing is being tested (see https://github.com/chef/inspec/blob/master/test/unit/mock/cmd/pip-show-jinja2 for an example, maybe name it "non-standard-pip-show-django"

I'd love to help you with the tests if you have any questions or concerns about them. Just let me know.

Again, THANK YOU for contributing to InSpec! It's great to have you.

tonybaloney:
@adamleff those changes have all been done. However the tests are failing, I think it's because it's checking for the existence of either these fake files and, the pip command which probably isn't on the Docker image. Or I may have made some other mistake

adamleff:
@tonybaloney thanks for your patience! I just pushed up a commit to your branch to do a bit of refactoring in order to solve your unit test dilemma.

Here's what happened... the logic in `resolve_pip_path` was correct, but because you didn't mock out the internal commands InSpec would run to see if even the default `pip` command would exist in the PATH, the `@pip_cmd` variable had `nil`. This cause your pip show mocks to never run because instead of calling `pip show django`, InSpec was actually calling `show django`

The way I found this was to use the `m` gem/command to run a subset of the unit tests... the test output can be super noisy, so getting only the output from a test/file I care about is helpful. You can run these yourself like this:

```
bundle install
bundle exec m test/unit/resources/pip_test.rb
```

... this will run only the unit tests located in `pip_test.rb` - that's how I saw the "command not mocked" output that led me to that discovery.

Please take a look at the commit I sent up and let me know if you have any questions. Basically, I moved some of the default handling back up to `initialize` and moved any handling of an unfound command to only the `skip_resource` call.  That way, the unit tests will still have a `@pip_cmd` of `pip` even if none is available.

If it looks good to you, and you're OK with it, I'll give this another review and we can work on getting this merged.

Thank you again for your help with this pull request!

tonybaloney:
all makes sense. I'll do some more reading before the next change 
all ok to merge 👍

adamleff (approved):
Super work, @tonybaloney! Thank you for your contribution, and congratulations on your first pull request to InSpec! Welcome to the community 🙂 

Once another maintainer reviews and approves, we'll merge this in!

arlimus (approved):
Amazing work @tonybaloney thank you so much!! 😁

arlimus:
And kudos @adamleff and @srenatus for all your help with the review!!

PR #2101, opened 2017-08-25
Support mixed-case group entries

adamleff:
The `group` resource downcased the input parameter unless the target
was a Windows node. However, it's completely legitimate for a Unix-y
node to have mixed case group and passwd entries.

This change does have the potential to break people that did not carefully
match their case when searching for a group, but we're currently blocking
people from using the group resource properly if they have mixed-case
entries.


arlimus (approved):
Great catch, thank you @adamleff !!

Issue #82, opened 2015-10-06
add xml resource


arlimus:
Support reading XML files via a resource

arlimus:
Kudos to @jonathanmorley for solving this one without additional dependencies 🎉

PR #2107, opened 2017-08-30
Add support for XML files

jonathanmorley:
To address #82


adamleff (changes requested):
Hi, @jonathanmorley, and thanks for submitting this PR!

Unfortunately, we currently have an (undocumented, shame on us) anti-nokogiri policy. It causes us too many issues, ranging from causing us issues trying to include it in Chef omnibus releases, to some of our users not being able to install the InSpec gem because their systems don't permit building of native extensions.

Any chance you can rework this to use REXML out of the Ruby stdlib instead?

adamleff (approved):
Solid, @jonathanmorley! Thank you for this awesome contribution!

arlimus (approved):
Amazing work and really fast turnaround @jonathanmorley , thank you so much!!!
I wasn't even aware rexml could do all of this from just within the stdlib, Kudos!!

PR #2109, opened 2017-08-30
Add with_toc_data option to markdown renderer

mivok:
This adds 'id="foo"' attributes to headers, which lets you link directly
to a specific section of the documentation.

Note: this doesn't actually generate a table of contents on any page, but it provides the anchors for you to link directly with something like <https://www.inspec.io/docs/reference/profiles/#profile_files>.

### DIFF
--- a/www/config.rb
+++ b/www/config.rb
@@ -54,4 +54,4 @@ activate :directory_indexes
 activate :syntax
 set :trailing_slash, false
 set :markdown_engine, :redcarpet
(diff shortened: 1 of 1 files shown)


adamleff (changes requested):
Thanks @mivok! I keep meaning to do this and then keep forgetting.

Can you please amend your commit with a DCO sign-off so I can merge? I'm planning on including this with the website release during tomorrow's InSpec release. Thanks!

adamleff (approved):

PR #2110, opened 2017-08-30
port resource: support ss instead of netstat

### LINKED ISSUE #1655: net-tools/netstat deprecated
### Description

net-tools package (which used to provide netstat) is deprecated.
See here: https://bugzilla.redhat.com/show_bug.cgi?id=1119297

This affects all newer Linux distro that provides minimal installation (RHEL 7.x, CentOS 7.x, Ubuntu 16.04) - ie all deployments in cloud that are using inspec as a validation tool.

RHEl 7.x
CentOS 7.x
Ubuntu 16.04

Inspec versions:
1.15.0 up to 1.19.1

### Replication Case

Use kitchen to run integration testing on an webserver cobook.
Platform is RHEL 7.x.
inpsec test contains the following piece of code:
` 
 describe port.where { protocol =~ /tcp/ && port == ihs_port } do
    it { should be_listening }
    its('processes') { should include 'httpd' }
    its('processes.length') { should eq 1 }
  end
`
The result is:
`
  ✖  webserver-restart: Server: HTTP Server deployment - Service restart behaviour (IHS) (2 failed)
     ✖  Port  with protocol =~ /tcp/ port == 80 should be listening
     expected `Port  with protocol =~ /tcp/ port == 80.listening?` to return true, got false
     ✖  Port  with protocol =~ /tcp/ port == 80 processes should include "httpd"
     expected [] to include "httpd"
`

### Possible Solutions

[Use ss instead of netstat](https://linux.die.net/man/8/netstat).
>This program is obsolete. Replacement for netstat is ss. Replacement for netstat -r is ip route. >Replacement for netstat -i is ip -s link. Replacement for netstat -g is ip maddr.

adamleff:
`netstat` is officially deprecated and is replaced with `ss`. This PR changes the port resource to use `ss` if it's available on the target system.

Resolves #1655

### DIFF
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -264,7 +264,7 @@ module Inspec::Resources
   end
 
   # extract port information from netstat
-  class LinuxPorts < PortsInfo
+  class LinuxPorts < PortsInfo # rubocop:disable Metrics/ClassLength
     ALLOWED_PROTOCOLS = %w{tcp tcp6 udp udp6}.freeze
 
     def info
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -265,9 +265,33 @@ module Inspec::Resources
 
   # extract port information from netstat
   class LinuxPorts < PortsInfo
+    ALLOWED_PROTOCOLS = %w{tcp tcp6 udp udp6}.freeze
+
     def info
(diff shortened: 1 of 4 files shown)


arlimus (approved):
Fantastic addition and improvement thank you so much @adamleff !!

PR #2117, opened 2017-09-01
Modify linux regular expression to handle process names with spaces

### LINKED ISSUE #1497: Processes resource fails in certain cases
### Description

Trying to use this spec:

```
control "agent-processes" do
    impact 1.0
    title "Testing status of Zabbix Agent processes"
    describe processes('zabbix_agentd') do
        its('users') { should include "zabbix" }
    end
end
```

Fails on certain distributions of Linux.

inspec version 1.11.0. Host OS is Ubuntu 14.04, running Kitchen using dokken. Tested on Guest OS in Dockers: Ubuntu 14.04 and 16.04, CentOS 6 and 7 .

### Replication Case

1. Run a Docker using CentOS 6: `docker run -it --rm centos:6 bash`
2. Run the processes regex from processes.rb: `ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command`
3. Note that all results are spread across two lines, breaking the matching. There are also spaces in the label.

Example output
```
[root@38ea131e50f3 /]# ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command
LABEL                             PID %CPU %MEM    VSZ   RSS TT       STAT  STARTED     TIME USER                             COMMAND
docker-default (enforce)
           1  0.0  0.0  11504  2624 ?        Ss   13:32:10 00:00:00 root                             bash
docker-default (enforce)
          13  0.0  0.0  13376  1876 ?        R+   13:32:22 00:00:00 root                             ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command
```

### Possible Solutions

A new regex was added to add labels on Linux in this commit: https://github.com/chef/inspec/commit/1811eb6666321659ada2a3fde66f4b70e9d102bb#diff-2d97e98326803a2c8b96f252c362f318R50 .

The regex is: 
`^([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+(\w{3} \d{2}|\d{2}:\d{2}:\d{2})\s+([^ ]+)\s+([^ ]+)\s+(.*)$`  .

This regex assumes that there are no spaces in the process label, which is not correct in this case.

I see a few options:

1. Remove label support (probably unwanted).
2. Fix the regex to support situations when results are spread across two/multiple lines and there are spaces in the label.
3. Add an option to disable labels.

ChadScott:
Fixes #1497

Tested on Ubuntu 14.04, Ubuntu 16.04, and CentOS 7.3

### DIFF
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -81,7 +81,7 @@ module Inspec::Resources
 
       if os.linux?
         command = 'ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command'
-        regex = /^(.+)\s+(\d+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+(\w{3} \d{2}|\d{2}:\d{2}:\d{2})\s+([^ ]+)\s+([^ ]+)\s+(.*)$/
+        regex = /^(.+?)\s+(\d+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+(\w{3} \d{2}|\d{2}:\d{2}:\d{2})\s+([^ ]+)\s+([^ ]+)\s+(.*)$/
       elsif os.windows?
         command = '$Proc = Get-Process -IncludeUserName | Where-Object {$_.Path -ne $null } | Select-Object PriorityClass,Id,CPU,PM,VirtualMemorySize,NPM,SessionId,Responding,StartTime,TotalProcessorTime,UserName,Path | ConvertTo-Csv -NoTypeInformation;$Proc.Replace("""","").Replace("`r`n","`n")'
         # Wanted to use /(?:^|,)([^,]*)/; works on rubular.com not sure why here?
--- a/lib/resources/processes.rb
+++ b/lib/resources/processes.rb
@@ -81,7 +81,7 @@ module Inspec::Resources
 
       if os.linux?
         command = 'ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command'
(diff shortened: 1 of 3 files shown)


adamleff:
Hi, @ChadScott! Thank you for your first PR to InSpec!

This is a great change, thank you. Can you please enhance the unit tests for this resource so that we make sure we don't have a future regression for this use case?

You'll need to add a sample line to the mock call here: https://github.com/chef/inspec/blob/master/test/unit/mock/cmd/ps-axoZ

... and then modify the tests accordingly here: 

https://github.com/chef/inspec/blob/master/test/unit/resources/processes_test.rb

Let me know if any of this is unclear or if you need any assistance. Thank you again!

adamleff (changes requested):
Also @ChadScott it looks like you've got some failing unit tests for the existing behavior (lots of extra untrimmed space, for example). Please take a look at the Travis job and you'll see what I mean.

ChadScott:
@adamleff the tests are failing in one case and it looks like it's because the `.+` is greedy instead of the `\s+`... I'm looking into how to better optimize that. Stand by.

ChadScott:
a3b70ab addresses your concerns, I believe.

ChadScott:
Well, now it appears to be failing on my CentOS instance. <sigh>

Let me hack at it a bit more.

ChadScott:
Disregard my local failure. It was the wrong gem (wasn't using my local copy)... I've validated this change against a real-world environment on Ubuntu 14.04, Ubuntu 16.04, and CentOS 7.3.

Should be okay to merge.

adamleff (approved):
Looks solid to me! Thanks, @ChadScott!

arlimus (approved):
Sweet, thank you so much @ChadScott !!
kudos @adamleff for your review

PR #2119, opened 2017-09-04
add nginx_conf accessors for http, servers, and locations

### LINKED ISSUE #1979: inspec.io docs for nginx_conf resource
Hey @arlimus, can you write up a doc page for your awesome new `nginx_conf` resource you added in #1889? :)

arlimus:
Resolves #1979

### DIFF
--- a/docs/resources/nginx_conf.md.erb
+++ b/docs/resources/nginx_conf.md.erb
@@ -1,0 +1,122 @@
+---
+title: About the nginx_conf Resource
+---
+
+# nginx_conf
+
+Use the `nginx_conf` InSpec resource to test configuration data for the NginX server located at `/etc/nginx/nginx.conf` on Linux and Unix platforms.
+
+**Stability: Experimental**
+
+## Syntax
+
+An `nginx_conf` resource block declares the client NginX configuration data to be tested:
+
+    describe nginx_conf.params['pid'] do
+      it { should cmp 'logs/nginx.pid' }
+    end
+
+where
+
+* `nginx_conf` is the resource to reference your NginX configuration
+* `params` accesses all its parameters
+* `params['pid']` selects the `pid` entry from the global NginX configuration
+* `{ should cmp 'logs/nginx.pid' }` tests if the PID is set to `logs/nginx.pid` (via `cmp` matcher)
+
+
+## Matchers
+
+This InSpec audit resource has the following matchers:
+
+### http
+
+Retrieves all `http` entries in the configuration file.
+
+    nginx_conf.http
+    => nginx_conf /etc/nginx/nginx.conf, http entries
+
+It provides further access to all individual entries, servers, and locations.
+
+    nginx_conf.http.entries
+    => [nginx_conf /etc/nginx/nginx.conf, http entry ...]
+
+    nginx_conf.http.servers
+    => [nginx_conf /etc/nginx/nginx.conf, server entry ...]
+
+    nginx_conf.http.locations
+    => [nginx_conf /etc/nginx/nginx.conf, location entry ...]
+
+You can access each of these from the array and inspect it further (see below).
+
+### servers
+
+Retrieve all `servers` entries in the configuration:
+
+    # all servers across all configs aggregated:
+    nginx_conf.servers
+    => [nginx_conf /etc/nginx/nginx.conf, server entry ...]
+
+    # servers that belong to a specific http entry:
+    nginx_conf.http.entries[0].servers
+    => [nginx_conf /etc/nginx/nginx.conf, server entry ...]
+
+Servers provide access to all their locations, parent http entry, and raw parameters:
+
+    server = nginx_conf.servers[0]
+
+    server.locations
+    => [nginx_conf /etc/nginx/nginx.conf, location entry ...]
+
+    server.parent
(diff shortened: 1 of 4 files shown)


docs/resources/nginx_conf.md.erb
```diff
@@ -0,0 +1,122 @@
+---
+title: About the nginx_conf Resource
+---
+
+# nginx_conf
+
+Use the `nginx_conf` InSpec resource to test configuration data for the NginX server located at `/etc/nginx/nginx.conf` on Linux and Unix platforms.
```
adamleff: Style-wise, the nginx project refers to the program as `nginx` (all lowercase) or `NGINX` (all uppercase) but never mixed case. We should probably respect their style and use `nginx`

lib/resources/nginx_conf.rb
```diff
@@ -92,4 +101,92 @@ def resolve_references(data, rel_path)
...
+    end
+
+    def locations
+      servers.map(&:locations).flatten
+    end
+
+    def to_s
+      @parent.to_s + ', http entries'
+    end
+    alias inspect to_s
```
adamleff, line 123: Unless there's a real specific need to do this, I'd rather leave `inspect` alone as it can prove useful when debugging.

lib/resources/nginx_conf.rb
```diff
@@ -92,4 +101,92 @@ def resolve_references(data, rel_path)
       Hash[data.map { |k, v| [k, resolve_references(v, rel_path)] }]
     end
   end
+
+  class NginxConfHttp
+    attr_reader :entries
+    def initialize(params, parent)
+      @parent = parent
+      @entries = params.map { |x| NginxConfHttpEntry.new(x, parent) }
```
adamleff: If I initialize `nginx_conf` with a path to a file that doesn't exist or has no content, `params` will be nil and you'll try to call `#map` on NilClass, which will not be fun :)

lib/resources/nginx_conf.rb
```diff
@@ -92,4 +101,92 @@ def resolve_references(data, rel_path)
...
+    end
+
+    filter = FilterTable.create
+    filter.add_accessor(:where)
+          .add(:locations, field: 'location')
+          .connect(self, :location_table)
+
+    def to_s
+      # go two levels up: 1. to the http entry and 2. to the root nginx conf
+      @parent.parent.to_s + ', server entry'
```
adamleff: I think we can be even more transparent with users, especially when they're debugging / exploring in InSpec shell. Otherwise you get output like this:

```
inspec> a.http.servers.entries
=> [nginx_conf /Users/aleff/projects/inspec/test/unit/mock/files/nginx.conf, server entry,
 nginx_conf /Users/aleff/projects/inspec/test/unit/mock/files/nginx.conf, server entry]
```

Can we do something like this instead?

```
def to_s
  server_name = ''
  unless params['server_name'].flatten.nil?
    server_name = params['server_name'].flatten.first
    server_name += ":#{params['listen'].flatten} unless params['listen'].flatten.nil?
  end
  @parent.parent.to_s + ", #{server_name} server entry"
end
```

^^^ not perfectly tested, but hopefully you get the gist of what I'm trying to describe.

adamleff (changes requested):
This is great step in a better direction, @arlimus - thank you!

We need to do some better defense against a user passing in a path that is not readable or doesn't exist lest we try to call methods on `nil`

Also, I'm curious as to your use of FilterTable in the classes where the table really only has two columns: the key and the contents. While it's elegant for us as developers, it's not elegant for our users that are going to be calling `#flatten` and `#first` a lot to get at the actual content (because of the nested arrays). I wonder if there's a better way to handle this.

docs/resources/nginx_conf.md.erb
arlimus: great catch!

lib/resources/nginx_conf.rb
arlimus: good point, also happens when the file has non NGINX content! fixed for now by preventing params from being `nil`

lib/resources/nginx_conf.rb
arlimus: we should decide if we want the inspec shell to have the pretty-printed resources or their `inspect` equivalent. for example: `nginx_conf` right now in the inspec shell will pretty-print to `=> nginx_conf /etc/nginx/nginx.conf`. without this change it would print the entire page full of param debug. I aimed for consistency and for less printing, especially on `http`, but i also get the argument for the counterpoint (not to forget fewer changes).

lib/resources/nginx_conf.rb
arlimus: Yeah, totally! I wanted to leave it out of this PR to figure out pretty-printing for all 3 new nginx portions in a follow-up PR. Are you ok if we issue this in a follow-up?

adamleff (approved):

adamleff:
Thanks for the params-nil cleanup :)

lib/resources/nginx_conf.rb
arlimus: Added both of these to this PR, thank you for the discussion!

adamleff (approved):
Looks great! Thanks @arlimus - let's get this out there.

PR #2127, opened 2017-09-08
Add attributes file readability error handling

### LINKED ISSUE #1943: Provide better error if the attributes file doesn't exist
### Description

A generic error message about "unable to find a parser" is provided if the filename provided as the value to `--attrs` doesn't actually exist

1.28.1

### Replication Case

```
inspec exec ~/profiles/test1 --attrs foo.yaml
Unable to find a parser for attributes file foo.yaml. Check to make sure the file exists and has the appropriate extension.
```

### Possible Solutions

Before looking for a valid secrets plugin for the attributes file provided, we should check to see if the file even exists and is readable. If not, we should provide a better error message and exit.

jerryaldrichiii:
This fixes #1943

### DIFF
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -274,17 +274,19 @@ module Inspec
     end
 
     def validate_attributes_file_readability!(target)
-      return if File.readable?(target)
-
       unless File.exist?(target)
         raise Inspec::Exceptions::AttributesFileDoesNotExist,
               "Cannot find attributes file '#{target}'. " \
               'Check to make sure file exists.'
       end
 
-      raise Inspec::Exceptions::AttributesFileNotReadable,
-            "Cannot read attributes file '#{target}'. " \
-            'Check to make sure file is readable.'
+      unless File.readable?(target)
+        raise Inspec::Exceptions::AttributesFileNotReadable,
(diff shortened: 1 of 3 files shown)


lib/inspec/runner.rb
```diff
@@ -270,5 +273,19 @@ def register_rule(rule)
 
       examples.each { |e| @test_collector.add_test(e, rule) }
     end
+
+    def validate_attributes_file_readability!(target)
+      return if File.readable?(target)
```
adamleff: While this is technically correct, it's an awkward implementation because this line is 100% connected to the exception at L286. If there were 3+ different conditions we wanted to check for in order to provide a unique error message for each, then this might be fine.

Please remove this line and wrap the exception at L286 with an `unless` block like the one above it so it's 100% clear the condition of that exception.

adamleff (changes requested):
Thanks @jerryaldrichiii - this is pretty close. One code cleanup item needed before merge.

lib/inspec/runner.rb
jerryaldrichiii: Can do! I'll have it return `true` to avoid the rubocop `Style/GuardClause` offense.

adamleff (approved):
LGTM - thanks @jerryaldrichiii!

PR #2134, opened 2017-09-11
Ignore linked container names when parsing docker containers

### LINKED ISSUE #1855: docker_container: running matcher does not work by name with linked containers
### Description

If containers are linked the `running` test will fail for the linked container.

debian stretch
inspec 1.25 (from ChefDK 1.4.3)

### Replication Case

`docker run -d --rm --name c1 centos:7 yes`
`docker run -d --rm --name c2 --link c1:container1 centos:7 yes`

```
describe docker_container('c1') do                                       
  it { should be_running }                                                         
end
```

### Possible Solutions

The test fails because in `lib/resources/docker_container.rb` line 94 the method `container_info` filters by name doing an exact match. However, as the container status is retrieved by doing `docker ps -a --no-trunc` the actual name for container `c1` in the replication case would be `c1,c2/container1`, that is, the concatenation of all names.

In Serverspec this works because the container status is retrieved individually by name using `docker inspect`.

adamleff:
If a container is linked to another container, the normal `docker ps` output does not include this information. However, when pulling the `.Names` field with `docker ps --format`, the linked container is listed in the name. This is confusing for users trying to use InSpec to audit a container.

This change strips any linked container names from the actual container name.

Fixes #1855


chris-rock (approved):
Thank you @adamleff

PR #2141, opened 2017-09-13
Add wildcard/multiple server support to nginx_conf resource

### LINKED ISSUE #1998: nginx_conf to support include by wildcard
### Description

i started a nginx docker container with docker run -it nginx bash and then i connected with inspec shell -t docker://<containerID> to the container. But i do not get a result with nginx_conf.params

```
atomic111:..tuff/chef_cookbooks ∅> inspec shell -t docker://elastic_heyrovsky
      thor (~> 0.19)
      rspec (~> 3)
      addressable (~> 2.4)
      winrm (~> 2.0)
      docker-api (~> 1.26)

You are currently running on:

    OS platform:  debian
    OS family:  debian
    OS release: 9.0

inspec> nginx_conf.params
=> {}
```

the nginx conf looks like:
```
root@c7c90079b007:/# cat /etc/nginx/nginx.conf 

user  nginx;
worker_processes  1;

error_log  /var/log/nginx/error.log warn;
pid        /var/run/nginx.pid;


events {
    worker_connections  1024;
}


http {
    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log  /var/log/nginx/access.log  main;

    sendfile        on;
    #tcp_nopush     on;

    keepalive_timeout  65;

    #gzip  on;

    include /etc/nginx/conf.d/*.conf;
}
```

If i do this on my arch linux host system, the nginx_conf resource is working. But it includes not the other configs, which are include in the config. see the chef-nginx-hardening. There is the 90-hardening.conf included and inspec does not read it.

inspec version is 1.30.0

I use an arch linux host system and startet the nginx docker container from hub.docker.com

### Replication Case

see above

### Possible Solutions

sorry, no plan :(

### LINKED ISSUE #2035: nginx_conf resource does not handle multiple virtual server definitions
### Description

nginx_conf resource does not parse and expose multiple virtual server definitions in the included conf files. It parses just one and adds it to params

OS: Cent 7

### Replication Case
nginx -T output (shows all included server defs)
image

nginx_conf.params output inspec shell ( includes only one of the servers)
image

### Possible Solutions
update nginx_conf parser to collect all the included servers from the conf files

jerryaldrichiii:
Fixes #2035 
Fixes #1998

### DIFF
--- a/lib/resources/nginx_conf.rb
+++ b/lib/resources/nginx_conf.rb
@@ -115,12 +115,16 @@ module Inspec::Resources
       return if data.nil? || conf.nil?
       # Step through all conf items and create combined return value
       data.merge!(conf) do |_, v1, v2|
-        # If both the data field and the conf field are arrays, then combine them
-        next v1 + v2 if v1.is_a?(Array) && v2.is_a?(Array)
-        # If both the data field and the conf field are maps, then deep merge them
-        next merge_config!(v1, v2) if v1.is_a?(Hash) && v2.is_a?(Hash)
-        # All other cases, just use the new value (regular merge behavior)
-        v2
+        if v1.is_a?(Array) && v2.is_a?(Array)
+          # If both the data field and the conf field are arrays, then combine them
+          v1 + v2
+        elsif v1.is_a?(Hash) && v2.is_a?(Hash)
+          # If both the data field and the conf field are maps, then deep merge them
+          merge_config!(v1, v2)
+        else
+          # All other cases, just use the new value (regular merge behavior)
+          v2
+        end
       end
     end
   end
--- a/lib/resources/nginx_conf.rb
+++ b/lib/resources/nginx_conf.rb
@@ -3,6 +3,7 @@
 # author: Christoph Hartmann
 
 require 'utils/nginx_parser'
+require 'utils/find_files'
 require 'forwardable'
 
 # STABILITY: Experimental
@@ -24,6 +25,8 @@ module Inspec::Resources
     "
 
     extend Forwardable
+
+    include FindFiles
 
     attr_reader :contents
 
@@ -93,8 +96,9 @@ module Inspec::Resources
       if data.key?('include')
         data.delete('include').flatten
             .map { |x| File.expand_path(x, rel_path) }
+            .map { |x| find_files(x) }.flatten
             .map { |path| parse_nginx(path) }
-            .map { |e| data.merge!(e) }
+            .map { |e| data.merge!(e) { |_, v1, v2| v1 + v2 } }
       end
 
       # Walk through the remaining hash fields to find more references
--- a/lib/resources/nginx_conf.rb
+++ b/lib/resources/nginx_conf.rb
@@ -105,7 +105,11 @@ module Inspec::Resources
       Hash[data.map { |k, v| [k, resolve_references(v, rel_path)] }]
     end
 
-    # Merge conf into data without overwriting data.
+    # Deep merge fields from NginxConfig.parse.
+    # A regular merge would overwrite values so a deep merge is needed.
+    # @param data [Hash] data structure from NginxConfig.parse
+    # @param conf [Hash] data structure to be deep merged into data
+    # @return [Hash] data structure with conf and data deep merged
     def merge_config!(data, conf)
       # Catch edge-cases
       return if data.nil? || conf.nil?
@@ -113,9 +117,9 @@ module Inspec::Resources
       data.merge!(conf) do |_, v1, v2|
         # If both the data field and the conf field are arrays, then combine them
         next v1 + v2 if v1.is_a?(Array) && v2.is_a?(Array)
-        # If both the data field and the conf field are maps, then merge them
+        # If both the data field and the conf field are maps, then deep merge them
         next merge_config!(v1, v2) if v1.is_a?(Hash) && v2.is_a?(Hash)
-        # All other cases, just use the new value
+        # All other cases, just use the new value (regular merge behavior)
         v2
       end
     end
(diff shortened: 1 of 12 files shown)


lib/resources/nginx_conf.rb
```diff
@@ -93,8 +96,9 @@ def resolve_references(data, rel_path)
       if data.key?('include')
         data.delete('include').flatten
             .map { |x| File.expand_path(x, rel_path) }
+            .map { |x| find_files(x) }.flatten
             .map { |path| parse_nginx(path) }
-            .map { |e| data.merge!(e) }
+            .map { |e| data.merge!(e) { |_, v1, v2| v1 + v2 } }
```
jerryaldrichiii: I feel like this is wrong...just not sure how to make it better. When loading server configs from other files (e.g. `/etc/nginx/conf.d/*.conf`) the `merge!` would clobber the `server` value and replace it. This appends the list to the key...but that assumes that `v1` and `v2` will always be lists...I don't like that assumption.

test/unit/mock/cmd/find-nginx-confd-example-conf
```diff
@@ -0,0 +1 @@
+/etc/nginx/conf.d/example.conf
```
adamleff: For completeness of testing, I'd change this to return multiple files to ensure the code actually iterates over each of them.

lib/resources/nginx_conf.rb
arlimus: should probably be an `each`; also let's extract this piece as it feels to cover some complexity (for deep merging)

adamleff: By "lists" to you mean `Array`s?  Why wouldn't they be? If that's a concern, how do we protect against it?

We own the entire codepath here, so if there's additional promises that need to be made, we can make them.

adamleff (changes requested):
This looks like a great first start.

I left a note about adding additional test coverage by having it find multiple files. We should make sure when it counters multiple files that it properly iterates of each.

We also need to expand the test coverage to ensure that data from each file is present. So if your sample files include multiple servers, we should have a test that is able to pluck unique data out of each server to ensure we're properly including all data and not accidentally clobbering another server record. That should help raise the confidence level of this change.

lib/resources/nginx_conf.rb
```diff
@@ -93,13 +96,29 @@ def resolve_references(data, rel_path)
...
             .map { |path| parse_nginx(path) }
-            .map { |e| data.merge!(e) }
+            .each { |conf| merge_config!(data, conf) }
       end
 
       # Walk through the remaining hash fields to find more references
       Hash[data.map { |k, v| [k, resolve_references(v, rel_path)] }]
     end
+
+    # Merge conf into data list of parameters.
```
arlimus: Needs better docs string 😁

arlimus (commented):
Agree with Adam's multi-file test case.
Otherwise this looks fantastic, thank you so much!!

jerryaldrichiii:
@adamleff @arlimus:

Feedback addressed! Let me know what you think.

arlimus (approved):
Thank you for the great updates!!

lib/resources/nginx_conf.rb
```diff
@@ -93,13 +96,33 @@ def resolve_references(data, rel_path)
...
+    # @param data [Hash] data structure from NginxConfig.parse
+    # @param conf [Hash] data structure to be deep merged into data
+    # @return [Hash] data structure with conf and data deep merged
+    def merge_config!(data, conf)
+      # Catch edge-cases
+      return if data.nil? || conf.nil?
+      # Step through all conf items and create combined return value
+      data.merge!(conf) do |_, v1, v2|
+        # If both the data field and the conf field are arrays, then combine them
+        next v1 + v2 if v1.is_a?(Array) && v2.is_a?(Array)
```
adamleff: Maybe it's just me, but this use of `next` feels pretty awkward. Any chance I can convince you guys to make this an `if ... elsif ... else` instead? I think it would read so much cleaner.

adamleff:
I'd really love to see some tests actually check the data we've parsed, not just the count of objects we expect. It can be done in a later PR though.

lib/resources/nginx_conf.rb
```diff
@@ -93,13 +96,37 @@ def resolve_references(data, rel_path)
       if data.key?('include')
         data.delete('include').flatten
             .map { |x| File.expand_path(x, rel_path) }
+            .map { |x| find_files(x) }.flatten
```
rx294, line 99: Thank you for the fix Jerry. 
However this solution does not always include files (included with '*') because of issue#2135 (find_files uses the command resource).
I just tested this code over an ssh connection and it does not parse all the included conf files;
It works fine when I tested by changing the find_files code to use bash resource instead of command resource.
Hope you can find a solution for it.

Thank you again Jerry,
Rony

adamleff: I know we have a logged issue about the command resource specifically, but we should log an issue for the `find_files` method too so we don't lose that context, and link it to this PR. I'm not going to hold off merging this PR since it's still a step in the right direction and will work fine for users not executing via sudo.

jerryaldrichiii: https://giphy.com/gifs/reaction-LSmULmByAQHQs/fullscreen

rx294: Haha sorry @jerryaldrichiii . 
Thank you @adamleff I have posted and issue for find_files. issue #2157

PR #2144, opened 2017-09-13
Build and tag docker image via Expeditor

adamleff:
In order to provide Docker images of all unstable, current, and stable builds of InSpec, and to avoid having to manually publish Docker images each time we release InSpec, Expeditor will now take care of this for us.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM ruby:alpine
 MAINTAINER Chef Software, Inc. <docker@chef.io>
 
-ARG CHANNEL=stable
 ARG VERSION=1.36.1
 ARG GEM_SOURCE=https://rubygems.org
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,13 @@
 FROM ruby:alpine
+MAINTAINER Chef Software, Inc. <docker@chef.io>
+
+ARG CHANNEL=stable
```
tduffield: If we are not using the CHANNEL arg, I don't think we need to keep it.

tduffield (approved):

Dockerfile
adamleff: Totally right - leftover from my testing of trying to just use the omnibus stuff. I'll kill it.

PR #2152, opened 2017-09-14
add example for checking last permissions octet

thomascate:
Recently had trouble figuring out how to check the last octet permissions aka everyone/everybody/all users perms. Added example to docs.

### DIFF
--- a/docs/resources/file.md.erb
+++ b/docs/resources/file.md.erb
@@ -61,6 +61,10 @@ an owner:
 
     it { should be_executable.by('owner') }
 
+any user:
+
+    it { should be_executable.by('others') }
+
 a user:
 
     it { should be_executable.by_user('user') }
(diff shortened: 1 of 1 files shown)


docs/resources/file.md.erb
```diff
@@ -61,6 +61,10 @@ an owner:
 
     it { should be_executable.by('owner') }
 
+any user:
```
adamleff: Can we change this (and the others) to "any user other than the owner or members of the file's group"? "any user" is actually incorrect. If the file is mode `004`, the owner and members of the group still won't be able to read it even though the "others" have read bit enabled.

Suggestions for a shorter string than what I provided are also welcome. :)

adamleff (changes requested):

adamleff:
Also @thomascate can you please sign-off your commit to satisfy the requirements of the [DCO](https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco)? Lemme know if you need help with that.

docs/resources/file.md.erb
thomascate: Good catch, couldn't come up with a shorter string so I used yours. :)

adamleff (approved):
Thanks @thomascate!

PR #2156, opened 2017-09-15
Add deprecation warning to auditd_rules resource

### LINKED ISSUE #2145: Add deprecation warnings to the existing auditd_rules resource
A new `auditd` resource has been created in #2133. As a result, we need to add deprecation warnings now so we can remove it in InSpec 2.0.

adamleff:
The auditd_rules resource has been replaced by the auditd resource. We are planning on removing the auditd_rules resource in InSpec 2.0. This change will provide a warning to any user using the old resource.

Resolves #2145

### DIFF
--- a/lib/resources/auditd_rules.rb
+++ b/lib/resources/auditd_rules.rb
@@ -87,6 +87,8 @@ module Inspec::Resources
         parse_content
         @legacy = nil
       end
+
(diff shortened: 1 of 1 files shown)


arlimus (approved):
Good point, thank you!

PR #2159, opened 2017-09-18
Add wildcard support to `Utils::FindFiles`

### LINKED ISSUE #2157: find_files does not handle wildcard '*' path
### Description

find_files does not capture files when a path with with wildcard '*' is provided.
This issue shows up when the backend is 'ssh' and sudo is active.

This is due to issue #2135 with the 'command' resource which find_files uses to run the commands.

jerryaldrichiii:
Wildcards are evaluated prior to applying `sudo` permissions. This means that running `sudo find /some/path/*.conf` will fail if the user does not have read permissions on `/some/path/` because the wildcard cannot expand before `sudo` is applied and `*.conf` isn't a file.

The solution for this is to run the command in a subshell that has the proper permissions (e.g. `sudo sh -c 'find /some/path/*.conf'`).  

This modifies `Utils::FindFiles` to use a subshell thus allowing wildcard support.

This fixes #2157


adamleff (approved):
Thanks for address this, @jerryaldrichiii!

arlimus (approved):
Great to have this fixed, thank you @jerryaldrichiii

PR #2163, opened 2017-09-20
Modify Upstart enabled check to use config file

### LINKED ISSUE #1834: Upstart is_enabled check on Trusty doesn't take into account the Chef's workaround
### Description

Chef's `Upstart` service provider disables a service by [commenting out the line](https://github.com/chef/chef/blob/a2890be2422453dee593ffb1f5eb6a49271c83be/lib/chef/provider/service/upstart.rb#L234-L239) of `start on` in the upstart file.

InSpec, on the other hand, uses `initctl show-config` to check if a service is enabled if the version of `initctl` is greater than 1.3.

The problem is that `initctl` on Trusty doesn't take into account the Chef's hack:

```console
$ cat /etc/init/my-service.conf | grep start
#start on runlevel [2345]
$ initctl show-config my-service
my-service
  start on runlevel [2345]
  stop on runlevel [!2345]
``` 

This means if I check the service in InSpec like this, it will fail:

```ruby
descrive upstart_service('my-service') do
  it { should_not be_enabled }
end
```

I'm using the InSpec from Chef DK:

* Chef Development Kit Version: 1.3.43
* InSpec 1.19.1
* Ubuntu 14.04.1 LTS
* initctl (upstart 1.12.1)

### Replication Case

I don't have a concrete minimal test case but it should be replicated like this:

Use Chef to disable an upstart service:
```ruby
service 'my-service' do
  provider Chef::Provider::Service::Upstart
  action :disable
end
```
then check it with InSpec:
```ruby
descrive upstart_service('my-service') do
  it { should_not be_enabled }
end
```
This fails.

### Possible Solutions

I see a few possible solutions:

* Add an additional check in [info_enabled](https://github.com/chef/inspec/blob/6c53188e8566c2785fd3da7a2e6940580860c1b6/lib/resources/service.rb#L357) to see if `start on` is commented out.
* Do not use `initctl`

### Stacktrace

I don't have a stacktrace.

jerryaldrichiii:
This modifies the enabled check for the `service` resource to use the service's config file instead of `initctl show-config`.

`initctl show-config` does not accurately show the state of a service if that service's config file is modified while the service is running.

This fixes #1834.

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -349,40 +349,21 @@ module Inspec::Resources
         description: nil,
         installed: true,
         running: running,
-        enabled: info_enabled(status, service_name),
+        enabled: info_enabled(service_name),
         type: 'upstart',
       }
     end
 
     private
 
-    def info_enabled(status, service_name)
+    def info_enabled(service_name)
       # check if a service is enabled
-      # http://upstart.ubuntu.com/cookbook/#determine-if-a-job-is-disabled
-      # $ initctl show-config $job | grep -q "^  start on" && echo enabled || echo disabled
-      # Ubuntu 10.04 show-config is not supported
-      # @see http://manpages.ubuntu.com/manpages/maverick/man8/initctl.8.html
-      support_for_show_config = Gem::Version.new('1.3')
-
-      if version >= support_for_show_config
-        config = inspec.command("#{service_ctl} show-config #{service_name}").stdout
-      else # use config file as fallback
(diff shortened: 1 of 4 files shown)


adamleff (approved):
Looks good to me! I wish `initctl` was smarter, but relying on the config files is the right thing to do. Thanks, @jerryaldrichiii!

arlimus (approved):
Makes a lot of sense, thank you for this @jerryaldrichiii !!

PR #2168, opened 2017-09-20
Support `false` for attribute value

### LINKED ISSUE #2164: Can't pass "false" as an attribute value
Assuming the following control file:

```ruby
true_test = attribute('true_test', description: 'foo')
false_test = attribute('false_test', description: 'foo')

puts "true_test class: #{true_test.class}"
puts "false_test class: #{false_test.class}"
```

... and the following attributes file:

```yaml
---
true_test: true
false_test: false
```

... you get the following results from `inspec exec`:

```
true_test class: TrueClass
false_test class: NilClass
```

Something is swallowing `false` values and needs to be fixed.

Originally logged in chef/kitchen-inspec#138

adamleff:
The logic in `Inspec::Attribute` prohibited the use of `false` (FalseClass) as a valid attribute. If the attribute value supplied was `false`, then it would fall back to the default value.

This change properly allows the use of `false` as a value, adds the initial tests for Inspec::Attribute, and also uses better attr_writer semantics for writing/storing the value.

Fixes #2164

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -3,6 +3,8 @@
 module Inspec
   class Attribute
     attr_accessor :name
+    attr_writer :value
+
     def initialize(name, options)
       @name = name
       @opts = options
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Sweet catch @adamleff thank you!!

PR #2170, opened 2017-09-21
Update method in which Pry hooks are removed

adamleff:
Pry 0.11 removed the clear_all method for removing all hooks. This change updates the way we clear hooks for the events we care about.


arlimus:
Thank you Adam!!

PR #2177, opened 2017-09-22
forgiving default attributes

arlimus:
When default attributes arent specified provide one that is much more forgiving.
See this https://github.com/chef/inspec/issues/2176

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -5,7 +5,17 @@ module Inspec
     attr_accessor :name
     attr_writer :value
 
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Looks good to me. Thank you @adamleff and @arlimus

adamleff (approved):
Would I ever do this in another Ruby project on purpose? Nope. Do I think it's a great idea to provide a great user experience for our project? You bet!

Thanks, @arlimus!

arlimus:
ahahahaha thank you @adamleff

PR #2179, opened 2017-09-22
Properly return postgres query errors on failure

adamleff:
When using the `query` method in the `postgres_session` resource, if the query fails, the `query` method attempts to call `skip_resource` with an error message. Not only does the `skip_resource` not properly work, but it also returns a `String` object back to the test which is probably going to try and call the `output` method on it to run the test.

This results in an error like this:

```
  Can't read
     ∅  undefined method `output' for "output":String
```

This change returns the full psql output as a Lines object to the user, including stderr, so they can at least get the error in their test output and avoids undefined method errors.


arlimus (approved):
Thank you @adamleff for fixing this 😁

PR #2184, opened 2017-09-25
Update postgres.rb

rndmh3ro:


aaronlippold:
Thanks for the catch. My bad.

adamleff:
Hi, @rndmh3ro - and thanks for your first PR to InSpec!

Before we can accept your change, we need your commits to be "signed off" according to the [Developer Certificate of Origin](https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#dco-sign-off-methods) (or "DCO"). Basically, your commit message either needs to have a DCO "Signed-off-by" line, or it needs to say "obvious fix" in the message. Either are fine for this change.

Could you please amend your change to include either of these so I can legally accept your change?

Thank you!
~Adam

adamleff:
There has not been any movement on this PR, so unfortunately I have to close it without a sign-off. If someone else would like to submit this change as an obvious fix or with proper DCO sign-off, we'll be happy to accept it. thanks!

PR #2202, opened 2017-10-02
Add output for port/protocol for host resource.

### LINKED ISSUE #2085: describe host(...) output
### Description

Is there any way to control output of inspec `host` resource to specify the port and protocol when it outputs results?

For example, I've coded:
```
  describe host("srv", port: 55001, protocol: 'tcp') do
    it { should be_reachable }
  end
```
and the output I get is:
```
Host srv
  should be reachable
```
where, in fact, I'd like to know which port connection is being checked, as in:
```
Host srv/55001/tcp
  should be reachable
```
1.32.1

jquick:
This fixes #2085. Port and protocol are now shown in output of the host resource if defined.

issue-2085

### DIFF
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -113,7 +113,11 @@ module Inspec::Resources
     end
 
     def to_s
-      "Host #{hostname}"
+      if port
+        "Host #{hostname} port #{port} proto #{protocol}"
+      else
+        "Host #{hostname}"
+      end
     end
 
     private
--- a/lib/resources/host.rb
+++ b/lib/resources/host.rb
@@ -113,11 +113,8 @@ module Inspec::Resources
     end
 
     def to_s
-      if port
-        "Host #{hostname} port #{port} proto #{protocol}"
-      else
-        "Host #{hostname}"
-      end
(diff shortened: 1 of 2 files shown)


lib/resources/host.rb
```diff
@@ -113,6 +113,7 @@ def ipaddress
     end
 
     def to_s
+      return "Host #{hostname} port #{port} proto #{protocol}" if port
```
adamleff: I think this is fine. I'd consider "building" the string instead of the early `return` so we're not duplicating the "Host hostname" bit... something like this:

```ruby
def to_s
  resource_name = "Host #{hostname}"
  resource_name += " port #{port} proto #{protocol}" if port

  resource_name
end
```

adamleff (approved):
This looks good to me, @jquick! My suggestion isn't a blocker but would avoid an early return. Thanks for the fix!

lib/resources/host.rb
jquick: Sounds good, just pushed out a update with the changes. Thanks for the help!

adamleff (approved):
Looks great, and tests are all green!

arlimus (approved):
Thank you @jquick !! 😁

PR #2203, opened 2017-10-02
Add Chef Automate support to `inspec compliance login`

### LINKED ISSUE #1795: Add Chef Automate support to `inspec compliance login`
The goal is to support Automate logins via `inspec compliance login` without requiring the existing `login_automate`.

Right now we have to different login methods for Chef Compliance and Chef Automate. We should harmonize those:

- `inspec compliance login_automate https://automate.compliance.test --insecure --user='admin' --ent='brewinc' --usertoken='%TOKEN%`
- 'inspec compliance login https://compliance.test --insecure --user='admin' --token='%TOKEN%'`

As you can see from the examples, only `token` and `ent` argument are different. We should use `--token` as the default and make `--ent` an optional argument. We may need a `--type` to indicate if this is a Chef Compliance or Automate server.

This issue should:
 - add Automate support to the `login` command (using the `--token` and `--ent` flag as required)
 - continue to support the `--usertoken` flag but output a deprecation warning
 - add a deprecation warning to the `login_automate` command

The removal of the `login_automate` command will come via a later change via #2154 in the next appropriate major release.

jerryaldrichiii:
This provides a single interface for logging into either Chef Automate or Chef Compliance servers. Server type is evaluated at run time via HTTP responses from designated endpoints.

This also moves the login logic from `Compliance::ComplianceCLI` to a separate set of modules in `Compliance::API`. Thus removing logic from Thor and allowing for more in depth Unit testing.

This also adds deprecation warnings for `login_automate` and `--usertoken`.

This fixes #1795.

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -9,7 +9,6 @@ require_relative 'api/login'
 
 module Compliance
   class ServerConfigurationMissing < StandardError; end
-  class CannotDetermineServerType < StandardError; end
 
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
@@ -244,9 +243,11 @@ module Compliance
     end
 
     def self.determine_server_type(url, insecure)
-      return :automate if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
-      return :compliance if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
-      raise CannotDetermineServerType, "Unable to determine if #{url} is a Chef Automate or Chef Compliance server"
+      if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
+        :automate
+      elsif Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
+        :compliance
+      end
     end
   end
 end
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -14,7 +14,6 @@ module Compliance
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
-
     extend Login
 
     # return all compliance profiles available for the user
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -5,13 +5,18 @@
 require 'net/http'
 require 'uri'
 
+require_relative 'api/login'
+
 module Compliance
-  class ServerConfigurationMissing < StandardError
-  end
+  class ServerConfigurationMissing < StandardError; end
+  class CannotDetermineServerType < StandardError; end
 
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
+
+    extend Login
+
     # return all compliance profiles available for the user
     def self.profiles(config)
       # Chef Compliance
@@ -237,6 +242,12 @@ module Compliance
       return nil unless config.key?('version')
       return nil unless config['version'].is_a?(Hash)
       config['version']['version']
+    end
+
+    def self.determine_server_type(url, insecure)
+      return 'automate' if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
+      return 'compliance' if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
+      raise CannotDetermineServerType
     end
   end
 end
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -244,8 +244,8 @@ module Compliance
     end
 
     def self.determine_server_type(url, insecure)
-      return 'automate' if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
-      return 'compliance' if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
+      return :automate if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
+      return :compliance if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
       raise CannotDetermineServerType, "Unable to determine if #{url} is a Chef Automate or Chef Compliance server"
     end
   end
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -14,7 +14,7 @@ module Compliance
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
-    extend Login
+    extend Compliance::API::Login
 
     # return all compliance profiles available for the user
     def self.profiles(config)
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -246,7 +246,7 @@ module Compliance
     def self.determine_server_type(url, insecure)
       return 'automate' if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
       return 'compliance' if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
-      raise CannotDetermineServerType
+      raise CannotDetermineServerType, "Unable to determine if #{url} is a Chef Automate or Chef Compliance server"
     end
   end
 end
--- a/lib/bundles/inspec-compliance/api/login.rb
+++ b/lib/bundles/inspec-compliance/api/login.rb
@@ -3,135 +3,143 @@
 # author: Dominik Ricter
 # author: Jerry Aldrich
 
-module Login
-  def login(options)
-    options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure'])
+module Compliance
+  class API
+    module Login
+      def login(options)
+        options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure'])
 
-    return Login::ComplianceServer.login(options) if options['server_type'] == 'compliance'
-    verify_thor_options(options)
+        return Login::ComplianceServer.login(options) if options['server_type'] == 'compliance'
+        Login::AutomateServer.login(options)
+      end
 
-    options['url'] = options['server'] + '/compliance'
-    token = options['dctoken'] || options['token']
-    puts store_access_token(options, token)
-  end
+      module AutomateServer
+        def self.login(options)
+          verify_thor_options(options)
 
-  private
+          options['url'] = options['server'] + '/compliance'
+          token = options['dctoken'] || options['token']
+          puts store_access_token(options, token)
+        end
 
-  def store_access_token(options, token)
-    token_info = if options['token']
-                   { type: 'usertoken', msg: 'automate user token' }
-                 else
-                   { type: 'dctoken', msg: 'data colletor token' }
-                 end
+        def self.store_access_token(options, token)
+          token_info = if options['token']
+                         { type: 'usertoken', msg: 'automate user token' }
+                       else
+                         { type: 'dctoken', msg: 'data colletor token' }
+                       end
 
-    config = Compliance::Configuration.new
+          config = Compliance::Configuration.new
 
-    config.clean
+          config.clean
 
-    config['automate'] = {}
-    config['automate']['ent'] = options['ent']
-    config['automate']['token_type'] = token_info[:type]
-    config['server'] = options['url']
-    config['user'] = options['user']
-    config['insecure'] = options['insecure'] || false
-    config['server_type'] = options['server_type']
+          config['automate'] = {}
+          config['automate']['ent'] = options['ent']
+          config['automate']['token_type'] = token_info[:type]
+          config['server'] = options['url']
+          config['user'] = options['user']
+          config['insecure'] = options['insecure'] || false
+          config['server_type'] = options['server_type']
 
-    config['token'] = token
-    config['version'] = Compliance::API.version(config)
-    config.store
-    "Stored configuration for Chef Automate: '#{config['server']}' with user: '#{config['user']}', ent: '#{config['automate']['ent']}' and your #{token_info[:msg]}"
-  end
+          config['token'] = token
+          config['version'] = Compliance::API.version(config)
+          config.store
+          "Stored configuration for Chef Automate: '#{config['server']}' with user: '#{config['user']}', ent: '#{config['automate']['ent']}' and your #{token_info[:msg]}"
+        end
 
-  # Automate login requires `--ent`, `--user`, and either `--token` or `--dctoken`
-  def verify_thor_options(o)
-    error_msg = []
+        # Automate login requires `--ent`, `--user`, and either `--token` or `--dctoken`
+        def self.verify_thor_options(o)
+          error_msg = []
 
-    error_msg.push('Please specify a server using `inspec compliance login https://SERVER`') if o['server'].nil?
-    error_msg.push('Please specify a user using `--user=\'USER\'`') if o['user'].nil?
-    error_msg.push('Please specify an enterprise using `--ent=\'automate\'`') if o['ent'].nil?
+          error_msg.push('Please specify a server using `inspec compliance login https://SERVER`') if o['server'].nil?
+          error_msg.push('Please specify a user using `--user=\'USER\'`') if o['user'].nil?
+          error_msg.push('Please specify an enterprise using `--ent=\'automate\'`') if o['ent'].nil?
 
-    if o['token'].nil? && o['dctoken'].nil?
-      error_msg.push('Please specify a token using `--token=\'AUTOMATE_TOKEN\'` or `--dctoken=\'DATA_COLLECTOR_TOKEN\'`')
+          if o['token'].nil? && o['dctoken'].nil?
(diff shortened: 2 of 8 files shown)


lib/bundles/inspec-compliance/api.rb
```diff
@@ -5,13 +5,18 @@
...
 module Compliance
-  class ServerConfigurationMissing < StandardError
-  end
+  class ServerConfigurationMissing < StandardError; end
+  class CannotDetermineServerType < StandardError; end
 
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
+
```
adamleff: nit: extra newline at class beginning

lib/bundles/inspec-compliance/api.rb
```diff
@@ -238,5 +243,11 @@ def self.server_version_from_config(config)
       return nil unless config['version'].is_a?(Hash)
       config['version']['version']
     end
+
+    def self.determine_server_type(url, insecure)
+      return 'automate' if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
```
adamleff: It's more idiomatic to return symbols for something like this rather than strings.

`return :automate if ...`

lib/bundles/inspec-compliance/api.rb
```diff
@@ -238,5 +243,11 @@ def self.server_version_from_config(config)
...
+      return 'compliance' if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
+      raise CannotDetermineServerType
```
adamleff: You should also add an exception message when raising.

`raise CannotDetermineServerType, "Unable to determine if #{url} is a Compliance or Automate server"`

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,134 @@
...
+module Login
+  def login(options)
+    options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure'])
+
+    return Login::ComplianceServer.login(options) if options['server_type'] == 'compliance'
+    verify_thor_options(options)
+
+    options['url'] = options['server'] + '/compliance'
+    token = options['dctoken'] || options['token']
+    puts store_access_token(options, token)
```
adamleff: I know it's a WIP, to just a reminder to remove this :)

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,134 @@
+# encodeing: utf-8
+# author: Christoph Hartmann
+# author: Dominik Ricter
+# author: Jerry Aldrich
+
+module Login
+  def login(options)
+    options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure'])
+
+    return Login::ComplianceServer.login(options) if options['server_type'] == 'compliance'
```
adamleff: Will you be doing the Automate login here too? Probably worth Zooming to review the expected flow, etc. would be good when you're ready :) This mid-method return with no comments as to why is confusing me as to your intentions for this method and how Automate login will be handled.

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    options['url'] = options['server'] + '/compliance'
+    token = options['dctoken'] || options['token']
+    puts store_access_token(options, token)
+  end
+
+  private
+
+  def store_access_token(options, token)
+    token_info = { type: 'dctoken', msg: 'data colletor token' } if options['dctoken']
```
adamleff: Will you ever not have either a dctoken or a token? What happens if token_info is not set?

I'd recommend changing it to something like:

```ruby
token_info = if options['dctoken']
               { # dctoken stuff }
             else
               { # non-dctoken stuff }
             end
```

adamleff (commented):
Just flipped through this quickly, some comments and questions, probably worth a Zoom review when it's no longer WIP.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Pick all the nits! Fixed.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Agreed. Fixed.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: This redirects to the `login` method of the `ComplianceServer` module if the server type is a Compliance server. I was returning here, since this will complete the login. Initially I had it just `exit 0` at the end, but that was causing me issues with the Unit tests.

Will investigate.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: `store_access_token` returns a string at this time. I will look at the flow and see if we should pass this up as a return from `login`.

Will investigate.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: I agree. Fixed.

jerryaldrichiii:
Agreed, a Zoom review would be helpful! Thanks for the initial feedback @adamleff.

clintoncwolfe:
We should also update the README docs to reflect that the `inspec compliance login_automate` command is no longer present.  See the [README in the bundle](https://github.com/chef/inspec/blob/ja/merge-login-commands/lib/bundles/inspec-compliance/README.md).  You've already removed it from the Thor-based CLI help, awesome!

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -18,78 +18,32 @@ def self.subcommand_prefix
...
-      desc: 'Automate user token'
-    option :user, type: :string,
-      desc: 'Automate username'
-    option :ent, type: :string,
+    option :ent, type: :string, required: false,
       desc: 'Enterprise for Chef Automate reporting'
-    option :insecure, aliases: :k, type: :boolean,
-      desc: 'Explicitly allows InSpec to perform "insecure" SSL connections and transfers'
-    def login_automate(server) # rubocop:disable Metrics/AbcSize
+    def login(server) # rubocop:disable Metrics/AbcSize
```
clintoncwolfe: You don't need the rubocop disable flag here anymore (when you refactored this out, the method became much simpler).    You can run a rubocop scan with 
```bundle exec rake lint```

File: lib/bundles/inspec-compliance/api.rb [hunk 8]
```diff
@@ -5,13 +5,17 @@
...
 module Compliance
-  class ServerConfigurationMissing < StandardError
-  end
+  class ServerConfigurationMissing < StandardError; end
+  class CannotDetermineServerType < StandardError; end
 
   # API Implementation does not hold any state by itself,
   # everything will be stored in local Configuration store
   class API # rubocop:disable Metrics/ClassLength
+    extend Login
```
clintoncwolfe: This may be a matter of taste, but I'd usually have the module name here more deeply namespaced - like Compliance::API::Login .  I don't know that we'd ever (within InSpec) have a name collision with a module named Login, but being specific avoids that possibility.

lib/bundles/inspec-compliance/api/login.rb
clintoncwolfe: I'd second @adamleff here - If you're switching between two sets of logic, and you've encapsulated one in a module, I'd expect the other logic - for Automate login - to also be encapsulated in a module (presumably Login::AutomateServer).

lib/bundles/inspec-compliance/cli.rb
jerryaldrichiii: Great find! Fixed.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: Originally I thought only to encapsulate the Compliance server stuff because it would be deprecated in the future leaving only `Login`...but the future is not now. I see your point.

Fixed.

jerryaldrichiii:
@clintoncwolfe I was certain I did a `grep -R login_automate` at one point and removed all references! Must have stashed and forgot about it somewhere.

Great catch! README.md in bundle is updated. I also removed references to `--usertoken` in all the places.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Agreed! Fixed.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: @adamleff let me know what you think of the flow now that I've modified the namespacing.

lib/bundles/inspec-compliance/cli.rb
jerryaldrichiii: Agreed, nice catch! Fixed.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: I always prefer going the idiomatic route. Fixed. Thanks @adamleff!

jerryaldrichiii:
Changing this PR to target `master`. This will not remove support for `login_automate` or `--usertoken`. Instead it will provide a deprecation warning. We will remove both in InSpec 2.0.

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,145 @@
+# encoding: utf-8
+# author: Christoph Hartmann
+# author: Dominik Ricter
+# author: Jerry Aldrich
+
+module Compliance
+  class API
+    module Login
+      def login(options)
+        options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure']).to_s
```
adamleff: Coercing to a string here seems... weird? We don't need it to be a string here or most of these methods -- only when we store the config to disk. Can we move the coercion there?

lib/bundles/inspec-compliance/api.rb
```diff
@@ -238,5 +242,11 @@ def self.server_version_from_config(config)
       return nil unless config['version'].is_a?(Hash)
       config['version']['version']
     end
+
+    def self.determine_server_type(url, insecure)
+      return :automate if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
+      return :compliance if Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
+      raise CannotDetermineServerType, "Unable to determine if #{url} is a Chef Automate or Chef Compliance server"
```
adamleff: Just tossing out a thought here... I prefer to have exceptions raised at the point of the implementation where "it matters" (for lack of a better phrase). I'd almost prefer to see this method return `nil` if it can't figure out the server type, and let the caller determine how to handle. That way other methods can use this if needed without having to worry about the stack unwinding on them.

```ruby
def self.determine_server_type(url, insecure)
  if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
    :automate
  elsif Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
    :compliance
  end
end
```

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,145 @@
...
+
+        return Login::ComplianceServer.login(options) if options['server_type'] == 'compliance'
```
adamleff: Here is where I'd raise an exception if we don't know the server type since this is actually where an error should be raised to the user:

```ruby
if options['server_type'] == :compliance
  Login::ComplianceServer.login(options)
elsif options['server_type'] == :automate
  Login::AutomateServer.login(options)
else
  raise CannotDetermineServerType, # .....
end
```

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,145 @@
...
+        Login::AutomateServer.login(options)
+      end
+
+      module AutomateServer
+        def self.login(options)
+          verify_thor_options(options)
+
+          options['url'] = options['server'] + '/compliance'
+          token = options['dctoken'] || options['token']
+          puts store_access_token(options, token)
```
adamleff: I realize I'm being a bit pedantic here, but it feels strange to me to have a method called `store_access_token` return a string intended to be user-facing messaging.

I'd rather see `store_access_token` return the `Compliance::Configuration` object it generated to successfully write the config to disk, and then a method that takes that config object and formats a message to the user. It will keep the concerns of the methods succinct and clear. This honestly just reads like a debug message to me. :)

```ruby
config = store_access_token(options, token)

# print out a success message to the user
puts format_status_message(config)
```

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,145 @@
...
+            msg = compliance_store_access_token(options, options['token'])
+          elsif options['user'] && options['password']
+            success, msg = compliance_login_user_pass(options)
+          elsif options['refresh_token']
+            success, msg = compliance_login_refresh_token(options)
+          end
+
+          raise msg unless success
+
+          puts msg
```
adamleff: Same feedback as above - I bet there's a good opportunity for reuse here, actually.

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -18,78 +18,67 @@ def self.subcommand_prefix
...
+    def login_automate(server)
+      warn '[DEPRECATION] `inspec compliance login_automate` is deprecated. Please use `inspec compliance login`'
       options['server'] = server
-      url = options['server'] + '/compliance'
 
-      if url && !options['user'].nil? && !options['ent'].nil? && (!options['dctoken'].nil? || !options['usertoken'].nil?)
-        msg = login_automate_config(url, options['user'], options['dctoken'], options['usertoken'], options['ent'], options['insecure'])
-        puts '', msg
-        exit 0
+      if options['usertoken']
```
adamleff: Despite what was in the issue that preceded this PR, we probably don't need this deprecation here since the whole command is deprecated. As long as the `inspec compliance login` command doesn't accept a `--usertoken` flag, we're covered.

adamleff (changes requested):
This is in great shape, @jerryaldrichiii. I'd like to see a bit of cleanup around where we raise exceptions and keeping method purposes focused. Let me know if any of my feedback is unclear.

jerryaldrichiii:
I made all the changes you suggested @adamleff. Feel free to give it another look! :smile:

adamleff (approved):
Lookin' good, @jerryaldrichiii - thanks!

lib/bundles/inspec-compliance/api.rb
```diff
@@ -5,13 +5,16 @@
 require 'net/http'
 require 'uri'
 
+require_relative 'api/login'
```
chris-rock, line 8: should we move this file to the api directory too? Feels strange to have a `api.rb` and `api` directory. What do you think?

lib/bundles/inspec-compliance/api.rb
```diff
@@ -238,5 +241,13 @@ def self.server_version_from_config(config)
       return nil unless config['version'].is_a?(Hash)
       config['version']['version']
     end
+
+    def self.determine_server_type(url, insecure)
+      if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
```
chris-rock, line 246: Could this also return 200 if the user is already logged in?

lib/bundles/inspec-compliance/api/login.rb
```diff
@@ -0,0 +1,150 @@
...
+      class CannotDetermineServerType < StandardError; end
+
+      def login(options)
+        raise ArgumentError, 'Please specify a server using `inspec compliance login https://SERVER`' unless options['server']
+
+        options['server_type'] = Compliance::API.determine_server_type(options['server'], options['insecure'])
+
+        case options['server_type']
+        when :automate
+          config = Login::AutomateServer.login(options)
```
chris-rock, line 18: Nice separation!

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -18,78 +18,64 @@ def self.subcommand_prefix
...
-        # TODO: we should login with the refreshtoken here
-      elsif !options['refresh_token'].nil?
-        _success, msg = login_refreshtoken(url, options)
-      else
-        puts 'Please run `inspec compliance login SERVER` with options --token or --refresh_token, --user, and --insecure or --not-insecure'
-        exit 1
-      end
-
-      puts '', msg
+      Compliance::API.login(options)
```
chris-rock, line 46: Nice clean-up! 👍

chris-rock (commented):
@jerryaldrichiii Awesome work. I just have some mini-questions before we can go forward

lib/bundles/inspec-compliance/api.rb
adamleff: That's actually pretty standard in Ruby.

jerryaldrichiii: Yeah, it's common in Ruby. I'm fine with it if you are @chris-rock.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: I'm not passing any headers to the `GET` request so it should always act as if the user is not logged in.

lib/bundles/inspec-compliance/api/login.rb
jerryaldrichiii: Thanks! Took some work that's for sure :sweat_smile:

lib/bundles/inspec-compliance/cli.rb
jerryaldrichiii: Thanks! :smiley:

chris-rock (approved):
Works like a charm!

PR #2205, opened 2017-10-03
ssl resource: properly raise error when unable to determine if port is enabled

### LINKED ISSUE #1205: resource fail (exception) aborts all tests

jquick:
This is related to #1205. This will fix the ssl resource for now until we redo the exceptions. Still looking around the code and need to build some unit tests for the ssl resource.
screen shot 2017-10-03 at 10 38 59 am

My fix here is to move the raise condition till later in the flow, specifically the enabled? method. This lets the raise get caught accordingly without killing the other tests.

### DIFF
--- a/test/unit/resources/ssl_test.rb
+++ b/test/unit/resources/ssl_test.rb
@@ -1,6 +1,4 @@
 # encoding: utf-8
-# author: Christoph Hartmann
-# author: Dominik Richter
 
 require 'helper'
 require 'inspec/resource'
--- a/test/unit/resources/ssl_test.rb
+++ b/test/unit/resources/ssl_test.rb
@@ -1,0 +1,57 @@
+# encoding: utf-8
+# author: Christoph Hartmann
+# author: Dominik Richter
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::SSL' do
+  it 'verify cipher enabled' do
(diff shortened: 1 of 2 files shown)


test/unit/resources/ssl_test.rb
```diff
@@ -0,0 +1,57 @@
+# encoding: utf-8
+# author: Christoph Hartmann
+# author: Dominik Richter
```
arlimus: Feel free to remove our names from the authors list :)

jquick: Sounds good, just pushed out a change to remove them.

adamleff (approved):
Thanks, @jquick!

chris-rock (approved):
Thank you for that improvement @jquick

arlimus (approved):
Sweet thank you @jquick :D

PR #2207, opened 2017-10-03
Hack day- dco - Resource documentation update

hannah-radish:
Light formatting changes, change order of example and matchers, slight
color changes

Signed-off-by: hannah-radish <hmaddy@chef.io>


adamleff (approved):
Thanks, @hannah-radish!

PR #2208, opened 2017-10-03
Clean-up kitchen-inspec reference doc

### LINKED ISSUE #1565: No suites run with kitchen inspec if `inspec_tests` directive is not specified.
### Description

The documentation for kitchen-inspec provides information on how to pull suites from a Chef Compliance Server, and how to pull suites from Supermarket; however, it does not clearly provide information on how to use an inspec suite from a local cookbook.

inspec: 1.16.1
kitchen-inspec: 0.17.0
platform version: workstation is Fedora 25, running tests against a RHEL 7.1 server.

### Replication Case

Put inspec tests in a cookbook, following instructions on http://inspec.io/docs/reference/plugin_kitchen_inspec.html/, ensuring that the folder structure is correct. When I run `kitchen verify`, no suites are executed (0 successful, 0 failures, 0 skipped). Since I was not pulling suites from a compliance server nor from Supermarket, did not include an `inspec_specs` directive; however, no suites were run.

### Possible Solutions

I think the easiest solution would be to update the documentation to indicate that, when a user intends to pull a suite from a local cookbook, the `inspec_tests` directive is still needed (this should probably be in a section with the folder structure information).

Or maybe the intent was that in the absence of this directive, kitchen-inspec should still search for suites to run (since the documentation here and in kitchen-inspec doesn't explicitly state that it's required). In this case, kitchen-inspec should probably be updated to actually run the suites available automatically.

### Stacktrace

N/A.

nathenharvey:
* Clarify the copy
* Fix the name of the kitchen configuration file.  `.kitchen.yml`, not
* `kitchen.yml`
* Use proper syntax for specifying recipes in the sample code
* Use a valid Supermarket-based profile in the sample code
* Demonstrate using local InSpec tests

Resolves #1565

Signed-off-by: Nathen Harvey <nharvey@chef.io>

### DIFF
--- a/docs/plugin_kitchen_inspec.md
+++ b/docs/plugin_kitchen_inspec.md
@@ -4,19 +4,19 @@ title: About kitchen-inspec
 
 # kitchen-inspec
 
-The `kitchen-inspec` driver enables InSpec to be used as a verifier within Kitchen.
+Use InSpec as a Kitchen verifier with `kitchen-inspec`.
 
-To use InSpec as a verifier, add it to the kitchen.yml file:
+Add the InSpec verifier to the `.kitchen.yml` file:
 
     verifier:
       name: inspec
 
-To define a suite that pulls its run-list from the Chef Compliance server:
+Use a compliance profile from the Chef Compliance server:
 
     suites:
       - name: compliance
         run_list:
-          - recipe[ssh-hardening]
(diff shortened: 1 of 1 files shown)


adamleff (approved):
Needs more "you"s, but looks good. :)

PR #2209, opened 2017-10-03
http resource: properly execute tests on remote target

### LINKED ISSUE #1436: `http` resource cannot access localhost of the target
### Description

The current http resource is executed on the inspec `workstation` where ruby is executed. A simple test like

```
describe http('http://localhost') do
end
```

demonstrates the complexity. It does not matter if I run:

```
$ hostname
inspechost
# run the tests locally
$ inspec exec http_test.rb
# run the tests agains remote target
$ inspec exec http_test.rb -t ssh://root@remotetarget.com
```

The current implementation will always run against `inspechost` and never against `remotetarget.com`. This may lead to confusion, since all other inspec resources are targeting the OS on `remotetarget.com`

1.10.0

### Possible Solutions

```
# rubocop:disable LineLength
if os['family'] == 'windows'
  curl_command = '&{ Invoke-WebRequest http://localhost:8080 } 2>&1 | Out-String'
else
  curl_command = 'curl http://localhost:8080/'
end
# rubocop:enable LineLength

describe command(curl_command) do
  its('stdout') { should match( ... ) }
end
```

This raises a conceptual question. I think, both use cases are valid run http test on that machine and test a machine from outside. We just need to make it more obvious for InSpec users.

adamleff:
Hack Day Project - Seattle Community Summit 2017!

Currently, the http resource always executes locally, even when scanning a remote machine with `--target` which leads to undesireable behavior.

This change adds the ability to remotely execute tests with curl. This behavior is currently opt-in with the `enable_remote_worker` flag, but will become the default behavior in InSpec 2.0. Deprecation warnings are emitted if the user is scanning a remote target but has not opted in to the new behavior.

Fixes #1436

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -22,51 +22,201 @@ module Inspec::Resources
         its('Content-Length') { should cmp 258 }
         its('Content-Type') { should cmp 'text/html; charset=UTF-8' }
       end
+
+      # properly execute the HTTP call on the scanned machine instead of the
+      # machine executing InSpec. This will be the default behavior in InSpec 2.0.
+      describe http('http://localhost:8080', enable_remote_worker: true) do
+        its('body') { should cmp 'local web server on target machine' }
+      end
     "
 
     def initialize(url, opts = {})
       @url = url
-      @method = opts.fetch(:method, 'GET')
-      @params = opts.fetch(:params, nil)
-      @auth = opts.fetch(:auth, {})
-      @headers = opts.fetch(:headers, {})
-      @data = opts.fetch(:data, nil)
-      @open_timeout = opts.fetch(:open_timeout, 60)
-      @read_timeout = opts.fetch(:read_timeout, 60)
-      @ssl_verify = opts.fetch(:ssl_verify, true)
+      @opts = opts
+
+      if use_remote_worker?
+        return skip_resource 'curl is not available on the target machine' unless inspec.command('curl').exist?
+        @worker = Worker::Remote.new(inspec, http_method, url, opts)
+      else
+        @worker = Worker::Local.new(http_method, url, opts)
+      end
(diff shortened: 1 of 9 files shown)


lib/resources/http.rb
```diff
@@ -22,51 +22,201 @@ class Http < Inspec.resource(1)
         its('Content-Length') { should cmp 258 }
         its('Content-Type') { should cmp 'text/html; charset=UTF-8' }
       end
+
+      # properly execute the HTTP call on the scanned machine instead of the
+      # machine executing InSpec. This will be the default behavior in InSpec 2.0.
```
arlimus, line 27: sweeeeeet!!

schisamo (commented):
Might I suggest you support multiple tools for executing the http commands? Making `curl` a hard requirement in InSpec 2.0 is not very delightful. You could follow a "fall through" pattern like we do in `mixlib-install` and try multiple approaches/tools to execute an HTTP requrest:
https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/generator/bourne/scripts/helpers.sh.erb#L236-L257

adamleff:
@schisamo I'm totally happy to support additional fall-through options. We already do this on the `port` resource, for example. This is an initial implementation and we'll happily accept enhancements to this... including an option to "force local worker" option if that's preferred. I haven't found anyone requesting that as of yet - it's all been people working-around the initial implementation of this by calling `command('curl')` so this felt like a good first step.

arlimus (approved):
What an amazing improvement, and great details on where this applies and doesn't. I love the direction for inspec2 here too 😁 Thank you!!

arlimus:
Great point on additional commands yay! If anyone stumbles on this please raise this as an issue if `curl` does not work for you

Issue #1936, opened 2017-06-16
port is_listening fails unless lsof is already installed


kplimack:
### Description

the `port` matcher works without needing additional dependencies installed.

### InSpec and Platform Version

1.28.0 / centos 6/7

### Replication Case

it returns false (not-listening) when indeed it is up.  installing `lsof` manually fixes this issue.

### Possible Solutions

the matcher should detect if lsof is installed and fall back on netstat, etc if it is not.

adamleff:
This should no longer be an issue. For Linux variants, we first try to use `ss` and then use `netstat`.  We only use `lsof` if the OS is AIX or Darwin. Please comment here if this is still an issue with additional information we can use to troubleshoot. Thanks!

PR #2210, opened 2017-10-03
Uses netstat to detect open ports on AIX

cattywampus:
Related to Issue #1936 - The `lsof` command is only provided in the AIX expansion pack so not all systems may have it installed. However netstat is available, but in true AIX fashion the command line arguments and output is not consistent with the Linux flavor of `netstat`. 

This PR creates a dedicated `AixPorts` implementation that defers to using netstat as the default port scanner for AIX and falls back to `lsof` if necessary. The implementation of this differs from the `LinuxPorts` netstat parser because:
* The argument flags are different
* The address information in the output is structured and parsed differently
* The output doesn't contain process/pid information
* Getting process/pid information requires an additional command

Disclaimer: I am by no means an AIX expert. I pieced this solution together based on this [blog post](http://www.ixbrian.com/blog/?p=70&cm_mc_uid=43755775972014843294058&cm_mc_sid_50200000=1507055147). I am completely open to changing any of this to make it better. All feedback welcome.

I created some mocked unit tests for aix and the port resource. I have also tested this by using `bin/inspec exec` with a `port` control pointing to a real AIX 7.1 server. Let me know if there is additional tests that need to be written to cover these changes.


adamleff:
@cattywampus how are you feeling on this PR? Is this ready for review now? I hesitated to review since I knew it was a WIP, and then I was traveling so I never got back around to looking more deeply in it.

Also, can you rebase / force push your branch up again so Travis will re-test? There was a period where Travis wasn't testing things because of a misconfiguration, and it appears this is one of them.

cattywampus:
@adamleff It's ready to be run through the gauntlet. Rebased and Travis is now passing. Yay!

adamleff (approved):
@cattywampus this looks solid! I don't know enough about AIX to say this is a good approach, but the code looks great and matches the style of the rest of this resource. Thank you for your contribution!

arlimus (approved):
Wow this looks great @cattywampus thank you so much for adding it!! 😁

PR #2216, opened 2017-10-05
Fix `only_if` behavior when used outside controls

### LINKED ISSUE #1914: Profile controls are not all being run with multiple rb files using only_if in some of them
### Description
Basically we have a profile with a list of controls. 

http.rb
php.rb
splunk.rb
sshd.rb
redhat_controls.rb

http.rb, php.rb, plunk.rb and sshd.rb all have only_if statements, which work great.   The redhat_controls does not have an only_if statement and is being skipped.   I renamed to controls_redhat.rb, now it seems all the controls are running.

1.26.0

client - 12.18.31

### Replication Case

create profile with multiple rb control files and include only_if statement, see if they are are checked running inspec

### Possible Solutions

Using only_if causing other controls to be skipped even though they do not have an only_if skip.   Does the order matter for the profiles

jerryaldrichiii:
This renames `@skip_profile` to `@skip_file` and modifies the scope of `only_if` (used outside of a control) to only apply to the control file that contains it instead of the entire profile.

This does this by exposing `@skip_file` from the control context so that it can be set back to `false` between loading control files in the profile context.

This fixes #1914

### DIFF
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -127,6 +127,7 @@ module Inspec
     end
 
     def load_control_file(*args)
+      control_eval_context.skip_file = false
       load_with_context(control_eval_context, *args)
     end
     alias load load_control_file
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -127,6 +127,7 @@ module Inspec
     end
 
     def load_control_file(*args)
+      # Set `skip_file` to `false` between file loads to prevent skips from spanning multiple control files
       control_eval_context.skip_file = false
       load_with_context(control_eval_context, *args)
     end
--- a/test/unit/profiles/profile_context_test.rb
+++ b/test/unit/profiles/profile_context_test.rb
@@ -55,12 +55,9 @@ describe Inspec::ProfileContext do
   let(:backend) { MockLoader.new.backend }
   let(:profile) { Inspec::ProfileContext.new(nil, backend, {}) }
 
-  def get_rule
-    profile.rules.values[0]
-  end
-
-  def get_checks
-    Inspec::Rule.prepare_checks(get_rule)
+  def get_checks(rule_index = 0)
+    rule = profile.rules.values[rule_index]
+    Inspec::Rule.prepare_checks(rule)
   end
 
   it 'must be able to load empty content' do
@@ -178,8 +175,8 @@ describe Inspec::ProfileContext do
       it 'doesnt extend into other control files' do
         profile.load_control_file(if_false + control)
         profile.load_control_file(control_2)
-        check_1 = get_checks
-        check_2 = Inspec::Rule.prepare_checks(profile.rules.values[1])
+        check_1 = get_checks(0)
+        check_2 = get_checks(1)
         check_1[0][1][0].resource_skipped.must_equal 'Skipped control due to only_if condition.'
         check_2[0][1][0].must_be_nil
       end
--- a/test/unit/profiles/profile_context_test.rb
+++ b/test/unit/profiles/profile_context_test.rb
@@ -121,7 +121,8 @@ describe Inspec::ProfileContext do
       let(:if_true) { "only_if { true }\n" }
       let(:if_false) { "only_if { false }\n" }
       let(:describe) { "describe nil do its(:to_i) { should eq rand } end\n" }
-      let(:control) { "control 1 do\n#{describe}end" }
+      let(:control) { "control 1 do\n#{describe}\nend\n" }
+      let(:control_2) { "control 2 do\n#{describe}\nend\n" }
 
       it 'provides the keyword' do
         profile.load(if_true)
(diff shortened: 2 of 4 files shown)


test/unit/profiles/profile_context_test.rb
```diff
@@ -173,6 +174,15 @@ def load(call)
         get_checks.length.must_equal 1
         get_checks[0][1][0].resource_skipped.must_equal 'Skipped control due to only_if condition.'
       end
+
+      it 'doesnt extend into other control files' do
+        profile.load_control_file(if_false + control)
+        profile.load_control_file(control_2)
+        check_1 = get_checks
+        check_2 = Inspec::Rule.prepare_checks(profile.rules.values[1])
```
adamleff: Can we add an optional parameter to get_checks to have it pull values other than the first? That way we're not defining this logic in multiple places.

adamleff (changes requested):
This looks fine to me. I'd like to see us keep the call to `Inspec::Rule.prepare_checks` consolidated so we're not calling a combination of `get_checks` and a hand-crafted call to `Inspec::Rule.prepare_checks`.

test/unit/profiles/profile_context_test.rb
jerryaldrichiii: I agree! Fixed :smile:

jerryaldrichiii:
I agree @adamleff. I modified `get_checks` to take a rule index.

adamleff:
@jerryaldrichiii This change does not work for me locally. I have a profile with five controls files. Three of them have, at the file level:

```ruby
only_if { 1 == 2 }
```

One of them has:
```ruby
only_if { 1 == 1 }
```

And one has no `only_if`

When I exec this profile with your branch, all controls execute. When I exec against master, I get the initially-broken behavior.

My expectation is that I would've had two control files run, three skipped.

adamleff:
Broken behavior on master:

```
  ✔  http: http
     ✔  http should cmp == "http"
  ↺  php: php
     ↺  Skipped control due to only_if condition.
  ↺  redhat: redhat
     ↺  Skipped control due to only_if condition.
  ↺  splunk: splunk
     ↺  Skipped control due to only_if condition.
  ↺  sshd: sshd
     ↺  Skipped control due to only_if condition.
```

... but redhat and splunk **should have* been run.

With this branch:

```
  ✔  http: http
     ✔  http should cmp == "http"
  ✔  php: php
     ✔  php should cmp == "php"
  ✔  redhat: redhat
     ✔  redhat should cmp == "redhat"
  ✔  splunk: splunk
     ✔  splunk should cmp == "splunk"
  ✔  sshd: sshd
     ✔  sshd should cmp == "sshd"
```

... but http, php, and sshd should've been skipped.

jerryaldrichiii:
Looking at your examples I see that your `only_if` statements are at the bottom of the files vs the top.

I did not test with them at the bottom. I will add functional tests to avoid this in the future. I will also modify this to have the `only_if` apply to the whole file regardless of position.

lib/inspec/profile_context.rb
```diff
@@ -127,6 +127,7 @@ def load_libraries(libs)
     end
 
     def load_control_file(*args)
+      control_eval_context.skip_file = false
```
adamleff: I'm curious what this is needed for... isn't this false by default? It also feels weird to me for this to be disclosed as a public API on the control eval context, but perhaps I'm missing an idiosyncrasy.

adamleff (commented):
@jerryaldrichiii this looks good - just a question before I approve.

lib/inspec/profile_context.rb
jerryaldrichiii: Ah, that was a tricky one! From what I could tell (and I could be wrong because this was quite a rabbit hole) is that we only create a `control_eval_context` once and it is set to `false` at that time. Then later on when an `only_if` evaluated to `false` we toggled that value to `true`.

The issue was caused because we never toggled that back to `false`. So I needed a way to set that to `false` between each evaluation/load of a control file. That was the only place I could find that fit the bill.

It also feels weird to me, but I'm not sure how to get around it. Suggestions are welcome!

adamleff: Your explanation is spot-on. The fact that you had to explain it is the problem. :) A one-line code comment above this line explaining why this is necessary will fit the bill.

```
# Set skip_file to false before we read in a new file, just in case the prior file was skipped
```

jerryaldrichiii: Great! I agree about the comment. It has been added.

jerryaldrichiii:
Looks like I'm failing Rubocop due to:
```
lib/inspec/plugins/resource.rb:40:5: C: Method has too many lines. [49/40]
    def __register(name, obj) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^
```

It's due to us defining a class in a method and the several methods I've added. @adamleff should we ignore this violation? If not, most of these methods can be defined on one line...but I don't like the way that looks. I'm open to other suggestions as well.

Also, thank you so much for all the effort you've put into these reviews. They helped the project and me greatly.

adamleff (approved):
Nicely done, @jerryaldrichiii!

chris-rock (approved):
Thank you @jerryaldrichiii That was a difficult one 🥇

PR #2217, opened 2017-10-05
Add nil check for sshd config file

### LINKED ISSUE #1778: undefined method `empty?' for nil:NilClass with sshd_conf
### Description
The resource sshd_config displays `(undefined method `empty?' for nil:NilClass)` if you don't use --sudo in the inspec command line. A more descriptive error message would much more helpful.

1.23.0 on OS X 10.10.5

### Replication Case
Using the following:
`control "con42" do
  impact 1.0
  title "verify that we are using sshd v2"
  desc "Using sshd v2."
  describe sshd_config do
    its('Protocol') { should cmp 2 }
  end
end `

Using the command line `inspec exec test.rb -t ssh://example.com` results in 

> `con42: verify that we are using sshd v2 (undefined method `empty?' for nil:NilClass)
     *  SSH Configuration Protocol
     undefined method `empty?' for nil:NilClass
`

Using the command line `inspec exec test.rb --sudo -t ssh://example.com results expect results
> `  ×  con42: verify that we are using sshd v2 (
     expected: "2"
          got: nil
'

### Possible Solutions

yes, you need to use --sudo but it you don't I feel its revealed a bug somewhere along the line and a more descriptive error message would much more help. Something like "permission denied. try adding --sudo to the inspec commandline"

If you have any questions I can be reached on the inspec slack channel as @gmoore1 or at gmoore1 at gci.com

jquick:
This fixes #1778. There was a issue where if the user did not have read permissions on /etc/ssh/sshd_config it would error out on the empty? check. The fix here is to also look for nil on the file content. Along with this I refactored the inspec file empty? check as it does not exist and was also erroring during my testing. Output with the changes:
screen shot 2017-10-05 at 10 43 37 am

### DIFF
--- a/lib/resources/ssh_conf.rb
+++ b/lib/resources/ssh_conf.rb
@@ -63,7 +63,7 @@ module Inspec::Resources
       end
 
       @content = file.content
-      if @content.nil? || (@content.empty? && File.size?(@conf_path))
+      if @content.nil? || (@content.empty? && !file.size.zero?)
         return skip_resource "Can't read file \"#{@conf_path}\""
       end
 
--- a/lib/resources/ssh_conf.rb
+++ b/lib/resources/ssh_conf.rb
@@ -63,7 +63,7 @@ module Inspec::Resources
       end
 
       @content = file.content
-      if @content.empty? && !file.empty?
+      if @content.nil? || (@content.empty? && File.size?(@conf_path))
         return skip_resource "Can't read file \"#{@conf_path}\""
       end
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -104,13 +104,16 @@ class MockLoader
       end
       md
     }
+    emptyfile = lambda {
+      mockfile.call('emptyfile')
+    }
 
     mock.files = {
       '/proc/net/bonding/bond0' => mockfile.call('bond0'),
       '/etc/ssh/ssh_config' => mockfile.call('ssh_config'),
       '/etc/ssh/sshd_config' => mockfile.call('sshd_config'),
       '/etc/ssh/sshd_config_does_not_exist' => mockfile.call('sshd_config_does_not_exist'),
-      '/etc/ssh/sshd_config_empty' => mockfile.call('sshd_config_empty'),
+      '/etc/ssh/sshd_config_empty' => emptyfile.call,
       '/etc/passwd' => mockfile.call('passwd'),
       '/etc/shadow' => mockfile.call('shadow'),
       '/etc/ntp.conf' => mockfile.call('ntp.conf'),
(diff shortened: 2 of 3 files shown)


lib/resources/ssh_conf.rb
```diff
@@ -63,7 +63,7 @@ def read_content
       end
 
       @content = file.content
-      if @content.empty? && !file.empty?
+      if @content.nil? || (@content.empty? && File.size?(@conf_path))
```
adamleff: We can't use the Ruby `File` class because this will always operate locally on the machine executing InSpec. If the user is scanning a remote target with `--target`, this call will still happen on the local machine. That's why the call to `file` was being done (lowercase `file` referring to the variable storing the `inspec.file` resource that will work both remotely and locally).

test/helper.rb
```diff
@@ -109,6 +109,8 @@ def md.directory?
       '/proc/net/bonding/bond0' => mockfile.call('bond0'),
       '/etc/ssh/ssh_config' => mockfile.call('ssh_config'),
       '/etc/ssh/sshd_config' => mockfile.call('sshd_config'),
+      '/etc/ssh/sshd_config_does_not_exist' => mockfile.call('sshd_config_does_not_exist'),
```
adamleff, line 115: I wonder if we should create a `fileempty` object, much like the `empty` object you see in the command mocks, so we don't have to create empty mock files.  Like this, but for files: https://github.com/chef/inspec/blob/master/test/helper.rb#L339

test/unit/resources/ssh_conf_test.rb
```diff
@@ -35,5 +35,18 @@
         '/etc/ssh/ssh_host_ecdsa_key',
       ]
     end
+
+    it 'check bad path' do
+      resource = load_resource('sshd_config', '/etc/ssh/sshd_config_does_not_exist')
+      _(resource.send(:read_content)).must_equal "Can't find file \"/etc/ssh/sshd_config_does_not_exist\""
+      assert_nil(resource.Protocol)
```
adamleff: We have not used `assert_nil` anywhere in the codebase yet and instead use `must_be_nil`. We should probably be consistent.

adamleff (changes requested):
@jquick this is a great first pass at this. We need to make sure we're always using InSpec resource, and not the ruby stdlib, when running commands or reading files... otherwise, we'll break the experience for remote scanned targets.

lib/resources/ssh_conf.rb
jquick: There is no inspec.file.empty? currently. Should we create one that does a size check accordingly?

adamleff: Since we basically just hand over logic on `file`s to Train, it should probably be implemented there. Can we do `file.size == 0` for now and address that in a future PR?

jquick: sounds good to me

test/helper.rb
jquick: Created the emptyfile object for the empty file mock. For this one I want it to point to file it cannot find.

adamleff (approved):
Great fix, @jquick - thanks!

chris-rock (approved):
Thank you @jquick for your contribution! This is a great improvement.

PR #2218, opened 2017-10-05
Update docs for the `os` resource

### LINKED ISSUE #2215: os resource does not work as expected
### Description

when using the 'os' resource in inspec shell, it appears the os resource does not work as expected. For instance:

inspec> os[:family]
=> "debian"
inspec> os[:debian]
=> nil
inspec> os[:family] == 'debian'
=> true
inspec> os[:windows]
=> nil

os[:debian] should NOT return nil.

`#inspec version`
      rake (>= 0)
1.40.0

`#uname -a`
Linux kali 4.6.0-kali1-amd64 #1 SMP Debian 4.6.4-1kali1 (2016-07-21) x86_64 GNU/Linux

### Replication Case

`#inspec shell`
inspec> os[:family]
=> "debian"
inspec> os[:family] == 'debian'
=> true
inspec> os[:debian]
=> nil

### Possible Solutions

NA

### Stacktrace

NA

adamleff:
The docs for the `os` resource did not have the proper parameters listed and also improperly had `os[:debian]` examples instead of `os.debian?`

Fixes #2215

### DIFF
--- a/docs/resources/os.md.erb
+++ b/docs/resources/os.md.erb
@@ -13,17 +13,17 @@ Use the `os` InSpec audit resource to test the platform on which the system is r
 An `os` resource block declares the platform to be tested. The platform may specified via matcher or control block name. For example, using a matcher:
 
     describe os[:family] do
-      it { should eq 'platform_name' }
+      it { should eq 'platform_family_name' }
     end
 
-or using the block name:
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @adamleff for taking such care for the details!

PR #2221, opened 2017-10-06
Support symbol keys in ObjectTraverser

### LINKED ISSUE #2036: YAML resource with symbol keys
### Description

The yaml resource assumes the yaml file has strings as keys

```
$ chef -v
Chef Development Kit Version: 2.0.28
chef-client version: 13.2.20
delivery version: master (17c1b0fed9be4c70f69091a6d21a4cbf0df60a23)
berks version: 6.2.0
kitchen version: 1.16.0
inspec version: 1.31.1
```

### Replication Case

For a given file on disk `file.yml`

```yaml
---
:key: value
```

And a test:

```yaml
describe yaml('file.yml') do
  its(:key) { should eq 'value' }
end
```

Outputs:

```
  ×  YAML file.yml max_revisions should eq "value"
     
     expected: "value"
          got: nil
     
     (compared using ==)
```

### Possible Solutions

The YAML resource inherits this logic from the JSON resource. As there is no such thing as Symbol in JSON, this isn't an issue for the JSON resource.

Looks like the problem lies with [`ObjectTraverser`](https://github.com/chef/inspec/blob/56986d1bdbb1b0dd2f0c7c60804e2ab617678940/lib/utils/object_traversal.rb#L16) that the JSON resource uses. [`ObjectTraverser`](https://github.com/chef/inspec/blob/56986d1bdbb1b0dd2f0c7c60804e2ab617678940/lib/utils/object_traversal.rb#L16) calls `#to_s` on the key, and the value comes back nil.

Solution could be to not call `#to_s`, but this may might cause issues elsewhere with other things that use [`ObjectTraverser`](https://github.com/chef/inspec/blob/56986d1bdbb1b0dd2f0c7c60804e2ab617678940/lib/utils/object_traversal.rb#L16).

adamleff:
As detected in #2036, it is not possible to extract values from a YAML file if the key is a symbol. This change refactors ObjectTraverser to support symbol keys before attempting to stringify them.

Fixes #2036

### DIFF
--- a/lib/utils/object_traversal.rb
+++ b/lib/utils/object_traversal.rb
@@ -3,22 +3,47 @@
 # author: Christoph Hartmann
 module ObjectTraverser
   def extract_value(keys, value)
-    key = keys.shift
-    return nil if key.nil? || value.nil?
+    return nil if value.nil?
 
-    if value.is_a?(Array)
-      value = if key.is_a?(Fixnum)
-                value[key]
-              elsif value.respond_to?(key.to_sym)
-                value.send(key.to_sym)
-              end
-    else
-      value = value[key.to_s].nil? ? nil : value[key.to_s]
-    end
+    key = keys.shift
+    return nil if key.nil?
+
+    # if the current value is not a Hash or Array, it is undefined
+    # behavior so value will be assigned nil by default.
+    value = if value.is_a?(Array)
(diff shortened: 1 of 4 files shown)


arlimus (approved):
Wow sweet catch and improvement!! Thank you so much 😁

PR #2222, opened 2017-10-06
processes resource: support busybox ps

### LINKED ISSUE #2034: Unable to detect running processes on Alpine-based Docker containers
### Description

When executing against a running Docker container based on Alpine Linux, `processes` appears unable to detect running processes. I am super new to InSpec so very open to the possibility I'm just doing something wrong.

InSpec 1.31.1 against Alpine 3.6 as Docker container on macOS 10.12.4

### Replication Case

Launch an Alpine container with a process running and attempt to assert the process is running with 
```ruby
describe processes(process_name) do
    it{ should exist }
end
```
Testcase implemented here https://github.com/MFAnderson/inspec-alpine-bug

adamleff:
This change enhances the processes resource to support the busybox ps command which is common on Alpine, for example. The way we map ps fields to the structs needed by FilterTable have also been refactored to be more flexible so we can support multiple formats in the future.

Also, the processes resource now allows the grep argument to be optional thus allowing a user to query all resources without passing in a match-all regex.

Fixes #2034

### DIFF
--- a/docs/resources/processes.md.erb
+++ b/docs/resources/processes.md.erb
@@ -18,7 +18,7 @@ A `processes` resource block declares the name of the process to be tested, and 
 
 where
 
-* `processes('process_name')` specifies the name of a process to check. If this is a string, it will be converted to a Regexp. For more specificity, pass a Regexp directly.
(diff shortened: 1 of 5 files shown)


arlimus (approved):
This is amazing thank you so much for adding the support for busybox ps!! 🎉

arlimus:
massaging travis a bit...

PR #2223, opened 2017-10-06
Fix loading profile files when executing multiple profiles

### LINKED ISSUE #2004: RuntimeProfile overwritten when executing multiple profiles
### Description

RuntimeProfile overwritten when executing multiple profiles, preventing a profile from reliably accessing its profile object via `inspec.profile` in a control file.

As `Inspec::Profile` objects are created, a `Inspec::RuntimeProfile` object is created its object is stored in `@backend.profile`. A single backend instance is created by `Inspec::Runner` and is passed to `Inspec::Profile` when creating the profile instance, so the last profile created wins and its RuntimeProfile instance will be the only one stored in the backend instance.

When controls are then loaded, no matter which profile they are from, the RuntimeProfile instance available via `inspec.profile` will be that of the last profile created. For example, if running `inspec exec` with three profiles, and the first profile attempts to call `inspec.profile.file`, it will fail because the file in question will not exist in the file list of the third profile.

InSpec v1.31.1

### Replication Case

 * Create two profiles, named "one" and "two"
 * Add a file: `one/files/sample.txt`
 * Add a control: `one/controls/example.rb` and call `inspec.profile.file('sample.txt')`
 * Run: `inspec exec one two` -- error recieved: `Cannot find file sample.txt in profile.`
 * Run: `inspec exec one` -- no error received.

For added detail, add a line similar to this at the type of each control file and see that the object IDs are the same and the metadata shown is from the last profile argument passed to `inspec exec`:

```ruby
puts "IN PROFILE one: (#{inspec.profile.object_id}) #{inspec.profile.instance_variable_get(:@src).metadata.inspect}\n\n"
```

### Possible Solutions

 * Create a unique backend instance for each profile instead of a single one for all profiles
 * Add a new class for tracking profile context/state, expose a DSL keyword for accessing it, and store the RuntimeProfile object there instead of on the Inspec::Backend object
 * A better idea I haven't thought of yet...?

I'd personally vote for creating a unique backend object for each target added to a runner. It shouldn't be impactful to performance with the exception of making a connection to the target per profile instead of a shared connection.

adamleff:
When running `inspec exec` with multiple profiles, such as:

inspec exec profile1 profile2

... profile1 control calls to `inspec.profile.file` will incorrectly try to pull files from profile2 because the RuntimeProfile object is stored on the backend object, and we share the backend object.

This change ensures each profile has a unique backend object to ensure the RuntimeProfile instance is not overwritten.

Fixes #2004

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -93,12 +93,20 @@ module Inspec
       @writable = options[:writable] || false
       @profile_id = options[:id]
       @cache = options[:cache] || Cache.new
-      @backend = options[:backend] || Inspec::Backend.create(options.select { |k, _| k != 'target' })
       @attr_values = options[:attributes]
       @tests_collected = false
       @libraries_loaded = false
       Metadata.finalize(@source_reader.metadata, @profile_id, options)
 
+      # if a backend has already been created, clone it so each profile has its own unique backend object
+      # otherwise, create a new backend object
+      #
+      # This is necessary since we store the RuntimeProfile on the backend object. If a user runs `inspec exec`
+      # with multiple profiles, only the RuntimeProfile for the last-loaded profile will be available if
+      # we share the backend between profiles.
+      #
+      # This will cause issues if a profile attempts to load a file via `inspec.profile.file`
+      @backend = options[:backend].nil? ? Inspec::Backend.create(options) : options[:backend].dup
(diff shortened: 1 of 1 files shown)


arlimus (approved):
Amazing fix, thank you Adam!!

PR #2225, opened 2017-10-06
Support PAX-formatted tar files, standardize file lists

### LINKED ISSUE #2028: InSpec cannot read vendored tar's with PAX extensions
### Description

Some tar files may be archived in a "special" way. When looking at them in the Ruby debugger, we see something called the pax header:

```
vendor/PaxHeaders.0/db143b033...
```

1.31.1

### Replication Case

Try `inspec json archive.tar.gz` on a file with vendored tar archives with pax.

### Possible Solutions

Do not embed vendored tar archives, use flat files instead which don't suffer from this problem.

### Other notes

Afaics, this may be an underlying problem in Ruby's TarReader. It doesn't seem to support the PAX header. It may affect other files than embedded vendored archives.

adamleff:
When a tar file is generated in PAX format, the files have an additional relative path prefix added to them. For example, instead of:

`inspec.yml`

... the file is listed as:

`./inspec.yml`

And the source reader plugin looks only for a "inspec.yml" file to determine the profile format.

This change addresses this issue by normalizing the file paths in the TarReader and accounting for the additional "./" prefix that may exist whenever the tar file is walked looking for a file to read its content.

Fixes #2028

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -11,7 +11,6 @@ end
 
 require 'minitest/autorun'
 require 'minitest/spec'
-require 'mixlib/shellout'
 require 'webmock/minitest'
 require 'mocha/setup'
 require 'fileutils'
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -486,21 +486,6 @@ class MockLoader
     dst
   end
 
-  def self.profile_pax_tgz(name)
-    path = File.join(home, 'mock', 'profiles', name)
-    dst = File.join(Dir.mktmpdir, "#{name}.tar.gz")
-
-    # generate relative paths
-    files = Dir.glob("#{path}/**/*").delete_if { |x| !File.file?(x) }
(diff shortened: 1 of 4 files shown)


test/helper.rb
```diff
@@ -471,12 +471,13 @@ def self.load_profile(name, opts = {})
 
   def self.profile_tgz(name)
     path = File.join(home, 'mock', 'profiles', name)
-    dst = File.join(Dir.tmpdir, Dir::Tmpname.make_tmpname(name, '.tar.gz'))
+    dst = File.join(Dir.mktmpdir, "#{name}.tar.gz")
```
arlimus, line 474: nice simplification 😁

arlimus (approved):
Thank you for supporting these tricky PAX-formatted files @adamleff !! I remember when I first stumbled into it and things got tricky... Kudos!

Issue #1783, opened 2017-05-08
Prevent `/` in profile name


chris-rock:
### Description

Briefly describe the issue and the expected behavior

### InSpec and Platform Version

1.22.0

### Replication Case

```
inspec init profile test/profile
Create new profile at /Users/chartmann/Development/compliance/inspec/test/profile
 * Create directory controls
 * Create file controls/example.rb
 * Create file inspec.yml
 * Create directory libraries
 * Create file README.md
 * Create file libraries/.gitkeep
➜  inspec git:(master) ✗ cat test/profile/inspec.yml 
name: test/profile
title: InSpec Profile
maintainer: The Authors
copyright: The Authors
copyright_email: you@example.com
license: All Rights Reserved
summary: An InSpec Compliance Profile
version: 0.1.0
```

### Possible Solutions

* Slug the name
* `inspec check` should ensure that only slugable url characters are used

adamleff:
Adding this as an enhancement and putting in the 2.0 milestone since this would be breaking behavior (though adding a warning now and an error later would be totally cool).

clintoncwolfe:
Just to clarify, is it acceptable to have profiles in subdirectories so long as their names are slugged?
```
inspec init profile test/profile
Create new profile at /Users/chartmann/Development/compliance/inspec/test/profile
#...
cat test/profile/inspec.yml 
name: test_profile
```

Or is the goal here to require profiles be at the same level on the filesystem?

adamleff:
@clintoncwolfe totally fine for there to be profiles in subdirectories and for the CLI to honor them, much like you discovered with `inspec init profile`. There just shouldn't be any non-sluggable characters in the profile name.

I just discovered that you can do `inspec init profile \{\{foo\}\}` and it will totally create a directory called `{{foo}}` and also set the profile name to `{{foo}}` in inspec.yml, which breaks horribly. We should fix that as part of this issue too... for whoever decides to pick this one up :)

clintoncwolfe:
We'll need to actively slug, at least based on this line, which will create invalid (space and slash containing) names when the target is a local path:

https://github.com/chef/inspec/blob/29c80110a81e1bb01055407ed40038a8f082d279/lib/inspec/metadata.rb#L215

adamleff:
[...]

PR #2231, opened 2017-10-09
Issue warning during check if profile name contains slash

### LINKED ISSUE #1954: Deprecate use of '/' in profile names
#1783 will cause profile names with `/` characters to throw an error. We should add a deprecation warning to `inspec check` now to ensure this is communicated prior to InSpec 2.0's release.

clintoncwolfe:
Fixes #1954 .   Also, in the case that no name can be determined and a name must be generated from the target (which is common in unit testing), replace slashes with dots.  This is sensible if the default name is arbitrary; if real-world use relies on having profiles with names like "tests from /path/to/tests", we can update the unit tests instead; however any real-world processes would have issues in InSpec 2.0 anyway due to #1783 .

### DIFF
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -11,7 +11,7 @@ require 'utils/spdx'
 
 module Inspec
   # Extract metadata.rb information
-  # A Metadata object may be created and finalized with invalid data.  
+  # A Metadata object may be created and finalized with invalid data.
   # This allows the check CLI command to analyse the issues.
   # Use valid? to determine if the metadata is coherent.
   class Metadata # rubocop:disable Metrics/ClassLength
@@ -115,10 +115,10 @@ module Inspec
         errors.push("Missing profile #{field} in #{ref}")
       end
 
-      if params[:name] =~ /[\/\\]/
+      if params[:name] =~ %r{[\/\\]}
         warnings.push("Profile names containing slashes (#{params[:name]}) are deprecated.")
       end
-      
+
       # if version is set, ensure it is correct
       if !params[:version].nil? && !valid_version?(params[:version])
         errors.push('Version needs to be in SemVer format')
@@ -221,7 +221,7 @@ module Inspec
       # create a new name based on the original target if it exists
       # Crudely slug the target to not contain slashes, to avoid breaking
       # unit tests that look for warning sequences
-      metadata.params[:name] = "tests from #{original_target}".gsub(/[\\\/]/, '.') unless original_target.to_s.empty?
+      metadata.params[:name] = "tests from #{original_target}".gsub(%r{[\\\/]}, '.') unless original_target.to_s.empty?
     end
 
     def self.finalize(metadata, profile_id, options, logger = nil)
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -11,6 +11,9 @@ require 'utils/spdx'
 
 module Inspec
   # Extract metadata.rb information
+  # A Metadata object may be created and finalized with invalid data.  
+  # This allows the check CLI command to analyse the issues.
+  # Use valid? to determine if the metadata is coherent.
   class Metadata # rubocop:disable Metrics/ClassLength
     attr_reader :ref
     attr_accessor :params, :content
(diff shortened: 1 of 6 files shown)


lib/inspec/metadata.rb
```diff
@@ -112,6 +115,10 @@ def valid # rubocop:disable Metrics/AbcSize
         errors.push("Missing profile #{field} in #{ref}")
       end
 
+      if params[:name] =~ %r{[\/\\]}
+        warnings.push("Profile names containing slashes (#{params[:name]}) are deprecated.")
```
adamleff: Can I suggest changing this to be a bit more explicit? Perhaps:

```
Your profile name (#{params[:name}) contains a slash which will not be permitted in InSpec 2.0. Please change your profile name in the `inspec.yml` file.
```

lib/inspec/metadata.rb
```diff
@@ -212,7 +219,9 @@ def self.finalize_name(metadata, profile_id, original_target)
       return unless metadata.params[:title].nil?
 
       # create a new name based on the original target if it exists
-      metadata.params[:name] = "tests from #{original_target}" unless original_target.to_s.empty?
+      # Crudely slug the target to not contain slashes, to avoid breaking
+      # unit tests that look for warning sequences
+      metadata.params[:name] = "tests from #{original_target}".gsub(%r{[\\\/]}, '.') unless original_target.to_s.empty?
```
adamleff: What do you think about using hyphens instead of periods? That's a more common word separator when slugging a URL, based on absolutely zero evidence I have to provide to you except my gut instinct and some googling. :)

I also wonder if this change is necessary... should we put this line in the `title` metadata instead and put something else in the `name`?

adamleff (commented):
This looks pretty solid. Just a couple questions/suggestions. Lemme know what you think.

lib/inspec/metadata.rb
clintoncwolfe: I was debating between dashes, underscores, and "something else shell-safe" (I settled on a period, as you see).  My concern with either dashes or underscores is that they're both common in both directory names and profile names, so slugging would lose the path separator....
```
my/profiles/from-steve/no-shenanigans/inspec.yml =>
my-profiles-from-steve-no-shenanigans
```
Maybe that's fine.  

Placing it (unslugged, as originally generated) in `title` works fine.  What should we put in `name` ?  "default" ?

adamleff: I'm pretty agnostic to the hyphens vs periods. I'll let someone else weigh in on that.

I guess what I'm coming back to on this particular change is if this is even necessary since this line is only hit when we're dealing with a file of tests outside of the construct of a profile. I suspect it's not even needed, but given your comment about wishing to potentially catch this in unit tests, I'm totally fine as-is.

adamleff: Since the CLI output will look funky when we strip the slashes, and profiles need to have names, how about we put the actual path as-is in the `title` so that way it at least shows in the output unaltered?

adamleff (approved):
Thanks, @clintoncwolfe! :)

test/unit/profiles/profile_test.rb
```diff
@@ -108,7 +108,6 @@
       it 'prints loads of warnings' do
         logger.expect :info, nil, ["Checking profile in #{home}/mock/profiles/#{profile_id}"]
         logger.expect :error, nil, ["Missing profile version in inspec.yml"]
-        logger.expect :warn, nil, ["Missing profile title in inspec.yml"]
```
chris-rock, line 111: I would expect that we see the result, if the inspec.yml is empty. Am I missing something?

clintoncwolfe: If the inspec.yml is missing, we generate a profile name.  Previously that name contained the path to the profile, which likely contained slashes.  As that's not permitted, we replaced the slashes; but to retain clarity, we now store the original path in the title.

So, a profile without metadata will now have both a name and a title generated, and thus there will be no warning about a missing title.

Discussion here: https://github.com/chef/inspec/pull/2231#discussion_r143976150

adamleff: @chris-rock this makes sense to me. It's only in the case where a profile exists, an inspec.yml exists, and the file is empty (or doesn't include a name). By default, we'll create a default name that includes the path (slugged), and a title that contains the actual path.

Since in the case of empty metadata / no name, we'll always have a title, then it's right to remove this test assertion.

chris-rock (approved):
Thank you @clintoncwolfe for this improvement and the great discussion!

PR #2234, opened 2017-10-09
PR to resolve the weird encoding issue within inspec shell

### LINKED ISSUE #1967: Weird encoding issue when launching inspec shell
### Description

Weird encoding issue when launching `inspec shell`

- inspec 1.25.1
- Ubuntu 16.04.2

**Does not work in:**
 - gnome-terminal 3.18.3
 - tilix 1.6.1
 - terminator 0.98-1

**Works in:**
 - xterm

Any ideas?

terminator
tilix
gnome-terminal
xterm

username-is-already-taken2:
This PR is an attempt to resolve #1967 

I have observed this on few different operating systems.    I use to suffer from the issue when you use the backspace key it would jump your cursor postion 10 places, which is why I think the function `readline_ignore` was added.  

From my testing today this was due to the use of colour when defining the prompt.  for whatever reason the extra chars I see are a result of `mark` calling out to the `readline_ignore` method and having `\001` & `\002` (the readline ignore flags) added.

The `readline_ignore` seems to work fine for the prompt but not for the `mark` method, so I revised the method not to use it.

Before ....
image
image
image

After ...
image
image

rightly or wrongly I've also amend the `mark` colour escape code, I don't envy you having to design a cross os ascii colour scheme, from my testing the same esc code resulted in different colours depending on the os lol :) (unless you wanted red but I thought red was bad :)  so for the mark method I opted for the system default and bumped the green for the prompt.

image
image

I've tested it on windows rhel and centos, I think it looks pretty good, but happy to go with what you think.

Lastly removing the couple of extra spaces within `print_target_info` is just for my ocd 

You can flame me in the morning if you don't agree, see you at the summit.

### DIFF
--- a/lib/inspec/shell.rb
+++ b/lib/inspec/shell.rb
@@ -42,7 +42,7 @@ module Inspec
 
       # configure pry shell prompt
       Pry.config.prompt_name = 'inspec'
-      Pry.prompt = [proc { "#{readline_ignore("\e[0;32m")}#{Pry.config.prompt_name}> #{readline_ignore("\e[0m")}" }]
+      Pry.prompt = [proc { "#{readline_ignore("\e[1m\e[32m")}#{Pry.config.prompt_name}> #{readline_ignore("\e[0m")}" }]
 
       # Add a help menu as the default intro
       Pry.hooks.add_hook(:before_session, 'inspec_intro') do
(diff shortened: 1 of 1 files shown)


adamleff (approved):
Escape characters and color encoding is a mystery to me, but this looks great!

chris-rock (approved):
Thank you @username-is-already-taken2 Another great PR! 🥇

PR #2235, opened 2017-10-10
Add non-halting exception support to resources

### LINKED ISSUE #1205: resource fail (exception) aborts all tests

jerryaldrichiii:
This adds two `Inspec::Exceptions` that can be used within resources to either skip or fail a test without halting execution.

This fixes #1205.

### DIFF
--- a/lib/inspec/exceptions.rb
+++ b/lib/inspec/exceptions.rb
@@ -6,7 +6,7 @@ module Inspec
     class AttributesFileDoesNotExist < ArgumentError; end
     class AttributesFileNotReadable < ArgumentError; end
     class ResourceFailed < StandardError; end
-    class SecretsBackendNotFound < ArgumentError; end
     class ResourceSkipped < StandardError; end
+    class SecretsBackendNotFound < ArgumentError; end
   end
 end
--- a/lib/inspec/exceptions.rb
+++ b/lib/inspec/exceptions.rb
@@ -5,6 +5,8 @@ module Inspec
   module Exceptions
     class AttributesFileDoesNotExist < ArgumentError; end
     class AttributesFileNotReadable < ArgumentError; end
+    class ResourceFailed < StandardError; end
     class SecretsBackendNotFound < ArgumentError; end
+    class ResourceSkipped < StandardError; end
   end
 end
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -40,9 +40,13 @@ module Inspec
     def __register(name, obj)
       cl = Class.new(obj) do
         def initialize(backend, name, *args)
+          @resource_skipped = nil
+          @resource_failed = nil
+
           # attach the backend to this instance
           @__backend_runner__ = backend
           @__resource_name__ = name
+
           # call the resource initializer
           begin
             super(*args)
@@ -63,20 +67,32 @@ module Inspec
           @example = example
         end
 
-        def resource_skipped
-          @resource_skipped if defined?(@resource_skipped)
-        end
-
         def skip_resource(message)
           @resource_skipped = message
         end
 
-        def resource_failed
-          @resource_failed if defined?(@resource_failed)
+        def resource_skipped?
+          !@resource_skipped.nil?
         end
+
+        def resource_skipped_message
+          @resource_skipped
+        end
+        # This alias exists for backwards compatibility
+        # It is more idiomatic to have a `?` and `_message` methods
+        # TODO: Remove in InSpec 2.0
+        alias_method :resource_skipped, :resource_skipped_message
 
         def fail_resource(message)
           @resource_failed = message
+        end
+
+        def resource_failed?
+          !@resource_failed.nil?
+        end
+
+        def resource_failed_message
+          @resource_failed
         end
 
         def inspec
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -78,10 +78,11 @@ module Inspec
         def resource_skipped_message
           @resource_skipped
         end
-        # This alias exists for backwards compatibility
-        # It is more idiomatic to have a `?` and `_message` methods
-        # TODO: Remove in InSpec 2.0
-        alias_method :resource_skipped, :resource_skipped_message
+
+        def resource_skipped
+          warn('[DEPRECATION] Use `resource_skipped_message` for the resource skipped message. This method will be removed in InSpec 2.0.')
+          @resource_skipped
+        end
 
         def fail_resource(message)
           @resource_failed = message
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -37,7 +37,7 @@ module Inspec
       Inspec::Resource.registry
     end
 
-    def __register(name, obj)
+    def __register(name, obj) # rubocop:disable Metrics/MethodLength
       cl = Class.new(obj) do
         def initialize(backend, name, *args)
           @resource_skipped = nil
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -44,7 +44,13 @@ module Inspec
           @__backend_runner__ = backend
           @__resource_name__ = name
           # call the resource initializer
-          super(*args)
+          begin
+            super(*args)
+          rescue Inspec::Exceptions::ResourceSkipped => e
+            skip_resource(e.message)
+          rescue Inspec::Exceptions::ResourceFailed => e
+            fail_resource(e.message)
+          end
         end
 
         def self.desc(description = nil)
@@ -63,6 +69,14 @@ module Inspec
 
         def skip_resource(message)
           @resource_skipped = message
+        end
+
+        def resource_failed
+          @resource_failed if defined?(@resource_failed)
+        end
+
+        def fail_resource(message)
+          @resource_failed = message
         end
 
         def inspec
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -39,9 +39,11 @@ module Inspec
 
     def __register(name, obj) # rubocop:disable Metrics/MethodLength
       cl = Class.new(obj) do
+        attr_reader :resource_exception_message
+
         def initialize(backend, name, *args)
-          @resource_skipped = nil
-          @resource_failed = nil
+          @resource_skipped = false
+          @resource_failed = false
 
           # attach the backend to this instance
           @__backend_runner__ = backend
@@ -68,31 +70,27 @@ module Inspec
         end
 
         def skip_resource(message)
-          @resource_skipped = message
+          @resource_skipped = true
+          @resource_exception_message = message
         end
 
         def resource_skipped?
-          !@resource_skipped.nil?
-        end
-
-        def resource_skipped_message
           @resource_skipped
         end
 
         def resource_skipped
-          warn('[DEPRECATION] Use `resource_skipped_message` for the resource skipped message. This method will be removed in InSpec 2.0.')
-          @resource_skipped
+          warn('[DEPRECATION] Use `resource_exception_message` for the resource skipped message. This method will be removed in InSpec 2.0.')
+          # Returning `nil` here to match previous behavior
+          return nil if @resource_skipped == false
+          @resource_exception_message
         end
 
         def fail_resource(message)
-          @resource_failed = message
+          @resource_failed = true
+          @resource_exception_message = message
         end
 
         def resource_failed?
-          !@resource_failed.nil?
-        end
-
-        def resource_failed_message
           @resource_failed
         end
 
--- a/test/unit/profiles/profile_context_test.rb
+++ b/test/unit/profiles/profile_context_test.rb
@@ -141,13 +141,19 @@ describe Inspec::ProfileContext do
       it 'alters controls when positive' do
         profile.load(if_false + control)
         get_checks.length.must_equal 1
+        get_checks[0][1][0].resource_skipped?.must_equal true
         get_checks[0][1][0].resource_skipped_message.must_equal 'Skipped control due to only_if condition.'
+        get_checks[0][1][0].resource_failed?.must_equal false
+        get_checks[0][1][0].resource_failed_message.must_be_nil
       end
 
       it 'alters non-controls when positive' do
         profile.load(if_false + describe)
         get_checks.length.must_equal 1
+        get_checks[0][1][0].resource_skipped?.must_equal true
         get_checks[0][1][0].resource_skipped_message.must_equal 'Skipped control due to only_if condition.'
+        get_checks[0][1][0].resource_failed?.must_equal false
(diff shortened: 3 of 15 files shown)


lib/inspec/plugins/resource.rb
```diff
@@ -65,6 +71,14 @@ def skip_resource(message)
           @resource_skipped = message
         end
 
+        def resource_failed
+          @resource_failed if defined?(@resource_failed)
```
adamleff: If `@resource_failed` is never assigned a value, it will return `nil` by default. Do we need this `defined?` check?

adamleff: If `@resource_failed` is never assigned a value, it will return `nil` by default. Do we need this `defined?` check?

adamleff (changes requested):
This is a great improvement. Small nitpick re: an unnecessary `defined?` check.

Also, I wonder if it would be cleaner to have something like this:

```ruby
def resource_failed?
  !@resource_failed.nil?
end

def resource_failure_message
  @resource_failed
end

def resource_skipped?
   # ... etc ...
end
```

It feels a bit better to me to have a boolean method for checking and only grab the message when necessary. This isn't a blocker though -- just wanted to open a conversation about it with you.

lib/inspec/exceptions.rb
```diff
@@ -5,6 +5,8 @@ module Inspec
   module Exceptions
     class AttributesFileDoesNotExist < ArgumentError; end
     class AttributesFileNotReadable < ArgumentError; end
+    class ResourceFailed < StandardError; end
     class SecretsBackendNotFound < ArgumentError; end
+    class ResourceSkipped < StandardError; end
```
adamleff: Can we keep this list alphabetized for consistency?

jerryaldrichiii: Hah! I originally had them called something else and took great care to make sure they were alphabetical. I renamed to both be `Resource` plus result so they would be together and forgot to sort.

Great catch @adamleff! Fixed in latest commit.

lib/inspec/plugins/resource.rb
jerryaldrichiii: I agree with you, I honestly just repeated what was there. I started down the path of correcting it and noticed that quite a few unrelated tests checked the value of `resource_skipped` and would need to be changed to `resource_skipped_message`.

The real worry here is that some of these tests are Unit tests for resources. Which got me thinking, if I modify `resource_skipped` to be a `?` and `_message` would that be a breaking change for some?

If so, should I:
  - Fix `resource_failed` only and have it be different than `resource_skipped` (and put in an issue for InSpec 2.0 to correct the difference)
  - Just leave it as is and put in an issue for InSpec 2.0 to fix both?

Thoughts?

adamleff: Just because something was there doesn't mean it's the right thing to do. :) We should always look for improvements while mucking around in the codebase.

The question to ask is: "are these methods part of an expected public API?" Is `resource_failed` used anywhere other than this class and its unit tests? Do the consumers of this method extend beyond the InSpec codebase (i.e. can a user call these methods from within their tests, and is it something we support)? And, if the answer is yes to either of these, can we do something to transition to what we want it to be?

If you agree that having `?` methods is a better fit here, then I'd like to encourage you to think about ways to get there without introducing breaking changes.  For example:

 - add the `?` methods
 - add the new `_message` methods
 - create method aliases for the non-`?` methods to the `_message` methods

I think you'll find it you do that, you gain a more idiomatic implementation while maintaining backward compatibilities (and we should add code comments explaining why the method aliases exist) and then we can remove those aliases in the future.

jerryaldrichiii: Excellent suggestions! I didn't think about an alias method. I pushed up another commit implementing what you recommended.

It also modifies other files that used `resource_skipped` in order to reduce work in the future when we deprecate. Let me know if I need to remove those changes.

jerryaldrichiii:
Looks like I'm failing Rubocop due to:
```
lib/inspec/plugins/resource.rb:40:5: C: Method has too many lines. [49/40]
    def __register(name, obj) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^
```

It's due to us defining a class in a method and the several methods I've added. @adamleff should we ignore this violation? If not, most of these methods can be defined on one line...but I don't like the way that looks. I'm open to other suggestions as well.

Also, thank you so much for all the effort you've put into these reviews. They've helped the project and me greatly.

adamleff:
@jerryaldrichiii just `rubocop:disable` it for that method. The length cops are super-annoying and don't really help, IMO... that's why they're generally disabled in chefstyle.  In fact, as you pointed out, it could encourage poorly-written code when people try to cram something into one line and lose readability.

lib/inspec/plugins/resource.rb
```diff
@@ -57,14 +67,34 @@ def self.example(example = nil)
...
+        def resource_skipped?
+          !@resource_skipped.nil?
+        end
+
+        def resource_skipped_message
+          @resource_skipped
+        end
+        # This alias exists for backwards compatibility
+        # It is more idiomatic to have a `?` and `_message` methods
+        # TODO: Remove in InSpec 2.0
```
adamleff: If you want to remove in 2.0, we need to change this to a deprecation warning now. :)

```
def resource_skipped
  warn '[DEPRECATED] Use `resource_skipped_message` for the resource skipped message. This method will be removed in InSpec 2.0.'
  @resource_skipped
end
```

... and then run through some manual tests to uncover if we are (or RSpec is) using it anywhere internally.

test/unit/profiles/profile_context_test.rb
```diff
@@ -141,13 +141,13 @@ def load(call)
       it 'alters controls when positive' do
         profile.load(if_false + control)
         get_checks.length.must_equal 1
-        get_checks[0][1][0].resource_skipped.must_equal 'Skipped control due to only_if condition.'
```
adamleff, line 144: Can I suggest a unit test that covers `resource_skipped` still, since we're leaving it in but deprecating it? That will help us in the future when we remove it, too.

test/unit/profiles/profile_context_test.rb
```diff
@@ -141,13 +141,13 @@ def load(call)
...
+        get_checks[0][1][0].resource_skipped_message.must_equal 'Skipped control due to only_if condition.'
       end
 
       it 'alters non-controls when positive' do
         profile.load(if_false + describe)
         get_checks.length.must_equal 1
-        get_checks[0][1][0].resource_skipped.must_equal 'Skipped control due to only_if condition.'
+        get_checks[0][1][0].resource_skipped_message.must_equal 'Skipped control due to only_if condition.'
```
adamleff: Can we include some `resource_skipped?` and `resource_failed?` tests in these, too? It appears we're missing coverage on those.

test/unit/profiles/profile_resource_exceptions_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    profile.runner_context.rules.values.each do |rule|
+      checks.push(Inspec::Rule.prepare_checks(rule))
+    end
+    checks
+  end
+
+  describe 'within initialize' do
+    it 'skips resource when `Inspec::Exceptions::ResourceSkipped` is raised' do
+      checks[0][0][1][0].resource_skipped_message.must_equal 'Skipping because reasons'
+      checks[0][0][1][0].resource_failed?.must_equal false
```
adamleff: Should we be checking `resource_skipped?` here too?

test/unit/profiles/profile_resource_exceptions_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    it 'fails resource when `Inspec::Exceptions::ResourceFailed` is raised' do
+      checks[1][0][1][0].resource_failed_message.must_equal 'Failing because reasons'
+      checks[1][0][1][0].resource_skipped?.must_equal false
```
adamleff: Should we be checking `resource_failed?` here too?

jerryaldrichiii:
Roger wilco. It has been fixed.

adamleff (changes requested):
This is a huge improvement, @jerryaldrichiii. I have some questions/suggestions. Reach out if anything is unclear.

lib/inspec/plugins/resource.rb
jerryaldrichiii: Will do :smile:

test/unit/profiles/profile_context_test.rb
jerryaldrichiii: I added a unit test to `profile_resource_exceptions_test` to cover it. :+1:

test/unit/profiles/profile_context_test.rb
jerryaldrichiii: Can do! I'll add test cases for each and their messages.

test/unit/profiles/profile_resource_exceptions_test.rb
jerryaldrichiii: Yup. :smile: I'll add test cases for each and their messages.

test/unit/profiles/profile_resource_exceptions_test.rb
jerryaldrichiii: Yup. :smile: I'll add test cases for each and their messages.

test/unit/profiles/profile_context_test.rb
jerryaldrichiii: https://github.com/chef/inspec/pull/2235/commits/8f9d9b1240bc97080af614bd71ea6e4537d55368#diff-d175555c026cd3a023e95ac973639b31R65

adamleff (approved):
@jerryaldrichiii this is great, thanks!

Before we merge, can you delete unneeded file `test/unit/mock/profiles/profile-with-resource-exceptions/libraries/.gitkeep` from this PR? Thanks!

jerryaldrichiii:
Done! Great catch :+1:

lib/inspec/plugins/resource.rb
```diff
@@ -57,12 +67,33 @@ def self.example(example = nil)
...
 
+        def skip_resource(message)
+          @resource_skipped = message
+        end
+
+        def resource_skipped?
+          !@resource_skipped.nil?
+        end
+
+        def resource_skipped_message
```
chris-rock: Why do we need to distinguish between failed and skip message. Can we have both at the same time?

lib/inspec/plugins/resource.rb
```diff
@@ -37,14 +37,24 @@ def __resource_registry
...
+
           # call the resource initializer
-          super(*args)
+          begin
+            super(*args)
+          rescue Inspec::Exceptions::ResourceSkipped => e
+            skip_resource(e.message)
+          rescue Inspec::Exceptions::ResourceFailed => e
+            fail_resource(e.message)
+          end
```
chris-rock, line 59: What happens in cases where we have an unhanded exception. I think we should handle those cases as well.

lib/inspec/plugins/resource.rb
```diff
@@ -37,14 +37,24 @@ def __resource_registry
...
+
           # attach the backend to this instance
           @__backend_runner__ = backend
           @__resource_name__ = name
+
           # call the resource initializer
-          super(*args)
+          begin
+            super(*args)
+          rescue Inspec::Exceptions::ResourceSkipped => e
```
chris-rock, line 55: It is great to see the separation here! Well done!

lib/inspec/plugins/resource.rb
```diff
@@ -37,14 +37,24 @@ def __resource_registry
       Inspec::Resource.registry
     end
 
-    def __register(name, obj)
+    def __register(name, obj) # rubocop:disable Metrics/MethodLength
       cl = Class.new(obj) do
         def initialize(backend, name, *args)
+          @resource_skipped = nil
```
chris-rock: Should we use an enum? Essentially we have stopped the execution and we did this for different reasons. The reason is important to say if this is okay or not.

chris-rock (commented):
@jerryaldrichiii This is a huge improvement and it is very beneficial. I just added a comment around msg handling and the how we can support adding more reasons easily in the future without breaking the api.

lib/inspec/plugins/resource.rb
jerryaldrichiii: Hmmm...you have a point. We shouldn't have both. Should we merge both into `resource_skipped_message`? Something else? Technically a failed resource isn't skipped since it tries to execute, but I don't want to rename it to `resource_message`. Thoughts?

adamleff: Part of the approach here was to ensure that we weren't botching something that was considered to be "public" or something RSpec was expecting. I'm +1 for consolidating as long as it doesn't break any backward compatibility.

I think calling it `exception_message` is probably sufficient.

lib/inspec/plugins/resource.rb
adamleff: Ugh, can't believe I missed this. An `else raise` is totally appropriate here.

adamleff: I take it back... Ruby will raise an exception if it's not one of the rescued classes. `else raise` is implicit. Given that we would just bubble up an exception anyways without this change (i.e. syntax issues in the initialize method), I think this is fine as-is.

adamleff: In other words, we currently unwind InSpec if there's a ruby syntax issue inside a resource. We should continue to do that in my opinion.

lib/inspec/plugins/resource.rb
jerryaldrichiii: I think merging into `exception_message` is a fine idea. Before, it was all contained in `resource_skipped` which should still work going forward.

jerryaldrichiii: Well fiddlesticks...
https://github.com/chef/inspec/blob/master/test/unit/resources/nginx_conf_test.rb#L100

This tests calls the instance variable `@resource_skipped` expecting a message. Since this test does it, is it safe to assume that any user could call the instance variable directly instead of the deprecated `resource_skipped` method?

If so it's no big, I'll just keep `@resource_skipped` and `@resource_failed` set to messages and evaluate in `resource_exception_message`...but I'd really like those to be bools.

adamleff: If you call an instance variable directly, you deserve breakage. 🙂  In no way can we support that as long-term behavior.

I'd fix the test and proceed with consolidation.

jerryaldrichiii: Agreed! Fixed.

chris-rock (approved):
Thank you for this improvement @jerryaldrichiii

PR #2239, opened 2017-10-11
Add inspec habitat profile setup command

adamleff:
Introduces a new `inspec habitat profile setup` command which will set up an existing profile repository with all the files necessary to build a Habitat package. This will prime a repository to be used by the Habitat Builder service.


chris-rock (approved):
Great improvement @adamleff

PR #2242, opened 2017-10-13
Squashed some unit test warnings

username-is-already-taken2:
Hi 

I've made a few changes to clear up some of the rspec messages when running through the tests.

I think a couple of the files weren't being executed due to them not having `_test` as a prefix so I just renamed those.

There looks to be a lot of warnings coming from the `parslet` gem which is fixed in a later build but you seem to be tided to the current version until the `toml` gem is updated.

Gary


adamleff (approved):
Superb, @username-is-already-taken2! Our unit test suite is pretty noisy, so anything we can do to eliminate unnecessary noise is excellent. Thank you!

chris-rock (approved):
Thank you @username-is-already-taken2 This change is very welcome!

PR #2244, opened 2017-10-16
Add missed 'html' to 'format' option explanation and arrange formatters in alphabetical order

strangeman:
### Before:
```
inspec help exec
...
      [--format=FORMAT]                            # Which formatter to use: cli, progress, documentation, json, json-min, junit
```

### After:
```
inspec help exec
...
      [--format=FORMAT]                            # Which formatter to use: cli, documentation, html, json, json-min, junit, progress
```

I found that `inspec` can produce reports in HTML format from StackOverflow thread, there is no info about it in official documentation. I think we should fix it.


adamleff (approved):
Hi, @strangeman! Thanks for your contribution to InSpec!

HTML output is currently provided "automatically" because RSpec provides it... it's not actually an InSpec feature itself. We may not support it as-is in the future, but for now, your change makes sense and there's no reason to hide that formatter.

Thank you for your PR!

PR #2246, opened 2017-10-17
Fix documentation of `shadow` resource

eramoto:
Describes actual behavior for the `shadow` InSpec audit resource.


adamleff:
Thanks for your PR, @eramoto! I thought there was more we could add to make this clearer, so I opened a new PR #2266 that supersedes this one. In addition, I've opened issue #2265 to add better FilterTable support to the `shadow` resource so it's consistent with our other resources that process tabular data, such as `passwd` and `crontab`

Thanks again for your PR!

PR #2250, opened 2017-10-17
Allow `inspec check` to ignore `only_if`

### LINKED ISSUE #2175: InSpec `check` and `only_if`
We currently execute all code in `only_if` statements during the `inspec check` operation. I have a feeling there may be little value in doing so, as the statement doesn't provide any useful input to the check command. It will execute the code in the block but it may also lead to executing code that doesn't fully compute (e.g. `nil` values because it doesn't talk to a target system). I wonder if we should ignore this block during the execution of `inspec check` to potentially prevent errors with dynamic computations.

On the upside, using `inspec check` in this case also catches potential errors when values aren't computed with the right data dynamically and may lead to better code in the long run. As a downside, users have to actually understand what is going on that is causing the failure and how to solve it, otherwise it only gets frustrating. A failure with dynamic computations at the moment (due to the nature of not catching raised failures) will lead `inspec check` to completely fail. It may be that a combination of catching failures and allowing the user to ignore the error may lead to relief as well.

### LINKED ISSUE #2237: syntax check on os.name.include?('Windows') failed but the code actually works
### Description
I have a only_if test to run a test only if the host is a windows server.
```
  only_if do
    os.name().include?('Windows Server')
  end
```
The code works, including running it in inspec shell:
```
inspec> os.name
=> "Windows 10 Pro"
inspec> os.name.include?("linux")
=> false
inspec> os.name.include?("Windows Server")
=> false
inspec> os.name.include?("Windows")
=> true
```
However, when running "inspec check", I received the following error:
```
>inspec check test.rb
test.rb:7:in `block (2 levels) in load_with_context': undefined method `include?' for nil:NilClass (NoMethodError)
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:103:in `only_if'
        from test.rb:6:in `block in load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:49:in `instance_eval'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:49:in `initialize'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/control_eval_context.rb:71:in `new'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/control_eval_context.rb:71:in `block (2 levels) in create'
        from test.rb:1:in `load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:146:in `instance_eval'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:146:in `load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:130:in `load_control_file'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:151:in `block in collect_tests'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:148:in `each'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:148:in `collect_tests'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:454:in `load_checks_params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:447:in `load_params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:141:in `params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:307:in `controls_count'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:278:in `check'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/cli.rb:69:in `check'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/bin/inspec:12:in `<top (required)>'
        from C:/opscode/inspec/bin/inspec:61:in `load'
        from C:/opscode/inspec/bin/inspec:61:in `<main>'
```
The content of the test.rb is shown below:
```
>type test.rb
control 'windows-feature' do
  impact 0.1
  title 'Collect the Windows Feature details on Windows Server'

  # the control will only run if the host is a windows server
  only_if do
    os.name().include?('Windows Server')
  end

  describe command('Get-WindowsFeature >> output.txt') do
  end
end
```
Briefly describe the issue and the expected behavior

As the code works, the check should pass.

```
>inspec version
1.40.0
```
```
inspec> os.name
=> "Windows 10 Pro"
```
### Replication Case
```
>type test.rb
control 'windows-feature' do
  impact 0.1
  title 'Collect the Windows Feature details on Windows Server'

  # the control will only run if the host is a windows server
  only_if do
    os.name().include?('Windows Server')
  end

  describe command('Get-WindowsFeature >> output.txt') do
  end
end
```

### Stacktrace
```
>inspec check test.rb
test.rb:7:in `block (2 levels) in load_with_context': undefined method `include?' for nil:NilClass (NoMethodError)
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:103:in `only_if'
        from test.rb:6:in `block in load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:49:in `instance_eval'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/rule.rb:49:in `initialize'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/control_eval_context.rb:71:in `new'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/control_eval_context.rb:71:in `block (2 levels) in create'
        from test.rb:1:in `load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:146:in `instance_eval'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:146:in `load_with_context'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile_context.rb:130:in `load_control_file'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:151:in `block in collect_tests'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:148:in `each'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:148:in `collect_tests'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:454:in `load_checks_params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:447:in `load_params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:141:in `params'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:307:in `controls_count'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/profile.rb:278:in `check'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/lib/inspec/cli.rb:69:in `check'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from C:/opscode/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.40.0/bin/inspec:12:in `<top (required)>'
        from C:/opscode/inspec/bin/inspec:61:in `load'
        from C:/opscode/inspec/bin/inspec:61:in `<main>'
```

jerryaldrichiii:
This fixes #2175
This fixes #2237

When using `inspec check` a mock Train backend is created. This means that the following would raise an error because `os.name` is `nil`

```
only_if { os.name.include?('anything') }
```

Since `inspec check` isn't concerned with the evaluation of `only_if` this skips those checks if the block given raises an error.

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -53,6 +53,7 @@ module Inspec
           @dependencies = dependencies
           @require_loader = require_loader
           @skip_file = false
+          @skip_only_if_eval = conf['check_mode'] || false
         end
 
         define_method :title do |arg|
@@ -70,7 +71,7 @@ module Inspec
         define_method :control do |*args, &block|
           id = args[0]
           opts = args[1] || {}
-          opts[:mock] = true if @backend.mock_transport?
+          opts[:skip_only_if_eval] = @skip_only_if_eval
           register_control(rule_class.new(id, profile_id, opts, &block))
         end
 
@@ -135,15 +136,9 @@ module Inspec
         define_method :only_if do |&block|
           return unless block
           return if @skip_file == true
+          return if @skip_only_if_eval == true
 
-          begin
-            return if block.yield == true
-          rescue
-            # Ignore error if a mock connection
-            # Example: `inspec check` with `only_if { os.name.include?('windows' }`
-            return if @backend.mock_transport?
-            raise
-          end
+          return if block.yield == true
 
           # Apply `set_skip_rule` for other rules in the same file
           profile_context_owner.rules.values.each do |r|
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -47,13 +47,13 @@ module Inspec
 
         attr_accessor :skip_file
 
-        def initialize(backend, conf, dependencies, require_loader)
+        def initialize(backend, conf, dependencies, require_loader, skip_only_if_eval)
           @backend = backend
           @conf = conf
           @dependencies = dependencies
           @require_loader = require_loader
           @skip_file = false
-          @skip_only_if_eval = conf['check_mode'] || false
+          @skip_only_if_eval = skip_only_if_eval
         end
 
         define_method :title do |arg|
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -138,11 +138,11 @@ module Inspec
 
           begin
             return if block.yield == true
-          rescue => e
+          rescue
             # Ignore error if a mock connection
             # Example: `inspec check` with `only_if { os.name.include?('windows' }`
             return if @backend.mock_transport?
-            raise e
+            raise
           end
 
           # Apply `set_skip_rule` for other rules in the same file
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -13,7 +13,7 @@ module Inspec
   # as the basic DSL of the control files (describe, control, title,
   # etc).
   #
-  class ControlEvalContext
+  class ControlEvalContext # rubocop:disable Metrics/ClassLength
     # Create the context for controls. This includes all components of the DSL,
     # including matchers and resources.
     #
@@ -70,6 +70,7 @@ module Inspec
         define_method :control do |*args, &block|
           id = args[0]
           opts = args[1] || {}
+          opts[:mock] = true if @backend.mock_transport?
           register_control(rule_class.new(id, profile_id, opts, &block))
         end
 
@@ -133,7 +134,16 @@ module Inspec
 
         define_method :only_if do |&block|
           return unless block
-          return if @skip_file == true || block.yield == true
+          return if @skip_file == true
+
+          begin
+            return if block.yield == true
+          rescue => e
+            # Ignore error if a mock connection
+            # Example: `inspec check` with `only_if { os.name.include?('windows' }`
+            return if @backend.mock_transport?
+            raise e
+          end
 
           # Apply `set_skip_rule` for other rules in the same file
           profile_context_owner.rules.values.each do |r|
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -78,7 +78,7 @@ module Inspec
       for_fetcher(fetcher, opts)
     end
 
-    attr_reader :source_reader, :backend, :runner_context
+    attr_reader :source_reader, :backend, :runner_context, :check_mode
     def_delegator :@source_reader, :tests
     def_delegator :@source_reader, :libraries
     def_delegator :@source_reader, :metadata
@@ -96,6 +96,7 @@ module Inspec
       @attr_values = options[:attributes]
       @tests_collected = false
       @libraries_loaded = false
+      @check_mode = options[:check_mode] || false
       Metadata.finalize(@source_reader.metadata, @profile_id, options)
 
       # if a backend has already been created, clone it so each profile has its own unique backend object
@@ -113,7 +114,7 @@ module Inspec
 
       @runner_context =
         options[:profile_context] ||
-        Inspec::ProfileContext.for_profile(self, @backend, @attr_values, options[:check_mode])
+        Inspec::ProfileContext.for_profile(self, @backend, @attr_values)
     end
 
     def name
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -113,7 +113,7 @@ module Inspec
 
       @runner_context =
         options[:profile_context] ||
-        Inspec::ProfileContext.for_profile(self, @backend, @attr_values)
+        Inspec::ProfileContext.for_profile(self, @backend, @attr_values, options[:check_mode])
     end
 
     def name
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -44,7 +44,7 @@ module Inspec
       @__checks = []
       @__skip_rule = nil
       @__merge_count = 0
-      @__mock = opts[:mock] || false
+      @__skip_only_if_eval = opts[:skip_only_if_eval]
 
       # evaluate the given definition
       instance_eval(&block) if block_given?
@@ -105,15 +105,9 @@ module Inspec
     # @return [nil]
     def only_if
       return unless block_given?
+      return if @__skip_only_if_eval == true
 
-      # Ignore error if a mock connection
-      # Example: `inspec check` with `only_if { os.name.include?('windows' }`
-      begin
-        @__skip_rule ||= !yield
-      rescue
-        return if @__mock == true
-        raise
-      end
+      @__skip_rule ||= !yield
     end
 
     # Describe will add one or more tests to this control. There is 2 ways
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -29,7 +29,7 @@ module Inspec
       @resource_dsl
     end
 
-    def initialize(id, profile_id, _opts, &block)
+    def initialize(id, profile_id, opts, &block)
       @impact = nil
       @title = nil
       @desc = nil
@@ -44,6 +44,7 @@ module Inspec
       @__checks = []
       @__skip_rule = nil
       @__merge_count = 0
+      @__mock = opts[:mock] || false
 
       # evaluate the given definition
       instance_eval(&block) if block_given?
@@ -104,7 +105,15 @@ module Inspec
     # @return [nil]
     def only_if
       return unless block_given?
-      @__skip_rule ||= !yield
+
+      # Ignore error if a mock connection
+      # Example: `inspec check` with `only_if { os.name.include?('windows' }`
+      begin
+        @__skip_rule ||= !yield
+      rescue => e
+        return if @__mock == true
+        raise e
+      end
     end
 
     # Describe will add one or more tests to this control. There is 2 ways
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -110,9 +110,9 @@ module Inspec
       # Example: `inspec check` with `only_if { os.name.include?('windows' }`
       begin
         @__skip_rule ||= !yield
-      rescue => e
+      rescue
         return if @__mock == true
-        raise e
+        raise
       end
     end
 
--- a/test/unit/profiles/control_eval_context_test.rb
+++ b/test/unit/profiles/control_eval_context_test.rb
@@ -31,7 +31,7 @@ EOF
   let(:eval_context) do
     c = Inspec::ControlEvalContext.create(profile_context, resource_dsl)
     # A lot of mocking here :(
-    c.new(backend, mock(), mock(), mock())
+    c.new(backend, {}, mock(), mock())
   end
 
   it 'accepts a context and a resource_dsl' do
@@ -40,7 +40,6 @@ EOF
 
   it 'provides rules with access to the given DSL' do
     profile_context.stubs(:current_load).returns({file: "<test content>"})
-    backend.stubs(:mock_transport?).returns(true)
     eval_context.instance_eval(control_content)
     profile_context.all_rules.each do |rule|
       # Turn each rule into an example group and run it, none of the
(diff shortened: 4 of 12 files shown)


lib/inspec/control_eval_context.rb
```diff
@@ -133,7 +134,16 @@ def to_s
 
         define_method :only_if do |&block|
           return unless block
-          return if @skip_file == true || block.yield == true
+          return if @skip_file == true
```
chris-rock, line 138: Can't we just set `@skip_file` to false if we are in `@backend.mock_transport?`

jerryaldrichiii: I don't think we can because Unit tests also use a mock backend. Doing so would make it impossible to test `only_if`.

lib/inspec/control_eval_context.rb
```diff
@@ -133,7 +134,16 @@ def to_s
...
+
+          begin
+            return if block.yield == true
+          rescue => e
+            # Ignore error if a mock connection
+            # Example: `inspec check` with `only_if { os.name.include?('windows' }`
+            return if @backend.mock_transport?
+            raise e
```
adamleff: You can eliminate the `e` here - `raise` within a `rescue` will re-raise the original exception.

lib/inspec/rule.rb
```diff
@@ -104,7 +105,15 @@ def source_file
...
       return unless block_given?
-      @__skip_rule ||= !yield
+
+      # Ignore error if a mock connection
+      # Example: `inspec check` with `only_if { os.name.include?('windows' }`
+      begin
+        @__skip_rule ||= !yield
+      rescue => e
+        return if @__mock == true
+        raise e
```
adamleff: You can eliminate the `e` here - `raise` within a `rescue` will re-raise the original exception.

adamleff (commented):
Just two quick clean-up items and this looks good.

I wonder if there's a better way to do this without having to pass around data as to whether we're mocking or not. We already tightly-couple ourselves to the mock backend for unit tests, so I'm afraid of future edge cases.

Would it make sense to introduce the concept of an "operating mode" class var we can set and query? For example, in the Thor method when `inspec check` is run:

```ruby
InSpec::OperatingMode.set_mode(:check)
```

... and we can flat-out skip only_if evaluation with something like:

```ruby
return if InSpec::OperatingMode.is_check?
# or
return if InSpec::OperatingMode.current_mode == :check
```

@jerryaldrichiii @arlimus @chris-rock -- thoughts?

jerryaldrichiii:
Thanks @adamleff didn't know that about `raise` in a rescue.

In regards to `OperatingMode`, I'm all for it. I think that is a cleaner approach.

chris-rock:
The more I think about it the more I am not sure we are solving the right thing here. We have two problems that I see here:

- 1. os.name is set to nil and crashed the execution
- 2. `only_if { os.name.include?('anything') }` any only if can crash

The first issue can also happen outside of the `only_if` block, therefore the proposed solution is not helping for those cases. An alternative would be to set the operating system in train if we are using the mock backend

For the second one. We could try to catch any exceptions for `only_if` blocks, since this can also happen during runtime. As proposed by @adamleff What do you think?

adamleff:
@chris-rock I think the `os.name` being nil and your `only_if` exception issue are symptoms of a larger issue... that issue is that we evaluate `only_if` blocks during an `inspec check`. In all other phases of InSpec operation, everything is fine. But given that `only_if` can contain other resources that could perhaps interact with a system not expected to be tested (i.e. the person's workstation), I think it's safe for us to declare that `only_if` blocks should never be eval'd during an `inspec check`

That's why I think we solve this with exposing a way for InSpec to communicate that we're in `check` mode... rather than catching exceptions. I'd much rather us never eval an `only_if` block unless we need to. That will solve both problems.

adamleff:
Hey @jerryaldrichiii! @chris-rock and I met today to discuss this change.

First off, we agree that we should probably add a PR to Train to correctly build an `os` object that responds to all the expected methods, and return "mock" or "unknown" such that issues like the one described in this PR don't happen. I'll open an issue on that in case you want to try your hand at it.

Secondly, we'd like to take a different approach with this PR. Essentially, rather than tightly coupling ourselves to the mock backend, we'd like to pass through a flag that indicates we're in "check mode" and then, if so, tell the `ControlEvalContext` to short-circuit the `only_if` eval. Essentially, it will look like this:

 - CLI: #check should plumb a `check_mode = true` option and pass it to Inspec::Profile
 - Inspec::ProfileContext.for_profile will need to accept a new parameter indicating whether it's in check mode, and Inspec::Profile needs to pass it when creating the context
 - Inspec::ControlEvalContext should accept a new "skip_only_if_eval" option. Inspec::ProfileContext should pass it if check_mode is true
 - The eval of the only_if should now return early if "skip_only_if_eval" rather than catching an exception and determining whether the mock backend is in use

Let me know if you have any questions.

lib/inspec/control_eval_context.rb
```diff
@@ -53,6 +53,7 @@ def initialize(backend, conf, dependencies, require_loader)
           @dependencies = dependencies
           @require_loader = require_loader
           @skip_file = false
+          @skip_only_if_eval = conf['check_mode'] || false
```
adamleff: I'd like to draw a line at what things care about the concept of "check mode". In my opinion, the line needs to be drawn at the profile. Controls (and their evaluation contexts) should not care about the mode in which InSpec was executed but rather whether they should care about only_if or not.

Can we find a way to gracefully pass a bool for skip_only_if_eval from the profile to the control instead of passing check_mode here?

adamleff (changes requested):

lib/inspec/control_eval_context.rb
jerryaldrichiii: What are your thought about just passing it all the way from the CLI, e.g. `o[:skip_only_if_eval] = true` vs `o[:check_mode]`?

jerryaldrichiii: What are your thoughts about just passing it all the way from the CLI, e.g. `o[:skip_only_if_eval] = true` vs `o[:check_mode] = true`?

adamleff: While you can claim YAGNI, what if something else in the future needs to care about whether we're in check_mode or not? And should the CLI really be directing how a control is evaluated?

The CLI knows it's in check mode, the profile needs to know it's in check mode (because it's the thing being checked), and then the profile should use that information to determine whether or not certain activities should take place. I still believe the line I drew and described above is the right separation of duties and logic.

jerryaldrichiii: I agree, I moved the line up to Profile and then passed `skip_only_if_eval` it to the `ControlEvalContext`

adamleff (approved):

lib/inspec/profile.rb
```diff
@@ -96,6 +96,7 @@ def initialize(source_reader, options = {})
       @attr_values = options[:attributes]
       @tests_collected = false
       @libraries_loaded = false
+      @check_mode = options[:check_mode] || false
```
adamleff, line 99: Little Ruby nugget - careful with `||` and default booleans. For example, if you wanted the default to be `true` and you wrote this like:

```ruby
@check_mode = options[:check_mode] || true
```

... it could never be false! If the user passed in `false` as the value in `options[:check_mode]`, it would fail the left side of the `||` and then right side would evaluate.

I tend to use `#fetch` on `Hash` for this instead, just in case:

```ruby
@check_mode = options.fetch(:check_mode, false)
```

test/unit/profiles/control_eval_context_test.rb
```diff
@@ -31,7 +31,7 @@ def foobar
   let(:eval_context) do
     c = Inspec::ControlEvalContext.create(profile_context, resource_dsl)
     # A lot of mocking here :(
-    c.new(backend, mock(), mock(), mock())
+    c.new(backend, {}, mock(), mock(), false)
```
adamleff: Can we add a comment here that enumerates the options ControlEvalContext expects so the next reader doesn't have to hop back to understand? I think that would help greatly.

adamleff (approved):
One request-for-comment, but otherwise, this is good to go.

lib/inspec/profile.rb
jerryaldrichiii: Ah, thanks for the tip. Never thought of that.

test/unit/profiles/control_eval_context_test.rb
jerryaldrichiii: Great idea, I pushed a commit that list the options that are mocked. Let me know if I need to modify the wording.

arlimus (approved):
Fantastic fix, thank you @jerryaldrichiii !!

PR #2253, opened 2017-10-18
Add new Windows helpers to find which release is being tested.

mattray:

### DIFF
--- a/docs/resources/os.md.erb
+++ b/docs/resources/os.md.erb
@@ -138,3 +138,24 @@ For example, both of the following tests should have the same result:
         its('processes') { should include 'xinetd' }
       end
     end
+
+## os.windows? helpers
+
+Use `os.windows*` helpers to enable more granular testing of Windows versions.
+
+* `windows?`
+* `windows_7?`
+* `windows_8?`
+* `windows_8_1?`
+* `windows_10?`
+* `windows_server_2008?`
+* `windows_server_2008_r2?`
+* `windows_server_2012?`
+* `windows_server_2012_r2?`
+* `windows_server_2016?`
+
+To test for Windows 2008R2 use:
+
+    describe os.windows_server_2008_r2? do
+       it { should eq true }
+    end
--- a/lib/resources/os.rb
+++ b/lib/resources/os.rb
@@ -58,54 +58,47 @@ module Inspec::Resources
     # Windows release numbers from https://en.wikipedia.org/wiki/Ver_(command)
     # 6.0.6001 6.0.6002
     def windows_server_2008?
-      return true if release =~ /6.0.600/
-      false
+      windows? and release =~ /6.0.600/ ? true : false
     end
 
     # 6.1.7600 6.1.7600.16385 6.1.7601
     def windows_7?
-      return true if release =~ /6.1.760/
-      false
+      windows? and release =~ /6.1.760/ ? true : false
     end
 
     # 6.1.7600 6.1.7600.16385 6.1.7601
     def windows_server_2008_r2?
-      windows_7?
+      windows? and release =~ /6.1.760/ ? true : false
     end
 
     # 6.2.9200 6.2.9200.16384
     def windows_8?
-      return true if release =~ /6.2.9200/
-      false
+      windows? and release =~ /6.2.9200/ ? true : false
     end
 
     # 6.2.9200
     def windows_server_2012?
-      return true if release =~ /6.2.9200/
-      false
+      windows? and release =~ /6.2.9200/ ? true : false
     end
 
     # 6.3.9600
     def windows_8_1?
-      return true if release =~ /6.3.9600/
-      false
+      windows? and release =~ /6.3.9600/ ? true : false
     end
 
     # 6.3.9600
     def windows_server_2012_r2?
-      windows_8_1?
+      windows? and release =~ /6.3.9600/ ? true : false
     end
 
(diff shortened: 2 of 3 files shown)


docs/resources/os.md.erb
```diff
@@ -138,3 +138,24 @@ For example, both of the following tests should have the same result:
         its('processes') { should include 'xinetd' }
       end
     end
+
+## os.windows? helpers
+
+Use `os.windows*` helpers to enable more granular testing of Windows versions.
```
chris-rock, line 144: It does not feel right to me to attach all those helpers to the os resource. It scope should be very limited.

lib/resources/os.rb
```diff
@@ -54,5 +54,58 @@ def params
...
+
+    # 6.3.9600
+    def windows_8_1?
+      return true if release =~ /6.3.9600/
+      false
+    end
+
+    # 6.3.9600
+    def windows_server_2012_r2?
+      windows_8_1?
```
chris-rock: This is not a safe way to implement this. I would rely on https://github.com/chef/train/blob/master/lib/train/extras/os_detect_windows.rb#L51

docs/resources/os.md.erb
chris-rock: Would it make sense to introduce a `windows` resource? I am not sure. What do you think @adamleff

chris-rock (commented):
Thank you @mattray Can you share the usecase why you need additional helper instead of relying on pure os versions and name? `os.name` includes that information

docs/resources/os.md.erb
adamleff: I don't think that feels right, either. Are we then going to need to have a `centos` and `linux` resource also?

adamleff (commented):
I'm personally 👎  on this change. I agree that it would likely make writing tests and custom resources for windows a bit cleaner, but I believe the data is already available in `os[:name]` and it feels like a slippery slope, requiring us to implement other OS-release-specific methods, such as `os.centos_7?` and similar which doesn't feel right to me.

I'd be in support of exposing an "OS short-name" or something similar on the `os` resource that looks like the method names provided though...

`windows_server_2012_r2`
`centos_7`

... so you could do things like:

```ruby
only_if { os.short_name == 'windows_server_2012_r2' }
```

... etc... but I think that would need to be implemented in Train, not in InSpec.

docs/resources/os.md.erb
mattray: Here is an example of the windows helpers in action.
https://github.com/mattray/inspec-profile-wannacry-exploit/blob/use_windows_helpers/controls/hotfixes.rb#L60

There are lots of security controls that are only applicable to certain versions of Windows. The intention is to provide a user-friendly means to know which version of Windows you're working with without having to parse the name or know the arcane versioning.

mattray:
I'm happy to move the code into train if that's where it belongs. As an author of controls and profiles I strongly want to have syntactic sugar available to me when dealing with operating systems. The whole point of InSpec is to make writing audits accessible, and I feel new developers would rather see

```os.windows_server_2016?```
than
```os.short_name == 'windows_server_2012_r2'```

Personally, I want every helper available through Chef-Sugar available in InSpec, because expecting an auditor to go and parse through OS internals seems undelightful.

File: lib/resources/os.rb (hunk 2 above)
Comment (mattray) [reply lib/resources/os.rb]: I was using the fact that windows_8_1 and 2012_r2 are the same to reduce the code footprint. I can revert it to duplicated code if that's what you mean.

Comment (mattray) [reply lib/resources/os.rb]: I should probably also add a windows? check to all of these too.

File: lib/resources/os.rb [hunk 3]
```diff
@@ -58,54 +58,47 @@ def to_s
     # Windows release numbers from https://en.wikipedia.org/wiki/Ver_(command)
     # 6.0.6001 6.0.6002
     def windows_server_2008?
-      return true if release =~ /6.0.600/
-      false
+      windows? and release =~ /6.0.600/ ? true : false
```
adamleff: We should use `&&` instead of `and` here. And we can probably eliminate the ternary with something like this:

```ruby
windows? && !release.match(/6.0.600/).nil?
```

adamleff:
I thought about this change some more and I'm going to change my vote from a 👎  to neither 👍  or 👎 .

In Linux variants, it's pretty common that if something applies to CentOS, it usually applies to all versions of CentOS. There are cases where that's not true, like the init system, but that's relatively easy to use the `os[:release]` value to make that determination, and it doesn't happen often enough to warrant having `os.centos6?` and `os.centos7?`helpers.

In Windows-land, the differences between 2008r2 and 2012 are large enough that it's likely these helpers will be used more often than an `os.centos6?` helper would be. That said, I personally do not see any difference in readability between:

```ruby
os.short_name == 'windows_server_2012_r2'
```

... and ...

```ruby
os.windows_server_2012_r2?
```

I'm :+1: on the perceived UX benefit, :-1: on the additional maintenance of adding these / keeping them up-to-date and having an inconsistency with what helpers we offer for what operating systems.

chris-rock:
> os.windows_server_2016?
than
os.short_name == 'windows_server_2012_r2'

In https://github.com/chef/train/pull/192 we are going to improve the `os.name` experience and make it more consistent across all operating systems. Therefore we are going to have a `os.name = 'windows_server_2012_r2_standard'` which is also consistent across all operating systems.

Therefore I vote against this change.

mattray:
Closing this to use ```os.name``` from https://github.com/chef/train/pull/192 so matchers will take the form of 
```os.name == 'windows_server_2008'```

PR #2257, opened 2017-10-18
mount resource: fix for Device-/Sharenames and Mountpoints including …

mgrobelin:
…whitespaces

Device-/Sharenames and Mountpoints on Linux may include whitespaces (\040), e.g. /etc/fstab entry like:

```
//fileserver.corp.internal/Research\040&\040Development /mnt/Research\040&\040Development cifs OTHER_OPTS
```

... results in a mount line like:

```
//fileserver.corp.internal/Research & Development on /mnt/Research & Development type cifs (OTHER_OPTS)
```

The Linux mount command replaces \040 with whitspace automatically, so this should be tributed.

I used a control like this:

```
    describe mount('/mnt/Research & Development') do
      it { should be_mounted }
      its('device') { should eq  '//fileserver.corp.internal/Research & Development' }
    end
```

Before:

```
  ×  whitespaces-1: Mount with whitespace within sharename and mountpoint. (1 failed)
     ✔  Mount /mnt/Research & Development should be mounted
     ×  Mount /mnt/Research & Development device should eq "//fileserver.corp.internal/Research & Development"

     expected: "//fileserver.corp.internal/Research & Development"
          got: "//fileserver.corp.internal/operational/Research"

     (compared using ==)
```

After:

```
  ✔  whitespaces-01: Mount with whitespace within sharename and mountpoint.
     ✔  Mount /mnt/Research & Development should be mounted
     ✔  Mount /mnt/Research & Development device should eq "//fileserver.corp.internal/Research & Development"
```

Signed-off-by: Markus Grobelin <grobi@koppzu.de>

### DIFF
--- a/lib/utils/parser.rb
+++ b/lib/utils/parser.rb
@@ -73,7 +73,7 @@ module LinuxMountParser
       type_split = mount_line.split(' type ')
       fs_path = type_split[0]
       other_opts = type_split[1]
-      fs, path = fs_path.match(/^(.+?)\son\s(\/.+?)$/).captures
+      fs, path = fs_path.match(%r{^(.+?)\son\s(/.+?)$}).captures
       mount = [fs, 'on', path, 'type']
       mount.concat other_opts.scan(/\S+/)
     else
@@ -106,9 +106,9 @@ module LinuxMountParser
   end
 
   # Device-/Sharename or Mountpoint includes whitespaces?
-  def includes_whitespaces? mount_line
+  def includes_whitespaces?(mount_line)
     ws = mount_line.match(/^(.+)\son\s(.+)\stype\s.*$/)
-    ws.captures[0].include? " " or ws.captures[1].include? " "
+    ws.captures[0].include? ' ' or ws.captures[1].include? ' '
   end
 end
 
--- a/lib/utils/parser.rb
+++ b/lib/utils/parser.rb
@@ -67,7 +67,7 @@ module LinuxMountParser
   # this parses the output of mount command (only tested on linux)
   # this method expects only one line of the mount output
   def parse_mount_options(mount_line, compatibility = false)
-    if includes_whitespaces? mount_line
+    if includes_whitespaces?(mount_line)
       # Device-/Sharenames and Mountpoints including whitespaces require special treatment:
       # We use the keyword ' type ' to split up and rebuild the desired array of fields
       type_split = mount_line.split(' type ')
--- a/lib/utils/parser.rb
+++ b/lib/utils/parser.rb
@@ -75,7 +75,7 @@ module LinuxMountParser
       other_opts = type_split[1]
       fs, path = fs_path.match(%r{^(.+?)\son\s(/.+?)$}).captures
       mount = [fs, 'on', path, 'type']
-      mount.concat other_opts.scan(/\S+/)
+      mount.concat(other_opts.scan(/\S+/))
     else
       # ... otherwise we just split the fields by whitespaces
       mount = mount_line.scan(/\S+/)
@@ -108,7 +108,7 @@ module LinuxMountParser
   # Device-/Sharename or Mountpoint includes whitespaces?
   def includes_whitespaces?(mount_line)
     ws = mount_line.match(/^(.+)\son\s(.+)\stype\s.*$/)
-    ws.captures[0].include? ' ' or ws.captures[1].include? ' '
+    ws.captures[0].include?(' ') or ws.captures[1].include?(' ')
   end
 end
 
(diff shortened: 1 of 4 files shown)


mgrobelin:
Please note, this PR do not address the case, when the Share-/Devicename and/or Mountpoint includes words like ` on ` or ` type `, e.g.: a `/etc/fstab` entry like:
```
/dev/sda1 /mnt/type\040on\040fire xfs [...]
```
Will most likely render wrong results / unexpected behavoir.

lib/utils/parser.rb
```diff
@@ -67,7 +67,19 @@ module LinuxMountParser
...
-    mount = mount_line.scan(/\S+/)
+    if includes_whitespaces? mount_line
+      # Device-/Sharenames and Mountpoints including whitespaces require special treatment:
+      # We use the keyword ' type ' to split up and rebuild the desired array of fields
+      type_split = mount_line.split(' type ')
+      fs_path = type_split[0]
+      other_opts = type_split[1]
+      fs, path = fs_path.match(%r{^(.+?)\son\s(/.+?)$}).captures
+      mount = [fs, 'on', path, 'type']
+      mount.concat other_opts.scan(/\S+/)
```
adamleff: We try to standardize on wrapping all method arguments in parentheses. Can you change this to:

```ruby
mount.concat(other_opts.scan(/\S+/))
```

... so it's clearer? Thanks!

lib/utils/parser.rb
```diff
@@ -92,6 +104,12 @@ def parse_mount_options(mount_line, compatibility = false)
 
     mount_options
   end
+
+  # Device-/Sharename or Mountpoint includes whitespaces?
+  def includes_whitespaces?(mount_line)
+    ws = mount_line.match(/^(.+)\son\s(.+)\stype\s.*$/)
+    ws.captures[0].include? ' ' or ws.captures[1].include? ' '
```
adamleff: Same here.. each of the `include?` method calls should wrap their arguments with parentheses, please. :)

lib/utils/parser.rb
```diff
@@ -67,7 +67,19 @@ module LinuxMountParser
   # this parses the output of mount command (only tested on linux)
   # this method expects only one line of the mount output
   def parse_mount_options(mount_line, compatibility = false)
-    mount = mount_line.scan(/\S+/)
+    if includes_whitespaces? mount_line
```
adamleff: As with the other review comments, please wrap arguments with parentheses:

```ruby
if includes_whitespaces?(mount_line)
```

lib/utils/parser.rb
```diff
@@ -67,7 +67,19 @@ module LinuxMountParser
...
+      # Device-/Sharenames and Mountpoints including whitespaces require special treatment:
+      # We use the keyword ' type ' to split up and rebuild the desired array of fields
+      type_split = mount_line.split(' type ')
```
adamleff: You might be able to use `rindex` to get the right-most ` type ` to avoid picking up a share name with the word ` type ` in it with spaces:

```ruby
type_index = mount_line.rindex(' type ')
fs_path = mount_line[0...type_index]
other_opts_index = type_index + 6 # since ' type ' is 6 chars
other_opts = mount_line[other_opts_index..-1]
# ... etc ...
```

I think this might decently well, but would need to be tested. Worth playing around with a bit?

adamleff (changes requested):
Looking pretty good! Just a couple of code-cleanup items and a suggestion on a potential way around the "type" being in device/share names.

lib/utils/parser.rb
mgrobelin: Definitely worth, but I left it out for now due to high complexity. Maybe later, but I cannot promise due to endofyear-crunch.

If one would address this edge-case, it might be clever to do the same for `BsdMountParser`, right? What would be your suggestion to keep it DRY?

adamleff: I'm OK leaving it out for this PR and we can circle back to it if someone raises a bug or if you feel like coming back for a second iteration. :)

I would definitely DRY it up if possible as long as it doesn't make it too complicated. I'm happy to help as needed!

lib/utils/parser.rb
adamleff: Missed one set of parentheses here and then I think we're good!

adamleff (changes requested):
@mgrobelin just one more set of parentheses was missed in your last commit from my initial review, and then I think we can move forward with this. Thanks!

adamleff (approved):
Thanks for your work on this fix, @mgrobelin!

chris-rock (approved):
Thank you @mgrobelin and @adamleff

PR #2259, opened 2017-10-18
Include ref when writing out inspec control objects

arlimus:
just like tags for all auto-generation and api-driven development :)


aaronlippold:
Just for rounding it out do we want to add 'remediation' which should fully flush out the object with respect to what is documented and supported.

adamleff:
@aaronlippold that's outside the scope of this PR, which simply ensures that refs are properly output when writing out a control object to disk. Adding "remediation" as a first-class citizen to the DSL would warrant an RFC for discussion. Right now, "remediation" is used as an example of a ref key, but isn't part of the DSL.

adamleff (approved):
Looks good, @arlimus!

PR #2262, opened 2017-10-23
Corrected name to be service_name

lhasadreams:
Found while following the docs to try out the "Profile Files" feature.
Tested with the fix


adamleff (approved):
Thanks, @lhasadreams! This will get updated on the website sometime Thursday when we do our weekly InSpec release. Thanks for contributing!

PR #2264, opened 2017-10-24
Fix regression when uploading compliance profiles

jerryaldrichiii:
When passing options to `Train` we need to ensure we do not pass the `'target'` key. See: https://github.com/chef/inspec/pull/1646

NOTE: When running `inspec exec` the key is `:target` so it is not removed, but since `options[:backend]` already exists this is not an issue.


adamleff (approved):
We will eventually need to circle back on how we handle options hashes like this, but this is the right solution given the current design. Thank you @jerryaldrichiii!

PR #2266, opened 2017-10-24
Fix the docs for the shadow resource

adamleff:
Inspired by the work in PR #2246

### DIFF
--- a/docs/resources/shadow.md.erb
+++ b/docs/resources/shadow.md.erb
@@ -29,17 +29,31 @@ A `shadow` resource block declares one (or more) users and associated user infor
       its('users') { should_not include 'forbidden_user' }
     end
 
-or with a filter:
+or with a single query:
 
-    describe shadow.uid(filter) do
-      its('users') { should cmp 'root' }
+    describe shadow.users('root') do
       its('count') { should eq 1 }
     end
 
-where
+or with a filter:
 
-* `homes`, `gids`, `passwords`, `shells`, `uids`, and `users` are valid accessors for `passwd`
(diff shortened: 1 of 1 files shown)


eramoto:
Thank you for your work on this documentation, this PR looks good to me.

BTW, could you please add fixes, which is to fix typo and remove information for /etc/passwd included in PR #2246, into this PR? Or, is there another PR having these fixes?

chris-rock (approved):
Thank you @eramoto and @adamleff for improving the documentation. It is super helpful!

adamleff:
@eramoto I will fix the typo that you corrected, thanks for pointing that out.

I'm afraid I don't understand what reference to `/etc/passwd` you're referring to. I don't see any reference to `passwd` in the proposed document, and the `passwords` property is still a valid property for `/etc/shadow`

adamleff:
@eramoto the `TThis` typo is now fixed. I'm going to merge this but I'm happy to help fix anything else you can identify as being incorrect, or I'll be happy to review another PR. Thank you!

eramoto:
Hi @adamleff ,

I apology for the insufficiency of explanation.

The following is the quote from the man-page passwd(5) (see also shadow(5)):
```
If the password field is a lower-case "x", then the encrypted password is actually stored in the shadow(5) file instead; there must be a corresponding line in the /etc/shadow file, or else the user account is invalid. 
```

`/etc/passwd` can have the letter x in the password field and `/etc/shadow` does not have it. So I think the following (the `passwords` description of the `shadow` resource) describes about `/etc/passwd`, and  is copied from the `passwords` property description of the `passwd` resource without any change.

* [#2246 L117](https://github.com/chef/inspec/pull/2246/files#diff-74ad57fe85d35cc8899975cf20666e1cL117)

or
* * In the `/etc/shadow` file, as indicated by the letter x (`x`)

adamleff:
@eramoto thanks for the clarification. I have pushed up #2269 that better clarifies how to use the `passwords` property and eliminates the reference to an `x` string which does not apply. Thanks for your help in getting our documentation to be more accurate.

PR #2267, opened 2017-10-26
Fix sorting of resources on website

adamleff:
Not sure how this ever worked without us force-sorting, but the
list of resources is now guaranteed to be alpha-sorted.


arlimus:
Sweeeeet thank you Adam!!

PR #2268, opened 2017-10-26
inspec compliance login: Ensure supplied server has a proper URI scheme

jerryaldrichiii:
An issue was discovered whereby a user logging in with `inspec compliance login` did not specify `https://` in their `SERVER` argument. This causes https://github.com/chef/inspec/blob/master/lib/fetchers/url.rb#L146 to fail because https://github.com/ruby/ruby/blob/trunk/lib/open-uri.rb#L33 doesn't interpret the option as a URL.

This adds `https://` to the front of `options['server']` if the URI doesn't contain a scheme, thus fixing the issue.


adamleff (approved):

chris-rock (approved):
Thank you @jerryaldrichiii

PR #2270, opened 2017-10-27
Use names when quering for containers

szymonpk:
Name does not work.

I am not sure if it changes across various docker versions or it was a typo from beginning. Anyway 'names' works now.

Obvious fix.


chris-rock (approved):
Thank you @szymonpk for highlighting this bug in our documentation. Docker only exposes `names` therefore each container can have multiple names.

adamleff (approved):
Thanks @szymonpk!

PR #2271, opened 2017-10-28
Wrapper profile and dependency with same filename in controls directory breaks inspec check

brianreumere:
### Description

When a wrapper profile has a filename in `controls/` that matches that of the dependency profile, running `inspec check` on the wrapper profile tries to read source code for the rule from its own `controls/` directory and fails.

Inspec 1.43.5
macOS 10.13

### Replication Case

Install the cis-ubuntu14.04lts-level1 profile to your Automate server. Create a wrapper profile with the following contents (probably have to adjust the compliance path).

inspec.yml:

```name: wrapper-cis
title: Wrapper
summary: Wrapper
version: 0.1.0

maintainer: The Authors
copyright: The Authors
copyright_email: example@example.com
license: Apache-2.0

supports:
  - os-family: linux
depends:
  - name: cis-ubuntu14.04lts-level1
    compliance: bcurran/cis-ubuntu14.04lts-level1
    version: 1.0.0-2
```

controls/translated-controls.rb:

```# encoding: utf-8
# copyright: 2017, The Authors

require_controls 'cis-ubuntu14.04lts-level1' do
  control 'xccdf_org.cisecurity.benchmarks_rule_7.4.4_Create_etchosts.deny'
end
```

Run `inspec check` on your wrapper profile.

### Possible Solutions

The easiest workaround is just renaming the file in `controls/` so it doesn't match anything in the dependency profile. This seems to be happening because nothing in `profile.rb` recognizes that the `location` variable in the `load_rule` method is referring to a file in the dependency profile (not a file in the wrapper profile) and when `Inspec::MethodSource.code_at(location, source_reader)` is called, it bails because `src = source_reader.tests[ref]` in `method_source.rb` returns some code, but not the code from the dependency profile, so it tries to read a line number that doesn't exist.

This doesn't happen with local dependencies because `location` in the `load_rule` method has an absolute path and doesn't match `source_reader.tests[ref]`:

```
[1] pry(Inspec::MethodSource)> location
=> {:ref=>"/Users/bcurran/git/inspec/profiles/cis-ubuntu14.04lts-level1/controls/translated-controls.rb", :line=>1011}
[2] pry(Inspec::MethodSource)> source_reader.tests
=> {"controls/translated-controls.rb"=>
  "# encoding: utf-8\n# copyright: 2017, The Authors\n\nrequire_controls 'cis-ubuntu14.04lts-level1' do\n  control 'xccdf_org.cisecurity.benchmarks_rule_7.4.4_Create_etchosts.deny'\n  control 'asdf'\nend\n"}
```

I'm not very well-versed in Ruby, but I attempted a fix here that removes `code` from the `controls[id]` hash and sets it conditionally immediately afterward, only using `code_at` if the current profile `name` matches the `profile_id` of the current rule, and sets it to `''` otherwise.

If this needs unit tests or anything let me know and I can take a shot at those. (Or if this is totally off-track and I'm missing something, let me know and I'll just file an issue 🙂.)

### Stacktrace

```
/opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/method_source.rb:23:in `rescue in code_at': Could not parse source at controls/translated-controls.rb:1011: unexpected $end (MethodSource::SourceNotFoundError)
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/method_source.rb:11:in `code_at'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:495:in `load_rule'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:472:in `block in load_checks_params'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:469:in `each'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:469:in `load_checks_params'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:458:in `load_params'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:152:in `params'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:318:in `controls_count'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/profile.rb:289:in `check'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/lib/inspec/cli.rb:69:in `check'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from /opt/inspec/embedded/lib/ruby/gems/2.3.0/gems/inspec-1.43.5/bin/inspec:12:in `<top (required)>'
        from /usr/local/bin/inspec:23:in `load'
        from /usr/local/bin/inspec:23:in `<main>'
```

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -492,6 +492,12 @@ module Inspec
         code: Inspec::MethodSource.code_at(location, source_reader),
         source_location: location,
       }
+      # Don't try to read code if the control is inherited from another profile
+      if rule.instance_variable_get(:@__profile_id) == name
+        controls[id][:code] = Inspec::MethodSource.code_at(location, source_reader)
+      else
+        controls[id][:code] = ''
+      end
 
       groups[file] ||= {
         title: rule.instance_variable_get(:@__group_title),
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -489,7 +489,6 @@ module Inspec
         refs: rule.ref,
         tags: rule.tag,
         checks: Inspec::Rule.checks(rule),
-        code: Inspec::MethodSource.code_at(location, source_reader),
         source_location: location,
       }
       # Don't try to read code if the control is inherited from another profile


adamleff:
@brianpcurran thanks for submitting this PR! I need to dig into this a bit deeper to understand what's going on, but your explanation has given us a head start.

I don't want you to think that we're ignoring you, but this PR is going to take a bit longer to go through given the area of the codebase in which the issue lies. I suspect there is a different way to go about fixing this, but I'll need some time to determine that for myself.

Thanks again for your submission. We'll be in touch.

brianreumere:
Sounds good, thanks Adam!

adamleff:
That's a nasty little bug, but I've successfully reproduced it. I'm going to dig in a bit more to see what's going on and then re-evaluate your change and see if that's the best approach. I'll be in touch. Thanks for your patience, @brianpcurran!

brianreumere:
Thanks for the update!

adamleff:
So, while your PR is a nice band-aid on this issue, I think the issue is deeper and should probably be fixed. The main problem is InSpec parses all the dependencies and collects the tests, but doesn't create any `Rule` objects until all the dependencies have been read. Then it tries to create `Rule` objects but only using the `SourceReader` of the main profile. The `MethodSource` stuff properly bails when it can't get the file content from the `SourceReader` (i.e. the control filename from the dependency profile doesn't exist in the main file, so there are no tests in that file).

Because the file names match, the `SourceReader` returns the tests from the main profile, and there isn't a line number / content match, hence the exception.

There are a couple of ways I can think of to solve this. First is to pass back the `SourceReader` along with any tests for each profile dependency so that way we read from the correct profile. This may be difficult given how the `require_controls` and `include_controls` stuff works. Another option is to build the `Rule` objects when we collect the tests so we don't need to read from anything after the dependencies are satisfied... all the reading will have been done already. Again, this may screw with the `require_controls` and `include_controls` stuff, so I'm not sure if that's a good idea either.

Unfortunately, I didn't write this part of InSpec so I'm flailing a bit, but I'll hopefully be able to spend more time on this week to see if I can't come up with a good solution.

pwelch:
@adamleff did you ever have a chance to investigate this one further? Do you know if the changes are still relevant?

jquick:
Thanks for this @brianpcurran ! https://github.com/inspec/inspec/pull/3264 was introduced which allows us to pull code fields from merged/wrapper profiles. I'll double check using check with the sourceReader on it but that should be the new direction to go. Thanks again for pointing this out and the suggested fix!

PR #2275, opened 2017-10-30
allow users to specify user/namespace when fetching profiles from Chef Automate

### LINKED ISSUE #1500: "inspec compliance upload" command does not allow to specify namespace
### Description

When uploading a profile to the Automate profile store using the `inspect compliance upload` command, the namespace is forced to be the namespace of the uploading user. It would be much better to be able to specify the namespace as part of the upload.

InSpec 1.7.2

arlimus:
By default it uses the username of the user that is logged into the system. However, the user can now specify the `--user` on the cli to list profiles from a user other than his own domain.

Fixes #1500

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -18,14 +18,15 @@ module Compliance
     # return all compliance profiles available for the user
     # the user is either specified in the options hash or by default
     # the username of the account is used that is logged in
-    def self.profiles(config, options = {})
+    def self.profiles(config)
+      owner = config['owner'] || config['user']
+
       # Chef Compliance
       if is_compliance_server?(config)
         url = "#{config['server']}/user/compliance"
       # Chef Automate
       elsif is_automate_server?(config)
-        user = options['user'] || config['user']
-        url = "#{config['server']}/profiles/#{user}"
+        url = "#{config['server']}/profiles/#{owner}"
       else
         raise ServerConfigurationMissing
       end
@@ -48,9 +49,8 @@ module Compliance
         elsif is_automate_server_pre_080?(config)
           mapped_profiles = profiles.values.flatten
         else
-          owner_id = config['user']
           mapped_profiles = profiles.map { |e|
-            e['owner_id'] = owner_id
+            e['owner_id'] = owner
             e
           }
         end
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
This works like a charm. Thank you @arlimus

adamleff (approved):
Looks good! Holding off on merging per a quick chat with @chris-rock and awaiting Travis to go green (they're having issues currently).

chris-rock:
I updated the PR to cover a little bit more. It supports the following now:

* list and upload a profile to another owner:
screen shot 2017-10-31 at 11 13 04 am

* profile inheritance with an alternative namespace

Just define the right namespace in `inspec.yml`:
```
depends:
  - name: profile
    compliance: drichter/profile
```
The following shows the result before this branch and after this branch
screen shot 2017-10-31 at 11 19 54 am

chris-rock:
fixes #1500

lib/bundles/inspec-compliance/api.rb
```diff
@@ -85,8 +88,12 @@ def self.version(config)
 
     # verifies that a profile
     def self.exist?(config, profile)
-      _msg, profiles = Compliance::API.profiles(config)
       owner, id, ver = profile_split(profile)
+
+      c = config.dup
```
adamleff: The `c` variable looks unused. Why is it needed? Looks like we can just delete lines 93 and 94, yes?

lib/bundles/inspec-compliance/cli.rb
```diff
@@ -145,11 +150,16 @@ def download(profile_name)
...
     option :overwrite, type: :boolean, default: false,
-      desc: 'Overwrite existing profile on Chef Compliance.'
+      desc: 'Overwrite existing profile on Server.'
+    option :owner, type: :string, required: false,
+      desc: 'Owner that should own the profile'
     def upload(path) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize, PerceivedComplexity, Metrics/CyclomaticComplexity
       config = Compliance::Configuration.new
       return if !loggedin(config)
 
+      c = config.dup
```
adamleff: What was your thinking for needing to `dup` this here vs. just using the existing `config`? We don't return it to the caller, and everything else we pass `config` to appears to potentially need the `owner` and the rest of the `config` object anyways. It just seems unclear to me.

We should also be more explicit with our variable names. `c` isn't very descriptive but it's used throughout this method now.

adamleff (changes requested):

adamleff (approved):
Thanks for your collective work on this, @arlimus and @chris-rock!

PR #2280, opened 2017-11-02
Properly compare profile version strings as SemVer

### LINKED ISSUE #1940: profile cis-rhel6-level1 version 1.4.0-1 which doesn't match >= 0
### Description

inspec compliance upload company_cis_rhel6 --overwrite

The profile cis-rhel6-level1 from {:compliance=>"comp_user/cis-rhel6-level1", :url=>"https://cit-automate.org/compliance/profiles/comp_user/cis-rhel6-level1/tar", :sha256=>"b41dad3b5c2a91924aa332562fb8d7abd16db3a64f42277ee07cb6174db03ca8"} has a version 1.4.0-1 which doesn't match >= 0

1.28, but issue is the same for 1.26, 1.27

### Replication Case

inspec compliance upload company_cis_rhel6 --overwrite

### Possible Solutions

Seems that the -1 with these profile version is not being accounted for

### Stacktrace

The profile cis-rhel6-level1 from {:compliance=>"comp_user/cis-rhel6-level1", :url=>"https://cit-automate.org/compliance/profiles/comp_user/cis-rhel6-level1/tar", :sha256=>"b41dad3b5c2a91924aa332562fb8d7abd16db3a64f42277ee07cb6174db03ca8"} has a version 1.4.0-1 which doesn't match >= 0

adamleff:
When configuring a profile dependency, if the dependent profile had a hyphen in it, it would not properly match the default version constraint of `>= 0`. This is because a hyphen indicates the version is a pre-release version and proper version matching would require the constraint to also be listed with a pre-release version string.

The proper solution is to use the `+` character instead which indicates a build number, which is what the hyphen was meant to convey. In the meantime, this change properly compares version strings as SemVer and also adds tests.

Fixes #1940

### DIFF
--- a/lib/inspec/dependencies/requirement.rb
+++ b/lib/inspec/dependencies/requirement.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require 'inspec/cached_fetcher'
 require 'inspec/dependencies/dependency_set'
+require 'semverse'
 
 module Inspec
   #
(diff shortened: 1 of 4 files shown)


chris-rock (approved):
Very nice improvement @adamleff

PR #2283, opened 2017-11-03
Fix classname in JUnit formatter

### LINKED ISSUE #2276: JUnit output has incorrect "class" attribute on <testcase>
### Description

Running inspec via Test-Kitchen with junit output produces an XML file with entries like:

```xml
<testcase name='Service DatadogAgent should be installed' class='datadog' time='0.383979'/>
```

I'm consuming this in Team City and it ignores the `class` attribute. From what I can see the attribute should be called `classname` (http://help.catchsoftware.com/display/ET/JUnit+Format)

1.36.1 on Windows (via ChefDK)

### Replication Case

Run inspec and output a junit file

### Possible Solutions

This is the line that generates the `class` attribute. The condition also looks strange - test for `:title` member and if it exists use the `:id` member instead?

https://github.com/chef/inspec/blob/e44de521b67c3cb0758432bb60f05510de9d1fd4/lib/inspec/rspec_json_formatter.rb#L896

adamleff:
The JUnit formatter currently incorrectly uses `class` instead of `classname` as an attribute.

Fixes #2276

### DIFF
--- a/lib/inspec/rspec_json_formatter.rb
+++ b/lib/inspec/rspec_json_formatter.rb
@@ -893,7 +893,14 @@ class InspecRspecJUnit < InspecRspecJson
   def build_result_xml(control, result)
     result_xml = REXML::Element.new('testcase')
     result_xml.add_attribute('name', result[:code_desc])
-    result_xml.add_attribute('class', control[:title].nil? ? 'Anonymous' : control[:id])
+    # if there is no control title, we are likely receiving test results from a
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @adamleff

adamleff:
The travis test is a valid failure and needs to be investigated, also holding off on merging this pending some discussion taking place in #2276

adamleff:
Test fixed, updated with suggestion from community members to prefix classname with the profile name to keep things unique and clear.

PR #2285, opened 2017-11-03
http resource: use proper syntax in `curl` header option

schisamo:
`curl` expects a valid header per RFC 2616 when using the `-H`/`--header` option. RFC 2616 declares header field/values should be separated using a colon (`:`):
https://tools.ietf.org/html/rfc2616#section-4.2


adamleff:
Ugh, I'm an idiot, that's my fault. Thanks for the fix, @schisamo.

adamleff (approved):

lehn-etracker:
Please backport that fix to 1.51

PR #2286, opened 2017-11-03
http resource: Add basic param handling to remote HTTP worker

schisamo:

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -202,7 +202,7 @@ module Inspec::Resources
           end
         end
 
-        def curl_command
+        def curl_command # rubocop:disable Metrics/AbcSize
           cmd = ["curl -i -X #{http_method}"]
           cmd << "--connect-timeout #{open_timeout}"
           cmd << "--max-time #{open_timeout+read_timeout}"
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -165,7 +165,7 @@ module Inspec::Resources
 
         def body
           run_curl
-          @body
+          @body&.strip
         end
 
         def response_headers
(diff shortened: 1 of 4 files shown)


adamleff (approved):

lib/resources/http.rb
```diff
@@ -165,7 +165,7 @@ def status
 
         def body
           run_curl
-          @body
+          @body.strip
```
chris-rock: What happens if body is nil?

schisamo: This would be the perfect spot for Ruby 2.3+'s safe navigation operator:
https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

```irb
➜  irb
Welcome to IRB. You are using ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin16]. Have fun ;)
>> nil&.strip #=> nil
>> 'foo'&.strip #=> "foo"
>> 'bar
'>
'>
'> '&.strip #=> "bar"
```

Looks like [InSpec is currently locked on Ruby 2.1+](https://github.com/chef/inspec/blob/master/inspec.gemspec#L27)....what is InSpec's policy surrounding minimum Ruby version?

adamleff: I don't know that we have a formal one, but probably should. Informally it's been "support the oldest version on normal maintenance" which is now 2.3 anyways. I'd be in favor of bumping to 2.3 in the gemspec. We already plan on bumping to 2.4 in the omnibus builds and that's on my backlog.

@chris-rock @arlimus thoughts?

chris-rock: I agree that bumping to 2.4 for omnibus and testing is a good idea. I do not agree that the inspec ruby code should assume that ruby 2.4 is there. In case you need it, the resource need to be able fail if ruby version x is not available.

schisamo: So what is the minimum Ruby version we can assume? Is it OK to assume 2.3 which is _the oldest version on normal maintenance_ as @adamleff mentioned?

adamleff (approved):
re-approved with the safe navigation operator and the master updates to support ruby 2.3 properly. Thanks, @schisamo!

PR #2287, opened 2017-11-03
Pin RubyGems to 2.6.14

adamleff:
2.7.0 seems to have introduced an issue causing bundler to fail to
install in our Jenkins pipeline.


adamleff:
http://manhattan.ci.chef.co//job/inspec-trigger-ad_hoc/37 ✅

scotthain (approved):
May want to leave a comment as to why, but that's just a nitpick

adamleff:
@scotthain comment added - thanks!

PR #2289, opened 2017-11-06
http resource: supply max-time option using read_timeout and open_timeout

### LINKED ISSUE #2288: http resource doesn't obey read_timeout
### Description

http resource obeys connect timeout (open_timeout) but no option is supplied for the read_timeout to the underlying curl call. The result is read_timeout is not observed.

 inspec (1.43.5)

### Replication Case

Create a site that permits connection but doesn't respond beyond that (i.e. hangs)

### Possible Solutions

Add behaviour to the curl command to include --max-time <seconds> as appropriate
Technically this should be open_timeout + read_timeout
as curl doesn't make a distinction between the two.

bdellegrazie:
Curl doesn't distinguish between them so need to use the sum of both as
the overall timeout.

fixes #2288

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -205,6 +205,7 @@ module Inspec::Resources
         def curl_command
           cmd = ["curl -i -X #{http_method}"]
           cmd << "--connect-timeout #{open_timeout}"
+          cmd << "--max-time #{open_timeout+read_timeout}"
           cmd << "--user \'#{username}:#{password}\'" unless username.nil? || password.nil?
           cmd << '--insecure' unless ssl_verify?
           cmd << "--data #{Shellwords.shellescape(request_body)}" unless request_body.nil?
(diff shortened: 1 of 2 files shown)


adamleff (changes requested):
@bdellegrazie thanks for your first PR to InSpec! Great contribution!

Unfortunately, you have some test failures. The curl commands get mocked by our MockLoader so we can test the resource end-to-end, and they use exact matches against the command. Because your PR changed the command, the mocks are no longer valid. You can find the mocks configured here:

https://github.com/chef/inspec/blob/master/test/helper.rb#L455-L466

Essentially, you need to add your new flag so the commands are properly mocked and then you should be fine.

You can run the tests manually on your workstation like so:

 - bundle install
 - bundle exec m test/unit/resources/http_test.rb

If you run the above BEFORE you fix the mocks, you should see errors indicating "command not mocked" that will give you more clues.

bdellegrazie:
@adamleff certainly, will do.

bdellegrazie:
@adamleff updated, please re-review at your earliest convenience, thanks!

adamleff (approved):
Super job, @bdellegrazie - thanks for your contribution!

chris-rock (approved):
Great work @bdellegrazie Thank you very much.

PR #2293, opened 2017-11-07
Require Ruby 2.3 and later

adamleff:
Ruby 2.1 is EOL, and Ruby 2.2 is on security fixes only. This moves
InSpec to support the current "normal maintenance" versions of Ruby
like Chef does and also bumps the versions used in Travis tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,40 +14,39 @@ before_install:
 - gem --version
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


schisamo (approved):

arlimus (approved):
Sweet thank you for the update @adamleff !!

PR #2295, opened 2017-11-07
Switch to tomlrb for TOML parsing

adamleff:
The `toml` gem has a very strict version dependency on an old version of parslet. This change switches us to use `tomlrb` instead which has no direct dependencies. This will allow us to bump up to a later version of parslet that has better error handling and insight into parser errors.


chris-rock (approved):
Thank you @adamleff

PR #2297, opened 2017-11-08
Fix gid filtering for etc_group resource

eramoto:
'etc_group' resource stores 'gid' as integer but the 'where' method
compares 'gid' as string. By this fix, the 'where' method get to compares
'gid' as integer.

### DIFF
--- a/lib/resources/etc_group.rb
+++ b/lib/resources/etc_group.rb
@@ -85,7 +85,7 @@ module Inspec::Resources
       conditions.each do |k, v|
         idx = fields[k.to_sym]
         next if idx.nil?
-        res = res.select { |x| x[idx] == v.to_s }
+        res = res.select { |x| x[idx].to_s == v.to_s }
       end
 
       EtcGroupView.new(self, res)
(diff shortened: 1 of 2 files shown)


lib/resources/etc_group.rb
```diff
@@ -85,7 +85,8 @@ def where(conditions = {})
       conditions.each do |k, v|
         idx = fields[k.to_sym]
         next if idx.nil?
-        res = res.select { |x| x[idx] == v.to_s }
```
adamleff, line 88: Rather than selectively converting the value to an integer only if the field is `gid`, since we intend on comparing all values as strings, we should probably ensure the value we're comparing to is a string too. I think we should just change this line to be:

```ruby
res = res.select { |x| x[idx].to_s == v.to_s }
```

... rather than only converting the `gid`. This has the same issue if someone is looking for groups that have no members... empty members is represented as `nil` but we convert the match value to a string, which will be `""` -- so we may as well convert everything.

adamleff (changes requested):
Thanks for your PR, @eramoto. I think we can take a simpler approach to this and have left you a comment with my recommendation.

eramoto:
Hi @adamleff ,
Thank you for your comment. I fixed this commit and submitted the updated PR.

adamleff (approved):
This is a great improvement, @eramoto - thank you for being responsive to my feedback, and double thanks for adding some great tests.

arlimus (approved):
Thank you @eramoto for fixing this, much appreciated!!

PR #2301, opened 2017-11-09
Habitat build works for all versions, eliminates rake

adamleff:
The Habitat plan has been modified to support building from the repo rather than relying on a gem being pushed to RubyGems. This allows us to build current packages at every merge rather than only pushing to Habitat Builder when we promote to stable.

This change also enables Expeditor to perform builds for us and removes the dependency on the rake task as it is no longer needed.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -10,6 +10,9 @@ docker_images:
       build_args:
         - GEM_SOURCE: http://artifactory.chef.co/omnibus-gems-local
 
+habitat_packages:
+  - inspec
+
 # Slack channel in Chef Software slack to send notifications about build failures, etc
 slack:
   notify_channel: inspec-notify
@@ -43,6 +46,10 @@ merge_actions:
         - "Omnibus: Skip Build"
         - "Expeditor: Skip All"
       only_if: built_in:bump_version
+  - built_in:trigger_habitat_package_build:
+      ignore_labels:
+        - "Habitat: Skip Build"
+        - "Expeditor: Skip All"
 
 artifact_actions:
   promoted_to_unstable:
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -1,5 +1,6 @@
 pkg_name=inspec
 pkg_origin=chef
+pkg_version=$(cat "$PLAN_CONTEXT/../VERSION")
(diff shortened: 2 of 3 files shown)


.expeditor/config.yml
```diff
@@ -10,6 +10,9 @@ docker_images:
       build_args:
         - GEM_SOURCE: http://artifactory.chef.co/omnibus-gems-local
 
+habitat_packages:
+  - inspec
```
schisamo, line 14: I believe this needs/should be a key...you might want to optionally export a container to docker hub:

```yaml
habitat_packages:
  - inspec:
      export:
        - docker
```

This would export the current build and tag with a version (`MAJOR.MINOR.PATCH-YYYYMMDDHHMMSS`), channel (`unstable`). We can then "promote" the image (e.g. update channel tags) along with the omnibus packages via Expeditor artifact_actions.

habitat/plan.sh
```diff
@@ -24,46 +25,63 @@ pkg_build_deps=(
 )
```
schisamo, line 25: Does it make sense to use `scaffolding-ruby` here?

schisamo (approved):
Left a few comments I am happy to chat with you about in the meatspace.

.expeditor/config.yml
adamleff: I was going off of the docs which make it seem like it can be a simple string as long as there were no options:

http://expeditor-docs.es.chef.io/configuration/habitat-packages/#habitat_packages

The InSpec docker image is built and consumed in a different way currently that I'd like to change in the future, but for now I need to keep them separate. Expeditor is at least handling the docker stuff for us, so that's cool.

habitat/plan.sh
adamleff: Unless something changed recently, the Ruby scaffolding worked well for rails, etc... but not for standalone ruby projects. Happy to consider a change in the future, but I'd like the focus of this PR to be on the automation of the existing package building vs. refactoring the package.

adamleff:
@schisamo thanks for the review! I replied to the comments and happy to Zoom or continue the conversation here in the PR further, for sure.

schisamo:
:shipit:

arlimus (approved):
Thank you @adamleff !! 👍

PR #2302, opened 2017-11-10
xinetd_conf resource: fix false positives when config file or directory doesn't exist

eramoto:
Not return 'nil' when xinetd_conf resource checks an empty file or a
non-existent file. So the following test succeeds, where /empty_file is
empty file.

  describe xinetd_conf("/empty_file") do
    it { should be_disabled }
  end

### DIFF
--- a/lib/resources/xinetd.rb
+++ b/lib/resources/xinetd.rb
@@ -53,12 +53,12 @@ module Inspec::Resources
       return @contents[path] if @contents.key?(path)
       file = inspec.file(path)
       if !file.file?
-        return skip_resource "Can't find file \"#{path}\""
+        raise Inspec::Exceptions::ResourceSkipped, "Can't find file: #{path}"
       end
 
       @contents[path] = file.content
       if @contents[path].nil? || @contents[path].empty?
-        return skip_resource "Can't read file \"#{path}\""
+        raise Inspec::Exceptions::ResourceSkipped, "Can't read file: #{path}"
       end
 
       @contents[path]
@@ -69,7 +69,6 @@ module Inspec::Resources
       flat_params = parse_xinetd(read_content)
       # we need to map service data in order to use it with filtertable
       params = { 'services' => {} }
-
       # map services that were defined and map it to the service hash
       flat_params.each do |k, v|
         name = k[/^service (.+)$/, 1]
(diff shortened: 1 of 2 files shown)


adamleff:
Thanks for the PR, @eramoto! There's actually a more graceful way for us to handle this now with raising exceptions, but I discovered a bug while building my reply to you. Let me get this bug fixed first (See #2307) and then I'll describe the change I believe will address this in a better way and in a manner in which we plan to write resources going forward.

lib/resources/xinetd.rb
```diff
@@ -53,12 +53,14 @@ def read_content(path = @conf_path)
       return @contents[path] if @contents.key?(path)
       file = inspec.file(path)
       if !file.file?
-        return skip_resource "Can't find file \"#{path}\""
+        skip_resource "Can't find file \"#{path}\""
```
adamleff: With #2307 now merged, a cleaner way to handle this would be to raise an actual Ruby exception. We have two to primary ones to choose from:

 * `Inspec::Exceptions::ResourceSkipped`: will skip the resource in the output
 * `Inspec::Exceptions::ResourceFailed`: will fail the resource in the output

One could make the argument that a failure is more appropriate here than a skip, but in this case, we should aim to keep consistency since that could be considered a breaking change. So, we can change this line to:

```ruby
raise Inspec::Exceptions::ResourceSkipped, "Can't find file: #{path}"
```

... and then you can remove the `return nil` below it since the `raise` will stop the resource from continuing on.

lib/resources/xinetd.rb
```diff
@@ -53,12 +53,14 @@ def read_content(path = @conf_path)
...
+        return nil
       end
 
       @contents[path] = file.content
       if @contents[path].nil? || @contents[path].empty?
-        return skip_resource "Can't read file \"#{path}\""
+        skip_resource "Can't read file \"#{path}\""
```
adamleff: Same as above - change to a raised exception, and then we can remove the `return nil`

adamleff (changes requested):
Hi @eramoto! I've provided some feedback on a way to handle this a bit more cleanly and what will become a more standard way of us handling resource failures in the future using Ruby exceptions.

eramoto:
Hi @adamleff ,
Thank you for your good opinion. I reflected this, also added this exception to the check against xinetd directory, and fixed a new bug due to using of an unreadable file.

adamleff (approved):
Nicely done, @eramoto!

chris-rock (approved):
Thank you for discovering and fixing that bug @eramoto

PR #2305, opened 2017-11-13
port resource: handle ss output from older iproute package

Wing924:
`port` resource use `ss -tulpen` to check ports when `ss` comand is available.
But `ss -tulpen` output are changed:
iproute-2.6.32-54.el6.x86_64:
```
Netid State      Recv-Q Send-Q  Local Address:Port Peer Address:Port
udp   UNCONN     0      0       *:111              *:*                 users:(("rpcbind",1123,6))
tcp   LISTEN     0      128     *:22               *:*                 users:(("sshd",3965,3))
```
iproute-2.6.32-20.el6.x86_64:
```
Netid            Recv-Q Send-Q  Local Address:Port Peer Address:Port
udp              0      0       *:111              *:*                 users:(("rpcbind",1123,6))
tcp              0      128     *:22               *:*                 users:(("sshd",3965,3))
```
iproute-2.6.32-20 lack State and we should also handle both output.


adamleff (approved):
Looks like a pretty reasonable change to me, @Wing924! Another option would be to look for the word `State` in the first line of the output, and then pass an option to the parse methods to determine what fields to pluck from. But this is a fine solution as well.

arlimus (approved):
Wow great addition @Wing924 , thank you so much 😄 

Kudos for the quick review Adam!!

PR #2307, opened 2017-11-13
Correctly format skip exceptions in formatters

adamleff:
PR #2235 allowed for resources to raise skip and fail exceptions and the RSpec formatters would do the right thing. These work inside initialize methods but not in any other method in the resource.

This change modifies the formatter to honor the ResourceSkipped exception properly. The ResourceFailed exception needs no additional handling as RSpec properly handles any exception thrown.


jerryaldrichiii (approved):
Great catch! Looks good to me.

I verified that using `raise Inspec::Exceptions::ResourceSkipped` in a method outside a control (e.g. ones loaded by FilterTable) incorrectly report as `failed` and with your changes report as `skipped`.

arlimus (approved):
Great catch, kudos Adam!! 😄

PR #2308, opened 2017-11-14
oracledb_session resource: fix credential passing to sql/sqlplus

dmytro-malovany:
Also, because of some reason sqlplus doesn't accept queries with \n, adding of ";" has fixed the problem

Signed-off-by: Dmytro Malovany <dmytro.malovannyy@gmail.com>

### DIFF
--- a/lib/resources/oracledb_session.rb
+++ b/lib/resources/oracledb_session.rb
@@ -60,7 +60,9 @@ module Inspec::Resources
         p = :parse_html_result
       end
 
-      command = "echo \"#{opts}\n#{verify_query(escaped_query)}\nEXIT\" | #{bin} -s #{@user}/#{@password}@//#{@host}:#{@port}/#{@service}"
+      query = verify_query(escaped_query)
+      query += ';' unless query.end_with?(';')
+      command = %{echo "#{opts}\n#{query}\nEXIT" | #{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service}}
       cmd = inspec.command(command)
 
       out = cmd.stdout + "\n" + cmd.stderr
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -416,7 +416,7 @@ class MockLoader
       "5c2bc0f0568d11451d6cf83aff02ee3d47211265b52b6c5d45f8e57290b35082" => cmd.call('mssql-getdate'),
       # oracle
       "bash -c 'type \"sqlplus\"'" => cmd.call('oracle-cmd'),
-      "ef04e5199abee80e662cc0dd1dd3bf3e0aaae9b4498217d241db00b413820911" => cmd.call('oracle-result'),
+      "527f243fe9b01fc7b7d78eb1ef5200e272b011aa07c9f59836d950107d6d2a5c" => cmd.call('oracle-result'),
       # nginx mock cmd
       %{nginx -V 2>&1} => cmd.call('nginx-v'),
       %{/usr/sbin/nginx -V 2>&1} => cmd.call('nginx-v'),


test/unit/resources/oracledb_session_test.rb
```diff
@@ -15,7 +15,7 @@
 
   it 'run a SQL query' do
     resource = load_resource('oracledb_session', user: 'SYSTEM', password: 'supersecurepass', host: '127.0.0.1', service: 'ORCL.localdomain', port: 1527)
-    query = resource.query('SELECT NAME AS VALUE FROM v$database;')
```
dmytro-malovany: this ";" is not needed here, inspec oracle profile contains queries without ";" at the end.

chris-rock:
Are you using the latest version of sqlplus? I tested it with the oracles version on macos `SQL*Plus: Release 12.1.0.2.0`. I think your approach is better though. Have you tested it with SQLCL too? http://www.oracle.com/technetwork/developer-tools/sqlcl/overview/index.html

dmytro-malovany:
@chris-rock , I have the same version of sqlplus:
```
/home/oracle> sqlplus -V
SQL*Plus: Release 12.1.0.2.0 Production
```
No, I didn't rest it with SQLCL.

lib/resources/oracledb_session.rb
```diff
@@ -56,11 +56,11 @@ def query(q)
         p = :parse_csv_result
       else
         bin = @sqlplus_bin
-        opts = "SET MARKUP HTML ON\nSET FEEDBACK OFF"
+        opts = "SET MARKUP HTML ON;\nSET FEEDBACK OFF;"
         p = :parse_html_result
       end
 
-      command = "echo \"#{opts}\n#{verify_query(escaped_query)}\nEXIT\" | #{bin} -s #{@user}/#{@password}@//#{@host}:#{@port}/#{@service}"
+      command = "echo \"#{opts}\n#{verify_query(escaped_query)};\nEXIT\" | #{bin} -s \"#{@user}\"/\"#{@password}\"@//#{@host}:#{@port}/\"#{@service}\""
```
adamleff: Have we tested this with a query where the user supplies a semicolon-terminated query? What happens if the query therefore ends up with `;;`?

I'd rather see us only add the semicolon if the user doesn't supply it. Something like this:

```ruby
command = %{echo "#{opts}\n#{query}\nEXIT" | #{bin} -s "#{@user}"/"#{@password}"@//#{@host}:#{@port}/"#{@service}"}
```

Note the use of `%{ }` to avoid having to do all that escaping of the double-quotes... should make it much more readable as well.

test/unit/resources/oracledb_session_test.rb
adamleff: If my suggestion is implemented above, this change should not be necessary. Let the resource either add or not add the trailing semicolon when necessary.

adamleff (changes requested):
Thanks for your contribution, @bratdim! I've left some suggestions I'd like to see implemented before we move forward with this change.

lib/resources/oracledb_session.rb
```diff
@@ -56,11 +56,11 @@ def query(q)
         p = :parse_csv_result
       else
         bin = @sqlplus_bin
-        opts = "SET MARKUP HTML ON\nSET FEEDBACK OFF"
+        opts = "SET MARKUP HTML ON;\nSET FEEDBACK OFF;"
```
adamleff: Are these necessary? My understanding is `SET` commands are not SQL statements and therefore do not require the semicolons. In my limited testing, the semicolons are not needed here.

dmytro-malovany:
Hi @adamleff thank you for the suggestions, I've implemented them and tested the code, it works fine.

adamleff (approved):
This looks good to me, @bratdim - thanks for your contribution!

chris-rock (approved):
Thank you @bratdim and @adamleff for pointing out the real issue! Awesome quick turnaround!

PR #2313, opened 2017-11-17
Remove debug message from unit test

eramoto:
Stops to output the meaningless message to standard output when testing.

Obvious fix.


adamleff (approved):
Thanks, @eramoto! That definitely should not be there. :)

PR #2317, opened 2017-11-17
json resource (et. al.): allow inspec check to succeed when using command

adamleff:
When using the `json` resource (or any of the resources that subclass JsonConfig), `inspec check` would fail if the content was supplied with the `command` option. This is because the `command` resource is mocked and an empty string would be returned for `stdout`. That content would be blindly passed to the `parse` method would which raise an exception and cause `inspec check` to fail.

This change refactors JsonConfig to be a bit cleaner and use some helper methods. Additionally, we use the new Exceptions to properly raise errors which are naturally caught by Inspec::Profile, etc.

### DIFF
--- a/lib/resources/csv.rb
+++ b/lib/resources/csv.rb
@@ -34,6 +34,8 @@ module Inspec::Resources
 
       # convert to hash
       csv.to_a.map(&:to_hash)
+    rescue => e
+      raise Inspec::Exceptions::ResourceFailed, "Unable to parse CSV: #{e.message}"
     end
 
     # override the value method from JsonConfig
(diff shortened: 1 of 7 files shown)


lib/resources/csv.rb
```diff
@@ -45,8 +47,10 @@ def value(key)
       @params.map { |x| x[key.first.to_s] }.compact
     end
 
-    def to_s
-      "Csv #{@path}"
+    # used by JsonConfig to build up a full to_s method
+    # based on whether a file path, content, or command was supplied.
+    def resource_base_name
```
chris-rock: should this be a private function then?

chris-rock (approved):
Great improvement @adamleff Just added a nit

lib/resources/csv.rb
adamleff: I've historically had issues with `to_s` using private methods. Perhaps it's fixed in later Ruby versions. I'll change it and we can see what happens.

chris-rock: to_s needs to be public, i am just talking about the `resource_base_name`

PR #2318, opened 2017-11-17
Fix minor typo

dragon788:
Amazingly doesn't appear anywhere else in the codebase from a brief search.

### DIFF
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -29,7 +29,7 @@ InSpec), you can use the `-t` flag. We support connecting using ssh,
 WinRm and docker. If no target is provided, we implicitly support the
 "local" target - i.e. tests running on the current machine running
 InSpec. For an ssh connection, use `-i` for specifying ssh key files,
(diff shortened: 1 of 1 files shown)


adamleff:
Hi, @dragon788! Welcome to the InSpec community, and thanks for your first PR to the project!

For legal reasons, all commits to the project must be signed-off according to the Developer Certificate of Origin (DCO). You can find more about the DCO here:

https://www.inspec.io/docs/reference/dsl_inspec/#exclude-specific-test

Essentially, we need you to amend your commit message to either include the `Signed-off-by` line (which is preferred), or a statement that you believe this is an "obvious fix" (which technically, it is!). Either way, the **commit** message has to be modified; we cannot simply modify the PR subject or description.

Would you be able to do that so that I may accept your change? 

Thank you again!

dragon788:
Having read the DCO stuff after making the PR, I see that I missed a part of the commit message for docs re: obvious fix, resubmitted.

PR #2320, opened 2017-11-17
Add clarification to INI resource re: settings in sections

### LINKED ISSUE #2316: INI resource documentation clarification
### Description

Enhance documentation to cover case where ini file has sections and how to reference the attributes

1.45.9

### Replication Case

Current doc only has simple example
```
describe ini('path') do
  its('setting_name') { should eq 'value' }
end
```

### Possible Solutions

If your ini file looks like
```
[section1]
attribute=value
```

To test the value would look something like

```
describe ini('path') do
  its(['section1','attribute']) { should eq 'value' }
end
```
The above worked for me.  Not sure if other method exists.

adamleff:
Fixes #2316

### DIFF
--- a/docs/resources/ini.md.erb
+++ b/docs/resources/ini.md.erb
@@ -18,7 +18,7 @@ An `ini` resource block declares the configuration settings to be tested:
 
 where
 
-* `'setting_name'` is a synchronization setting defined in the INI file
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Great improvement @adamleff

PR #2322, opened 2017-11-20
Remove bundler install during Appveyor tests

adamleff:
It appears that the Appveyor images changed recently and the Ruby
installs for all version (2.2, 2.3, and 2.4) already include Bundler
installed. Attempting to install it will hang the tests at an
interactive prompt asking if you want to override the install, and
--force'ing the install causes path issues and Rubygems won't be able
to activate the gem.

Removing the bundler install appears to solve the problem nicely.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,8 +3,6 @@ platform:
   - x64
 
 environment:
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @adamleff

PR #2328, opened 2017-11-22
Allow crontab resource to read crontab at user specified paths.

miah:
This PR adds support for reading crontab files at user specified paths as requested in https://github.com/chef/inspec/issues/1461.

Example usage:
```
describe crontab('/etc/cron.d/some_crontab') do
  its('commands') { should include '/path/to/some/script' }
end
```

### DIFF
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -104,7 +104,7 @@ module Inspec::Resources
     end
 
     def path?
-      @destination.include?('/')
+      @destination.to_s.include?('/')
     end
   end
 end
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -1,5 +1,4 @@
 # encoding: utf-8
-# author: Adam Leff
 
 require 'utils/parser'
 require 'utils/filter'
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -96,8 +96,12 @@ module Inspec::Resources
     filter.connect(self, :params)
 
     def to_s
-      @destination.nil? ? 'crontab for current user' : "crontab for user #{@destination}"
+      @destination.nil? ? 'crontab for current user' : path_or_user
     end
+
+    def path_or_user
+      pou = @destination.include?('/') ? 'path' : 'user'
+      "crontab for #{pou} #{@destination}"
     end
   end
 end
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -100,8 +100,12 @@ module Inspec::Resources
     end
 
     def path_or_user
-      pou = @destination.include?('/') ? 'path' : 'user'
+      pou = path? ? 'path' : 'user'
       "crontab for #{pou} #{@destination}"
+    end
+
+    def path?
+      @destination.include?('/')
     end
   end
 end
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -38,7 +38,8 @@ module Inspec::Resources
     end
 
     def read_crontab
-      inspec.command(crontab_cmd).stdout.lines.map { |l| parse_crontab_line(l) }.compact
+      ct = path? ? inspec.file(@destination).content : inspec.command(crontab_cmd).stdout
+      ct.lines.map { |l| parse_crontab_line(l) }.compact
     end
 
     def parse_crontab_line(l)
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -31,8 +31,8 @@ module Inspec::Resources
 
     include CommentParser
 
-    def initialize(user = nil)
-      @user   = user
+    def initialize(destination = nil)
+      @destination = destination
       @params = read_crontab
 
       return skip_resource 'The `crontab` resource is not supported on your OS.' unless inspec.os.unix?
@@ -73,7 +73,7 @@ module Inspec::Resources
     end
 
     def crontab_cmd
-      @user.nil? ? 'crontab -l' : "crontab -l -u #{@user}"
+      @destination.nil? ? 'crontab -l' : "crontab -l -u #{@destination}"
     end
 
     filter = FilterTable.create
@@ -96,7 +96,8 @@ module Inspec::Resources
     filter.connect(self, :params)
 
     def to_s
-      @user.nil? ? 'crontab for current user' : "crontab for user #{@user}"
+      @destination.nil? ? 'crontab for current user' : "crontab for user #{@destination}"
+    end
     end
   end
 end
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -24,6 +24,10 @@ module Inspec::Resources
       describe crontab.where { command =~ /a partial command string/ } do
         its('entries.length') { should cmp 1 }
       end
+
+      describe crontab('/etc/cron.d/some_crontab') do
+        its('commands') { should include '/path/to/some/script' }
+      end
     "
 
     attr_reader :params
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -44,7 +44,7 @@ module Inspec::Resources
     end
 
     def read_crontab
-      ct = path? ? inspec.file(@destination).content : inspec.command(crontab_cmd).stdout
+      ct = @path.nil? ? inspec.command(crontab_cmd).stdout : inspec.file(@path).content
       ct.lines.map { |l| parse_crontab_line(l) }.compact
     end
 
--- a/lib/resources/crontab.rb
+++ b/lib/resources/crontab.rb
@@ -8,7 +8,7 @@ module Inspec::Resources
     name 'crontab'
     desc 'Use the crontab InSpec audit resource to test the contents of the crontab for a given user which contains information about scheduled tasks owned by that user.'
     example "
-      describe crontab({user:'root'}) do
+      describe crontab({user: 'root'}) do
         its('commands') { should include '/path/to/some/script' }
       end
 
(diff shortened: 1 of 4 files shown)


lib/resources/crontab.rb
```diff
@@ -96,7 +100,16 @@ def crontab_cmd
     filter.connect(self, :params)
 
     def to_s
-      @user.nil? ? 'crontab for current user' : "crontab for user #{@user}"
+      @destination.nil? ? 'crontab for current user' : path_or_user
+    end
+
+    def path_or_user
```
adamleff: We should probably make this a private method.

lib/resources/crontab.rb
```diff
@@ -73,7 +77,7 @@ def parse_crontab_line(l)
     end
 
     def crontab_cmd
-      @user.nil? ? 'crontab -l' : "crontab -l -u #{@user}"
+      @destination.nil? ? 'crontab -l' : "crontab -l -u #{@destination}"
```
adamleff: I don't think this works... `-u` specifically takes a user. On a stock Ubuntu 16.04 box, which has a `/etc/cron.d/mdadm` file, `crontab -l -u /etc/cron.d/mdadm` fails. I think you just need to read in the file.

lib/resources/crontab.rb
```diff
@@ -25,21 +24,26 @@ class Crontab < Inspec.resource(1)
...
+      @destination = destination
       @params = read_crontab
 
       return skip_resource 'The `crontab` resource is not supported on your OS.' unless inspec.os.unix?
     end
 
     def read_crontab
-      inspec.command(crontab_cmd).stdout.lines.map { |l| parse_crontab_line(l) }.compact
+      ct = path? ? inspec.file(@destination).content : inspec.command(crontab_cmd).stdout
+      ct.lines.map { |l| parse_crontab_line(l) }.compact
```
adamleff: Non-user crontabs, such as those in `/etc/cron.d` follow a different format. Specifically, before the command, there is another field for the user via which the command should be run. We'd need to adapt the resource to handle that new format for non-user crontabs.

test/unit/mock/files/crondotd
```diff
@@ -0,0 +1,10 @@
+#
```
adamleff, line 1: I think we need to provide a better mock file that has users in it. Here's the contents of the `/etc/cron.d/mdadm` file from the Ubuntu vagrant box I just spun up:

```
#
# cron.d/mdadm -- schedules periodic redundancy checks of MD devices
#
# Copyright © martin f. krafft <madduck@madduck.net>
# distributed under the terms of the Artistic Licence 2.0
#

# By default, run at 00:57 on every Sunday, but do nothing unless the day of
# the month is less than or equal to 7. Thus, only run on the first Sunday of
# each month. crontab(5) sucks, unfortunately, in this regard; therefore this
# hack (see #380425).
57 0 * * 0 root if [ -x /usr/share/mdadm/checkarray ] && [ $(date +\%d) -le 7 ]; then /usr/share/mdadm/checkarray --cron --all --idle --quiet; fi
```

adamleff (changes requested):
Hi @miah! I love that we're enhancing this resource. However, I think some incorrect assumptions were made regarding the format of non-user crontab files and how they can be read in. Perhaps I tested on a different OS than you did, so I'm happy to be wrong about this as well.

lib/resources/crontab.rb
```diff
@@ -25,21 +24,26 @@ class Crontab < Inspec.resource(1)
...
+      end
     "
 
     attr_reader :params
 
     include CommentParser
 
-    def initialize(user = nil)
-      @user   = user
+    def initialize(destination = nil)
```
chris-rock: I am a little bit unsure about the overloading of the destination. It can be a user name or a path. Wouldn't it be cleaner to use something like:
`describe crontab({path: '/etc/cron.d/some_crontab'})` and `describe crontab({user: 'chris'})`

test/unit/mock/files/crondotd
miah: Thanks, my precise64 doesn't have this; even after I installed mdadm. =)

lib/resources/crontab.rb
miah: Yes will do!

miah:
This PR updates the crontab resource to parse system and user crontab files. To parse a system crontab file pass a `{path: '/some/path'}` to the resource. The resource still accepts a user passed in as a string `describe crontab('root')`. You may now match against the `user` field.

```
describe crontab({user: 'root'}) do
  its('commands') { should include '/path/to/some/script' }
  its('user') { should be 'root' }
end

describe crontab({path: '/etc/cron.d/some_crontab'}) do
  its('commands') { should include '/path/to/some/script' }
end
```

lib/resources/crontab.rb
```diff
@@ -1,15 +1,15 @@
 # encoding: utf-8
-# author: Adam Leff
 
 require 'utils/parser'
 require 'utils/filter'
 
+# rubocop:disable Metrics/ClassLength
```
adamleff: To be consistent with the way we disable in other files, can you please move this to the class definition itself?

```ruby
class Crontab < Inspec.resource(1) # rubocop:disable Metrics/ClassLength
```

lib/resources/crontab.rb
```diff
@@ -1,15 +1,15 @@
...
 module Inspec::Resources
   class Crontab < Inspec.resource(1)
     name 'crontab'
     desc 'Use the crontab InSpec audit resource to test the contents of the crontab for a given user which contains information about scheduled tasks owned by that user.'
     example "
-      describe crontab('root') do
+      describe crontab({user: 'root'}) do
```
adamleff: The `{ }` braces aren't necessary, Ruby will happily see that you're passing in a hash if you use the 2.0-and-greater syntax like you're doing:

```ruby
describe crontab(user: 'root') do
```

lib/resources/crontab.rb
```diff
@@ -25,51 +25,35 @@ class Crontab < Inspec.resource(1)
       describe crontab.where { command =~ /a partial command string/ } do
         its('entries.length') { should cmp 1 }
       end
+
+      describe crontab({path: '/etc/cron.d/some_crontab'}) do
```
adamleff: Same here, unnecessary `{ }` characters :)

lib/resources/crontab.rb
```diff
@@ -25,51 +25,35 @@ class Crontab < Inspec.resource(1)
...
     "
 
     attr_reader :params
 
     include CommentParser
 
-    def initialize(user = nil)
-      @user   = user
+    def initialize(opts = {})
+      Hash[opts.map { |k, v| [k.to_sym, v] }] if opts.respond_to?(:fetch)
```
adamleff: We're doing a check to see if it's a Hash on each of these 3 lines, so can we consolidate it so it reads a bit more clearly?

```ruby
if opts.respond_to?(:fetch)
  # user supplied a Hash containing options
  @user = opts[:user]
  @path = opts[:path]
else
  # user passed in a string, and the prior behavior only supported
  # user crontabs, so assume they're supplying a user.
  @user = opts
  @path = nil
end
```

lib/resources/crontab.rb
```diff
@@ -25,51 +25,35 @@ class Crontab < Inspec.resource(1)
...
+      @user = opts.respond_to?(:fetch) ? opts.fetch(:user, nil) : opts
+      @path = opts.fetch(:path, nil) if opts.respond_to?(:fetch)
       @params = read_crontab
 
       return skip_resource 'The `crontab` resource is not supported on your OS.' unless inspec.os.unix?
     end
 
     def read_crontab
-      inspec.command(crontab_cmd).stdout.lines.map { |l| parse_crontab_line(l) }.compact
+      ct = @path.nil? ? inspec.command(crontab_cmd).stdout : inspec.file(@path).content
```
adamleff: We do the `@path.nil?` check a few times. What do you think about adding a helper method or to?

```ruby
def is_system_crontab?
  !@path.nil?
end

def is_user_crontab?
  !@user.nil?
end
```

We could use that here, in the `to_s` method, and in `parse_crontab_line`.

lib/resources/crontab.rb
```diff
@@ -25,51 +25,35 @@ class Crontab < Inspec.resource(1)
...
 
-    def initialize(user = nil)
-      @user   = user
+    def initialize(opts = {})
+      Hash[opts.map { |k, v| [k.to_sym, v] }] if opts.respond_to?(:fetch)
+      @user = opts.respond_to?(:fetch) ? opts.fetch(:user, nil) : opts
+      @path = opts.fetch(:path, nil) if opts.respond_to?(:fetch)
       @params = read_crontab
 
       return skip_resource 'The `crontab` resource is not supported on your OS.' unless inspec.os.unix?
```
adamleff: While you're in here, can you change this to raise an exception?

```ruby
raise Inspec::Exceptions::ResourceSkipped, 'The `crontab` resource is not supported on your OS.' unless inspec.os.unix?
```

And we should also add an exception if the user didn't supply us enough information (i.e. they gave us a hash that contains neither user or path):

```ruby
raise Inspec::Exceptions::ResourceFailed, "A user or path must be supplied" if @user.nil? && @path.nil?
```

adamleff (changes requested):
@miah nice work! A couple of clean-up items and I think this will be ready to go.

miah:
Hi,

I have addressed that list of clean up items. Please let me know if there is anything else I should consider. =)

adamleff (approved):
Alright, @miah! This looks great. Conditionally approved pending green tests after a rebase once we square away the unrelated Train issue.

lib/resources/crontab.rb
```diff
@@ -25,51 +24,40 @@ class Crontab < Inspec.resource(1)
       describe crontab.where { command =~ /a partial command string/ } do
         its('entries.length') { should cmp 1 }
       end
+
+      describe crontab(path: '/etc/cron.d/some_crontab') do
```
arlimus, line 28: love the new options specification!

arlimus (approved):
What a wonderful improvement and update, thank you so much @miah !! 😄 

Kudos Adam for your guidance as well :)

PR #2336, opened 2017-11-24
[OI] correctly check if auditd version is >= 2.3

wunzeco:
This change fixes the bug reported here, https://github.com/chef/inspec/issues/2334.

Further info:
"auditd" resource does not work for any auditd installation that its
listed rules (output of "auditctl -l") start with "LIST_RULES:".
Unfortunately, there is a wrong assumption made in the resource code
that kind of assumes output of "auditctl -l" for audit >= 2.3 does
not start with "LIST_RULES:". On a ubuntu 14.04 host with auditd 2.3.2
listed rules do indeed start with "LIST_RULES:"

Issue raised on Inspec github repo: https://github.com/chef/inspec/issues/2334

As workaround, "file" resource can be used though not ideal.

### DIFF
--- a/lib/resources/auditd.rb
+++ b/lib/resources/auditd.rb
@@ -34,9 +34,10 @@ module Inspec::Resources
 
     def initialize
       @content = inspec.command('/sbin/auditctl -l').stdout.chomp
+      @version = inspec.command('/sbin/auditctl -v').stdout.chomp
       @params = []
 
-      if @content =~ /^LIST_RULES:/
+      if @version.to_f < 2.3
         return skip_resource 'The version of audit is outdated. The `auditd` resource supports versions of audit >= 2.3.'
       end
       parse_content


jburns12:
@wunzeco This PR changes the logic of how the auditd resource determines its behavior, which will actually break the resource. It should check for LIST_RULES because its behavior is dependent upon the syntax of the output from auditctl, not necessarily the version of auditd. If anything perhaps the documentation should be changed. The difference will be in the inspec control that you're writing, not in the resource itself...at this point I would check out using the auditd_rules resource instead, but @adamleff may have another solution.

adamleff:
@jburns12 I'm not sure I completely agree with you. On the surface, this appears to be a completely valid change because our own "skip_resource" messaging says that we only support `>= 2.3`. Rather than checking the version, we make an assumption as to the version based on the output NOT containing `LIST_RULES`. If there are versions of auditd >= 2.3 that return output containing `LIST_RULES`, our own messaging is incorrect.

I'm fine with saying that this resource only supports certain versions of auditd, but it should be accurate and not based on the output of the command. Since I'm not well-versed in auditd, someone smarter than me will have to help me figure out where that dividing line is.

Even if we were to continue to base our decision as to whether we do or don't support the version of auditd on the scanned node (based on whether the output contains `LIST_RULES` or not), we need to provide good messaging to the user as to how to fix the problem, which will likely involve telling them a version they need to upgrade to.

adamleff:
There hasn't been additional movement or discussion on this PR, so I'm going to close it for now. I'm happy to reopen or accept a new PR when we're ready to proceed. Thank you!

PR #2340, opened 2017-11-27
http resource: properly support HEAD request with remote worker

### LINKED ISSUE #2335: HTTP Resource HEAD timeouts
### Description

When using the http resource with head parameter, the curl command that executes gives a timeout because it´s malformed.

InSpec version: 1.43.8
Platform: Amazon Linux AMI 2017.03 (Based on RHEL Fedora)

### Replication Case

Create a describe block like this:
```
describe http('http://google.com', method: 'HEAD', enable_remote_worker: true) do
  its('status') { should eq 200 }
end
```

And execute inspec with `-l debug`. 
You will see that the command who is been executing is `curl -i -X HEAD --connect-timeout 60 'http://google.com'`

### Possible Solutions

Delete the `-X HEAD` from the method 'HEAD' of the http resource. With the `-i` it´s enough but you can add `--head` if you want.

To workaround this I used the command resource:
```
describe command("curl --head 'http://google.com'") do
  its('stdout') {should match /HTTP\/1.1 200 OK/ }
end
```

### Stacktrace

```
[2017-11-24T13:55:14+01:00] DEBUG: Registering rule (generated from test.rb:31 235b7ba485b901fc871767aa31bd44a2)
D, [2017-11-24T13:55:14.299651 #5964] DEBUG -- : [SSH] user@IP<{:user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>15, :auth_methods=>["none", "publickey"], :keys_only=>nil, :keys=>nil, :password=>"<hidden>", :forward_agent=>nil, :verify_host_key=>false}> (curl -i -X HEAD --connect-timeout 60 'http://google.com')
```

adamleff:
The existing method of adding `-X HEAD` to the curl command does not work properly and can cause timeouts because curl doesn't properly close the connection. The correct way is to use curl's own `--head` flag.

Fixes #2335

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -203,7 +203,17 @@ module Inspec::Resources
         end
 
         def curl_command # rubocop:disable Metrics/AbcSize
-          cmd = ["curl -i -X #{http_method}"]
+          cmd = ['curl -i']
+
+          # Use curl's --head option when the method requested is HEAD. Otherwise,
+          # the user may experience a timeout when curl does not properly close
+          # the connection after the response is received.
+          if http_method.casecmp('HEAD') == 0
+            cmd << '--head'
+          else
+            cmd << "-X #{http_method}"
+          end
+
           cmd << "--connect-timeout #{open_timeout}"
           cmd << "--max-time #{open_timeout+read_timeout}"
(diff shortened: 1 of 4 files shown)


chris-rock (approved):
Thank you @adamleff

PR #2341, opened 2017-11-27
Remove rubocop from example Gemfiles

adamleff:
The pinned version of Rubocop in some of the TK examples' Gemfiles was a very old version with known vulnerabilities. Since these are just examples and have no Rake tasks that rely on them, I removed rubocop outright from the example Gemfiles.


chris-rock (approved):
Thank you @adamleff

PR #2342, opened 2017-11-27
wmi resource: properly escape quotes in WMI query

### LINKED ISSUE #2260: WMI resource does not escape double-quotes
### Description

I cannot specify a filter on a WMI resource that is a string containing double-quotes. Doing so will make it so WMI does not find any resources matching the given filter.

Currently using v1.36.1 (part of ChefDK 2.3.4) on MacOS 10.12.6 targeting Window Server 2016

### Replication Case

```ruby
# Does not work:
describe wmi({
  class: 'win32_service',
  filter: 'name like "%winrm%"'
}) do
  its('Status') { should cmp 'ok' }
  its('State') { should cmp 'Running' }
  its('ExitCode') { should cmp 0 }
  its('DisplayName') { should eq 'Windows Remote Management (WS-Management)'}
end

# Does work:
describe wmi({
  class: 'win32_service',
  filter: "name like '%winrm%'"
}) do
  its('Status') { should cmp 'ok' }
  its('State') { should cmp 'Running' }
  its('ExitCode') { should cmp 0 }
  its('DisplayName') { should eq 'Windows Remote Management (WS-Management)'}
end
```

### Possible Solutions

This is an issue with key-value pairs being serialized into an array of flags for PowerShell. Currently it looks like this:

```ruby
args.each { |key, value| params += " -#{key} \"#{value}\"" }
```

Replacing it like so would solve the problem:

```ruby
args.each { |key, value| params += " -#{key} \"#{value.gsub('"', '`"')}\"" }
```

[Offending line](https://github.com/chef/inspec/blob/1479feecd3e1c3e164f477d3d199d10d394b8b57/lib/resources/wmi.rb#L70)

### Stacktrace

_No stacktrace, only test failure_

TheLonelyGhost:
Fixes #2260.

This modifies an existing test to include that scenario and fixes it as 2 separate commits.

### DIFF
--- a/test/integration/default/controls/wmi_spec.rb
+++ b/test/integration/default/controls/wmi_spec.rb
@@ -14,7 +14,7 @@ end
 # Use win32_service with filter, it returns a single service object
 describe wmi({
   class: 'win32_service',
-  filter: "name like '%winrm%'"
+  filter: 'name like "%winrm%"'
 }) do
   its('Status') { should cmp 'ok' }
   its('State') { should cmp 'Running' }
--- a/lib/resources/wmi.rb
+++ b/lib/resources/wmi.rb
@@ -27,7 +27,7 @@ module Inspec::Resources
 
     def initialize(wmiclass = nil, opts = nil)
       # verify that this resource is only supported on Windows
-      return skip_resource 'The `windows_feature` resource is not supported on your OS.' unless inspec.os.windows?
+      return skip_resource 'The `wmi` resource is not supported on your OS.' unless inspec.os.windows?
 
       @options = opts || {}
       # if wmiclass is not a hash, we have to handle deprecation behavior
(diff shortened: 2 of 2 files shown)


adamleff (approved):
You know more about Windows than I do, but this seems like a reasonable change. Thanks, @TheLonelyGhost!

test/integration/default/controls/wmi_spec.rb
```diff
@@ -14,7 +14,7 @@
 # Use win32_service with filter, it returns a single service object
 describe wmi({
   class: 'win32_service',
-  filter: "name like '%winrm%'"
+  filter: 'name like "%winrm%"'
```
chris-rock, line 17: Isn't this breaking existing users that used the old syntax?

TheLonelyGhost: I'm not sure I understand what you mean. The original test checks if a nested string works with single-quotes (`'`) whereas this test checks against nested double-quotes (`"`).

This small change makes it so the same test covers 2 concepts instead of just one.

Since the wmi resource implementation executes powershell with nested `"`'s using a string composed with `"`'s, it makes sense to check that the string given here uses `"`'s and that it escapes them properly.

Additionally, this line of the test checks that the `filter` option actually works as expected.

What's the question here? What concern is there about a breaking change?

chris-rock: Seems like it was late during my review. All good, thank you for catching this one!

chris-rock (approved):
Thank you for catching this issue @TheLonelyGhost

PR #2347, opened 2017-11-28
allow override of attribute identifier 

chris-rock:
This PR tackles two improvements for attribute objects:

- it allows the override of attribute identifier
- supports values for default different from strings


adamleff (approved):
Looks good, thanks @chris-rock!

PR #2349, opened 2017-11-28
Allow skipping/failing resources in FilterTable

jerryaldrichiii:
`FilterTable` is commonly used in the class body of a resource and is evaluated during an `instance_eval`. This means that if you raise an exception (e.g. SkipResource) it will halt `inspec exec` and `inspec check`.

This adds an `ExceptionCatcher` class that will postpone evaluation until test execution.

This allows `inspec check` and `inspec exec` to perform as intended when skipping/failing a resource in `FilterTable`.

Huge thanks to @adamleff for providing the starting code/ideas!

### DIFF
--- a/lib/utils/filter.rb
+++ b/lib/utils/filter.rb
@@ -7,10 +7,9 @@ module FilterTable
   module Show; end
 
   class ExceptionCatcher
-    def initialize(original_resource, original_exception, accessors)
+    def initialize(original_resource, original_exception)
       @original_resource = original_resource
       @original_exception = original_exception
-      @accessors = accessors
     end
 
     # This method is called via the runner and signals RSpec to output a block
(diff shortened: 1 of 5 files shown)


adamleff (approved):

lib/utils/filter.rb
```diff
@@ -6,6 +6,48 @@
...
+      @original_exception.is_a?(Inspec::Exceptions::ResourceFailed)
+    end
+
+    def resource_exception_message
+      @original_exception.message
+    end
+
+    # Capture message chains and return `ExceptionCatcher` objects
+    def method_missing(*)
+      ExceptionCatcher.new(@original_resource, @original_exception)
```
chris-rock: Since we do not keep the chain, could we just return `self`?

chris-rock (approved):
Great handling of InSpec/RSpec, smart solution for the chaining. Thank you @jerryaldrichiii @adamleff and @miah and @jquick for the team work to make this happen!

lib/utils/filter.rb
adamleff: Yes, we totally can. I was originally misled thinking `self` was going to be the original object that was throwing the exception, but I tricked myself because I was overriding `to_s` and `inspec` in `ExceptionCatcher` to match the original exception for InSpec/RSpec formatter purposes.

I just pushed up a change to change this to `self` instead.

chris-rock: Nice :-)

PR #2350, opened 2017-11-29
Unique export file for security policy resource

jquick:
The SecurityPolicy resource has a potential race condition when exporting the policy. If two runners were both working on the SecurityPolicy resource one could delete the policy export file before the other finished the export. This would cause the policy test to fail on the second runner. 

This change is to create a SecureRandom hex and append it to the policy export file. This resolves the issue as each instance will have its own unique file.

### DIFF
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -119,12 +119,8 @@ module Inspec::Resources
       # store file content
       cmd = inspec.command("Get-Content #{export_file}")
       return skip_resource "Can't read security policy" if cmd.exit_status.to_i != 0
-      @content = cmd.stdout
 
-      if @content.empty?
-        return skip_resource "Can't read security policy"
-      end
-      @content
+      @content = cmd.stdout
     ensure
       # delete temp file
       inspec.command("Remove-Item #{export_file}").exit_status.to_i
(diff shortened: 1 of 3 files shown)


lib/resources/security_policy.rb
```diff
@@ -108,22 +109,25 @@ def to_s
...
+      cmd = inspec.command("secedit /export /cfg #{export_file}")
       return nil if cmd.exit_status.to_i != 0
 
       # store file content
-      cmd = inspec.command('Get-Content win_secpol.cfg')
+      cmd = inspec.command("Get-Content #{export_file}")
       return skip_resource "Can't read security policy" if cmd.exit_status.to_i != 0
       @content = cmd.stdout
 
-      if @content.empty? && !file.empty?
```
jquick: This was removed as it seems to be legacy code. We don't have any local `file` variable. In the future we could look into using the Train::File to capture the content of the policy.

adamleff (approved):
Nice fix, @jquick.

lib/resources/security_policy.rb
chris-rock: I think we should remove the content check completely. If the content is empty, we should not expect that to be okay, so it is probably a failure. Can we add a test that ensures the parsing is not crashing if we have an empty file?

chris-rock (approved):
Thank you @jquick

PR #2353, opened 2017-11-29
Update command resource to check for mock backend

jquick:
This change removes the command errors when using a mock backend that does not set a os.name. Ex: `command(docker).exist? is not supported on your OS: unknown` 

This change goes in tandem with https://github.com/chef/train/pull/221. It can be merged early if needed but we will not see the errors go away till both are merged and inspec is updated to use the new version of train.

### DIFF
--- a/lib/resources/command.rb
+++ b/lib/resources/command.rb
@@ -45,9 +45,9 @@ module Inspec::Resources
       result.exit_status.to_i
     end
 
-    def exist?
+    def exist? # rubocop:disable Metrics/AbcSize
       # silent for mock resources
-      return false if inspec.os.name.nil?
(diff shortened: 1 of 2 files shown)


adamleff (approved):
I think this is a fine approach. Another would be to do `inspec.backend.is_a?(Train::Transports::Mock::Connection)` but either way, we're tying ourselves to a specific Train implementation detail, and I'd rather name the mock OS something other than "unknown" and leave "unknown" for OSs we literally cannot detect.

arlimus (approved):
Great catch @jquick :D thank you

arlimus:
Fully agree Adam, great points on `unknown`

PR #2363, opened 2017-12-03
Fix incorrect case in paragraph.

jordancraig:
Incorrect case used in paragraph to describe interactive debugging with Pry.

Obvious fix.


adamleff (approved):
A 100% obvious fix, indeed! Thanks for your first contribution to InSpec, @jordancraig 🙂

PR #2364, opened 2017-12-04
key_rsa resource: fix inline shell documentation help, wrong resource name used in examples

eramoto:
The resource name is incorrect in the example in key_rsa.rb.


adamleff (approved):
That is quite a silly error on our part. Thanks for the fix, @eramoto!

PR #2368, opened 2017-12-04
security_policy resource: use PID for filename instead of random

jquick:
There is an issue with the security policy when used with the new Train caching. Since it creates a unique export name every call it will not be cached. This change was originally created to protect from parallel runners. Moving it to Process.pid will allow it to be cached as well as protect from parallel runners. 


jerryaldrichiii:
Excellent catch @jquick!

adamleff (approved):

arlimus:
Thank you for fixing this @jquick !!

PR #2371, opened 2017-12-05
Lazy filtertable schema opt

aduric:
As per https://github.com/chef/inspec/issues/2370

This enables deferred function execution in the `FilterTable`, like so:

```
  # assuming we always pass some things to the procs
  def self.lookup_access_keys(table_row, condition)
      # condition is presumably the username
      # table_row is a hash with the other info about a user
  end

   # You could also define a proc out here in a variable
   login_count = Proc.new(table_row, condition) do
      # Same args as above
      # go fetch login count for this user from CloudTrail
   end

  filter = FilterTable.create
    filter.add_accessor(:where)
            .add(:username,     field: 'username')
            .add(:access_keys, lazy: self.method(:lookup_access_keys) ) # No need to specify or pass args here
            .add(:login_count, lazy: login_count)
```

There is also a test script created to play around with this implementation here https://gist.github.com/aduric/7f6ba3f3998b2c14d000cbc0e4e627ad

### DIFF
--- a/lib/utils/filter.rb
+++ b/lib/utils/filter.rb
@@ -125,10 +125,11 @@ module FilterTable
       end
     end
 
-    def resolve_column(func, args, field)
+    def resolve_column(func, field)
       @params.each_with_index do |line, i|
-        mapped_args = args.map { |x| line[x] }
-        @params[i][field] = func.call(*mapped_args)
+        mapped_args = func.parameters.map { |x| x[1] }
+        mapped_vals = line.values_at(*mapped_args)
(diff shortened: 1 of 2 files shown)


jquick:
Thanks for this proof of concept @aduric. Sorry this has sat so long. The team is thinking of redoing some of the filter table internals and this likely will be involved in some flavor. Due to the rebuild this will likely not be merged directly but will be used as a blueprint.

aduric:
Thanks for your reply. I'll close this.

PR #2376, opened 2017-12-06
Fix inspec appveyor test with the new local train transport

jquick:
The new Train local transport tries to use the new pipe runner here:
https://github.com/chef/inspec/blob/master/test/helper.rb#L84

This change forces it to use the generic runner which ensures we don't create pipes for every test. This PR requires https://github.com/chef/train/pull/225

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@
 source 'https://rubygems.org'
 gemspec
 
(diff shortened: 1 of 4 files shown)


arlimus:
Great to see it working :D

adamleff (approved):
Excellent, thanks!

PR #2377, opened 2017-12-06
Resolve merge issue with json-config vs thor defaults

jquick:
When testing the backend_cache via json-config we noticed that the Thor defaults were overriding the json-config settings. This was due to a hash merge using the wrong hash as the source. Rearranging the hash merge and using the json-config options as the override hash fixes this issue.

Along with this change it was requested for debug lines to ensure backend_cache was enabled or not. 

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -120,7 +120,7 @@ module Inspec
 
     def merged_opts
       # argv overrides json
-      Thor::CoreExt::HashWithIndifferentAccess.new(options_json.merge(options))
+      Thor::CoreExt::HashWithIndifferentAccess.new(options.merge(options_json))
     end
 
     def options_json
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -59,16 +59,24 @@ module Inspec
         desc: 'A list of controls to run. Ignore all other tests.'
       option :format, type: :string,
         desc: 'Which formatter to use: cli, progress, documentation, json, json-min, junit'
-      option :color, type: :boolean, default: true,
+      option :color, type: :boolean,
         desc: 'Use colors in output.'
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
       option :cache, type: :string,
         desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
-      option :create_lockfile, type: :boolean, default: true,
+      option :create_lockfile, type: :boolean,
         desc: 'Write out a lockfile based on this execution (unless one already exists)'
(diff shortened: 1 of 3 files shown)


lib/inspec/base_cli.rb
```diff
@@ -120,7 +120,7 @@ def opts
 
     def merged_opts
       # argv overrides json
-      Thor::CoreExt::HashWithIndifferentAccess.new(options_json.merge(options))
+      Thor::CoreExt::HashWithIndifferentAccess.new(options.merge(options_json))
```
adamleff: This will fix the "Thor defaults incorrectly overriding the JSON options" problem, but it introduces the "JSON options overriding user-provided Thor CLI flags" problem. And the comment on line 122 indicates the intention - flags on the CLI should override the JSON.

I think we may have to remove default assignments in Thor and handle them on our own:

* remove all defaults in Thor option declarations
* build a hash of all the options that have defaults, and assign their default values
* merge in the JSON options
* merge in the Thor options

test/unit/base_cli_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+require 'helper'
+require 'thor'
+
+describe 'BaseCLI' do
+  let(:cli) { Inspec::BaseCLI.new }
+
+  describe 'merge_options' do
+    it 'json options override cli' do
+      json = { "backend_cache":true }
```
adamleff: `read_config` doesn't return JSON but rather parsed JSON. Can we change the variable name to `parsed_json` so it's clear? initially I thought you were trying to represent actual JSON here and thought you forgot the quotes around the whole string. 🙂 

Also, this Hash has a key of a symbol named `:backend_cache` with the way you've written it. `JSON.parse` won't return it like that, so we should probably change this to:

```ruby
parsed_json = { 'backend_cache' => true }
```

test/unit/base_cli_test.rb
```diff
@@ -0,0 +1,31 @@
...
+      cli_options = { 'json_config' => 'dummy', 'backend_cache' => false }
+      cli.instance_variable_set(:@options, cli_options)
+      cli.expects(:read_config).returns(json)
```
adamleff: Maybe add a `.with` to this so we show how the options plumb all the way to the config reading?

```ruby
cli.expects(:read_config).with('dummy').returns(json)
```

adamleff (changes requested):
Hi @jquick! I think this change will introduce a problem. Details below.

adamleff (approved):
Looks great. I especially like how you appease my OCD with line spaces.

arlimus (approved):
Awesome fix and coding, thank you @jquick !!

PR #2378, opened 2017-12-07
Update default cli options to be uniq per command type

jquick:
https://github.com/chef/inspec/pull/2377 was released to solve the default command issue with json-config. While this fix works it makes the default cli commands global. This means any command you run will have these defaults. This could cause unintended issues as shell and other inspec commands will be getting exec defaults.

This change breaks out the defaults by command type so we only load the correct defaults.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -73,9 +73,11 @@ module Inspec
 
     def self.default_options
       {
-        color: true,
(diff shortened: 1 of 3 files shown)


adamleff (approved):
Way to think ahead, @jquick - thank you very much for this improvement. :)

arlimus (approved):
Kudos!

PR #2382, opened 2017-12-07
Rubocop: disable GuardClause and ClassLength

adamleff:
Neither of these cops is helping us write good code, and GuardClause specifically is actually contributing to us writing sometimes-unclear code. Disabling both of these cops and removing all unnecessary disable decorators in the codebase.


arlimus (approved):
Looks good

Issue #431, opened 2016-02-08
Bastion host


Fodoj:
What is the best way to forward specs through bastion host?

chris-rock:
@Fodoj If I get you right, you need some kind of gateway setting right?

Fodoj:
Yep

chris-rock:
InSpec has no native gateway setting yet. At this point, I propose to use ssh to bind the remote ssh port via the gateway ssh to a local port with `-L`. Then just let inspec connect to ssh://127.0.0.1:yourport

chris-rock:
For reference, in Serverspec it can be solved as described: http://pmyjavec.com/code/2014/12/03/serverspec-behind-jump-server.html

This feature should be implemented in https://github.com/chef/train/blob/master/lib/train/transports/ssh.rb

robcoward:
I'm already using the Net::SSH::Proxy::Command gem as mentioned in @chris-rock's link above to get serverspec to test infrastructure on remote non-routable networks via a bastion host. Would really love to see something similar usable with Inspec. 

My specific use-case here is working in a secure environment with different security domains, we are not allowed to route traffic between two security domains. All connections have to go through some form of jumpbox / proxy server where there is a protocol break and traffic can be "inspected". My secondary use-case is to be able to test infrastructure built on a cloud hosting provider where for good reasons, not all the 'internal' servers are accessible from the internet and have to also be accessed via a jumpbox.

zambien:
I would also like to see this added.

unitsche:
+1

thiagonache:
I did small patch, but I'm not sure how to get proxy variable from kitchen yml
https://github.com/chef/train/compare/master...thiagonache:master

can you guys help?

james-s-nduka:
+1
This is a feature Hashicorp Terraform has for remote execution on instances. It would be really helpful for inspec testing

jesstracy:
+1

heathsnow:
Here's what I've come up with so far:
```
task_thread = []
%w(port_map spec_test).each do |task|
  task_thread << Thread.new do
    case task
    when 'port_map'  
      puts 'mapping 1337 local to 22 remote'
      IO.popen('ssh -A -L1337:FOO_SERVER:22 ubuntu@BASTION_HOST')
    when 'spec_test'  
      puts 'running spec tests...'
[...]

PR #2385, opened 2017-12-10
Added support for proxy_command for remote SSH connections

cbeckr:
**NOTE** This is just a simple wrapper which relies on the underlying support being added to train: https://github.com/chef/train/pull/227
Solves #431.

Signed-off-by: Christian Becker <c.becker@mediaevent.services>

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -45,6 +45,8 @@ module Inspec
         desc: 'Allow remote scans with self-signed certificates (WinRM).'
       option :json_config, type: :string,
         desc: 'Read configuration from JSON file (`-` reads from stdin).'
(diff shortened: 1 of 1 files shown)


aaronlippold:
@chris-rock We may also want to expose and layer in the support for the test-kitchen `download` function as part of this effort as pushing the results from the bastion to the calling system would be a common use case. https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/transport/ssh.rb#L186

jquick:
Note: chef/train#227 was merged in recently. Once InSpec pins to the new version we can push this in.

jquick (approved):
Thanks @cbeckr  - This is now good to go!

miah (approved):

PR #2388, opened 2017-12-12
Enhance `package` resource error handling

### LINKED ISSUE #1692: NilClass errors when package info fails
Appears to be a bug in the `package` resource if the command used to find the package details fails.

```
              1) System Package amazon-ssm-agent should be installed
                   Failure/Error: info[:installed] == true

                        NoMethodError:
                               undefined method `[]' for nil:NilClass
```

`info` should return an empty hash until we have better exception handling.

### LINKED ISSUE #2373: Package resource passes when no packages are defined
### Description

When describing a package and not defining a package name, the inspec test will pass.

Confirmed on 1.45.13, CentOS 6.5

### Replication Case

Create a package resource with no packages defined:

```
describe package do
  it { should be_installed }
end
```

```
System Package
✔ should be installed
```

### Possible Solutions

I think, this should fail or generate an error that "package name" is undefined for "package" control/resource

### Stacktrace

No stack trace as test passes.

Thanks!

jerryaldrichiii:
This does the following:
  - Modifies `info` to return an empty hash instead of `nil`
  - Adds a failure case if package name is not specified
  - Adds a skip case if no package manager can be found
  - Changes `skip_resource` to use the `ResourceSkipped` exception
  - Raises an error if JSON parsing fails for Windows/MacOS/Darwin

This closes #1692
This closes #2373

### DIFF
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -21,6 +21,8 @@ module Inspec::Resources
     "
 
     def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize
+      raise Inspec::Exceptions::ResourceFailed, 'Package name required but not specified' if package_name.nil?
+
       @package_name = package_name
       @name = @package_name
       @cache = nil
@@ -45,7 +47,7 @@ module Inspec::Resources
       elsif ['hpux'].include?(os[:family])
         @pkgman = HpuxPkg.new(inspec)
       else
-        return skip_resource 'The `package` resource is not supported on your OS yet.'
+        raise Inspec::Exceptions::ResourceSkipped, 'The `package` resource is not supported on your OS yet.'
       end
 
       evaluate_missing_requirements
@@ -53,13 +55,11 @@ module Inspec::Resources
 
     # returns true if the package is installed
     def installed?(_provider = nil, _version = nil)
-      return false if info.nil?
       info[:installed] == true
     end
 
     # returns true it the package is held (if the OS supports it)
     def held?(_provider = nil, _version = nil)
-      return false if info.nil?
       return false unless info.key?(:held)
       info[:held] == true
     end
@@ -67,14 +67,12 @@ module Inspec::Resources
     # returns the package description
     def info
       return @cache if !@cache.nil?
-      return nil if @pkgman.nil?
       @pkgman.info(@package_name)
     end
 
     # return the package version
     def version
       info = @pkgman.info(@package_name)
-      return nil if info.nil?
       info[:version]
     end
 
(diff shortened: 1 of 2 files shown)


lib/resources/package.rb
```diff
@@ -20,7 +20,9 @@ class Package < Inspec.resource(1)
       end
     "
 
-    def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize
+    def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
```
adamleff: Maybe we should change this to:

```ruby
def initialize(package_name, opts = {})
```

... such that Ruby enforces that `package_name` must be specified? While I agree that the error message isn't as clear as the one you've provided, we should be consistent with other resources that have required parameters. 

We can always add some arity magic in the future with better error messaging, but for now, we should be consistent.

lib/resources/package.rb
```diff
@@ -45,36 +47,32 @@ def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize
       elsif ['hpux'].include?(os[:family])
         @pkgman = HpuxPkg.new(inspec)
       else
-        return skip_resource 'The `package` resource is not supported on your OS yet.'
+        raise Inspec::Exceptions::ResourceSkipped, 'The `package` resource is not supported on your OS yet.'
```
adamleff, line 48: Thanks for this - in the same vein as my prior comment, since we're adopting exceptions as the new normal, moving toward that is a good thing.

lib/resources/package.rb
```diff
@@ -45,36 +47,32 @@ def initialize(package_name = nil, opts = {}) # rubocop:disable Metrics/AbcSize
...
     # returns true if the package is installed
     def installed?(_provider = nil, _version = nil)
-      return false if info.nil?
       info[:installed] == true
     end
 
     # returns true it the package is held (if the OS supports it)
     def held?(_provider = nil, _version = nil)
-      return false if info.nil?
       return false unless info.key?(:held)
```
adamleff: I think we can remove this too and just it fall through to the boolean comparison, yes?

adamleff (changes requested):
This looks like some good cleanup, @jerryaldrichiii - just a couple of comments.

lib/resources/package.rb
jerryaldrichiii: Yup, I see we set it to a bool farther down. Good catch!

adamleff (approved):
Thanks for this improvement, @jerryaldrichiii.

arlimus (approved):
Fantastic improvement, thank you Jerry!

PR #2389, opened 2017-12-12
xml resource: update docs with alternate syntax for keys with periods

### LINKED ISSUE #2380: XML resource docs need updating: key with periods fails.
### Description

According to the current documentation, something like
``` 
describe xml('/data/backup/current/security.xml') do
    its('security/users/user/groups/org.artifactory.model.xstream.security.UserGroupImpl') { should include 'somestring' }
end
```
should function, but instead this results in an error trying to call 'artifactory' as a method in this case.
```
undefined method `artifactory' for []:Array
```

inspec 1.42.3
RHEL 7.4

### Replication Case

Try the xml resource with a leaf key that has periods.

### Possible Solutions

An alternate syntax, in this case:
```
its(['security/users/user/groups/org.artifactory.model.xstream.security.UserGroupImpl']) { should include 'somestring' }
```
works fine, but this is not documented on the website, therefore the behavior should be corrected (as a string should act like a string) or the docs should be updated with such an example using square brackets.

adamleff:
Fixes #2380

### DIFF
--- a/docs/resources/xml.md.erb
+++ b/docs/resources/xml.md.erb
@@ -38,6 +38,10 @@ where
 
 <br>
 
+In the event the path contains an element which contains periods, the alternate syntax can be used:
+
+    its(['root/name.with.a.period']) { should cmp 'so_many_dots' }
+
 ## Examples
 
 The following examples show how to use this InSpec audit resource.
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Thank you Adam!!

PR #2390, opened 2017-12-12
Replace exec --cache with --vendor-cache

jquick:
Fixes https://github.com/chef/inspec/issues/2339

This change deprecates the current `--cache` exec flag in favor of the `--vendor-cache` flag. Deprecation warnings have been added and code updated.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -64,7 +64,7 @@ module Inspec
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
       option :cache, type: :string,
-        desc: '[DEPRECATION] Please use --vendor-cache - this will be removed in InSpec 2.0'
+        desc: '[DEPRECATED] Please use --vendor-cache - this will be removed in InSpec 2.0'
       option :vendor_cache, type: :string,
         desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
       option :create_lockfile, type: :boolean,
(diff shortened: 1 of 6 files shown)


lib/inspec/base_cli.rb
```diff
@@ -64,6 +64,8 @@ def self.exec_options
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
       option :cache, type: :string,
+        desc: '[DEPRECATION] Please use --vendor-cache - this will be removed in InSpec 2.0'
```
adamleff: Can we change this to `[DEPRECATED]`? It's cleared to the users that it's deprecated, vs. us declaring that it's a deprecation. A nitpick, I know, but grammar matters :)

lib/inspec/cli.rb
```diff
@@ -155,6 +155,13 @@ def exec(*targets)
     o = opts(:exec).dup
     configure_logger(o)
 
+    # check for deprecated --cache
+    # TODO: REMOVE for inspec 2.0
+    if o.key?('cache')
+      o[:vendor_cache] = o[:cache]
+      o[:logger].warn '[DEPRECATION] The use of `--cache` is being deprecated in InSpec 2.0. Please use `--vendor-cache` instead.'
```
adamleff: Same here - `DEPRECATED` instead of `DEPRECATION`

adamleff (changes requested):
@jquick this looks great. Thanks for taking the time to cleanup the keys throughout the code as well.

Just a couple of grammar nits, but otherwise this is great.

adamleff (approved):

arlimus (approved):
Great to be so clear on this now! Kudos 😁

PR #2391, opened 2017-12-12
Split unit tests from functional

adamleff:
Functional tests can sometimes be picky. Moving them to their own matrix item so they're smaller and faster to re-run if needed.


arlimus (approved):
Love this change, small but effective 👍

PR #2392, opened 2017-12-12
Updates the docs for mysql_session

burtlo:
* Fixes usage of 'output' to 'stdout'
* Fixes the format of the examples to proper 4 space ident to become code blocks
* Adds examples for 'exit_status' and 'stderr'
* Modifies the matchers section to point to the command resource


burtlo:
I noticed this resource does not use params like some of the resources. I see a comment here https://github.com/chef/inspec/pull/1933#pullrequestreview-44674923 that talks about that being a nice to have. Happy to make that change but I don't know if that should be done with backwards compatibility in mind, deprecations, or simply break the order dependent parameters.

Until then, these docs better show how to work with the resource.

adamleff (approved):
Thanks, @burtlo!

PR #2393, opened 2017-12-13
Add platform resource and platform supports

### LINKED ISSUE #1740: inspec.yml supports: release wildcards/comparisons
We're trying to cover multiple "release" values without having to specify the exact release values when there are more than one.

RHEL 6 ends up looking like this:
supports:
  - os-family: redhat
    release: 6.4
  - os-family: redhat
    release: 6.5
  - os-family: redhat
    release: 6.6
  - os-family: redhat
    release: 6.7
  - os-family: redhat
    release: 6.8
  - os-family: redhat
    release: 6.9

Windows 2012
supports:
  - os-family: windows
    release: 6.2.9200
  - os-family: windows
    release: 6.3.9200
  - os-family: windows
    release: 6.3.9600

If it could support a wild card or comparisons
- os-family: redhat
   release: 6.*

- os-family: redhat
   release: < 6.1
   release: > 6.9

### LINKED ISSUE #1908: Double inheritance does not honor the os-family and controls are running for platform that does not pertain
### Description

Main Profile:

company_linux_profile
         inpec.yml file has the following:
                   depends:
   - name: company_rhel6_linux_compliance
     compliance: comp_user/company_rhel6_linux_compliance
   - name: company_rhel7_linux_compliance
     compliance: comp_user/company_rhel7_linux_compliance

Now if you look at the profiles company_rhel6_linux_compliance and company_rhel7_linux_compliance

profile inspec.yml  company_rhel6_linux_compliance
supports:
   - os-family: redhat
     release: 6.0
   - os-family: redhat
     release: 6.5
   - os-family: redhat
     release: 6.6
   - os-family: redhat
     release: 6.7
   - os-family: redhat
     release: 6.8
   - os-family: redhat
     release: 6.9

profile inspec.yml  company_rhel7_linux_compliance
supports:
   - os-family: redhat
     release: 7.0
   - os-family: redhat
     release: 7.1
   - os-family: redhat
     release: 7.2
   - os-family: redhat
     release: 7.3

This causes both platforms to run on a server that may not be an OS 6.X or 7.X.

inspec 1.26.0

### Replication Case

### Possible Solutions

os-family should be supported so you are only dealing with one profile with many versions of OS profiles.

### Stacktrace

### LINKED ISSUE #2188: inspec.yml should support which version of InSpec is required for a profile
### Description

When a new resource is added to InSpec it would be nice to be able to ensure that profiles that use it are not used with older releases of InSpec.

### Replication Case

Try using a new resource (like windows_hotfix) with an older version of InSpec

### Possible Solutions

Add ```inspec_version: '>1.39' ``` to the inspec.yml to use SemVer to check for versions.

### Stacktrace

```
./windows_hotfix_spec.rb:8:in `load_with_context': undefined method `windows_hotfix' for #<#<Class:0x007f80c6399a88>:0x007f80c6398e80> (NoMethodError)
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile_context.rb:146:in `instance_eval'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile_context.rb:146:in `load_with_context'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile_context.rb:130:in `load_control_file'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile.rb:151:in `block in collect_tests'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile.rb:148:in `each'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/profile.rb:148:in `collect_tests'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/runner.rb:89:in `block in load'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/runner.rb:78:in `each'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/runner.rb:78:in `load'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/runner.rb:99:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/base_cli.rb:83:in `run_tests'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/lib/inspec/cli.rb:159:in `exec'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/inspec-1.35.1/bin/inspec:12:in `<top (required)>'
	from /opt/chefdk/bin/inspec:264:in `load'
	from /opt/chefdk/bin/inspec:264:in `<main>'
```

### LINKED ISSUE #2404: Add supports keyword to InSpec resources and profile yaml
### Description

This change is part of the larger platforms RFC here: #1661. This change adds in the supports field for resources and profiles. 

## Platform in profiles

Profiles can specify a list of supported platforms. This is done via the `supports` field and includes:

* `os-name`: for OS-specific platform names
* `os-family`: for OS-specific platform families
* `platform`: to target platform families or names
* `platform-name`: for specific platform names
* `platform-family`: for general platform families
* `release`: for the platform version

The `support` entries are combined via `OR` and each entry is made up of fields combined via `AND`. Example:

```
# This profile will support any version of BSD and RHELv6.
# i.e. Platform  matches  (family: bsd) OR (name: redhat AND release: 6)
- supports:
  - os-family: bsd
  - os-name: redhat
    release: 6
```

## Platforms in resources

Resources can specify a list of supported platforms. This is done via the `supports` field. It behaves just like `supports` in `inspec.yml` (section above).

Due to the Ruby language, a few names needed adjustments: `os_name`, `os_family`, `platform`, `platform_name`, `platform_family`

Full example:

```
# This resource will support any version of BSD and RHELv6.
# i.e. Platform  matches  (family: bsd) OR (name: redhat AND release: 6)
class GordonConf < Inspec.resource(1)
  name 'gordon_config'
  supports os_family: 'bsd'
  supports os_name: 'redhat', release: 6
```

jquick:
This change adds the platform support options to resources and enhances the support options for profiles. This change is part of the larger platforms RFC here: https://github.com/chef/inspec/issues/1661
Fixes #2404

This change also:
Fixes https://github.com/chef/inspec/issues/1908
- The rework of platform supports fixes this issue.

Fixes https://github.com/chef/inspec/issues/1740
- The fix here was to add in bash wildcard for release.

Fixes https://github.com/chef/inspec/issues/2188
- Move the unused supports:inspec to a proper inspec_version check.

Examples of the new output:
screen shot 2017-12-19 at 9 40 35 am

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -123,12 +123,12 @@ module Inspec
 
           unless profile_context_owner.profile_supports_platform?
             platform = inspec.platform
-            msg = "Platform #{platform.name}/#{platform.release} is not supported with profile '#{profile_context_owner.profile_id}'"
+            msg = "Profile #{profile_context_owner.profile_id} is not supported on platform #{platform.name}/#{platform.release}."
             ::Inspec::Rule.set_skip_rule(control, msg)
           end
 
           unless profile_context_owner.profile_supports_inspec_version?
-            msg = "Current InSpec version #{Inspec::VERSION} is not supported with profile '#{profile_context_owner.profile_id}'"
+            msg = "Profile #{profile_context_owner.profile_id} is not supported on InSpec version (#{Inspec::VERSION})."
             ::Inspec::Rule.set_skip_rule(control, msg)
           end
 
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -121,6 +121,17 @@ module Inspec
             ::Inspec::Rule.set_skip_rule(control, true)
           end
 
+          unless profile_context_owner.profile_supports_platform?
+            platform = inspec.platform
+            msg = "Platform #{platform.name}/#{platform.release} is not supported with profile '#{profile_context_owner.profile_id}'"
+            ::Inspec::Rule.set_skip_rule(control, msg)
+          end
+
+          unless profile_context_owner.profile_supports_inspec_version?
+            msg = "Current InSpec version #{Inspec::VERSION} is not supported with profile '#{profile_context_owner.profile_id}'"
+            ::Inspec::Rule.set_skip_rule(control, msg)
+          end
+
           profile_context_owner.register_rule(control, &block) unless control.nil?
         end
 
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -117,7 +117,7 @@ module Inspec
         end
 
         define_method :register_control do |control, &block|
-          if @skip_file || !profile_context_owner.profile_supports_os?
+          if @skip_file
             ::Inspec::Rule.set_skip_rule(control, true)
           end
 
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -36,6 +36,7 @@ module Inspec
       summary
       description
       version
+      inspec_version
     }.each do |name|
       define_method name.to_sym do |arg|
         params[name.to_sym] = arg
@@ -53,8 +54,7 @@ module Inspec
     end
 
     def inspec_requirement
-      inspec = params[:supports].find { |x| !x[:inspec].nil? } || {}
-      Gem::Requirement.create(inspec[:inspec])
+      Gem::Requirement.create(params[:inspec_version])
     end
 
     def supports_runtime?
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -54,11 +54,13 @@ module Inspec
     end
 
     def inspec_requirement
-      inspec = params[:supports].find { |x| !x[:inspec].nil? }
-      if inspec
-        Inspec::Log.warn '[DEPRECATED] The use of inspec.yml `supports:inspec` is being deprecated in InSpec 2.0. Please use `inspec_version` instead.'
-        Gem::Requirement.create(inspec[:inspec])
+      inspec_in_supports = params[:supports].find { |x| !x[:inspec].nil? }
+      if inspec_in_supports
+        Inspec::Log.warn '[DEPRECATED] The use of inspec.yml `supports:inspec` is deprecated and will be removed in InSpec 2.0. Please use `inspec_version` instead.'
+        Gem::Requirement.create(inspec_in_supports[:inspec])
       else
+        # using Gem::Requirement here to allow nil values which
+        # translate to [">= 0"]
         Gem::Requirement.create(params[:inspec_version])
       end
     end
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -52,35 +52,6 @@ module Inspec
       # already.
     end
 
-    def is_supported?(os, entry)
-      name = entry[:'os-name'] || entry[:os]
-      family = entry[:'os-family']
-      release = entry[:release]
-
-      # return true if the backend matches the supported OS's
-      # fields act as masks, i.e. any value configured for os-name, os-family,
-      # or release must be met by the backend; any field that is nil acts as
-      # a glob expression i.e. is true
-
-      # os name is both saved in :family and :name, so check both
-      name_ok = name.nil? ||
-                os[:name] == name || os[:family] == name
-
-      family_check = family.to_s + '?'
-      family_ok = family.nil? || os[:family] == family ||
-                  (
-                    os.respond_to?(family_check) &&
-                    # this call will return true if the family matches
-                    os.method(family_check).call
-                  )
-
-      # ensure we do have a string if we have a non-nil value eg. 16.06
-      release_ok = release.nil? || os[:release] == release
-
-      # we want to make sure that all matchers are true
-      name_ok && family_ok && release_ok
-    end
-
     def inspec_requirement
       inspec = params[:supports].find { |x| !x[:inspec].nil? } || {}
       Gem::Requirement.create(inspec[:inspec])
@@ -91,18 +62,8 @@ module Inspec
       inspec_requirement.satisfied_by?(running)
     end
 
-    def supports_transport?(backend)
-      # with no supports specified, always return true, as there are no
-      # constraints on the supported backend; it is equivalent to putting
-      # all fields into accept-all mode
-      return true if params[:supports].empty?
-
-      found = params[:supports].find do |entry|
-        is_supported?(backend.os, entry)
-      end
-
-      # finally, if we found a supported entry, we are good to go
-      !found.nil?
+    def supports_platform?(backend)
+      backend.platform.supported?(params[:supports])
     end
 
     # return all warn and errors
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -54,7 +54,13 @@ module Inspec
     end
 
     def inspec_requirement
-      Gem::Requirement.create(params[:inspec_version])
+      inspec = params[:supports].find { |x| !x[:inspec].nil? }
+      if inspec
+        Inspec::Log.warn '[DEPRECATED] The use of inspec.yml `supports:inspec` is being deprecated in InSpec 2.0. Please use `inspec_version` instead.'
+        Gem::Requirement.create(inspec[:inspec])
+      else
+        Gem::Requirement.create(params[:inspec_version])
+      end
     end
 
     def supports_runtime?
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -28,6 +28,12 @@ module Inspec
       __resource_registry[@name].desc(description)
     end
 
+    def supports(criteria = nil)
+      return if criteria.nil?
+      Inspec::Resource.supports[@name] ||= []
+      Inspec::Resource.supports[@name].push criteria
+    end
+
     def example(example = nil)
       return if example.nil?
       __resource_registry[@name].example(example)
@@ -37,17 +43,21 @@ module Inspec
       Inspec::Resource.registry
     end
 
-    def __register(name, obj) # rubocop:disable Metrics/MethodLength
-      cl = Class.new(obj) do
+    def __register(name, obj) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+      cl = Class.new(obj) do # rubocop:disable Metrics/BlockLength
         attr_reader :resource_exception_message
 
         def initialize(backend, name, *args)
           @resource_skipped = false
           @resource_failed = false
+          @supports = Inspec::Resource.supports[name]
 
           # attach the backend to this instance
           @__backend_runner__ = backend
           @__resource_name__ = name
+
+          # check resource supports
+          check_supports unless @supports.nil?
 
           # call the resource initializer
           begin
@@ -67,6 +77,12 @@ module Inspec
         def self.example(example = nil)
           return @example if example.nil?
           @example = example
+        end
+
+        def check_supports
+          status = inspec.platform.supported?(@supports)
+          skip_msg = "Platform #{inspec.platform.name}/#{inspec.platform.release} is not supported with resource [#{@__resource_name__}]"
+          fail_resource(skip_msg) unless status
         end
 
         def skip_resource(message)
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -81,8 +81,8 @@ module Inspec
 
         def check_supports
           status = inspec.platform.supported?(@supports)
-          skip_msg = "Platform #{inspec.platform.name}/#{inspec.platform.release} is not supported with resource [#{@__resource_name__}]"
-          fail_resource(skip_msg) unless status
+          skip_msg = "Platform #{inspec.platform.name}/#{inspec.platform.release} is not supported with resource '#{@__resource_name__.capitalize}'"
+          skip_resource(skip_msg) unless status
         end
 
         def skip_resource(message)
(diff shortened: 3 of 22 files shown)


jquick:
If the undocumented:
```
supports:
  - inspec: ">= 1.0.0"
```
Is actually being used by anyone, we can add a deprecation warning and remove in 2.0.

lib/inspec/control_eval_context.rb
```diff
@@ -117,10 +117,21 @@ def to_s
...
 
         define_method :register_control do |control, &block|
-          if @skip_file || !profile_context_owner.profile_supports_os?
+          if @skip_file
             ::Inspec::Rule.set_skip_rule(control, true)
           end
 
+          unless profile_context_owner.profile_supports_platform?
+            platform = inspec.platform
+            msg = "Platform #{platform.name}/#{platform.release} is not supported with profile '#{profile_context_owner.profile_id}'"
```
adamleff: "supported with" feels like it might not be the right English to use. I'm wondering if this might be better:

`"Profile #{profile_context_owner.profile_id} is not supported on platform #{platform.name}/#{platform.release}."`

lib/inspec/control_eval_context.rb
```diff
@@ -117,10 +117,21 @@ def to_s
...
+            ::Inspec::Rule.set_skip_rule(control, msg)
+          end
+
+          unless profile_context_owner.profile_supports_inspec_version?
+            msg = "Current InSpec version #{Inspec::VERSION} is not supported with profile '#{profile_context_owner.profile_id}'"
```
adamleff: Same feedback on this one... "Profile blah is not supported on the current InSpec version (#{Inspec::VERSION})."

lib/inspec/metadata.rb
```diff
@@ -52,57 +53,23 @@ def supports(sth, version = nil)
...
-      release_ok = release.nil? || os[:release] == release
-
-      # we want to make sure that all matchers are true
-      name_ok && family_ok && release_ok
-    end
-
     def inspec_requirement
-      inspec = params[:supports].find { |x| !x[:inspec].nil? } || {}
-      Gem::Requirement.create(inspec[:inspec])
+      inspec = params[:supports].find { |x| !x[:inspec].nil? }
```
adamleff: Given that we overload the use of the keyword `inspec` in a lot of places, I'm going to engage SUPER PEDANT MODE and suggest that this variable be `inspec_in_supports` or something clearer.

lib/inspec/metadata.rb
```diff
@@ -52,57 +53,23 @@ def supports(sth, version = nil)
...
+      if inspec
+        Inspec::Log.warn '[DEPRECATED] The use of inspec.yml `supports:inspec` is being deprecated in InSpec 2.0. Please use `inspec_version` instead.'
```
adamleff: Technically, we're deprecating it now, and it will be removed in 2.0.

lib/inspec/metadata.rb
```diff
@@ -52,57 +53,23 @@ def supports(sth, version = nil)
...
+        Gem::Requirement.create(inspec[:inspec])
+      else
+        Gem::Requirement.create(params[:inspec_version])
```
adamleff: It might be worth a comment here for the next reader who is not familiar with `Gem::Requirement.create` that if you pass in `nil`, it will create a constraint of `[">= 0"]` so it's totally OK for this to be `nil`.

lib/inspec/plugins/resource.rb
```diff
@@ -28,6 +28,12 @@ def desc(description = nil)
       __resource_registry[@name].desc(description)
     end
 
+    def supports(criteria = nil)
+      return if criteria.nil?
+      Inspec::Resource.supports[@name] ||= []
+      Inspec::Resource.supports[@name].push criteria
```
adamleff: Can you wrap the parameter in parentheses so we remain consistent in the codebase?

```ruby
Inspec::Resource.supports[@name].push(criteria)
```

lib/inspec/plugins/resource.rb
```diff
@@ -69,6 +79,12 @@ def self.example(example = nil)
           @example = example
         end
 
+        def check_supports
+          status = inspec.platform.supported?(@supports)
+          skip_msg = "Platform #{inspec.platform.name}/#{inspec.platform.release} is not supported with resource '#{@__resource_name__.capitalize}'"
```
adamleff: Same feedback as above re: the message structure.

lib/resources/platform.rb
```diff
@@ -0,0 +1,92 @@
...
+module Inspec::Resources
+  class PlatformResource < Inspec.resource(1)
+    name 'platform'
+    desc 'Use the platform InSpec resource to test the platform on which the system is running.'
+    example "
+      describe platform.name do
+        it { should eq 'redhat' }
+      end
+
+      describe platform.family?('unix') do
```
adamleff: I wonder if we should try to exploit RSpec's magic treatment of methods with question marks and see if we can't get it to be like this to make it more readable and more similar to existing resources:

```ruby
describe platform do
  it { should be_in_family('unix') }
end
```

That should work if we create a method `def in_family?(family_name)`, right? or am I making that up?

test/unit/resources/platform_test.rb
```diff
@@ -0,0 +1,99 @@
+# encoding: utf-8
+# author: Christoph Hartmann
```
adamleff: We can yank the author lines here.

test/unit/resources/platform_test.rb
```diff
@@ -0,0 +1,99 @@
...
+# author: Dominik Richter
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::Os' do
```
adamleff: Looks like you're testing the OS resource, but the file is named `platform_test`?

adamleff (changes requested):

jquick:
@adamleff  - Thanks for the feedback. These issues have been addressed and a new output has been posted to the description.

test/unit/resources/platform_test.rb
```diff
@@ -0,0 +1,97 @@
+# encoding: utf-8
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::Platform' do
+  let(:resource) { resource = MockLoader.new(:ubuntu1504).load_resource('os') }
```
adamleff: Should this be `load_resource('platform')`?

adamleff: or rather, should we actually be doing this twice... once for platform, and once for os?

jquick: Yeah this should be platform instead of the os subclass.

adamleff (approved):

lib/resources/platform.rb
```diff
@@ -0,0 +1,92 @@
+# encoding: utf-8
+
+module Inspec::Resources
+  class PlatformResource < Inspec.resource(1)
+    name 'platform'
+    desc 'Use the platform InSpec resource to test the platform on which the system is running.'
+    example "
+      describe platform.name do
+        it { should eq 'redhat' }
```
chris-rock: i think this should be
```
describe platform do 
 its('name') { should eq 'redhat' } 
```

lib/resources/platform.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+
+      describe platform do
+        it { should be_in_family('unix') }
+      end
+    "
+
+    def initialize
+      @platform = inspec.backend.os
```
chris-rock, line 18: should this be `inspec.backend.platform`?

test/unit/mock/profiles/supported_inspec/inspec.yml
```diff
@@ -1,2 +1 @@
-supports:
-- inspec: 0.18
+inspec_version: 0.18
```
chris-rock, line 1: What was the reason behind this change? I cannot remember.

chris-rock (commented):
Thank you @jquick This is a great move forward. I added some comments.

lib/resources/platform.rb
jquick: Yeah it should, for now this breaks unit tests as the mock transport in train only mocks up os. I'll put a PR out for train to fix the mock transport and we can update this after.
https://github.com/chef/train/pull/230

test/unit/mock/profiles/supported_inspec/inspec.yml
jquick: We wanted supports to only have platform conditions in it. So we didn't have a mixture.

lib/inspec/metadata.rb
```diff
@@ -52,57 +53,25 @@ def supports(sth, version = nil)
...
-      return true if params[:supports].empty?
-
-      found = params[:supports].find do |entry|
-        is_supported?(backend.os, entry)
-      end
-
-      # finally, if we found a supported entry, we are good to go
-      !found.nil?
+    def supports_platform?(backend)
+      backend.platform.supported?(params[:supports])
```
chris-rock, line 74: I am a little bit confused by that. We have a backend (aka train connection) that exposes platform and we have the platform resource. It is really difficult to differentiate this now.

test/unit/mock/profiles/supported_inspec/inspec.yml
chris-rock: @arlimus Is that matching your expectations?

lib/resources/platform.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def in_family?(family)
+      @platform.family_hierarchy.include?(family)
+    end
+
+    def families
+      @platform.family_hierarchy
+    end
+
+    def supported?(supports)
```
chris-rock, line 47: I really like that we have one method that rules them all. I am just not sure if this is not better suited for train than this resource.

lib/inspec/metadata.rb
jquick: This is very valid, we have:
`backend.platform` -> resource
`backend.backend.platform` -> train platform object

@adamleff  what do you think about this?

lib/resources/platform.rb
jquick: Yeah it could live in either. The only issue with train is we would be sending InSpec specific keys to the support library.

test/unit/mock/profiles/supported_inspec/inspec.yml
adamleff: I'd like to add a 👍  to this change to make is super-clear. Especially since we don't want resource authors to add an inspec version check to the `supports` inside a resource definition itself, because that would make zero sense. Separating it and isolating it to the `inspec.yml` is cleaner, IMO.

lib/inspec/metadata.rb
adamleff: Naming things is hard, and we've abused some of that terminology because of the origin of some of the Train code (coming from Test Kitchen, where some of the words mean different things than how we're using them).

I honestly can't think of a different way of differentiating them, and I think it's no less confusing than us having the `os` resource and the `os` exposed by Train. These are implementation details for authors of InSpec than they are for the consumers of it.

@chris-rock do you have suggestions on ways to clarify it?

lib/resources/platform.rb
adamleff: I think InSpec is the right place for this logic to live. I see Train as the thing that surfaces the information about the platform, and provides the means to connect to it. Those that consume that information can decide whether what they intend to do is supported by the information Train has surfaced.

If @chris-rock feels strongly about moving this logic into Train, then this method could be transformed to something that puts together a cleaned-up hash of the Train-specific keys, and then passes them to a `supported?` method on the Train side.

However, I still think this lives within InSpec.

test/unit/mock/profiles/supported_inspec/inspec.yml
arlimus: I'm 👍 on changing this here.

The resource  cross-dependency issue looks broad and should be tackled separately

lib/resources/platform.rb
arlimus: Looks good here 👍 Agreed it could live in either and I like Adam's explanation too.

lib/inspec/metadata.rb
arlimus: I think it's ok for now. Yeah, it is confusing and we also had it with `os` unfortunately. Unless we find something better of course 😁

arlimus (approved):
Sweet work @jquick thank you so much!!

Kudos for your reviews @adamleff and @chris-rock

PR #2395, opened 2017-12-13
docs: fix XML resource example to use alternate syntax

adamleff:


chris-rock (approved):
Thank you @adamleff

PR #2397, opened 2017-12-14
Deprecate and warn when comparing against OS name with capitals/spaces

### LINKED ISSUE #2407: Force lowercase/underscore platform names
### Description

https://github.com/chef/train/pull/228 now converts all platform names to lowercase and underscores. This change is a breaking change for InSpec. We need to update the os/platform code to auto correct this for profiles with a warning saying this is going to be finalized in InSpec 2.0.

jquick:
This change supports the new train lowercase/underscored platform names. https://github.com/chef/train/pull/228

Fixes https://github.com/chef/inspec/issues/2407

This change will auto convert any profiles using the wrong format and give a warning to fix by 2.0.

### DIFF
--- a/lib/resources/os.rb
+++ b/lib/resources/os.rb
@@ -36,17 +36,19 @@ module Inspec::Resources
         if other =~ /[A-Z ]/
           cleaned = other.downcase.tr(' ', '_')
           Inspec::Log.warn "[DEPRECATED] Platform names will become lowercase in InSpec 2.0. Please match on '#{cleaned}' instead of '#{other}'"
+          super(cleaned)
+        else
+          super(other)
         end
-        super(other)
       end
     end
 
-    def [](name)
+    def [](key)
       # convert string to symbol
-      name = name.to_sym if name.is_a? String
-      return self.name if name == :name
+      key = key.to_sym if key.is_a? String
+      return name if key == :name
 
-      inspec.backend.os[name]
+      inspec.backend.os[key]
     end
 
     def name
--- a/lib/resources/os.rb
+++ b/lib/resources/os.rb
@@ -29,44 +29,10 @@ module Inspec::Resources
       end
     end
 
-    # This is a string override for os.name.
-    # TODO: removed in inspec 2.0
-    class NameCleaned < String
-      def ==(other)
-        if other =~ /[A-Z ]/
-          cleaned = other.downcase.tr(' ', '_')
-          Inspec::Log.warn "[DEPRECATED] Platform names will become lowercase in InSpec 2.0. Please match on '#{cleaned}' instead of '#{other}'"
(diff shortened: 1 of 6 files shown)


jquick:
@adamleff - I think this is ready for review, it will want to go in tandem with the train release. https://github.com/chef/train/pull/228

Note: I will redo this to be off the platform resource once that gets merged in.

lib/resources/os.rb
```diff
@@ -27,15 +27,33 @@ class OSResource < Inspec.resource(1)
...
 
+    # This is a string override for os.name.
+    # TODO: removed in inspec 2.0
+    class NameCleaned < String
+      def ==(other)
+        if other =~ /[A-Z ]/
+          cleaned = other.downcase.tr(' ', '_')
+          Inspec::Log.warn "[DEPRECATED] Platform names will become lowercase in InSpec 2.0. Please match on '#{cleaned}' instead of '#{other}'"
+        end
+        super(other)
```
adamleff: We want to compare against `cleaned` if we had to translate, right?

lib/resources/os.rb
```diff
@@ -27,15 +27,33 @@ class OSResource < Inspec.resource(1)
...
+      end
+    end
+
     def [](name)
       # convert string to symbol
       name = name.to_sym if name.is_a? String
+      return self.name if name == :name
```
adamleff: For clarity purposes, we should probably change the method parameter from `name` to `key` or something like that so we're not overloading the use of `name` here.

adamleff (changes requested):
Can we add a couple of unit tests for `NameCleaned`? I'm thinking something for when translation was necessary, and something for when it wasn't? Just some checks using the `==` method you define should be good.

test/unit/resources/os_test.rb
```diff
@@ -6,6 +6,24 @@
 require 'inspec/resource'
 
 describe 'Inspec::Resources::Os' do
+  it 'verify NamedCleaned with uppercase' do
```
adamleff: Forgive the pun, but can we `Clean` this up a bit?

```ruby
let(:cleaner) { Inspec::Resources::OSResource::NameCleaned }

it 'does something' do
  cleaner.new('blah')
  # some test here
end
```

It feels repetitious to be defining `cleaner` in each test like that. If you don't want to use a `let`, we should at least consolidate two lines per test:

```ruby
it 'does something' do
  name = Inspec::Resources::OSResource::NameCleaned.new('something')
  # the test
end
```

adamleff (changes requested):

adamleff (approved):
Conditionally approved pending a train release and a commit here that uses it.

adamleff (approved):

PR #2400, opened 2017-12-15
add iis_app_pool resource

strocknar:
Adds a resource to check AppPool settings in IIS


lib/resources/iis_app_pool.rb
```diff
@@ -0,0 +1,97 @@
...
+      its('runtime_version') { should eq 'v4.0' }
+      its('pipeline_mode') { should eq 'Integrated' }
+    end
+  "
+
+  def initialize(pool_name)
+    @pool_name = pool_name
+    @pool_path = "IIS:\\AppPools\\#{@pool_name}"
+    @cache = nil
+    @inspec = inspec
```
adamleff: You shouldn't need to put this in an instance variable. Methods within this resource should be able to simply call `inspec.command` and similar directly, without using `@inspec`

lib/resources/iis_app_pool.rb
```diff
@@ -0,0 +1,97 @@
...
+  def username
+    iis_app_pool[:username]
+  end
+
+  def exists?
+    !iis_app_pool[:pool_name].empty?
+  end
+
+  def to_s
+    "iis_app_pool '#{@pool_path}'"
```
adamleff: Since we never dynamically change the `@pool_path` based on later interactions (i.e. it's aways `IIS:\\AppPools\whatever_the_user_supplied`), maybe we should just make this `"iis_app_pool #{@pool_name}"` for readability?

It feels redundant to have `iis_app_pool` and `IIS:\\AppPools` right next to each other, I think.

lib/resources/iis_app_pool.rb
```diff
@@ -0,0 +1,97 @@
...
+  def iis_app_pool
+    return @cache unless @cache.nil?
+
+    command = "Import-Module WebAdministration; Get-Item '#{@pool_path}' | Select-Object * | ConvertTo-Json"
+    cmd = @inspec.command(command)
+
+    begin
+      pool = JSON.parse(cmd.stdout)
+    rescue JSON::ParserError => _e
+      return {}
```
adamleff: I think it might be better to fail the resource here with a good error message:

```ruby
raise Inspec::Exceptions::ResourceFailed, "Unable to parse app pool JSON"
```

lib/resources/iis_app_pool.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    # map our values to a hash table
+    info = {
+      pool_name: pool['name'],
+      version: pool['managedRuntimeVersion'],
+      e32b: pool['enable32BitAppOnWin64'],
+      mode: pool['managedPipelineMode'],
+      processes: pool['processModel']['maxProcesses'],
+      # Has to be a separate command to get "XX:XX:XX" format
+      timeout: @inspec.command("Get-ItemProperty #{@pool_path} processmodel.idletimeout | select-object value").stdout,
```
adamleff: Sorry, I know so little about Windows... how does it come back natively?  Can we do a conversion in Ruby instead? I'd be happy to help.

lib/resources/iis_app_pool.rb
```diff
@@ -0,0 +1,97 @@
...
+      user_identity_type: pool['processModel']['identityType'],
+      username: pool['processModel']['userName'],
+    }
+
+    @cache = info unless info.nil?
```
adamleff: `info` can never be `nil` if you've made it this far. I would just change line 83 to be `@cache = { ... `

adamleff (changes requested):
@strocknar hello! Thanks for your first contribution to InSpec, and welcome to the community!

So happy to see yet another Windows resource addition to InSpec. This is great. I've left you some comments and suggestions.

Also, we need to add some unit tests for this. The process will be to write some tests in `test/unit/resources`, mock out the commands your tests would run in `test/helper.rb` and add the mock output to `test/unit/mock`. There's a lot of prior art in there, but I'm happy to help you in any way I can. Just give us a shout here if you're stuck.

Again, great first submission! We're looking forward to helping you push it across the finish line.

lib/resources/iis_app_pool.rb
strocknar: Removed this.

lib/resources/iis_app_pool.rb
strocknar: line 83 changed

lib/resources/iis_app_pool.rb
strocknar: I changed this to build it locally as opposed to my hacky string. Also added in settings for days, hours, minutes, seconds to test those individually if needed.

lib/resources/iis_app_pool.rb
strocknar: Raised exception instead of returning nothing

lib/resources/iis_app_pool.rb
strocknar: change to use `@pool_name` instead.

adamleff:
@strocknar this looks pretty solid! However, before we can accept this into InSpec, we need to add some unit tests as I mentioned before. Do you need some additional guidance on how to do so?

strocknar:
@adamleff I have not yet had a chance to look into the unit tests. It's been a busy couple weeks. I'm hoping to get into them this week, though.

adamleff:
No problem, @strocknar! No rush, and we appreciate your contribution. Let us know if you get stuck.

strocknar:
@adamleff What's the best way to run these tests so that I can verify that what I'm doing is actually correct?

adamleff:
@strocknar from the top-level directory of your locally-cloned InSpec repo:

 * `bundle install`
 * `bundle exec m test/unit/resource/iis_app_pool_test.rb` (or whatever the file is named

That will run the unit tests in that particular file. To run the full suite, you can type: `bundle exec rake test`

You can/should also run `bundle exec rake lint` before commit/pushing to validate that all your changes pass lint testing too! Hope that helps!

strocknar:
@adamleff I'm obviously doing something wrong, but I can't quite put my finger on what it is. My super simple beginning `iis_app_pool_test.rb` is below, but when I run it I get an error. I based it off of the `wmi_test.rb` (which runs fine) and it looks pretty much identical. Any tips on why this might be failing to find the `new` method in `MockLoader` (as that is what it looks like is happening to me)?

ERROR:
```
  1) Error:
Inspec::Resources::IisAppPool#test_0001_verify default app pool settings:
NoMethodError: undefined method `new' for nil:NilClass
    /Users/rvankleeck/projects/inspec/test/helper.rb:481:in `load_resource'
    /Users/rvankleeck/projects/inspec/test/unit/resources/iis_app_pool_test.rb:9:in `block (2 levels) in <top (required)>'
```

IIS_APP_POOL_TEST.RB
```ruby
require 'helper'
require 'inspec/resource'

describe 'Inspec::Resources::IisAppPool' do
  it 'verify default app pool settings' do
    resource = MockLoader.new(:windows).load_resource('iis_app_pool', 'DefaultAppPool')
    _(resource.send('to_s')).must_equal 'iis_app_pool \'DefaultAppPool\''
  end
end
```

jquick:
Hey @strocknar ! Looked into this a bit. The issue here is not your test!. You need to require your new resource in `inspec/lib/inspec/resource.rb`. You can just add it under the other `iis_app`:

`require 'resources/iis_app_pool'`

This should get your resource into the mix and your test should then work. Let me know if I can assist more!

strocknar:
@adamleff @jquick 

Thanks for the assistance. I have added mock tests as requested.

jquick (approved):
Thanks @strocknar !

chris-rock (approved):
Thank you @strocknar Great work

PR #2401, opened 2017-12-15
Fix OWCA detection for `compliance login`

### LINKED ISSUE #2375: Inspec Compliance Login Failures on "Opsworks Automate"
### Description

When attempting to login to the Inspec Compliance server at AWS (it's a fresh Opsworks + Chef Automate build) I receive the following error which has no information on google (yet).

```/Users/josiah/.rvm/gems/ruby-2.4.0/gems/inspec-1.47.0/lib/bundles/inspec-compliance/api/login.rb:24:in `login': Unable to determine if https://chef-opsworks-<redacted>.us-west-2.opsworks-cm.io is a Chef Automate or Chef Compliance server (Compliance::API::Login::CannotDetermineServerType)```

Mac OS Sierra 10.12.1
Inspec 1.47.0

### Replication Case

- Install gem "inspec", this happens on both v1.43.8 and when I updated to the latest Inspec v1.47.0
- Let Opsworks on AWS provision you an Opsworks + Chef Automate server
- Note: I have added "Chef Manage" web interface to the server via CLI
- Attempt to login via CLI:
```inspec compliance login https://chef-opsworks-<redacted>.us-west-2.opsworks-cm.io --insecure --user='admin' --token='<redacted>' --ent='default'```

### Possible Solutions

The `inspec compliance login` command attempts to hit up to two endpoints to figure out if what you’re asking us to login to is an older-generation Chef Compliance server, or a newer Chef Automate server. Given that you are referring to an Opsworks instance, I believe, it’s possible that those endpoints are filtered or returning data we don’t expect.

### Stacktrace

https://gist.github.com/dcijj/9a25994f05999ed744894259276a5760

### PR DESCRIPTION
**Pull request by jerryaldrichiii**
This adds detection logic for logging into OpsWorks Chef Automate via `inspec compliance login`.

Currently, the OWCA `/compliance/version` endpoint returns a 200 code instead of a 401.  This is due to a redirect when doing a GET request without authentication headers. This may change in the future, but I recommend leaving this in to support older versions.

Would welcome thoughts on the matter though.

Fixes #2375

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -270,7 +270,7 @@ module Compliance
           "Received 401 from #{url}#{automate_endpoint} " \
           'assuming target is a Chef Automate instance',
         )
-        return true
+        true
       when '200'
         # Chef Automate currently returns 401 for `/compliance/version` but some
         # versions of OpsWorks Chef Automate return 200 and a Chef Manage page
@@ -280,15 +280,21 @@ module Compliance
             "Received 200 from #{url}#{automate_endpoint} " \
             'assuming target is an OpsWorks Chef Automate instance',
           )
-          return true
+          true
         else
           Inspec::Log.debug(
             "Received 200 from #{url}#{automate_endpoint} " \
             'but did not receive the Chef Manage page. ' \
             'Continuing with detection attempts',
           )
-          return false
+          false
         end
+      else
+        Inspec::Log.debug(
+          "Received neither 200 nor 401 from #{url}#{automate_endpoint}. " \
+          'Continuing with detection attempts',
+        )
+        false
       end
     end
 
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -251,11 +251,28 @@ module Compliance
     end
 
     def self.determine_server_type(url, insecure)
-      if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
-        :automate
-      elsif Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
-        :compliance
+      automate_endpoint = '/compliance/version'
+      response = Compliance::HTTP.get(url + automate_endpoint, nil, insecure)
+      case response.code
+      when '401'
+        return :automate
+      when '200'
+        # Chef Automate currently returns 401 for `/compliance/version` but some
+        # versions of OpsWorks Chef Automate return 200 and a Chef Manage page
+        # when unauthenticated requests are received.
+        if response.body.include?('Are You Looking For the Chef Server?')
+          Inspec::Log.debug(
+            "Received 200 from #{url}#{automate_endpoint} " \
+            'assuming target is an OpsWorks Chef Automate instance',
+          )
+          return :automate
+        end
       end
+
+      # All versions of Chef Compliance return 200 for `/api/version`
+      compliance_endpoint = '/api/version'
+      response = Compliance::HTTP.get(url + compliance_endpoint, nil, insecure)
+      return :compliance if response.code == '200'
     end
   end
 end
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -251,10 +251,14 @@ module Compliance
     end
 
     def self.determine_server_type(url, insecure)
-      return :automate if target_is_automate_server?(url, insecure)
-      return :compliance if target_is_compliance_server?(url, insecure)
-      Inspec::Log.debug('Could not determine server type using known endpoints')
-      nil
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/unit/bundles/inspec-compliance/api_test.rb [hunk 1]
```diff
@@ -247,32 +247,65 @@
...
+        .with(url + automate_endpoint, headers, insecure)
+        .returns(good_response)
+
+      Compliance::API.determine_server_type(url, insecure).must_equal(:automate)
+    end
+
+    # Chef Automate currently returns 401 for `/compliance/version` but some
+    # versions of OpsWorks Chef Automate return 200 and a Chef Manage page when
+    # unauthenticated requests are received.
+    it 'returns `:automate` when a 200 is received from `https://URL/compliance/version`' do
```
adamleff, line 273: Can we add a test that checks that it returns `nil` if we hit the server, get a 200, but do not get the Chef Manage page? That's the other edge case here.

adamleff (changes requested):

test/unit/bundles/inspec-compliance/api_test.rb
jerryaldrichiii: Can do! Great suggestion.

jerryaldrichiii:
@adamleff I added some debug messages for `determine_server_type`, but it made the method a bit wordy in my opinion so I split out the logic. Let me know if you want me to drop those commits and put the debug messages with `determine_server_type`.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -251,11 +251,55 @@ def self.server_version_from_config(config)
     end
 
     def self.determine_server_type(url, insecure)
-      if Compliance::HTTP.get(url + '/compliance/version', nil, insecure).code == '401'
-        :automate
-      elsif Compliance::HTTP.get(url + '/api/version', nil, insecure).code == '200'
-        :compliance
+      return :automate if target_is_automate_server?(url, insecure)
```
adamleff: Given that this is a pretty short method now that you've broken things out, I think it's totally fine to leave this as an `if...elsif` structure. No short-circuiting needed. In fact, I think it reads cleaner:

```ruby
if target_is_automate_server?(url, insecure)
  :automate
elsif target_is_compliance_server?(url, insecure)
  :compliance
else
  Inspec::Log.debug('Could not determine server type using known endpoints')
  nil
end
```

lib/bundles/inspec-compliance/api.rb
```diff
@@ -251,11 +251,55 @@ def self.server_version_from_config(config)
...
+          return true
+        else
+          Inspec::Log.debug(
+            "Received 200 from #{url}#{automate_endpoint} " \
+            'but did not receive the Chef Manage page. ' \
+            'Continuing with detection attempts',
+          )
+          return false
+        end
       end
```
adamleff: We'll return `nil` if we get a non-200 / non-401 status code. I think we should have a default `else` in the `case` statement that's just `false`

And if you do that, you can eliminate the `return` statements in 269, 279, and 286 and just have them be `true` or `false` accordingly. The `return`s are superfluous.

adamleff (changes requested):
Just a few cleanup items and this should be good to go.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Ah! Great catch. I will make it so.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: I agree, will fix.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -276,15 +280,21 @@ def self.target_is_automate_server?(url, insecure)
...
             'but did not receive the Chef Manage page. ' \
             'Continuing with detection attempts',
           )
-          return false
+          false
         end
+      else
+        Inspec::Log.debug(
+          "Received neither 200 nor 401 from #{url}#{automate_endpoint}. " \
+          'Continuing with detection attempts',
```
adamleff: This assumes that whatever called this method has additional detection attempts to make :)

A better message, to keep it contained, would be `"Received unexpected status code #{response.code} from #{url}#{automate_endpoint} - assuming this is not an Automate server."`

... and whoever is calling can determine whether there are additional detection attempts to make, and now the user gets to see what status code they actually got.

adamleff (changes requested):
One small suggested change to a debug log message.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Ah, great catch. Fixing now.

adamleff (approved):

chris-rock (approved):
Thank you @jerryaldrichiii for working out the details!

PR #2402, opened 2017-12-15
Add option to allow SimpleConfig to strip trailing whitespace

miah:
This PR Updates `SimpleConfig` to allow users to strip trailing whitespace from parsed values. The `:strip_values` option is defaulted to false, users must explicitly enable it.

Note the whitespace after www.inspec.test:
```
SimpleConfig.new('ServerAlias www.inspec.test ', assignment_regex: /^\s*(\S+)\s+(.*)\s*$/, multiple_values: true, strip_values: true)
```

Will result in params including `{ 'ServerAlias' => ['www.inspec.test'] }`

This PR also updates the `apache_conf` resource to enable `:strip_values` by default.

Because Apache allows you to pass multiple values, and define the setting more than once you can still end up with parameters like this:

```
# apache2.conf
ServerAlias io.inspec.test inspec.test
ServerAlias www.inspec.test
```

`{ 'ServerAlias' => ['io.inspec.test inspec.test', 'www.inspec.test'] }`

But they will no longer include trailing whitespace =)

Fixes https://github.com/chef/inspec/issues/1241

### DIFF
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -83,6 +83,7 @@ module Inspec::Resources
           raw_conf,
           assignment_regex: /^\s*(\S+)\s+(.*)\s*$/,
           multiple_values: true,
+          strip_values: true,
         ).params
         @params.merge!(params)
 
--- a/lib/utils/simpleconfig.rb
+++ b/lib/utils/simpleconfig.rb
@@ -74,6 +74,7 @@ class SimpleConfig
     if opts[:multiple_values]
       @vals[m[1]] ||= []
       @vals[m[1]].push(parse_values(m, opts[:key_values]))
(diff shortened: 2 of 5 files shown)


adamleff:
So, I'm a bit on-the-fence on this change. Part of me thinks that this feature shouldn't be necessary; the assignment_regex should just be written properly to not be greedy and pull in spaces. Am I crazy for thinking that?

Also, I think I'm going to engage SUPER PEDANT MODE and suggest that (if we proceed with this change) the parameter be named `strip_whitespace` to make it clearer to the user what the outcome will be.

Let's chat about this sometime tomorrow if you don't mind!

miah:
I was also on the fence with this change. I worked with @jerryaldrichiii on a regex solution to this. New PR incoming =)

adamleff:
@miah thank you for going back and taking a second look at it after the initial feedback!

PR #2403, opened 2017-12-15
Add support for Darwin Directory Service groups

### LINKED ISSUE #2330: group resource is broken on macOS
### Description

The group resource looks at /etc/group like a normal UNIX machine for groups, but on macOS groups created with dscl are put into the Directory Service, which is not always reflected in the /etc/group. 

In fact, the /etc/groups reads: `Note that this file is consulted directly only when the system is running in single-user mode. At other times this information is provided by Open Directory. See the opendirectoryd(8) man page for additional information about OpenDirectory.`

Tell us which version of InSpec (`inspec version` or SHA of a commit) and Operating System distribution you are using

### Replication Case

Chef cookbook:
```
group 'SomeGroup`
```
InSpec:
```
describe group('SomeGroup') do
  it { should exist }
end
```

### Possible Solutions

On macOS the group resource could look at `dscl . -ls /Groups` to get a more complete list of groups.

jerryaldrichiii:
This allows users to verify groups added by Chef on OS X.

The current method that `UnixGroup` uses is to check the contents of `/etc/group`, but OS X adds groups to Directory Service and not `/etc/group`. This modifies the `group` resource on Darwin to use `dscacheutil` to get group info.

This fixes #2330

### DIFF
--- a/lib/resources/groups.rb
+++ b/lib/resources/groups.rb
@@ -14,7 +14,11 @@ module Inspec::Resources
     # returns nil, if no group manager was found for the operating system
     def select_group_manager(os)
       if os.unix?
-        @group_provider = UnixGroup.new(inspec)
+        @group_provider = if os.darwin?
+                            DarwinGroup.new(inspec)
+                          else
+                            UnixGroup.new(inspec)
+                          end
       elsif os.windows?
         @group_provider = WindowsGroup.new(inspec)
       end
@@ -151,6 +155,35 @@ module Inspec::Resources
   class UnixGroup < GroupInfo
     def groups
       inspec.etc_group.entries
+    end
+  end
+
+  # OSX uses opendirectory for groups, so `/etc/group` may not be accurate
+  class DarwinGroup < GroupInfo
+    def groups
+      raw_output = inspec.command('dscacheutil -q group').stdout
+      raw_group_info = raw_output.split("\n\n")
+
+      group_info = []
+      regex = /^([^:]*?)\s*:\s(.*?)\s*$/
+      raw_group_info.each do |data|
+        group_info << inspec.parse_config(data, assignment_regex: regex).params
+      end
+
+      match_etc_group_output(group_info)
+    end
+
+    # Converts the `dscacheutil` groups to match `inspec.etc_group.entries`
+    def match_etc_group_output(dscacheutil_groups)
+      dscacheutil_groups.each { |g| g['gid'] = g['gid'].to_i }
+
+      # Convert `users` to `members` and ` ` to `,`
+      dscacheutil_groups.each do |g|
+        next if g['users'].nil?
+        g['members'] = g['users']
+        dscacheutil_groups.delete(:users)
+        g['members'].tr!(' ', ',')
+      end
     end
   end
 
--- a/lib/resources/groups.rb
+++ b/lib/resources/groups.rb
@@ -13,15 +13,13 @@ module Inspec::Resources
     # select group provider based on the operating system
     # returns nil, if no group manager was found for the operating system
     def select_group_manager(os)
-      if os.unix?
-        @group_provider = if os.darwin?
-                            DarwinGroup.new(inspec)
-                          else
-                            UnixGroup.new(inspec)
-                          end
-      elsif os.windows?
-        @group_provider = WindowsGroup.new(inspec)
-      end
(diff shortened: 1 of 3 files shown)


lib/resources/groups.rb
```diff
@@ -14,7 +14,11 @@ module GroupManagementSelector
     # returns nil, if no group manager was found for the operating system
     def select_group_manager(os)
       if os.unix?
-        @group_provider = UnixGroup.new(inspec)
+        @group_provider = if os.darwin?
```
adamleff: Is there a reason to nest this under a `os.unix?` check? I think it might be cleaner to do this:

```ruby
if os.darwin?
  @group_provider = DarwinGroup.new(inspec)
elsif os.unix?
  @group_provider = UnixGroup.new(inspec)
elsif os.windows?
  @group_provider = WindowsGroup.new(inspec)
end
```

... or, if you want to use the conditional assignment stuff:

```ruby
@group_provider = if os.darwin?
                    DarwinGroup.new(inspec)
                  elsif os.unix?
                    UnixGroup.new(inspec)
                  elsif os.windows?
                    WindowsGroup.new(inspec)
                  end
```

lib/resources/groups.rb
```diff
@@ -154,6 +158,35 @@ def groups
...
+      regex = /^([^:]*?)\s*:\s(.*?)\s*$/
+      raw_group_info.each do |data|
+        group_info << inspec.parse_config(data, assignment_regex: regex).params
+      end
+
+      match_etc_group_output(group_info)
+    end
+
+    # Converts the `dscacheutil` groups to match `inspec.etc_group.entries`
+    def match_etc_group_output(dscacheutil_groups)
```
adamleff: What's the rationale for breaking this out into a separate method?

lib/resources/groups.rb
```diff
@@ -154,6 +158,35 @@ def groups
...
+      dscacheutil_groups.each { |g| g['gid'] = g['gid'].to_i }
+
+      # Convert `users` to `members` and ` ` to `,`
+      dscacheutil_groups.each do |g|
+        next if g['users'].nil?
+        g['members'] = g['users']
+        dscacheutil_groups.delete(:users)
```
adamleff: This seems curious to me... In the line above, you use the string version of the key, but then you delete the symbol?

lib/resources/groups.rb
```diff
@@ -154,6 +158,35 @@ def groups
...
+    end
+
+    # Converts the `dscacheutil` groups to match `inspec.etc_group.entries`
+    def match_etc_group_output(dscacheutil_groups)
+      dscacheutil_groups.each { |g| g['gid'] = g['gid'].to_i }
+
+      # Convert `users` to `members` and ` ` to `,`
+      dscacheutil_groups.each do |g|
+        next if g['users'].nil?
+        g['members'] = g['users']
```
adamleff: The `#delete` method returns the content of the key that was deleted. You should be able to consolidate this line and the line after to:

```ruby
g['members'] = g.delete('users')
```

adamleff (changes requested):
Great first go at this one, @jerryaldrichiii.

lib/resources/groups.rb
jerryaldrichiii: I can see that, I originally did it that way because `os.unix?` was at the top and triggering first. I suppose we could be `os.darwin?` first and achieve the same outcome.

I think the latter is cleaner so I'll do that.

lib/resources/groups.rb
jerryaldrichiii: I wanted to isolate the logic to keep the `groups` method clean. In hindsight, I should have made it a private method. I'll move the logic back into `groups` though since it's only a few lines an a comment will suffice.

lib/resources/groups.rb
jerryaldrichiii: Oh neat! Great tip. That's much cleaner.

lib/resources/groups.rb
jerryaldrichiii: Shoot, missed that. Great catch. This coupled with the tip you gave me about `#delete` cleans this up quite nicely.

adamleff (approved):

chris-rock (approved):
Thank you @jerryaldrichiii

PR #2408, opened 2017-12-21
grub_conf resource: fix menuentry detection

### LINKED ISSUE #2333: grub_conf resource parsing the wrong kernel
### Description

The grub_conf resource is parsing the wrong menuentry on RedHat and Centos 7 OS. All the time i get the menuentry from the last menuentry of the grub.cfg, but this is not the default menuentry.

inspec version: 1.45.9

OS: RHEL 7.4 and Centos 7.1

You can use this box `https://github.com/holms/vagrant-centos7-box/releases/download/7.1.1503.001/CentOS-7.1.1503-x86_64-netboot.box` for your testing, if you do not have a rhel 7 box available.

### Replication Case

For this case i use the mentioned Centos Box. It is the same behavior on the RHEL 7.4 box.

open a inspec shell and execute `inspec> grub_conf('/boot/grub2/grub.cfg').kernel
=> ["linux16", "/vmlinuz-0-rescue-3dde825218a845849d79c10075e675fb", "root=/dev/mapper/centos-root", "ro", "rd.lvm.lv=centos/root", "rd.lvm.lv=centos/swap", "crashkernel=auto", "biosdevname=0", "net.ifnames=0"]`

The Grub config looks like:
```
#
# DO NOT EDIT THIS FILE
#
# It is automatically generated by grub2-mkconfig using templates
# from /etc/grub.d and settings from /etc/default/grub
#

### BEGIN /etc/grub.d/00_header ###
set pager=1

if [ -s $prefix/grubenv ]; then
  load_env
fi
if [ "${next_entry}" ] ; then
   set default="${next_entry}"
   set next_entry=
   save_env next_entry
   set boot_once=true
else
   set default="${saved_entry}"
fi

if [ x"${feature_menuentry_id}" = xy ]; then
  menuentry_id_option="--id"
else
  menuentry_id_option=""
fi

export menuentry_id_option

if [ "${prev_saved_entry}" ]; then
  set saved_entry="${prev_saved_entry}"
  save_env saved_entry
  set prev_saved_entry=
  save_env prev_saved_entry
  set boot_once=true
fi

function savedefault {
  if [ -z "${boot_once}" ]; then
    saved_entry="${chosen}"
    save_env saved_entry
  fi
}

function load_video {
  if [ x$feature_all_video_module = xy ]; then
    insmod all_video
  else
    insmod efi_gop
    insmod efi_uga
    insmod ieee1275_fb
    insmod vbe
    insmod vga
    insmod video_bochs
    insmod video_cirrus
  fi
}

terminal_output console
if [ x$feature_timeout_style = xy ] ; then
  set timeout_style=menu
  set timeout=5
# Fallback normal timeout code in case the timeout_style feature is
# unavailable.
else
  set timeout=5
fi
### END /etc/grub.d/00_header ###

### BEGIN /etc/grub.d/00_tuned ###
set tuned_params=""
set tuned_initrd=""
### END /etc/grub.d/00_tuned ###

### BEGIN /etc/grub.d/01_users ###
if [ -f ${prefix}/user.cfg ]; then
  source ${prefix}/user.cfg
  if [ -n "${GRUB2_PASSWORD}" ]; then
    set superusers="root"
    export superusers
    password_pbkdf2 root ${GRUB2_PASSWORD}
  fi
fi
### END /etc/grub.d/01_users ###

### BEGIN /etc/grub.d/10_linux ###
menuentry 'Red Hat Enterprise Linux Server (3.10.0-693.5.2.el7.x86_64) 7.4 (Maipo)' --class red --class gnu-linux --class gnu --class os --unrestricted $menuentry_id_option 'gnulinux-3.10.0-693.5.2.el7.x86_64-advanced-3e592f82-80c7-4583-abcc-66a9fc689109' {
        load_video
        set gfxpayload=keep
        insmod gzio
        insmod part_msdos
        insmod xfs
        set root='hd0,msdos1'
        if [ x$feature_platform_search_hint = xy ]; then
          search --no-floppy --fs-uuid --set=root --hint-bios=hd0,msdos1 --hint-efi=hd0,msdos1 --hint-baremetal=ahci0,msdos1 --hint='hd0,msdos1'  67022177-300d-4be3-bd6a-b9f4a515c81d
        else
          search --no-floppy --fs-uuid --set=root 67022177-300d-4be3-bd6a-b9f4a515c81d
        fi
        linux16 /vmlinuz-3.10.0-693.5.2.el7.x86_64 root=/dev/mapper/rhel-root ro crashkernel=auto rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap biosdevname=0 net.ifnames=0 rhgb quiet net.ifnames=0 
        initrd16 /initramfs-3.10.0-693.5.2.el7.x86_64.img
}
menuentry 'Red Hat Enterprise Linux Server (3.10.0-693.el7.x86_64) 7.4 (Maipo)' --class red --class gnu-linux --class gnu --class os --unrestricted $menuentry_id_option 'gnulinux-3.10.0-693.el7.x86_64-advanced-3e592f82-80c7-4583-abcc-66a9fc689109' {
        load_video
        set gfxpayload=keep
        insmod gzio
        insmod part_msdos
        insmod xfs
        set root='hd0,msdos1'
        if [ x$feature_platform_search_hint = xy ]; then
          search --no-floppy --fs-uuid --set=root --hint-bios=hd0,msdos1 --hint-efi=hd0,msdos1 --hint-baremetal=ahci0,msdos1 --hint='hd0,msdos1'  67022177-300d-4be3-bd6a-b9f4a515c81d
        else
          search --no-floppy --fs-uuid --set=root 67022177-300d-4be3-bd6a-b9f4a515c81d
        fi
        linux16 /vmlinuz-3.10.0-693.el7.x86_64 root=/dev/mapper/rhel-root ro crashkernel=auto rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap biosdevname=0 net.ifnames=0 rhgb quiet net.ifnames=0 
        initrd16 /initramfs-3.10.0-693.el7.x86_64.img
}
menuentry 'Red Hat Enterprise Linux Server (0-rescue-90120aa86be64049928408085fcd9a4d) 7.4 (Maipo)' --class red --class gnu-linux --class gnu --class os --unrestricted $menuentry_id_option 'gnulinux-0-rescue-90120aa86be64049928408085fcd9a4d-advanced-3e592f82-80c7-4583-abcc-66a9fc689109' {
        load_video
        insmod gzio
        insmod part_msdos
        insmod xfs
        set root='hd0,msdos1'
        if [ x$feature_platform_search_hint = xy ]; then
          search --no-floppy --fs-uuid --set=root --hint-bios=hd0,msdos1 --hint-efi=hd0,msdos1 --hint-baremetal=ahci0,msdos1 --hint='hd0,msdos1'  67022177-300d-4be3-bd6a-b9f4a515c81d
        else
          search --no-floppy --fs-uuid --set=root 67022177-300d-4be3-bd6a-b9f4a515c81d
        fi
        linux16 /vmlinuz-0-rescue-90120aa86be64049928408085fcd9a4d root=/dev/mapper/rhel-root ro crashkernel=auto rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap biosdevname=0 net.ifnames=0 rhgb quiet net.ifnames=0 
        initrd16 /initramfs-0-rescue-90120aa86be64049928408085fcd9a4d.img
}
if [ "x$default" = 'Red Hat Enterprise Linux Server (3.10.0-693.5.2.el7.x86_64) 7.4 (Maipo)' ]; then default='Advanced options for Red Hat Enterprise Linux Server>Red Hat Enterprise Linux Server (3.10.0-693.5.2.el7.x86_64) 7.4 (Maipo)'; fi;
### END /etc/grub.d/10_linux ###

### BEGIN /etc/grub.d/20_linux_xen ###
### END /etc/grub.d/20_linux_xen ###

### BEGIN /etc/grub.d/20_ppc_terminfo ###
### END /etc/grub.d/20_ppc_terminfo ###

### BEGIN /etc/grub.d/30_os-prober ###
### END /etc/grub.d/30_os-prober ###

### BEGIN /etc/grub.d/40_custom ###
# This file provides an easy way to add custom menu entries.  Simply type the
# menu entries you want to add after this comment.  Be careful not to change
# the 'exec tail' line above.
### END /etc/grub.d/40_custom ###

### BEGIN /etc/grub.d/41_custom ###
if [ -f  ${config_directory}/custom.cfg ]; then
  source ${config_directory}/custom.cfg
elif [ -z "${config_directory}" -a -f  $prefix/custom.cfg ]; then
  source $prefix/custom.cfg;
fi
### END /etc/grub.d/41_custom ###
```

### Possible Solutions

For rhel 7 based os the resource must read at first the `/etc/default/grub`, because for rhel 7 based systems there is a string defined `GRUB_DEFAULT=saved` and for debian based os is there a integer like `GRUB_DEFAULT=0` defined to identify the default menuentry, which kernel should be loaded as default. Then the resource must read the `/boot/grub2/grubenv`, there is the string defined `saved_entry=Red Hat Enterprise Linux Server (3.10.0-693.5.2.el7.x86_64) 7.4 (Maipo)` to identify the menuentry on rhel 7 based os to load the default kernel. 

That means the line 83 (https://github.com/chef/inspec/blob/master/lib/resources/grub_conf.rb#L83) in the grub_conf resource must be adjusted.

An other idea is:

Just parse the complete grub.cfg and create a hash with all menuentries and parameters. Then the use can decide which kernel he want to inspect. Also add a filter to the resource to identify the default kernel, like `grub_config.default`.

jerryaldrichiii:
This does the following:
  - Corrects Grub2 bug where last entry was always selected
  - Adds support for specifying a Grub2 menu entry by name
  - Adds support for using `GRUB_DEFAULT=saved` with Grub2
  - Adds more Unit tests

This fixes #2333


lib/resources/grub_conf.rb
```diff
@@ -71,35 +73,62 @@ def to_s
...
-          menu_entry += 1
-          break
+    lines.each_with_index do |line, index|
+      next unless line =~ /^menuentry\s+.*/
+      entry = {}
+      entry['insmod'] = []
+
+      # Extract name from menuentry line
+      capture_data = line.match(/(?:^|\s+).*menuentry\s*['|"](.*)['|"]\s*--/)
+      entry['name'] = capture_data.captures[0]
```
adamleff: What if `.match` doesn't match the line, and then `capture_data` is `nil`?

lib/resources/grub_conf.rb
```diff
@@ -71,35 +73,62 @@ def to_s
...
+
+      menu_entries << entry
+    end
+
+    menu_entries
+  end
+
+  def default_menu_entry(menu_entries, default)
+    if default == 'saved'
+      grubenv_contents = read_file(@grubenv_path)
```
adamleff: Careful with this new helper method... this will raise SkipResource exceptions if the file isn't there or is empty or isn't readable. Are we guaranteed it will be there and with content?

adamleff (commented):
This looks like a good first step. Another potential option would be to use the `parselet` stuff that's used in the `nginx_conf` resource, but might be overkill here. I'd encourage you to at least take a look at it and understand how it works and decide which one is cleaner. What you have here is probably fine with a bit more of a defensive stance.

adamleff:
@jerryaldrichiii can you please rebase on master so we can try to get travis fixed for this PR?

jerryaldrichiii:
@adamleff I have rebased and pushed. :+1:

lib/resources/grub_conf.rb
jerryaldrichiii: We should be fine since we don't make it to that bit unless `line` contains `menuentry`. That being said, I'm all for adding an exception to ensure the user gets a better error than a `nil` error.

lib/resources/grub_conf.rb
jerryaldrichiii: Great catch! That is not a guarantee according to https://www.gnu.org/software/grub/manual/grub/html_node/Environment-block.html

I will fall back to the 0th entry the same way as if the `saved_entry` value is not up to date.

lib/resources/grub_conf.rb
adamleff: It's possible that a line can contain `menuentry` but still not pass your regex... unless you're willing to tell me you'd bet 3 months' salary that EVERY line in that file that has the word "menuentry" in it will ALWAYS match that regex. :)

jerryaldrichiii: Hah! Solid point. It would have to contain menuentry at the start of the line followed by at least 1 whitespace. I am not willing to take that bet though!

Let me know if failing the resource is appropriate here.

lib/resources/grub_conf.rb
```diff
@@ -71,35 +73,74 @@ def to_s
...
-          break
+    lines.each_with_index do |line, index|
+      next unless line =~ /^menuentry\s+.*/
+      entry = {}
+      entry['insmod'] = []
+
+      # Extract name from menuentry line
+      capture_data = line.match(/(?:^|\s+).*menuentry\s*['|"](.*)['|"]\s*--/)
+
+      if capture_data.captures[0]
```
adamleff: I still think it's possible for `capture_data` to POSSIBLY return nil, and then this line will throw a "undefined method on NilClass" error.

If you really want to fail the whole resource if we can't parse the line, this should probably look like this:

```ruby
if capture_data.nil? || capture_data.captures[0].nil?
  raise Inspec::Exceptions::ResourceFailed "Cannot extract menuentry name from line: #{line}"
end

entry['name'] = capture_data.captures[0]
```

You could also use the safe navigation operator to your advantage here as well:

```ruby
entry['name'] = capture_data&.captures[0]
if entry['name'].nil?
  raise Inspec::Exceptions::ResourceFailed "Cannot extract menuentry name from line: #{line}"
end
```

adamleff (changes requested):
This is looking pretty good, @jerryaldrichiii. I still think we need a bit of defensive coding around the `menuentry` line and the regex parsing of it. I've left you a couple suggestions.

lib/resources/grub_conf.rb
jerryaldrichiii: I agree, better to raise anything but "undefined method for NilClass".

I went with your former suggestion. The safe navigation operator is neat, but it threw me for a loop the first time I saw it and I wouldn't want it to trip anyone else up later.

lib/resources/grub_conf.rb
```diff
@@ -71,35 +73,73 @@ def to_s
...
 
   def grub2_parse_kernel_lines(content, conf)
-    # Find all "menuentry" lines and then parse them into arrays
-    menu_entry = 0
+    menu_entries = extract_menu_entries(content)
+
+    if @kernel == 'default'
+      default_menu_entry(menu_entries, conf['GRUB_DEFAULT'])
+    else
+      menu_entries.select { |entry| entry['name'] == @kernel }[0]
```
adamleff: If we're always grabbing the first, we could change the `select` to a `find` which always returns the first element where the block evals `true`

lib/resources/grub_conf.rb
```diff
@@ -71,35 +73,73 @@ def to_s
...
       end
+
+      menu_entries << entry
+    end
+
+    menu_entries
+  end
+
+  def default_menu_entry(menu_entries, default)
+    if default == 'saved'
```
adamleff: This is one of those places where I'd probably do this:

```ruby
return menu_entries[default.to_i] unless default == 'saved'

grubenv_contents = ....
```

Anytime I see a single-line else or a really small else block, I'd try and short-circuit the method early and eliminate one level of indentation.

Not a blocker for this, just something to look out for in the future.

adamleff (approved):

lib/resources/grub_conf.rb
jerryaldrichiii: Neat! I didn't know that, will update.

lib/resources/grub_conf.rb
jerryaldrichiii: Great catch! I changed it because I believe that way to be much cleaner.

arlimus (approved):
I love the improvements in the detection @jerryaldrichiii thank you for adding this!!

@adamleff kudos for your great review as well

PR #2410, opened 2017-12-21
bugfix: default attributes for nil and false

arlimus:
Traditionally those would translated DEFAULT_ATTRIBUTE. but that was wrong, it should have been nil or false or whatever the user supplied.


chris-rock (approved):
Thank you @arlimus

PR #2423, opened 2017-12-27
xml resource: support fetching attributes

### LINKED ISSUE #2409: Cannot use XML resource with XPath to test attribute value
### Description

We're trying to check a connection string that is stored in an xml file as a node's attribute value. The xml resource doesn't seem to yield correct results.

```
Official Chef Inspec Docker Container (latest)
    Inspec Version: 1.48.0

    OS platform: alpine
    OS family: linux
    OS release: 3.4.6
```
The replication case below was tested against SLES 11 SP4 and Ubuntu 16.04.3 LTS.

### Replication Case
**database.xml** (Store as `/tmp/database.xml` for test purposes)

```xml
<beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util" xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd" default-lazy-init="true">
<bean id="dataSource" class="com.myapplication.base.persistency.support.DecoratedDataSource" parent="oracleBaseDataSource" scope="singleton" lazy-init="true">
	<property name="type" ref="ORACLE_11" />
	<property name="name" value="MYAPPLICATION" />
	<property name="url" value="jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD" />
	<property name="username" value="MYUSERNAME" />
	<property name="password">
	<bean class="com.mycorp.core.util.crypt.PasswordHelper" factory-method="decrypt">
	<constructor-arg value="" />
</bean>
</property>
</bean>
</beans>
```

**test.rb**

```ruby
testfile = '/tmp/database.xml'
datasource = 'jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD'

describe file(testfile) do
    it { should exist }
end
if file(testfile).exist?
    describe xml(testfile) do
        its(['//property[@name="url"]/@value']) { should eq datasource }
    end
end
```

### Possible Solutions

We've tried multiple XPath combinations in the xml describe block:

```
inspec> if file(testfile).exist?
inspec>   describe xml(testfile) do
inspec>     its(['//property[@name="url"]/@value']) { should eq datasource }
inspec>   end
inspec> end

Profile: inspec-shell
Version: (not specified)

  XML /tmp/database.xml
     ∅  undefined method `text' for #<REXML::Attribute:0x0000556eab506fa8>

```

```
inspec> if file(testfile).exist?
inspec>   describe xml(testfile) do
inspec>     its(['//property[@name="url"]/@value/text()']) { should eq datasource }
inspec>   end
inspec> end

Profile: inspec-shell
Version: (not specified)

  XML /tmp/database.xml
     ∅  ["//property[@name=\"url\"]/@value/text()"] should eq "jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD"

     expected: "jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD"
          got: []

     (compared using ==)


```

tarcinil:
fixes #2409


tarcinil:
Travis has failed due to changes in the `x509_certificate /tmp/mycert.pem`, that I believe came from updates to their build/openssl library.

Awaiting direction from @adamleff  as it is outside the scope of my changes.

```
     ∅  subject_dn should eq "/C=US/O=Foo Bar/OU=Lab/CN=www.f00bar.com"
     
     expected: "/C=US/O=Foo Bar/OU=Lab/CN=www.f00bar.com"
          got: "/C=US/ST= /L= /O=Foo Bar/OU=Lab/CN=www.f00bar.com"
     
     (compared using ==)
     
     ✔  subject.C should eq "US"
     ∅  issuer_dn should eq "/C=US/O=Foo Bar/OU=Lab/CN=www.f00bar.com"
     
     expected: "/C=US/O=Foo Bar/OU=Lab/CN=www.f00bar.com"
          got: "/C=US/ST= /L= /O=Foo Bar/OU=Lab/CN=www.f00bar.com"
     
     (compared using ==)
```

test/unit/resources/xml_test.rb
```diff
@@ -26,4 +26,22 @@
...
+    end
+
+    it 'retrieves empty array if xpath cannot be found' do
+      _(resource.send('missing')).must_equal []
+    end
+
+    it 'retrieves attribute value through xpath' do
+      _(resource.send('//property[@name="url"]/@value')).must_equal ['jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD']
+      _(resource.send('/beans/bean[@id="dataSource"]/property[@name="url"]/@value')).must_equal ['jdbc:oracle:thin:@databaseserver.domain.tld:1521/DBO.DOMAIN.TLD']
+
```
adamleff: Extra line here, but not a big deal!

lib/resources/xml.rb
```diff
@@ -20,7 +20,16 @@ def parse(content)
     end
 
     def value(key)
-      REXML::XPath.each(@params, key.first.to_s).map(&:text)
+      output = []
+      REXML::XPath.each(@params, key.first.to_s) do |element|
+        if element.is_a?(REXML::Attribute)
+          output.push(element.to_s)
+        elsif element.is_a?(REXML::Element)
```
adamleff, line 27: Should this be a default `else`? Is there any chance we'd stumble across something that's not a `REXML::Attribute` or `REXML::Element` that wouldn't respond to a `#text` method call?

adamleff (changes requested):
On a roll, @tarcinil :) Quick question below regarding your logic when looking at attributes vs. elements, but otherwise, this is a great enhancement.

lib/resources/xml.rb
tarcinil: @adamleff I cannot say with 100% confidence.  I believe the only remaining type would be context which contains data about parsing.

I think a default statement would be nice, but it should fall into `ResourceSkipped` or `ResourceFailed` so that it doesn't run into the same issue that I am fixing with this PR.  That would allow it to sanely fail rather than with a MethodError.

test/unit/resources/xml_test.rb
tarcinil: I will fix that.  I had noticed it and it was on my list of things to do tonight.

lib/resources/xml.rb
adamleff: I would be in support of that!

tarcinil:

lib/resources/xml.rb
```diff
@@ -20,7 +20,18 @@ def parse(content)
...
+          output.push(element.text)
+        else
+          raise Inspec::Exceptions::ResourceSkipped, 'We have encountered an unknown class from REXML'
```
adamleff: Let's fail this, and let's give a little more help to the user:

```ruby
raise Inspec::Exceptions::ResourceFailed, "Unknown XML object received (#{element.class}): #{element}"
```

That way the user can either try to fix on their own, or they'll be able to give us more helpful debugging information to fix a bug.

adamleff (changes requested):
Just one change to the exception message and I think this is good to go!

adamleff (approved):
Great work, @tarcinil 🙂

megamorf:
Hey @adamleff, do you know when this will be merged? Thanks :-)

megamorf:
Hey @adamleff, do you know when this will be merged? Thanks :-)

adamleff:
@megamorf it should be merged this week. Last week, all Chef employees were meeting for our annual all-hands event and this needs one more maintainer approval before we can merge it in. It should be included in this week's upcoming release.

chris-rock (approved):
Cool stuff @tarcinil

PR #2429, opened 2017-12-30
mssql_session resource: add port parameter

### LINKED ISSUE #2148: mssql_session resouce missing "port" attribute
My MSSQL server is running on custom port, normally I connect to it using sqlcmd with a custom option:
sslcmd -S "localhost,1533" ...

Looking at [mssql_session resource](https://github.com/chef/inspec/blob/master/lib/resources/mssql_session.rb) there is no such functionality.

Could you please add the "port" attribute to this resource?
Thanks.

tarcinil:
fixes #2148 

also adds unit testing around hostname/port/instance and mocks them to the same mssql-getdata in helper.rb

adds integration test for mssql_session based on appveyor services (https://www.appveyor.com/docs/services-databases/#sql-server-2012), however, no integration testing is done on appveyor so I removed the appveyor.yml services section in c28deda


lib/resources/mssql_session.rb
```diff
@@ -25,26 +25,27 @@ class MssqlSession < Inspec.resource(1)
...
       if opts[:pass]
         warn '[DEPRECATED] use `password` option to supply password instead of `pass`'
       end
       @host = opts[:host] || 'localhost'
+      @port = opts[:port] || '1433'
       @instance = opts[:instance]
 
       # check if sqlcmd is available
-      return skip_resource('sqlcmd is missing') if !inspec.command('sqlcmd').exist?
+      raise Inspec::Exceptions::ResourceFailed, 'sqlcmd is missing' unless inspec.command('sqlcmd').exist?
```
adamleff: This is technically a breaking change since we mark the resource as "skipped" and not failed. In principle, I think I agree with you that we should fail, but we have to do that during a major version bump.

For now, we should change this to `Inspec::Exceptions::ResourceSkipped`

lib/resources/mssql_session.rb
```diff
@@ -25,26 +25,27 @@ class MssqlSession < Inspec.resource(1)
...
       # check that database is reachable
-      return skip_resource("Can't connect to the MS SQL Server.") if !test_connection
+      raise Inspec::Exceptions::ResourceFailed, "Can't connect to the MS SQL Server." unless test_connection
```
adamleff: Same change as above - we need to skip the resource rather than fail it for now.

adamleff (changes requested):
@tarcinil fantastic PR - thank you! Just two small changes we need to make to ensure backward compatibility for now (even though I think your change is an accurate one). Comments below.

tarcinil:

adamleff (approved):
Looks great, @tarcinil!

chris-rock (approved):
Awesome work 👍  @tarcinil

PR #2430, opened 2018-01-01
firewalld resource: prepend rule string only when necessary

### LINKED ISSUE #2383: Firewalld test fails
### Description

Try to use firewalld test. Rule is there but inspec says the rule doesnt exist

Inspec: 1.48.0
OS: Ubuntu 16.04
RVM: Ruby 2.4.1p111

Target OS:
CentOS 7.4.1708

### Replication Case

Create a script "firewall.rb"

```
control 'firewall' do
   title 'Check firewall'
   describe firewalld do
     it { should be_running }
     its('default_zone') { should eq 'public' }
     it { should have_rule_enabled('family=ipv4 service name=http accept', 'public')}
   end
end
```

Create a firewall rule on OS in firewalld

`firewall-cmd --add-rich-rule 'rule family=ipv4 service name=http accept'`

Test inspec
`inspec exec firewall.rb -t ssh://user@host --sudo`

Output is similar to:

```
~/.rvm/gems/ruby-2.4.1/gems/net-ssh-3.2.0/lib/net/ssh/transport/cipher_factory.rb:97: warning: constant OpenSSL::Cipher::Cipher is deprecated
~/.rvm/gems/ruby-2.4.1/gems/net-ssh-3.2.0/lib/net/ssh/transport/cipher_factory.rb:72: warning: constant OpenSSL::Cipher::Cipher is deprecated

Profile: tests from firewall.rb (tests from firewall.rb)
Version: (not specified)
Target:  ssh://user@host:22

  ∅  firewall: Check firewall (1 failed)
     ✔  firewalld should be running
     ×  firewalld should have rule enabled "family=ipv4 service name=http accept", "public"
     expected #has_rule_enabled?("family=ipv4 service name=http accept", "public") to return true, got false
     ✔  firewalld default_zone should eq "public"

Test Summary: 2 successful, 1 failure, 0 skipped
```

With following code in regard
[https://github.com/chef/inspec/blob/7fc7942ab1d074820cdd8a7259a78e64ef2627f5/lib/resources/firewalld.rb#L87-L90](https://github.com/chef/inspec/blob/7fc7942ab1d074820cdd8a7259a78e64ef2627f5/lib/resources/firewalld.rb#L87-L90)

I executed `firewall-cmd --zone=public --query-rich-rule='rule family="ipv4" service name="http" accept'`
which results in "yes"

### Possible Solutions

Only conclusion. Target system is ok. Rule is in place, and executing the test command over there results in a positive lookup. So somewhere between te lookup and then end result its "yes" gets converted to "false"

### Stacktrace

Not applicable

tarcinil:
fixes #2383

adding rule string if it doesn't already contain rule string.

Correcting resource name in firewalld from etc_hosts_deny

adding tests for both branches of the statement created in firewalld

Signed-off-by: Vern Burton <me@vernburton.com>


lib/resources/firewalld.rb
```diff
@@ -85,8 +85,8 @@ def has_port_enabled_in_zone?(query_port, query_zone = default_zone)
     end
 
     def has_rule_enabled?(rule, query_zone = default_zone)
-      rule = 'rule ' + rule
-      firewalld_command("--zone=#{query_zone} --query-rich-rule=#{rule}") == 'yes'
+      rule = "rule #{rule}" until rule.include?('rule')
```
adamleff: Should this be `unless` instead of `until`? `until` is usually reserved for loop control.

Also, in case the word "rule" could appear elsewhere in the string, we might want to be defensive about this and make this: `unless rule.start_with?('rule')`

adamleff (changes requested):
Another great contribution, @tarcinil - thank you! A comment left below.

lib/resources/firewalld.rb
tarcinil: @adamleff `until` is also favored for negative conditions.

https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#until-for-negatives

I agree with the comment about being defensive about the location of "rule" and I will plan on changing the branch when I get home from work.

adamleff: True, but again, it's in the context of loop control (`while` and `until`). `unless` is the preferred statement in this case, and would be consistent with the rest of the codebase.

Think of it this way: `until` is the opposite of `while`, like `unless` is the opposite of `if`.

tarcinil:

adamleff (approved):
Fantastic, @tarcinil - thanks!

chris-rock (approved):
Thank you @tarcinil

PR #2431, opened 2018-01-02
Fix `x509_certificate` integration tests

jerryaldrichiii:
An update to the openssl cookbook modified the defaults for `state` and
`city` in the `openssl_x509` resource. That change modified the output
of `issuer_dn` and `subject_dn` in InSpec's `x509_certificate` resource.

This modifies the expected output of the integration tests to match
these new defaults.

See: https://github.com/chef-cookbooks/openssl/commit/06f790bb34453ccad38ed2bf6c233cd9f2775861


adamleff (approved):
Thanks @jerryaldrichiii!

PR #2432, opened 2018-01-02
Fix Docker build in the Expeditor pipeline

adamleff:
The new ruby:alpine image already has a /share directory, so our attempt to create one is breaking the Docker builds. This change updates the Dockerfile to try to make the /share directory but silently fail if it's already there.


chris-rock (approved):
Thank you @adamleff

PR #2435, opened 2018-01-02
Modify `inspec json` to use `check_mode`

### LINKED ISSUE #2434: Enable "check_mode" during an "inspec json" operation
When running `inspec check`, we enable "check_mode" on the `Profile` object in order to ensure certain operations that should only happen during an `exec` don't take place, such the evaluation of `only_if` block contents (which is pointless to do during an `inspec check`).

We need to enable the same for `inspec json` as that's another example of an operation that just checks the validity of a profile and outputs the contents, so it should mimic the settings of `inspec check` as well.

## Example Control

```ruby
control 'RHEL-06-000324' 
  # standard control stuff
  only_if { package('GConf2').installed? }
end
```

## Expected Outcome

InSpec output's the JSON representation of the profile.

## Current Outcome

```
NoMethodError: undefined method `info' for nil:NilClass
  /home/travis/.chefdk/gem/ruby/2.4.0/bundler/gems/inspec-c4d678a1df06/lib/resources/package.rb:67:in `info'
```

jerryaldrichiii:
This modifies `inspec json` to make it not evaluate code inside of `only_if` blocks.

This closes #2434

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -36,6 +36,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     o = opts.dup
     o[:ignore_supports] = true
     o[:backend] = Inspec::Backend.create(target: 'mock://')
+    o[:check_mode] = true
 
     profile = Inspec::Profile.for_target(target, o)
     dst = o[:output].to_s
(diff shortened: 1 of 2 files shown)


adamleff (commented):
Thanks for the quick fix, @jerryaldrichiii!

adamleff (approved):

arlimus (approved):
Sweet fix thank you @jerryaldrichiii

PR #2437, opened 2018-01-03
package resource: fix NilClass errors on arch linux

jerryaldrichiii:
This modifies `.info` to return `{}` in cases where the package manager cannot be determined. This matches the behavior of `@pkgman.info`.

### DIFF
--- a/test/unit/resources/package_test.rb
+++ b/test/unit/resources/package_test.rb
@@ -139,6 +139,7 @@ describe 'Inspec::Resources::Package' do
   # undefined
   it 'verify package handling on unsupported os' do
     resource = MockLoader.new(:undefined).load_resource('package', 'curl')
+    _(resource.info).must_be_empty
     _(resource.resource_skipped?).must_equal true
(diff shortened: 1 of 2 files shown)


test/unit/resources/package_test.rb
```diff
@@ -139,6 +139,7 @@
   # undefined
   it 'verify package handling on unsupported os' do
     resource = MockLoader.new(:undefined).load_resource('package', 'curl')
+    _(resource.info).must_be_empty
```
adamleff: The only downside with using `must_be_empty` here is that if `resource.info` ever had a regression introduced that returned an empty string, this would still pass.

It might be safer to say `.must_equal({})`

adamleff (changes requested):

test/unit/resources/package_test.rb
jerryaldrichiii: Great catch! I have updated it to match what you suggested.

adamleff (approved):

chris-rock (approved):
Thank you @jerryaldrichiii

PR #2438, opened 2018-01-03
load local dependencies in inspec shell

### LINKED ISSUE #1486: InSpec Shell should support custom resources
### Description

InSpec shell doesn't load my custom resources.

all

### Replication Case

Just run inspec shell, there is no option to tell it which dependency profile to pull or include nor resource to load locally.

### Possible Solutions

Add an option to load resource packs.

arlimus:
for loading dependencies from local folders. mainly used for development.

depends

Fixes #1486

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -194,6 +194,8 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: 'A single command string to run instead of launching the shell'
   option :format, type: :string, default: nil, hide: true,
     desc: 'Which formatter to use: cli, documentation, html, json, json-min, junit, progress'
+  option :depends, type: :array, default: [],
+    desc: 'A list of local folders that are loaded as dependencies (e.g. for resources)'
   def shell_func
     diagnose
     o = opts.dup
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -195,7 +195,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   option :format, type: :string, default: nil, hide: true,
     desc: 'Which formatter to use: cli, documentation, html, json, json-min, junit, progress'
   option :depends, type: :array, default: [],
-    desc: 'A list of local folders that are loaded as dependencies (e.g. for resources)'
+    desc: 'A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell'
   def shell_func
     diagnose
     o = opts.dup
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -40,6 +40,28 @@ $ inspec shell -t winrm://UserName:Password@windowsmachine:1234  # Login to wind
 $ inspec shell -t docker://container_id # Login to a docker container.
 ```
 
+## Resource packs
+
+The InSpec shell may use additional keywords provided in resource packs.
+A resource pack is a profile that defines new language terms that can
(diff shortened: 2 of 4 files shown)


aaronlippold:
That's really cool. Now the follow-up to this would be to add train support for AWS.

lib/inspec/cli.rb
```diff
@@ -194,6 +194,8 @@ def detect
     desc: 'A single command string to run instead of launching the shell'
   option :format, type: :string, default: nil, hide: true,
     desc: 'Which formatter to use: cli, documentation, html, json, json-min, junit, progress'
+  option :depends, type: :array, default: [],
+    desc: 'A list of local folders that are loaded as dependencies (e.g. for resources)'
```
adamleff: Proposed alteration to make it clear what type of folder we're expecting:

`'A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell'`

lib/inspec/cli.rb
```diff
@@ -194,6 +194,8 @@ def detect
     desc: 'A single command string to run instead of launching the shell'
   option :format, type: :string, default: nil, hide: true,
     desc: 'Which formatter to use: cli, documentation, html, json, json-min, junit, progress'
+  option :depends, type: :array, default: [],
```
adamleff, line 197: What do you think about calling the option `--load-profiles` instead of `--depends`? I know `--depends` is technically correct re: implementation and consistent with the inspec.yml, but I wonder if `--load-profiles` is a clearer option as to what it feels like to the user. Don't feel obligated to take this suggestion, just wanted to toss it out there.

adamleff (approved):
Super great change, @arlimus, and closes a long-standing feature request too!

Two suggestions, neither of which are blocking.

adamleff:
Could we add a quick section to https://www.inspec.io/docs/reference/shell/ also?

aaronlippold:
@arlimus @adamleff do we need to update or add any options to `kitchen-inspec` as well to connect support for this? Something akin to when `controls:` and `attributes:` were added? Looking forward to testing this today or tomorrow 👍

adamleff:
The change is shell-specific. It has nothing to do with Test Kitchen or otherwise.

aaronlippold:
Ok. Just double checking. Thanks.

lib/inspec/cli.rb
arlimus: Much nicer words, thank you, i'll add it!

lib/inspec/cli.rb
arlimus: Load-profiles is interesting too yeah 🤔 

To add a small argument: "load profiles" would technically only load the libraries of the profile and not the controls, i.e. not the entire profile. I was pondering "load libraries" (but that may lead users to specify the libraries folder which we don't want either; traversing dependencies for all resources is beautiful right now)

There may be a better term...

adamleff: `--load-profile-libs`?

adamleff: I can't think of anything terribly better. If `--load-profile-libs` doesn't feel right, and you think `--depends` is clear enough, please merge when ready. :)  We can always add an alias if someone comes up with something better.

arlimus: Good idea, i'll make sure to add the docs you suggested below. Otherwise we can deprecate and find something else. Naming is hard...

PR #2441, opened 2018-01-05
filesystem resource: inspect linux filesystems

### LINKED ISSUE #1831: Missing "df" resource
Although there is a mount resource: 
https://www.inspec.io/docs/reference/resources/mount/
which is great, but if the filesystem size needs to be checked, we would require a "df" like solution, since doing it by bash isn't the best: 

  describe command('echo "scale=1;$(/usr/bin/df -m /boot --output=size | sed "/blocks/d; s/ *//g")/1024" | /usr/bin/bc | sed "s/^\./0\./g"') do
   its('stdout') { should match '0.2' }
  end

Please implement a "df" resource too. 

Many thanks.

tarcinil:
fixes #1831

adding df resource

```
describe df('/') do
  its('size') { should be >= 32000 }
end
```

I do convert the `cmd.stdout` to an integer.  I believe that as a full-blown resource, that pattern was better rather than relying on `cmp` to deal with it.


lib/resources/df.rb
```diff
@@ -0,0 +1,28 @@
...
+    example "
+      describe df('/') do
+        its('size') { should be >= 32000 }
+      end
+    "
+    attr_reader :partition
+
+    def initialize(partition)
+      @partition = partition
+      @cmd = inspec.command("df #{@partition} --output=size | sed \"/blocks/d; s/ *//g\"")
```
adamleff: Since you made `partition` an `attr_reader` (and it's available as `partition` as a method argument), you can drop the `@` here.

adamleff: I don't think there's any value in storing the command in an instance variable since it's only ever used in the `size` method... other than to perhaps prevent the command from being executed twice.

Instead, we can memoize the size value in the `size` method to avoid this. Example below.

adamleff: `--output=size` is not a globally available flag. For example, it doesn't work on macOS or FreeBSD, to my knowledge.

We need to use the new `supports` keyword here to limit this resource to only executing on the operation systems on which this is supported. I would recommend we start with the "linux" family to start. @jquick, could you chime in here and help @tarcinil with the proper `supports` call that will accomplish this?

lib/resources/df.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def size
+      @cmd.stdout.to_i
```
adamleff: So, rather than storing the `command` object in an instance var, I'd change this to something like this:

```ruby
@size ||= begin
            cmd = inspec.command("df #{@partition} --output=size | sed \"/blocks/d; s/ *//g\"").stdout
            raise Inspec::Exceptions::ResourceFailed, "Unable to get available space for partition #{partition}" if cmd.stdout.nil? || cmd.stdout.empty? || !cmd.exit_status.zero?
            cmd.stdout.to_i
          end
```

This will ensure the command is only run once (as we'll cache the size in an instance variable) and also does some additional error handling.

I'd also suggest just getting the raw output from the command and doing the string manipulation/parsing in Ruby instead of piping to `sed`.

lib/resources/df.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def to_s
+      "DiskSpace #{@partition}"
```
adamleff: Same here - can just change this to `#{partition}` thanks to it being made into an `attr_reader`

test/unit/resources/df_test.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+# author: Christoph Hartmann
```
adamleff: You can remove these two author lines.

adamleff (changes requested):
Great first pass, @tarcinil! Left you a bunch of comments, let me know if anything is unclear.

lib/resources/df.rb
tarcinil: I was unable to find an existing example of `supports` as it pertained to resources.

I used a pre-existing pattern of `inspec.os.linux?` which appears to depend on chef/train in the background.

tarcinil: I have found the `supports?`/`supported?` functions from `lib/inspec/plugins/resource.rb` and `lib/resources/platform.rb`, but I am unable to find documentation that would allow me to implement it.  I await implementation details from @jquick.

jquick: Hey @tarcinil. You are correct that the documentation is lacking for the new supports features for resources, this will be addressed. This was implemented via https://github.com/chef/inspec/issues/1661 (section 4). For this resource you will want to add something like:
`supports os_family: 'linux'`
After you set the 'name' on line 3 in your new resource. This will skip any tests using this resource unless the platform is a linux base.

tarcinil:
@adamleff Comments have been addressed, ready for another review.

lib/resources/df.rb
```diff
@@ -14,15 +12,26 @@ class DfResource < Inspec.resource(1)
...
       @partition = partition
-      @cmd = inspec.command("df #{@partition} --output=size | sed \"/blocks/d; s/ *//g\"")
     end
 
     def size
-      @cmd.stdout.to_i
+      if inspec.os.linux?
+        command = "df #{partition} --output=size"
+      else
+        raise Inspec::Exceptions::ResourceFailed, 'The `df` resource is not supported on your OS.'
```
adamleff: The `supports` stuff you added takes care of this for you so you can remove all of this logic and just declare your command. 🙂 

And you should move the command into the `@size` memoifier so it looks something like this and is the first line of the `size` method:

```ruby
@size ||= begin
  cmd = inspec.command("df #{partition} --output=size")
  raise ...
```

adamleff (changes requested):
This is looking really solid, @tarcinil! One last small fix and we'll be good to go, I think.

lib/resources/df.rb
tarcinil: I thought as much as well, but would it be nice to have for future support of other OS families?

adamleff: Sure, but if we're not going to account for that in this PR, we should leave it out.

If we add future family support, I'd recommend doing two things:

1. Flag the resource for the additional family support using the `supports` method
2. Change the line to:

```ruby
cmd = inspec.command(df_command_for_family)
```

... and supply a new `df_command_for_family` method that returns a command string appropriate for the node under test.

adamleff (approved):
Excellent work, @tarcinil!

adamleff:
Hey @tarcinil - one last thing, and I'm sorry I didn't think about this sooner. I think we should rename this resource to `filesystem`. If we were to add Windows support in the future, those users wouldn't really understand why the resource is called `df`.

Also, if we switch to use a different method of inspecting filesystem properties other than the `df` command, it would be confusing.

What do you think?

tarcinil:
@adamleff, I think that is a good argument.  I will look at making the changes on Tuesday most likely.

tarcinil:
I have time to fix this tonight.  It has been a busy week between getting engaged and snowmageddon 2018.

adamleff:
No problem, @tarcinil! Hope you're surviving the snow... no rush on this, and I thank you again for your contribution!

adamleff (approved):
Excellent work, @tarcinil! Thanks for introducing this new resource. I'm sure we'll be able to continue adding to it in the future.

chris-rock (approved):
Thank you @tarcinil

PR #2450, opened 2018-01-05
Adds glossary.md file

kagarmoe:
This is the glossary file that we've developed. We should decide about where it lives. Does it get its own page? Does it belong somewhere else?

Signed-off-by: kgarmoe <kgarmoe@chef.io>

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -23,7 +23,7 @@ end
 
 ### **its('blah') { should cmp '123' }** is an _individual test_, where
 
-  * `blah'` is a _property_
+  * `'blah'` is a _property_
   * { should cmp '123' }is a _condition statement_
   * `should`  is the _condition_
   * `cmp`  is the _matcher_
@@ -95,6 +95,6 @@ end
 ## Properties with advanced usage:
 
 ### Some properties may have advanced usage:
-#### **its `('jared.sort.first.monkey') { should be `loud` }**
+#### **its `('jared.sort.first.monkey') { should be `loud` }`**
 
   * `jared.sort.first.monkey` is an example of the `jared` property with an advanced usage
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -9,33 +9,43 @@ describe foo('/path/to/foo.txt') do
     it { should_not be_ridiculous }
 end
 ```
-#### Basic Elements:
-* describe **`foo`**
-    * **`foo`** is the _`resource`_
+## Basic Elements:
 
-* describe foo **`('/path/to/foo.txt')`**
-   *  **`('/path/to/foo.txt')`** is the _`resource parameter`_
+### describe **foo**, where
 
-#### Tests:
-* **`its('blah') { should cmp '123' }`**  is an _`individual test`_
-    * **`{ should cmp '123' }`** is a _`condition statement`_
-    * **`should`** is the `condition`
-    * **`cmp`** is the `matcher`
-    * **`123`** is the `expected` result
+        "foo" is the _resource_
 
-* **`{ should exist }`** is a _`condition statement`_
-    * `should` is the condition
-    * `exist` is the matcher
+### describe foo **('/path/to/foo.txt')**, where
 
-* **`{ should be_reasonable }`** is a _`condition statement`_
-    * **`should`** is the _`condition`_
-    *  **`be_reasonable`** is the _`matcher`_
+        "('/path/to/foo.txt')" is the resource parameter
 
-* **`{ should_not be_ridiculous }`** is a _`negative condition statement`_
-    * **`should_not`** is the _`negative condition`_
-    * **`be_ridiculous`** is the _`matcher`_
+## Tests:
+
+### **its('blah') { should cmp '123' }** is an _individual test_, where
+
+        "blah'" is a property
+        { should cmp '123' }is a condition statement
+        "should" is the condition
+        "cmp" is the matcher
+        "123" is the expected result
+
+### **{ should exist }** is a _condition statement_, where
+
+        "should" is the condition
+        "exist" is the matcher
+
+### **{ should be\_reasonable }** is a _condition statement_, where
+    
+        "should" is the condition
+        "be_reasonable" is the matcher
+
+### **{ should\_not be\_ridiculous }** is a _negative condition statement_, where
+
+        "should_not" is the negative condition
+        "be_ridiculous" is the matcher
 
 ## Advanced Syntax
+
 ```
 describe foos('/path/to/foo.txt', ssl_verify: true).where { names == 'blah' } do
     its('jared') { should cmp >= 123 }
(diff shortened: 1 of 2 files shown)


docs/glossary.md
```diff
@@ -0,0 +1,76 @@
...
+    * { should cmp >= 123 } is a conditional statement that uses a matcher with an operator and expected value.
+        * **`cmp`** is the _`matcher`_
+        * **`>=`** is the _`operator`_ (some matchers accept operators)
+        * **`123`** is the _`expected value`_
+
+#### Properties with advanced usage:
+* some properties may have advanced usage:
+    * its **`('jared.sort.first.monkey') { should be "loud" }`**
+       * **`jared.sort.first.monkey`** is an example of the **`jared`** property with some advanced usage
+    * **`its(['jared', 'monkey.with.dots') { should be "loud" }`**
```
adamleff: Missing a closed bracket after `'monkey.with.dots'`: 

`its(['jared', 'monkey.with.dots']) { should be "loud" }`

docs/glossary.md
```diff
@@ -0,0 +1,76 @@
...
+   *  **`where  { names == 'blah' }`** is an example of a *`filter`*. Some resources support one or more filters.
+   *  **`where  { names == 'blah' }`**  example of a filter clause [research SQL terms for this.]
+   *  **`{ names == 'blah' , has spots}`** are filter criteria
+        *  `names` compares output to 'blah'
+        *  `has spots` evaluates to 'true' or 'false'
+
+#### Properties:
+* **its('jared') { should cmp >= 123 }**
+    * **`jared`** is the _`property`_
+    * { should cmp >= 123 } is a conditional statement that uses a matcher with an operator and expected value.
```
adamleff: We should wrap the `{ should cmp >= 123 }` with backticks for consistent markdown code formatting.

adamleff (commented):
I really like the way this has turned out. I'd be in favor of this going in the left-hand nav in the docs section of the site. 

@arlimus and @chris-rock, what do you think of this page?

arlimus (commented):
I love the idea of the glossary and having everything at your fingertips in a simple view! I agree with Adam's review for a few more improvements before we can merge this. Also the formatting is a bit tricky in markdown 🤔 it's nevertheless an awesome start and something that will help a lot of people understand it better

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+  * `names` compares output to `blah`
+  * `has spots` evaluates to `true` or `false`
+
+## Properties:
+
+### **its('jared') { should cmp >= 123 }**
+
+  * `jared` is the _property_
+  *
```
adamleff: Extra bullet.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+```
+## Basic Elements:
+
+### describe **foo**, where
+
+  * `foo` is the _resource_
+
+### describe foo **('/path/to/foo.txt')**, where
+
+  * `('/path/to/foo.txt')` is the _resource parameter_
```
adamleff: Parameters go inside the parentheses, but are not actually part of the parameter. I think this should be:

```
  * `'/path/to/foo.txt'` is the _resource parameter_
```

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Tests:
+
+### **its('blah') { should cmp '123' }** is an _individual test_, where
+
+  * `'blah'` is a _property_
```
adamleff: Technically, the property is `blah` - the single quotes are not part of the property.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+  * { should cmp '123' }is a _condition statement_
```
adamleff: Missing a space between `}` and `is`, and `{ should cmp '123' }` should be wrapped in backticks.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+  * `should`  is the _condition_
+  * `cmp`  is the _matcher_
+  * `123`  is the _expected result_
```
adamleff: Technically, since the example has `123` wrapped in quotes, the expected result is actually `'123'` and not `123` - string vs. integer. This should probably be:

```
 * `'123'` is the _expected result_
```

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Advanced Elements:
+
+### describe **foos**, where
+
+  * `foos` is a _plural resource_
+
+### describe foos **('/path/to/foo.txt', ssl_verify: true)**, where
+
+  * `('/path/to/foo.txt', ssl_verify: true)` are the _resource parameters_. Resources take one or more parameters.
```
adamleff: As with above, the parentheses are not technically part of the parameters. I think this should be:

```
 * `'/path/to/foo.txt'` and `ssl_verify: true` are the _resource parameters_. Resources take one or more parameters.
```

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Filters:
+
+### describe **foos('/path/to/foo.txt', ssl_verify: true)** where { names == 'blah' } do
+
+  * `('/path/to/foo.txt', ssl_verify: true)` are the _resource parameters_. Resources take one or more parameters.
```
adamleff: Same as above re: parentheses.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+  * Note: Filters are used on plural resources
```
adamleff: Does this belong here, or in the section below where we actually define a filter?

I'm also not sure if we want to say this. We have other resources, like `passwd` that doesn't LOOK like a "plural resource" because it doesn't end in `s`, but it's totally filterable.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+  * `foos` is a _plural resource_
+
+### describe foos **('/path/to/foo.txt', ssl_verify: true)**, where
+
+  * `('/path/to/foo.txt', ssl_verify: true)` are the _resource parameters_. Resources take one or more parameters.
+
+## Filters:
+
+### describe **foos('/path/to/foo.txt', ssl_verify: true)** where { names == 'blah' } do
```
adamleff: Missing a period before `where`... should be:

```
.where { names == 'blah '}
```

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+  * `('/path/to/foo.txt', ssl_verify: true)` are the _resource parameters_. Resources take one or more parameters.
+  * Note: Filters are used on plural resources
+
+### **where  { names == 'blah' }** is an example of a **filter**. Some resources support one or more filters.
```
adamleff: Should include `.` before `where`

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+
+  * `where  { names == 'blah' }` is an example of a _filter clause_
```
adamleff: Same re: missing period.

docs/glossary.md
```diff
@@ -0,0 +1,100 @@
...
+### describe **foos('/path/to/foo.txt', ssl_verify: true)** where { names == 'blah' } do
+
+  * `('/path/to/foo.txt', ssl_verify: true)` are the _resource parameters_. Resources take one or more parameters.
+  * Note: Filters are used on plural resources
+
+### **where  { names == 'blah' }** is an example of a **filter**. Some resources support one or more filters.
+
+  * `where  { names == 'blah' }` is an example of a _filter clause_ 
+
+### **{ names == 'blah' , has spots}** are filter criteria
```
adamleff: This is actually bad syntax, and `has spots` would never work.

```
{ names == 'my-name' && spots == true }
```
... would be a better example.

adamleff (dismissed):
Hi @kagarmoe - bunch of feedback to make sure this is super-accurate technically. Let me know if you have any questions.

docs/glossary.md
kagarmoe: OK

docs/glossary.md
kagarmoe: OK

docs/glossary.md
kagarmoe: OK

docs/glossary.md
kagarmoe: OK

docs/glossary.md
kagarmoe: tenor-140626189
Would not have seen that one on my own!

docs/glossary.md
kagarmoe: OK

docs/glossary.md
kagarmoe: Some resources, such as `etc_hosts` are explicitly plural, while others, such as `passwd` are implicitly plural.

docs/glossary.md
kagarmoe: ok

docs/glossary.md
kagarmoe: ok

docs/glossary.md
kagarmoe: I respect you for going to the mat on that.

docs/glossary.md
kagarmoe: ok

docs/glossary.md
```diff
@@ -0,0 +1,99 @@
...
+### describe foos **('/path/to/foo.txt', ssl_verify: true)**, where
+
+  * `'/path/to/foo.txt'` and `ssl_verify: true` are the _resource parameters_. Resources take one or more parameters.
+
+## Filters:
+
+### describe foos ('/path/to/foo.txt', ssl_verify: true)**.where { names == 'blah' }** 
+
+  * `.where  { names == 'blah' }` is an example of a **filter**. 
+  * `.where  { names == 'blah' }` is an example of a _filter clause_
```
jquick: This line confused me. Is this line of code both a `filter` and a `filter clause`?

jquick (approved):

chris-rock (commented):
I think this is a great start. Thank you for doing the groundwork @kagarmoe  We want to make this more visual in the next iterations though.

chris-rock (approved):

PR #2455, opened 2018-01-12
Update firewalld.rb

markri:
Prevent the rule being parsed as different arguments by adding quotes on the rule


adamleff:
Thanks for your contribution, @markri! This change is already accounted for in PR #2430 which is just awaiting a second review and should be merged in next week. Therefore, I'm going to be closing this PR.

PR #2456, opened 2018-01-12
new docker_service resource to inspect Docker Swarm services

mattlqx:
This change adds the `docker_service` resource for Docker swarm mode services. This
branches off some of the common elements (id, exists) into a `DockerObject` module along
with a utility function for parsing the image/repo string. That function was implemented
separately by `docker_image` and `docker_container`, now with a third resource, it made
sense to consolidate that into an included module. I used the most comprehensive
implementation. Existing classes had to be slightly modified for the genericization.

Signed-off-by: Matt Kulka <mkulka@parchment.com>


kmf:
Excited for this!

lib/resources/docker_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def sanitize_options(opts)
+      opts.merge(parse_components_from_image(opts[:image]))
+    end
+
+    def object_info
+      return @info if defined?(@info)
+      opts = @opts
+      @info = inspec.docker.services.where {
+        name == opts[:name] || image == opts[:image] || (!id.nil? && !opts[:id].nil? && (id == opts[:id] || id.start_with?(opts[:id])))
```
adamleff, line 90: Can you explain why the special treatment around `id`? Specifically checking to see if the parameters are not `nil` before using them to compare? This is probably worthy of a code comment explaining why.

adamleff (commented):
@mattlqx this is a stellar change, very high quality code and documentation contribution. Thank you for that!

I have one question for you in-line before I approve to move this forward. Thank you!

lib/resources/docker_service.rb
mattlqx: Good question. It's already in master in the `docker_container` and `docker_image` resources. https://github.com/chef/inspec/blob/master/lib/resources/docker_container.rb#L111

It seems to be saying that if the object has an ID and the resource has an ID, then match if they're equal or if the ID is an abbreviated match (as Docker accepts arbitrarily shorter hash matches for objects if there's only one possibility). I'm not aware of any circumstances where an object in Docker wouldn't have an ID so that check may be over-safe but the rest is pretty legit.

mattlqx:
Thanks for the praise. This change is mostly a light refactor of prior art though so most of it should go to @chris-rock . This should make it easier to round out the collection of Docker objects though with `network`, `config` and `secret`. I might go deeper in the near future.

adamleff (approved):

arlimus (approved):
@mattlqx This is amazing work, huge kudos for the rewrite and improvements as well as the `docker_service` resource!!

kmf:
@mattlqx thanks for this awesomeness

PR #2457, opened 2018-01-12
http resource: make header keys case insensitive

### LINKED ISSUE #2453: http headers documentation is incorrect
### Description

The documentation on https://www.inspec.io/docs/reference/resources/http/ is incorrect.

Capitalization of http headers like `its('headers.Location') { should cmp 'http://foo.com' }` results in a false failure like:

```
  expected: "https://foo.com"
          got:
```

Because `headers.Location` is empty.

The correct way to do it is `its('headers.location') { should cmp 'http://foo.com' }`

1.45.13

### Replication Case

This will pass when the L in location is lower case (like below) and fail if it's upper case.

```
  describe http('https://api.particle.io', enable_remote_worker: true) do
    its('status') { should eq 302 }
    its('headers.location') { should cmp 'https://docs.particle.io/reference/api/' }
  end
```

### Possible Solutions

Update the documentation. Help me understand how I could have done this myself. I looked for an "update the docs" link on the Inspec docs site but couldn't find it.

adamleff:
HTTP header keys are currently case-sensitive, and the local and remote workers currently store the keys in different formats due to the different tools generating them.

This change ensures the ability to fetch headers by key is case-insensitive.

Fixes #2453

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -241,20 +241,11 @@ module Inspec::Resources
       end
 
       def [](requested_key)
-        unless requested_key.match(/[[:upper:]]/).nil?
-          warn '[DEPRECATION] HTTP resource: only use lower-case letters when specifying a header name.'
-        end
-
         fetch(requested_key.downcase, nil)
       end
 
       def method_missing(requested_key)
-        requested_key = requested_key.to_s
-        unless requested_key.match(/[[:upper:]]/).nil?
-          warn '[DEPRECATION] HTTP resource: only use lower-case letters when specifying a header name.'
-        end
-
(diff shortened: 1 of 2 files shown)


adamleff:
Added a deprecation in this PR that should be removed in 3.0. Removing in 2.0 seems rather sudden given that the 2.0 release is currently targeted for next month.

adamleff:
I originally had some deprecation logic in here to warn users if they tried to fetch a key containing upper-case letters but I reconsidered. HTTP headers are case-insensitive per the RFC, there's no reason not to leave this logic in place long-term to make it easier on our users.

chris-rock (approved):
Thank you @adamleff

PR #2459, opened 2018-01-15
pip resource docs: mention custom path to pip

mgrobelin:
... as introduced by GH-2097

The docs lacks a note on how to pass a custom path to the `pip` resource, which is especially useful when checking packages within virtualenv's or if python2/python3 (pip2/pip3) is installed on the same system.

Signed-off-by: Markus Grobelin <grobi@koppzu.de>


adamleff (approved):
Sweet, thanks @mgrobelin!

PR #2461, opened 2018-01-15
SMTP example for the ini resource was incorrect and did not work

djenniex-bene:
SMTP example in the resource documentation was incorrect and did not work when used in a test application.

### DIFF
--- a/docs/resources/ini.md.erb
+++ b/docs/resources/ini.md.erb
@@ -53,8 +53,8 @@ The following examples show how to use this InSpec audit resource.
 For example, a PHP INI file located at contains the following settings:
 
     [mail function]
-    ; SMTP = smtp.gmail.com
-    ; smtp_port = 465
+    SMTP = smtp.gmail.com
+    smtp_port = 465
 
 and can be tested like this:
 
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you for  spotting this and your first contribution @djenniex-bene

adamleff (changes requested):
Unfortunately, the docs are still wrong. The `php.ini` file is kind of a bad example because it's evaluated as PHP code, meaning lines beginning with `;` are treated as comments. However, standard INI files don't have that support.

So, when you parse the example in our docs, you actually get params that contain the semicolon:

```
inspec> ini('/tmp/test.ini').params
=> {"mail function"=>{"; SMTP"=>"smtp.gmail.com", "; smtp_port"=>"465"}}
```

So, we should change the doc example to read:

```
[mail function]
SMTP = smtp.gmail.com
smtp_port = 465
```

adamleff (approved):
Thanks for the docs fix, @djenniex-bene, and thank you for your first contribution to InSpec! Welcome to the community!

PR #2462, opened 2018-01-16
Update security_policy resource to return Names, not SIDs

ViolentOr:
Added possibility to translate SID to human-readable name. As it would be a breaking change, i added optional 'translate_sid' param to the resource.

Calling the resource 
```ruby
security_policy
```
would return previous output.

Calling it 
```ruby
security_policy(translate_sid: true)
```
will attempt to translate sid to human-readable name.

For example
```ruby
describe security_policy(translate_sid: true) do       
  its('SeNetworkLogonRight') { should include "NT AUTHORITY\\Authenticated Users" }
end
```
will return

>  Security Policy
>      [PASS]  SeNetworkLogonRight should include "NT AUTHORITY\\Authenticated Users"

### DIFF
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -154,7 +154,7 @@ module Inspec::Resources
         if @translate_sid
           val.split(',').map { |v|
             object_name = inspec.command("(New-Object System.Security.Principal.SecurityIdentifier(\"#{v.sub('*S', 'S')}\")).Translate( [System.Security.Principal.NTAccount]).Value").stdout.to_s.strip
-            object_name.blank? ? v.sub('*S', 'S') : object_name
+            object_name.empty? || object_name.nil? ? v.sub('*S', 'S') : object_name
           }
         else
           val.split(',').map { |v|
--- a/lib/resources/security_policy.rb
+++ b/lib/resources/security_policy.rb
@@ -151,10 +151,10 @@ module Inspec::Resources
         val.to_i
       # special handling for SID array
       elsif val =~ /[,]{0,1}\*\S/
-        if (@translate_sid)
+        if @translate_sid
           val.split(',').map { |v|
-              temp = inspec.command("(New-Object System.Security.Principal.SecurityIdentifier(\"#{v.sub('*S', 'S')}\")).Translate( [System.Security.Principal.NTAccount]).Value").stdout.to_s.strip
-              v = temp.blank? ? v.sub('*S', 'S') : temp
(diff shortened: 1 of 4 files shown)


lib/resources/security_policy.rb
```diff
@@ -74,8 +74,16 @@ class SecurityPolicy < Inspec.resource(1)
       describe security_policy do
         its('SeNetworkLogonRight') { should include 'S-1-5-11' }
       end
+
+      describe security_policy(translate_sid: true) do
+        its('SeNetworkLogonRight') { should include 'NT AUTHORITY\\Authenticated Users' }
+      end
     "
 
+    def initialize(pars = {})
```
adamleff: To be consistent with other resources, can you change this from `pars` to `opts`?

lib/resources/security_policy.rb
```diff
@@ -142,10 +150,17 @@ def extract_value(val)
...
         val.to_i
       # special handling for SID array
-      elsif val =~ /^\*\S/
-        val.split(',').map { |v|
-          v.sub('*S', 'S')
-        }
+      elsif val =~ /[,]{0,1}\*\S/
+        if @translate_sid
+          val.split(',').map { |v|
+            temp = inspec.command("(New-Object System.Security.Principal.SecurityIdentifier(\"#{v.sub('*S', 'S')}\")).Translate( [System.Security.Principal.NTAccount]).Value").stdout.to_s.strip
```
adamleff: A better variable name here would be `policy_name` or something similar to that, rather than `temp`

adamleff (changes requested):
Hi, @ViolentOr! Thank you for your contribution!

I have some suggestions for you on some better naming of variables. Also, before we can accept this, we need to add some unit tests to cover this new behavior. The two tests I can think of right away are:

 - set translate_sid to true and expect InSpec to successfully translate
 - set translate_sid to true, expect InSpec to fail to translate, and ensure the SID is returned instead of the failed translated name

Please let us know if you need guidance on how to proceed with adding those unit tests. Again, thanks for your contribution!

ViolentOr:
@adamleff thank you for suggestions. Some of them are addressed.
As for

> Also, before we can accept this, we need to add some unit tests to cover this new behavior.

I'll do my best. But i don't have ruby-development environment, it would be kind of walking in the dark. Tests are not hard though, i think i'll manage =)

Update:
/sigh
I guess, I'll have to install ruby after all. Oh, well. I'll do it later.

ViolentOr:
@adamleff take a look now, please.

I'm not sure, what is wrong with the TravisCI test... but it doesn't look like this PR is the reason of it =)

adamleff (approved):
Looks good to me, thanks @ViolentOr!

arlimus (approved):
I love the addition of translated SIDs, thank you for contributing this @ViolentOr !!

Issue #1728, opened 2017-04-27
Remove deprecations for InSpec 2.0


adamleff:
This issue will serve as a checklist of items we need to remove from the codebase as part of the InSpec 2.0 release. Since these are currently deprecations, they will become expected errors, etc. with InSpec 2.0. Therefore, these should not be removed until right before the 2.0 release.

InSpec core:

- [x] #1723: `assignment_re` and `key_vals` config keys
- [x] #2339: remove exec `--cache` flag
- [x] #2397: remove lowercase os name protection
- [x] remove `resource_skipped` (replaced by `resource_exception_message`)

InSpec matchers:

* [x] `contain_legacy_plus`, should have been removed in 1.0
* [x] `contain_match`, should have been removed in 1.0
* [ ] `contain_duplicates`, should have been removed in 1.0
* [x] `with_version` in favor of `its('version')`
* [x] `belong_to_group` in favor of `its('groups') { should include(...) }`
* [x] `belong_to_primary_group` in favor of `its('group') { should eq ... }`
* [x] `contain` in favor of `its('content') { should include(...) }`
* [x] `with_version` in favor of `its('version') { should eq ... }`

InSpec resources:

* `passwd`
  * [ ] `passwd.count`, should have been removed in 1.0
  * [ ] `passwd.usernames`, should have been removed in 1.0
  * [ ] `passwd.username`, should have been removed in 1.0
  * [ ] `passwd.uid`, should have been removed in 1.0
* `auditd_rules` - replaced with `auditd` in #2133

arlimus:
Thank you for keeping track of this @adamleff !!

arlimus:
Added a few more.

arlimus:
Some of these might be old and we can use create discussion threads if needed :)

username-is-already-taken2:
Hi there, 

I would like to find a way of helping you with getting inspec to 2.0.  May I ask how do you manage the PR's for the release?  For something like removing passwd.count from the passwd resource.  If I was to raise a PR that removes that functionailty do you store these up or apply them to a 2.0 branch that you raise to master at some point?

Never been involved with something like this, happy to chip away at the dockey work in the evening if it helps.

Gary

adamleff:
Hi, @username-is-already-taken2! We greatly appreciate your willingness to help!

This kind of problem is always tricky... no matter what approach, we're going to be dealing with merge conflicts, etc.  So here's what we agreed to do a while back:

[...]

PR #2464, opened 2018-01-17
Add new "reporter" system (replacement for "formatters"), support multiple reporters per run

jquick:
Fixes https://github.com/chef/inspec/issues/2415

This PR includes the full reporter/formatter refactor for InSpec. There should be no breaking changes in this PR and it should be fully compatible with InSpec 1.x.

This change also adds the new `--show-progress` flag which will show the standard dot notation progress but shows our text output instead of the rspec progress text output. 

This is currently designed with the ability to send a hash of info to a reporter object. This is called from the runner level so should be very easy to add any attributes or profile metadata that a future reporter may need.

Fixes https://github.com/chef/inspec/issues/2412
- profile_sha256 added to the json_min output.

Fixes https://github.com/chef/inspec/issues/1823
- generated_with_inspec_version added to the profile output.

Fixes https://github.com/chef/inspec/issues/2346
- multiple reports are part of this change.

Fixes https://github.com/chef/inspec/issues/1754
- you can output any report to a file.

Fixes https://github.com/chef/inspec/issues/1546
- Resource with options now show on title


aaronlippold:
👍

lib/inspec/base_cli.rb
```diff
@@ -57,12 +57,16 @@ def self.exec_options
       profile_options
       option :controls, type: :array,
         desc: 'A list of controls to run. Ignore all other tests.'
+      # TODO: remove in inspec 2.0
       option :format, type: :string,
-        desc: 'Which formatter to use: cli, progress, documentation, json, json-min, junit'
+        desc: '[DEPRECATED] Please use --format - this will be removed in InSpec 2.0'
```
mgrobelin: typo: `--format` is deprecated in favor of `--reporter`

mgrobelin:
Great additions, thank you!

lib/inspec/base_cli.rb
jquick: Thanks!

lib/inspec/base_cli.rb
```diff
@@ -57,12 +57,16 @@ def self.exec_options
       profile_options
       option :controls, type: :array,
         desc: 'A list of controls to run. Ignore all other tests.'
+      # TODO: remove in inspec 2.0
       option :format, type: :string,
-        desc: 'Which formatter to use: cli, progress, documentation, json, json-min, junit'
+        desc: '[DEPRECATED] Please use --reporter - this will be removed in InSpec 2.0'
```
adamleff: I'd argue that we're too close to add deprecations for 2.0. I think I said unofficially that Jan 1 was the last day for 2.0 deprecations. I'd mark this for removal in 3.0.

lib/inspec/base_cli.rb
```diff
@@ -57,12 +57,16 @@ def self.exec_options
...
+      option :reporter, type: :array,
+        desc: 'Which reporter(s) to use: cli, documentation, html, progress, json, json-min, json-rspec, junit'
```
adamleff: I'd change this to "Enable one or more output reporters: cli, documentation..."

lib/inspec/base_cli.rb
```diff
@@ -57,12 +57,16 @@ def self.exec_options
...
       option :color, type: :boolean,
         desc: 'Use colors in output.'
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
+      # TODO: remove in inspec 2.0
```
adamleff: Seems silly to say this, but I would not include this in this PR. We should already have it tracked on our 2.0-deprecation-removal issue #1728

lib/inspec/base_cli.rb
```diff
@@ -71,11 +75,16 @@ def self.exec_options
...
+      option :show_progress, type: :boolean,
+        desc: 'Show progress while executing tests.'
     end
 
     def self.default_options
       {
         exec: {
+          reporter: {
+            'cli' => nil,
+          },
```
adamleff: Do we need to include a default_option for the new `show_progress` option as well?

lib/inspec/cli.rb
```diff
@@ -32,24 +32,27 @@ class Inspec::InspecCLI < Inspec::BaseCLI
...
     o = opts.dup
+    diagnose(o)
     o[:ignore_supports] = true
     o[:backend] = Inspec::Backend.create(target: 'mock://')
     o[:check_mode] = true
 
     profile = Inspec::Profile.for_target(target, o)
+    info = profile.info
+    # add in inspec version
+    info[:generated_with_inspec_version] = Inspec::VERSION
```
adamleff: This feels like an unnecessarily long key... would `:inspec_version` be sufficient?

lib/inspec/cli.rb
```diff
@@ -193,16 +195,19 @@ def detect
   target_options
   option :command, aliases: :c,
     desc: 'A single command string to run instead of launching the shell'
+  # TODO: remove in inspec 2.0
   option :format, type: :string, default: nil, hide: true,
-    desc: 'Which formatter to use: cli, documentation, html, json, json-min, junit, progress'
+    desc: '[DEPRECATED] Please use --format - this will be removed in InSpec 2.0'
```
adamleff: Same as above - please target this for 3.0.

lib/inspec/reporters.rb
```diff
@@ -0,0 +1,27 @@
...
+    when 'json'
+      output = Inspec::Reporters::Json.new(config).render
+    when 'json-min'
+      output = Inspec::Reporters::JsonMin.new(config).render
+    when 'junit'
+      output = Inspec::Reporters::Junit.new(config).render
+    end
+
+    path = config[:path]
+    if path.nil? or path.strip == '-'
```
adamleff: use `||` instead of `or`

lib/inspec/reporters/cli.rb
```diff
@@ -0,0 +1,396 @@
...
+        'unknown'  => '?',
+      }.freeze
+    end
+
+    MULTI_TEST_CONTROL_SUMMARY_MAX_LEN = 60
+
+    def render
+      run_data[:profiles].each do |profile|
+        @control_count = 0
+        output ''
```
adamleff: Let's be consistent and add parentheses around method parameters:

```ruby
output('')
```

... and fix that across this PR.

lib/inspec/reporters/cli.rb
```diff
@@ -0,0 +1,396 @@
...
+        output format_message(
+          indentation: 5,
+          message: 'No tests executed.',
+        ) if @control_count == 0
+      end
+
+      output ''
+      print_profile_summary
+      print_tests_summary
+      @output
```
adamleff: This feels like a code smell that subclasses need to know to return an instance variable containing they output they've been creating (by calling a method that appends to an instance variable), which is pretty implementation-specific.

I'd encourage you to refactor this a bit. I'm thinking something needs to change in `Inspec::Reporters.render` so it looks something like this:

```ruby
def self.render(config)
  reporter = case config[:name]
             when 'cli'
               Inspec::Reporters::CLI.new(config).render
             when 'json'
               Inspec::Reporters::Json.new(config).render
             # etc...
             end
  reporter.render # does the actual rendering of text and storage, and should not attempt to do it again if we've already stored rendered text
  reporter.rendered_output # returns the actual output, which can be called multiple times if eneded
end
```

lib/inspec/reporters/json.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Inspec::Reporters
+  class Json < Base
+    def render
+      output clean_json.to_json
+    end
+
+    private
+
+    def clean_json
```
adamleff: The way this is implemented leaves us open to regressions if someone adds additional fields to the base formatter.

Since there's an associated schema, I'd rather see us actually build the object to match the schema rather than delete fields. That way if additional fields are added, they don't end up in this output.

lib/inspec/runner.rb
```diff
@@ -101,6 +101,31 @@ def run(with = nil)
       run_tests(with)
     end
 
+    def render_output(run_data)
+      # Test kitchen and the audit cookbook just call the runner
+      # so we need to default to cli here if its not set.
+
+      # copy over any of the old syntex --format
+      # until its remove in 2.0
```
adamleff: grammar nit: `until it's removed in 3.0`

lib/inspec/runner.rb
```diff
@@ -101,6 +101,31 @@ def run(with = nil)
       run_tests(with)
     end
 
+    def render_output(run_data)
```
adamleff: It's kind of a bummer that some of this is a subset of the logic already in the CLI code. Can we move this out into a helper somewhere, either here in the Runner or elsewhere, so that both the CLI and the Runner can use it without needing to dual-define it? It seems like TK could benefit from the additional logic already available in the CLI code.

lib/inspec/runner_rspec.rb
```diff
@@ -98,13 +99,30 @@ def reset
...
-      'junit' => 'InspecRspecJUnit',
-    }.freeze
+    # Set optional formatters and output
+    #
+    #
+    def set_optional_formatters
+      return if @conf[:reporter].nil?
+      if @conf[:reporter].key?('json-rspec')
+        if @conf[:reporter]['json-rspec'].nil?
+          RSpec.configuration.add_formatter(Inspec::Formatters::RspecJson)
```
adamleff: We could probably use an explanation comment here and at line 119 explaining why we can't pass a second argument of `nil` - i.e. what happens if we do that?

adamleff (changes requested):
Great job, @jquick - this is going to be wonderful. Here's my first pass.

jquick:
@adamleff - I think I have addressed all your comments. If you are ok with the direction of them I will add some additional tests around the new work.

adamleff:
@jquick yes - your latest commit in dc28ee2 is looking MUCH better. Lemme know when you've finished adding tests etc. and I'll give it another pass.

jquick:
@adamleff - Tests added, should be good for another look over.

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,58 @@ def self.exec_options
...
+      opts
+    end
+
     private
 
     # helper method to run tests
     def run_tests(targets, opts)
       o = opts.dup
-      log_device = opts['format'] == 'json' ? nil : STDOUT
+      log_device = o[:reporter]&.keys&.grep(/json/)&.any? ? nil : STDOUT
```
adamleff: This will match `json`, `rspec-json`, and `json-min`, right? Is that intended?  I think that's probably OK. We probably want the same treatment for `junit` too?

A future improvement on this might be to have a method on each reporter that we can call called `suppress_log_output?` or something similar so we do this more programmatically than just string matching.

lib/inspec/reporters.rb
```diff
@@ -0,0 +1,30 @@
...
+    case config[:name]
+    when 'cli'
+      reporter = Inspec::Reporters::CLI.new(config)
+    when 'json'
+      reporter = Inspec::Reporters::Json.new(config)
+    when 'json-min'
+      reporter = Inspec::Reporters::JsonMin.new(config)
+    when 'junit'
+      reporter = Inspec::Reporters::Junit.new(config)
+    end
```
adamleff: Is it possible for this method to be called with a reporter name string we're not expecting? Should we have a default `else` here that raises a `NotImplementedError`?

lib/inspec/reporters/base.rb
```diff
@@ -0,0 +1,23 @@
...
+  class Base
+    attr_reader :run_data
+
+    def initialize(config)
+      @run_data = config[:run_data]
+      @output = ''
+    end
+
+    def output(str)
+      @output << "#{str}\n"
```
adamleff, line 11: Alternatively, one could make `@output` an array that we continually `<<` onto, and `rendered_output` can `join("\n")`

adamleff (changes requested):
I think we're in the home stretch on this one, @jquick!

lib/inspec/reporters/base.rb
jquick: This is a good note. I think I prefer to keep it a string at this time but will keep this idea for the future.

jquick:
@adamleff - I believe everything is updated.

adamleff (approved):
Excellent work, @jquick!

lib/inspec/cli.rb
```diff
@@ -128,35 +134,34 @@ def vendor(path = nil)
...
   rescue StandardError => e
     pretty_handle_exception(e)
   end
 
   desc 'exec PATHS', 'run all test files at the specified PATH.'
   exec_options
   def exec(*targets)
-    diagnose
     o = opts(:exec).dup
+    diagnose(o)
```
chris-rock, line 162: I think we should verify at this point that we got a valid reporter. Otherwise this happens:
```
b inspec exec supermarket://dev-sec/ssh-baseline --reporter cli,json                                                  x1 7d jq/formatter-refactor[d12b6d3e]
[2018-02-01T11:55:54+01:00] WARN: URL target https://github.com/dev-sec/ssh-baseline transformed to https://github.com/dev-sec/ssh-baseline/archive/master.tar.gz. Consider using the git fetcher
bundler: failed to load command: inspec (/usr/lib/ruby/gems/2.4.0/bin/inspec)
NotImplementedError: 'cli,json' is not a valid reporter type.
  /pub/go/src/github.com/chef/inspec/lib/inspec/reporters.rb:19:in `render'
  /pub/go/src/github.com/chef/inspec/lib/inspec/runner.rb:115:in `block in render_output'
  /pub/go/src/github.com/chef/inspec/lib/inspec/runner.rb:109:in `each'
  /pub/go/src/github.com/chef/inspec/lib/inspec/runner.rb:109:in `render_output'
  /pub/go/src/github.com/chef/inspec/lib/inspec/runner.rb:133:in `run_tests'
  /pub/go/src/github.com/chef/inspec/lib/inspec/runner.rb:101:in `run'
  /pub/go/src/github.com/chef/inspec/lib/inspec/base_cli.rb:145:in `run_tests'
  /pub/go/src/github.com/chef/inspec/lib/inspec/cli.rb:172:in `exec'
  /usr/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
  /usr/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
  /usr/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
  /usr/lib/ruby/gems/2.4.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
  /pub/go/src/github.com/chef/inspec/bin/inspec:12:in `<top (required)>'
  /usr/lib/ruby/gems/2.4.0/bin/inspec:23:in `load'
  /usr/lib/ruby/gems/2.4.0/bin/inspec:23:in `<top (required)>'
```
Inspec download the profile although it will not be able to report, which slows down user feedback. I think we can be more quick if we do an upfront checking.

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
+      # Catch any dynamic inspec runners or tests
+      opts['reporter'] = { 'cli' => nil } if opts['reporter'].nil?
+
+      opts
+    end
+
     private
 
+    def suppress_log_output?(opts)
+      match = %w{json json-min json-rspec junit} & opts['reporter'].keys
```
chris-rock: Is it okay to output the logs for the html reporter?

lib/inspec/schema.rb
```diff
@@ -147,6 +147,7 @@ class Schema
       'properties' => {
         'id' => { 'type' => 'string' },
         'profile_id' => { 'type' => %w{string null} },
+        'profile_sha256' => { 'type' => 'string' },
```
chris-rock, line 150: @arlimus should it be simply be `sha256` or `profile_sha256`?

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
+          stdout += 1 if v.nil?
+        end
+
+        raise ArgumentError, 'The option --reporter can only have a single report outputting to stdout.' if stdout > 1
+
+        opts['reporter'] = reports
+      end
+
+      # Catch anyone using the legacy --format option
+      unless opts['format'].nil?
```
chris-rock: I found this not intutive, if i use `inspec exec supermarket://dev-sec/ssh-baseline --format json-min --reporter junit` I would expect that reporter overwrites `format`. format should only be used if reporter is not set. In addition I would like to see an deprecation warning on cli if `format` is still used.

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
         },
       }
     end
 
+    def self.clean_reporters(opts)
+      # Parse report options from CLI
+      if opts['reporter'].is_a?(Array)
+        stdout = 0
+        reports = {}
+        opts['reporter'].each do |report|
```
chris-rock: This is not a safe implementation. I can easily break it with:

```
{ 
    "backend_cache": true,
    "reporter": { 
        "junit" : null,
        "cli" : null
    }
}
```
And then calling `inspec exec supermarket://dev-sec/ssh-baseline --json-config config.json `

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
+          k, v = report.split(':')
```
chris-rock: Interessting hidden feature: `inspec exec supermarket://dev-sec/ssh-baseline --reporter json:a.json junit:a.xml` Is that documented somewhere?

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
+          v = nil if v == '-'
+          reports[k] = v
+          stdout += 1 if v.nil?
+        end
+
+        raise ArgumentError, 'The option --reporter can only have a single report outputting to stdout.' if stdout > 1
+
+        opts['reporter'] = reports
```
chris-rock: I propose a slightly different structure for internal reporter representation:
```
{
  "backend_cache": true,
  "reporter": {
    "junit": [
      {
        "file": "a.xml"
      }
    ],
    "json": [
      {
        "file": "a.json"
      }
    ]
  }
}
```
This allows us:
- more options for future reporter (e.g. the automate reporter that is coming with auth headers etc)
- we can parse the cli options into that structure and implement one validator method in the reporter implementation
- we do not need to distingish between cli and json config

We may enable yml as configuration in future. Above config would look like:
```
---
backend_cache: true
reporter:
  junit:
  - file: a.xml
  json:
  - file: a.json

```

test/unit/base_cli_test.rb
```diff
@@ -13,7 +13,7 @@
       Inspec::BaseCLI.stubs(:default_options).returns(default_options)
 
       opts = cli.send(:merged_opts, :exec)
-      expected = { 'format' => 'json', 'backend_cache' => false }
+      expected = { 'format' => 'json', 'backend_cache' => false, 'reporter' => { 'cli' => nil } }
```
chris-rock: I think we should not have json and cli reporter enabled at the same time

lib/inspec/reporters.rb
```diff
@@ -0,0 +1,32 @@
...
+    case config[:name]
+    when 'cli'
+      reporter = Inspec::Reporters::CLI.new(config)
+    when 'json'
+      reporter = Inspec::Reporters::Json.new(config)
+    when 'json-min'
+      reporter = Inspec::Reporters::JsonMin.new(config)
+    when 'junit'
+      reporter = Inspec::Reporters::Junit.new(config)
+    else
```
chris-rock, line 20: Just to be sure I understand the implementation correct, the `html` & `documentation`  reporter just works by accident?
@adamleff What do you think? Should this be `rspec-documentation`, `rspec-html`?

lib/inspec/reporters/json.rb
```diff
@@ -0,0 +1,132 @@
...
+
+module Inspec::Reporters
+  class Json < Base
+    def render
+      report = {
+        platform: platform,
+        profiles: profiles,
+        statistics: { duration: run_data[:statistics][:duration] },
+        version: run_data[:version],
+        controls: controls,
```
chris-rock, line 13: `controls` is deprecated and will be removed in InSpec 2.0, right?

chris-rock (changes requested):
Huge improvement @adamleff @jquick Well done! This makes everything so much easier and more extensible. I added some things we need to improve but really nothing conceptional.

lib/inspec/base_cli.rb
```diff
@@ -71,24 +73,70 @@ def self.exec_options
...
+        stdout = 0
+        reports = {}
+        opts['reporter'].each do |report|
+          k, v = report.split(':')
+          v = nil if v == '-'
+          reports[k] = v
+          stdout += 1 if v.nil?
+        end
+
+        raise ArgumentError, 'The option --reporter can only have a single report outputting to stdout.' if stdout > 1
```
chris-rock: This is interesting, we have the undocumented behavior to write the content to file. I think this error should like to an inspec docs page that explains multiple reporting options.

lib/inspec/base_cli.rb
adamleff: Yeah, that's a cool find actually. So we can either raise an exception if `opts['reporter']` isn't an array, or we can support both Array and Hash syntax. I'd prefer the latter, and then raise if we get anything other than Array or Hash.

lib/inspec/base_cli.rb
adamleff: Only hidden because we were waiting for feedback from you and Dom before writing the website docs. :) It won't be hidden upon acceptance.

lib/inspec/base_cli.rb
adamleff: That's definitely more flexible. I don't know that it needs to be addressed in this PR since this an internal representation and nothing the user sees, and we can translate into whatever internal structure we need as our own needs and featureset changes... but it shouldn't be a big deal to do it in this PR if you feel strongly about it.

lib/inspec/base_cli.rb
adamleff: To ensure backward compatibility (because of our desire to get this in to 1.x), `--format` and `--reporter` have to be able to co-exist. And a big deliverable of this change is that multiple reporters/formatters can be used simultaneously - a HUGE persistent request from the community. I think it's only unintuitive because you're reading the code, but a everyday InSpec user shouldn't even notice unless they become aware of the new feature and start using it.

Totally agree with the deprecation warning though. That could be added in a 2.x version though because it's way too late to deprecate this for 2.0.

lib/inspec/base_cli.rb
adamleff: Nope. I also wonder if we're making this hard on ourselves and shouldn't just change the default log destination to STDERR instead of STDOUT. If we do that, all of this selective suppression can go away because people can just redirect STDOUT to a file if they care without mucking with the reporter output. Even though it's technically a breaking change, part of me thinks that outputting log output to the same stream as formatted data is a bug, and I'd be happy to have us change this without requiring a major version bump. I don't think the log output is consumed by many except when debugging something.

lib/inspec/cli.rb
adamleff: That's great feedback, I agree.

lib/inspec/reporters.rb
adamleff: Not by accident... on purpose :) We handle that in the runner because that's where we need to invoke and register the RSpec-provided formatters. We have lots of users in the community that use the doc and html formatters, especially in CI pipelines or Test Kitchen.  This class is just setting up our new Reporter stuff that we own.

I'd vote against changing the names for two reasons:

 - it would be a breaking change, and lots of our TK community use the documentation formatter
 - it further calls attention to the fact that RSpec is used, and I'd love to be able to hide RSpec details from our users if at all possible.

That said, if further separation is required in the future, I'd totally support it. It would just need to be in a major version bump 3.0 or later at this point.

lib/inspec/reporters/json.rb
adamleff: Yes - addressed in #2487 which is blocked on this PR and will target the `release-2.0` branch.

test/unit/base_cli_test.rb
adamleff: Can you explain why?

adamleff:
Awesome review, @chris-rock - thank you so much.

lib/inspec/schema.rb
aaronlippold: I think `profile_sha256` communicates better and more clearly as to what the SHA is for.

When we then add things like: a id/'SHA' for the system, an id/sha for the 'run', etc. it will keep things more clear and seperated.

test/unit/base_cli_test.rb
aaronlippold: @chris-rock actually having them both active is one of the primary use cases, where we want the console output in Jenkins or gitlab CI/CD to record the CLI output and we want to produce an artifact json of the results that can be saved / processed. Or a user is running a kitchen job on their laptop and they want to see the CLI but also need to produce JSON to push to someone else or another system.

lib/inspec/cli.rb
jquick: Yes, this now happens during the option merging.

jquick:
@chris-rock - I think I have hit on everything you commented on. I will be adding documentation in another PR I think as this one is quite large already. Let me know what else you see.

lib/inspec/base_cli.rb
```diff
@@ -80,74 +80,97 @@ def self.exec_options
...
     end
 
-    def self.clean_reporters(opts)
-      # Parse report options from CLI
+    def self.parse_reporters(opts)
+      # merge in any legacy formats as reporter
+      # this method will only be used for ad-hoc runners
+      if !opts['format'].nil? && opts['reporter'].nil?
+        warn '[DEPRECATED] The option --format is being is being deprecated and will be removed in inspec 3.0. Please use --reporter'
+        opts['reporter'] = [opts['format']]
```
adamleff: If `opts['format']` has the chance of being an Array, this would do an array-inside-an-array which may not be what you intended.

If you just want to ensure that it's always an array, it's generally safer to do:

```ruby
opts['reporter'] = Array(opts['format'])
```

... while won't double-encode into another Array.

It's probably safe in this case since this gets populated via Thor and we control the type there, but something to keep in mind for the future.

lib/inspec/base_cli.rb
```diff
@@ -80,74 +80,97 @@ def self.exec_options
...
+        opts.delete('format')
+      end
+
+      # parse out cli to proper report format
       if opts['reporter'].is_a?(Array)
-        stdout = 0
         reports = {}
         opts['reporter'].each do |report|
           k, v = report.split(':')
```
adamleff: To help future readers, instead of `k, v` it might be clearer to use `reporter_name, target`

lib/inspec/base_cli.rb
```diff
@@ -80,74 +80,97 @@ def self.exec_options
...
         opts['reporter'] = reports
       end
 
-      # Catch anyone using the legacy --format option
-      unless opts['format'].nil?
-        opts['reporter'] = { opts['format'] => nil }
-        opts.delete('format')
+      # add in stdout if not specified
+      if opts['reporter'].is_a?(Hash)
+        opts['reporter'].each do |k, v|
```
adamleff: Same feedback here re: variable names

lib/inspec/base_cli.rb
```diff
@@ -80,74 +80,97 @@ def self.exec_options
...
+      ]
+
+      reporters.each do |k, _v|
+        raise NotImplementedError, "'#{k}' is not a valid reporter type." unless valid_types.include?(k)
+      end
+
+      # check to make sure we are only reporting one type to stdout
+      stdout = 0
+      reporters.each_value do |v|
+        stdout += 1 if v&.[]('stdout')&.==(true)
```
adamleff: heh, gotta love the safe navigation operator! But this syntax is super-hard to grok, especially the `[]` usage here.

In this case, I'd recommend something a bit more standard/legacy for readability reasons:

```ruby
stdout += 1 if v.key?('stdout') && v['stdout'] == true
```

... or, honestly, if you're validating the reporter data structure before you even get to this point, you can probably just do:

```ruby
stdout += 1 if v['stdout'] == true
```

lib/inspec/base_cli.rb
```diff
@@ -80,74 +80,97 @@ def self.exec_options
...
 
     def suppress_log_output?(opts)
-      match = %w{json json-min json-rspec junit} & opts['reporter'].keys
+      return false if opts['reporter'].nil?
+      match = %w{json json-min json-rspec junit html} & opts['reporter'].keys
       unless match.empty?
         match.each do |m|
           # check to see if we are outputting to stdout
-          return true if opts['reporter'][m].nil?
+          return true if opts['reporter'][m]&.[]('stdout') == true
```
adamleff: Again, while I love the safe-navigation operator, I think using it with things like `[]` makes it hard to read and understand. Are you not guaranteed at this point that `opts['reporter'][m]` will be a Hash?

lib/inspec/cli.rb
```diff
@@ -168,8 +168,13 @@ def exec(*targets)
       o[:logger].warn '[DEPRECATED] The use of `--cache` is being deprecated in InSpec 2.0. Please use `--vendor-cache` instead.'
     end
 
-    # run tests
-    run_tests(targets, o)
+    runner = Inspec::Runner.new(o)
+    targets.each { |target| runner.add_target(target) }
+
+    exit runner.run
+  rescue ArgumentError, RuntimeError, Train::UserError => e
```
adamleff, line 175: I'm curious what caused you to add this. This will swallow potentially useful stacktraces from dependencies that may throw these types of errors without helpful messaging.

lib/inspec/cli.rb
```diff
@@ -222,7 +227,7 @@ def shell_func
     exit res unless run_type == :ruby_eval
 
     # No InSpec tests - just print evaluation output.
-    res = (res.respond_to?(:to_json) ? res.to_json : JSON.dump(res)) if o['reporter'].keys.include?('json')
+    res = (res.respond_to?(:to_json) ? res.to_json : JSON.dump(res)) if o['reporter']&.keys&.include?('json')
```
adamleff: This is a much cleaner use of safe-navigation operator IMO.

adamleff:
@jquick I love the use of the safe-navigation operator but think in many cases it has detracted from the potential readability.

When dealing with Hashes, rather than doing `&.[]('key')`, I'd recommend doing `.fetch('key', nil)`.

I think this:

```ruby
my_hash&.[]('key')&.some_method
```

isn't as readable as this:

```ruby
my_hash&.fetch('key', nil)&.some_method
```

Perhaps personal opinion, but I'd error on being more verbose than clever on some of those examples.

jquick:
@adamleff  - Thanks for the feedback. I agree, sometimes I am not a fan of the safe-navigation and have thought about disabling the rubocop rule for some instances for readability. I do think the fetch syntax is much cleaner then the brackets, so thanks for that tip.

lib/inspec/cli.rb
jquick: This was pulled from the run_tests method so we capture these errors along with the Train specific ones.

adamleff: Got it. As discussed, I think this is fine. What I would love to see in the future is:

 - we capture all exceptions and just hand to `pretty_handle_exception`
 - `pretty_handle_exception` is modified to pint out both the class and the message, and then print the backtrace if log level is debug.

I think this provides a better UX, but totally doesn't have to be done in this PR.

miah (approved):

lib/inspec/reporters.rb
chris-rock: That makes sense @adamleff

chris-rock (approved):
Awesome improvement @jquick @adamleff

clintoncwolfe (commented):
So, it's big, and awesome.  I'm not sure about the UX, though - it's not obvious how to redirect multiple reporters to files (I eventually found it in the source - using junit:foo.xml, for example.)  I didn't see any docs file included in the PR, and the Thor help is pretty scant.

jquick:
@clintoncwolfe  - Yeah, the docs file will be coming in another PR (likely including this and the 2.0 reporter changes), this one is already passed max and I need to work with @kagarmoe to see the best place. I did updated Thor a tad to show the output path syntax. I didn't feel like I could add more verbiage to the description though as its already quite long.

chris-rock:
I agree with @clintoncwolfe feedback. I created #2546 to track this as pre-InSpec 2.0 work.

PR #2468, opened 2018-01-17
package resource: Add architecture support

### LINKED ISSUE #2463: Support for checking architecture within package resource
### Description
We have a requirement for installing the same package(compat-libstdc++-33) for a couple of architectures. The inspec "package" resource supports checking the version of a package installed but not architecture. This is good to have if we want to install the same package for different architectures for various reasons. Some example code like below.

describe package(‘my_xyz_package’) do
   it { should be_installed }
   its(os.arch) { should match(‘i686’)  }
   its(os.arch) { should match(‘x86_64’)  }
 end

The InSpec version is 1.25.1 and OS distribution we are planning is RHEL 7.x.
### Replication Case

It's not a bug so no replication possible at this point.

### Possible Solutions

Jerry Aldrich comes up with below code using command resource for my development to continue which worked flawlessly.

describe command('rpm -qa | grep compat-libstdc++-33') do
  its('stdout') { should match /x86_64/ }
  its('stdout') { should match /i686/ }
end

### Stacktrace

No Stacktrace possible at this point.

jerryaldrichiii:
This adds support for `architecture` to the package resource. Example:

```
describe package('curl') do
  its('architecture') { should cmp 'x86_64' }
end
```

This also adds support for multiple architectures. Example:
```
describe package('compat-libstdc++-33') do
  its('architecture') { should include 'i686' }
  its('architecture') { should include 'x86_64' }
end
```

This does not add `architecture` support for the following platforms due
to the query command not providing enough info:
  - AIX
  - HP-UX
  - macOS/Darwin
  - Solaris 11
  - Windows

This closes #2463

### DIFF
--- a/docs/resources/package.md.erb
+++ b/docs/resources/package.md.erb
@@ -118,8 +118,8 @@ You can also use the `cmp OPERATOR` matcher to perform comparisions using the ve
 
 `cmp` understands version numbers using Gem::Version, and can use the operators `==, <, <=, >=, and >`.  It will compare versions by each segment, not as a string - so '7.4' is smaller than '7.30', for example.
 
-### architecture
+### arch
 
-The `architecture` matcher tests the named package's architecture on the system:
+The `arch` matcher tests the named package's architecture on the system:
 
-its('architecture') { should incluce 'x86_64' }
+its('arch') { should incluce 'x86_64' }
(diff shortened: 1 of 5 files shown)


jerryaldrichiii:
The latest commit shortens `architecture` to `arch`. It is a personal preference and can be dropped if needed.

jerryaldrichiii:
After some internal discussion, I realize this is probably the wrong approach. Since we have the `packages` resource and we are essentially checking multiple packages I'm going to close this PR and instead add functionality to the `packages` resource to solve this.

PR #2469, opened 2018-01-18
packages resource: Add `architectures` support

### LINKED ISSUE #2463: Support for checking architecture within package resource

jerryaldrichiii:
This adds support for `architectures` to the `packages` resource.

Example:

```
describe packages(/compat-libstdc++-33/) do
  its('architectures') { should include 'x86_64' }
  its('architectures') { should include 'i686' }
end
```

This also adds documentation for the `packages` resource and closes #2463

### DIFF
--- a/docs/resources/packages.md.erb
+++ b/docs/resources/packages.md.erb
@@ -1,0 +1,66 @@
+---
+title: About the packages Resource
+---
+
+# packages
+
+Use the `packages` InSpec audit resource to test the properties of multiple packages on the system.
+
+<br>
+
+## Syntax
+
+A `packages` resource block declares a regular expression search to select packages
+
+    describe packages(/name/) do
+      its('statuses') { should cmp 'installed' }
+    end
+
+<br>
+
+## Examples
+
+The following examples show how to use this InSpec audit resource.
+
+### Verify that no `xserver` packages are installed
+
+    describe package(/xserver/) do
+      its('statuses') { should_not cmp 'installed' }
+    end
+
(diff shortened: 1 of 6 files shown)


adamleff (approved):
Excellent work, @jerryaldrichiii.

arlimus (approved):
Oh yay sweet work @jerryaldrichiii thank you for adding architectures 😁

Wolfsrudel:
Is this implemented already? It is not working in 1.51.2 on my installation. :/

adamleff:
It should be working as it was in the 1.51.1 build. Please open a new issue with as much detail as possible so we can troubleshoot it. Thanks.

PR #2473, opened 2018-01-20
service resource: attempt a SysV fallback if SystemD unit file is not found

### LINKED ISSUE #1708: systemctl is-enabled fails on Debian for SysV units
### Description

On Debian 8, a SysV unit (such as the stock `apache2` package) will fail the `systemctl is-enabled` command because Debian 8's systemd doesn't include the `systemd-sysv-install` shim.

1.19.1

### Replication Case

**recipe:**

```ruby
package 'apache2'

service 'apache2' do
  action :enable
end
```

**test:**

```ruby
describe service('apache2') do
  it { should be_enabled }
end
```

**inspec output:**

```
  Service apache2
     ∅  should be enabled
     expected that `Service apache2` is enabled
```

**manual command run:**

```
root@default-debian-87:~# systemctl is-enabled apache2
Failed to get unit file state for apache2.service: No such file or directory
root@default-debian-87:~# echo $?
1
```

### Possible Solutions

It would be great if InSpec could detect this failure and try the init script in `/etc/init.d` as a fallback.

jerryaldrichiii:
This modifies the `enabled?` check to fallback to `sysv_service` in the event that a `.service` file cannot be found.

For example: On Debian 8.7 the stock apache2 package does not deploy a `.service` file but deploys a SysV style service. This causes `systemctl is-enabled` to fail when the service is in fact enabled.

This closes #1708

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -196,8 +196,8 @@ class MockLoader
     }
 
     cmd_exit_1 = lambda { |x = nil|
-      stderr = ::File.read(::File.join(scriptpath, 'unit/mock/cmd', x)) if x
-      mock.mock_command('', '', stderr || '', 1)
+      stderr = x.nil? ? '' : File.read(File.join(scriptpath, 'unit/mock/cmd', x))
+      mock.mock_command('', '', stderr, 1)
     }
 
     mock.commands = {
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -195,12 +195,10 @@ class MockLoader
       mock.mock_command('', '', '', 0)
     }
 
-    cmd_stderr = lambda { |x|
-      stderr = ::File.read(::File.join(scriptpath, '/unit/mock/cmd/'+x))
-      mock.mock_command('', '', stderr, 1)
+    cmd_exit_1 = lambda { |x = nil|
+      stderr = ::File.read(::File.join(scriptpath, 'unit/mock/cmd', x)) if x
+      mock.mock_command('', '', stderr || '', 1)
     }
-
-    cmd_exit_1 = mock.mock_command('', '', '', 1)
 
     mock.commands = {
       '' => empty.call,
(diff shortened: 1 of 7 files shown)


test/helper.rb
```diff
@@ -195,6 +195,11 @@ def md.directory?
       mock.mock_command('', '', '', 0)
     }
 
+    cmd_stderr = lambda { |x|
+      stderr = ::File.read(::File.join(scriptpath, '/unit/mock/cmd/'+x))
```
adamleff: We don't need to do string concatenation or interpolation when we're already in a `File.join` call. It's also customary to remove the leading and trailing slashes and let `File.join` do its thing:

```ruby
stderr = ::File.read(::File.join(scriptpath, 'unit/mock/cmd', x))
```

I also question whether we really need another helper here, especially when this is also doing an `exit 1` like `cmd_exit_1` is already doing. Should we just enhance `cmd_exit_1` to accept an optional path to stderr output?

adamleff (changes requested):
Functionally looks good, @jerryaldrichiii. I think we can cleanup the test helper before merging though.

test/helper.rb
jerryaldrichiii: I agree with all those points! Update incoming.

Thanks for your always stellar reviews!

test/helper.rb
```diff
@@ -195,13 +195,11 @@ def md.directory?
       mock.mock_command('', '', '', 0)
     }
 
-    cmd_stderr = lambda { |x|
-      stderr = ::File.read(::File.join(scriptpath, '/unit/mock/cmd/'+x))
-      mock.mock_command('', '', stderr, 1)
+    cmd_exit_1 = lambda { |x = nil|
+      stderr = ::File.read(::File.join(scriptpath, 'unit/mock/cmd', x)) if x
```
adamleff: You might think I'm nitpicking here, but I think we can eliminate the conditional in the following line and just let `stderr` always be the right thing (rather than `mock_command` having another in-line conditional):

```ruby
stderr = x.nil? ? '' : ::File.read(::File.join(scriptpath, 'unit/mock/cmd', x))
mock.mock_command('', '', stderr, 1)
```

That's a bit more direct and clear, would you agree?

adamleff (changes requested):

test/helper.rb
jerryaldrichiii: Yup, I agree, elegant use of the ternary. Not nitpicky at all! 😄

adamleff (approved):
LGTM, @jerryaldrichiii - thanks!

arlimus (approved):
Great handling of a corner-case, love the fallback @jerryaldrichiii thank you!

tas50:
@jerryaldrichiii You've solved my #1 inspec bug. Thanks

PR #2479, opened 2018-01-25
Doc fixes the apache_conf properties returns an array

burtlo:
Updates the `apache_conf`'s attributes: `timeout`, `allowoverride`, `maxclients`, `Listen`

### DIFF
--- a/docs/resources/apache_conf.md.erb
+++ b/docs/resources/apache_conf.md.erb
@@ -37,7 +37,7 @@ The following examples show how to use this InSpec audit resource.
 ### Test ports for SSL
 
     describe apache_conf do
-      its('Listen') { should eq '443'}
+      its('Listen') { should include '443' }
     end
 
 <br>
@@ -57,5 +57,5 @@ For example:
 
     describe apache_conf do
       its('MaxClients') { should eq 100 }
-      its('Listen') { should eq '443'}
+      its('Listen') { should include '443' }
     end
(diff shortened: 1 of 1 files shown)


burtlo:
I don't know if this is intended. This is what I experienced with InSpec version `1.45.13`.

burtlo:
Testing this against CentOS 6.7 is where I started. Then I moved to Ubuntu 14.04. The default for that platform does not set at `maxclients` which results in a `nil`.

```
∅  /etc/apache2/apache2.conf MaxClients should include "300"
     expected nil to include "300", but it does not respond to `include?`
```

This feels like a bug to me with the `apache_conf` properties collector. Like it should create an empty array.

adamleff:
@burtlo I'd rather see these changed to `cmp` instead of `include`.  `cmp` will strip a single-item-array value out and do type matching if needed.

```
inspec> describe [443] do
inspec>   it { should cmp 443 }
inspec> end

Profile: inspec-shell
Version: (not specified)

  [443] should
     ✔  cmp == 443
```

This also avoids the issue of calling `include` on a potential `nil` value.

I don't know that I agree that the current properties returning nil are bugs. We don't know of all the various Apache configuration directives that may exist and if multiple values are supported, so if someone calls for one that isn't a valid directive, or asks for a directive that is not present in the configuration file, `nil` feels like the right value to return.

adamleff (changes requested):
Requesting `cmp` instead of `include` in the documents to address the discovered issue.

burtlo:
Alright, updated to use `cmp`. I threw in another example to demonstrate order independent comparison of multiple ports with the `=~` operator.

adamleff (approved):

Issue #2418, opened 2017-12-22
Fail resources when required content is not available


adamleff:
Historically, resources have called `skip_resource` when a required component is not available, such as a command or a file not being present. In most cases, it would be more appropriate for this to be a failure since InSpec was not able to complete the assertion. With the proper handling of `Inspec::Exceptions::ResourceFailed`, we can accomplish this now.

Since this would be a breaking change, this should be targeted for a major release bump. Let's see if we can get this in for InSpec 2.0.

clintoncwolfe:
I'm closing as a duplicate - this is covered in much more detail on #3174

Issue #1300, opened 2016-11-18
custom resource subclass of existing inpec resource no longer works


jeremymv2:
### Description

I’ve got a simple custom profile with a custom library that is a subclass of `JsonConfig` resource:
```
class NodeAttributes <JsonConfig
  name 'node'
  def initialize
    super('/tmp/node.json')
  end
end
```

That used to work, but now I get
```
libraries/node.rb:2:in `load_with_context': uninitialized constant #<Class:#<#<Class:0x007fa05aa40b08>:0x007fa05aa409f0>>::JsonConfig (NameError)
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:146:in `instance_eval'
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:146:in `load_with_context'
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:135:in `load_library_file'
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:124:in `block in load_libraries'
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:122:in `each'
    from /Users/jmiller/.chefdk/gem/ruby/2.3.0/gems/inspec-1.4.1/lib/inspec/profile_context.rb:122:in `load_libraries'
```
  
is that expected?

### InSpec and Platform Version

inspec 1.4.1  Mac OSX

### Replication Case

The profile is here: https://github.com/jeremymv2/test-inspec/tree/master/test/integration/profile

jeremymv2:
@chris-rock took a little peek into this.  The uninitialized constant error above can be easily fixed by adding `include Inspec::Resources` to the anonymous class that gets created for providing the profile context to be instance_eval'd against, just after this line:  https://github.com/chef/inspec/blob/master/lib/inspec/library_eval_context.rb#L36

However, the next issue is that the `node` object above never gets registered (as verified in pry) and is therefore unavailable in the inspec dsl scope.  It appears that the `Inspec.resource()` method is artificially creating an opportunity for resource registration at compile time that `class NodeAttributes <JsonConfig` doesn't allow for.

[...]

PR #2481, opened 2018-01-25
/etc/libuser.conf Configuration Resource

djhaynes:
This pull request adds support for a resource to check the configuration of the /etc/libuser.conf file.


lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
+require 'utils/simpleconfig'
+require 'utils/object_traversal'
+
+module Inspec::Resources
+  class EtcLibUser < Inspec.resource(1)
+    name 'etc_libuser'
+    desc 'Use the etc_libuser InSpec audit resource to check the configuration of /etc/libuser.conf.
+          Due to limitations of the RSpec engine and for consistency, please use the method described here
```
adamleff: I don't think this needs to be directly in the resource documentation. Perhaps you can suggest a different place for it, but this warning applies to anything that may have a `.` character in it across many resources and I would not want to litter this warning throughout all resource documentation.

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+          [import]
+          # Data from these files is used when libuser.conf does not define a value.
+          # The mapping is documented in the man page.
+          login_defs = /etc/login.defs
+          default_useradd = /etc/default/useradd
+          ...
+          ---end sample /etc/libuser.conf---
+
+          describe etc_libuser do
+            its(['import','login_defs']){should eq ['/etc/login.defs']}
```
adamleff: Please provide spacing that is consistent with the other documentation and examples. I'd also like to request that you change this example to use the `cmp` matcher so users don't have to worry about array syntax unless necessary:

```ruby
its(['import', 'login_defs']) { should cmp '/etc/login.defs' }
```

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+          end
+         "
+
+    include ObjectTraverser
+    attr_reader :params, :raw_content
+
+    class UnknownEtcLibUserConf < StandardError; end
+
+    def initialize(conf_path = nil)
+      @conf_path = conf_path || inspec.os_env('LIBUSER_CONF').content || '/etc/libuser.conf'
```
adamleff: Can you provide a use case for grabbing the path via an environment variable? This is not standard in our resources and is not something we advocate for a variety of reasons.

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+      file = inspec.file(@conf_path)
+      return skip_resource "The `etc_libuser` resource is not supported on your OS. Can't find '#{@conf_path}'." unless file.file?
```
adamleff: I think this should actually be a failure. If someone writes a test and the file isn't there, we cannot perform the requested assertion. We've recently changed to use exceptions for this also so it makes it easier to raise them anywhere in a resource, not just the `initialize` method.

I'd change this to:

```ruby
raise Inspec::Exceptions::FailResource, "Can't find '#{@conf_path}'" unless file.file?
```

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
+require 'utils/simpleconfig'
+require 'utils/object_traversal'
+
+module Inspec::Resources
+  class EtcLibUser < Inspec.resource(1)
+    name 'etc_libuser'
```
adamleff, line 6: On what operating systems should this resource run? We should add `supports` line(s) accordingly, such as:

```
supports os_family: 'linux'
```

... that way it won't attempt to run on Windows machines.

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+
+          describe etc_libuser do
+            its(['import','login_defs']){should eq ['/etc/login.defs']}
+          end
+         "
+
+    include ObjectTraverser
+    attr_reader :params, :raw_content
+
+    class UnknownEtcLibUserConf < StandardError; end
```
adamleff: Where is this used?

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+      @raw_content = load_raw_content(@conf_path)
+      @params = parse(@raw_content)
+    end
+
+    def parse(content)
+      @params = SimpleConfig.new(content, multiple_values: true).params
+    end
+
+    def to_s
+      'Libuser Configuration'
```
adamleff: Should we add the file name here in case a user test multiple libuser configs in the same profile and therefore would need to know which is which?

perhaps:

```ruby
def to_s
  "LibUser Config from #{@conf_path}"
end
```

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    def value(key)
+      # uses ObjectTraverser.extract_value to walk the hash looking for the key,
+      # which may be an Array of keys for a nested Hash.
+      extract_value(key, params)
+    end
+
+    def load_raw_content(opts)
+      # if the opts isn't a hash, we assume it's a path to a file
+      unless opts.is_a?(Hash)
```
adamleff, line 68: I'm unclear why this is needed. The resource is written such that only a filename parameter is expected. When would a user supply a Hash? Is this method just copy/pasta from another resource implementation? Looks like it may have been copied from the `json` resource and should probably get removed.

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+      end
+    end
+
+    def load_raw_from_file(path)
+      file = inspec.file(path)
+
+      # these are currently ResourceSkipped to maintain consistency with the resource
+      # pre-refactor (which used skip_resource). These should likely be changed to
+      # ResourceFailed during a major version bump.
+      raise Inspec::Exceptions::ResourceSkipped, "No such file: #{path}" unless file.file?
```
adamleff: These should probably change to `ResourceFailed` instead of skipped... looks like this may have been copied from another resource where we skip to maintain backward compatibility, but per #2418, we plan on changing these to failures soon. Since this is a new resource, this can fail right out of the gate.

lib/resources/etc_libuser.rb
```diff
@@ -0,0 +1,108 @@
...
+      raise Inspec::Exceptions::ResourceSkipped, "File #{path} is empty or is not readable by current user" if file.content.nil? || file.content.empty?
+
+      file.content
+    end
+
+    def load_raw_from_command(command)
```
adamleff, line 95: Again, this resource isn't written to expect a user to supply a command to run to get the content of a libuser.conf file, but rather only accept a path to a file. This method should not be necessary.

test/unit/resources/etc_libuser.rb
```diff
@@ -0,0 +1,48 @@
+# encoding: utf-8
+# author: Thomas Cate
```
adamleff: I don't think this is you... did Thomas Cate take part in the writing of these tests?

adamleff (changes requested):
Hi, @djhaynes, and thanks for your contribution!

This looks like a great start, however it looks like a bunch of unnecessary code got copied into this resource from another resource. I've left you comments accordingly, but we definitely need to get some cleanup done before we can review again. You also have some rubocop failures which is causing the Travis failures we see.

Additionally, this project uses the [Developer Certificate of Origin](https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) method for ensuring contributors have the legal authority to contribute. We'll need to you to either reword each of your commits to include the `Signed-off-by` line, or you can take this opportunity to squash them down to a single commit and add the `Signed-off-by` line then. But we will not be able to accept your contribution until it's signed off accordingly.

Please let us know if you have any questions. Thanks!

djhaynes:
Hi, @adamleff , thanks for the quick reply and all the feedback! Sorry, for all the errors, I did not see CONTRIBUTING.md before I issued the pull request. I will go through all your individual feedback and make the necessary updates. 

As far as the code that I copied from the JSON resource, it was because of issue #1300 which prevented me from making my resource a subclass of the existing JSON resource. Is there a better way to handle this while the issue is still open?

Thanks,

Danny

adamleff:
Hi, @djhaynes!

Core resources can definitely subclass other core resources. It's custom resources that currently have an issue. See the xml, csv, and ini resources as an example.

However, upon review of this resource, I'm not sure this is the right call for this resource. Additionally, the libuser.conf looks a lot like an INI-style file. Is there any reason we can't use the already-existing `ini` resource to parse this file?

~Adam

lib/resources/etc_libuser.rb
djhaynes: The reason that I added it was because the man page for /etc/libuser.conf (https://linux.die.net/man/5/libuser.conf) states:

"Files
/etc/libuser.conf The default location of the configuration file. Can be overridden by the LIBUSER_CONF environment variable, except in set-uid or set-gid programs."

djhaynes:
Hi, @adamleff !

I did not use the INI resource because it uses SimpleConfig with default parameters which includes "multiple_values: false". The man page for /etc/libuser.conf (https://linux.die.net/man/5/libuser.conf) states that settings can have multiple values. Specifically, it says: 

"A variable can have more than one value, specified by using more than one line defining that variable. All currently defined variables accept only the first value and ignore the others, if any."

adamleff:
@djhaynes thanks for the explanation!

So, based on that excerpt from the man page, while variables can have multiple definitions, it seems that only the first is accepted and any other assignment is ignored. Is that your read of it, as well?

If so, technically `multiple_values: false` is the right call, but perhaps we need a signal that only the first should be taken (because SimpleConfig takes the last).

So I'm wondering if the right path for this resource is:

 - add an option to SimpleConfig, such as `keep_first_value_only: true` which defaults to `false`
 - subclass the existing ini resource
 - override the `parse` method to use SimpleConfig with our new keep-the-first-value flag

I don't know that it makes sense to put the environment variable detection into the resource... detecting that feels like out-of-scope for the resource itself and rather should be put back onto the consumer of the resource to provide a path to a libuser.conf file they want InSpec to look at.

djhaynes:
@adamleff no problem!

Yes, that was how I interpreted it as well. However, I wasn't sure if there was a use case when an administrator might want to check that there were not any unwanted values even if not actually used by libuser. I also thought since we were taking the time to create a resource for this configuration file, we should support the ability to check all the values.

As far as how to proceed, your approach seems reasonable. It would also allow me to get rid of that code from the JSON resource assuming that I won't run into that bug again.

I can go either way with having the environment variable detection and I am willing to defer to your preferences since you are the InSpec expert :) and I am new to all this. I was thinking that the environment variable lookup would make it easier for profile control authors. If we do decided to take it out, we should probably add documentation somewhere to make it explicit for the profile control authors in case they are looking at the man page too and see that text.

test/unit/resources/etc_libuser.rb
djhaynes: No it is not. This should be fixed now.

adamleff:
> However, I wasn't sure if there was a use case when an administrator might want to check that there were not any unwanted values even if not actually used by libuser.

That's an interesting concern. However, we want to make sure the resource is an accurate representation of how the program that's consuming the configuration file would interpret it. Otherwise, there's a disconnect between InSpec and reality. For example, we wouldn't want to test an nginx configuration file differently than how nginx itself would interpret it.

> I was thinking that the environment variable lookup would make it easier for profile control authors.

I'd recommend removing it for now and we can see if we get strong feedback from the community of users that this is necessary. My concern is that InSpec, when testing remotely, does not acquire a PTY and does not evaluation "profile" files. So if a user has this set in their .bash_profile for example, the `os_env` resource won't see it and they'll get unwanted results. When dealing with compliance scanning, explicit behavior is always preferred over implicit behavior.

lib/resources/etc_libuser.rb
djhaynes: Yeah, I directly copied this code from the JSON resource too. Would you like me to remove it now? Or, should I just wait until you determine the best approach for moving forward (e.g. updating SimpleConfig and subclassing INI). I am happy to do whatever you think is best. Thanks!

lib/resources/etc_libuser.rb
djhaynes: Yeah, I think this is a good idea. I will update the code now.

lib/resources/etc_libuser.rb
djhaynes: Natively, it should run on Linux systems and I am happy to make this change. However, I wrote the resource such that it just checked for the config path rather than operating system because it would allow you to evaluate the configuration file on any system (since it does not require any OS-specific commands). This may not be something you want to support, but, I know it was a use case for some vendors when I worked on the OVAL project. Just let me know either way.

adamleff: OK, that sounds reasonable for now, though I'm not sure it makes sense to me. :)  If we don't want to limit it to certain operating system families, I'd definitely like to see the logic to try and pull from the environment variable removed so this focuses on either the sane default, or whatever path the user supplies.

lib/resources/etc_libuser.rb
adamleff: I believe subclassing the `ini` resource and modifying `SimpleConfig` is the right approach. So much of this resource code, including this method, will likely need to be removed. :)

djhaynes:
Hi @adamleff !

I still seem to be running into the "uninitialized constant" error when I try to subclass from the INI resource. I get the error when I use the following combinations suggested from (https://github.com/chef/inspec/issues/1300).


Attempt-1: 
&nbsp;&nbsp;  class EtcLibUser < IniConfig
Error-1:
&nbsp;&nbsp;  bundler: failed to load command: bin/inspec (bin/inspec)
&nbsp;&nbsp;  NameError: uninitialized constant IniConfig

Attempt-2: 
&nbsp;&nbsp;  class EtcLibUser < Inspec::Resources::IniConfig
Error-2:
&nbsp;&nbsp;  bundler: failed to load command: bin/inspec (bin/inspec)
&nbsp;&nbsp;  NameError: uninitialized constant Inspec::Resources::IniConfig

Attempt-3: 
&nbsp;&nbsp;  class EtcLibUser < ::Inspec::Resources::IniConfig
Error-3:
&nbsp;&nbsp;  bundler: failed to load command: bin/inspec (bin/inspec)
&nbsp;&nbsp;  NameError: uninitialized constant Inspec::Resources::IniConfig

Attempt-4:  
&nbsp;&nbsp;  module Inspec::Resources
&nbsp;&nbsp;  class EtcLibUser < IniConfig
Error-4:
&nbsp;&nbsp;  bundler: failed to load command: bin/inspec (bin/inspec)
&nbsp;&nbsp;  NameError: uninitialized constant Inspec::Resources::IniConfig

Is there something that I am doing incorrectly :)? 

Also, I am running InSpec Version 1.5.1 on OpenSuse 42.3.

Thanks!

-Danny

adamleff:
@djhaynes I'm not 100% sure why you're running into that error. Perhaps it's a local dev environment / bundler issue?

Here's what I just did:

 * created a new file at `lib/resources/leff_test.rb`
 * updated the file with the following content:

```ruby
module Inspec::Resources
  class LeffTest < IniConfig
    name 'leff_test'
    desc 'A test resource.'
    example 'an example'

    private

    def resource_base_name
      'LeffTest'
    end
  end
end
```

 * Added an entry to `lib/inspec/resource.rb` to load it in: `require 'resources/leff_test`
 * Went into the InSpec Shell (`bundle exec inspec shell`) and tried to use the resource:

```
inspec> leff_test('/tmp/foo.txt')
=> LeffTest /tmp/foo.txt
```

... as you can see, it worked as expected.

djhaynes:
Thanks @adamleff !

It did not work for me, but, good to know the issue isn't with InSpec! 

I will look more closely at my development environment to see if I can track down the issue.

Will you be updating SimpleConfig to take that new option?

Thanks again!

adamleff:
@djhaynes no, I hadn't planned on it. I hope you can attempt to tackle that in your PR! If not, feel free to log an issue in the repo asking for a feature request (and referring to this issue for context).

djhaynes:
Okay, not a problem! I can check it out after I resolve my development environment issues.

Thanks,

Danny

jquick:
Hey @djhaynes ! Thanks for working on this. Do you need additional help with your env issues?

djhaynes:
Hi @jquick! 

Yes, do you have any thoughts on how/where I would go about starting the investigation of my environment issues? I am fairly new to InSpec and Ruby and I am not really sure where to begin. 

Thanks!

-Danny

jquick:
Hey @djhaynes - I think I see the issue here. Since your resource is required before the IniConfig, its not defined yet. Try putting:

`require 'resources/ini'`

At the top of your resource. To make sure it gets loaded before you subclass it.

clintoncwolfe:
Hi @djhaynes - are you still actively working on this?

djhaynes:
At the moment, I am not actively working on this. At some point, I would like to get back to this, but, if you are interested in working on it, let me know.

clintoncwolfe:
For the time being, let's close this, and if you wish to revisit it, just comment or re-open.  Thanks!

PR #2487, opened 2018-01-29
Update reporter with breaking inspec 2.0 changes.

jquick:
This change includes all reporter breaking changes for the 2.0 release. 

Fixes https://github.com/chef/inspec/issues/2290
- Time fields are now rfc3339 for json/json-min

Fixes https://github.com/chef/inspec/issues/1716
- Controls section has been removed from json

Fixes https://github.com/chef/inspec/issues/1825
- Unique exit codes have been added

DOCS will be coming in https://github.com/chef/inspec/issues/2546 for these changes.

This also includes removing the old support:inspec version deprecation.


adamleff:
This is conditionally approved by me. :) I'll add official removal once the reporter work is merged and this PR is updated to point toward the release-2.0 branch instead.

I think we should clean up the multiple layers of abstraction needed to bubble up the exit code. This isn't the fault of THIS PR, it's always been a bit weird. I don't think we need to go through 4-5 methods to get to the exit code and expect that something upstream is going to properly call `exit`. 

Personally, I think the CLI classes should be responsible for exiting with the right exit code, not the Runner class... since the only time the exit code matters is the CLI. Would you mind taking a few minutes to think about how we might clean that up? I think we might be able to just add a method on the Runner called `exit_code` and have it fetch, and then the CLI class can ask for it.

jquick:
The cli exit issue was addressed here #2464.

clintoncwolfe (commented):
Code looks sensible, and should get the intended effects.  I did run the functional tests, and got an unrelated failure - is it a regression, or a broken window?
```
Finished in 113.519541s, 1.3390 runs/s, 5.2766 assertions/s.

  1) Failure:
example inheritance profile#test_0002_can vendor profile dependencies from the profile path [/Users/cwolfe/sandbox/inspec/mergo/inspec/test/functional/inspec_vendor_test.rb:28]:
Expected "Dependencies for profile /private/var/folders/bd/vrncq_yd2zd9r_kkq87nm66r0000gn/T/d20180209-79752-13xbma9/examples/inheritance successfully vendored to /private/var/folders/bd/vrncq_yd2zd9r_kkq87nm66r0000gn/T/d20180209-79752-13xbma9/examples/inheritance/vendor\n" to include "Dependencies for profile /var/folders/bd/vrncq_yd2zd9r_kkq87nm66r0000gn/T/d20180209-79752-13xbma9/examples/inheritance successfully vendored to /var/folders/bd/vrncq_yd2zd9r_kkq87nm66r0000gn/T/d20180209-79752-13xbma9/examples/inheritance/vendor"
```

adamleff:
@clintoncwolfe that's a macOS-specific error in the functional tests because `/var -> /private/var`. That test needs to be updated to be a bit more permissive in its matching.

aaronlippold:
For the labeling of ‘critical’, ‘major’, ‘minor’ - we will need a way to ‘pass in’ a new category matrix. i.e. in the IC/MIL space - 1.0 would be ‘critical’, 0.0 is N/A and High >= 0.7, Medium = 0.6 - 0.4, Low <= 0.3. 

I think we will find very quickly that ‘hard coding’ these will lead us to issues we could address fairly easily. 

We end up having to parse the json data, check controls for impact 0.0, etc and do a translation for different labeling systems between agencies and such. IC / DOD use a mix of `CAT1, CAT2,...` or `critical, high, med, low` etc. Gov/Fed ( non-military )  uses `high, med, low, na` etc. Let’s make it easy to support :slightly_smiling_face:

The use case is to support ‘Inherited Controls’  ( controls handled by someone else ) in dependent profiles the ‘top’ and ‘bottom’ of the `impact` are reserved for the logic cases of Critical and N/A controls and how we report and transform the data to support authorization decisions. 

My suggestion is - we add the ability to ‘set’ these in the inspec.yml file as something like:

```control_impact:
  - "critical" : [1.0]
  - "high": [0.7,0.9]
  - "medium" : [0.4,0.6]
  - "low" : [0.1,0.3]
  - "NA": [0.0]
```
We could also have a `defaults` setting which uses the base settings you have ‘hard coded’. If I just wanted to add in ‘NA’ as an example.

```control_impact:
  - NA: [0.0]
```

https://github.com/chef/inspec/pull/2487/ in the formatter/base.rb -> L97-L140 is what we would need to make a bit more dynamic.

aaronlippold:
As a follow-up. 

It would be useful for us to add a flag to the cli’s in general to say `--report=failed` or `--report=skipped` or `--report=high` or `--report=critical,major`  such that it would only report out the controls with the category in the given class. For the use case where Joe Blow is working on building the ‘hardening’ - it can make it easier to iterate on the work.

The default case is of course `--report=all`

lib/inspec/formatters/base.rb
```diff
@@ -84,6 +85,60 @@ def results
...
+    def statistics
+      failed = 0
+      skipped = 0
+      passed = 0
+      critical = 0
+      major = 0
+      minor = 0
+
+      all_unique_controls.each do |control|
+        next unless control[:results]
```
aaronlippold: Are we sure that 'skipped' controls still have a `results` array?

jquick:
@aaronlippold - Thanks for the comments here. These are great ideas but out of scope for this issue. I put these suggestions in a followup enhancement ticket. https://github.com/chef/inspec/issues/2615

lib/inspec/formatters/base.rb
```diff
@@ -84,6 +85,60 @@ def results
...
+    def all_unique_controls
+      unique_controls = Set.new
+      run_data[:profiles].each do |profile|
+        profile[:controls].map { |control| unique_controls.add(control) }
+      end
+
+      unique_controls
+    end
+
+    def statistics
```
chris-rock, line 97: I am a little bit worried of putting the stats into the reporter. Because this leads to additional problems/questions: stats per profile etc.

lib/inspec/formatters/base.rb
```diff
@@ -84,6 +85,60 @@ def results
...
+      passed = 0
+      critical = 0
+      major = 0
+      minor = 0
+
+      all_unique_controls.each do |control|
+        next unless control[:results]
+        if control[:results].any? { |r| r[:status] == 'failed' }
+          failed += 1
+          if control[:impact] >= 0.7
```
chris-rock: We need to be aware that this introduces a clear definition of what 0.7, 08 etc means. Before we always avoided that and I am not sure if we should rely on that as a core assumptions in inspec.

chris-rock:
I discussed this topic with @arlimus and we reached the following agreement:

- stats about control groups do not belong into inspec
- summary should be focused on `total`, `failed`, `skipped` & `passed` controls
- exit codes need to be adapted:
  - 0 run ok + all passed:  
  - 100 run okay + failures
  - 101 run okay + skipped only

lib/inspec/formatters/base.rb
jquick: Yeah, we need it here so the base runner knows what exit code to use. If we do it upstream then it would be hard to capture an accurate exit code for a rspec format. Removing the major/minor/critical will minimize this though.

lib/inspec/formatters/base.rb
```diff
@@ -23,6 +23,7 @@ def dump_summary(summary)
       run_data[:version] = Inspec::VERSION
       run_data[:statistics] = {
         duration: summary.duration,
+        controls: statistics,
```
chris-rock, line 26: Lets not add stats as default to our json reporter.

jquick: Ill remove it from the json report so it never shows up but leave this here for now for the exit code info.

chris-rock (approved):
Thank you @jquick

lib/inspec/formatters/base.rb
aaronlippold: @jquick @chris-rock what data is in statistics now that we have removed the grouping? If we are going to have the exit codes which use the statistics data - I think it is reasonable to have it. One stat we use is the 'timestamp' of the start of the run

PR #2489, opened 2018-01-29
Kg/apache

kagarmoe:
Signed-off-by: Kimberly Garmoe kgarmoe@chef.io

### DIFF
--- a/docs/resources/apache.md.erb
+++ b/docs/resources/apache.md.erb
@@ -1,0 +1,71 @@
+---
+title: About the apache Resource
+---
+
+# apache
+
+Use the `apache` InSpec audit resource to test the state of the Apache server on Linux/Unix systems. 
+
+<br>
+
+## Syntax
+
+An `apache` InSpec audit resource block declares settings that should be tested:
+
+    describe apache do
+      its('setting_name') { should cmp 'value' }
+    end
+
+where
+
+* `'setting_name'` is description of the Apache configuration file
+* `{ should cmp 'value' }` is the value that is expected
+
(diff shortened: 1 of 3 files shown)


adamleff:
Hey, @kagarmoe. So your example of `should exist` isn't a good one because this resource doesn't provide an existence property.

But now that I look at this resource, I personally think it's pointless. It doesn't actually do any inspection of the target node... it just sets up some data that could be completely wrong. I think this resource should actually be deleted outright in InSpec 2.0 and the conf_path stuff that `apache_conf` resource uses just baked into that resource.

@arlimus @chris-rock @jquick - thoughts? Perhaps it's too late to deprecate and remove in InSpec 2.0, but I think we should probably remove in 3.0.

adamleff:
If we decide to leave it (though I have strong feelings about core resources just setting up unverified data for use by other resources), I think we should leave this undocumented.

jquick:
Yeah I don't really see the use for it. I would say just pull in whatever we need into the apache_conf.

kagarmoe:
I don't have strong feelings about the resource itself, however, it is discoverable with in the inspec shell with `help resources`. Generally, user facing resources should be documented, even when they are decidedly mundane. I do have strong feelings about that.

Also, I've refactored this resource as a template for all of the others. Are we all OK with this format?

chris-rock:
I would say, lets merge with the apache conf, remove the separate file and deprecate the apache resource for inspec 3.0

adamleff:
Superseded by #2494

PR #2494, opened 2018-01-30
apache resource: document and deprecate

adamleff:
The `apache` resource only serves as a way to surface sane per-OS defaults for
a standard Apache install. Internally, this is only used by the `apache_conf`
resource and doesn't actually perform any inspection.

This PR adopts the documentation work in #2489, deprecates the `apache` resource,
and updates `apache_conf` to remove its reliance on the deprecated `apache` resource.

Supersedes #2489

### DIFF
--- a/docs/resources/apache.md.erb
+++ b/docs/resources/apache.md.erb
@@ -1,0 +1,71 @@
+---
+title: About the apache Resource
+---
+
+# apache
+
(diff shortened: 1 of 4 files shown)


kagarmoe (approved):

chris-rock (approved):
Thank you @adamleff

PR #2495, opened 2018-01-30
Changes 'has' to 'uses' in docs matcher section

kagarmoe:


kagarmoe:
Decided to do this a different way.

PR #2513, opened 2018-02-02
Refactors crontab docs & adds 'property examples'

kagarmoe:

### DIFF
--- a/docs/resources/crontab.md.erb
+++ b/docs/resources/crontab.md.erb
@@ -18,7 +18,7 @@ A `crontab` resource block declares a user (which defaults to the current user, 
 
 <br>
 
-## Examples
+## Resource Parameter Examples
 
 The following examples show how to use this InSpec audit resource.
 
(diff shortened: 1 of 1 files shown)


docs/resources/crontab.md.erb
```diff
@@ -65,6 +65,21 @@ The following examples show how to use this InSpec audit resource.
 
 <br>
 
+## Property Examples
+
+  ### Test a special time string (i.e., @yearly /root/anual_report.sh)
```
adamleff: I don't think this is worthy of a header. Also, the example below isn't a "yearly" but a "weekly" example.

Perhaps a note instead below the example:

> Note: InSpec will automatically interpret crontab-supported special time strings. For example, a crontab entry set to run `@yearly` can be tested as if the entry was manually configured to run on January 1, 12 AM.

adamleff (dismissed):

jquick (approved):

chris-rock (approved):
Thank you @kagarmoe

PR #2516, opened 2018-02-02
Updates directory resource

kagarmoe:

### DIFF
--- a/docs/resources/directory.md.erb
+++ b/docs/resources/directory.md.erb
@@ -13,7 +13,7 @@ Use the `directory` InSpec audit resource to test if the file type is a director
 A `directory` resource block declares the location of the directory to be tested, and then one (or more) matchers:
 
     describe directory('path') do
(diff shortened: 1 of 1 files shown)


docs/resources/directory.md.erb
```diff
@@ -13,7 +13,7 @@ Use the `directory` InSpec audit resource to test if the file type is a director
 A `directory` resource block declares the location of the directory to be tested, and then one (or more) matchers:
 
     describe directory('path') do
-      it { should MATCHER 'value' }
+      it { should cmp 'value' }
```
adamleff: There aren't any top-level matchers for the `directory` resource that return a directly-comparable value, so we should add a property into this example:

```ruby
its('property') { should cmp 'value' }
```

adamleff (dismissed):

jquick (approved):

chris-rock (approved):
Thank you @kagarmoe

PR #2517, opened 2018-02-02
Refactors docker_container resource

kagarmoe:

### DIFF
--- a/docs/resources/docker_container.md.erb
+++ b/docs/resources/docker_container.md.erb
@@ -49,31 +49,31 @@ The following examples show how to use this InSpec resource.
 
 ### id
 
-The `id` matcher tests the container id:
+The `id` property tests the container id:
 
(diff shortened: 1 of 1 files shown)


docs/resources/docker_container.md.erb
```diff
@@ -39,55 +43,57 @@ Alternatively, you can pass in the container id:
...
 
     its('command') { should eq 'nc -ll -p 1234 -e /bin/cat' }
+
+
+### Verify a running container:
+
+    describe docker_container('an-echo-server') do
+      it { should exist }
+      it { should be_running }
+      its('id') { should_not eq '' }
```
adamleff, line 86: Future PR can potentially change these to use `cmp` instead of `eq` if we're going to standardize on that. But this is technically correct as-is so I'm going to move it forward.

adamleff (approved):

PR #2520, opened 2018-02-02
http resource: Make remote worker the default

### LINKED ISSUE #2228: http resource: enable_remote_worker should be true by default in InSpec 2.0
The `http` resource now supports a remote worker via curl, but it is currently opt-in behavior. In InSpec 2.0, this should become the default behavior and the deprecation warnings removed.

jerryaldrichiii:
This makes it so the `http` resource will execute `curl` on the target machine by default.

This does not remove the functionality to make the HTTP call via Ruby on the machine executing InSpec.

This closes #2228

### DIFF
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -41,9 +41,6 @@ module Inspec::Resources
       # Run locally if InSpec is ran locally and remotely if ran remotely
       if inspec.local_transport?
         @worker = Worker::Local.new(http_method, url, opts)
-      elsif !inspec.command('curl').exist?
-        raise Inspec::Exceptions::ResourceSkipped,
-              'curl is not available on the target machine'
       else
         @worker = Worker::Remote.new(inspec, http_method, url, opts)
       end
@@ -150,6 +147,11 @@ module Inspec::Resources
         attr_reader :inspec
 
         def initialize(inspec, http_method, url, opts)
+          unless inspec.command('curl').exist?
+            raise Inspec::Exceptions::ResourceSkipped,
+                  'curl is not available on the target machine'
+          end
+
           @inspec = inspec
           super(http_method, url, opts)
         end
--- a/lib/resources/http.rb
+++ b/lib/resources/http.rb
@@ -29,7 +29,9 @@ module Inspec::Resources
       @opts = opts
 
       # Prior to InSpec 2.0 the HTTP test had to be instructed to run on the
-      # remote target machine. This warning will be removed.
+      # remote target machine. This warning will be removed after a few months
+      # to give users an opportunity to remove the unused option from their
+      # profiles.
       if opts.key?(:enable_remote_worker) && !inspec.local_transport?
         warn 'Ignoring `enable_remote_worker` option, the `http` resource ',
              'remote worker is enabled by default for remote targets and ',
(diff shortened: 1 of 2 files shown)


adamleff (changes requested):
@jerryaldrichiii, we don't give users the option with other resources whether to force a resource to execute locally vs. remotely, so we shouldn't do it here either. The flexibility to choose was done for backward compatibility only.

So, I think this PR should do the following:

 * remove support for the `enable_remote_worker` flag
 * emit a warning if they try to use it
 * run locally if local
 * run remotely if remote

jerryaldrichiii:
I agree @adamleff. Like you said offline, if local support is needed the user can leverage Ruby's `Net::HTTP`.

I removed support for `enable_remote_worker` in the latest commit. Let me know what you think.

lib/resources/http.rb
```diff
@@ -22,23 +22,28 @@ class Http < Inspec.resource(1)
...
     def initialize(url, opts = {})
       @url = url
       @opts = opts
 
-      if use_remote_worker?
-        return skip_resource 'curl is not available on the target machine' unless inspec.command('curl').exist?
-        @worker = Worker::Remote.new(inspec, http_method, url, opts)
-      else
+      # Prior to InSpec 2.0 the HTTP test had to be instructed to run on the
+      # remote target machine. This warning will be removed.
```
adamleff: This is a bit confusing... when will it be removed? 🙂 

How about we update this comment to "This warning will be removed after a few months to give users an opportunity to remove the unused option from their profiles."

lib/resources/http.rb
```diff
@@ -22,23 +22,28 @@ class Http < Inspec.resource(1)
...
+      if opts.key?(:enable_remote_worker) && !inspec.local_transport?
+        warn 'Ignoring `enable_remote_worker` option, the `http` resource ',
+             'remote worker is enabled by default for remote targets and ',
+             'cannot be disabled'
+      end
+
+      # Run locally if InSpec is ran locally and remotely if ran remotely
+      if inspec.local_transport?
         @worker = Worker::Local.new(http_method, url, opts)
+      elsif !inspec.command('curl').exist?
```
adamleff: Code-flow-wise, it would be way cleaner to couple this check with the remote worker itself. I initially did not do that because we didn't have the ability to raise exceptions to skip resources and you could only really call `skip_resource` during initialize. But now, since we're here, we can clean this up and this `if` statement will read much cleaner to the new user.

I would change this whole conditional to look like this:

```ruby
if inspec.local_transport?
  @worker = Worker::Local.new(http_method, url, opts)
else
  @worker = Worker::Remote.new(inspec, http_method, url, opts)
end
```

And then move the `curl` check to the initialize of the `Worker::Remote` class. Make sense?

adamleff (changes requested):
Much better. Let's clean up the `curl` check now that we can raise exceptions since we're already in here mucking with this stuff.

lib/resources/http.rb
jerryaldrichiii: Yup, that is much cleaner! Thanks for the historical context. I will make it happen.

jquick (approved):

jquick:
@adamleff - You mind giving this another look?

adamleff (approved):
Looks good. Not sure why the travis/appveyor changes are on this, maybe a rebase is needed, or maybe they should be pulled out and put in a separate PR? Either way, this PR should be good to go now.

jerryaldrichiii:
I rebased and force pushed. Travis/AppVeyor commits are removed :smile:

chris-rock (approved):
Thank you @jerryaldrichiii

PR #2533, opened 2018-02-06
Fix travis-ci bundler issue

jquick:
Bundler is now included in ruby and set as a default gem. As such we need to remove the force uninstall bundler command (as you cannot remove a default gem). Leaving in the install command will confirm its always there.


chris-rock (approved):
Thank you @jquick

PR #2538, opened 2018-02-08
Prevent Slashes in profile names

miah:
This PR prevents slashes in existing profile names, and when they are created with `inspec init`.

Fixes https://github.com/chef/inspec/issues/1783


jquick (approved):

chris-rock (changes requested):
While this is great for prevention the slash, it compromises on the usability of the tool

I find it is very compelling to just call:
```
inspec init profile path/to/directory
```
In case we have a slash in it, we have two options:
- slug the profile name from the path
- use last directory as the name

miah:
@chris-rock Thats a valid concern. I'll take another look at this and make it work that way. Thanks!

miah:
Oof. I rebased off master and that commit history looks awful now. I can delete and create a new PR if needed.

This PR now allows you to initialize a PR like "init profile foo/bar/baz" which would place the profile in "foo/bar/baz" and name the profile "baz".

chris-rock:
I think the PR needs to be opened against master not release 2.0. A new PR may be the better solution.

miah:
Oh right it's pointing to the wrong bench because I opening this a while
ago. Will re-pr this.

miah:
Going to re-open this against master.

clintoncwolfe:
Superceded by #3175

PR #2545, opened 2018-02-08
Changes indented code blocks to code fences adds style

burtlo:
* Fixed a small typo with a single quoted string
* Added `$` prompt where I saw it was needed
* Changed some inspec shell code blocks to use `shell`. It's not right but `irb` is not currently supported in the gem. We can talk if you want to add an inspec shell lexer they aren't too hard to make.
* Ran a script to change over all the indented code blocks in the resources doc to become code fences with the ruby language specified.
* Added a stylesheet and took a stab at picking some colors that would look appealing on the site. I don't know if there were some guidelines that I should have been aware of or used.

## Example of final rendering

The view 'docs/reference/dsl_inspec'

image

An example of the poor rendering of inspec shell: 

image

## Color Commentary

will-smith-i-make-this-look-good-men-in-black

https://www.youtube.com/watch?v=AjPau5QYtYs

### DIFF
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -107,7 +107,7 @@ The following test shows how to audit machines running PostgreSQL to ensure that
 ```ruby
 control 'postgres-7' do
   impact 1.0
-  title 'Don't allow empty passwords'
+  title "Don't allow empty passwords"
   describe postgres_session('user', 'pass').query("SELECT * FROM pg_shadow WHERE passwd IS NULL;") do
    its('output') { should eq('') }
   end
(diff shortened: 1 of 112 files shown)


burtlo:
I can continue to rebase this content if this is interesting to people and they are interested in this work eventually becoming part of the site.

burtlo:
I'll rebase to keep this relevant.

burtlo:
I see that the site went live with highlight.js and it produces the same results from the two samples I provided. I guess the only argument against this is the 44.9 KB of JavaScript being sent down the pipe to the user the first time they visit the site. The evaluation and paint operation appear trivial. A JS error on the site has the ability to block the rendering.

PR #2552, opened 2018-02-08
Fixes docs for file example

burtlo:
* The version test should be a string
* There was a missing closing brace in the regex examples
* Some white-spaced got cleaned up
* I added the file name into the variable for the pg examples

### DIFF
--- a/docs/resources/file.md.erb
+++ b/docs/resources/file.md.erb
@@ -51,9 +51,7 @@ The `content` property tests if contents in the file match the value specified i
 
 The following complete example tests the `pg_hba.conf` file in PostgreSQL for MD5 requirements.  The tests look at all `host` and `local` settings in that file, and then compare the MD5 checksums against the values in the test:
 
-    hba_config_file = '/etc/postgresql/9.1/main/pg_hba.conf'
-
-    describe file(hba_config_file) do
+    describe file('/etc/postgresql/9.1/main/pg_hba.conf') do
       its('content') { should match(%r{local\s.*?all\s.*?all\s.*?md5}) }
       its('content') { should match(%r{host\s.*?all\s.*?all\s.*?127.0.0.1\/32\s.*?md5}) }
       its('content') { should match(%r{host\s.*?all\s.*?all\s.*?::1\/128\s.*?md5}) }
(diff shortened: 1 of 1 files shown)


burtlo:
Another issue, probably more with the matcher and InSpec, but I will mention it here for discussion. 

I noticed that when the file does not exist and you use the following matchers:

```ruby
describe file('that-does-not-exist') do
  it { should be_setgid }
  it { should be_sticky }
  it { should be_setuid }
end
```

The results are not very descriptive:

```
  File that-does-not-exist
      ∅  undefined method `>' for false:FalseClass
      ∅  undefined method `>' for false:FalseClass
      ∅  undefined method `>' for false:FalseClass
```

jquick (approved):

jquick:
@burtlo - Good points, we should probably use fail_resource if it does not exist with a better message.

docs/resources/file.md.erb
```diff
@@ -162,6 +164,8 @@ or:
 
 ### Test the contents of a file for MD5 requirements
 
+    hba_config_file = 'pg_hba.conf'
+
     describe file(hba_config_file) do
```
chris-rock: 'pg_hba.conf' should be inplace and include the full path. Its unlikely to work

chris-rock (changes requested):
Thank you for highlighting the issues @burtlo Lets improve the fix to give a real pg config path

jquick:
@burtlo - You mind updating this with the requested changes?

burtlo:
Sure, I'll go learn what the real path and then update the pull request.

burtlo:
Updated and rebased.

docs/resources/file.md.erb
```diff
@@ -51,10 +51,12 @@ The `content` property tests if contents in the file match the value specified i
 
 The following complete example tests the `pg_hba.conf` file in PostgreSQL for MD5 requirements.  The tests look at all `host` and `local` settings in that file, and then compare the MD5 checksums against the values in the test:
 
+    hba_config_file = '/etc/postgresql/9.1/main/pg_hba.conf'
```
jquick: Can we move the path into the describe block? I'm not a fan of showing this format and it would keep us uniform with the rest of this documentation.

jquick (changes requested):

docs/resources/file.md.erb
burtlo: Sure. Happy to oblige.

jquick (approved):
Thanks @burtlo !

chris-rock:
@burtlo I do not understand the matcher part. `sticky`, `setgid` and `setupid` do not return true and should not. Therefore the following tests do not make sense. Therefore it is not an inspec issue.
```
describe file('that-does-not-exist') do
  it { should be_setgid }
  it { should be_sticky }
  it { should be_setuid }
end
```

chris-rock (approved):
Thank you @burtlo for this improvement. It makes things more clear.

burtlo:
That's my goal in life. Eventually, I'll find my way through all this ...

PR #2554, opened 2018-02-08
Use the version of LicenseScout that comes with the Omnibus gem.

tduffield:
LicenseScout is being refactored. We have released a 1.x version that
is pinned within the Omnibus gem.

Signed-off-by: Tom Duffield <tom@chef.io>


chris-rock (approved):
Thank you @tduffield

adamleff (approved):

PR #2601, opened 2018-02-09
Fix runner report and json newline

jquick:
There were a few issues with the new reporter work:

1) The report function on the runner was not working properly with the new report types. This is used for the audit cookbook. This has been resolved.
2) json/json-min were sending a newline at the end of their report. This has been resolved.


jerryaldrichiii (approved):

miah (approved):

chris-rock (approved):
Thank you @jquick

clintoncwolfe (commented):
My only comment is that I wish there were clear tests reproducing the regression, and showing that the change fixes it.

jquick:
Tests were added to confirm the regressions do not happen again. We capture the report output without newlines for the json. Tests were also added around the new report output to confirm its working as expected.

PR #2606, opened 2018-02-11
Add caution to "mode": include leading zero in file mode specifications

atsaloli:


aaronlippold (changes requested):
Great first start. We need to make a few updates and don't forget to sign your commit or we will be blocked getting it through.

You should use active language, something like:

The `mode` property uses the OCTAL format to compare modes, so you must write your numeric mode specification starting with zero. See: https://en.wikipedia.org/wiki/Leading_zero#0_as_a_prefix.

For example, the mode specification of `644`, ("-rw-r--r--" ) should be written as:

`{ should cmp '0644' }`, as shown above.

If not, it will be taken as a the decimal '1024' in octal by Ruby, and your test will fail.

( Note to implementer:

Does it makes sense to also have this section?? We may also want to 'link' to the `is_sticky`, `is_guid` and `is_suid` properties here as well as it connects logically using a style like below )

If you want to validate the `sticky bit` or the `suid/guid` bits, you can add the needed octet as:
`{ should cmp '00644' }`
... ( suid example )
... ( guid example )

For the best readability, use the provided `matchers` - <link to sticky>,<be_suid> and <be_guid>

`{ should be_sticky }` 
and
`{ should be_suid }`
`{ should be_guid }`

atsaloli:
Thanks @aaronlippold !  Who is the "implementer" you addressed?

jquick (commented):
Thanks @atsaloli ! This is a great addition. Some small personal suggestions.

docs/resources/file.md.erb
```diff
@@ -87,6 +87,22 @@ The `mode` property tests if the mode assigned to the file matches the specified
 
     its('mode') { should cmp '0644' }
 
+You must write the numeric mode specification starting with zero, to tell Ruby
```
jquick: Would you mind changing `Ruby` to `InSpec`? We have a lot of non-tech people using these docs and it may be more clear.

docs/resources/file.md.erb
```diff
@@ -87,6 +87,22 @@ The `mode` property tests if the mode assigned to the file matches the specified
...
+
+For example, write:
+
+    { should cmp '0644' }
+ 
+If you write it as
+
+    { should cmp '644' }
+
+then Ruby will understand it as _decimal_ 644, which is octal 1024 or
```
jquick: Same as above.

atsaloli:
Great! I appreciate your attention to usability, @jquick 

I forgot to add the 

    Signed-off-by: Aleksey Tsalolikhin <aleksey@verticalsysadmin.com>

to the last commit, and I'm not sure how to fix that at this point.   There is no IP in it anyway.

jquick:
Hey @atsaloli you should be able to do:
```
git commit --amend -s
then save/write
git push -f
```
To sign it off.

docs/resources/file.md.erb
```diff
@@ -87,6 +87,22 @@ The `mode` property tests if the mode assigned to the file matches the specified
 
     its('mode') { should cmp '0644' }
 
+You must write the numeric mode specification starting with zero, to tell InSpec
+the value is [octal](https://en.wikipedia.org/wiki/Leading_zero#0_as_a_prefix).
```
kagarmoe: InSpec [octal](https://en.wikipedia.org/wiki/Leading_zero#0_as_a_prefix) values begin the numeric mode specification with zero.

docs/resources/file.md.erb
```diff
@@ -87,6 +87,22 @@ The `mode` property tests if the mode assigned to the file matches the specified
...
+
+For example, write:
+
+    { should cmp '0644' }
+ 
+If you write it as
```
kagarmoe: Without the zero prefix for the octal value, InSpec will interpret it as the _decimal_ value 644, which is octal 1024 or`-----w-r-T`, and any test for a file that is `-rw-r--r--` will fail.

kagarmoe (changes requested):

atsaloli:
@kagarmoe: your wording is much better, I like how direct and simple it is.

I've fixed the sign-off and the wording; and I've rebased.

aaronlippold:
The implementer is you, @jquick and @kagarmoe ... i.e. asking - do you guys think we should also address these realted issues here - guid, suid and stickly bit.

jquick:
@aaronlippold - I think that would be a great follow up pr from someone. For now I think the scope of this ticket is complete. Thanks for your work on this @atsaloli !

jquick (approved):

kagarmoe (approved):

atsaloli:
Thank you all!!  :)

PR #2607, opened 2018-02-11
Host resource: use bash over netcat in Linux

jvale:
Netcat's presence is widely considered a security issue, and thus not always available. This solution only uses timeout (from coreutils) and bash builtins, so less likely to require installing additional packages.

It also adds UDP support for free and moves a bit further in the direction of solving the concerns raised in https://github.com/chef/inspec/issues/1439.


adamleff (changes requested):
I think this is a great improvement, but I think we need to leave the netcat support in as a fallback. Not all Linux versions have a bash installed that was compiled with the `/dev/tcp` and `/dev/udp` support. For example, I believe the default Debian compile instructions for bash still include `--disable-net-redirections`

jvale:
Fair enough, will work on that.

jvale:
As suggest, brought back netcat. It is the preferred option, if available on the box. Also added support for UDP in Darwin since the code was pretty much already there.

lib/resources/host.rb
```diff
@@ -147,6 +147,69 @@ def missing_requirements(_protocol)
...
+
+      @has_nc = inspec.command('nc').exist?
+      @has_ncat = inspec.command('ncat').exist?
+      @has_net_redirections = inspec.command("strings `which bash` | grep -qE '/dev/(tcp|udp)/'").exit_status == 0
+    end
+
+    def missing_requirements(protocol)
+      missing = []
+
+      if %w{tcp udp}.include?(protocol) and !@has_nc and !@has_ncat
```
jquick: Can you change the `and`'s to `&&`? We try to avoid and/or due to the lower precedence.

lib/resources/host.rb
```diff
@@ -147,6 +147,69 @@ def missing_requirements(_protocol)
...
+          missing << 'netcat must be installed'
+        end
+      end
+
+      missing
+    end
+
+    def ping(hostname, port, protocol)
+      if %w{tcp udp}.include?(protocol)
+        if @has_nc or @has_ncat
```
jquick: Same as above, if you could change to `||`

jquick (commented):
Thanks @jvale ! This is a really nice addition to the host resource. Just some small suggestions.

lib/resources/host.rb
jvale: Had no idea about the precedence difference, guess I learned something. :) Done.

lib/resources/host.rb
jvale: ...and also done.

jquick (approved):
Thanks @jvale !

jquick:
@adamleff - You mind giving another pass on this one?

adamleff (approved):

PR #2610, opened 2018-02-12
Force a default reporter for ad-hoc runners

jquick:
There is a potential issue with ad-hoc runners if they do not set a format or reporter. If this happens no reporters run. This change forces a cli report in that case.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -93,7 +93,7 @@ module Inspec
       }
     end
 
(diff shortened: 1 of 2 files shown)


adamleff (approved):
This looks good. I'd also consider adding a test where if someone passes in `--format` (such as `--format documentation`) and no `--reporter` that we don't try and default to CLI in that case as well.

chris-rock (approved):
Thank you

PR #2616, opened 2018-02-12
Fix the /private/var osx issue causing functional tests to fail

jquick:
This change fixes the functional changes for osx.
screen shot 2018-02-12 at 9 39 47 am

### DIFF
--- a/test/functional/inspec_vendor_test.rb
+++ b/test/functional/inspec_vendor_test.rb
@@ -26,7 +26,7 @@ describe 'example inheritance profile' do
       out.stderr.must_equal ''
       out.exit_status.must_equal 0
       # this fixes the osx /var symlink to /private/var causing this test to fail
-      out.stdout.gsub!(%r{\/private\/var}, '/var')
+      out.stdout.gsub!('/private/var', '/var')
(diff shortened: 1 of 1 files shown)


jerryaldrichiii (approved):

test/functional/inspec_vendor_test.rb
```diff
@@ -25,6 +25,8 @@
       out = inspec('vendor --overwrite', "cd #{dir} &&")
       out.stderr.must_equal ''
       out.exit_status.must_equal 0
+      # this fixes the osx /var symlink to /private/var causing this test to fail
+      out.stdout.gsub!(%r{\/private\/var}, '/var')
```
adamleff: I don't even think you need to do this as a regex. `gsub` and `gsub!` accept a string match too.

```
[1] pry(main)> 'Something blah /private/var/dir_a/dir_b/private/blah more stuff'.gsub('/private/var', '/var')
=> "Something blah /var/dir_a/dir_b/private/blah more stuff"
```

jquick:
@adamleff - agreed, updated and tested.

adamleff (approved):

PR #2618, opened 2018-02-12
Add json fields to schema and add tests.

jquick:
There are a few fields that were getting included into the json results/controls output even though they were not listed in the schema. The fields are:

Message: simple error message when an exception happends
exception: exception error class if its not a traditional rspec error
backtrace: backtrace dump if its not a traditional rspec error

This PR adds them to the schema along with testing. Another talk we should have is if we actually want these fields in the json.


miah (approved):

chris-rock (approved):
Thank you @jquick

PR #2620, opened 2018-02-12
Skelatal resource: aws_s3_bucket_object

dromazmj:
* This new resource allows testing against an aws s3 bucket object.
* You can test the objects acls and detect if an object is public


docs/resources/aws_s3_bucket_object.md
```diff
@@ -0,0 +1,81 @@
+---
```
clintoncwolfe: Now that we're in InSpec Core, the doc files should be named `.md.erb`.

docs/resources/aws_s3_bucket_object.md
```diff
@@ -0,0 +1,81 @@
...
+title: About the aws_s3_bucket_object Resource
+---
+
+# aws_s3_bucket_object
+
+Use the `aws_s3_bucket_object` InSpec audit resource to test properties of a single AWS bucket object.
+
```
clintoncwolfe: Here, I think it might be worth explaining that each object has a 'key', which might be thought of as a path.

docs/resources/aws_s3_bucket_object.md
```diff
@@ -0,0 +1,81 @@
...
+This InSpec audit resource has the following special matchers. For a full list of available matchers (such as `exist`) please visit our [matchers page](https://www.inspec.io/docs/reference/matchers/).
+
+### be_public
+
+The `be_public` matcher tests if the object has potentially insecure access controls. This high-level matcher detects several insecure conditions, which may be enhanced in the future. Currently, the matcher reports an insecure object if any of the following conditions are met:
+
+  1. A object ACL grant exists for the 'AllUsers' group
+  2. A object ACL grant exists for the 'AuthenticatedUsers' group
+
+Note: This resource does not detect insecure object ACLs.
```
clintoncwolfe: This looks like copypasta from s3_bucket - in fact, this resource specifically DOES detect object ACLs.  You might mention that it doesn't detect _bucket_ ACLs.

docs/resources/aws_s3_bucket_object.md
```diff
@@ -0,0 +1,81 @@
+---
+title: About the aws_s3_bucket_object Resource
+---
+
+# aws_s3_bucket_object
```
clintoncwolfe: Now that we're in core InSpec, we need to escape all underscores that aren't quoted.  So here we need `aws\_s3\_bucket\_object`

docs/resources/aws_s3_bucket_object.md
```diff
@@ -0,0 +1,81 @@
...
+Use the `aws_s3_bucket_object` InSpec audit resource to test properties of a single AWS bucket object.
+
+
+<br>
+
+## Limitations
+
+S3 object security is a complex matter.  For details on how AWS evaluates requests for access, please see [the AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/how-s3-evaluates-access-control.html).  S3 buckets and the objects they contain support three different types of access control: bucket ACLs, bucket policies, and object ACLs.
+
+As of January 2018, this resource supports evaluating bucket object ACLs. In particular, users of the `be_public` matcher should carefully examine the conditions under which the matcher will detect an insecure bucket.  See the `be_public` section under the Matchers section below.
```
clintoncwolfe: I'd suggest not saying "bucket object ACLs" - say merely "object ACLs" for clarity.

lib/resources/aws/aws_s3_bucket_object.rb
```diff
@@ -0,0 +1,78 @@
...
+  include AwsSingularResourceMixin
+  attr_reader :bucket_name, :key
+
+  def to_s
+    # keep the format that aws uses.
+    "s3://#{@bucket_name}/#{@key}"
+  end
+
+  def object_acl
+    @object_acl ||= BackendFactory.create(inspec_runner).get_object_acl(bucket: bucket_name, key: key).grants
```
clintoncwolfe: You'll need to break this one-liner apart a bit, and wrap the get_object_acl call in a block:

```
return @object_acl if defined? @object_acl
catch_aws_errors do
  @object_acl = BackendFactory.create(inspec_runner).get_object_acl(bucket: bucket_name, key: key).grants
end
@object_acl
```

lib/resources/aws/aws_s3_bucket_object.rb
```diff
@@ -0,0 +1,78 @@
...
+    )
+    if validated_params.empty? or !validated_params.key?(:bucket_name) or !validated_params.key?(:key)
+      raise ArgumentError, 'You must provide a bucket_name and key to aws_s3_bucket_object.'
+    end
+    validated_params
+  end
+
+  def fetch_from_api
+    backend = BackendFactory.create(inspec_runner)
+
```
clintoncwolfe: Wrap this whole begin block in a call to catch_aws_errors

test/aws/default/build/s3.tf
```diff
@@ -132,3 +141,28 @@ resource "aws_s3_bucket_policy" "allow-02" {
...
 POLICY
 }
+
+#=================================================================#
+#                       S3 Bucket Object
+#=================================================================#
+resource "aws_s3_bucket_object" "inspec_logo_public" {
+  bucket = "${aws_s3_bucket.public_bucket_for_objects.id}"
+  key    = "inspec-logo-public"
+  source = "inspec-logo.png"
```
clintoncwolfe: :100:

test/unit/resources/aws_s3_bucket_object_test.rb
```diff
@@ -0,0 +1,192 @@
...
+#                               Recall
+#=============================================================================#
+
+class AwsS3BucketObjectRecallTest < Minitest::Test
+  def setup
+    AwsS3BucketObject::BackendFactory.select(AwsMSBOSB::Basic)
+  end
+
+  def test_searching
+    assert_equal(true, AwsS3BucketObject.new(bucket_name: 'Public Bucket', key: 'public_file.jpg').exists?)
```
clintoncwolfe: You can just say `assert(thing_you_expect_to_be_true)`

test/unit/resources/aws_s3_bucket_object_test.rb
```diff
@@ -0,0 +1,192 @@
...
+    assert_equal(false, AwsS3BucketObject.new(bucket_name: 'Public Bucket', key: 'NonExistingObject').exists?)
```
clintoncwolfe: And you can just do `refute(thing_you_expect_to_be_false)`

clintoncwolfe (changes requested):
Very good work!  Mainly docs changes, and then wrapping API calls in our new (crude) exception handler.

test/integration/aws/default/build/cloudtrail.tf
```diff
@@ -186,6 +186,9 @@ resource "aws_cloudtrail" "trail_1" {
   cloud_watch_logs_group_arn = "${aws_cloudwatch_log_group.trail_1_log_group.arn}"
   cloud_watch_logs_role_arn  = "${aws_iam_role.cloud_watch_logs_role.arn}"
   kms_key_id                 = "${aws_kms_key.trail_1_key.arn}"
+  provisioner "local-exec" {
```
clintoncwolfe: This looks like an unrelated change - could you remove it?

clintoncwolfe (changes requested):
Thanks - your last set of changes looks good, but I think a stray commit (adding a local delay in cloudtrail TF config) slipped in.  Passes unit and integration tests.  Great work!  Please remove the delay (if you think we need it, let's open an issue and discuss, and then add in a separate PR) and then we'll be all set!

test/integration/aws/default/build/cloudtrail.tf
dromazmj: Sorry about that! That is actually a fix I had to do in order to get the aws setup to work on my system.  It looks like it just needed to wait a little for other resources to finish before it builds that resource.

clintoncwolfe (approved):

jquick (approved):
Thanks @dromazmj !

PR #2621, opened 2018-02-12
Allow ad-hoc runners to use rspec formats.

jquick:
Currently ad-hoc runners cannot use rspec format as the reporter cleaner happens after the runner is initiated. This change re-orders the initialize for the runner, as well as changing symbols to strings for the rspec reporters. 

The thor option does some magic to make some string keys work as symbols. In a future issue we should really look into making everything symbols and unifying it.  


miah (approved):

chris-rock (approved):
Thank you @jquick

PR #2626, opened 2018-02-13
Kg/k module

kagarmoe:


docs/resources/kernel_module.md.erb
```diff
@@ -39,6 +40,12 @@ where
 
 The following examples show how to use this InSpec audit resource.
 
+### version
+
+The `version` property tests if the named module version is on the system:
+
+    its(:version) { should eq '3.2.2' }
```
jquick: Do we have other resources with symbol properties? I confirmed 'version' works as well.

docs/resources/kernel_module.md.erb
```diff
@@ -1,6 +1,7 @@
 ---
 title: About the kernel_module Resource
 platform: linux
+=======
```
jerryaldrichiii: Is this line needed? Could we remove it?

docs/resources/kernel_module.md.erb
```diff
@@ -40,11 +41,17 @@ where
 
 The following examples show how to use this InSpec audit resource.
 
+### version
+
+The `version` property tests if the named module version is on the system:
+
+    its('version') { should eq '3.2.2' }
+
 ### Test a modules 'version'
```
jerryaldrichiii: Not part of your change but could we change this to `module's`

docs/resources/kernel_module.md.erb
```diff
@@ -40,11 +41,17 @@ where
 
 The following examples show how to use this InSpec audit resource.
 
+### version
+
+The `version` property tests if the named module version is on the system:
```
jerryaldrichiii: I would reword this to something like:

```
The `version` property tests if the named module on the system has the correct version:
```

jerryaldrichiii (changes requested):
Looks great @kagarmoe just a few suggestions/questions.

docs/resources/kernel_module.md.erb
```diff
@@ -94,14 +100,21 @@ The following examples show how to use this InSpec audit resource.
...
+### be_disabled
 
-    its(:version) { should eq '3.2.2' }
+The `be_disabled` matcher tests if the module is disabled:
+
+    it { should be_disabled }
+
+### be_loaded
+
+The `be_loaded` matcher tests if the module is a loadable kernel module:
```
jerryaldrichiii: Could we change this to:

```
The `be_loaded` matcher tests if the named module is loaded:
```

jerryaldrichiii (changes requested):
Great job! Only one suggestion.

jerryaldrichiii (approved):
Looks great @kagarmoe thank you for all the iterations.

jquick (approved):
Thanks @kagarmoe !

PR #2634, opened 2018-02-13
Update `inspec detect` to support APIs/Families

### LINKED ISSUE #2611: InSpec detect platform support
The `inspec detect` command currently caters to only operating systems. This command's wording needs to be changed to "platforms". The code should also be moved from the os.rb to platform.rb resource. 

We should update so family is taken from train so it works for both api/os. Release could also be used for api versions if we set them in train. Arch should hide if not set?

We should also wrap the detect block to not blow up with a train stack trace like shell/exec have.

jerryaldrichiii:
This closes #2611 

This does the following to `inspec detect`:
  - Modifies it to use the `platform` resource
  - Changes the output to mention Platform and show the family hierarchy
  - Changes the JSON output by changing `family` to `families`
  - Adds better error messaging (no more stacktraces!)
  - Adds support for APIs such as AWS/Azure
  - Hides Arch from API platforms output (not applicable)

### DIFF
--- a/lib/resources/platform.rb
+++ b/lib/resources/platform.rb
@@ -25,6 +25,10 @@ module Inspec::Resources
       end
     end
 
+    def families
+      @platform.family_hierarchy
+    end
+
     def name
       @platform.name
     end
@@ -46,8 +50,19 @@ module Inspec::Resources
       @platform.family_hierarchy.include?(family)
     end
 
-    def families
-      @platform.family_hierarchy
+    def params
+      h = {
+        name: name,
+        families: families,
+      }
+
+      # Avoid adding Release/Arch for APIs
+      unless in_family?('api')
+        h[:release] = release
+        h[:arch] = arch
+      end
+
+      h
     end
 
     def supported?(supports)
--- a/lib/resources/platform.rb
+++ b/lib/resources/platform.rb
@@ -54,11 +54,11 @@ module Inspec::Resources
       h = {
         name: name,
         families: families,
+        release: release,
       }
 
-      # Avoid adding Release/Arch for APIs
+      # Avoid adding Arch for APIs (not applicable)
       unless in_family?('api')
-        h[:release] = release
         h[:arch] = arch
       end
 
--- a/test/functional/inspec_detect_test.rb
+++ b/test/functional/inspec_detect_test.rb
@@ -16,19 +16,7 @@ describe 'inspec detect' do
     stdout.must_include "\nRelease:   \e[0;36m"
   end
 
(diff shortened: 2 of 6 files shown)


miah (approved):

jquick (approved):
This looks good to me. Example output of the new detect:
screen shot 2018-02-14 at 8 31 24 am

lib/resources/platform.rb
```diff
@@ -46,8 +50,19 @@ def in_family?(family)
...
-      @platform.family_hierarchy
+    def params
+      h = {
+        name: name,
+        families: families,
+      }
+
+      # Avoid adding Release/Arch for APIs
+      unless in_family?('api')
+        h[:release] = release
```
chris-rock: According to https://github.com/chef/inspec/issues/1661 section 5. The platform resource release is not optional. Do we have a sdk version that we could use for release of an api?

chris-rock (commented):
Great improvement @jerryaldrichiii I added one comment.

chris-rock:
We may want to add a schema to verify that the output conforms to our standard.

test/functional/inspec_detect_test.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    stdout = res.stdout
+    stdout.must_include "\n== Platform Details\n\n"
+    stdout.must_include "\nName:      \e[0;36m"
+    stdout.must_include "\nFamilies:  \e[0;36m"
+
+    stdout.wont_include "\nArch:"
+    stdout.wont_include "\nRelease:"
+  end
+end
```
chris-rock: Can we add a json test for aws and azure?

lib/resources/platform.rb
jquick: I can throw in a train PR to add the sdk version as the release for the transports then it will be available here.

jerryaldrichiii: Cool beans, thanks @jquick. I'll move release out of the omit category.

test/functional/inspec_detect_test.rb
jerryaldrichiii: Sure can. Latest update has tests for API + JSON. Let me know what you think.

lib/resources/platform.rb
jquick: Fields added via https://github.com/chef/train/pull/258

jquick:
apis now shows release:
screen shot 2018-02-14 at 1 30 28 pm

chris-rock (approved):
Thank you @jerryaldrichiii

PR #2635, opened 2018-02-14
New Resource aws_config_recorder

dromazmj:
* Adds functionality to be able to test an AWS Configuration Recorder


docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
+---
```
clintoncwolfe: Docs files should now end in .md.erb

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+title: About the aws_config_recorder Resource
+---
+
+# aws_config_recorder
```
clintoncwolfe: Escape underscores in unquoted doc strings

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+
+Use the `aws_config_recorder` InSpec audit resource to test properties of your aws config service.
```
clintoncwolfe: Capitalize: "AWS Config service".  That helps make it clear we're talking about a specific offering from AWS.

clintoncwolfe: Might also be good to describe what a Config Recorder is.  What does it represent in the Config Service?  What does it interact with?

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+
+    describe aws_config_recorder(recorder_name: 'my-recorder') do
+      it { should be_recording }
+    end
+    
+## Properties
+
+### role_arn
+
+Provides the IAM role arn associated with the configuration recorder.
```
clintoncwolfe: A few words about what the role is used for might be good.

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+    describe aws_config_recorder(username: 'bob')
+      its('role_arn') { should eq 'arn:aws:iam::721741954427:role/My_Recorder' }
+    end
+    
+### resource_types
+
+Provides a list of AWS resource types for which the AWS Config records configuration will change.
+
+    describe aws_config_recorder(username: 'bob')
+      its('resource_types') { should eq ['AWS::EC2::CustomerGateway', 'AWS::EC2::EIP'] }
```
clintoncwolfe: Array equality is problematic - there are sorting issues, for example (see #2739 for some discussion on that).  In examples, I recommend using the `include` matcher and splitting the tests:

```
its('resource_types') { should include 'AWS::EC2::CustomerGateway' }
its('resource_types') { should include 'AWS::EC2::EIP' }
```

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+    end
+    
+<br>
+
+## Matchers
+
+### be_all_supported
+
+Provides whether AWS Config records configuration is changed for each type of regional resource.
```
clintoncwolfe: Ok, we need to rename this one for fluency, and also smooth out that description.  How about:

```
be_recording_all_resource_types
record_all_resource_types  # This would require a matcher_alias call
```
And for the prose:
"Indicates if the ConfigurationRecorder will record changes for all resources, regardless of type.  If this is true, resource_types is ignored."

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+
+Provides the IAM role arn associated with the configuration recorder.
+
+    describe aws_config_recorder(username: 'bob')
+      its('role_arn') { should eq 'arn:aws:iam::721741954427:role/My_Recorder' }
+    end
+    
+### resource_types
+
+Provides a list of AWS resource types for which the AWS Config records configuration will change.
```
clintoncwolfe: Might also add "Note that if `be_recording_all_resource_types` is true, this property is meaningless and may be an empty list."

docs/resources/aws_config_recorder.md
```diff
@@ -0,0 +1,68 @@
...
+
+### be_all_supported
+
+Provides whether AWS Config records configuration is changed for each type of regional resource.
+
+      it { should be_all_supported }
+      
+### have_include_global_resource_types
+
+Provides whether or not the AWS Config includes support for all global resource types.
```
clintoncwolfe: "Indicates whether the ConfigurationRecorder will record changes for global resource types (such as IAM Users)."

I think we should rename this:
```
be_recording_all_global_types
record_all_global_types  # This would require a matcher_alias call
```
Better ideas welcome!

lib/resources/aws/aws_config_recorder.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def has_include_global_resource_types?
+    @include_global_resource_types
+  end
+
+  def status
+    return unless @exists
+    backend = BackendFactory.create(inspec_runner)
+    @resp = backend.describe_configuration_recorder_status(@query)
```
clintoncwolfe: Wrap in catch_aws_errors

lib/resources/aws/aws_config_recorder.rb
```diff
@@ -0,0 +1,91 @@
...
+    @status = @resp.configuration_recorders_status.first.to_h
```
clintoncwolfe: I'm always leery at accepting the first result from a possibly unsorted array.  How do we know it's the right one?

lib/resources/aws/aws_config_recorder.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    validated_params
+  end
+
+  def fetch_from_api
+    backend = BackendFactory.create(inspec_runner)
+    @query = { configuration_recorder_names: [@recorder_name] }
+
+    @resp = backend.describe_configuration_recorders(@query)
```
clintoncwolfe: wrap in catch_aws_errors

lib/resources/aws/aws_config_recorder.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  class Backend
+    class AwsClientApi < AwsBackendBase
+      BackendFactory.set_default_backend(self)
+      self.aws_client_class = Aws::ConfigService::Client
+
+      def describe_configuration_recorders(query)
+        aws_service_client.describe_configuration_recorders(query)
+      rescue Aws::ConfigService::Errors::NoSuchConfigurationRecorderException
```
clintoncwolfe: Don't catch exceptions here - catch them in fetch_from_api.  It's part of the contract of the AWS API that it will throw that exception.  If you catch it here, that means we're not testing for exception handling elsewhere.

test/aws/default/verify/controls/aws_config_recorder.rb
```diff
@@ -0,0 +1,52 @@
+fixtures = {}
+[
+  'role_for_config_recorder_arn',
+  'config_recorder_name',
+].each do |fixture_name|
+  fixtures[fixture_name] = attribute(
+    fixture_name,
+    default: "default.#{fixture_name}",
+    description: 'See ../build/iam.tf',
```
clintoncwolfe: copypasta - config.tf

test/unit/resources/aws_config_recorder_test.rb
```diff
@@ -0,0 +1,153 @@
...
+              all_supported: false,
+              include_global_resource_types: false,
+              resource_types: ['AWS::EC2::CustomerGateway', 'AWS::EC2::EIP']
+            }),
+          ]
+        }),
+        'empty' => {}
+      }
+      return recorders[query[:configuration_recorder_names][0]] unless recorders[query[:configuration_recorder_names][0]].nil?
+      recorders['empty']
```
clintoncwolfe, line 132: You should raise `Aws::ConfigService::Errors::NoSuchConfigurationRecorderException` here

clintoncwolfe (changes requested):
Very good! A few matchers to rename and doc edits, and some changes in exception handling.   Your tests are very thorough - great work, this will be very useful!

lib/resources/aws/aws_config_recorder.rb
dromazmj: The first one will always work because you are specifying the recording name.  So then it will always return an array of one element

clintoncwolfe (approved):

jquick (approved):
This looks good to me! Thanks @dromazmj !

Issue #2633, opened 2018-02-13
Normalize internal implementation of 3 AWS resources


clintoncwolfe:
### Description

Three AWS resources, aws_iam_password_policy, aws_ec2_instance, and aws_iam_root_user were implemented early in the life of the project, using dependency injection for testing and 'resource' style AWS API calls.  Those choices are valid in isolation, but all other AWS resources use a standardized approach using the AwsSingularResourceMixin or AwsPluralResourceMixin, which provides a variety of validation, testing, and (with #2613) exception handling facilities, allowing the resource code to be brief and distinct to the needs of the resource (and the AWS API that backs it).

Until the three resources are refactored, we must duplicate (or omit) any such shared functionality in the resources.

### InSpec and Platform Version

InSpec 2.0 pre-release, branch release-2.0, 469369fa

PR #2636, opened 2018-02-14
Catch cloud exceptions and document connection info

### LINKED ISSUE #2612: Azure/Aws connection documentation
Documentation: It is not easy to get started. We need documentation about how to setup (e.g. get the access token) and we need to support the default aws/azure client configuration to make the experience seamless. (e.g.  ~/.aws/credentials and ~/.aws/config)

### LINKED ISSUE #2613: Azure/Aws resources should not hard fail on API exceptions
A profile needs to be executable, but a control may fail. We currently throw exceptions for resources that stop full profile execution, resulting in no report. This will lead to issues if we expect json output e.g in Chef Automate. The controls needs to fail (ie, have a failure outcome but absorb any exceptions) if the resource cannot be executed, but we need to generate a report.

We should hook into the fail_resource if the resource connection is bad saying such.

clintoncwolfe:
Adds a document describing the credentials setup for AWS and Azure, fixes #2612 

Adds a method, `catch_aws_errors`, to the AwsResourceMixin.  Intercepts connection exceptions and provides a friendlier error message.  Intercepts all other AWS errors and calls `fail_resource`, allowing the inspec run to continue while marking the test/control failed.  Added wrapper around lazy-fetched properties, and copy-pasted the code to get three more resources covered (#2633 logged as tech debt).  

Adds a method, `catch_azure_errors`, to AzureResourceBase.  Catches Azure exceptions (there appears to only be one class; they are otherwise distinguished by a JSON structure).  Due to how properties are dynamically generated, it was necessary to add an instance variable as a sentinel for whether the resource had failed, and then treat any unknown property as nil, which is ugly. 

The AWS and Azure changes together fix #2613.

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -4,7 +4,7 @@ New in InSpec 2.0, you may now use certain InSpec resources to audit properties 
 
 In the initial release of InSpec 2.0, support for selected AWS and Azure resources is included.
 
-## AWS Platform Support in InSpec 2.0
+## AWS Platform Support
 
 ### Setting up AWS credentials for InSpec
 
@@ -18,17 +18,40 @@ InSpec uses the standard AWS authentication mechanisms.  Typically, you will cre
 
 You may provide the credentials to InSpec by setting the following environment variables: `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_KEY_ID`.  You may also use `AWS_PROFILE`, or if you are using MFA, `AWS_SESSION_TOKEN`.  See the [AWS Command Line Interface Docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html) for details.
 
+Once you have your environment variables set, you can verify your credentials by running:
+
+```bash
+you$ inspec detect -t aws:// 
+
+== Platform Details
+Name:         aws
+Families:     cloud, api
+```
+
 #### Using the InSpec target option to provide credentials
 
-Look for a file in your home directory named `~/.aws/credentials`.  If it does not exist, create it.  Add your credentials as a new profile, in INI format:
+Look for a file in your home directory named `~/.aws/credentials`.  If it does not exist, create it.  Choose a name for your profile; here, we're using the name 'auditing'.  Add your credentials as a new profile, in INI format:
 
 ```
-[my-profile-name]
+[auditing]
 aws_access_key_id = AKIA....
 aws_secret_access_key = 1234....abcd
 ```
 
 You may now run InSpec using the `--target` / `-t` option, using the format `-t aws://region/profile`.  For example, to connect to the Ohio region using a profile named 'auditing', use `-t aws://us-east-2/auditing`.
+
+To verify your credentials, 
+```bash
+you$ inspec detect -t aws:// 
+
+== Platform Details
+Name:         aws
+Families:     cloud, api
+```
+
+#### Verifying your credentials
+
+To verify your credentials
 
 ## Azure Platform Support in InSpec 2.0
 
@@ -46,7 +69,7 @@ The information from the SPN can be specified either in a file `~/.azure/credent
 
 #### Using the Azure Credentials File
 
-The simplest way is to create the file `~/.azure/credentials` with the following format. InSpec is configured to look for this file by default, so no settings are required.
+By default InSpec is configured to look at ~/.azure/credentials, and it should contain:
 
 ```
 [<SUBSCRIPTION_ID>]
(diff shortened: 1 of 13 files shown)


lib/resources/aws/aws_iam_user.rb
```diff
@@ -82,6 +82,8 @@ def fetch_from_api
 
     # TODO: consider returning Inspec AwsIamAccessKey objects
     @access_keys = backend.list_access_keys(user_name: username).access_key_metadata
```
jquick, line 84: does this need wrapped in catch_aws_errors?

jquick (approved):
Looking good! Thanks for this @clintoncwolfe

docs/platforms.md
```diff
@@ -0,0 +1,88 @@
...
+
+Look for a file in your home directory named `~/.aws/credentials`.  If it does not exist, create it.  Add your credentials as a new profile, in INI format:
+
+```
+[my-profile-name]
+aws_access_key_id = AKIA....
+aws_secret_access_key = 1234....abcd
+```
+
+You may now run InSpec using the `--target` / `-t` option, using the format `-t aws://region/profile`.  For example, to connect to the Ohio region using a profile named 'auditing', use `-t aws://us-east-2/auditing`.
```
chris-rock, line 41: Are you referring to the profile from the `.aws/credentials` or is that a different one. If so, we may use the same name for the profile. I'd like the `auditing` profile name.

docs/platforms.md
```diff
@@ -0,0 +1,88 @@
+# Using InSpec 2.0 with Platforms
+
+New in InSpec 2.0, you may now use certain InSpec resources to audit properties of things that aren't individual machines - for example, an Amazon Web Services S3 bucket.
+
+In the initial release of InSpec 2.0, support for selected AWS and Azure resources is included.
+
+## AWS Platform Support in InSpec 2.0
```
chris-rock: I think we do not need to name InSpec 2.0 here again

docs/platforms.md
```diff
@@ -0,0 +1,88 @@
...
+Look for a file in your home directory named `~/.aws/credentials`.  If it does not exist, create it.  Add your credentials as a new profile, in INI format:
+
+```
+[my-profile-name]
+aws_access_key_id = AKIA....
+aws_secret_access_key = 1234....abcd
+```
+
+You may now run InSpec using the `--target` / `-t` option, using the format `-t aws://region/profile`.  For example, to connect to the Ohio region using a profile named 'auditing', use `-t aws://us-east-2/auditing`.
+
```
chris-rock, line 42: Should we include the `inspec detect` as an example to verify that the credentials and the connection works?

chris-rock (approved):
@clintoncwolfe Great additions. I added some questions regarding the documentation. We can fix those in follow-up PRs

docs/platforms.md
```diff
@@ -0,0 +1,88 @@
...
+The simplest way is to create the file `~/.azure/credentials` with the following format. InSpec is configured to look for this file by default, so no settings are required.
+
+```
+[<SUBSCRIPTION_ID>]
+client_id = "<CLIENT_ID>"
+client_secret = "<CLIENT_SECRET>"
+tenant_id = "<TENANT_ID>"
+```
+
+So to run InSpec now it is as simple as running:
```
miah: This could read better like "With the credentials are in place you may now execute InSpec"

docs/platforms.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Azure CLI: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authenticate-service-principal-cli
+- PowerShell: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authenticate-service-principal
+- Azure Portal: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-service-principal-portal
+
+The information from the SPN can be specified either in a file `~/.azure/credentials`, as environment variables, or using InSpec target URIs.
+
+#### Using the Azure Credentials File
+
+The simplest way is to create the file `~/.azure/credentials` with the following format. InSpec is configured to look for this file by default, so no settings are required.
```
miah: "By default InSpec is configured to look at `~/.azure/credentials`, and it should contain:" or "By default InSpec is configured to look at `~/.azure/credentials`, and the format is:"

docs/platforms.md
clintoncwolfe: 👍

docs/platforms.md
clintoncwolfe: This may be misleading.  Train's AWS support doesn't test the connection until an API call is actually tried.

PR #2641, opened 2018-02-14
New Skeletal Resource aws_config_delivery_channel

dromazmj:
This PR adds functionality to test an aws delivery channel and its properties.


test/aws/default/build/iam.tf
```diff
@@ -99,4 +99,51 @@ resource "aws_iam_group" "administrators" {
 
 output "iam_group_administrators" {
   value = "${aws_iam_group.administrators.name}"
+}
+
+#======================================================#
+#                    IAM Roles
+#======================================================#
+
+resource "aws_iam_role" "role_for_config_delivery_channel" {
```
clintoncwolfe: I think I'd prefer to keep the supporting fixtures for Config in config.tf .  In other words, move the iam_role and iam_role_profile into config.tf.  The idea being, the Terraform files group together all of the pieces needs to test closely related InSpec resources.  So, it's not "what it is", but rather, "what it is used for"

test/aws/default/build/s3.tf
```diff
@@ -76,6 +76,16 @@ output "s3_bucket_access_logging_not_enabled_name" {
   value = "${aws_s3_bucket.acess_logging_not_enabled.id}"
 }
 
+resource "aws_s3_bucket" "bucket_for_delivery_channel" {
```
clintoncwolfe: Same as above - move config-related support fixtures to config.tf

test/aws/default/build/sns.tf
```diff
@@ -14,6 +14,14 @@ output "sns_topic_recall_hit_arn" {
   value = "${aws_sns_topic.sns_test_topic_1.arn}"
 }
 
+resource "aws_sns_topic" "sns_topic_for_delivery_log" {
```
clintoncwolfe: Same as above - move config-related support fixtures to config.tf

test/aws/default/verify/controls/aws_config_delivery_channel.rb
```diff
@@ -0,0 +1,47 @@
...
+].each do |fixture_name|
+  fixtures[fixture_name] = attribute(
+    fixture_name,
+    default: "default.#{fixture_name}",
+    description: 'See ../build/iam.tf',
+  )
+end
+
+#======================================================#
+#               aws_config_recorder - Singular
```
clintoncwolfe: Copypasta - this is the delivery channel, not the config recorder

test/aws/default/verify/controls/aws_config_delivery_channel.rb
```diff
@@ -0,0 +1,47 @@
...
+
+#-------------------  Properties -------------------#
+control "aws_config_delivery_channel properties" do
+  describe aws_config_delivery_channel(fixtures['delivery_channel_01']) do
+    its('s3_bucket_name') { should eq fixtures['s3_bucket_for_delivery_channel'] }
+    its('s3_key_prefix') { should eq nil }
+    its('sns_topic_arn') { should eq fixtures['sns_topic_for_delivery_log_arn'] }
+    its('delivery_frequency') { should eq 'TwentyFour_Hours' }
+  end
+end
```
clintoncwolfe: This is an excellent integration test!  Great work @dromazmj !

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
+---
+title: About the aws_config_delivery_channel Resource
+---
+
+# aws_config_delivery_channel
+
+Use the `aws_config_delivery_channel` InSpec audit resource to test properties of your aws config service.
```
clintoncwolfe: Capitalize AWS and Config.  That helps make it clearer that we're talking about the service offered by AWS known as Config.  Also, we need to clarify what this InSpec resource specifically is used for.... maybe:

The AWS Config service can monitor and record changes to your AWS resource configurations.  
Use the `aws_config_delivery_channel` InSpec audit resource to examine how the AWS Config service delivers those change notifications.

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Syntax
+
+An `aws_config_delivery_channel` resource block declares the tests for a single AWS configuration delivery channel.
+
+    describe aws_config_delivery_channel('my_channel') do
+      it { should exist }
+    end
+
+    describe aws_config_delivery_channel(recorder_name: 'my-channel') do
```
clintoncwolfe: Copypasta - channel_name, not recorder_name

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+      it { should exist }
+    end
+
+<br>
+
+## Examples
+
+The following examples show how to use this InSpec audit resource.
+
+### Test if the recorder is active and recording.
```
clintoncwolfe: This example is copypasta - this resource is a delivery channel, not a recorder.

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+
+    describe aws_config_delivery_channel(recorder_name: 'my-recorder') do
+      it { should be_recording }
+    end
+    
+## Properties
+    
+### s3_bucket_name
+
+Provides the name of the s3 bucket that the channel takes snaphots of and logs configuration history.
```
clintoncwolfe: The delivery channel *writes* configuration changes to the bucket.  The channel does not snapshot or read the bucket.

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+
+    describe aws_config_delivery_channel(channel_name: 'my_channel')
+      its('s3_bucket_name') { should eq 'my_bucket' }
+    end
+    
+### s3_key_prefix
+
+Provides the prefix of the s3 bucket.
```
clintoncwolfe: Might expand a bit here - 'Provides the s3 object key prefix (or "path") under which configuration data will be recorded.'

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Test if the recorder is active and recording.
+
+    describe aws_config_delivery_channel(recorder_name: 'my-recorder') do
+      it { should be_recording }
+    end
+    
+## Properties
+    
+### s3_bucket_name
```
clintoncwolfe, line 40: Might clarify that s3_bucket_name is an optional value - a delivery channel can talk to SNS, S3, or both.  That info could also go at the top of the doc, where we describe the resource.

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+    describe aws_config_delivery_channel(channel_name: 'my_channel')
+      its('s3_key_prefix') { should eq '' }
+    end
+    
+### sns_topic_arn
+
+Provides the ARN of the SNS topic for which the channel sends notifications about configuration changes.
+
+    describe aws_config_delivery_channel(channel_name: 'my_channel')
+      its('sns_topic_arn') { should eq '' }
```
clintoncwolfe: Provide a dummy ARN for the SNS topic.

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+    end
+    
+### delivery_frequency
+
+Provides how often the AWS Config sends snapshots to the s3 bucket in the delivery channel.
+
+Its values can be One_Hour, Three_Hours, Six_Hours, Twelve_Hours, TwentyFour_Hours
```
clintoncwolfe: I think this might be a case where we want to be better than the AWS API.  InSpec users might want to compare this ("ensure config snapshots are delivered at least as frequently as every 3 hours").  It's also an integer, this just seems silly.  

What if we parsed the value returned by AWS, and then offered up an integral valued property, `delivery_frequency_in_hours` ?

docs/resources/aws_config_delivery_channel.md
```diff
@@ -0,0 +1,77 @@
...
+    describe aws_config_delivery_channel(channel_name: 'my_channel')
+      its('delivery_frequency') { should eq '' }
+    end
+    
+    
+<br>
+
+## Matchers
+
+This resource contains no matchers.
```
clintoncwolfe: This resource provides no matchers, aside from the standard `exists` matcher.

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -0,0 +1,67 @@
+class AwsConfigurationDeliveryChannel < Inspec.resource(1)
+  name 'aws_config_delivery_channel'
+  desc 'Verifies settings for AWS Configuration Delivery Channel'
+  example "
+    describe aws_config_delivery_channel do
+      it { should exist }
+      it { should be_default }
```
clintoncwolfe: be_default? Copypasta?

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -0,0 +1,67 @@
+class AwsConfigurationDeliveryChannel < Inspec.resource(1)
```
clintoncwolfe: I'd suggest we rename this - the name of the AWS service is Config, not Configuration

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -0,0 +1,67 @@
...
+      raise ArgumentError, 'You must provide a channel_name to aws_config_delivery_channel'
+    end
+
+    validated_params
+  end
+
+  def fetch_from_api
+    backend = BackendFactory.create(inspec_runner)
+    query = { delivery_channel_names: [@channel_name] }
+    @resp = backend.describe_delivery_channels(query)
```
clintoncwolfe: FYI, per #2636 , we should now wrap every call we make to AWS in a block, like this:
```
catch_aws_errors do
  @resp = backend.describe_delivery_channels(query)
end
```
If your API call indicates a search miss by throwing an exception, as some do, you are still obliged to catch it.  catch_aws_errors is intended for more general things - bad creds, insufficient perms, etc.

clintoncwolfe (changes requested):
Very good work - what a great contribution!  You seem to have taken all of the changes as we moved into core in stride - very good.  I did find some copypasta, and a propery I'd like to throw some parsing onto, but otherwise, few changes needed.  👏

docs/resources/aws_config_delivery_channel.md
```diff
@@ -26,52 +29,51 @@ An `aws_config_delivery_channel` resource block declares the tests for a single
 
 The following examples show how to use this InSpec audit resource.
 
-### Test if the recorder is active and recording.
+### Test how frequent the channel writes configuration changes to the s3 bucket.
 
-    describe aws_config_delivery_channel(recorder_name: 'my-recorder') do
-      it { should be_recording }
+    describe aws_config_delivery_channel(channel_name: 'my-recorder') do
+      its(delivery_frequency) { should be > 3 }
```
clintoncwolfe: If the property has been changed to be `delivery_frequency_in_hours`, need to update it here too.

lib/resources/aws/aws_config_delivery_channel.rb
clintoncwolfe: Still need to wrap the API call in the error-catcher.

test/aws/default/verify/controls/aws_config_delivery_channel.rb
```diff
@@ -14,7 +14,7 @@
 end
 
 #======================================================#
-#               aws_config_recorder - Singular
+#         aws_config_delivery_channel - Plural
```
clintoncwolfe: Not a big deal, but this is a Singular resource, not a Plural.

clintoncwolfe (dismissed):
Almost there - just a couple of tweaks and we can bring it in.  Great work!

clintoncwolfe:
Looks good, unit tests are good, but integration tests fail - need to rename the bucket.

```
* aws_s3_bucket.bucket_for_delivery_channel: Error validating S3 bucket name: only lowercase alphanumeric characters and hyphens allowed in "inspec-bucket_for_delivery_channel-cwolfe-01.chef.io"
```

dromazmj:
@clintoncwolfe I'm having trouble reproducing this error.  Can you check that the bucket doesnt already exist before trying to setup aws

clintoncwolfe:
I saw the error when trying to create the fixtures (in `rake test:aws:setup:default`).  It's not a bucket name collision - AWS gives a different error for that.

AWS requires bucket names to be valid DNS names, which means they can't have things like underscores.    So I think changing the underscores to dashes [here](https://github.com/chef/inspec/pull/2641/files/de4552cc8ad672d5aabce7529cca0f4ca92453c9#diff-f2463f11eabdc193a0d2f875804c4999R80) will likely fix the issue.

dromazmj:
I made the changes, let me know if it still has problems for you.  Also I noticed that we changed the location of the aws integration tests in the master branch.  Do we need to change that in this PR as well?

clintoncwolfe:
Yes, moving the test files will make it easier to merge.  Thanks!

jquick (approved):
Looks good! Thanks @dromazmj !

test/integration/aws/default/build/config.tf
```diff
@@ -33,4 +33,101 @@ output "role_for_config_recorder_arn" {
...
+  name           = "delivery_channel_01"
+  s3_bucket_name = "${aws_s3_bucket.bucket_for_delivery_channel.bucket}"
+  depends_on     = ["aws_config_configuration_recorder.config_recorder_for_delivery_channel"]
+  sns_topic_arn  = "${aws_sns_topic.sns_topic_for_delivery_channel.arn}"
+  snapshot_delivery_properties = {
+    delivery_frequency = "TwentyFour_Hours"
+  }
+}
+
+resource "aws_config_configuration_recorder" "config_recorder_for_delivery_channel" {
```
clintoncwolfe: It looks like the default AWS account limit for config recorders is 1 - so when I run the integration tests, it fails here.  I think we'll have to re-use the one defined for the `aws_config_recorder` tests.

```
Error applying plan:

1 error(s) occurred:

* aws_config_configuration_recorder.config_recorder_for_delivery_channel: 1 error(s) occurred:

* aws_config_configuration_recorder.config_recorder_for_delivery_channel: Creating Configuration Recorder failed: MaxNumberOfConfigurationRecordersExceededException: Failed to put configuration recorder 'config_recorder_for_delivery_channel' because the maximum number of configuration recorders: 1 is reached.
	status code: 400, request id: 47d66458-2dea-11e8-a668-9734f608a028
```

clintoncwolfe: AWS has a limit of 1 config recorder, so this fails.  I've submitted a PR on your fork - https://github.com/dromazmj/inspec/pull/1 - to make the Delivery Channel tests re-use the config recorder created by the `aws_config_recorder` tests.  Once that's in, it otherwise passes, and looks great!

Here's the error:
```
Error applying plan:

1 error(s) occurred:

* aws_config_configuration_recorder.config_recorder_for_delivery_channel: 1 error(s) occurred:

* aws_config_configuration_recorder.config_recorder_for_delivery_channel: Creating Configuration Recorder failed: MaxNumberOfConfigurationRecordersExceededException: Failed to put configuration recorder 'config_recorder_for_delivery_channel' because the maximum number of configuration recorders: 1 is reached.
	status code: 400, request id: 47d66458-2dea-11e8-a668-9734f608a028
```

clintoncwolfe (changes requested):
One quick change to the integration tests - I sent a PR to your fork, so should be ready after that.  Look great!, and thanks!

clintoncwolfe (approved):

PR #2642, opened 2018-02-14
Update shadow resource to use FilterTable

miah:
This PR updates the shadow resource to use FilterTable rather than a custom filter implementation.
I've also increased test coverage =)

Fixes https://github.com/chef/inspec/issues/2265

### DIFF
--- a/lib/resources/shadow.rb
+++ b/lib/resources/shadow.rb
@@ -47,9 +47,9 @@ module Inspec::Resources
     filter = FilterTable.create
     filter.add_accessor(:where)
           .add_accessor(:entries)
-          .add(:user, field: 'user')
-          .add(:password, field: 'password')
-          .add(:last_change, field: 'last_change')
+          .add(:users, field: 'user')
+          .add(:passwords, field: 'password')
+          .add(:last_changes, field: 'last_change')
           .add(:min_days, field: 'min_days')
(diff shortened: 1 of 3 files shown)


jquick (approved):
Looks good! Thanks @miah

lib/resources/shadow.rb
```diff
@@ -32,87 +30,55 @@ class Shadow < Inspec.resource(1)
...
+
+    filter.add(:users) { |u, _| u.entries.map(&:user) }
+
+    filter.add(:count) { |i, _|
+      i.entries.length
+    }
+
+    filter.connect(self, :params)
+
+    alias users user
```
chris-rock: Should we add some deprecations here? Based on that improvement, I am not sure what you think users should use.

chris-rock (commented):
Huge improvements. The only part that is missing to me is a clear statement how the resource should be used. This would require to update the examples, docs and highlight deprecations for InSpec 3.0

lib/resources/shadow.rb
miah: Yes, I think we should deprecate the singular forms of user and password. I can update this PR with deprecation notices for those targeting 3.0. I'll update the docs too =)

chris-rock (approved):
Great improvement. Thank you @miah

PR #2645, opened 2018-02-15
Adds os metadata to 'z' resources

kagarmoe:


docs/resources/zfs_dataset.md.erb
```diff
@@ -1,5 +1,6 @@
 ---
 title: About the zfs_dataset Resource
+os: FreeBSD
```
chris-rock, line 3: zfs is not restricted to freebsd, it works on netbsd and macos too

PR #2646, opened 2018-02-15
Adds OS to header

kagarmoe:


docs/resources/yum.md.erb
```diff
@@ -1,5 +1,6 @@
 ---
 title: About the yum Resource
+os: RHEL
```
jquick, line 3: Will probably be a few more here, like amazon linux and some others. Ill get you a list

PR #2653, opened 2018-02-15
New Skeletal Resource aws_s3_buckets

dromazmj:
This PR adds a new resource for testing all or a group of s3 buckets


jquick (approved):
Thanks for this @dromazmj ! This one will also need rebased.

docs/resources/aws_s3_buckets.md
```diff
@@ -0,0 +1,49 @@
+---
+title: About the aws_s3_buckets Resource
+---
+
+# aws_s3_buckets
```
clintoncwolfe: We need to escape all underscores in titles, etc - anywhere they are unquoted.  Thanks!

docs/resources/aws_s3_buckets.md
```diff
@@ -0,0 +1,49 @@
...
+
+Use the `aws_s3_buckets` InSpec audit resource to list all buckets in a single account.
```
clintoncwolfe: You might also refer to the `aws_s3_bucket` resource to perform in-depth auditing of an individual bucket.

docs/resources/aws_s3_buckets.md
```diff
@@ -0,0 +1,49 @@
...
+The control will pass if the resource contains at least one bucket.
+
+    # Test if there are any buckets
+    describe aws_s3_buckets
+      it { should exist }
+    end
+
+## Properties
+
+### bucket_names
```
clintoncwolfe: :100: for using plural property names on a plural resource!

docs/resources/aws_s3_buckets.md
```diff
@@ -0,0 +1,49 @@
...
+
+Provides an array of strings containing the names of the buckets.
+
+    # Examine a specific subnets cidr_blocks
+    describe aws_s3_buckets do
+      its('bucket_names') { should eq ['my_bucket'] }
```
clintoncwolfe: Might also include an `include` test (which won't care about array length or order):

`its('bucket_names') { should include 'my_bucket' }`

lib/resources/aws/aws_s3_buckets.rb
```diff
@@ -0,0 +1,52 @@
+# author: Matthew Dromazos
+# author: Sam Cornwell
+class AwsS3Buckets < Inspec.resource(1)
+  name 'aws_s3_buckets'
+  desc 'Verifies settings for AWS S3 Buckets in bulk'
+  example "
+    describe aws_s3_bucket do
+      its('Bucket_names') { should eq ['my_bucket'] }
```
clintoncwolfe: That `Bucket_names` should probably be `bucket_names`.

lib/resources/aws/aws_s3_buckets.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  include AwsPluralResourceMixin
+
+  # Underlying FilterTable implementation.
+  filter = FilterTable.create
+  filter.add_accessor(:where)
+        .add_accessor(:entries)
+        .add(:exists?) { |x| !x.entries.empty? }
+        .add(:bucket_names, field: :name)
+        .add(:creation_dates, field: :creation_date)
```
clintoncwolfe: Looks like this isn't documented.  Unless that was intentional, it should be added to the docs.

test/aws/default/verify/controls/aws_s3_buckets.rb
```diff
@@ -0,0 +1,27 @@
...
+  # You should be able to test if there are any buckets
+  describe aws_s3_buckets do
+    it { should exist }
+  end
+end
+
+control "aws_s3_buckets properties" do
+  # you should be able to test the cidr_block of a subnet
+  describe aws_s3_buckets do
+    its('bucket_names') { should include fixtures['s3_bucket_public_name'], fixtures['s3_bucket_private_name'] }
```
clintoncwolfe: Could you add a negative test here?  Perhaps

`its('bucket_names') { should_not include 'i-dont-exist' }`

test/unit/resources/aws_s3_buckets_test.rb
```diff
@@ -0,0 +1,56 @@
...
+
+#=============================================================================#
+#                               Test Fixtures
+#=============================================================================#
+module AwsMSBB
+  class Basic < AwsBackendBase
+    def list_buckets
+      fixtures = [
+        OpenStruct.new({
+          name: "bucket_01",
```
clintoncwolfe: If you're going to include creation date, be sure to include it here.

test/unit/resources/aws_s3_buckets_test.rb
```diff
@@ -0,0 +1,56 @@
...
+#=============================================================================#
+module AwsMSBB
+  class Basic < AwsBackendBase
+    def list_buckets
+      fixtures = [
+        OpenStruct.new({
+          name: "bucket_01",          
+        }),
+        OpenStruct.new({
+          name: "bucket_02",
```
clintoncwolfe: Also - s3 bucket names must be DNS compliant - so you can use dashes, but not underscores.

test/unit/resources/aws_s3_buckets_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    assert(basic.bucket_names.include?('bucket_01'))
+    refute(basic.bucket_names.include?(nil))
+  end
+end
+
+#=============================================================================#
+#                               Test Fixtures
+#=============================================================================#
+module AwsMSBB
+  class Basic < AwsBackendBase
```
clintoncwolfe: You should also define a test fixture class that is empty, so we can test the negative side of `exist`s.

test/unit/resources/aws_s3_buckets_test.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def test_constructor_no_args_ok
+    AwsS3Buckets.new
+  end
+
+  def test_constructor_reject_unknown_resource_params
+    assert_raises(ArgumentError) { AwsS3Buckets.new(bla: 'blabla') }
+  end
+end
+
```
clintoncwolfe, line 22: Here is where you could add a 'Recall' section, where you test `exist` against an empty fixture and populated one.

clintoncwolfe (changes requested):
Great work again @dromazmj !  Just a few docs tweaks, and it would be great to have a negative unit test for `exist`.  Looking forward to pulling this in!

clintoncwolfe:
Unit and integration tests passed - thanks!

test/unit/resources/aws_s3_buckets_test.rb
dromazmj: Good call!

dromazmj:
@clintoncwolfe good catch on the creation date.  When I was making this resource I thought that would be a good property to have but for now I think its best to just get the resource through and if there is a use case for creation date I can add it.  Thanks!

clintoncwolfe (approved):
Great work!

PR #2654, opened 2018-02-15
Add `platform` tags to resource documentation and remove trailing whitespace

jerryaldrichiii:
This will allow for sorting of resources in the future.

This also removes trailing whitespace.

### DIFF
--- a/docs/resources/aide_conf.md.erb
+++ b/docs/resources/aide_conf.md.erb
@@ -1,5 +1,6 @@
 ---
 title: The aide_conf Resource
+platform: linux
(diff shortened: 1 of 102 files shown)


kagarmoe (approved):
Thanks!

jquick (approved):

jquick:
A few notes here:
1) These platform info are in the header metadata and will not be displayed on the docs.
2) The `linux` tag here is really referring to "anything that is not windows, so all unix based systems.
 - this will be enhanced later to be more defined.

PR #2665, opened 2018-02-16
Prevent resources from loading if supports check fails

### LINKED ISSUE #2662: API supports not working correctly
### Description

When using aws on azure you get errors along with azure on aws. The resource supports should be preventing this and skipping tests accordingly.

jquick:
This change prevents resources from initializing if we are not on a supported platform. This also fixes the issue where resource_fail/skip were not working correctly when using a where clause on a resource.

Fixes #2662 

### DIFF
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -43,8 +43,8 @@ module Inspec
       Inspec::Resource.registry
     end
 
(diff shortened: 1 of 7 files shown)


chris-rock (approved):
Looks good to me

clintoncwolfe (approved):
Looks good - I see where the init is avoided if the supports check failed

PR #2667, opened 2018-02-16
Fix legacy reporter output to file

### LINKED ISSUE #2656: Support legacy format output
### Description

The old inspec reporter supported passing in `:output` option to the inspec runner. This would use the rspec output code to write to a file. With the new reporter work we need to update the code here:
https://github.com/chef/inspec/blob/master/lib/inspec/base_cli.rb#L101

To also look for `:output` and if its set then set the file path. This will use the new file writer accordingly. 

This needs to go into the 1-stable and master branches

jquick:
This fixes the issue with test-kitchen when trying to output the results to a file. That use to use the old rspec file writer but now it set to use the new format. 

Fixes #2656 

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -91,11 +91,18 @@ module Inspec
       }
     end
 
-    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize
+    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       # merge in any legacy formats as reporter
       # this method will only be used for ad-hoc runners
       if !opts['format'].nil? && opts['reporter'].nil?
         warn '[DEPRECATED] The option --format is being is being deprecated and will be removed in inspec 3.0. Please use --reporter'
+
+        # see if we are using the legacy output to write to files
+        if opts['output']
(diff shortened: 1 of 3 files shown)


miah (approved):

lib/inspec/base_cli.rb
```diff
@@ -91,11 +91,18 @@ def self.default_options
...
-    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize
+    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       # merge in any legacy formats as reporter
       # this method will only be used for ad-hoc runners
       if !opts['format'].nil? && opts['reporter'].nil?
         warn '[DEPRECATED] The option --format is being is being deprecated and will be removed in inspec 3.0. Please use --reporter'
+
+        # see if we are using the legacy output to write to files
+        if opts['output']
+          opts['format'] = "#{opts['format']}:#{opts['output']}"
```
chris-rock: I think we should print that this is deprecated and will be removed in inspec 3.0

chris-rock (approved):
Thank you @jquick

lib/inspec/base_cli.rb
jquick: Added:
screen shot 2018-02-17 at 3 06 54 pm

chris-rock (approved):
Thank you @jquick

PR #2669, opened 2018-02-16
Move `TESTING_AGAINST_AWS.md` to `test/aws`

jerryaldrichiii:


miah (approved):

jquick:
I think this should go in docs somewhere and exposed

jerryaldrichiii:
@jquick I was thinking since this is very specific to our integration tests that it should live there.

I can move it though in case others want to replicate how we test internally. Any suggestions?

jquick:
Yeah good point, @kagarmoe / @clintoncwolfe id like your views on this also. Im not sure where it should live to get exposure.

clintoncwolfe (changes requested):
I'm OK with the file moving, so long as we make a link in the README in the Developing / Contributing section (I think we have a mention, but no link, now).  In terms of website docs, I'm not sure it's needed - this is pretty "inside baseball" material, and anyone looking to contribute would be reading the README on github anyway.

jerryaldrichiii:
@clintoncwolfe I added a link. Let me know if that works for you 😄

clintoncwolfe (approved):

jquick (approved):
Thanks @jerryaldrichiii !

PR #2671, opened 2018-02-16
Remove duplicated encryption key test.

juliandunn:
You are testing the same thing twice as this already exists down on line 23


jquick (approved):
Thanks for this!

PR #2674, opened 2018-02-16
Add correct `supports platform` to resources.

miah:
The PR updates all the resources to specify their supported platforms.

Fixes https://github.com/chef/inspec/issues/2661

### DIFF
--- a/lib/resources/aide_conf.rb
+++ b/lib/resources/aide_conf.rb
@@ -1,11 +1,11 @@
 # encoding: utf-8
-# author: Jen Burns, burnsjennifere@gmail.com
 
 require 'utils/filter'
 require 'utils/parser'
 module Inspec::Resources
   class AideConf < Inspec.resource(1)
     name 'aide_conf'
+    supports platform: 'unix'
     desc 'Use the aide_conf InSpec audit resource to test the rules established for
       the file integrity tool AIDE. Controlled by the aide.conf file typically at /etc/aide.conf.'
     example "
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -7,9 +7,8 @@ require 'utils/find_files'
 module Inspec::Resources
   class ApacheConf < Inspec.resource(1)
     name 'apache_conf'
-    supports platform: 'unix'
-    supports os_family: 'linux'
-    supports os_family: 'debian'
+    supports platform: 'linux'
+    supports platform: 'debian'
     desc 'Use the apache_conf InSpec audit resource to test the configuration settings for Apache. This file is typically located under /etc/apache2 on the Debian and Ubuntu platforms and under /etc/httpd on the Fedora, CentOS, Red Hat Enterprise Linux, and Arch Linux platforms. The configuration settings may vary significantly from platform to platform.'
     example "
       describe apache_conf do
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -1,7 +1,5 @@
 # encoding: utf-8
 # copyright: 2015, Vulcano Security GmbH
-# author: Dominik Richter
-# author: Christoph Hartmann
 
 require 'utils/simpleconfig'
 require 'utils/find_files'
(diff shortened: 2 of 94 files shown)


lib/resources/apache_conf.rb
```diff
@@ -1,14 +1,13 @@
...
-# author: Dominik Richter
-# author: Christoph Hartmann
 
 require 'utils/simpleconfig'
 require 'utils/find_files'
 
 module Inspec::Resources
   class ApacheConf < Inspec.resource(1)
     name 'apache_conf'
+    supports platform: 'unix'
```
jquick: Since we already have os_family here defined lets update it to be:
```
supports platform: 'linux'
supports platform: 'debian'
```
instead of unix, and remove the os_family's

lib/resources/filesystem.rb
```diff
@@ -1,6 +1,7 @@
 module Inspec::Resources
   class FileSystemResource < Inspec.resource(1)
     name 'filesystem'
+    supports platform: 'unix'
```
jquick: same here change this to linux and remove the os_family one

jquick (changes requested):
Thanks @miah , This is looking really nice!

lib/resources/apache_conf.rb
miah: alright

lib/resources/filesystem.rb
miah: got it

jquick (approved):
I think this is a solid platform to start with supports. We can move more restricted later but this will keep them from erroring when used with api's.

lib/resources/csv.rb
```diff
@@ -1,13 +1,13 @@
...
-# author: Christoph Hartmann
-# author: Dominik Richter
 
 # Parses a csv document
 # This implementation was inspired by a blog post
 # @see http://technicalpickles.com/posts/parsing-csv-with-ruby
 module Inspec::Resources
   class CsvConfig < JsonConfig
     name 'csv'
+    supports platform: 'unix'
```
chris-rock, line 9: We need to be careful, since unix and windows is not covering all operating systems. Even we have esx and cisco not being part of unix right now. While thinking about it, we may should fix that in our train config...

lib/resources/iptables.rb
```diff
@@ -24,6 +22,7 @@
 module Inspec::Resources
   class IpTables < Inspec.resource(1)
     name 'iptables'
+    supports platform: 'unix'
```
chris-rock: This would need to be linux

lib/resources/aide_conf.rb
```diff
@@ -1,11 +1,11 @@
 # encoding: utf-8
-# author: Jen Burns, burnsjennifere@gmail.com
```
chris-rock, line 2: Have we agreed somewhere that we remove the author headers?

chris-rock (commented):
I think it is okay to add supports for `unix` + `windows` for now, but we need to follow up and add the real support to the resource and remove the checks from initialize.

lib/resources/aide_conf.rb
jquick: Yes, github accurately keeps history of authors. This is not the main focus of this PR but is being included for now.

chris-rock: Lets try to keep a PR focused on one thin in the future

miah: No problem, our Team Working Agreement states that we should remove these when the file is touched so thats what I did.

lib/resources/iptables.rb
miah: Fixed

lib/resources/csv.rb
jquick: For now lets add:
`platform: 'esx'`
`platform: 'cisco'`

To csv, json, and yaml. Once the OS changes go out here https://github.com/chef/train/pull/261 we can change them to 'os'. But I don't want to force out a detect change this close to 2.0 release.

jquick: I've added this to the PR.

jquick:
@chris-rock - I think this is in a pretty safe spot now with the new changes.

chris-rock (approved):
@miah @jquick This is good for a first pass. As a next iteration, we need to focus on:
- update train to reflect os
- clean up the supports logic for resources and move that out of the initialize method
- some resources do not work on unix but only on linux, we need to be more accurate. It works right now, since we have the supports logic still in initialize

PR #2679, opened 2018-02-17
Updates README to use function instead of alias

burtlo:
I really liked the inclusion of the alias for the docker run version
of installing inspec. The instructions tried to make it clear that
when you create the alias that it was for only your current directory
but it sounded more like your current working directory wherever you
were so I thought that it would be better to actually make that work.

By changing it to a function you now can delay the processing of the
$(pwd) subshell so that it will mount your current working directory
when you run it. This makes it really easy to use the InSpec CLI
and the installation this way is on par with the other ways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,20 +84,19 @@ gem install inspec
 
 ### Usage via Docker
 
-Download the image and define an alias for convenience:
+Download the image and define a function for convenience:
 
 ```
 docker pull chef/inspec
-alias inspec='docker run -it --rm -v $(pwd):/share chef/inspec'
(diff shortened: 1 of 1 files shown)


jquick (commented):
This is a pretty nice improvement. What do you think about doing a one line function so its easier to just slap into your terminal instead of having multi line render?
`function inspec { docker run -it --rm -v $(pwd):/share chef/inspec $@; }`

burtlo:
I also like one-liners.
> I recently decided to sell my vacuum cleaner as all it was doing was gathering dust.

jquick (approved):
Thanks @burtlo !

jerryaldrichiii (approved):

PR #2684, opened 2018-02-17
Revert download button

hannah-radish:
I don’t agree with pushing the download button down for the new features - mainly because it’s very unlikely that most users will click the “show all new features” button and now we have hid a primary CTA under secondary content


arlimus (approved):
lgtm, don't hide it away

PR #2685, opened 2018-02-18
quote password when generating mysql command string

tolland:
If a mysql password contains shell characters, login fails...
```
# mysql -uroot -p$@be|11d{>4s6D> -h localhost  -s -e "show databases;"
-su: 11d{: command not found
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
```

with fix;

```
# mysql -uroot -p'$@be|11d{>4s6D>' -h localhost  -s -e "show databases;"
mysql: [Warning] Using a password on the command line interface can be insecure.
Database
information_schema
mysql
performance_schema
sys
```


burtlo:
Would a password with a single-quote cause this solution to break?

tolland:
Yes, it seems it would by a quick test that single quotes need escaping as well. The obvious escaping doesn't seem to work;

    mysql -uroot -p'abc12*%HUHUIH3\'' -h localhost
    > ^C

This does;

```
# mysql -uroot -p$'abc12*%HUHUIH3\'' -h localhost
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
```

as does;

```
# mysql -uroot -p'abc12*%HUHUIH3'"'"'' -h localhost
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
```

tolland:
or to keep the password out of the command args;

```
export MYSQL_PWD=$(cat <<'EOMYSQL'
abc12*%HUHUIH3'
EOMYSQL
)
$ mysql -uroot -h localhost
Welcome to the MySQL monitor.  Commands end with ; or \g.
```

burtlo:
These all sound like good tests.

I see this unit test file for that resource. https://github.com/chef/inspec/blob/master/test/unit/resources/mssql_session_test.rb but there are no tests that assert the command built and executed.

There is an example of a more complex test specified for the Postgres session resource https://github.com/chef/inspec/blob/master/test/unit/resources/postgres_session_test.rb

tolland:
That latter solution would also be preferred because html format output current includes the password from the mysql_session params;
screenshot_2018-02-18_21-10-37

adamleff (changes requested):
Can we use `Shellwords.shellescape` instead? We should probably wrap both the user and the password in a `shellescape` call which should do all the work for us and handle single-quotes-in-the-password as well.

tolland:
Shellwords.shellescape and some tests were added

jquick:
@adamleff - You mind giving this another pass?

adamleff (approved):
Great work, @tolland!

jquick (commented):
Thanks @tolland !

PR #2689, opened 2018-02-18
Fix legacy reporter output to file (inspec-v1.x)

jquick:
This fixes the issue with test-kitchen when trying to output the results to a file. That use to use the old rspec file writer but now it set to use the new format.

This is the same PR as https://github.com/chef/inspec/pull/2667 but is going to the 1.x branch.

Signed-off-by: Jared Quick jquick@chef.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: required
 branches:
   only:
     - master
+    - 1-stable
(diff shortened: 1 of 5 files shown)


chris-rock (approved):
Thank you @jquick

PR #2694, opened 2018-02-19
Fixes the deprecation warning text for report and output.

burtlo:
Remove an extra 'is being' from the 'is being is being'


jquick:
@burtlo - Thanks for this! can you also update the test here?:
https://github.com/chef/inspec/blob/master/test/unit/base_cli_test.rb#L124

jquick (approved):
Thanks @burtlo !

PR #2707, opened 2018-02-20
Always make sure the copyright footer has the correct year

juliandunn:


jquick (approved):
Thanks for this!

www/source/layouts/_footer.slim
```diff
@@ -45,4 +45,4 @@ footer#main-footer.dark-bg
                 img src="/images/chef-logo.png"
 
   p.strict-center.pad-under-xs
-    small.t-gray &copy; Chef Software 2017
+    small.t-gray &copy; Chef Software #{ Time.now.strftime("%Y") }
```
chris-rock: Should this be © 2015-2018 ?

juliandunn: that's probably safer, yeah... I'll change it

miah (approved):

PR #2708, opened 2018-02-20
Fix infrastructure code example and typo in parse_config_file reference

juliandunn:
On the website I think you meant to use `parse_config_file` in the example. I also found a typo in the help text while looking this up.

### DIFF
--- a/www/source/index.html.slim
+++ b/www/source/index.html.slim
@@ -243,8 +243,8 @@ header.bg-gradient.margin-top-offset.short-bg.relative
                 code   its('mode') { should cmp 0644 }
                 code end
                 br
-                code describe parse_config_file('/etc/myapp.conf') do
-                code   its('port') { should cmp 8080 }
+                code describe apache_conf do
+                code   its('Listen') { should cmp 8080 }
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @juliandunn !

www/source/index.html.slim
```diff
@@ -243,7 +243,7 @@ header.bg-gradient.margin-top-offset.short-bg.relative
                 code   its('mode') { should cmp 0644 }
                 code end
                 br
-                code describe myapp.conf do
+                code describe parse_config_file('/etc/myapp.conf') do
```
chris-rock: That implies that we can load arbitrary config files. While I agree that myapp.conf may not be the best, I would love to have a resource that fits better. Maybe `nginx` resource?

juliandunn: sure, I can do that. the trouble is that `nginx_conf` will result in a lengthy example, so I will use an `apache_conf` example instead.

jerryaldrichiii (approved):
Looks great! Love the more applicable example!

PR #2710, opened 2018-02-20
Fixes the paths for assets loading for javascript and stylesheets

burtlo:
When viewing a non-index page like tutorials I was getting a lot of errors:

image

* The paths were relative and that was causing problems on non-index pages.
* jquery apparently needs to be loaded before the menumaker.

I can't get the demo to generate those stylesheets/javascript so I can't
verify if the `script/inspec_tutorial.js` and `dist/inspect_tutorial.js` are the
same/similar from the web as one is minified and the other is not.

### DIFF
--- a/www/source/layouts/layout.slim
+++ b/www/source/layouts/layout.slim
@@ -10,8 +10,8 @@ html
 
     title = current_page.data.title || "InSpec"
 
-    link rel='stylesheet' type='text/css' href='css/inspec_tutorial.css'
-    link rel='stylesheet' type='text/css' href='stylesheets/vendor/fontawesome.min.css'
+    link rel='stylesheet' type='text/css' href='/css/inspec_tutorial.css'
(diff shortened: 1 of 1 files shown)


burtlo:
Whitespace also tends to get cleaned up around me.

jquick (approved):
This seems fine to me and the rendered netlife looks correct.

jquick:
@hannah-radish - You mind giving this a once over?

burtlo:
Both of the tutorial JS files added here: https://github.com/chef/inspec/commit/959509fd70dceb0c29e4eceb0911ba1a53e91ca6. 

There isn't much explanation around them, but I assume it was intentional. If both should exist, then should the non-minified script file also receive the minification treatment?

hannah-radish (approved):
@burtlo Thanks so much for taking a look at this and fixing it! NICE WORK!! and I appreciate it ❤️

PR #2712, opened 2018-02-20
Update shell detect to work with platforms

jquick:
The shell detect intro was using its own os logic here. This pr's updates it to use the same logic as the cli detect command. This now works for platforms accordingly.
screen shot 2018-02-20 at 6 48 17 am

Fixes https://github.com/chef/inspec/issues/2711
Fixes https://github.com/chef/inspec/issues/2687

### DIFF
--- a/lib/resources/platform.rb
+++ b/lib/resources/platform.rb
@@ -65,6 +65,23 @@ module Inspec::Resources
       h
     end
 
+    def detect(indent: 0, color: 39)
+      str = ''
+      params.each { |item, info|
+        data = info
+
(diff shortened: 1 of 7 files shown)


lib/resources/platform.rb
```diff
@@ -65,6 +65,23 @@ def params
       h
     end
 
+    def detect(indent: 0, color: 39)
+      str = ''
```
chris-rock: Should that be private? I am not sure if we want to expose this via our resource

chris-rock (approved):
Very nice improvement

PR #2721, opened 2018-02-21
Reword `it` block in `inspec check` tests to match actual test

jerryaldrichiii:


miah (approved):

jquick (approved):
Thanks @jerryaldrichiii

PR #2292, opened 2017-11-07
nginx resource: support quoted identifiers

adamleff:
An nginx config may contain configuration settings that are quoted, such as a map entry:

```
"~^\/opcache-api" 1;
```

The `nginx_conf` resource was failing to properly parse these.

Fixes #2278

PR #2726, opened 2018-02-22
nginx_conf resource: Fix include paths with quotes

jerryaldrichiii:
This modifies `nginx_conf` to strip surrounding quotation marks before attempting to expand the file paths.

Without this, the values from the NGINX config get passed to `File.expand_path` and have undesirable results. 

Example:
```
[1] pry(main)> good = "/foo/bar/baz"
=> "/foo/bar/baz"
[2] pry(main)> bad = '"/foo/bar/baz"'
=> "\"/foo/bar/baz\""
[3] pry(main)> File.expand_path(good)
=> "/foo/bar/baz"
[4] pry(main)> File.expand_path(bad)
=> "/Users/jerry/Projects/github.com/chef/inspec/\"/foo/bar/baz\""
```


adamleff:
While this is one way to address the problem, I'm not sure this is the correct approach. Quotes can be used in an nginx config to wrap values that contain spaces, but the quotes aren't actually part of the value unless they are escaped. If we're seeing quotes in a value and the quotes are just wrapping the value, the parser probably needs to be modified to account for that. We've had to address this at least once before - see #2292. We probably need to do something similar for values too, not just identifiers.

jerryaldrichiii:
@adamleff I moved the logic to remove quotes to `NginxParser`. Thank you for the suggestion and your prior work done with `identifiers`.

adamleff (approved):
This looks solid, @jerryaldrichiii. I know the PEG structure can be daunting, but this looks like it should work nicely. Good job.

Since @arlimus did the majority of the work on the parslet stuff for nginx_conf, I'd like to explicitly request his approval on this one, too.

arlimus (dismissed):
Found a tiny edge-case to cover and added it as a test.
Thank you for this great work and the review :grin:

jerryaldrichiii:
@arlimus I took a stab at solving for your quote in quote edge case. Let me know what you think! 😄

jquick (approved):
Thanks @jerryaldrichiii

PR #2730, opened 2018-02-22
package resource: Fix `brew` package detection

jerryaldrichiii:
This allows for package detection via `brew` to handle cases where a
particular package formula exists but is not installed.


miah (approved):

lib/resources/package.rb
```diff
@@ -194,15 +194,23 @@ class Brew < PkgManagement
...
       pkg = JSON.parse(cmd.stdout)[0]
-      {
-        name: pkg['name'],
-        installed: true,
-        version: pkg['installed'][0]['version'],
-        type: 'brew',
-      }
+      # If package exists but is not installed, then `brew` output will not
+      # contain `pkg['installed'][0]['version']
+      unless pkg.dig('installed', 0, 'version')
```
adamleff: `unless`...`else` is really frowned-upon in Rubyland.

Let's just make this:

```ruby
return {} unless pkg.dig('installed', 0, 'version')

# the package is installed, so we build up our pkg info hash
{
  name: ...
}
```

adamleff (changes requested):
Great improvement, but we can be more idiomatically correct Ruby-wise.

lib/resources/package.rb
jerryaldrichiii: I like that much more. Implemented in latest commit.

Thanks

adamleff (approved):

jquick (approved):
Thanks @jerryaldrichiii !

PR #2737, opened 2018-02-23
Fix inspec check to work with platforms

jquick:
The InSpec check command uses the mock backend which does not mock a full aws/azure api. Also some of these create methods on the fly depending on the api request. We always skip supports when loading resources for testing (mock backend) which includes inspec check. To get around this issue if we get a NotDefined raise coming from the train connection and we are in a mock backend we will capture and skip.

This allows `inspec check` to check for all its syntax issues accordingly. Without a proper api backend to hit I don't see how we can give more feedback.

### DIFF
--- a/lib/inspec/plugins/resource.rb
+++ b/lib/inspec/plugins/resource.rb
@@ -43,7 +43,7 @@ module Inspec
       Inspec::Resource.registry
     end
 
-    def __register(name, obj) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    def __register(name, obj) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       cl = Class.new(obj) do # rubocop:disable Metrics/BlockLength
         attr_reader :resource_exception_message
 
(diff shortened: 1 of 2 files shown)


lib/inspec/plugins/resource.rb
```diff
@@ -59,7 +59,8 @@ def initialize(backend, name, *args)
           # check resource supports
           supported = true
           supported = check_supports unless @supports.nil?
-          if defined?(Train::Transports::Mock::Connection) && backend.backend.class == Train::Transports::Mock::Connection
+          test_backend = defined?(Train::Transports::Mock::Connection) && backend.backend.class == Train::Transports::Mock::Connection
+          if test_backend
```
jerryaldrichiii: This could be rewritten as:

`return if !supported && !test_backend`

chris-rock (approved):
This seems reasonable to me for now. We should start to think about a solution that creates a more predictive backend implementation.

clintoncwolfe (approved):
Aside from a fluke functional test failure, this worked fine.  I agree we need to come up with a better way of handling this in the future; intercepting MethodMissing (even with your well-thought-out safeties) is probably a bad habit!

PR #2742, opened 2018-02-25
http resource: Support OPTIONS method

cbeckr:
This patchset adds support for the HTTP OPTIONS method by using `run_request` since Faraday does not support an `options` method.
See https://github.com/lostisland/faraday/issues/305 for background.


jquick (approved):
Thanks for this @cbeckr. Pulled this down and looks good to me!

clintoncwolfe (approved):
Thanks for this - code looks good, and very good tests!

adamleff (approved):
Solid! Thanks for your first contribution to InSpec, @cbeckr :)

PR #2747, opened 2018-02-26
Ensure we have a proper exit code and report data for ad-hoc runners

jquick:
This PR confirms we have the proper @run_data for a ad-hoc report call. It also ensures we pass the rspec error code back if no data was returned.


jquick:
I just confirmed with @alexpop that this is fixing the test issues. This should be good to go.

alexpop (approved):
Thanks for the quick turnaround!

miah (approved):

PR #2753, opened 2018-02-27
Docs: Describe support boundary between RSpec and InSpec

clintoncwolfe:
This adds a bullet list to the "InSpec and Friends" document, describing the differences between InSpec and RSpec, and laying out a support boundary (taking docs.inspec.io as the support surface).

My first draft of this generally aligns with my understanding of the team's perspective, though any over-bluntness / misperceptions are my own.  Feedback very welcome.

### DIFF
--- a/docs/inspec_and_friends.md
+++ b/docs/inspec_and_friends.md
@@ -40,13 +40,12 @@ control "sshd-11" do
 end
 ```
 
-That said, InSpec is not RSpec. InSpec does not support arbitrary RSpec features. Some key differences:
+That said, InSpec is not RSpec. Some key differences:
 
- * RSpec `describe` blocks are intended to be nested. In InSpec, `describe` 
- blocks should not be nested; instead use `control` blocks to describe a 
- higher-level grouping of tests.
- * RSpec is a tool designed for software engineers. It thus supports a very large range of matchers, to enable testing of software engineering constructs such as exceptions, Object Oriented Programming relationships, and so on. In contrast, InSpec is aimed at compliance practioners and infrastructure testers, so our focus is providing a few, well-supported, easy-to-use universal matchers, such as `cmp`.
- * RSpec is a separate project from InSpec. Its release schedule and feature set are beyond the control of the InSpec team. While you are free to use RSpec-specific features in your InSpec profiles, the InSpec team can only support those features described at [docs.inspec.io](https://docs.inspec.io).
+ * In InSpec, `describe` blocks should not be nested; instead use `control` blocks to describe a higher-level grouping of tests.
+ * The RSpec `shared_example` sonstruct is not suppoerted.  Instead, create a simple custom resource that executes repetitious tasks.
+ * InSpec is aimed at compliance practitioners and infrastructure testers, so our focus is providing a few, well-supported, easy-to-use [universal matchers](https://www.inspec.io/docs/reference/matchers/), such as `cmp`. In contrast, RSpec is a tool designed for software engineers. It thus supports a very large range of matchers, to enable testing of software engineering constructs such as exceptions, Object Oriented Programming relationships, and so on. 
+ * While InSpec uses parts of the RSpec project and codebase, it is a separate project from InSpec. Rspec's release schedule and feature set are beyond the control of the InSpec team. While it is possible to use many of the RSpec core features within InSpec profiles, InSpec can only guarantee that the features described at [docs.inspec.io](https://docs.inspec.io) will function correctly. Some RSpec core functionality may be removed in future versions of InSpec as needed to ensure stability in the InSpec project.
 
 ## Serverspec
 
--- a/docs/inspec_and_friends.md
+++ b/docs/inspec_and_friends.md
@@ -9,7 +9,7 @@ relate to each other.
 
 ## RSpec
 
-RSpec is an awesome framework that is widely used by software engineer to test
+RSpec is an awesome framework that is widely used by software engineers to test
 Ruby code. It enables test-driven development (TDD) and helps developers to write
 better code every day.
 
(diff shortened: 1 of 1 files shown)


docs/inspec_and_friends.md
```diff
@@ -9,8 +9,8 @@ relate to each other.
 
 ## RSpec
 
-RSpec is an awesome framework that is widely used to test Ruby code. It
-enables test-driven development (TDD) and helps developers to write
+RSpec is an awesome framework that is widely used by software engineer to test
```
jerryaldrichiii: Could we change this to `software engineers`?

jerryaldrichiii (changes requested):
Looks great! I think it's is the perfect amount of bluntness. I only found one thing that needed changing.

I  also noticed that the first bit of text has 1 space after  a period, but your other bits have 2 spaces. I have no preference on 1 vs 2 spaces, but thought I'd let you know.

jerryaldrichiii (approved):
Great work @clintoncwolfe thanks for this!

kagarmoe (commented):
List the InSpec method first, then contrast with RSpec. The point is to educate users about InSpec.

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,14 @@ control "sshd-11" do
 end
 ```
 
+That said, InSpec is not RSpec. InSpec does not support arbitrary RSpec features. Some key differences:
+
+ * RSpec `describe` blocks are intended to be nested. In InSpec, `describe` 
+ blocks should not be nested; instead use `control` blocks to describe a 
+ higher-level grouping of tests.
+ * RSpec is a tool designed for software engineers. It thus supports a very large range of matchers, to enable testing of software engineering constructs such as exceptions, Object Oriented Programming relationships, and so on. In contrast, InSpec is aimed at compliance practioners and infrastructure testers, so our focus is providing a few, well-supported, easy-to-use universal matchers, such as `cmp`.
```
chris-rock: Also we do not want to have a matcher per resource, since it requires us to split logic that belongs together. We are not accepting new matcher without a very good reason.

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,14 @@ control "sshd-11" do
...
+ * RSpec is a separate project from InSpec. Its release schedule and feature set are beyond the control of the InSpec team. While you are free to use RSpec-specific features in your InSpec profiles, the InSpec team can only support those features described at [docs.inspec.io](https://docs.inspec.io).
```
chris-rock: We should discourage the use or RSpec-specific features

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,14 @@ control "sshd-11" do
 end
 ```
 
+That said, InSpec is not RSpec. InSpec does not support arbitrary RSpec features. Some key differences:
```
adamleff: One key difference I might point out, since it comes up from time-to-time, is that InSpec doesn't support `shared_examples`... instead, we recommend creating a custom resource that performs the repetitious tests as needed.

docs/inspec_and_friends.md
adamleff: While I agree with this sentiment, I don't know that it needs specific call-out, especially in this section. I think it might be confusing since this document is focused on the differences rather than a declaration of what types of changes we'll accept moving forward.

docs/inspec_and_friends.md
adamleff: A strong discouragement of RSpec-specific features will likely alienate some members of the community. I don't know that there's any technical or logistical reason to flat-out discourage it... it's more philosophical.

I might change the wording on this to a middle-ground, such as:

> While InSpec uses parts of the RSpec project and codebase, it is a separate project from InSpec. Its release schedule and feature set are beyond the control of the InSpec team. While it is possible to use many of the RSpec core features within InSpec profiles, InSpec can only guarantee that the features described at [docs.inspec.io](https://docs.inspec.io) will function correctly. Some RSpec core functionality may be removed in future versions of InSpec as needed to ensure stability in the InSpec project.

I think this sets a better "you're on your own, we may accidentally screw you if you use RSpec core features" tone without flat-out discouraging it.

adamleff (commented):
I'm very appreciative that you put this into words, @clintoncwolfe. No blockers from POV, just some suggestions.

docs/inspec_and_friends.md
burtlo: > **arbitrary** - based on random choice or personal whim, rather than any reason or system.
I would consider a different adjective. The text that you use below to describe RSpec's focus as a tool in supporting software engineers feels like it could help describe the goal of their project versus the compliance focus of this project.

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,14 @@ control "sshd-11" do
...
+
+ * RSpec `describe` blocks are intended to be nested. In InSpec, `describe`
```
burtlo: I definitely agree with @kagarmoe suggestions of communicating the InSpec approach contrasted with the RSpec approach.

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,14 @@ control "sshd-11" do
 end
 ```
 
+That said, InSpec is not RSpec. InSpec does not support arbitrary RSpec features. Some key differences:
+
+ * RSpec `describe` blocks are intended to be nested. In InSpec, `describe` 
+ blocks should not be nested; instead use `control` blocks to describe a
```
burtlo: A code sample could help with the comparison story that you are describing.

docs/inspec_and_friends.md
burtlo: > ... so our focus is providing a few, well-supported, easy-to-use universal matchers, such as `cmp`.
Particular in a contrasting document it feels like you may be saying that RSpec is not. Which I don't believe is your intention.

A smaller set of universal matchers that are focused on readability by a wider audience. This is again where a code sample may help convey the story you are trying to tell.

burtlo: With a focus on matchers if feels like a link to matchers content would be appropriate.

docs/inspec_and_friends.md
burtlo: The language construct `control` is mention here. Would adding a link to the appropriate to an appropriate docs page?

docs/inspec_and_friends.md
burtlo: > InSpec is built on top of RSpec and uses it as the underlying foundation to execute tests. It uses the key strengths of RSpec, easily execute tests and a DSL to write tests, but extends the functionality for use as compliance audits. 

For clarity and consistency, I would update DSL to state Domain Specific Language (DSL). 

This is not a part you edited. But if you agree I think it would help the overall document.

burtlo (commented):
Suggests to reduce flammability and bring clarity through examples.

docs/inspec_and_friends.md
```diff
@@ -40,6 +40,13 @@ control "sshd-11" do
 end
 ```
 
+That said, InSpec is not RSpec. Some key differences:
+
+ * In InSpec, `describe` blocks should not be nested; instead use `control` blocks to describe a higher-level grouping of tests.
+ * The RSpec `shared_example` sonstruct is not suppoerted.  Instead, create a simple custom resource that executes repetitious tasks.
```
burtlo: "sonstruct is not suppoerted" to "construct is not supported"

arlimus (approved):
Thank you all!

PR #2756, opened 2018-02-28
Fix AWS IAM access keys example

duffn:
There is no `created_age`. This should be `created_days_ago`.

Running the example as currently provided, throws an exception.

```
aws/controls/iam.rb:10:in `block (2 levels) in load_with_context': undefined local variable or method `created_age' for #<#<Class:0x00007fd3aea43150>:0x00007fd3b11d8468> (NameError)
```

### DIFF
--- a/docs/resources/aws_iam_access_keys.md.erb
+++ b/docs/resources/aws_iam_access_keys.md.erb
@@ -41,7 +41,7 @@ The following examples show how to use this InSpec audit resource.
 
 ### Disallow access keys created more than 90 days ago
 
-    describe aws_iam_access_keys.where { created_age > 90 } do
+    describe aws_iam_access_keys.where { created_days_ago > 90 } do
       it { should_not exist }
     end 
 


clintoncwolfe (approved):
Great catch! Congratulations on your first contribution to this project!

docs/resources/aws_iam_access_keys.md.erb
```diff
@@ -41,7 +41,7 @@ The following examples show how to use this InSpec audit resource.
 
 ### Disallow access keys created more than 90 days ago
 
-    describe aws_iam_access_keys.where { created_age > 90 } do
+    describe aws_iam_access_keys.where { created_days_ago > 90 } do
```
aaronlippold, line 44: hi would `key_age` be a more concise property here?

aaronlippold (commented):
would `key_age` be a better way to label this?

docs/resources/aws_iam_access_keys.md.erb
trickyearlobe: Or even just `age` since we already know we are describing a key ;)

clintoncwolfe: I'd refer to keep the word "created" in there - that would distinguish it from other potential meanings (perhaps age since rotation, for example).  And I'd like to keep `days` in there as well, since there is almost always confusion over units.

aaronlippold: I agree it is a good thing to have the units - Matt and I ran against this same issue with S3 ACLs - and that a standard interface / functions made sense. So in general perhaps we should try to come up with a common pattern to unify how our resources handle this 'age' issue.

I think there are three main units which we should cover at some point:  

- epoch time
- days
- months  

I wonder if we can use something like this as a standard:

```
{ age_created.days > 90 }
{ age_created.moths <= 6 }
{ age_created.epoch ... }
```
I am not sure if folks really use epoch anymore but in general having a standard function of `days` and `months` in the FilterTable for all things that require a 'time' delta I think makes sense. 

Most of our security guidance will want to ensure that things are only so old - passwords, keys, user accounts, even s3 buckets or documents stored in buckets - etc. etc.

jquick (approved):
Thanks @duffn !

PR #2761, opened 2018-03-01
Fix aws-iam-users pagination

### LINKED ISSUE #2760: aws-iam-users only retrieves 100 users
### 🤔 Replication Case
With more than 100 active users in the AWS account

```
describe aws_iam_users do
  its('entries.length') { should be >100 }
end
```

### 💁 Possible Solutions

Loop over the `backend.list_users` function in `fetch_from_api` using the `marker` and `is_truncated` variables in the returned API object.

I will submit a PR for this issue shortly.

trickyearlobe:
PROBLEM: aws-iam-users resource only retrieves 100 records due to pagination
in the AWS IAM list_users function.

FIX: Iterate over all the pages using the AWS pagination variables `marker`
and `is_truncated`

Fixes #2760 and part of #2582 (which concerns additional resources)

### DIFF
--- a/lib/resources/aws/aws_iam_users.rb
+++ b/lib/resources/aws/aws_iam_users.rb
@@ -37,17 +37,21 @@ class AwsIamUsers < Inspec.resource(1)
     raw_params
   end
 
+  def fetch_from_api_paginated(backend)
+    table = []
+    page_marker = nil
+    loop do
+      api_result = backend.list_users(marker: page_marker)
+      table += api_result.users.map(&:to_h)
+      page_marker = api_result.marker
+      break unless api_result.is_truncated
+    end
+    table
+  end
+
   def fetch_from_api
     backend = BackendFactory.create(inspec_runner)
-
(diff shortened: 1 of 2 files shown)


clintoncwolfe (changes requested):
This is much simpler than the approach I was going to take (in which `list_users` may be treated as an iterator).  Unfortunately this breaks the unit tests, though the fix is simple.  Since `list_users` is now being called with params, alter lines 106 and 131 of `test/unit/resources/aws_iam_users_test.rb` to read `def list_users(criteria = {})`

trickyearlobe:
Cheers @clintoncwolfe 
Next time I'll run both lint and unit before I raise the PR

clintoncwolfe (approved):

clintoncwolfe:
Thanks - looks great!  I've borrowed your approach and applied it to several other resources.

trickyearlobe:
Sweet... that means I don't have to finish my generic `AwsApiDepaginator` :)

jquick (approved):
Thanks @trickyearlobe !

PR #2762, opened 2018-03-01
AWS API Pagination fixes

### LINKED ISSUE #2582: AWS API Pagination Ignored
<a href="https://github.com/clintoncwolfe"></img></a> **Issue by [clintoncwolfe](https://github.com/clintoncwolfe)**
_Friday Jan 05, 2018 at 16:05 GMT_
_Originally opened as https://github.com/chef/inspec-aws/issues/171_

----

In many locations where we make an API call to AWS using the "client" style interface (as opposed to the "resource" style interface, which tends to use iterators), the API often has a limit on the number of records it will return before paginating.  Currently we don't detect this, which means we're not actually testing many resources.

Identify API calls which we touch that may be sensitive to this issue.

Determine a common solution if possible.  One approach might be to use the Resource approach.

And of course, fix.

- [x] aws_iam_groups - list_groups call
- [x] aws_iam_policy - list_policies call
- [x] aws_iam_policies - list_policies call
- [x] aws_kms_keys - list_keys call
- [x] aws_iam_access_keys - list_users call
- [ ] aws_iam_users - list_users call - #2761 

All plural resources should be examined.
- [x] aws_cloudtrail_trails - describe_trails is not paginated
- [x] aws_subnets - describe_subnets is not paginated
- [x] aws_vpcs - describe_vpcs is not paginated
- [x] aws_security_groups - describe_security_groups call is only paginated if we supply the `max_results` option.

clintoncwolfe:
Fixes #2582 for most current resources (#2761 fixes `aws_iam_users`, and this PR fixes all others).  Uses a simple loop, sending the token back to AWS when needed.  Unit tests and integration tests are happy.

### DIFF
--- a/lib/resources/aws/aws_iam_access_keys.rb
+++ b/lib/resources/aws/aws_iam_access_keys.rb
@@ -85,9 +85,14 @@ class AwsIamAccessKeys < Inspec.resource(1)
             # Swallow - a miss on search results should return an empty table
           end
         else
-          # TODO: pagination check and resume
-          iam_client.list_users.users.each do |info|
-            user_details[info.user_name] = info
+          pagination_opts = {}
+          loop do
+            api_result = iam_client.list_users(pagination_opts)
+            api_result.users.each do |info|
+              user_details[info.user_name] = info
+            end
+            break unless api_result.is_truncated
+            pagination_opts[:marker] = api_result.marker
           end
         end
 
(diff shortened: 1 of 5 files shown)


jquick (approved):
Thanks @clintoncwolfe !

miah (approved):

Issue #2784, opened 2018-03-06
Add `aws_ec2_instances` plural resource


trickyearlobe:
### 🎛 Description

Inspec contains singular and plural resources for various AWS objects, but seems to be missing a plural resource for EC2 instances.

Its quite helpful to have a plural resource for performing filter driven tests across large numbers of objects.

I have some basic code which

* exposes a collection of EC2 instances
* filters them using Inspec's built in FilterTable class.

I'll submit a PR for discussion and improvement.

skpaterson:
This is available in the new InSpec AWS resource pack:
https://github.com/inspec/inspec-aws/blob/master/docs/resources/aws_ec2_instances.md

Issue #2779, opened 2018-03-05
Resources inconsistently define exist? / exists? methods


burtlo:
### 🎛 `exist?` and `exists?`use is inconsistent across resources

🙋 feature request

The `user` resource defines `exist?` but not `exists?`.
The `file` resource defines `exists?` but not `exist?`.

Having a consistent method or supporting both will make it an easier experience when working with InSpec resources.

### 🌍 InSpec and Platform Version

InSpec 2.0.17, Mac 10.13.13

### 🤔 Replication Case

Given that you have a user that exists with the name `mr_robot` and a file that exists at the path `fsociety.dat`.

```
$ inspec shell
inspec> user('mr_robot').exists?
=> true
inspec> file('fsociety.dat').exists?
NoMethodError: undefined method `exists?' for File fsociety.dat:#<Class:0x00007fa7830dfca0>
from (pry):36:in `load_with_context'
inspec> file('fsociety.dat').exist?
=> true
```

### 💁 Possible Solutions

* Choose either `exist?` or `exists?`.

I do not get a feeling that a standard has been defined for all resources. It feels like there is usage of both as I have found in this instance. It sounds like the community would need to agree on what is preferred, the methods not in alignment would need to be deprecated, the documentation would need to be upgraded, and then the deprecated choice would need to be removed in the next release.

* Support both `exist?` and `exists?`

As a standard has not been defined perhaps the best approach is to support both. Review existing resources and add aliases to ensure they support both methods. It may still be a good idea to review the documentation to ensure the usage is consistent.

* Take no action

When defining controls with the expectations around existence the differing methods do not appear to impact the individual test case written.

jquick:
This is a good point! I know ruby moved to `exist?` and deprecated `exists?` for its File methods. Not saying we should because of that but noting it. I would say lets pick one and deprecate the other.

My vote is to move them to `exist?`

jquick:
Note, if we plan on updating the file to support `exist?` That would be a Train change.

miah:
__What’s happening? Why was this issue closed?__
This issue was closed due to some much needed review of legacy issues or issues that were spawned in older versions of InSpec, i.e. < v3.

__Why do I care?__
You would care about this if this was an issue that you are still seeing and/or feel needs to be addressed in the current version of InSpec.

__What do I need to do?__
If this issue is no longer important, no further action is necessary. However, if you think this is something that should be addressed, please open a new issue and refer to the original issue in the description.

PR #2785, opened 2018-03-06
Add new plural resource aws_ec2_instances

trickyearlobe:
Sample implementation to implement #2784 
Does this look like its on the right lines?

* Contains a possible implementation
* Contains unit tests
* Contains a documentation page

TODO's
* Doesn't implement the new style AWS Backend (yet)
* Unit tests are a bit thin?


docs/resources/aws_ec2_instances.md.erb
```diff
@@ -0,0 +1,60 @@
...
+
+An `aws_ec2_instances` resource block declares the tests for multiple AWS EC2 instance with filtering
+on many instance properties.
+
+    # test without filters
+    describe aws_ec2_instances do
+      its('entries.count') ( should eq 22 )
+    end
+
+    # Test with a filter (as params)
```
clintoncwolfe, line 22: We use the word "criteria" here now - "resource params" are things you pass to the constructor.

docs/resources/aws_ec2_instances.md.erb
```diff
@@ -0,0 +1,60 @@
...
+    describe aws_ec2_instances.where(instance_id: 'i-01a2349e944580000') do
+      it { should exist }
+    end
+
+    # Test with a filter (as a block)
+    describe aws_ec2_instances.where{ public_ip_address != nil && state == 'running'} do
+      its('instance_id') { should eq [] }
```
clintoncwolfe, line 29: It would be good to add a comment explaining what the example is aiming for.  Here, you I think you're saying you don't want any running instances with public IPs.  We're also trying to encourage tests to be more "fluent, so you might try `it { should_not exist }` rather than checking for an empty instance_id array.

docs/resources/aws_ec2_instances.md.erb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    # Test with a wildcard filter (as a block)
+    describe aws_ec2_instances.where{ private_dns_name == /.*eu-west.*/' } do
+      its('instance_id') { should eq [] }
```
clintoncwolfe, line 34: Let's use `exist` here as well.

docs/resources/aws_ec2_instances.md.erb
```diff
@@ -0,0 +1,60 @@
...
+
+    describe aws_ec2_instances.where(instance_id: 'i-01a2349e944580000') do
+      it('entries') { should eq [] }
+    end
+
+<br>
+
+## Properties
+
+`ami_launch_index`, `image_id`, `instance_id`, `instance_type`, `kernel_id`, `key_name`, `launch_time`, `monitoring`, `placement`, `platform`, `private_dns_name`, `private_ip_address`, `product_codes`, `public_dns_name`, `public_ip_address`, `ramdisk_id`, `state`, `state_transition`, `subnet_id`, `vpc_id`, `architecture`, `block_device_mappings`, `client_token`, `ebs_optimized`, `ena_support`, `hypervisor`, `iam_instance_profile`, `instance_lifecycle`, `elastic_gpu_associations`,  `network_interfaces`, `root_device_name`, `root_device_type`, `security`, `source_dest_check`, `spot_instance_request_id`,  `sriov_net_support`, `state_reason`, `tags`, `virtualization_type`
```
clintoncwolfe, line 53: I think you may have used `aws_ec2_instance` as a starting point; it was a very early resource and is underdocumented.  At this point, we need docs for each individual property, with a brief example.

For a plural resource, this is a very generous property list, which is awesome.   That said, it's a lot of work to document and test all of these.  Consider stripping this down to the ones you need, and letting the resource grow in the future.

docs/resources/aws_ec2_instances.md.erb
```diff
@@ -0,0 +1,60 @@
...
+
+<br>
+
+## Matchers
+
+This InSpec resource has no special matchers.
+For a full list of available matchers, please visit our [matchers page](https://www.inspec.io/docs/reference/matchers/).
```
clintoncwolfe, line 60: Please refer to this as the "universal matchers" page.  We're trying to clarify that :)

docs/resources/aws_ec2_instances.md.erb
clintoncwolfe, line 53: Also, as this is a plural resource, each of its properties should be named in the plural:
```
describe aws_ec2_instances do
  # Ensure we are running only vegetarian AMIs
  its('image_ids') { should_not include 'ami-deadbeef' }
end
```

Mentally, I substitute "they" for "it" and "theirs" for "its".  (Which I think would be a good feature, personally).

lib/resources/aws/aws_ec2_instances.rb
```diff
@@ -0,0 +1,72 @@
+# author: Christoph Hartmann
+class AwsEc2Instances < Inspec.resource(1)
+  name 'aws_ec2_instances'
+  desc 'Verifies settings for an EC2 instance'
+  supports platform: 'aws'
+  example "
+    describe_ec2_instances.where { public_ip_address !=nil } do
+      its('instance_id') { should eq [] }
```
clintoncwolfe, line 8: `it { should_not exist }`

lib/resources/aws/aws_ec2_instances.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  # TODO: rewrite to avoid direct injection, using AwsPluralResourceMixin
+  def initialize(opts = {}, conn = nil)
+    @opts = opts
+    @ec2_client = conn ? conn.ec2_client : inspec_runner.backend.aws_client(Aws::EC2::Client)
+  end
+
+  # Set up some accessors and the filter table
+  filter = FilterTable.create
+  filter.add_accessor(:where).add_accessor(:entries)
```
clintoncwolfe, line 20: Ahh, I see why you're not using `exist` - it's not there!  You should add it.  It's simple, 
`filter.add(:exists?) { |x| !x.entries.empty? }`

That's pretty much standard in all of our plural resources.  Well, [pretty much](#2779).

lib/resources/aws/aws_ec2_instances.rb
```diff
@@ -0,0 +1,72 @@
...
+    filter.add(field, field: field)
+    # Add an accessors to access properties without the filter table
+    define_method(field) do
+      list.map(&:field)
+    end
+  end
+  filter.connect(self, :list)
+
+  # Give uses of `inspec shell` a few hints about what they can test
+  def possible_properties
```
clintoncwolfe, line 32: Hmmm, I was not aware of this.  Great use of a feature!

lib/resources/aws/aws_ec2_instances.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  # Calls self.class.list to gain access to 'class instance variables'
+  def list
+    self.class.list(@ec2_client)
+  end
+
+  def self.list(ec2_client)
+    # Make sure we only call depaginate_api once
+    @aws_ec2_instances_cache ||= depaginate_api(ec2_client)
```
clintoncwolfe, line 51: I think this violates our caching philiosophy.  To date, I think we've been careful to cache only connections, and avoid caching any data returned from an API.  We don't know how long an InSpec run may live - if they are scanning S3 on a per-object basis, it could be hours - and this would result in stale EC2 data.  Additionally, to date we've treated Train as being the component responsible for providing a cache authority (with facilities for disabling, expiring, etc).  @jquick, @chris-rock , what do you think?

lib/resources/aws/aws_ec2_instances.rb
```diff
@@ -0,0 +1,72 @@
...
+  name 'aws_ec2_instances'
+  desc 'Verifies settings for an EC2 instance'
+  supports platform: 'aws'
+  example "
+    describe_ec2_instances.where { public_ip_address !=nil } do
+      its('instance_id') { should eq [] }
+    end
+  "
+
+  # TODO: rewrite to avoid direct injection, using AwsPluralResourceMixin
```
clintoncwolfe, line 12: Have a look at more recent plural resources - aws_security_groups, for example.  They use the AwsPluralResource mixin to handle several things, including resource param validation and testing backend setup.

test/unit/resources/aws_ec2_instances_test.rb
```diff
@@ -0,0 +1,191 @@
+require 'helper'
+
+class TestAwsEc2Instances < Minitest::Test
+
+#=============================================================================#
```
clintoncwolfe, line 5: So, this obviously uses a different style that the rest of AWS unit tests - which is OK in general - but we really do have to avoid dependency injection in the constructor.  If you rework this to use the BackendFactory and place your fixture data in a dummy backend, that would be fine.  Have a look at the other AWS plural resources - aws_iam_policies or aws_security_groups would be typical.

test/unit/resources/aws_ec2_instances_test.rb
```diff
@@ -0,0 +1,191 @@
+require 'helper'
```
clintoncwolfe, line 1: Really need some integration tests - under `test/integration/aws/default/verify/controls/`.  You may also need to add more EC2 instances to get enough variety to test the filters.  I'm happy to help.

clintoncwolfe (changes requested):
Hi!  Great submission.  There are several things that I'd like to see adjusted to match other AWS resource standards, and we'd need to see the docs and integration tests expanded.  Thanks for your contribution!

docs/resources/aws_ec2_instances.md.erb
trickyearlobe: The property list is dynamically generated from properties exposed by the AWS API... would you prefer I explicitly define them ?

docs/resources/aws_ec2_instances.md.erb
trickyearlobe: ok... i better define `exist` then ;-)

I will probably still show an example like this with more comments about why I'm doing it because customers often ask (in other resources) how to know which ones failed. When it fails its nice to get something like
```
   expected: [ ]
   got: ['i-3892934', 'i-4769259' ]
```
instead of 
```
    expected: false
    got: true
```

docs/resources/aws_ec2_instances.md.erb
trickyearlobe: The properties are dynamically defined by iterating over the properies exposed in the AWS API. I can leave them dynamic and use `pluralize` from `ActiveSupport`, or I can statically define them (See previous comment)

lib/resources/aws/aws_ec2_instances.rb
trickyearlobe: Note, that `its('instance_id) { should eq []}` makes Inspec spit out a list of instance ID's that failed... instead of just a rather uninformative `failed` message

lib/resources/aws/aws_ec2_instances.rb
trickyearlobe: Will do...

lib/resources/aws/aws_ec2_instances.rb
trickyearlobe: I also avoid it because I want an explicit list of failed instances, rather than a `failed` message.

lib/resources/aws/aws_ec2_instances.rb
trickyearlobe: This is where I get the list of properties that need to be dynamically defined using `define_method` (see note about excessive number of exposed properties and the documentation burden)

lib/resources/aws/aws_ec2_instances.rb
trickyearlobe: We should probably review that policy...

Take a look at performance of `aws_iam_users`. 

- It takes approx 1 min to pull all the users in an AWS account with 215 users.
- If I use `describe aws_iam_users` twice, it runs for twice the time.
- If I modify it to cache at class level instead of instance level its runtime remains roughly constant.

Lazy loading in the filter table may help, but
 - it still takes significant time for the API calls
 - it generates a ton of unnecessary network traffic
 - it still runs slow if a lazy-load field is actually used

Not sure we really care about data being a few seconds/minutes stale.

test/unit/resources/aws_ec2_instances_test.rb
trickyearlobe: I'll have to check how you're doing those... might need to ping you for help.

lib/resources/aws/aws_ec2_instances.rb
clintoncwolfe: We had a chat in the InSpec team standup on 2018-03-19, and my concerns were mostly assuaged.  

* I agree that seconds or minutes are not a concern.  However, some resources - s3 objects for example - have an API that is aggressively unscalable (requiring, for example, a separate API call for each object to determine ACLs).  Running that on a S3 bucket with thousands or millions of objects would take hours (and yes, it's likely the wrong tool - just pointing out that hour-long executions are conceivable).
* You can disable Train caching with a CLI option to `inspec exec`
* We can add a doc warning that results may be cached
* Relatively few users would fetch EC2 instances, do something long running, then check EC2 instances again (I'd love to hear a common use case if you have one)
* With caching enabled, we can think of the scan as being a point-in-time compliance check.  Without caching, it's like multiple scans.
* It is possible to clear the cache, and we can document that better so that affected users will have a workaround.

All that said, my only concern is where the code is located - I think we want caching in the general case in Train.  @jquick , any advice?

docs/resources/aws_ec2_instances.md.erb
clintoncwolfe: Hmm, I see - I like that error message much better.  How could we get that UX in the error message, while still having better readability of the profile code?

```ruby
describe aws_ec2_instances.where{ private_dns_name == /.*eu-west.*/' } do
      # Most plural resources define 'entries', which simply returns the array-of-hashes that FilterTable acts on.  
      # be_empty is provided by RSpec
      its('entries') { should be_empty }
     # You could alias it to 'results' for a bit better readability
      is('results') { should be_empty }
     # but Instance IDs is probably your best bet for brevity
      is('instance_ids') { should be_empty }
end
```

It does barf out the whole structure when failing - which could be good or bad, I suppose.  Focusing on instnace ID would be the most clear.

Here's a shell session, trying it on users.
```
inspec> describe aws_iam_users do
inspec>   its('entries') { should be_empty }
inspec> end

Profile: inspec-shell
Version: (not specified)

  IAM Users
     ×  entries should be empty
     expected `[#<struct :exists?=nil, has_mfa_enabled=false, has_console_password=false, :password_ever_used?=false...e, :password_never_used?=true, password_last_used_days_ago=nil, user_name="test-user-readonly-ec2">].empty?` to return true, got false

```

clintoncwolfe:
@trickyearlobe , would you mind if I adopted this PR?

lib/resources/aws/aws_ec2_instances.rb
clintoncwolfe: Refs #3127

jquick:
Thanks so much for this @trickyearlobe ! At this point we are going to close this and likely use it as building blocks for a future resource. We appreciate all you have done and setting the foundation!

PR #2792, opened 2018-03-07
return 1 as exit code for commands that are not available

### LINKED ISSUE #2787: Inspec CLI returns 0 for a non-existent command
### 🐛 Description

inspec CLI returns 0 when a non-existent command is provided at command line.
Since it's possible to add plugin commands to inspec; it is realistic to want to test whether the command exists or not.
If the plugin command is not correctly loaded then inspec should not return 0 for that command.

1.51.21
OSX 10.12

### 🤔 Replication Case

```
➜  my_repo git:(master) ✗ inspec party help
Could not find command "party".
➜  my_repo git:(master) ✗ echo $?
0
```

### 💁 Possible Solutions

Not sure exactly where the fix is needed; but `Could not find command` comes directly from thor so presumably inspec is not checking for and responding to thor errors in at least one place.

### 💻 Stacktrace

None, just the terminal output above

chris-rock:
Fixes #2787

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -8,6 +8,11 @@ require 'inspec/profile_vendor'
 
 module Inspec
   class BaseCLI < Thor
+    # https://github.com/erikhuda/thor/issues/244
+    def self.exit_on_failure?
+      true
+    end
+
     def self.target_options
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @chris-rock !

PR #2795, opened 2018-03-07
New Resource: MS Message Queue

TheLonelyGhost:
This is derived from a custom resource I wrote for internal use. I did my best to convert it from being a custom resource drawing from files in a profile to being part of the inspec standard library, but I'm at a complete loss for how to best add automated tests for it. Help?

---

Here's the theory behind this implementation:

Server 2008r2 does not support any of the later MSMQ APIs exposed in later .NET versions. We need to either backport them or find a more neutral way of getting to the same information. Luckily, there was [someone else on the internet](https://github.com/jlevitt/MSMQSecurity) who had the same issue and decided to open source their solution. It was written in C# which means we can use it with PowerShell via something called P/Invoke.

This C# library calls into some really old DLLs to interact with MSMQ internals at a super low level. It converts them to C# data structures and we can manipulate them from there. That doesn't help us though. We need them in PowerShell.

Here's a deep-dive of how we're traversing the different layers:

```
Ruby (inspec) -> PowerShell -> C# -> C (dll) -> C# -> PowerShell -> Ruby (inspec)
```

Since there's a lot of serialization and deserialization with all of this, there are lower-level accessors for more sensitive checks. This inspec resource provides the `#allowed?` method for digging into binary comparison of the rights that user is allowed for the given queue, but comparing it at the PowerShell level. Same with `#denied?`.

The comparison at the PowerShell level is to reduce the number of conversions before comparing, thereby reducing the number of moving parts. The side effect of it is that we have a winrm call for each check on a user's permissions. This is not cached. This has room for improvement.

---

Yes, it's ugly. Yes, it's a nightmare to maintain. No, I don't want to maintain it long-term. It's the only solution I've found to accomplish what we need for checking MSMQ permissions.

### DIFF
--- a/lib/resources/ms_message_queue.rb
+++ b/lib/resources/ms_message_queue.rb
@@ -89,26 +89,8 @@ module Inspec::Resources
       end
     end
 
-    # If negating `#allowed?`, use this method instead. A lot may be lost serializing and deserializing the
-    # binary values, so it's best to just have PowerShell handle the comparison.
     def denied?(permission, opts = {})
-      script = allowed_wrapper(permission, opts) + "\n\n"
-      script += 'If (($mask -band $target_perm) -ne $target_perm) { Exit 98; } Else { Exit 99; }'
-
-      cmd = inspec.powershell(script)
-
-      case cmd.exit_status
-      when 98 then return true
-      when 99 then return false
-      else
-        # It should never reach here, but for debugging, this will be left in
-        obj = {
-          code: cmd.exit_status,
-          stderr: cmd.stderr,
-          stdout: cmd.stdout,
-        }
-        raise "Problem with P/Invoke: #{obj.inspect}"
-      end
+      !allowed?(permission, opts)
     end
 
     private
--- a/lib/resources/ms_message_queue.rb
+++ b/lib/resources/ms_message_queue.rb
@@ -1,0 +1,222 @@
+# encoding: utf-8
+
+require 'utils/pinvoke_wrapper'
+
+module Inspec::Resources
+  # NOTE: This resource (currently) only works with private queues on the localhost. It supports Windows 2008r2 up to 2016
+  class MicrosoftMessageQueue < Inspec.resource(1)
+    name 'ms_message_queue'
+    supports platform: 'windows'
+    desc <<-EOH
+      Provides visibility into a Microsoft Message Queue
+    EOH
+
+    example <<-EOH
+      describe ms_message_queue('.\\private$\\MyQueueHere') do
+        it { should exist }
+        its('QueueName') { should eq 'MyQueueHere' }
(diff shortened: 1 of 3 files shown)


lib/resources/ms_message_queue.rb
```diff
@@ -0,0 +1,222 @@
...
+          stdout: cmd.stdout,
+        }
+        raise "Problem with P/Invoke: #{obj.inspect}"
+      end
+    end
+
+    # If negating `#allowed?`, use this method instead. A lot may be lost serializing and deserializing the
+    # binary values, so it's best to just have PowerShell handle the comparison.
+    def denied?(permission, opts = {})
+      script = allowed_wrapper(permission, opts) + "\n\n"
```
chris-rock: could this be?:
```
!allowed?(permission, opts)
```

chris-rock (commented):
This is awesome work @TheLonelyGhost . I worry though that it is difficult to maintain for us, since we do not have the 2008 environment to ensure it works. Would it be an approach to reduce this implementation to the pure powershell solution for newer windows versions only. You could still use your solution in your environment. But once you're environment moves on, you just switch the built-in resource then.

lib/resources/ms_message_queue.rb
TheLonelyGhost: Possibly...? I'm having trouble thinking of an example where negating the other method's `If (($mask -band $target_perm) -eq $target_perm)` result wouldn't work the same as this version.

TheLonelyGhost:
I totally get that it would be difficult to maintain (which is why I wanted to put it here 😜) but figured I'd offer it up as an option to integrate into core. Oddly enough, it might be easier to maintain in core due to issues with profile inheritance, but that's another discussion.

For testing against Server 2008R2, I've been using [an image on vagrantcloud](https://app.vagrantup.com/salesforce/boxes/server2008r2) with virtualbox. I work with the guy who uploaded the image and we use it almost every day.

As for the forward-thinking of a PowerShell-only solution...

=== incoming rant ===

I've found `Get-ACL` to be painfully inconsistent between 2008r2, 2012, and 2016, and that's just for filesystem permissions: a well-documented API in .NET. I have very little faith that a better solution will be made for MSMQ permissions when there is no .NET API for it and the only CLI access is a set of opaque PowerShell modules released for 2012+. Nothing was backported for 2008r2. Given the same version of PowerShell on both 2008r2 and 2012, 2008r2 did not have support for those powershell commands and 2012 did.

=== /rant ===

I suppose all that ^^^ is misdirected frustration at Microsoft. I totally get if you think this is too much complexity to take into the InSpec code base.

clintoncwolfe:
I think for the time being this is best left as a custom resource.  Closing for now.

PR #2801, opened 2018-03-08
Refine deprecated methods to be consistent with supported fields in shadow file.

miah:
After much thought the deprecations from #2642  were for the wrong methods.

Plural method names feel much more natural when working with this
resource because you can have more than a single result.

Consider a match like `shadow.user(/^www/)`, this could return multiple
users, so `shadow.users` feels more natural here.

The problem is that the fields we're matching in the shadow file itself
are singular. Each entry is for a user, which has a password, and some
other fields. A user never has `passwords` in the shadow file, only a
`password`.

This is made more obvious when you use the `filter` method.

When we use this filter: `shadow.filter(min_days: 20, max_days: 30)` we
are matching fields in the shadow file and not using our matcher
methods. This means that if there is a discrepancy between our matcher
methods, and the shadow fields the user could end up confused. Like I did =)

This PR changes:

Changed matchers to match shadow fields.
Updated documentation to reflect changes.
Updated tests to reflect changes.
Re-add `filter` method, and add a test for it.
Renamed variable for FilterTable to be less confusing.
Renamed query argument for methods to be consistent.

### DIFF
--- a/docs/resources/shadow.md.erb
+++ b/docs/resources/shadow.md.erb
@@ -27,12 +27,12 @@ These entries are defined as a colon-delimited row in the file, one row per user
 A `shadow` resource block declares one (or more) users and associated user information to be tested:
 
     describe shadow do
-      its('users') { should_not include 'forbidden_user' }
+      its('user') { should_not include 'forbidden_user' }
     end
 
 or with a single query:
 
-    describe shadow.users('root') do
+    describe shadow.user('root') do
       its('count') { should eq 1 }
     end
 
@@ -44,9 +44,9 @@ or with a filter:
 
 The following properties are available:
 
-* `users`
-* `passwords`
-* `last_changes`
+* `user`
+* `password`
+* `last_change`
 * `min_days`
 * `max_days`
 * `warn_days`
@@ -65,13 +65,13 @@ The following examples show how to use this InSpec audit resource.
 ### Test for a forbidden user
 
     describe shadow do
-      its('users') { should_not include 'forbidden_user' }
+      its('user') { should_not include 'forbidden_user' }
     end
 
 ### Test that a user appears one time
 
-    describe shadow.users('bin') do
-      its('passwords') { should cmp 'x' }
+    describe shadow.user('bin') do
+      its('password') { should cmp 'x' }
       its('count') { should eq 1 }
     end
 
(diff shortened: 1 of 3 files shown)


jquick (approved):
This looks good. Thanks for catching this @miah !

docs/resources/shadow.md.erb
```diff
@@ -89,56 +89,58 @@ The `count` matcher tests the number of times the named user appears in `/etc/sh
...
 
-### expiry_date
-
-The `expiry_date` matcher tests the number of days a user account has been disabled:
-
-    its('expiry_date') { should eq '' }
+### user
 
-### inactive_days
+The `user` matcher tests if the user name exists `/etc/shadow`:
```
jerryaldrichiii: Could we change this to `username`?

docs/resources/shadow.md.erb
```diff
@@ -89,56 +89,58 @@ The `count` matcher tests the number of times the named user appears in `/etc/sh
...
+    its('password') { should cmp '*' }
 
-### max_days
+### last_change
 
-The `max_days` matcher tests the maximum number of days after which a password must be changed:
+The `last_change` matcher tests the last time a password was changed:
 
-    its('max_days') { should eq 90 }
+    its('last_change') { should eq '' }
```
jerryaldrichiii: Will `last_change` ever be `''`? Or, will it be days since the user was created and the password was set?

docs/resources/shadow.md.erb
```diff
@@ -89,56 +89,58 @@ The `count` matcher tests the number of times the named user appears in `/etc/sh
...
-The `users` matcher tests if the user name exists `/etc/shadow`:
+    its('warn_days') { should eq 7 }
 
-    its('users') { should eq 'root' }
+### inactive_days
 
-### warn_days
+The `inactive_days` matcher tests the number of days a user must be inactive before the user account is disabled:
+
+    its('inactive_days') { should eq '' }
```
jerryaldrichiii: Wouldn't this be `should be_empty`?

docs/resources/shadow.md.erb
```diff
@@ -89,56 +89,58 @@ The `count` matcher tests the number of times the named user appears in `/etc/sh
...
+
+### expiry_date
+
+The `expiry_date` matcher tests the number of days a user account has been disabled:
+
+    its('expiry_date') { should eq '' }
```
jerryaldrichiii: Wouldn't this be `should be_empty`?

jerryaldrichiii (changes requested):
Looks good overall! Just a few minor changes.

docs/resources/shadow.md.erb
miah: Yes, I'll correct those. Thanks =)

docs/resources/shadow.md.erb
miah: Its generally going to be a timestamp (epoch since last change), but its entirely possible to be empty too. 

Eg, system users e.g. `bin` or `daemon` may not have a password set, or any values for the other fields. Its entirely dependent on the `useradd` program doing the right thing and referencing `login.defs`. Its also possible to update the shadow file without `useradd` so anything can happen really.

docs/resources/shadow.md.erb
miah: Sure.

jerryaldrichiii (approved):
Looks good! Thanks @miah for the changes.

PR #2815, opened 2018-03-14
Override unf_ext with Solaris patch version

jquick:
This change loads in unf_ext with the solaris fix. This is a temporary bind until:

https://github.com/knu/ruby-unf_ext/pull/39

Can be merged/released accordingly.

Credit to @scotthain

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,11 @@
 # encoding: utf-8
 source 'https://rubygems.org'
+
+# override for unf_ext until
+# https://github.com/knu/ruby-unf_ext/pull/39
+# is merged and released
+gem 'unf_ext', '=0.0.7.6', :git => 'https://github.com/jquick/ruby-unf_ext.git'
+
 gemspec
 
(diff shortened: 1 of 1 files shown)


miah (approved):
We should also create a issue with a goal of un-forking after they've merged our patch upstream so we don't forget =)

trickyearlobe:
We should also find a way to avoid using `unf_ext` because gems that need compilation are a major problem in enterprise environments.

In the enterprise, build tools are usually prohibited on prod machines, and the libs/headers are unlikely to be installed. Often they don't even have internet access.

Managing distribution of pre-compiled gems is a major PITA (unless they are in omnibus)

miah:
I agree, this is happening now because we added the Azure gem. It might be
possible to change that upstream I haven't looked too deeply.

jquick:
We can also do this on the Train level instead of `inspec` which may be cleaner for now.

jquick:
I am moving this fix to train. Closing.

jquick:
@trickyearlobe - This is only for testing, you can install inspec after release via the omnibus packages which pre-compile it all for you.

PR #2824, opened 2018-03-15
Documentation: `supports` attribute of custom resource DSL

TheLonelyGhost:
I wasn't sure the best way to document this, so I figured I'd give it an initial pass and get some feedback. Thoughts?

### DIFF
--- a/docs/dsl_resource.md
+++ b/docs/dsl_resource.md
@@ -37,15 +37,25 @@ require a higher version.
 
 The following attributes can be configured:
 
-* name - Identifier of the resource (required)
-* desc - Description of the resource (optional)
-* example - Example usage of the resource (optional)
-* supports - Platform restrictions of the resource. Without specifying anything, it implicitly supports all OS's and releases (optional) (starting in InSpec 2.0)
+<dl>
+  <dt>name</dt>
+  <dd>Identifier of the resource (required)</dd>
+  <dt>desc</dt>
+  <dd>Description of the resource (optional)</dd>
+  <dt>example</dt>
+  <dd>Example usage of the resource (optional)</dd>
+  <dt>supports</dt>
+  <dd><small>(InSpec 2.0+)</small> Platform restrictions of the resource (optional)</dd>
+</dl>
 
 The following methods are available to the resource:
 
-* inspec - Contains a registry of all other resources to interact with the operating system or target in general.
-* skip\_resource - A resource may call this method to indicate that requirements aren't met. All tests that use this resource will be marked as skipped.
+<dl>
+  <dt>inspec</dt>
+  <dd>Contains a registry of all other resources to interact with the operating system or target in general.</dd>
+  <dt>skip_resource</dt>
+  <dd>A resource may call this method to indicate that requirements aren't met. All tests that use this resource will be marked as skipped.</dd>
+</dl>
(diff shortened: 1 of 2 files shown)


docs/dsl_resource.md
```diff
@@ -37,14 +37,25 @@ require a higher version.
...
+<dl>
+  <dt>name</dt>
+  <dd>Identifier of the resource (required)</dd>
+  <dt>desc</dt>
+  <dd>Description of the resource (optional)</dd>
+  <dt>example</dt>
+  <dd>Example usage of the resource (optional)</dd>
+  <dt>supports</dt>
+  <dd><small>(InSpec 2.0+)</small> Platform restrictions of the resource (optional)</dd>
+</dl>
```
kagarmoe: Our house style avoids HTML in Markdown. (At the same time, I am honestly impressed by your use of definition lists!)

On supports: We use platform for OS/AWS/Azure. This usage appears to be for InSpec's semver. I suggest:
- supports - InSpec version restrictions for this resource.

docs/dsl_resource.md
```diff
@@ -37,14 +37,25 @@ require a higher version.
...
 The following methods are available to the resource:
 
-* inspec - Contains a registry of all other resources to interact with the operating system or target in general.
-* skip\_resource - A resource may call this method to indicate, that requirements aren't met. All tests that use this resource will be marked as skipped.
+<dl>
+  <dt>inspec</dt>
+  <dd>Contains a registry of all other resources to interact with the operating system or target in general.</dd>
+  <dt>skip_resource</dt>
+  <dd>A resource may call this method to indicate that requirements aren't met. All tests that use this resource will be marked as skipped.</dd>
+</dl>
```
kagarmoe: Same as above: Our house style avoids HTML in Markdown docs.

kagarmoe (changes requested):
Thanks for your work! Our house style generally avoids HTML in Markdown. We use lists instead of definition lists. We use "Platform" for OS/AWS/Azure, see suggested wording.

kagarmoe (changes requested):
Thanks for your work! Our house style generally avoids HTML in Markdown. We use lists instead of definition lists. We use "Platform" for OS/AWS/Azure, see suggested wording.

docs/dsl_resource.md
TheLonelyGhost: Actually, `supports` is supposed to indicate what platform a resource is known to work with. Where a lot of resources are written to `skip_resource` under certain platform-specific circumstances (e.g., `etc_fstab` isn't going to work on Windows), this new feature simplifies that duplication of checking the platform so it's an attribute like `desc` or `example`.

I see where it was confusing about it representing the inspec version the resource supports. I was just trying to differentiate that it was a feature only supported in InSpec 2.x, not the 1.x line that many people are still working off of (if using ChefDK, for instance).

kagarmoe (approved):

TheLonelyGhost:
FWIW, I'm not quite sure why the Travis build is failing now. No source code changes occurred between the prior passing build and this failing build.

jquick (approved):
Thanks @TheLonelyGhost !

jquick:
Confirmed signoff is on the last commit.

PR #2831, opened 2018-03-15
Add Info about nginx_conf param

pwelch:
Add information about setting the path to the NGINX server configuration
file if it is not in the default path.


jquick (approved):
Thanks @pwelch !

jerryaldrichiii (approved):
Looks good to me, thanks @pwelch!

PR #2841, opened 2018-03-20
Fix typo in some docs

eramoto:
Also includes fixes of typo such as 
 - PostgreSQL
 - TCPMUX
 - auditctl


jquick (approved):
Thanks so much for helping with this!

PR #2842, opened 2018-03-20
Update audit_policy supports to 'windows'

### LINKED ISSUE #2829: [BUG] Wrong supports platform in  inspec/lib/resources/audit_policy.rb
### 🎛 Description

[BUG]
In the resource audit_policy there is the wrong supports platform set.
In the current files it states "supports platform: 'unix' but these test is designed for windows.

 inspec/lib/resources/audit_policy.rb

Version 2.0.47 and also i see this in the current git-hub release Version 2.0.49

### 🤔 Replication Case

If you run an Audit_Policy tests against a windows system with winRM the result is always "skipped"
[SKIP] Resource Audit_policy is not  supported on platform windows_server_2016_standard/10.....

### 💁 Possible Solutions

possible Solution: Change supports platform from 'unix' to 'windows'.
At least in my tests it works as i expected.
### 💻 Stacktrace

Trace not needed think,.

jquick:
This updates the supports from `unix` to `windows` for the audit_policy resource.

Fixes #2829 

### DIFF
--- a/lib/resources/audit_policy.rb
+++ b/lib/resources/audit_policy.rb
@@ -24,7 +24,7 @@
 module Inspec::Resources
   class AuditPolicy < Inspec.resource(1)
     name 'audit_policy'
-    supports platform: 'unix'
+    supports platform: 'windows'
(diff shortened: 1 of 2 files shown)


jerryaldrichiii (approved):
Good catch, I feel like this one was my fault somehow. Thanks for fixing!

jquick:
closing, this is a dupe of https://github.com/chef/inspec/pull/2850

PR #2845, opened 2018-03-20
The resource is login_defs not login_def

burtlo:
The main index is displaying the wrong resource name. The url to the resource
is not correct either. This should generate the correct entry and path in the
docs.


jquick (approved):
Thanks @burtlo

PR #2848, opened 2018-03-20
Remove supports binding for generic resources

jquick:
We ran into an issue were the AWS transport was trying to parse some YAML. This resource was previously locked down to only OS's. This change opens up the generic resources(yaml, csv, json) to be able to be used with any transport.

Fixes https://github.com/chef/inspec/issues/2846


jerryaldrichiii (approved):
Great catch @jquick, looks good to me.

arlimus (approved):
Thank you yay!! :confetti_ball:

PR #2850, opened 2018-03-20
Correct support platform for audit_policy

### LINKED ISSUE #2829: [BUG] Wrong supports platform in  inspec/lib/resources/audit_policy.rb

pwelch:
Audity Policy resource is only used on the windows platform and
was incorrectly set to unix.

Fixes #2829

Co-authored-by: Trevor Bramble <tbramble@chef.io>
Co-authored-by: Paul Welch <pwelch@chef.io>

### DIFF
--- a/lib/resources/audit_policy.rb
+++ b/lib/resources/audit_policy.rb
@@ -24,7 +24,7 @@
 module Inspec::Resources
   class AuditPolicy < Inspec.resource(1)
     name 'audit_policy'
-    supports platform: 'unix'
+    supports platform: 'windows'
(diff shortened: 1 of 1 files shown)


jquick (approved):
Thanks @pwelch !

miah (approved):

PR #2854, opened 2018-03-21
Add unf_ext configuration to omnibus to allow building against our forked unf_ext.

miah:
Installing the InSpec gem on Solaris is currently failing, because of a new dependency on unf_ext which was brought in when we added Azure to Train. We have submitted a PR upstream to get unf_ext fixed, but until then hope to provide a solution through omnibus. When upstream is merged we can remove our fork, and undo this change but still provide omnibus for Solaris and AIX.

Unfortunately, we do not currently release omnibus builds of InSpec for Solaris or AIX. I have another PR opened to add Solaris and AIX to our InSpec omnibus build matrix.

Until https://github.com/knu/ruby-unf_ext/pull/39 is merged we have a fork of unf_ext living at https://github.com/jquick/ruby-unf_ext.

### DIFF
--- a/omnibus/config/software/inspec.rb
+++ b/omnibus/config/software/inspec.rb
@@ -24,6 +24,7 @@ dependency 'rubygems'
 dependency 'bundler'
 dependency 'rb-readline'
 dependency 'appbundler'
+dependency 'unf_ext'
 
 license :project_license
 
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @miah !

scotthain (approved):

PR #2866, opened 2018-03-22
new resource: aws rds instance (singular)

HackerShark:
Porting over the singular rds resource from the aws-inspec git repo

Signed-off-by: HackerShark <melsharkawi@mitre.org>


jquick (approved):
Thanks @HackerShark !

test/integration/aws/default/build/rds.tf
```diff
@@ -0,0 +1,25 @@
+# Contains resources and outputs related to testing the aws_rds_* resources.
+
+#======================================================#
+#                    RDS Instances
+#======================================================#
+
+
+resource "aws_db_instance" "default" {
```
clintoncwolfe, line 8: This is just a preference, but I like to name test fixtures with a more useful name - perhaps "alpha".  I just don't like the word "default", when there isn't anything default about it.

test/integration/aws/default/build/rds.tf
```diff
@@ -0,0 +1,25 @@
...
+  allocated_storage    = 20
+  storage_type         = "gp2"
+  engine               = "mysql"
+  engine_version       = "5.6.37"
+  instance_class       = "db.t2.micro"
+  identifier           = "test-instance-id"
```
clintoncwolfe, line 14: To avoid testing collisions, we need to prefix identifiers.  Typically we use ${terraform.env}.

test/integration/aws/default/build/rds.tf
```diff
@@ -0,0 +1,25 @@
...
+  name                 = "test_instance"
+  username             = "testuser"
+  password             = "testpassword"
+  parameter_group_name = "default.mysql5.6"
+  skip_final_snapshot  = true
+}
+
+output "rds_db_instance_id" {
```
clintoncwolfe, line 22: And then you could name this rds_db_instance_id_alpha, which future-proofs it against when we need to have a second fixture with different settings... but yeah, it's just a preference.

clintoncwolfe (approved):
This is a great contribution!  My only real concern is making the name of the test DB a bit more unique, using terraform.env .  But that is not something that should block merging, unless you want to.  Thanks!

clintoncwolfe:
Hey, my fears of long integration tests with RDS were unfounded, for now - with a tiny RDS instance, we went from 5 min => 8 min.  That's tolerable.

PR #2871, opened 2018-03-23
registry_key resource was returning an incorrect value

omar-irizarry:
When key value was greater than 2147483647 the registry_key resource was returning an incorrect value failing the test even though the expected value and the actual value in the registry were the same.


jquick:
Hey @omar-irizarry ! Thanks for your first contribution! Welcome to the family.

This looks great, always enjoy enhancing the windows resources.

jquick (approved):
Thanks for this!

jerryaldrichiii (approved):
Looks good to me! Great catch on the `(default)` handling!

PR #2872, opened 2018-03-23
Mitigate trivial warning output on test

eramoto:
To be easier to find out any problem when testing, mitigates that trivial warnings are output due to initializing arguments and removing redundant methods. 

Changes to mitigate each warning on this PR are described in detail below:

* Sets the initial value to the variables in initialize method.
  ```
  lib/bundles/inspec-habitat/profile.rb:290: warning: instance variable @cli_config not initialized
  lib/inspec/profile.rb:143: warning: instance variable @supports_platform not initialized
  lib/inspec/profile.rb:150: warning: instance variable @supports_runtime not initialized
  lib/inspec/profile_context.rb:169: warning: instance variable @current_load not initialized
  lib/resources/http.rb:130: warning: instance variable @response not initialized
  lib/resources/http.rb:177: warning: instance variable @ran_curl not initialized
  lib/utils/filter.rb:236: warning: instance variable @resource not initialized
  lib/utils/filter.rb:244: warning: instance variable @resource not initialized
  ```

* Fixes passing a ambiguous argument.
  ```
  test/unit/reporters/cli_test.rb:233: warning: ambiguous first argument; put parentheses or a space even after `/' operator
  ```

* Since `has_data_disks` and `count` method are defined as method in azure_virtual_machine_data_disk.rb after being added as accessor, removes the accessors
  ```
  lib/resources/azure/azure_virtual_machine_data_disk.rb:66: warning: method redefined; discarding old has_data_disks?
  lib/utils/filter.rb:223: warning: previous definition of has_data_disks? was here
  lib/resources/azure/azure_virtual_machine_data_disk.rb:71: warning: method redefined; discarding old count
  lib/utils/filter.rb:223: warning: previous definition of count was here 
  ```

* Since `openssl_version/compiler_info/support_info` method are redefined as unique method after being defined by define_method method, removes previous definitions.
  ```
  lib/resources/nginx.rb:43: warning: method redefined; discarding old openssl_version
  lib/resources/nginx.rb:38: warning: previous definition of openssl_version was here
  lib/resources/nginx.rb:48: warning: method redefined; discarding old compiler_info 
  lib/resources/nginx.rb:38: warning: previous definition of compiler_info was here
  lib/resources/nginx.rb:53: warning: method redefined; discarding old support_info
  lib/resources/nginx.rb:38: warning: previous definition of support_info was here
  ```

* Since `content` method is defined as method after being defined as attribute reader, removes the reader.
  ```
  lib/resources/os_env.rb:39: warning: method redefined; discarding old content
  ```

* Since `setup` method is multiply defined in the same class name, organizes to remove them.
  ```
  test/unit/resources/aws_s3_bucket_test.rb:56: warning: method redefined; discarding old setup
  test/unit/resources/aws_s3_bucket_test.rb:36: warning: previous definition of setup was here
  test/unit/resources/aws_s3_bucket_test.rb:153: warning: method redefined; discarding old setup
  test/unit/resources/aws_s3_bucket_test.rb:56: warning: previous definition of setup was here 
  ```

* Since `test_be_public_public_acl` method is defined twice and is the same in content, removes one.
  ```
  test/unit/resources/aws_s3_bucket_test.rb:166: warning: method redefined; discarding old test_be_public_public_acl
  test/unit/resources/aws_s3_bucket_test.rb:157: warning: previous definition of test_be_public_public_acl was here
  ```

### DIFF
--- a/lib/resources/azure/azure_virtual_machine_data_disk.rb
+++ b/lib/resources/azure/azure_virtual_machine_data_disk.rb
@@ -17,8 +17,6 @@ module Inspec::Resources
     filter = FilterTable.create
     filter.add_accessor(:where)
           .add_accessor(:entries)
-          .add_accessor(:has_data_disks?)
-          .add_accessor(:count)
           .add(:exists?) { |x| !x.entries.empty? }
           .add(:disk, field: :disk)
           .add(:number, field: :number)
--- a/lib/resources/nginx.rb
+++ b/lib/resources/nginx.rb
@@ -34,7 +34,7 @@ module Inspec::Resources
       read_content
     end
 
-    %w{compiler_info error_log_path http_client_body_temp_path http_fastcgi_temp_path http_log_path http_proxy_temp_path http_scgi_temp_path http_uwsgi_temp_path lock_path modules_path openssl_version prefix sbin_path service support_info version}.each do |property|
+    %w{error_log_path http_client_body_temp_path http_fastcgi_temp_path http_log_path http_proxy_temp_path http_scgi_temp_path http_uwsgi_temp_path lock_path modules_path prefix sbin_path service version}.each do |property|
       define_method(property.to_sym) do
         @params[property.to_sym]
       end
--- a/test/unit/resources/aws_s3_bucket_test.rb
+++ b/test/unit/resources/aws_s3_bucket_test.rb
@@ -29,14 +29,14 @@ class AwsS3BucketConstructor < Minitest::Test
   end
 end
 
-#=============================================================================#
-#                               Search / Recall
-#=============================================================================#
 class AwsS3BucketPropertiesTest < Minitest::Test
   def setup
     AwsS3Bucket::BackendFactory.select(AwsMSBSB::Basic)
   end
 
+  #===========================================================================#
+  #                               Search / Recall
+  #===========================================================================#
   def test_recall_no_match_is_no_exception
     refute AwsS3Bucket.new('NonExistentBucket').exists?
   end
@@ -46,17 +46,10 @@ class AwsS3BucketPropertiesTest < Minitest::Test
   end
 
   # No need to handle multiple hits; S3 bucket names are globally unique.
-end
 
-#=============================================================================#
-#                               Properties
-#=============================================================================#
-
-class AwsS3BucketPropertiesTest < Minitest::Test
-  def setup
-    AwsS3Bucket::BackendFactory.select(AwsMSBSB::Basic)
-  end
-
+  #===========================================================================#
+  #                               Properties
+  #===========================================================================#
   #---------------------Bucket Name----------------------------#  
   def test_property_bucket_name
     assert_equal('public', AwsS3Bucket.new('public').bucket_name)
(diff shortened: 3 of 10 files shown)


lib/resources/nginx.rb
```diff
@@ -34,7 +34,7 @@ def initialize(nginx_path = '/usr/sbin/nginx')
       read_content
     end
 
-    %w{compiler_info error_log_path http_client_body_temp_path http_fastcgi_temp_path http_log_path http_proxy_temp_path http_scgi_temp_path http_uwsgi_temp_path lock_path modules_path openssl_version prefix sbin_path service support_info version}.each do |property|
+    %w{error_log_path http_client_body_temp_path http_fastcgi_temp_path http_log_path http_proxy_temp_path http_scgi_temp_path http_uwsgi_temp_path lock_path modules_path prefix sbin_path service version}.each do |property|
```
jquick, line 37: Same here, why are removing these properties?

lib/resources/azure/azure_virtual_machine_data_disk.rb
```diff
@@ -17,8 +17,6 @@ class AzureVirtualMachineDataDisk < AzureResourceBase
     filter = FilterTable.create
     filter.add_accessor(:where)
           .add_accessor(:entries)
-          .add_accessor(:has_data_disks?)
-          .add_accessor(:count)
```
jquick, line 21: Can you explain why we are removing these?

jquick (commented):
Thanks @eramoto ! Really like cleaning up test output. A few questions just to clarify if you don't mind.

lib/resources/azure/azure_virtual_machine_data_disk.rb
clintoncwolfe: Looks like they are also defined as methods - see lines 64 and 69.  Adding them as accessors would indeed trigger a method redefined warning.

lib/resources/nginx.rb
clintoncwolfe: compiler info is defined as a method on line 48; openssl_version on 43, and support_info on 53.

test/unit/resources/aws_s3_bucket_test.rb
```diff
@@ -29,14 +29,14 @@ def test_constructor_reject_unknown_resource_params
   end
 end
 
-#=============================================================================#
```
clintoncwolfe, line 32: This re-organization isn't quite how we like to do it.  Usually we have a test class for each major section - so Search/ Recall would be a class, and Properties would be a class as well.

jquick (approved):
Thanks @eramoto !

clintoncwolfe (commented):
I like the changes - they certainly quiet down some of the noisy unit tests!  Thanks for contributing.  It might be helpful to explain in your PR description a bit more about what you're doing (deleting attribute readers was a bit alarming!  But they were all benign, after checking.).  My only nitpick is that the test file organization for aws_s3_bucket isn't usual, but I don't think that should block merging.  Thanks for contributing!

eramoto:
Thanks for your helpful feedback @clintoncwolfe .

I agree with it (I am not familiar with InSpec's policy, so I just organized to remove the target methods).
If you want, I will try to create individual class for them instead of this fix in aws_s3_bucket_test.rb.

clintoncwolfe (approved):

PR #2875, opened 2018-03-24
Fixes the YAML documentation

burtlo:
* Updated the initial example and tried to give it a little more real context.
* Fixes the inline content example. 
* Fixes the incorrect usage below when parsing the kitchen configuration file.

### DIFF
--- a/docs/resources/yaml.md.erb
+++ b/docs/resources/yaml.md.erb
@@ -40,7 +40,7 @@ Like the `json` resource, the `yaml` resource can read a file, run a command, or
       its('state') { should eq 'open' }
     end
 
-    describe yaml({ content: \"key1: value1\nkey2: value2\" }) do
+    describe yaml({ content: "\"key1: value1\nkey2: value2\"" }) do
       its('key2') { should cmp 'value2' }
     end
 
@@ -53,7 +53,7 @@ The following examples show how to use this InSpec audit resource.
 ### Test a kitchen.yml file driver
 
     describe yaml('.kitchen.yaml') do
-      its('driver.name') { should eq('vagrant') }
+      its(['driver','name']) { should eq('vagrant') }
     end
 
 <br>


docs/resources/yaml.md.erb
```diff
@@ -11,36 +11,36 @@ Use the `yaml` InSpec audit resource to test configuration data in a Yaml file.
 
 ## Syntax
 
-A `yaml` resource block declares the configuration data to be tested. Assume the following Yaml file:
+A `yaml` resource block declares the configuration data to be tested. Assume the following Yaml file called 'filename.yml':
```
kagarmoe: Keep "Assume the following YAML file:" -- the  filename is part of what is assumed.

docs/resources/yaml.md.erb
```diff
@@ -11,36 +11,36 @@ Use the `yaml` InSpec audit resource to test configuration data in a Yaml file.
...
 
-    name: foo
-    array:
-      - zero
-      - one
+    name: myservice
+    hosts:
+      - ip1.host.com
+      - ip2.host.com
```
kagarmoe: The purpose of this example to is set up an array and then demonstrate how to query the array. This is one of the places for using the conventions `foo, bar, baz`, `tmp\example`, and `example.com`.

docs/resources/yaml.md.erb
```diff
@@ -11,36 +11,36 @@ Use the `yaml` InSpec audit resource to test configuration data in a Yaml file.
...
-      its('name') { should eq 'foo' }
-      its(['array', 1]) { should eq 'one' }
+      its('name') { should eq 'myservice' }
+      its(['hosts', 0]) { should eq 'ip1.host.com' }
     end
 
 where
 
-* `name` is a configuration setting in a Yaml file
-* `should eq 'foo'` tests a value of `name` as read from a Yaml file versus the value declared in the test
```
kagarmoe: Again, let's stay with the simpler example. However, the second bullet point could be: 

-* `should eq 'foo'` tests the value of `name` as read from a Yaml file against the value declared in the test.

docs/resources/yaml.md.erb
```diff
@@ -11,36 +11,36 @@ Use the `yaml` InSpec audit resource to test configuration data in a Yaml file.
...
+* `name` retrieves the 'myservice' value from file. That value is compared against the expressed expected value of 'myservice'.
+* `['hosts, 0']` retrieves the first element from the `hosts` array. That value is compared against the expressed expected value of 'ip1.host.com'.
 
 Like the `json` resource, the `yaml` resource can read a file, run a command, or accept content inline:
 
-    describe yaml('config.yaml') do
-      its(['driver', 'name']) { should eq 'vagrant' }
+    describe yaml('/etc/cloud/cloud.cfg') do
+      its('ssh_pwauth') { should eq 0 }
     end
```
kagarmoe: Keep the simpler example. We don't want to send novices to look up `ssh_pwauth`.

docs/resources/yaml.md.erb
```diff
@@ -11,36 +11,36 @@ Use the `yaml` InSpec audit resource to test configuration data in a Yaml file.
...
 
     describe yaml({ command: 'retrieve_data.py --yaml' }) do
       its('state') { should eq 'open' }
     end
 
-    describe yaml({ content: \"key1: value1\nkey2: value2\" }) do
+    describe yaml({ content: "\"key1: value1\nkey2: value2\"" }) do
```
kagarmoe, line 43: Thanks!

docs/resources/yaml.md.erb
```diff
@@ -53,7 +53,7 @@ The following examples show how to use this InSpec audit resource.
 ### Test a kitchen.yml file driver
 
     describe yaml('.kitchen.yaml') do
-      its('driver.name') { should eq('vagrant') }
+      its(['driver','name']) { should eq('vagrant') }
```
kagarmoe, line 56: Thanks!

kagarmoe (changes requested):
Thanks for your suggestions! The specificity of your thought in the examples you laid out runs somewhat counter to the purpose of the basic examples in this document. The general level resources start out with general examples that serve as description and instruction for using the resource. These are sometimes followed by one or two specific uses, such as the  kitchen file driver example. Thanks for fixing the syntax there!

docs/resources/yaml.md.erb
burtlo: `config.yaml` is not right then. It should read `.kitchen.yml`. But then I am asking myself the question - why would I want to verify the kitchen configuration with InSpec?

docs/resources/yaml.md.erb
burtlo: The file name is not mentioned anywhere and then one is used below in the example. I thought it would be useful to see where the filename came from but fine with changing it back.

docs/resources/yaml.md.erb
burtlo: Really this and the next change are the ones I care about. I'll update the commit to include these two things.

kagarmoe (approved):

jquick (approved):
Thanks @burtlo !

PR #2883, opened 2018-03-27
Pw/pip windows bug

### LINKED ISSUE #2855: Pip inspec resource fails on Windows when there is a new version of Pip available
### 🎛 Description
[BUG]
When using the Pip resource on Windows the check fails when a new version of Pip is available.

$ chef --version
Chef Development Kit Version: 2.5.3
chef-client version: 13.8.5
delivery version: master (73ebb72a6c42b3d2ff5370c476be800fee7e5427)
berks version: 6.3.1
kitchen version: 1.20.0
inspec version: 1.51.21

Running inspec using kitchen.

### 🤔 Replication Case

Have a Windows 2016 Core box with an out-of-date Pip version. Let's say 9.0.1 and the latest version is 9.0.2.

Have a test that verifies if a Pip packages is installed, like the one in the examples for Jinja2

### 💁 Possible Solutions

https://github.com/chef/inspec/blob/master/lib/resources/pip.rb#L40

The command Inspec runs is `pip show [package_name]`. Inspec will parse the outcome of that command in key/value pairs based on a regex. It seems that is silently fails parsing the info due to the `You are using pip version 9.0.1, however version 9.0.2 is available.` message.

pwelch:
When pip on windows has a newer version available it spits out a stderr message along with the results of the `pip show` command. 

This adds logic to ignore the pip message that there is a newer version. 

Fixes #2855

### DIFF
--- a/lib/resources/pip.rb
+++ b/lib/resources/pip.rb
@@ -27,7 +27,6 @@ module Inspec::Resources
       @package_name = package_name
       @pip_cmd = pip_path || default_pip_path
 
-      return skip_resource 'python not found' if paths['Python'].nil?
       return skip_resource 'pip not found' if @pip_cmd.nil?
     end
 
@@ -68,7 +67,7 @@ module Inspec::Resources
     end
 
     def cmd_successful?
-      result = inspec.command("#{@pip_cmd} show #{@package_name}")
+      result = cmd
       return true if result.exit_status == 0
 
       if result.exit_status != 0
@@ -84,15 +83,15 @@ module Inspec::Resources
       false
     end
 
-    # Paths of Python and Pip
+    # Paths of Python and Pip on windows
     # {"Pip" => nil, "Python" => "/path/to/python"}
     #
-    # @return [Hash] of paths
-    def paths
-      return @__paths if @__paths
+    # @return [Hash] of windows_paths
+    def windows_paths
+      return @__windows_paths if @__windows_paths
       cmd = inspec.command('New-Object -Type PSObject | Add-Member -MemberType NoteProperty -Name Pip -Value (Invoke-Command -ScriptBlock {where.exe pip}) -PassThru | Add-Member -MemberType NoteProperty -Name Python -Value (Invoke-Command -ScriptBlock {where.exe python}) -PassThru | ConvertTo-Json')
 
-      @__paths = JSON.parse(cmd.stdout)
+      @__windows_paths = JSON.parse(cmd.stdout)
     end
 
     # Default path of python pip installation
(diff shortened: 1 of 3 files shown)


lib/resources/pip.rb
```diff
@@ -62,18 +63,52 @@ def to_s
 
     private
 
+    def cmd
+      @__cmd ||= inspec.command("#{@pip_cmd} show #{@package_name}")
+    end
+
+    def cmd_successful?
+      result = inspec.command("#{@pip_cmd} show #{@package_name}")
```
jquick: Could we just use `cmd` here instead of doing the `inspec.command` again?

lib/resources/pip.rb
```diff
@@ -62,18 +63,52 @@ def to_s
...
+      end
+
+      false
+    end
+
+    # Paths of Python and Pip
+    # {"Pip" => nil, "Python" => "/path/to/python"}
+    #
+    # @return [Hash] of paths
+    def paths
```
jquick: You mind updating this to `windows_paths`. Since its windows only command it may read better.

jquick (commented):
Really nice work here @pwelch ! Just a few comments.

lib/resources/pip.rb
pwelch: @jquick good catch! Thanks!

lib/resources/pip.rb
pwelch: Fixed

lib/resources/pip.rb
```diff
@@ -26,16 +26,16 @@ class PipPackage < Inspec.resource(1)
...
     end
 
     def info
       return @info if defined?(@info)
 
       @info = {}
       @info[:type] = 'pip'
-      cmd = inspec.command("#{@pip_cmd} show #{@package_name}")
-      return @info if cmd.exit_status != 0
+      return @info if !cmd_successful?
```
TrevorBramble:
```ruby
return @info unless cmd_successful?
```

lib/resources/pip.rb
```diff
@@ -62,19 +62,56 @@ def to_s
 
     private
 
+    def cmd
+      @__cmd ||= inspec.command("#{@pip_cmd} show #{@package_name}")
+    end
+
+    def cmd_successful?
+      result = cmd
```
TrevorBramble: Seems like every instance of `result` in this method could just be `cmd`? (And would actually read better that way?)

lib/resources/pip.rb
```diff
@@ -62,19 +62,56 @@ def to_s
...
+      false
+    end
+
+    # Paths of Python and Pip on windows
+    # {"Pip" => nil, "Python" => "/path/to/python"}
+    #
+    # @return [Hash] of windows_paths
+    def windows_paths
+      return @__windows_paths if @__windows_paths
+      cmd = inspec.command('New-Object -Type PSObject | Add-Member -MemberType NoteProperty -Name Pip -Value (Invoke-Command -ScriptBlock {where.exe pip}) -PassThru | Add-Member -MemberType NoteProperty -Name Python -Value (Invoke-Command -ScriptBlock {where.exe python}) -PassThru | ConvertTo-Json')
```
TrevorBramble: Can we break this out onto multiple lines?
```ruby
cmd = inspec.command(
  'New-Object -Type PSObject' \
  ' | Add-Member -MemberType NoteProperty -Name Pip -Value (Invoke-Command -ScriptBlock {where.exe pip}) -PassThru' \
  ' | Add-Member -MemberType NoteProperty -Name Python -Value (Invoke-Command -ScriptBlock {where.exe python}) -PassThru' \
  ' | ConvertTo-Json')
```
or better...
```ruby
cmd = inspec.command([
  'New-Object -Type PSObject',
  'Add-Member -MemberType NoteProperty -Name Pip -Value (Invoke-Command -ScriptBlock {where.exe pip}) -PassThru',
  'Add-Member -MemberType NoteProperty -Name Python -Value (Invoke-Command -ScriptBlock {where.exe python}) -PassThru',
  'ConvertTo-Json'
].join(' | '))

TrevorBramble (commented):
Some nitpicks

pwelch:
@TrevorBramble changes for feedback applied.

jquick (approved):
Thanks @pwelch !

TrevorBramble (approved):

jquick:
Confirmed commit was signed. DCO bot is bugged.

PR #2885, opened 2018-03-27
Change route_table_id Regular Expression for correctness

### LINKED ISSUE #2741: aws_route_table resource should validate route table ID lengths properly
### Description

As reported by @burtlo on #2740 , `aws_route_table` allows `rtb-` IDs of arbitrary length, when they should be limited to 8 hex chars.  Be sure to check that limit in AWS docs; many IDs are being lengthened (e.g. EC2 instance IDs); but even the long identifiers have a known length.

InSpec 2.0.17, Mac 10.13.13

### Replication Case

Create a control file that references a 9-digit aws_route_table ID.  When running it, InSpec will report it as missing; instead, it should fail the control with an ID validation failure error message.

TrevorBramble:
Without the terminating character ($), it just accepted any characters
at all after the initial matching set.

Also add some tests to assure we're raising appropriately.

Fixes #2741.

Co-authored-by: Trevor Bramble <tbramble@chef.io>
Co-authored-by: Joshua Padgett <jpadgett@chef.io>

Signed-off-by: Trevor Bramble <tbramble@chef.io>

### DIFF
--- a/lib/resources/aws/aws_route_table.rb
+++ b/lib/resources/aws/aws_route_table.rb
@@ -26,9 +26,11 @@ class AwsRouteTable < Inspec.resource(1)
       allowed_scalar_type: String,
     )
 
-    if validated_params.key?(:route_table_id) && validated_params[:route_table_id] !~ /^rtb\-[0-9a-f]{8}/
-      raise ArgumentError, 'aws_route_table Route Table ID must be in the' \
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
Looks great - Thanks!

jquick (approved):
Thanks @TrevorBramble !

PR #2891, opened 2018-03-28
Require a key attribute for the key_rsa resource

### LINKED ISSUE #2865: Defining an attribute without a default value generates a stacktrace
### 🎛 Description

This a 🐛 bug report

Defining a profile attribute without a default value results in a stacktrace that tells the user it cannot convert the InSpec::Attribute::DEFAULT_ATTRIBUTE to String.

InSpec 2.1.0
Workstation: Mac 10.13.3

### 🤔 Replication Case

Define an attribute without a default value that you plan to use within a control that expects a String.

```ruby
pass = attribute('pass', description: 'passphrase for the secret-key')

describe key_rsa('/home/chef/.ssh/secret-key',pass) do
  it { should be_private }
end
```

Execute a scan:

```shell
$ inspec exec myprofile --sudo -t ssh://username:password@target
/opt/inspec/embedded/lib/ruby/gems/2.4.0/gems/inspect-2.1.0/lib/resources/key_rsa.rb:32:in `read': can't convert InSpec::Attribute::DEFAULT_ATTRIBUTE to String (InSpec::Attribute::DEFAULT_ATTRIBUTE#to_str gives InSpec::Attribute::DEFAULT_ATTRIBUTE) (TypeError)
```

### 💁 Possible Solutions

* Fail the run without generating a stacktrace.

The control should fail but not fail the run.

* Inform the user that they need to provide a default value

* Prompt the user to enter a default value at run-time.

### 💻 Stacktrace

SEE ABOVE

omar-irizarry:
Checking that passphrase is not of type Attribute::DEFAULT_ATTRIBUTE before calling OpenSSL::PKey.read.
Also added a catch to exception OpenSSL::PKey::PKeyError to prevent a stack trace error if the passphrase was incorrect.

### DIFF
--- a/lib/resources/key_rsa.rb
+++ b/lib/resources/key_rsa.rb
@@ -28,7 +28,6 @@ module Inspec::Resources
     def initialize(keypath, passphrase = nil)
       @key_path = keypath
       @passphrase = passphrase
-      default_attribute?(passphrase)
       @key = read_pkey(read_file_content(@key_path, allow_empty: true), @passphrase)
     end
 
--- a/lib/resources/key_rsa.rb
+++ b/lib/resources/key_rsa.rb
@@ -3,6 +3,7 @@
 require 'openssl'
 require 'hashie/mash'
 require 'utils/file_reader'
+require 'utils/pkey_reader'
 
 module Inspec::Resources
   class RsaKey < Inspec.resource(1)
@@ -22,18 +23,13 @@ module Inspec::Resources
     "
 
     include FileReader
+    include PkeyReader
 
     def initialize(keypath, passphrase = nil)
       @key_path = keypath
       @passphrase = passphrase
-      if @passphrase.is_a? Inspec::Attribute::DEFAULT_ATTRIBUTE
-        return fail_resource 'Please provide default value for attribute'
-      end
-      begin
-        @key = OpenSSL::PKey.read(read_file_content(@key_path, allow_empty: true), @passphrase)
-      rescue OpenSSL::PKey::PKeyError
-        return fail_resource 'passphrase Error'
-      end
+      default_attribute?(passphrase)
+      @key = read_pkey(read_file_content(@key_path, allow_empty: true), @passphrase)
     end
 
     def public?
--- a/lib/resources/key_rsa.rb
+++ b/lib/resources/key_rsa.rb
@@ -26,7 +26,14 @@ module Inspec::Resources
     def initialize(keypath, passphrase = nil)
       @key_path = keypath
       @passphrase = passphrase
-      @key = OpenSSL::PKey.read(read_file_content(@key_path, allow_empty: true), @passphrase)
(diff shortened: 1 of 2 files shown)


jquick (changes requested):
Thanks @omar-irizarry ! This is a really nice addition. Do you mind adding a unit test for this case?

lib/resources/key_rsa.rb
```diff
@@ -26,7 +26,14 @@ class RsaKey < Inspec.resource(1)
     def initialize(keypath, passphrase = nil)
       @key_path = keypath
       @passphrase = passphrase
-      @key = OpenSSL::PKey.read(read_file_content(@key_path, allow_empty: true), @passphrase)
+      if @passphrase.is_a? Inspec::Attribute::DEFAULT_ATTRIBUTE
+        return fail_resource 'Please provide default value for attribute'
+      end
+      begin
```
chris-rock: We should also consider moving the logic out of the initialize method, so that we make sure it is a control error and its not happening during initialization.

lib/utils/pkey_reader.rb
```diff
@@ -0,0 +1,17 @@
+module PkeyReader
+  def read_pkey(filecontent, passphrase)
+    default_attribute?(passphrase)
+    begin
+      key = OpenSSL::PKey.read(filecontent, passphrase)
+    rescue OpenSSL::PKey::PKeyError
+      raise Inspec::Exceptions::ResourceFailed, 'passphrase Error'
+    end
+    key
+  end
```
TrevorBramble: You could restructure this to clarify the flow like this:

```ruby
def read_pkey(filecontent, passphrase)
  default_attribute?(passphrase)

  OpenSSL::PKey.read(filecontent, passphrase)
rescue OpenSSL::PKey::PKeyError
  raise Inspec::Exceptions::ResourceFailed, 'passphrase error'
end
```

lib/utils/pkey_reader.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def default_attribute?(passphrase)
+    if passphrase.is_a? Inspec::Attribute::DEFAULT_ATTRIBUTE
+      raise Inspec::Exceptions::ResourceFailed, 'Please provide default value for attribute'
+    end
+  end
```
TrevorBramble: Predicate methods (`foo?`) are expected to return `true`/`false`. I think it's right to pull out the test and exception raising into another method that can be used as a guard statement in `read_pkey()`. So maybe this method could be renamed `ensure_not_default` or `raise_if_default` or similar instead?

TrevorBramble (changes requested):
Thanks for your contribution! Have some small changes I'd like to see before merging. =^)

TrevorBramble (approved):
Thank you!

jquick (approved):
Thanks!

PR #2892, opened 2018-03-28
Add AWS hardware MFA matcher

### LINKED ISSUE #2750: Add hardware MFA matchers to aws_iam_root_user
### 🎛 Description

🙋 feature request

See https://github.com/chef/inspec-aws/pull/124 - add support for _hardware_ MFA detection.  Ensure that we have matchers for "any" mfa, virtual MFA, and hardware MFA.  A positive integration test for this is out of scope.

pwelch:
Adding a hardware as well as a virtual MFA matcher for aws_iam_root_user
resource

Fixes #2750

Signed-off-by: Paul Welch <pwelch@chef.io>

### DIFF
--- a/lib/resources/aws/aws_iam_root_user.rb
+++ b/lib/resources/aws/aws_iam_root_user.rb
@@ -46,8 +46,10 @@ class AwsIamRootUser < Inspec.resource(1)
     summary_account['AccountMFAEnabled'] == 1
   end
 
-  def has_virtual_mfa_devices?
+  def has_virtual_mfa_enabled?
     virtual_mfa_devices.each do |device|
+      # if the root account has a Virtual MFA device then
+      # it will have a special serial number ending in 'root-account-mfa-device'
       if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
         device['serial_number']
         return true
@@ -56,8 +58,8 @@ class AwsIamRootUser < Inspec.resource(1)
     false
   end
 
-  def has_hardware_mfa_devices?
-    @__hardware_devices ||= !@client.list_mfa_devices.mfa_devices.empty?
+  def has_hardware_mfa_enabled?
+    has_mfa_enabled? && !has_virtual_mfa_enabled?
   end
 
   def to_s
@@ -73,6 +75,8 @@ class AwsIamRootUser < Inspec.resource(1)
   end
 
   def virtual_mfa_devices
-    @__virtual_devices ||= @client.list_virtual_mfa_devices.virtual_mfa_devices
+    catch_aws_errors do
+      @__virtual_devices ||= @client.list_virtual_mfa_devices.virtual_mfa_devices
+    end
   end
 end
--- a/lib/resources/aws/aws_iam_root_user.rb
+++ b/lib/resources/aws/aws_iam_root_user.rb
@@ -46,6 +46,20 @@ class AwsIamRootUser < Inspec.resource(1)
     summary_account['AccountMFAEnabled'] == 1
   end
 
+  def has_virtual_mfa_devices?
+    virtual_mfa_devices.each do |device|
+      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
+        device['serial_number']
+        return true
+      end
+    end
+    false
+  end
+
+  def has_hardware_mfa_devices?
+    @__hardware_devices ||= !@client.list_mfa_devices.mfa_devices.empty?
+  end
+
   def to_s
     'AWS Root-User'
   end
@@ -56,5 +70,9 @@ class AwsIamRootUser < Inspec.resource(1)
     catch_aws_errors do
       @summary_account ||= @client.get_account_summary.summary_map
     end
+  end
+
+  def virtual_mfa_devices
+    @__virtual_devices ||= @client.list_virtual_mfa_devices.virtual_mfa_devices
   end
 end
--- a/lib/resources/aws/aws_iam_root_user.rb
+++ b/lib/resources/aws/aws_iam_root_user.rb
@@ -46,16 +46,12 @@ class AwsIamRootUser < Inspec.resource(1)
     summary_account['AccountMFAEnabled'] == 1
   end
 
+  # if the root account has a Virtual MFA device then it will have a special
+  # serial number ending in 'root-account-mfa-device'
   def has_virtual_mfa_enabled?
-    virtual_mfa_devices.each do |device|
-      # if the root account has a Virtual MFA device then
-      # it will have a special serial number ending in 'root-account-mfa-device'
-      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
-        device['serial_number']
-        return true
-      end
-    end
-    false
+    mfa_device_pattern = %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device}
+
+    virtual_mfa_devices.any? { |d| mfa_device_pattern =~ d['serial_number'] }
   end
 
   def has_hardware_mfa_enabled?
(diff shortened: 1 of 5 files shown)


test/unit/resources/aws_iam_root_user_test.rb
```diff
@@ -44,4 +44,52 @@ def test_has_mfa_enabled_returns_false_when_account_mfa_devices_is_zero
 
     assert_equal false, AwsIamRootUser.new(@mock_conn).has_mfa_enabled?
   end
-end
+
+  def test_has_virtual_mfa_enabled_returns_true_when_account_vmfa_devices_is_one
```
clintoncwolfe: :100: for meaningful unit test names

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -46,6 +46,20 @@ def has_mfa_enabled?
     summary_account['AccountMFAEnabled'] == 1
   end
 
+  def has_virtual_mfa_devices?
```
clintoncwolfe: So, it turns out that even though the API returns an array, you can only have one MFA device total per root account or IAM user (see [FAQ](https://aws.amazon.com/iam/faqs/#MFA_FAQs), 'Q. Can I have multiple authentication devices active for my AWS account?').  

Based on that, I think we should rename this matcher to be `has_virtual_mfa_enabled?`.

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -46,6 +46,20 @@ def has_mfa_enabled?
...
+    virtual_mfa_devices.each do |device|
+      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
+        device['serial_number']
+        return true
+      end
+    end
+    false
+  end
+
+  def has_hardware_mfa_devices?
```
clintoncwolfe: Likewise, this should be renamed `has_hardware_mfa_enabled?`.

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -46,6 +46,20 @@ def has_mfa_enabled?
     summary_account['AccountMFAEnabled'] == 1
   end
 
+  def has_virtual_mfa_devices?
+    virtual_mfa_devices.each do |device|
+      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
```
clintoncwolfe: Might want to add a comment here - if the root account has a VMFA, it will have this special serial number, ending in 'root-account-mfa-device'.

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -57,4 +71,8 @@ def summary_account
       @summary_account ||= @client.get_account_summary.summary_map
     end
   end
+
+  def virtual_mfa_devices
+    @__virtual_devices ||= @client.list_virtual_mfa_devices.virtual_mfa_devices
```
clintoncwolfe: Need to wrap this in a `catch_aws_errors` block.

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -46,6 +46,20 @@ def has_mfa_enabled?
...
+      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
+        device['serial_number']
+        return true
+      end
+    end
+    false
+  end
+
+  def has_hardware_mfa_devices?
+    @__hardware_devices ||= !@client.list_mfa_devices.mfa_devices.empty?
```
clintoncwolfe: Alas, this isn't the right logic.  `list_mfa_devices` [will list mfa devices for a particular user](https://docs.aws.amazon.com/sdkforruby/api/Aws/IAM/Client.html#list_mfa_devices-instance_method), or the _currently authenticated user_ - which would be the test harness user, not the root user.  You can't pass the username 'root', either:
```
[cwolfe@lodi inspec-review]$ aws --profile inspec-aws-test-default iam list-mfa-devices --user-name root

An error occurred (NoSuchEntity) when calling the ListMFADevices operation: The user with name root cannot be found.
```

So that's no fun.  However, the CIS PDF tells us that if the account has MFA enabled (which we know from `get_account_summary`, and we have a predicate for it in `has_mfa_enabled?`) BUT has no virtual MFA device (which you have added a predicate for), then it must have a hardware MFA.  So I think this can simply be:

```
   has_mfa_enabled? && !has_virtual_mfa_enabled?
```

test/unit/resources/aws_iam_root_user_test.rb
```diff
@@ -44,4 +44,52 @@ def test_has_mfa_enabled_returns_false_when_account_mfa_devices_is_zero
...
+
+  def test_has_hardware_mfa_enabled_returns_false_when_account_hardware_devices_is_zero
+    test_list_hardware_mfa_devices = OpenStruct.new(
+      mfa_devices: []
+    )
+    @mock_client.expect :list_mfa_devices, test_list_hardware_mfa_devices
+
+    assert_equal false, AwsIamRootUser.new(@mock_conn).has_hardware_mfa_devices?
+  end
+end
```
clintoncwolfe: One missing piece: you need to add the two new matchers to the docs, at `docs/resources/aws_iam_root_account.md.erb`.  Thanks!

clintoncwolfe (changes requested):
A great start!  I suggest a couple of renames, and I found a logic error.  And we need docs!  Great unit tests, though - this is coming along nicely!

lib/resources/aws/aws_iam_root_user.rb
pwelch: TIL!

lib/resources/aws/aws_iam_root_user.rb
pwelch: 👍

pwelch:
@clintoncwolfe Thanks for the feedback! Pushed up changes.

lib/resources/aws/aws_iam_root_user.rb
```diff
@@ -46,6 +46,22 @@ def has_mfa_enabled?
...
+    virtual_mfa_devices.each do |device|
+      # if the root account has a Virtual MFA device then
+      # it will have a special serial number ending in 'root-account-mfa-device'
+      if %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device} =~
+        device['serial_number']
+        return true
+      end
+    end
+    false
+  end
```
TrevorBramble: How about extracting the pattern to an explaining variable and using `Enumerable#any?()` instead?
```ruby
# if the root account has a Virtual MFA device then it will have a special
# serial number ending in 'root-account-mfa-device'
def has_virtual_mfa_enabled?
  mfa_device_pattern = %r{arn:aws:iam::\d{12}:mfa\/root-account-mfa-device}

  virtual_mfa_devices.any? { |device| mfa_device_pattern =~ device['serial_number'] }
end
```

test/unit/resources/aws_iam_root_user_test.rb
```diff
@@ -44,4 +44,54 @@ def test_has_mfa_enabled_returns_false_when_account_mfa_devices_is_zero
...
   end
-end
+
+  def test_has_virtual_mfa_enabled_returns_true_when_account_vmfa_devices_is_one
+    test_list_virtual_mfa_devices = OpenStruct.new(
+      virtual_mfa_devices: [Aws::IAM::Types::VirtualMFADevice.new(
+        serial_number: "arn:aws:iam::123456789011:mfa/root-account-mfa-device",
+        user: Aws::IAM::Types::User.new(
+          user_id: "123456789011",
+          arn:     "arn:aws:iam::123456789011:root",
```
TrevorBramble: I think we're aligning single quotes in this project?

TrevorBramble (changes requested):

clintoncwolfe:
I'm happy with the docs and the code changes.  One last thing, could you addd some integration tests?  Thanks!

pwelch:
@clintoncwolfe integration tests added.

TrevorBramble (approved):

clintoncwolfe (approved):

jquick (approved):
Thanks @pwelch !

PR #2894, opened 2018-03-28
powershell resource: Add support other OSs

jerryaldrichiii:
This adds `powershell` resource support for non-Windows OSs via `pwsh` and Base64 encoded commands.


pwelch (approved):
Looks good to me 👍

jquick (approved):
Thanks @jerryaldrichiii !

PR #2901, opened 2018-03-29
add systemd service for amazon linux 2

zakhark:
Right now service Amazon Linux is treated as upstart_service while Amazon Linux 2 is systemd.
It causes issues like https://github.com/chef/inspec/issues/2639

On the Amazon Linux 2 Candidate, os[:release] returns "2.0", so I'm using .start_with?('2.') here in case they'll use something 2.x in the future.
(Amazon Linux 1 returns something like "2016.09")

Signed-off-by: Zakhar Kleyman <zakhar.kleyman@mongodb.com>

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -162,7 +162,11 @@ module Inspec::Resources
       elsif %w{aix}.include?(platform)
         SrcMstr.new(inspec)
       elsif %w{amazon}.include?(platform)
-        Upstart.new(inspec, service_ctl)
+        if os[:release].start_with?('2.')
+          Systemd.new(inspec, service_ctl)
+        else
+          Upstart.new(inspec, service_ctl)
+        end
       elsif os.solaris?
         Svcs.new(inspec)
       end
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -162,10 +162,10 @@ module Inspec::Resources
       elsif %w{aix}.include?(platform)
         SrcMstr.new(inspec)
       elsif %w{amazon}.include?(platform)
-        if os[:release].start_with?('2.')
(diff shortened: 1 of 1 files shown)


zakhark:
I see that the https://travis-ci.org/chef/inspec/builds/359998634 test for the original commit that was missing DCO was green but the current https://travis-ci.org/chef/inspec/builds/359999391 failed.

The only difference between those 2 was the DCO message, I don't think that could have caused the test to fail. Can it be a flaky test?

Let me know If you have any suggestions on how to approach this (re-run the test, fix something in the code, submit a new pr, etc).

TrevorBramble:
@zakhark Green on re-run. Possibly flaky!

jquick (approved):
Thanks @zakhark !

lib/resources/service.rb
```diff
@@ -162,7 +162,11 @@ def select_service_mgmt # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCom
       elsif %w{aix}.include?(platform)
         SrcMstr.new(inspec)
       elsif %w{amazon}.include?(platform)
-        Upstart.new(inspec, service_ctl)
+        if os[:release].start_with?('2.')
```
chris-rock: Since 2.x will be the new default, I am curious if we should turn the logic the other way: use systemd as a default for amazon linux and just use upstart if the version string starts with `20\d\d`

chris-rock (commented):
Thank you @zakhark for uncovering this!

lib/resources/service.rb
zakhark: That's a very good point, I've updated the logic

miah (approved):
Thanks @zakhark !

chris-rock (approved):
Thank you @zakhark

PR #2902, opened 2018-03-30
Add automate reporter

jquick:
Adds the automate reporter to send directly to automate. Currently this report is only usable via the `--json-config` as it requires additional metadata.

Fixes https://github.com/chef/inspec/issues/2547

### DIFF
--- a/lib/inspec/formatters/base.rb
+++ b/lib/inspec/formatters/base.rb
@@ -69,6 +69,7 @@ module Inspec::Formatters
         name: platform(:name),
         release: platform(:release),
         target: backend_target,
+        uuid: platform(:uuid),
       }
     end
 
--- a/lib/inspec/formatters/base.rb
+++ b/lib/inspec/formatters/base.rb
@@ -183,7 +183,12 @@ module Inspec::Formatters
 
     def platform(field)
       return nil if @backend.nil?
-      @backend.platform[field]
+      begin
+        @backend.platform[field]
+      rescue Train::Error => e
+        Inspec::Log.error(e.message)
+        nil
+      end
     end
 
     def backend_target
--- a/lib/inspec/formatters/base.rb
+++ b/lib/inspec/formatters/base.rb
@@ -181,6 +181,9 @@ module Inspec::Formatters
       end
     end
 
+    # This formatter runs for all reports and we cannot error on missing fields.
+    # Return nil if not found or Train error. If needed, we will raise an error inside
+    # the proper report.
     def platform(field)
       return nil if @backend.nil?
       begin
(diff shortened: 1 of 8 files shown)


lib/inspec/reporters/automate.rb
```diff
@@ -0,0 +1,64 @@
...
+        return false
+      end
+    end
+
+    private
+
+    # This hashes the passed string into SHA1.
+    # Then it downgrades the 160bit SHA1 to a 128bit
+    # then we format it as a valid UUIDv5.
+    def uuid_from_string(string)
```
chris-rock, line 66: Interesting idea 👍

lib/inspec/reporters/automate.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      uri = URI(@config['url'])
+      req = Net::HTTP::Post.new(uri.path, headers)
+      req.body = enriched_report.to_json
+      begin
+        Inspec::Log.debug "Posting report to Chef Automate: #{uri.path}"
+        Net::HTTP.start(uri.hostname,
+                        uri.port,
+                        use_ssl: uri.scheme == 'https',
+                        verify_mode: OpenSSL::SSL::VERIFY_NONE) do |http|
```
chris-rock: Can we make this a config option? I would like to go to a world where we verify it

chris-rock (changes requested):
Short and simple 👍 Thank you @jquick

lib/inspec/reporters/automate.rb
```diff
@@ -0,0 +1,76 @@
...
+require 'json'
+require 'net/http'
+
+module Inspec::Reporters
+  class Automate < Json
+    def initialize(config)
+      super(config)
+
+      # default to using no ssl for sending reports
+      @config['use_ssl'] = @config['use_ssl'] || false
```
jerryaldrichiii: Would `@config['use_ssl'] ||= false` work here?

jerryaldrichiii (approved):
Looks good to me, thanks @jquick

lib/inspec/formatters/base.rb
```diff
@@ -182,7 +183,12 @@ def format_expectation_message(example)
 
     def platform(field)
       return nil if @backend.nil?
-      @backend.platform[field]
+      begin
+        @backend.platform[field]
+      rescue Train::Error => e
+        Inspec::Log.error(e.message)
+        nil
```
chris-rock, line 193: I think we should document, why it is okay to ignore the error here

lib/inspec/reporters/automate.rb
chris-rock: I think `use_ssl` is not precise, because we use ssl in any case. The only difference is that in one case we validate the certificate and in the other we don't. Maybe `@config['verify_ssl'] = @config['verify_ssl'] || false`

chris-rock (approved):
Great work @jquick

PR #2904, opened 2018-03-31
Update example resource syntax

jerryaldrichiii:
This makes the example resource satisfy Rubocop and adds a few other recent changes (e.g. `supports`)

### DIFF
--- a/test/functional/inspec_exec_json_test.rb
+++ b/test/functional/inspec_exec_json_test.rb
@@ -58,7 +58,7 @@ describe 'inspec exec with json formatter' do
         "license" => "Apache-2.0",
         "summary" => "Demonstrates the use of InSpec Compliance Profile",
         "version" => "1.0.0",
-        "sha256" => "8eed5154c9fa0174067ab475c0ad4a053f772590d129eb324101fe43ef30794d",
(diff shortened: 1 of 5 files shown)


jquick (approved):
Thanks @jerryaldrichiii !

miah (approved):

test/functional/inspec_exec_json_test.rb
```diff
@@ -58,7 +58,7 @@
         "license" => "Apache-2.0",
         "summary" => "Demonstrates the use of InSpec Compliance Profile",
         "version" => "1.0.0",
-        "sha256" => "8eed5154c9fa0174067ab475c0ad4a053f772590d129eb324101fe43ef30794d",
+        "sha256" => "57709d3a3d5cd06f4179be7e6fbe254c09e3af25ce274e474d52623e34487cc4",
```
TheLonelyGhost, line 61: I understand why this needs to be regenerated, but how are you figuring out what value to put in here?

jerryaldrichiii: Run the test, then copy the one that is generated vs the one I expect. Nothing magical.

PR #2905, opened 2018-04-02
Add Cisco IOS `enable_password` support

jerryaldrichiii:
This adds support for using the `ssh://` URI with Cisco IOS devices via swapping the transport after platform detection.

This also adds support for passing the IOS enable password via the CLI.

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -40,7 +40,7 @@ module Inspec
     #
     # @param [Hash] config for the transport backend
     # @return [TransportBackend] enriched transport instance
-    def self.create(config) # rubocop:disable Metrics/AbcSize
+    def self.create(config) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       conf = Train.target_config(config)
       name = Train.validate_backend(conf)
       transport = Train.create(name, conf)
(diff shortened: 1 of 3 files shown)


lib/inspec/backend.rb
```diff
@@ -53,6 +53,14 @@ def self.create(config) # rubocop:disable Metrics/AbcSize
         raise "Can't connect to transport backend '#{name}'."
       end
 
+      # Cisco IOS requires a special implementation of `Net:SSH`. This uses the
+      # SSH transport to identify the platform, but then replaces that transport
+      # with the CiscoIOS transport in order to behave as expected later.
+      if name == 'ssh' && connection.platform.cisco_ios?
```
jquick: This bothers me a bit, as we use the ssh connection to run commands to see what platform its on and then override the platform. I think we should just have the user use `cisco_ios` transport when needed.

chris-rock: Feels like we have some conceptual parts that we need to discuss about the cisco specific transport

jerryaldrichiii:
I agree @jquick, I've removed the connection swap logic from this PR and will update my Train PR to do the connection swap instead.

jquick (approved):
Thanks @jerryaldrichiii

jquick:
🛑  DO NOT MERGE 🛑 

This PR requires a new train release pin.

jquick:
Safe to merge.

miah (approved):
Thanks @jerryaldrichiii

PR #2908, opened 2018-04-03
Added a description to steer people to correct resource

username-is-already-taken2:
Whilst revising for my inspec exam I noticed that these two resources didn't have a description.  I appreciate they are just helper resources but they are still listed when running `help resources` within an inspec shell.

So I added a description to steer people in the right direction.

Signed-off-by: DigitalGaz <digitalgaz@hotmail.com>


chris-rock (approved):
Thank you @username-is-already-taken2 for the clarification

jquick (approved):
Thanks @username-is-already-taken2 !

PR #2910, opened 2018-04-03
Makes JSON resource enumerable, despite method_missing magic

TheLonelyGhost:
Per Slack discussion:

```ruby
obj = json(content: '[{"Name": "abc"}, {"Name": "def"}]')

obj.each do |node|
  describe node do
    it { should eq nil }
  end
end
```

Above does not respond as expected due to method_missing magic. One expects 2 tests to execute, but each does not iterate over its data. This PR corrects that and includes all of the same enumerable abilities on the `json` (and by extension, `yaml`) resource that its data contains.

Yes, I am aware this could also be accomplished by doing `obj.params.each do` instead of `obj.each do`, but I don't feel this is very intuitive. We're exposing the `:[]` method via method_missing, why not make it react like a hash or array too?

### DIFF
--- a/lib/resources/json.rb
+++ b/lib/resources/json.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 require 'utils/object_traversal'
+require 'utils/enumerable_delegation'
 require 'utils/file_reader'
 
 module Inspec::Resources
@@ -37,6 +38,9 @@ module Inspec::Resources
       # load the raw content from the source, and then parse it
       @raw_content = load_raw_content(opts)
       @params = parse(@raw_content)
+
+      # If the JSON content is enumerable, make this object enumerable too
+      extend EnumerableDelegation if @params.respond_to?(:each)
     end
 
     # Shorthand to retrieve a parameter name via `#its`.
--- a/lib/utils/enumerable_delegation.rb
+++ b/lib/utils/enumerable_delegation.rb
@@ -1,0 +1,9 @@
+# encoding: utf-8
+
+module EnumerableDelegation
+  include Enumerable
+
+  def each(&block)
+    @params.each(&block)
+  end
+end
--- a/test/unit/resources/json_test.rb
+++ b/test/unit/resources/json_test.rb
@@ -32,6 +32,14 @@ describe 'Inspec::Resources::JSON' do
     it 'doesnt resolve symbol-notation names' do
       _(resource.send(:'x.y.z')).must_be_nil
     end
+
+    it 'is enumerability matches its data' do
+      enum = load_resource('json', content: '["a","b"]')
+      not_enum = load_resource('json', content: '525600')
+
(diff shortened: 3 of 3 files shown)


jquick:
Thanks @TheLonelyGhost ! This looks pretty slick. You mind adding a unit test for this?

TheLonelyGhost:
@jquick I believe I did under `test/unit/resources/json_test.rb`. Or is there something more you’d like me to do with regards to test support?

jquick (approved):
Thanks @TheLonelyGhost !

lib/resources/json.rb
```diff
@@ -37,6 +38,9 @@ def initialize(opts)
       # load the raw content from the source, and then parse it
       @raw_content = load_raw_content(opts)
       @params = parse(@raw_content)
+
+      # If the JSON content is enumerable, make this object enumerable too
+      extend EnumerableDelegation if @params.respond_to?(:each)
```
TrevorBramble, line 43: Reading above, and `parse()`, it looks like there should always be a valid hash for `@params`, what's the scenario you have in mind where this condition would be false?

lib/utils/enumerable_delegation.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+
+module EnumerableDelegation
```
TrevorBramble, line 3: This is purpose built for params instead of delegation to some injected receiver, so I wonder if Params should be part of the name? Maybe `EnumerableParamsDelegator`?

TrevorBramble (commented):
This is great, just have a couple of questions. =^)

lib/resources/json.rb
TheLonelyGhost: I originally considered `json(content: 'null')` or something similar that is neither an array nor a JSON object. Upon actually testing this just now, I see that `JSON.parse()` actually fails with anything other than those two. That was news to me.

[From the look of things](https://stackoverflow.com/questions/18419428/what-is-the-minimum-valid-json) "valid JSON" was once defined as either an array or object, but ECMAScript 5 expanded it to include any data type such as `"fizz buzz"`, `null`, `3.14`, and others. It doesn't appear ruby's JSON parser has caught up yet.

At this point, this line doesn't really hurt anything, but builds in support for when and if the JSON parser is updated to the ES5 standard definition of "valid" JSON.

lib/utils/enumerable_delegation.rb
TheLonelyGhost: I didn't like the dependency on the params attribute either, but minimum viable solution and all. I agree that your suggestion is a better name, but I'm definitely more interested in if you have a solution that delegates to an injected receiver. Ideas?

clintoncwolfe (approved):
Seems sensible, and the test makes sense and passes.  Thanks, @TheLonelyGhost

PR #2911, opened 2018-04-04
[docs] Amended Syntax to correct formatting

username-is-already-taken2:
Hi 

I noticed the package resource in the docs had an example that wasn't rendering correctly

package_doc

So I've amended that 
_I need to fess up that I couldn't workout how to stop middleman rendering the `_` 's to `<em>` tags so I changed the wording_ :)

package_doc_after

To test the website build I needed to install node.js, so I added a hint to the readme.

Best Regards

Gary


miah:
Thanks @username-is-already-taken2 ! This looks good, but we'll need you to update this PR with a [DCO](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) Sign-Off.

username-is-already-taken2:
Thanks @miah , sorry I missed the `-s` off my commit.

I also might have done someone wrong as I pulled in the latest changes from origin before amending my previous commit so I've ended up with two commits on this pr.

So don't know if that's bad or I got it in the wrong order, I'm still learning git and know enough to be dangerous ;)

If I need to start again I can re do it

jquick (approved):
Thanks @username-is-already-taken2 !

PR #2918, opened 2018-04-05
Policy Statement Search capability for aws_iam_policy

### LINKED ISSUE #2749: AWS IAM policy statement search facility
### 🎛 Description

Several IAM and S3 resources support inline policies.  Additionally, there are self-contained IAM policies, which may be attached to IAM users, groups, and roles.  Policies are composed of a list of Statements, each with certain well-formed properties.  You can read about the data structure in the [AWS docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_statement.html).

Many use cases require determining whether a particular statement is present in a policy (or not).  A very rich example is CIS AWS Benchmark v1.1, Recommendation 1.18, which requires a long list of statements to be in effect on a Role - but does not specify if they are to be an inline policy or an attached policy.

2.0.26

### 💁 Possible Solutions

Suppose my-policy looks like:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:ListAllMyBuckets",
        "s3:GetBucketLocation"
      ],
      "Resource": "arn:aws:s3:::*"
    },
}
```

Possible proposed syntax:

```
describe aws_iam_policy('my-policy') do
  # We'd have to override include here, since we would want subset hash matching, not exact
  its('statements') { should_not include(Effect: 'Allow', Action: '*') }
  # Or...
  it { should have_statement(Effect: 'Allow', Action: '*') }
  # Actions, Resources, and Conditions can all take a single element as a string, or allow an array.
  # How would we handle that?
  # I'd think this should evaluate true - which means we can't just do a simple hash comparison.
  it { should have_statement(Action: 's3:GetBucketLocation') }
end
```

So, an IAM policy has its policy baked into it.  Other resources (IAM users, groups, roles, S3 buckets...) support inline policies as well; it makes sense then to push the has_statement? matcher out to a mixin, or a similar solution.

It's less clear to me what to when you need to know if a statement is in effect on a resource that may also have policies attached to it.  Assuming that (say) in `aws_iam_role` is able to enumerate its attached policies by ARN, does the InSpec user then have to fetch any attached policies and examine them, as well?

Refs #2748 #2584 #2562

clintoncwolfe:
This PR adds one low-level and two mid-level features for auditing the statements embedded in an IAM Policy.

While the features are currently only available on `aws_iam_policy`, it was designed to be easily refactored to become a mixin, allowing us to enable this feature on other policy-bearing objects, such as users, groups, roles, S3 buckets, Lambdas, etc.

Low-level: 
* `policy` is a property returning a struct as returned by the AWS API, after decoding and inflation to Ruby data structures.  It is intended to appeal to those who wish to perform in-depth traversal of the data structures in Ruby code.

Mid-level :
* `statement_count` is a property returning a numeric count of statements.  This can be used to detect if an unexpected statement was added or removed.
* `have_statement` is a matcher which provides a sophisticated search facility into the statement list.  This can be used to detect the presence (or absence) of particular statements.

Fixes #2749 

Unit and integration test run:
screen shot 2018-04-05 at 14 15 57

### DIFF
--- a/docs/resources/aws_iam_policy.md.erb
+++ b/docs/resources/aws_iam_policy.md.erb
@@ -5,9 +5,9 @@ platform: aws
 
 # aws\_iam\_policy
 
-Use the `aws_iam_policy` InSpec audit resource to test properties of a single managed AWS IAM Policy.
+Use the `aws_iam_policy` InSpec audit resource to test detailed properties of a single managed AWS IAM Policy.  Use `aws_iam_policies` to audit IAM policies in bulk.
 
-A policy is an entity in AWS that, when attached to an identity or resource, defines their permissions. AWS evaluates these policies when a principal, such as a user, makes a request. Permissions in the policies determine if the request is allowed or denied.
+A policy is an entity in AWS that, when attached to an identity or resource, defines their permissions. AWS evaluates these policies when a principal, such as a user, makes a request. Permissions (known as Statements) in the policies determine if the request is allowed or denied.
 
 Each IAM Policy is uniquely identified by either its policy\_name or arn.
 
@@ -50,11 +50,22 @@ The following examples show how to use this InSpec audit resource.
       it { should be_attached }
     end
 
+### Examine the policy statements
+
+    describe aws_iam_policy('my-policy') do
+      # Verify that there is at least one statement allowing access to S3
+      it { should have_statement(Action: 's3:PutObject', Effect: 'allow') }
+
+      # have_statement does not expand wildcards.  If you want to verify 
+      # they are absent, an explicit check is required.
+      it { should_not have_statement(Action: 's3:*') }
+    end
+
 <br>
 
 ## Properties
 
-* `arn`, `attachment_count`, `attached_groups`, `attached_roles`,`attached_users`, `default_version_id`
+* `arn`, `attachment_count`, `attached_groups`, `attached_roles`,`attached_users`, `default_version_id`, `policy`, `statement_count`
 
 ## Property Examples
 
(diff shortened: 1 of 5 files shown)


docs/resources/aws_iam_policy.md.erb
```diff
@@ -142,3 +175,59 @@ The test will pass if the identified policy attached the specified role.
...
+* `Resource` - Expresses the operation's target. Acceptable values are ARNs, including the '*' wildcard. `Resource` may also use a list of ARN values.
+
+Please note the following about the behavior of `have_statement`:
+* `Action`, `Sid`, and `Resource` allow using a regular expression as the search critera instead of a string literal.
+* it does not support wildcard expansion; to check for a wildcard value, check for it explicitly. For example, if the policy includes a statement with `"Action": "s3:*"` and the test checks for `Action: "s3:PutObject"`, the test _will not match_. You must write an additional test checking for the wildcard case. 
+* it supports searching list values. For example, if a statement contains a list of 3 resources, and a `have_statement` test specifes _one_ of those resources, it will match.
+* `Action` and `Resource` allow using a list of string literals or regular expressions in a test, in which case _all_ must match on the _same_ statement for the test to match. Order is ignored.
+* it does not support the `Principal` nor `Conditional` key, nor any of `NotAction`, `NotPrincipal`, or `NotResource`.
```
kagarmoe: it does not support the `Principal` or `Conditional` keys, or any of `NotAction`, `NotPrincipal`, or `NotResource`.

docs/resources/aws_iam_policy.md.erb
```diff
@@ -106,10 +117,32 @@ The 'default_version_id' value of the specified policy.
...
+### policy
+
+Returns the default version of a policy document as a Ruby hash. This hash contains the enforced policy statements and is useful for performing checks that cannot be expressed using higher-level matchers like `have_statement`.
+
+For details regarding the contents of this structure, refer to the [AWS IAM Policy JSON Reference](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies.html).  A set of examples is [also available](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_examples.html).
+
+Example: 
+
+    describe aws_iam_policy('my-policy') do
+      # TODO - example?
```
miah: Forgotten TODO?

miah (changes requested):
Thanks @clintoncwolfe ! I love the readability of this resource. Apart from the few documentation changes needed I think this looks great!

lib/resources/aws/aws_iam_policy.rb
```diff
@@ -50,8 +69,133 @@ def attached_to_role?(role_name)
...
+      backend = BackendFactory.create(inspec_runner)
+      gpv_response = backend.get_policy_version(policy_arn: arn, version_id: default_version_id)
+      @policy = JSON.parse(URI.decode_www_form_component(gpv_response.policy_version.document))
+    end
+    @policy
+  end
+
+  def statement_count
+    return nil unless exists?
+    policy['Statement'].count
```
jquick, line 86: Will statement always be populated?

jquick (approved):
Thanks @clintoncwolfe ! Looks good.

miah (approved):

lib/resources/aws/aws_iam_policy.rb
clintoncwolfe: Yes - it's invalid otherwise, and AWS won't accept an invalid policy JSON document.

PR #2922, opened 2018-04-06
Update PowerShell requirement and add an asterisk

jerryaldrichiii:
We require a minimum of PowerShell 5.0 to use Windows as a remote target.


chris-rock (approved):
Thank you for the clarification @jerryaldrichiii

jquick (approved):

PR #2928, opened 2018-04-09
Make names for AWS Config service objects optional

### LINKED ISSUE #2900: Do not require resource names for AWS Config objects
### 🎛 Description

🐛 bug report

Bug, or at least a misfeature.  `aws_config_recorder` and `aws_config_delivery_channel` currently require you to provide the name of the Recorder or Delivery Channel.  However, AWS Config Recorders and Delivery Channels are singleton-per-region (and our InSpec runs are per-region).

If the resource name was optional, the AWS API is able to list the existing recorders, and you could get the singleton recorder without specifying a name.

We should add code that throw an exception if more than one resurce is returned, in case AWS starts supporting multiple recorder.  That behavior would match our typical response when multiples match.

2.1.17

### 💁 Possible Solutions

Don't require a resource_name; when you call `describe_config_recorder` with no args, it will return all (presumably, one or zero) recorders.

clintoncwolfe:
As described on #2900, we were unnecessarily requiring an AWS resource name for the two Config Service inspec resources, `aws_config_recoorder` and `aws_config_delivery_resource`.  If you don't provide a name in the query, the API gives you all config resources, and since they are singletons, it's easy to pick the right one.

Also found we had no unit tests for `aws_config_delivery_resource`; added them.

Also some general code cleanup in both affected resources.

Fixes #2900

### DIFF
--- a/docs/resources/aws_config_delivery_channel.md.erb
+++ b/docs/resources/aws_config_delivery_channel.md.erb
@@ -9,11 +9,15 @@ to an S3 Bucket, an SNS or both.
 
 Use the `aws_config_delivery_channel` InSpec audit resource to examine how the AWS Config service delivers those change notifications.
 
+As of April 2018, each AWS region may have only one Delivery Channel.
+
 <br>
 
-## Syntax
+## Resource Parameters
 
-An `aws_config_delivery_channel` resource block declares the tests for a single AWS Config delivery channel.
+An `aws_config_delivery_channel` resource block declares the tests for a single AWS Config Delivery Channel.
+
+You may specify the Delivery Channel name:
 
     describe aws_config_delivery_channel('my_channel') do
       it { should exist }
@@ -23,45 +27,35 @@ An `aws_config_delivery_channel` resource block declares the tests for a single 
       it { should exist }
     end
 
+However, since you may only have one DDelivery Channel per region, and InSpec connections are per-region, you may also omit the `channel_name` to obtain the one Delivery Channel (if any) that exists:
+
+    describe aws_config_delivery_channel do
+      it { should exist }
+    end
+
 <br>
 
 ## Examples
 
 The following examples show how to use this InSpec audit resource.
 
-### Test how frequent the channel writes configuration changes to the s3 bucket.
+### Test how frequently the channel writes configuration changes to the s3 bucket.
 
     describe aws_config_delivery_channel(channel_name: 'my-recorder') do
       its(delivery_frequency_in_hours) { should be > 3 }
     end
     
 ## Properties
-    
-### s3_bucket_name
 
-Provides the name of the s3 bucket that the channel sends configuration changes to.  This is an optional value since a Delivery Channel can also talk to an SNS.
+### channel\_name
 
-    describe aws_config_delivery_channel(channel_name: 'my_channel')
-      its('s3_bucket_name') { should eq 'my_bucket' }
+Returns the name of the Delivery Channel.
+
+    describe aws_config_delivery_channel do
+      its('channel_name') { should cmp 'my-channel' }
     end
-    
-### s3_key_prefix
 
-Provides the s3 object key prefix (or "path") under which configuration data will be recorded.
-
-    describe aws_config_delivery_channel(channel_name: 'my_channel')
-      its('s3_key_prefix') { should eq 'log/' }
-    end
-    
-### sns_topic_arn
-
-Provides the ARN of the SNS topic for which the channel sends notifications about configuration changes.
-
-    describe aws_config_delivery_channel(channel_name: 'my_channel')
-      its('sns_topic_arn') { should eq 'arn:aws:sns:us-east-1:721741954427:sns_topic' }
-    end
-    
-### delivery_frequency_in_hours
+### delivery\_frequency\_in\_hours
 
 Provides how often the AWS Config sends configuration changes to the s3 bucket in the delivery channel.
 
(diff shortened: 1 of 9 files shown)


jquick (approved):
Looks good! Thanks.

docs/resources/aws_config_delivery_channel.md.erb
```diff
@@ -23,57 +27,70 @@ An `aws_config_delivery_channel` resource block declares the tests for a single
       it { should exist }
     end
 
+However, since you may only have one DDelivery Channel per region, and InSpec connections are per-region, you may also omit the `channel_name` to obtain the one Delivery Channel (if any) that exists:
```
TrevorBramble: DDelivery

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -28,29 +28,29 @@ def validate_params(raw_params)
...
     validated_params
   end
 
   def fetch_from_api
     backend = BackendFactory.create(inspec_runner)
-    query = { delivery_channel_names: [@channel_name] }
-    catch_aws_errors do
-      @resp = backend.describe_delivery_channels(query)
+    query = @channel_name ? { delivery_channel_names: [@channel_name] } : {}
+    response = {}
```
TrevorBramble: Does the assignment on line 39 sometimes return nil? If so, maybe just `response = backend.describe_delivery_channels(query) || {}` instead?

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -28,29 +28,29 @@ def validate_params(raw_params)
...
+    begin
+      response = backend.describe_delivery_channels(query)
+    rescue Aws::ConfigService::Errors::NoSuchDeliveryChannelException
```
TrevorBramble: Dropping this rescue to the bottom of the method would clear up the "happy path" for readers and removes the need to early return when handling the exception.

```ruby
def fetch_from_api
  # ...
rescue Aws::ConfigService::Errors::NoSuchDeliveryChannelException
  @exists = false
end
```

lib/resources/aws/aws_config_delivery_channel.rb
```diff
@@ -28,29 +28,29 @@ def validate_params(raw_params)
...
-    @delivery_frequency_in_hours = @channel[:config_snapshot_delivery_properties][:delivery_frequency] unless @channel[:config_snapshot_delivery_properties].nil?
+    @exists = !response.empty?
+    return unless exists?
+
+    channel = response.delivery_channels.first.to_h
+    @channel_name = channel[:name]
+    @s3_bucket_name = channel[:s3_bucket_name]
+    @s3_key_prefix = channel[:s3_key_prefix]
+    @sns_topic_arn = channel[:sns_topic_arn]
+    @delivery_frequency_in_hours = channel[:config_snapshot_delivery_properties][:delivery_frequency] unless channel[:config_snapshot_delivery_properties].nil?
```
TrevorBramble: You can remove the conditional clause with [dig](http://ruby-doc.org/core-2.5.0/Hash.html#method-i-dig). =^)
```ruby
@delivery_frequency_in_hours = channel.dig(:config_snapshot_delivery_properties, :delivery_frequency)
```

lib/resources/aws/aws_config_recorder.rb
```diff
@@ -50,35 +50,34 @@ def validate_params(raw_params)
...
-    @query = { configuration_recorder_names: [@recorder_name] }
+    query = @recorder_name ? { configuration_recorder_names: [@recorder_name] } : {}
+
+    response = nil
+    begin
+      response = backend.describe_configuration_recorders(query)
+    rescue Aws::ConfigService::Errors::NoSuchConfigurationRecorderException
+      @exists = false
+      return
+    end
```
TrevorBramble: Same as the other `fetch_from_api`, re: `rescue` at method level instead of introducing a `begin` block.

test/unit/resources/aws_config_delivery_channel_test.rb
```diff
@@ -0,0 +1,128 @@
...
+#=============================================================================#
+#                               Test Fixtures
+#=============================================================================#
+
+module AwsMDCSB
+  class Basic < AwsBackendBase
+    def describe_delivery_channels(query = {})
+      fixtures = {
+        'default' => Aws::ConfigService::Types::DescribeDeliveryChannelsResponse.new(
+          :delivery_channels => [
```
TrevorBramble, line 110: Sometimes I see `{:foo => :bar}` and sometimes I see `{foo: :bar}`, particularly in the Rubocop-ignored test code. Shouldn't we always use JSON style for symbol keys?

test/unit/resources/aws_config_delivery_channel_test.rb
```diff
@@ -0,0 +1,128 @@
...
+              config_snapshot_delivery_properties: {
+                 delivery_frequency: 'Three_Hours'
+              },
+            },
+          ]
+        ),
+      }
+      return fixtures['default'] if query.empty?
+      return fixtures[query[:delivery_channel_names][0]] unless fixtures[query[:delivery_channel_names][0]].nil?
+      raise Aws::ConfigService::Errors::NoSuchDeliveryChannelException.new(nil, nil)
```
TrevorBramble, line 125: These look like three `case`s to me. ;^)

TrevorBramble (commented):
Loads of great additions and improvements. I have some suggestions to offer but I think only the typo is a blocker. ;^)

lib/resources/aws/aws_config_delivery_channel.rb
clintoncwolfe: I've moved the `rescue` to be method-level, as you asked.  I generally dislike doing that, however.  By moving the rescue away from the "dangerous" code, it obscures _which_ line(s) are potentially dangerous.  While this code only has one such dangerous expression, other methods may need to call multiple APIs that may fail.

There is an argument there for shorter, more granular methods.

lib/resources/aws/aws_config_delivery_channel.rb
clintoncwolfe: No, it never returns a `{}`.

```
inspec> c = inspec.backend.aws_client(Aws::ConfigService::Client)
=> #<Aws::ConfigService::Client>
inspec> c.describe_delivery_channels
=> #<struct Aws::ConfigService::Types::DescribeDeliveryChannelsResponse delivery_channels=[]>
inspec> c.describe_delivery_channels({})
=> #<struct Aws::ConfigService::Types::DescribeDeliveryChannelsResponse delivery_channels=[]>
inspec> c.describe_delivery_channels(delivery_channel_names:['nonesuch'])
Aws::ConfigService::Errors::NoSuchDeliveryChannelException: Cannot find delivery channel with specified name 'nonesuch'.
from /Users/cwolfe/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/aws-sdk-core-2.11.21/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
inspec> c.describe_delivery_channels.delivery_channels
=> []
inspec> c.describe_delivery_channels.delivery_channels.empty?
=> true
```

Thanks for mentioning it, though - that means this is wrong:
```
@exists = !response.empty?
```

Should be:
```
@exists = !response.delivery_channels.empty?
```

test/unit/resources/aws_config_delivery_channel_test.rb
clintoncwolfe: If it ignored by rubocop, that means we as a team have decided it is code that is not worthwhile to enforce style conventions.  We can open that for discussion if you'd like.

TrevorBramble (approved):
Thanks for considering my feedback!

PR #2934, opened 2018-04-09
Add warning when returning DEFAULT_ATTRIBUTE

### LINKED ISSUE #2865: Defining an attribute without a default value generates a stacktrace

TrevorBramble:
When the anonymous DEFAULT_ATTRIBUTE class is used, log a warning.

We now pass in the attribute name to that class so it can be used in the log message.

Fixes #2865 
Complement of #2891 

image

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -6,7 +6,17 @@ module Inspec
     attr_writer :value
 
     DEFAULT_ATTRIBUTE = Class.new do
+      def initialize(name)
+        @name = name
+      end
+
       def method_missing(*_)
+        Inspec::Log.warn(
+          "Returning DEFAULT_ATTRIBUTE for '#{@name}'. "\
+          "Use --attrs to provide a value for '#{@name}' or specify a default  "\
+          "value with `attribute('#{@name}', default: 'somedefault', ...)`.",
+        )
+
         self
       end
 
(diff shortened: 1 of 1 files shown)


jquick (approved):
I like this. Thanks @TrevorBramble !

jerryaldrichiii (approved):
Looks good to me. Love it!

miah (approved):
Thanks @TrevorBramble !

PR #1511, opened 2017-02-24
Added oracle linux to rhel platform section of 'service' resource.

carldjohnston:
Tests currently skipping over 'service' resource on oracle linux. 

Broken when chef/train was fixed to detect oracle linux correctly.

PR #2935, opened 2018-04-10
Add cloudlinux under redhat family

### LINKED ISSUE #2932: Service resource doesn't work on CloudLinux
### 🎛 Description

Hi. I am getting this warning when I am trying to use inspec service resource on a CloudLinux server.

  Service nrpe
     âº  The `service` resource is not supported on your OS yet.

Can you please fix it? Its basically CentOS.

inspec 1.49.2
CloudLinux release 7.4 (Georgy Grechko)

### 🤔 Replication Case

describe service('nrpe') do
  it { should be_installed }
  it { should be_enabled }
  it { should be_running }
end

tarcinil:
Train is already able to add cloudlinux under os and platforms.
```
You are currently running on:

    Name:      cloudlinux
    Release:   7.4
    Arch:      x86_64
```

Adding cloudlinux to the `select_service_mgmt` method so the correct init system can be returned for it.  Adding tests for the OS and adding it to the helper.

fixes #2932

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -134,9 +134,9 @@ module Inspec::Resources
         else
           SysV.new(inspec, service_ctl || '/usr/sbin/service')
         end
-      elsif %w{redhat fedora centos oracle}.include?(platform)
+      elsif %w{redhat fedora centos oracle cloudlinux}.include?(platform)
         version = os[:release].to_i
-        if (%w{redhat centos oracle}.include?(platform) && version >= 7) || (platform == 'fedora' && version >= 15)
+        if (%w{redhat centos oracle cloudlinux}.include?(platform) && version >= 7) || (platform == 'fedora' && version >= 15)
           Systemd.new(inspec, service_ctl)
         else
           SysV.new(inspec, service_ctl || '/sbin/service')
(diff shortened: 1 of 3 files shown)


chris-rock:
Thank you @tarcinil for this improvement. It feels like we should add `cloudlinux` to https://github.com/chef/train/blob/master/lib/train/platforms/detect/specifications/os.rb as a RHEL system, too. Then InSpec can use its `platform.rhel?` to match that system too

tarcinil:
@chris-rock I thought about that as well.  I will work on that this evening or tomorrow evening.

EDIT: This was created in chef/train#281.

jquick:
@tarcinil - Your train changes were merged and released. This looks like it bases the changes off os name so it would have to stay `cloudlinux`. If all redhat were valid you could do `os.family = redhat`

jquick:
I don't think this is needed anymore since we have the train change in place which adds cloudlinux as a redhat platform. Closing this for now. Please reopen if there is something additional that needs added.

KrisShannon:
I missed this closed pull request when I added #3293

I have inspec-2.2.35 and train-1.4.21 and I still needed this patched for CloudLinux.

In pull #1511 (Added oracle linux to rhel platform section of 'service' resource.) there is a comment from @chris-rock about wrlinux that I think is why you couldn't just use `platform.rhel?`

chris-rock:
I agree with @tarcinil since the service implementation does not use the family identifier, we would need to add cloudlinux here. We may consider refactoring the service resource to use the rhel family instead.

jquick (approved):
Thanks @tarcinil

miah (approved):
Thanks @tarcinil

jquick:
Thanks for this @tarcinil, sorry for the lateness on this and confusion!

Issue #2884, opened 2018-03-27
Check for Terraform update for integration tests


clintoncwolfe:
### 🎛 Description

🙋 feature request

We currently pin TF to 0.10.x, based needing to stay on a version that worked with the azure and aws plugins needed.  Is that still true?  TF is currently on 0.11.x, can we move to that?

### 🌍 InSpec and Platform Version

2.1.14, master tip a5ff79d30

### 🤔 Replication Case

Running `bundle exec rake test:aws` will require TF 0.10 and pinned plugins.

### 💁 Possible Solutions

Try pinning to 0.11.x, and loosening the azure and aws pins to see if the latest works.

aaronlippold:
We have been running on the current TF without issues but @rx294 and @dromazmj would know the best answer with respect to the integration tests

PR #2937, opened 2018-04-10
 #2810 - Add check if aws s3 bucket is encrypted.

### LINKED ISSUE #2810: Feature request: Check if AWS S3 bucket is encrypted
Check if an s3 bucket is encrypted using the default encryption - https://docs.aws.amazon.com/AmazonS3/latest/dev/bucket-encryption.html

Many compliances have the requirement that data at rest is encrypted.

UranusBytes:
Fixes  #2810 - Add check if aws s3 bucket is encrypted.

Required terraform aws provider >= 1.6

Signed-off-by: Jeremy Phillips <github@uranusbytes.com>


test/unit/resources/aws_s3_bucket_test.rb
```diff
@@ -285,5 +293,18 @@ def get_bucket_logging(query)
       end
       buckets[query[:bucket]]
     end
+
+    def get_bucket_encryption(query)
+      buckets = {
+          'public' => OpenStruct.new({ server_side_encryption_configuration: OpenStruct.new({ rules: [] }) })
```
TrevorBramble: You can omit these curly braces and lose some of the noise here. `OpenStruct.new(server_side_encryption_configuration: OpenStruct.new(rules: []))`

test/unit/resources/aws_s3_bucket_test.rb
```diff
@@ -285,5 +293,18 @@ def get_bucket_logging(query)
...
+      }
+      if query[:bucket].eql?'private'
```
TrevorBramble: Add a space before the `'private'` argument, please. (Or wrap it with parentheses.)

test/unit/resources/aws_s3_bucket_test.rb
```diff
@@ -285,5 +293,18 @@ def get_bucket_logging(query)
...
+        raise Aws::S3::Errors::ServerSideEncryptionConfigurationNotFoundError.new(nil, nil)
+      end
+      unless buckets.key?(query[:bucket])
+        raise Aws::S3::Errors::NoSuchBucket.new(nil, nil)
+      end
+      buckets[query[:bucket]]
```
TrevorBramble: I think putting these three cases into an actual `case` statement would make the purpose of this method more clear.

```ruby
def get_bucket_encryption(query)
  buckets = {
    'public' => OpenStruct.new(server_side_encryption_configuration: OpenStruct.new(rules: []))
  }

  case
  when query[:bucket].eql? 'private'
    raise Aws::S3::Errors::ServerSideEncryptionConfigurationNotFoundError.new(nil, nil)
  when !buckets.key?(query[:bucket])
    raise Aws::S3::Errors::NoSuchBucket.new(nil, nil)
  else
    buckets[query[:bucket]]
  end
end
```

In writing that, I realized the second and third cases are actually just a missing use of [`fetch()`](http://ruby-doc.org/core-2.5.0/Hash.html#method-i-fetch), which I think is even better than the `case` approach.
```ruby
def get_bucket_encryption(query)
  buckets = {
    'public' => OpenStruct.new(server_side_encryption_configuration: OpenStruct.new(rules: []))
  }

  if query[:bucket].eql? 'private'
    raise Aws::S3::Errors::ServerSideEncryptionConfigurationNotFoundError.new(nil, nil)
  end

  buckets.fetch(query[:bucket]) { raise Aws::S3::Errors::NoSuchBucket.new(nil, nil) }
end
```

clintoncwolfe:
Was Blocked by #2884, now unblocked

lib/resources/aws/aws_s3_bucket.rb
```diff
@@ -89,6 +94,16 @@ def fetch_bucket_policy
...
 
+  def fetch_bucket_encryption_configuration
+    catch_aws_errors do
+      begin
+        @has_default_encryption_enabled ||= !BackendFactory.create(inspec_runner).get_bucket_encryption(bucket: bucket_name).server_side_encryption_configuration.nil?
+      rescue Aws::S3::Errors::ServerSideEncryptionConfigurationNotFoundError
+        @has_default_encryption_enabled = false
+      end
+    end
+  end
```
TrevorBramble: A few things here. =^)

Line 100 is hard to read for how long it is. Going vertical with those method calls off `BackendFactory` would make it more readable.

We can do a single assignment to `@has_default_encryption_enabled` from whatever you return in the `catch_aws_errors` block.

```ruby
def fetch_bucket_encryption_configuration
  @has_default_encryption_configuration ||= catch_aws_errors do
    begin
      !BackendFactory.create(inspec_runner)
        .get_bucket_encryption(bucket: bucket_name)
        .server_side_encryption_configuration
        .nil?
    rescue Aws::S3::Errors::ServerSideEncryptionConfigurationNotFoundError
      false
    end
  end
end
```

test/integration/aws/default/build/s3.tf
```diff
@@ -85,6 +85,32 @@ output "s3_bucket_access_logging_not_enabled_name" {
...
 }
 
+resource "aws_s3_bucket" "default_encryption_enabled" {
+  bucket = "inspec-testing-defencrypt-enabled-${terraform.env}.chef.io"
+  acl    = "private"
+
+  server_side_encryption_configuration {
+      rule {
+        apply_server_side_encryption_by_default {
+          sse_algorithm     = "aws:kms"
```
TrevorBramble, line 95: [nitpick] Extra spacing between `sse_algorithm` and `=`.

lib/resources/aws/aws_s3_bucket.rb
```diff
@@ -35,6 +35,11 @@ def public?
       bucket_policy.any? { |s| s.effect == 'Allow' && s.principal == '*' }
   end
 
+  def has_default_encryption_enabled?
+    return unless @exists
```
TrevorBramble: [nitpick] Predicate methods should only ever return explicit `true` or `false` values. So `return false unless @exists` would be the more accurate guard statement.

TrevorBramble (commented):
Hi, thanks for your contribution!

Please consider the following suggestions.

jquick:
Hey @UranusBytes, we merged in the new terraform pins and you will need to rebase your branch. Sorry about that.

TrevorBramble (approved):
Thanks!

jquick (approved):
Thanks @UranusBytes

PR #2939, opened 2018-04-10
Ensure @params in shadow resource always has a valid value.

### LINKED ISSUE #2913: Breaking change in resource filters since 2.0.45
### 🎛 Description

We have profiles that we create and test with inspec 1.x; and now need to test and maintain for both inspec 1.x AND 2.x

inspec-2.1.21 via `bundle install`

### 🤔 Replication Case

The following control works in Inspec 1.51 and Inspec 2.0.45 but fails in Inspec 2.1

```
control "Ensure inactive password lock is 30 days or less" do
  title "Ensure inactive password lock is 30 days or less"
  users = shadow.filter(password:/[^!|*]/).params.map { |x| x['user'] }
  unless users.empty?
    users.each do |user|
      describe shadow.users(user) do
        its('inactive_days') { should cmp <= 30 }
      end
    end
  else
    describe 'user password' do
      skip 'no user with password detected'
    end
  end
end
```

### 💁 Possible Solutions

Make the filter method backward compatible

### 💻 Stacktrace

```
root@510dc6aa1821:/workdir# cd /workdir && bundle exec inspec check my_profile
[DEPRECATION] The shadow `users` property is deprecated and will be removed in InSpec 3.0.  Please use `user` instead.
bundler: failed to load command: inspec (/workdir/vendor/bundle/ruby/2.5.0/bin/inspec)
NoMethodError: undefined method `find_all' for nil:NilClass
  /workdir/vendor/bundle/ruby/2.5.0/gems/inspec-2.1.21/lib/resources/shadow.rb:82:in `block in filter'
  /workdir/vendor/bundle/ruby/2.5.0/gems/inspec-2.1.21/lib/resources/shadow.rb:79:in `each'
  /workdir/vendor/bundle/ruby/2.5.0/gems/inspec-2.1.21/lib/resources/shadow.rb:79:in `filter'
  my_profile/controls/my-controls.rb:3:in `block in load_with_context'
  /workdir/vendor/bundle/ruby/2.5.0/gems/inspec-2.1.21/lib/inspec/rule.rb:50:in `instance_eval'
  /workdir/vendor/bundle/ruby/2.5.0/gems/inspec-2.1.21/lib/inspec/rule.rb:50:in `initialize'
```

miah:
Add tests for method chained shadow resource with readable and unreadable shadow files.

Ensure @params always has a safe value, otherwise we may stacktrace when
unable to read /etc/shadow and invoked with method chaining.

My only complaint with this is that when the `/etc/shadow` file is unreadable, you won't see the error when chaining; you'll need to test the file or a specific user.

image

Fixes #2913

### DIFF
--- a/lib/resources/shadow.rb
+++ b/lib/resources/shadow.rb
@@ -34,7 +34,6 @@ module Inspec::Resources
 
     include FileReader
 
-    attr_reader :params
     attr_reader :lines
 
     def initialize(path = '/etc/shadow', opts = nil)
@@ -43,7 +42,7 @@ module Inspec::Resources
       @filters = opts[:filters] || ''
       @raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
       @lines = @raw_content.to_s.split("\n")
-      @params = @lines.map { |l| parse_shadow_line(l) }
+      params
     end
 
     filtertable = FilterTable.create
@@ -74,7 +73,7 @@ module Inspec::Resources
 
     def filter(query = {})
       return self if query.nil? || query.empty?
-      res = @params
+      res = params
       filters = ''
       query.each do |attr, condition|
         condition = condition.to_s if condition.is_a? Integer
@@ -121,10 +120,14 @@ module Inspec::Resources
       "/etc/shadow#{f}"
     end
 
+    def params
+      @params = @lines.nil? ? [] : @lines.map { |l| parse_shadow_line(l) }
+    end
+
     private
 
     def map_data(id)
-      @params.map { |x| x[id] }
+      params.map { |x| x[id] }
     end
 
     # Parse a line of /etc/shadow
--- a/lib/resources/shadow.rb
+++ b/lib/resources/shadow.rb
@@ -37,11 +37,9 @@ module Inspec::Resources
     attr_reader :params
 
     def initialize(path = '/etc/shadow', opts = {})
+      @opts = opts
       @path = path || '/etc/shadow'
-      @filters = opts[:filters] || ''
-      raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
-      @lines = raw_content.to_s.split("\n")
-      @params = set_params
+      @filters = @opts[:filters] || ''
     end
 
     filtertable = FilterTable.create
@@ -68,7 +66,7 @@ module Inspec::Resources
       i.entries.length
     }
 
-    filtertable.connect(self, :params)
+    filtertable.connect(self, :set_params)
 
     def filter(query = {})
       return self if query.nil? || query.empty?
@@ -117,7 +115,7 @@ module Inspec::Resources
     def lines
       warn '[DEPRECATION] The shadow `lines` property is deprecated and will be removed' \
         ' in InSpec 3.0.'
-      @lines
+      shadow_content.to_s.split("\n")
     end
 
     def to_s
@@ -127,12 +125,16 @@ module Inspec::Resources
 
     private
 
+    def shadow_content
+      @opts[:content] || read_file_content(@path, allow_empty: true)
+    end
+
     def set_params
-      Array(@lines).map { |l| parse_shadow_line(l) }
+      @params ||= Array(shadow_content.to_s.split("\n")).map { |l| parse_shadow_line(l) }
     end
 
     def map_data(id)
-      params.collect { |x| x[id] }
+      set_params.collect { |x| x[id] }
     end
 
     # Parse a line of /etc/shadow
--- a/lib/resources/shadow.rb
+++ b/lib/resources/shadow.rb
@@ -34,15 +34,14 @@ module Inspec::Resources
 
     include FileReader
 
-    attr_reader :lines
+    attr_reader :params
 
-    def initialize(path = '/etc/shadow', opts = nil)
-      opts ||= {}
+    def initialize(path = '/etc/shadow', opts = {})
       @path = path || '/etc/shadow'
       @filters = opts[:filters] || ''
-      @raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
-      @lines = @raw_content.to_s.split("\n")
-      params
+      raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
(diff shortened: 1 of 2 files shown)


jerryaldrichiii (approved):
Looks good to me! Thanks @miah!

lib/resources/shadow.rb
```diff
@@ -43,7 +42,7 @@ def initialize(path = '/etc/shadow', opts = nil)
       @filters = opts[:filters] || ''
       @raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
       @lines = @raw_content.to_s.split("\n")
-      @params = @lines.map { |l| parse_shadow_line(l) }
+      params
```
TrevorBramble: I'd really rather see this assignment preserved, rather than treating the setup as a side-effect of calling a method.

```ruby
def initialize
  @params = set_params(@lines)
end

def set_params(lines)
  lines.nil? ? [] : lines.map { |l| parse_shadow_line(1) }
end
```

lib/resources/shadow.rb
```diff
@@ -121,10 +120,14 @@ def to_s
       "/etc/shadow#{f}"
     end
 
+    def params
+      @params = @lines.nil? ? [] : @lines.map { |l| parse_shadow_line(l) }
+    end
```
TrevorBramble: Couple of directions this could go to make the point more obvious.

You could separate the guard statement:
```ruby
def params(lines)
  return [] if lines.nil?

  lines.map { |l| parse_shadow_line(1) }
end
```
And that maps directly to your logic in the ternary. But what I think you're actually concerned about is calling `map` on `nil`? So this would actually be more direct and clear:
```ruby
def params(lines)
  Array(lines).map { |l| parse_shadow_line(1) }
end
```
`Array(nil)` will produce the empty array, and pass through the `map` call to return what you expect, just that empty array. So there's one path through the code and all the needed safety.

lib/resources/shadow.rb
```diff
@@ -121,10 +120,14 @@ def to_s
...
+
     private
 
     def map_data(id)
-      @params.map { |x| x[id] }
+      params.map { |x| x[id] }
```
TrevorBramble: Consider using `collect` here instead of `map` to make it clear that you're collecting a new set of values from the input data. (Same for the parsed shadow lines in the `params` method.)

test/unit/resources/shadow_test.rb
```diff
@@ -86,6 +86,24 @@
     end
   end
 
+  describe 'when method chained' do
+    let(:unreadable_shadow) { load_resource('shadow', '/etc/unreadable_shadow') }
+    it 'can read /etc/shadow and #filter matches user with no password and inactive_days' do
```
TrevorBramble: Separate sections with newlines please. =^)

lib/resources/shadow.rb
TrevorBramble: While we're there, `@raw_content` and `@lines` are only used in the initializer (if you pass `@lines` to the `params` method) so why not make them local variables instead? You could even inline the lines parsing because there's no conditional logic involved.
```
def initialize(path = '/etc/shadow', opts = nil)
  # ...
  params(raw_content.to_s.split("\n"))
end
```

...And while we're there
```ruby
def initialize(path = '/etc/shadow', opts = nil)
  opts ||= {}
  # ...
end
```
This makes me so sad. Would you mind?
```ruby
def initialize(path = '/etc/shadow', opts = {})
  # ...
end
```

TrevorBramble (changes requested):
You asked for it! ;^)

lib/resources/shadow.rb
miah: I like this `@params = params(@raw_content.to_s.split("\n"))` but when I set up our params this way, when `filter` is called I'm still getting a nil value.

```
NoMethodError: undefined method `find_all' for nil:NilClass
  /Users/miah/projects/github/tmp/inspec/lib/resources/shadow.rb:80:in `block in filter'
```

So what I've done is preserved `@lines` and reference that from set_params.

TrevorBramble: Ah, I missed `def lines` elsewhere. Still, I'd pass it into the new method that makes your params and use it as a local variable there.

lib/resources/shadow.rb
```diff
@@ -35,15 +35,13 @@ class Shadow < Inspec.resource(1)
     include FileReader
 
     attr_reader :params
-    attr_reader :lines
 
-    def initialize(path = '/etc/shadow', opts = nil)
-      opts ||= {}
+    def initialize(path = '/etc/shadow', opts = {})
```
TrevorBramble, line 39: Thank you!

TrevorBramble (approved):

lib/resources/shadow.rb
```diff
@@ -35,15 +35,13 @@ class Shadow < Inspec.resource(1)
...
       @path = path || '/etc/shadow'
```
chris-rock: I am wondering if we could check if the file exist and throw an exception here? Otherwise we absorb the error.

lib/resources/shadow.rb
```diff
@@ -35,15 +35,13 @@ class Shadow < Inspec.resource(1)
...
       @filters = opts[:filters] || ''
-      @raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
-      @lines = @raw_content.to_s.split("\n")
-      @params = @lines.map { |l| parse_shadow_line(l) }
+      raw_content = opts[:content] || read_file_content(@path, allow_empty: true)
```
chris-rock: I think it would be beneficial to move the read logic out of the initialize function and do this during runtime not object creation time. This would allow us to do less during `inspec check` runs.

lib/resources/shadow.rb
miah: Definitely can do this, but the file resource we use for reading and content gathering also does this; but we do absorb the error when method chained.

TrevorBramble (approved):
<emoji>

jquick (approved):
Thanks @miah

PR #2941, opened 2018-04-10
Fixes configuration for Azure integrationt tests

dmccown:
Updates some calls in train so we can access the configuration options
for some integration tests.

Signed-off-by: David McCown <dmccown@chef.io>

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -162,7 +162,8 @@ namespace :test do
       sh("cd #{integration_dir}/build/ && terraform workspace new #{tf_workspace}")
 
       # Generate Azure crendentials
-      creds = Train.create('azure').connection.connect
+      connection = Train.create('azure').connection
+      creds = connection.options
 
       # Determine the storage account name and the admin password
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -162,7 +162,9 @@ namespace :test do
       sh("cd #{integration_dir}/build/ && terraform workspace new #{tf_workspace}")
 
       # Generate Azure crendentials
-      creds = Train.create('azure').connection.connect
+      connection = Train.create('azure').connection
+      connection.connect
```
TrevorBramble: That shouldn't be necessary? https://github.com/chef/train/blob/a931a4dccf8d01b0a99de28d5c17b2b484e6ce06/lib/train/transports/azure.rb#L50

dmccown: Ahh nice catch. I'll update.

dmccown: I updated and confirmed it works without the explicit connect.

TrevorBramble (approved):
Awesome!

jquick (approved):
Thanks @dmccown

Issue #2940, opened 2018-04-10
os_env resource returns only user's environment variable on Windows


omar-irizarry:
### 🎛 Description

🐛 bug report

os_env resource returns only the user's environment variable.  There is no way to verify system's environment variables

### 🌍 InSpec and Platform Version

InSpec Version 2.1.30
Windows 2016

### 🤔 Replication Case

Create variables as the ones in the screenshot below
![image](https://user-images.githubusercontent.com/25246117/38583531-48c70e0c-3ce1-11e8-8c19-3f4decd9266d.png)

create the following test:

```
describe os_env('TEMP') do
  its('content') { should eq 'C:\TEMP' }
end

```
run test and you will get the following output:

![image](https://user-images.githubusercontent.com/25246117/38583739-f0eb5bce-3ce1-11e8-83b5-0132408daa4e.png)

### 💁 Possible Solutions

Add a user parameter to os_env resource.

### 💻 Stacktrace

None

clintoncwolfe:
Fixed on #2945

PR #2945, opened 2018-04-11
os_env resource returns only user's environment variable on Windows

omar-irizarry:
Enhance os_env resource to deal with scenario explained in issue #2940 

os_env resource may be called with one parameter, this keeps the current functionality
```
describe os_env('variablename') do
  its('content') {should eq 'content' }
end
```

os_env resource may be called with two parameters
```
describe os_env('variablename', 'target') do
  its('content') {should eq 'content' }
end
```
target values may be system or user


jquick (approved):
Thanks @omar-irizarry! This is a great enhancement. Would you mind adding an example here?https://github.com/chef/inspec/blob/master/docs/resources/os_env.md.erb

lib/resources/os_env.rb
```diff
@@ -22,8 +22,15 @@ class OsEnv < Inspec.resource(1)
       end
     "
 
-    def initialize(env = nil)
+    def initialize(env = nil, target = nil)
       @osenv = env
+      @target = unless target.nil?
```
clintoncwolfe, line 27: An inline `unless` with a block dual-arm `if` block.   Yes!

clintoncwolfe (approved):
There's a conditional in there that is, um, unorthodox, but it's sounds.  I tested it on win2016 on an EC2 instance, all's well.  Thanks @omar-irizarry !

PR #2949, opened 2018-04-12
Document the inspec_version metadata field

clintoncwolfe:
Like it says, just document the inspec_version field used in inspec.yml .


jquick (approved):
Thanks @clintoncwolfe

PR #2958, opened 2018-04-13
aws_iam_policy statement search fix for degenerate policies

clintoncwolfe:
I found an odd bug in `aws_iam_policy`; it assumed all JSON policy structures would have an Array at position policy['Statement'].  Apparently, if you have exactly one statement, you are permitted to skip the Array, and simply place your statement directly at policy['Statement'], like this:

```
# arn:aws:iam::aws:policy/AWSCertificateManagerReadOnly
{
    "Version": "2012-10-17",
    "Statement": {
        "Effect": "Allow",
        "Action": [
            "acm:DescribeCertificate",
            "acm:ListCertificates",
            "acm:GetCertificate",
            "acm:ListTagsForCertificate"
        ],
        "Resource": "*"
    }
}
```

As `statement_count` and `have_statement` assume that is an Array, hijinks and stacktraces ensue.

This PR includes test cases to replicate the bug, and then fixes it.

### DIFF
--- a/lib/resources/aws/aws_iam_policy.rb
+++ b/lib/resources/aws/aws_iam_policy.rb
@@ -85,7 +85,7 @@ class AwsIamPolicy < Inspec.resource(1)
     return nil unless exists?
     # Typically it is an array of statements
     if policy['Statement'].is_a? Array
-      policy['Statement'].count    
+      policy['Statement'].count
     else
       # But if there is one statement, it is permissable to degenerate the array,
       # and place the statement as a hash directly under the 'Statement' key
@@ -138,7 +138,7 @@ class AwsIamPolicy < Inspec.resource(1)
     recognized_criteria
   end
 
-  def has_statement__normalize_criteria(criteria)    
+  def has_statement__normalize_criteria(criteria)
     # Transform keys into lowercase symbols
     criteria.keys.each do |provided_key|
       criteria[provided_key.downcase.to_sym] = criteria.delete(provided_key)
@@ -149,10 +149,10 @@ class AwsIamPolicy < Inspec.resource(1)
 
   def has_statement__normalize_statements
     # Some single-statement policies place their statement
-    # directly in policy['Statement'], rather than in an 
+    # directly in policy['Statement'], rather than in an
     # Array within it.  See arn:aws:iam::aws:policy/AWSCertificateManagerReadOnly
     # Thus, coerce to Array.
-    policy['Statement'] = [ policy['Statement'] ] if policy['Statement'].is_a? Hash
+    policy['Statement'] = [policy['Statement']] if policy['Statement'].is_a? Hash
     policy['Statement'].map do |statement|
       # Coerce some values into arrays
       %w{Action Resource}.each do |field|
--- a/lib/resources/aws/aws_iam_policy.rb
+++ b/lib/resources/aws/aws_iam_policy.rb
@@ -83,7 +83,14 @@ class AwsIamPolicy < Inspec.resource(1)
 
   def statement_count
     return nil unless exists?
-    policy['Statement'].count
+    # Typically it is an array of statements
+    if policy['Statement'].is_a? Array
+      policy['Statement'].count    
+    else
+      # But if there is one statement, it is permissable to degenerate the array,
+      # and place the statement as a hash directly under the 'Statement' key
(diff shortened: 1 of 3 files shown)


lib/resources/aws/aws_iam_policy.rb
```diff
@@ -83,7 +83,14 @@ def policy
...
     return nil unless exists?
-    policy['Statement'].count
+    # Typically it is an array of statements
+    if policy['Statement'].is_a? Array
+      policy['Statement'].count    
+    else
+      # But if there is one statement, it is permissable to degenerate the array,
+      # and place the statement as a hash directly under the 'Statement' key
+      return 1
+    end
```
TrevorBramble, line 93: I think this came up in another PR, but `Array()` is designed to return a given array, or wrap a non-array item so that you are certain to have an array. Seems like a good fit here! `Array(policy['Statement']).count`

Otherwise maybe treat the not-an-array case as a condition to guard against and sling it under the existing guard statement?
```ruby
def statement_count
  return nil unless exists?
  # AWS spec allows either a single item or an array of items
  return 1 unless policy['Statement'].respond_to?(:count)

  policy['Statement'].count
end
```

(The comment is just How The AWS API Always Works, right? I think this came up a few weeks ago. So I wonder if the comment is necessary/helpful.)

lib/resources/aws/aws_iam_policy.rb
```diff
@@ -131,7 +138,7 @@ def has_statement__validate_criteria(raw_criteria)
     recognized_criteria
   end
 
-  def has_statement__normalize_criteria(criteria)
+  def has_statement__normalize_criteria(criteria)
```
TrevorBramble: Trailing whitespace

lib/resources/aws/aws_iam_policy.rb
```diff
@@ -141,6 +148,11 @@ def has_statement__normalize_criteria(criteria)
   end
 
   def has_statement__normalize_statements
+    # Some single-statement policies place their statement
+    # directly in policy['Statement'], rather than in an 
+    # Array within it.  See arn:aws:iam::aws:policy/AWSCertificateManagerReadOnly
+    # Thus, coerce to Array.
+    policy['Statement'] = [ policy['Statement'] ] if policy['Statement'].is_a? Hash
```
TrevorBramble: `Array()` =^)

TrevorBramble (changes requested):
Good find. Highly recommend using `Array()` over the conditions and clauses, but trailing whitespace is the only blocker. =^)

lib/resources/aws/aws_iam_policy.rb
clintoncwolfe: I do use Array() in other places in the code, and even in this file.  It's useful.  But in this case, it does not behave as desired:
```
irb(main):001:0> h = {a: 1, b: 2}
=> {:a=>1, :b=>2}
irb(main):002:0> Array(h)
=> [[:a, 1], [:b, 2]]
```
Desired result - wrap the hash in an array: `[{:a=>1, :b=>2}]`

Hash does in fact respond to count.  Returns the number of keys.  So, if policy['Statement'] is a hash with keys like 'Action', 'Resource, 'Effect' - we'd return 3 here, which is incorrect.  That's why the conditional is there, and I'm doing a class check. I'd normally do `kind_of?`, but rubocop wants us to do `is_a?`.  I suppose I could duck-type on 'key?' or something?

I do feel the comment is necessary.  The official grammar for the Policy JSON is daunting (at least to me), and without the comment, the logic would seem a bit bizarre.

lib/resources/aws/aws_iam_policy.rb
clintoncwolfe: No, this is where it doesn't work.  I tried it with Array here, and got the array-of-arrays-containing-keys-and-values and discussed above.

lib/resources/aws/aws_iam_policy.rb
TrevorBramble: Ahhh, right, right. I didn't realize the single-item case was a hash. (See also `respond_to?(:count)`.)

TrevorBramble (approved):

jquick (approved):
Thanks @clintoncwolfe

PR #2962, opened 2018-04-13
Policy statement search: don't stacktrace on missing field

clintoncwolfe:
If a criterion was provided to the `aws_iam_policy` `have_statement` system, but one or more statements was missing that key, a run abort and stacktrace would occur.

Because required fields vary with AWS service, and mutually exclusive fields exist (for example, Action and NotAction), this condition can arise.  It was discovered while looping over all AWS-managed profiles, and blowing up on a statement that had no Action component (because it had a NotAction component).

This change will cause the search to not-match if a criterion is presented for a statement that does not provide the field.


jquick (approved):
Thanks @clintoncwolfe

PR #2963, opened 2018-04-16
Add A2 support to the inspec-compliance toolset

jquick:
This is the initial support for Automate 2 with the compliance tools. This supports login/upload/download/exec/logout calls accordingly.

### DIFF
--- a/lib/bundles/inspec-compliance/api.rb
+++ b/lib/bundles/inspec-compliance/api.rb
@@ -4,6 +4,7 @@
 
 require 'net/http'
 require 'uri'
+require 'json'
 
 require_relative 'api/login'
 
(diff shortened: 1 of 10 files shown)


lib/bundles/inspec-compliance/api.rb
```diff
@@ -18,12 +19,15 @@ class API
     # return all compliance profiles available for the user
     # the user is either specified in the options hash or by default
     # the username of the account is used that is logged in
-    def self.profiles(config)
+    def self.profiles(config) # rubocop:disable PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/AbcSize, Metrics/MethodLength
```
TrevorBramble, line 22: You're getting dinged for these things because it's past time to treat API as an abstract parent class and make concrete subclasses for the differing targets; `API_Compliance`, `API_Automate`, and `API_Automate2`. Almost all of the conditional code you had to expand here goes away.

jquick: Yeah, I really wanted to break up a lot of this code into sub classes like this. I will take a shot at it and see if I can fit in into the timeline.

jquick: Started down this path a bit tonight, its much cleaner but there is a LOT to do here. We call the compliance stuff all over and will need to be re-done with instance classes and caching. I will create a ticket to get this refactor in as its needed but out of scope for now.

jquick: https://github.com/chef/inspec/issues/2964 added, thanks for the feedback.

miah (approved):

jquick:
Confirmed with @vjeffrey that these changes are working as expected.

chris-rock (approved):
Thank you @jquick

PR #2968, opened 2018-04-17
Upgrade Terraform version pins for integration testing

### LINKED ISSUE #2884: Check for Terraform update for integration tests

clintoncwolfe:
Moves version pin for Terraform to be `~> 0.11.0`, azure provider to `1.3.3`, and aws to `1.13`.  I found the aws provider v`1.14` panicked; so I backed down to 1.13 without incident.

Also fixed a handful of TF errors which caused things to fail idempotency.

Fixes #2884 
Unblocks #2937

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ profile-1.0.0.tar.gz
 terraform.tfstate*
 terraform.tfstate.backup
 inspec-azure.plan
+inspec-aws-*.plan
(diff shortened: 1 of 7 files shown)


jquick (approved):
Very nice, thanks @clintoncwolfe !

miah (approved):
Awesome! Thanks @clintoncwolfe

PR #2970, opened 2018-04-17
Amazon linux service mgmt detection

meringu:
Looks like when support for Amazon Linux 2 was added, detection for Amazon 1 had this bug introduced:

```ruby
# Current behavior
> '2015.03'.start_with?('20\d\d')
=> false

# Proposed behavior
> '2015.03 =~ /^20\d\d/
=> 0
```

This change also adds a test for both Amazon linux 1 and 2 service detection. I haven't touched the tests for inspec before, please let me know If I should improve my change.

### DIFF
--- a/lib/resources/service.rb
+++ b/lib/resources/service.rb
@@ -162,7 +162,7 @@ module Inspec::Resources
       elsif %w{aix}.include?(platform)
         SrcMstr.new(inspec)
       elsif %w{amazon}.include?(platform)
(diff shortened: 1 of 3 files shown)


jquick (approved):
Thanks for this @meringu ! Nice enhancement

miah (approved):
Thank you @meringu !

meringu:
Thanks @jquick and @miah!

PR #2991, opened 2018-04-25
Ease access to hierarchical attributes

ghost:
### 🎛 Description

Inspec allows to pass various sources of attributes and pass them to the executed profiles. The mentioned sources shall be in YAML format. 

InSpec version 2.1.54.1 on any platform

### 🤔 Replication Case

YAML format allows to define attributes in an hierarchical manner, example

```
pod:
  regional: de
  timezone: UTC
host:
  name: foobarbaz
```

Accessing those attributes may look as follows currently

```
pod = attribute('pod', default: {}, description: 'The POD details')
regional = pod.fetch('regional', 'en')
timezone = pod.fetch('timezone', 'GMT+2')

```

### 💁 Possible Solutions

I personally feel uncomfortable with that as it is lengthy and the description may lack required details. Due to that I propose to change the behavior the way it allows to access the specific attributes down the hierarchy without breaking existing behavior: 

```
regional = attribute('pod:regional', default: 'en', description: 'The POD region')
timezone = attribute('pod:timezone', default: 'GMT+2', description: 'The POD timezone')

```

This pull request attempts to ensure that.

### DIFF
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -189,7 +189,14 @@ module Inspec
       # we need to return an attribute object, to allow dermination of default values
       attr = Attribute.new(name, options)
       # read value from given gived values
-      attr.value = @conf['attributes'][attr.name] unless @conf['attributes'].nil?
+      if @conf['attributes']
+        if @conf['attributes'][attr.name]
+          attr.value = @conf['attributes'][attr.name]
+        else
+          # Test if both, the given name and attributes are hierarchical and pick up the value down the hierarchy
+          attr.value = attr.name.split(':').inject(@conf['attributes'], :[])
+        end
+      end
       @attributes.push(attr)
       attr.value
     end
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -187,8 +187,6 @@ module Inspec
 
     def register_attribute(name, options = {})
       # we need to return an attribute object, to allow dermination of default values
-      attr = Attribute.new(name, options)
-      # read value from given gived values
       attr = Attribute.lookup(name, options, @conf['attributes'])
       @attributes.push(attr)
       attr.value
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -7,7 +7,6 @@ require 'inspec/resource'
 require 'inspec/library_eval_context'
 require 'inspec/control_eval_context'
 require 'inspec/require_loader'
-require 'json'
 require 'securerandom'
 require 'inspec/objects/attribute'
 
@@ -190,14 +189,7 @@ module Inspec
       # we need to return an attribute object, to allow dermination of default values
       attr = Attribute.new(name, options)
       # read value from given gived values
-      if @conf['attributes']
-        if @conf['attributes'][attr.name]
-          attr.value = @conf['attributes'][attr.name]
-        else
-          # Test if both, the given name and attributes are hierarchical and pick up the value down the hierarchy
-          attr.value = name.split(':').inject(JSON.parse(@conf['attributes'].to_json), :[])
-        end
-      end
+      attr = Attribute.lookup(name, options, @conf['attributes'])
       @attributes.push(attr)
       attr.value
     end
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -7,6 +7,7 @@ require 'inspec/resource'
 require 'inspec/library_eval_context'
 require 'inspec/control_eval_context'
 require 'inspec/require_loader'
+require 'json'
 require 'securerandom'
 require 'inspec/objects/attribute'
 
@@ -194,7 +195,7 @@ module Inspec
           attr.value = @conf['attributes'][attr.name]
         else
           # Test if both, the given name and attributes are hierarchical and pick up the value down the hierarchy
-          attr.value = attr.name.split(':').inject(@conf['attributes'], :[])
+          attr.value = name.split(':').inject(JSON.parse(@conf['attributes'].to_json), :[])
         end
       end
       @attributes.push(attr)
--- a/test/unit/profiles/profile_context_test.rb
+++ b/test/unit/profiles/profile_context_test.rb
@@ -451,21 +451,23 @@ describe Inspec::ProfileContext do
     let(:attrs_profile) {Inspec::ProfileContext.new(nil, backend, {'attributes' => attrs})}
 
     it 'can handle hierarchical attributes' do
-      assert_equal(attrs_profile.register_attribute('foo'), attrs['foo'])
-      assert_equal(attrs_profile.register_attribute(1337), attrs[1337])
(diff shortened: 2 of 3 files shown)


jquick:
Thanks for this @afischer-opentext-com , do you mind adding a test here https://github.com/chef/inspec/blob/master/test/unit/profiles/profile_context_test.rb for this use case?

clintoncwolfe:
Thanks for the contribution!  I love the idea of making this easier and more obvious.

However, I think we need to discuss this syntax.  I'm not sure this is the right approach.  

First, I don't think adding logic in profile_context is quite the right place - I'd think it would be new accessors on [Attribute](https://github.com/chef/inspec/blob/master/lib/inspec/objects/attribute.rb).

Secondly, if you're putting hashes in YAML, you may also be putting in arrays, nested hashes, etc.  It's not clear to me that this would support arrays - would `some:2:thing` work?  It seems to assume all path components are strings, not integers.

Finally, if I think an attribute is a hash or other structure, wouldn't I want to access it like one?

```
hash_attr = attribute('some_key', ...)
value = hash_attr[:second_key][2][:yet_another_key]
```

In other words, I think this functionality could be exposed (and documented) via `:[]`. That would be less surprising, and potentially more obvious.

Again, thanks for the contribution, it's got us thinking about how to make things easier!

ghost:
Thanks much for the constructive feedback. 

For completeness and my learning experience I added a unit test case as requested and changed the code the way keys are available by accessing their string value (using JSON serialization). 

I agree that arrays and hashes without keys of type String may be an issue. I used the method register_attribute() as the change here is imho least invasive. With a better understanding of where attributes are used I would have added a class Attributes which would look for a passed hierarchical attribute key. Without knowing if the idea is accepted that code change is quite large :)

I understand that the proposed solution will not be implemented that way. The good news is that you like the idea at least. Not sure how to proceed, feel free to do whatever you believe is right.

clintoncwolfe:
To boil down what I suggest, can you imagine implementing something like this?  Note that the Attibute class already exists [here](https://github.com/chef/inspec/blob/master/lib/inspec/objects/attribute.rb); you'd just be adding a method to it.

```
# In attribute.rb

def [](key) 
  # Probably do something involving fetch()
  # You could return a bare value, or possibly one dressed up as an Attribute
end
```

test/unit/profiles/profile_context_test.rb
```diff
@@ -424,4 +424,49 @@ def get_rule
       }.must_raise LoadError
     end
   end
+
+  describe 'attribute loading' do
```
clintoncwolfe, line 428: Great unit test!

ghost:
I changed the code as requested and added a method named _lookup_ which is now used by the ProfileContext and which can also be used to query a nested Attribute value. This is now more code, mainly due to the support of Arrays during the lookup (as mentioned this was an issue before) and due to the improvement that data types are now maintained. 

May you have another look?

ghost:
Hi @clintoncwolfe, what can I do to move this forward a bit? :) Are there further optimizations you would propose?

clintoncwolfe:
👀

clintoncwolfe:
Hey @afischer-opentext-com , we'll need this to have merge messages signed off, using the `-s` option to git.  Have a look at our [DCO policy](https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

ghost:
I forgot it with my last commit, added the sign-off now. Thank you!

clintoncwolfe:
It looks like there are a number of Rubocop violations - that's what we use for code style checking.  Try running
```
bundle exec rubocop lib/inspec
```

lib/inspec/profile_context.rb
```diff
@@ -189,7 +189,7 @@ def register_attribute(name, options = {})
       # we need to return an attribute object, to allow dermination of default values
       attr = Attribute.new(name, options)
```
clintoncwolfe: It looks like `attr` is being set twice, and the first value discarded.  Can the first be removed?

clintoncwolfe (changes requested):
I'm working on a set of functional tests for this. The behavior when it actually parses a YAML file is different.

clintoncwolfe:
I'm afraid this change may not be needed after all.  It appears that Attribute already supports `[]`, though it has some rough edges. @jquick ?

ghost:
Hi, and thanks for having another look! What I did now:
* I fixed the bug with the duplicate attr assignment. 
* I fixed all rubocop messages where applicable. 
  * I did not fix but suppress the Exception suppression warning in regards to the Arrays key index. 
  * I refactored the lookup method the way Rubocop no more complains about its complexity. 

Yes, the [] method existed and it still exists. That is already indicated in the example of the description. I used fetch there, as it is more false tolerant and does not simply let the profile die in case a key does not exist. The difference is that the "new" [] method also supports the "foo:bar:baz" syntax. If nothing is found, the old behavior comes into play (currently line 99 and below). 

Btw, this PR would also solve the issue with the Symbols as YAML keys as mentioned in #3157.

clintoncwolfe:
OK, I see.  I was confused: when this PR was opened, I was under the impression there was _no way_ to access sub-keys of hashes or sub-elements of arrays.  But with `[]` already present, clearly there is a way to access sub-elements.

I do not feel introducing a new syntax for data access is advisable.  It breaks expectations about how Ruby works, and we want people to be able to use pre-existing knowledge of either Ruby in general, or Chef in particular.  So, I'm afraid we have to reject this proposal of using "foo:bar:baz" syntax.

If you are concerned about failing when accessing a nonexistent sub-key, I suggest you look into [`dig()`](http://ruby-doc.org/core-2.5.0/Hash.html#method-i-dig), and check to see if that works, or could be made to work, on Attribute.  If you choose to add this, please start a separate PR, branching from current master.

If you are concerned about using symbols as keys, keep in mind that in the real world (outside of a unit test) the attribute files are being parsed by a YAML parser, which always stringifies the keys.  So, I believe the keys will always be strings, except for array indicies.

Thanks for all the effort on this, and I'm sorry that we can't accept the change.  If you have any other ideas for how to improve InSpec, I highly recommend you open an issue, and we can discuss it before work begins.  We value your time and hope you contribute more in the future!

ghost:
Understood.

The motivitation for this to reduce the amount of code needed in the profiles, to improve their maintainability. This optional piece would have not broke the Ruby feature set and as the specific feature (Attributes) relates to InSpec, the documentation would still need to be consulted first. 
I understand the point with the other Chef products. That's of course something I can't coordinate ;-)
Finally, no argument but fyi, Saltstack folks have something [similar](https://docs.saltstack.com/en/2017.7/ref/modules/all/salt.modules.pillar.html#salt.modules.pillar.get). 

Anyways, at least it was a great learning experience. Thank you and cu later! :-)

P.S. For completeness - I will remove the branch - I attach the diff to this PR, in case someone should be interested.


[2991.diff.txt](https://github.com/inspec/inspec/files/2122445/2991.diff.txt)

PR #3000, opened 2018-04-27
Update Habitat plan

jerryaldrichiii:
This updates the included Habitat plan to do the following:
  - Include binaries needed for certain resources (Example: `curl`)
  - Use `gem install/build` instead of Bundler
  - Use a wrapper binary to ensure GEM_HOME and GEM_PATH are correct
  - Perform build/install steps in a cache directory instead of `/src`

Many thanks to @miah @tduffield!

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -15,9 +15,17 @@ pkg_deps=(
   core/libxslt
   core/net-tools
   core/ruby
+
+  # Needed for some InSpec resources
+  core/bind
+  core/curl
+  core/docker
+  core/less
+  core/mysql-client
+  core/netcat
+  core/postgresql
 )
 pkg_build_deps=(
-  core/bundler
   core/gcc
   core/make
   core/readline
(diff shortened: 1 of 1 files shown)


jquick (approved):
Thanks @jerryaldrichiii

habitat/plan.sh
```diff
@@ -15,9 +15,17 @@ pkg_deps=(
...
   core/ruby
+
+  # Needed for some InSpec resources
+  core/bind
+  core/curl
+  core/docker
+  core/less
+  core/mysql-client
+  core/netcat
+  core/postgresql
```
chris-rock, line 26: Do we need a full pg server or just the clients?

vjeffrey (approved):
yes please!

habitat/plan.sh
tduffield: Habitat doesn't currently have a package that contains only the clients. Could be worth making one though...

PR #3007, opened 2018-04-30
cmp should recognise a string being a negative int

### LINKED ISSUE #3005: cmp does not recognise a string being a negative int
### 🎛 Description

`'-1' should cmp <= -1` returns false when it should be true

2.1.54 from bundle install

### 💁 Possible Solutions

I think it's because this line [here](https://github.com/chef/inspec/blob/master/lib/matchers/matchers.rb#L214) does not account for the string representing a negative number - I'll get a PR up that accounts for this and adds a unit test 

```
def integer?(value)
    !(value =~ /\A0+\Z|\A[1-9]\d*\Z/).nil?
end
```

james-stocks:
fixes #3005

`cmp` will now compare a negative integer string (e.g. `-'12'`) as the number -12. Previously the `-` character caused it not to be identified as an integer.

Includes new integration tests

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -211,7 +211,7 @@ end
 RSpec::Matchers.define :cmp do |first_expected| # rubocop:disable Metrics/BlockLength
 
   def integer?(value)
-    !(value =~ /\A0+\Z|\A[1-9]\d*\Z/).nil?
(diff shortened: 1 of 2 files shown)


jquick (approved):
Looks good to me! Thanks @james-stocks

clintoncwolfe (approved):
Looks and works well.  Thanks!

PR #3008, opened 2018-04-30
Split inspec into a core gem.

miah:
This adds a `inspec-core.gemspec` to allow InSpec to be dependent on `train-core` which is distributed without support for additional cloud providers.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,12 +2,14 @@
 # encoding: utf-8
 
 require 'bundler'
-require 'bundler/gem_tasks'
+require 'bundler/gem_helper'
 require 'rake/testtask'
 require 'passgen'
 require 'train'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
+
+Bundler::GemHelper.install_tasks name: 'inspec'
 
 def prompt(message)
   print(message)
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -8,8 +8,7 @@ Gem::Specification.new do |spec|
   spec.authors       = ['Dominik Richter']
   spec.email         = ['dominik.richter@gmail.com']
   spec.summary       = 'Just InSpec'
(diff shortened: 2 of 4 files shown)


Rakefile
```diff
@@ -2,13 +2,15 @@
...
 require 'bundler'
-require 'bundler/gem_tasks'
+require 'bundler/gem_helper'
 require 'rake/testtask'
 require 'passgen'
 require 'train'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
 
+Bundler::GemHelper.install_tasks name: 'inspec'
```
clintoncwolfe, line 12: Would it make sense to have a Rake task for building the inspec-core gem, as well?

inspec-core.gemspec
```diff
@@ -0,0 +1,43 @@
...
+  spec.email         = ['dominik.richter@gmail.com']
+  spec.summary       = 'Just InSpec'
+  spec.description   = 'Core InSpec, local support only. See `inspec` for full support.'
+  spec.homepage      = 'https://github.com/chef/inspec'
+  spec.license       = 'Apache-2.0'
+
+  spec.files = %w{README.md MAINTAINERS.toml MAINTAINERS.md LICENSE
+                  inspec-core.gemspec Gemfile CHANGELOG.md} +
+               Dir.glob('{bin,docs,examples,lib}/**/*', File::FNM_DOTMATCH)
+                  .reject { |f| File.directory?(f) || f =~ /aws|azure|gcp/ }
```
clintoncwolfe, line 18: We should (eventually) make this regex more obvious - possibly make it a variable or env var.  This will be a brittle point.

clintoncwolfe (approved):
Just a couple of maintainability comments, nothing to block for today.  Great work!

jquick (approved):
Thanks @miah !

PR #3011, opened 2018-05-02
Fix the A2 compliance ssl flag

jquick:
The current code uses the old A1 headers which uses `ssl_verify_mode`. The new post request requires `verify_mode`.


clintoncwolfe (approved):
Looks good - thanks @jquick !

miah (approved):

PR #3015, opened 2018-05-03
1.x - Adjust version test to allow for inspec 2.x being available

james-stocks:
The test for `inspec version` fails on the 1.x branch because it does not allow for the extra "new version available" text that correctly appears now that inspec 2.x is released.
This commit adjusts the test to expect a newer version notification.


miah (approved):

jquick (approved):
Thanks

Issue #2774, opened 2018-03-02
Windows Feature resource incorrectly sees features as installed


tas50:
### 🎛 Description

This a 🐛 bug report

The windows feature matcher currently sees InstallPending state as being installed. InstallPending isn't installed until a reboot so auditing that the features are installed will lead you to believe they are in fact fully installed when they're not.

### 🌍 InSpec and Platform Version

Inspec 1.51.21
Windows 2016

### 🤔 Replication Case

Install a feature like SNMP that requires a reboot

### 💁 Possible Solutions

???

jquick:
@tas50 - Thanks for reporting this!

kekaichinose:
__What’s happening? Why was this issue closed?__
This issue was closed due to some much needed review of legacy issues or issues that were spawned in older versions of InSpec, i.e. < v3.

__Why do I care?__
You would care about this if this was an issue that you are still seeing and/or feel needs to be addressed in the current version of InSpec.

__What do I need to do?__
If this issue is no longer important, no further action is necessary. However, if you think this is something that should be addressed, please open a new issue and refer to the original issue in the description.

PR #3020, opened 2018-05-04
Correct Windows Feature state detection

akacodemonkey:
When using inspec on Windows 2016 server found that the windows feature always marked feature as not installed.  Looked into it and found the command doesn't import server manager.  This changes loads server manager on init if required.  The Get-WindowsFeature then works and correctly identifies a WIndows Feature as installed Get-WindowsFeature

Edit (@clintoncwolfe): Move title to description

### DIFF
--- a/lib/resources/windows_feature.rb
+++ b/lib/resources/windows_feature.rb
@@ -40,6 +40,10 @@ module Inspec::Resources
       @feature = feature
       @cache = nil
 
+	  module_cmd = "if (!(get-module servermanager)){ import-module servermanager}"
+      cmd = inspec.command(module_cmd)
+	  cmd.stdout
+	  
       # verify that this resource is only supported on Windows
       return skip_resource 'The `windows_feature` resource is not supported on your OS.' if !inspec.os.windows?
     end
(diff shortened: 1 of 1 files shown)


clintoncwolfe:
Hi @akacodemonkey ! Thanks so much for your contribution, and welcome to the InSpec contributors!

I have one question, and two small changes needed.

Please review the [Developer Certification of Origin (DCO)](https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).  This certifies that we can accept the code without impacting licensing.  To add your DCO attestation, please do an amend commit with the `-s` (sign-off) flag set. On the command line, that looks like:
```
git commit --amend -s --no-edit
git push -f
```

Second, we use Rubocop to handle code formatting.  Could you run rubocop on your code?  `bundle exec rake lint` will show you the issues, and then it's usually just a matter of whitespace.  Once you've done that, commit and push again.

Finally, this is the actual question - we have an open issue that says we detect Pending features as installed, #2774. Do you think this might be related?    Could you add a test checking?

Thanks again!

akacodemonkey:
Doesn't cover the pending (yet marked as installed) I'm afraid.  Logically if its pending it is installed, so maybe the wrong check being performed?

akacodemonkey:
Doesn't cover the pending issue I'm afraid.  Logically if pending it is installed, just requires a reboot for the OS to be able to us it, so maybe the wrong check being performed

PR #3022, opened 2018-05-04
Fix the A2 vendoring with depends on the A2 server

jquick:
There is a issue pulling the profile info when vendoring on the A2 server. This fixes the issue by pulling the profile owner and id from the profile tag instead of the url.


jerryaldrichiii (approved):
Looks good to me, thanks @jquick!

clintoncwolfe (approved):
Thanks @jquick

PR #3029, opened 2018-05-08
Enhance groups resource with members property

Happycoil:
Adds a `members` property to the `groups` resource.

Discussed a while ago in https://github.com/chef/inspec/issues/1218

### DIFF
--- a/lib/resources/groups.rb
+++ b/lib/resources/groups.rb
@@ -107,7 +107,7 @@ module Inspec::Resources
     end
 
     def gid
-      flatten_entry(group_info, "gid")
+      flatten_entry(group_info, 'gid')
     end
 
     # implements rspec has matcher, to be compatible with serverspec
@@ -117,7 +117,7 @@ module Inspec::Resources
 
     def members
       return unless inspec.os.windows?
-      flatten_entry(group_info, "members")
+      flatten_entry(group_info, 'members')
     end
 
     def local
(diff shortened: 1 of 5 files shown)


jquick (approved):
Nice enhancement! Thanks @Happycoil

jquick (changes requested):
Actually, can we flatten the members? Would be really slick if something like this would work:
screen shot 2018-05-10 at 11 09 32 am

Happycoil:
@jquick it is [flattening](https://github.com/chef/inspec/pull/3029/files/4049af788e6c2c92ac39b94f3d5bac9d365cf585#diff-2569a99ba47f832af43b95a79118d3f2R134) in the `group` resource, not sure if it makes sense to also flatten the property in `groups`?

```
  Group Administrators
     ✔  members should include "Administrator"
  Groups with name == "Administrators"
     ×  members should include "Administrator"
     expected [["Administrator", "vagrant"]] to include "Administrator"
```

That said, [the spec](https://github.com/chef/inspec/pull/3029/files/4049af788e6c2c92ac39b94f3d5bac9d365cf585#diff-c65aa11000aab57e2d0f91b5c9079351R60) should probably test `should include 'Administrator'` rather than `should cmp ['Administrator']`.

Happycoil:
Looks like one of the builds stalled.

jquick (approved):
Ah yes, I see now. Thanks for clarifying this. This looks good to me.

miah (approved):
Thanks @Happycoil

PR #3031, opened 2018-05-09
mssql_session default port and local_mode

Happycoil:
This attempts to address issues raised in https://github.com/chef/inspec/issues/2948

It introduces a new `local_mode` parameter to avoid breaking people relying on the default host. It sets the default value of `port` to `nil`, as this shouldn't break anyone.

It also throws `ResourceFailed` instead of warning (https://github.com/chef/inspec/issues/1205)


lib/resources/mssql_session.rb
```diff
@@ -29,15 +29,18 @@ class MssqlSession < Inspec.resource(1)
...
       @password = opts[:password] || opts[:pass]
       if opts[:pass]
         warn '[DEPRECATED] use `password` option to supply password instead of `pass`'
       end
-      @host = opts[:host] || 'localhost'
-      @port = opts[:port] || '1433'
+      @local_mode = opts[:local_mode]
+      unless local_mode?
+        @host = opts[:host] || 'localhost'
+        @port = opts[:port]
```
jquick: While this is probably fine as `sqlcmd` defaults to `1433` I think i'd like to keep the explicit set incase we change what this uses under the hood.

jquick (commented):
Very nice addition @Happycoil !

lib/resources/mssql_session.rb
Happycoil: We'd want to retain the ability to set port to nil to support dynamic ports though. Forgive my ignorance, but does ruby have an equivalent of `$PSBoundParameters` from powershell so we can tell if a parameter is explicitly passed?

jquick: Something like:
```
opts[:port] = 1433 unless opts.key?(:port)
```
should only set it if the key is not previously defined.

Happycoil:
@jquick I think this should work. We have to explicitly pass `nil` now, otherwise the port will be set to `1433`.

jquick (approved):
Thanks @Happycoil !

miah (approved):
Thank you @Happycoil

jquick:
AppVeyor build looks fine https://ci.appveyor.com/project/chef/inspec/build/1.0.6378, Overriding checks.

PR #3036, opened 2018-05-10
AWS resource permission docs

miah:
Here is some documentation for required permissions for the AWS resources. 

I've also removed some stray spaces =)

### DIFF
--- a/docs/resources/aws_cloudtrail_trail.md.erb
+++ b/docs/resources/aws_cloudtrail_trail.md.erb
@@ -150,6 +150,6 @@ The test will pass if the identified trail has log file integrity validation is 
 
 ## AWS Permissions
 
-Your IAM User will need the `cloudtrail:DescribeTrails` action set to allow.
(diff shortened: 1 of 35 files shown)


jquick (approved):
Fantastic work @miah !

clintoncwolfe (changes requested):
Very good work and very thorough!  Thanks @miah !

My only request would be to adjust the phrasing on the sentences like:
```Your IAM User will need the `cloudtrail:DescribeTrails` action set to allow.```

A couple of things stick out, there:
 * It took me a bit to realize "set to allow" means that you need permissions with Effect set to Allow.  Adding capitalization and the word Effect should clarify that.
* The credentials may or may not represent an IAM User - it could be an assumed Role, or a temporary STS creds set.  You could use the generic term "Principal", or some other generic phrasing. (using the term IAM User might get confusing when discussing perms on `aws_iam_user` and related inspec resources)

I think that could be changed with a pair of global replacements - if it's a lot of effort, I don't think it's worth it.

clintoncwolfe (approved):

PR #3037, opened 2018-05-10
Add git dependency to habitat plan.

phiggins:
I ran into the same error as https://github.com/chef/inspec/issues/1270 today while using the inspec hab package. Since there are already dependencies on things like mysql and postgres to support various inspec features, git seemed like something that should be there.


miah (approved):
Thanks @phiggins !

jquick (approved):
Thanks @phiggins

PR #3038, opened 2018-05-10
Backport ci verify scripts

miah:
These are run after omnibus builds to verify that inspec is functional.

https://github.com/chef/inspec/pull/2732


jquick (approved):
Thanks @miah

PR #3040, opened 2018-05-11
handle nil properties in iis_site

Happycoil:
As it stands we get a ``undefined method `[]' for nil:NilClass`` error if the following test is run after deleting Default Web Site:

```
describe iis_site("Default Web Site") do
  its('bindings') { should eq nil }
end
```

This change returns `nil` instead of trying to index into `nil`.

It also fixes a typo in iis_site_spec.

### DIFF
--- a/lib/resources/iis_site.rb
+++ b/lib/resources/iis_site.rb
@@ -40,19 +40,19 @@ module Inspec::Resources
     end
 
     def app_pool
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @Happycoil

miah (approved):
Thank you @Happycoil

Issue #2786, opened 2018-03-06
Documentation for http resource has an error


jekriske-lilly:
### 🎛 Description

This a 🐛 bug report 

https://github.com/chef/inspec/blob/master/docs/resources/http.md.erb#L157
this is a boolean true or false not the string 'true' or 'false'

### 🌍 InSpec and Platform Version

1.45.13 MacOS and Linux

### 🤔 Replication Case
see https://github.com/chef/inspec/blob/master/docs/resources/http.md.erb#L157
`ssl_verify: 'true') do`

### 💁 Possible Solutions

Line should read:
`ssl_verify: true) do`

### 💻 Stacktrace
When run through kitchen (This error needs work btw):

```
Test the main page (undefined method `lines' for nil:NilClass)
     ×  http GET on https://localhost:443 status
     undefined method `lines' for nil:NilClass
```

when run directly (error much more clear):
```

Test the main page (**SSL_connect returned=1 errno=0 state=error: certificate verify failed**)
     ×  http GET on https://localhost:443 status
```

jquick:
@jekriske-lilly - Thanks for this!

PR #3043, opened 2018-05-11
Fix http resource documentation #2786

miah:
true should not be a string.

Fixes https://github.com/chef/inspec/issues/2786


clintoncwolfe (approved):
Thanks @miah - does what it says!

jquick (approved):
Thanks @miah

PR #3044, opened 2018-05-13
Fix matcher output when the args to the matcher are a hash for two resources

### LINKED ISSUE #2925: Some reporters display an empty Hash for matchers that take Hash arguments
### 🎛 Description

🐛 bug report

If a matcher takes a hash as arguments, the output displays an empty hash in both pass and failure of tests.  This makes it hard to identify which tests are failing, and very hard to determine why.

cli, documentation, html, progress, and json-rspec exhibit this behavior.

Preferred behavior would be to pretty-print the hash contents.

json and json-min reporters appear to be immune to this - here's a snippet from running the AWS integration tests with `--reporter json`:
```json
{"status":"failed","code_desc":"EC2 Security Group sg-de7859b5 should allow out only {:from_port=>6000, :to_port=>6007}
```

2.1.30

### 🤔 Replication Case

The AWS integration tests display this behavior:

```
$ bundle exec rake test:aws
# ... snip ...
×  EC2 Security Group sg-de7859b5 should allow in {}
     expected `EC2 Security Group sg-de7859b5.allow_in?({})` to return true, got false
     ✔  EC2 Security Group sg-de7859b5 should not allow in {}
     ×  EC2 Security Group sg-de7859b5 should allow in {}
# .... snip ...
```

Refs #2918, #2876

clintoncwolfe:
Fixes #2925 

2925 indicated that reporters would output an empty hash when a matcher that accepted hash arguments was used.  It turned out that works fine - it's just that two resources, `aws_iam_security_group` and `aws_iam_policy` were both using Hash#delete to whitelist criteria in matchers, without `dup`ing the Hash.  So, if all went well, the input hash would be emptied; and the reporter would faithfully include an empty hash in the test description.

This fix simply adds `dup` calls to the matcher hash arguments.  To test, run the AWS integration tests and observe test output for the two affected resources.

### DIFF
--- a/lib/resources/aws/aws_iam_policy.rb
+++ b/lib/resources/aws/aws_iam_policy.rb
@@ -93,8 +93,9 @@ class AwsIamPolicy < Inspec.resource(1)
     end
   end
 
-  def has_statement?(raw_criteria = {})
+  def has_statement?(provided_criteria = {})
     return nil unless exists?
+    raw_criteria = provided_criteria.dup # provided_criteria is used for output formatting - can't delete from it.
     criteria = has_statement__normalize_criteria(has_statement__validate_criteria(raw_criteria))
     @normalized_statements ||= has_statement__normalize_statements
     statements = has_statement__focus_on_sid(@normalized_statements, criteria)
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @clintoncwolfe

miah (approved):

PR #3045, opened 2018-05-13
Two fixes to FilterTable `where` criteria handling

### LINKED ISSUE #2929: FilterTable block filters run an extra iteration with Trace objects
### 🎛 Description

🐛 bug report

When using a block-based filter and using field methods, the block executes once for each row in the FilterTable (correctly), then once more with the field methods returning FilterTable::Trace objects.

I would expect the block to execute once for each row in the table, with the field methods returning their proper values.

2.1.33

### 🤔 Replication Case

I found it while working on #2926 .  Here's an example, using a made-up resource.

```ruby
describe primary_colors.where { puts color_name; color_name =~ /e/ } do
  it { should exist }
end
```

Output from puts:
```
red
green
blue
#<FilterTable::Trace:0x00007f80c2d36928 @chain=[]>
```

### 💁 Possible Solutions

Meditate on FilterTable.

### 💻 Stacktrace

Here's an example from a proposed unit test on `aws_vpcs`.
```
     hit = AwsVpcs.new.where do
      require 'byebug'; byebug
      IPAddr.new('10.0.0.0/8').include? cidr_block
    end
```

Here's a screencap of poking at the values.  There were two legit entries.

screen shot 2018-04-09 at 14 35 05

The last few lines are especially problematic.  Because FilterTable::Trace implements `method_missing` - and returns itself from it!  - it will match many things, thus creating false positives.

### LINKED ISSUE #2943: FilterTable does not validate keys when filtering in named parameter mode
### 🎛 Description

🐛 bug report

FilterTable, the guts behind all of the plural resources, allows InSpec resources to filter a tabular result set using criteria, either in block mode, or by passing a hash of criteria and values, in which case the operator is assumed to be 'equalish'.  

Each criterion is explicitly defined in the resource source code, using the `add` method of the FilterTable suite.

I've discovered that the `where` method doesn't care what keys you send:

```
(byebug) u = AwsIamUsers.new.where(garbage_i_made_up: true)
IAM Users with garbage_i_made_up == true
(byebug) u.entries
[]
(byebug) u.exists?
false
```

This is problematic, because we often recommend looking for insecure conditions using a plural resource and negated existence:

(dummy resource)
```ruby
describe things.where(is_safe => false) do
  it { should_not exist }
end
```

In this case, what if `is_safe` did not exist, and the correct criteria was `is_secure` ?  This would result in a false negative condition.

2.1.33

### 💁 Possible Solutions

FilterTable knows what you intend to use as criteria because of `add` or else the keys of the internal table.  It should raise a warning and fail the resource.

clintoncwolfe:
Partially Fixes #2929 , in which code in a block provided to `where` gets evaluated one more time than expected.  This was determined to be harmless (the extra evaluation occurs to stringify the where block for output, and filtration is not affected by the extra eval).  However, because the extra evaluation is in a different context than the filtration evaluations (a context in which all missing method return a Trace object), it is possible to have exceptions occur in the stringification eval, but not in the filtration evals.  This PR intercepts all such exceptions _and ignores them_.  Stringification still works but has less detail in such an event.

Fixes #2943 , in which Hash criteria passed to a method-call-style `where` call are not validated.  This could lead to user errors.  This PR changes behavior such that all criteria must be a key of one of the hash-rows of the raw data; if the raw data is empty (zero rows) all provided criteria are treated as valid.  That is dubious; feedback welcome.

A set of functional tests and a simple profile fixture are provided.

### DIFF
--- a/test/unit/mock/profiles/filter_table/controls/where_iteration.rb
+++ b/test/unit/mock/profiles/filter_table/controls/where_iteration.rb
@@ -1,16 +1,22 @@
 title 'Verify `where` filter blocks iterate properly - issue 2929'
 
-control 'When the data has no rows, there should be zero iterations' do
+control 'When the data has no rows, there should be zero iterations witth one where call' do
   $count = 0
-  describe iteration_counter([]).where { $count += 1; true } do
+  describe simple_plural([]).where { $count += 1; true } do
     subject { $count }
     it { should cmp 0 }
   end
 end
 
+control 'When the data has no rows, entries should report zero count' do
+  describe simple_plural([]).where { true } do
+    its('entries.count') { should be_zero }
+  end
+end
+
 control 'When the data has no rows, field accessors should not be called' do
   $field_accessor_values = []
-  describe iteration_counter([]).where { $field_accessor_values << id; true } do
+  describe simple_plural([]).where { $field_accessor_values << id; true } do
     subject { $field_accessor_values }
     it { should be_empty }
   end
@@ -18,7 +24,7 @@ end
 
 control 'When the data has one row, there should be one iteration' do
   $count = 0
-  describe iteration_counter([{id: 1 }]).where { $count += 1; true } do
+  describe simple_plural([{id: 1 }]).where { $count += 1; true } do
     subject { $count }
     it { should cmp 1 }
   end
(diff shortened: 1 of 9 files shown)


jquick (approved):
Nice addition @clintoncwolfe !

test/unit/mock/profiles/filter_table/controls/where_iteration.rb
```diff
@@ -0,0 +1,59 @@
+title 'Verify `where` filter blocks iterate properly - issue 2929'
+
+are_we_worried_about_2929_breaking_filtration = false
+
+only_if { are_we_worried_about_2929_breaking_filtration }
+
+control 'When the data has no rows, there should be zero iterations witth one where call' do
```
miah: 'witth'

miah (approved):
Thank you @clintoncwolfe

PR #3046, opened 2018-05-14
Add train GCP transport

jquick:

This change includes the Train pin for the new GCP features as well as a berkshelf update which is needed to support it.

This is manually testing of Train master for now. Will be updated once the Train gem is released.

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -2,6 +2,11 @@
 source 'https://supermarket.chef.io'
 
 cookbook 'apt'
+cookbook 'yum'
+cookbook 'postgresql'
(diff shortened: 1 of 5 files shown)


clintoncwolfe (commented):
Why the Berkshelf changes? (what's the connection  to the train pin?)

jquick:
GCP depends had some conflicts with Berkshelf. This was resolved by upgrading it on train and inspec.

clintoncwolfe (approved):

miah (approved):

Issue #3041, opened 2018-05-11
RFC: FilterTable / Plural Resource Standardization and Improvements


clintoncwolfe:
### 🎛 Description

This is an RFC around FilterTable, the support library which implements the "pluralization" features of many resources.

Suggestions:

1.  Provide `where` and `entries` facilities by default.
2. Deprecate the `add_accessor` configuration method; it only has meaning for `where` and `entries`
3.  Rename `params` to `raw_data`
4.  Provide `exist?` by default
5.  Provide `count` by default
6.  Ensure that any field that appears in a raw data hash-row may be used as a data accessor in block-form where
7.  Add aliases to RSpec: alias `they` to `it` and `their` to `its`.
8.  Provide simple array properties by default for each field seen in the raw data, but allow that to be overridden.
9.  Assign names to anonymous classes.
10. Ensure that block-valued properties (which do not sensibly have a per-row value) do not appear in the structs returned by `entries`.  One example is `exist?`.

Aside from some deprecations - which can be implemented as warnings now, or excluded from the RFC - none of these changes are necessarily a breaking change.

### 🌍 InSpec and Platform Version

2.1.72

### 🤔 Replication Case

For 1, 4, 5 and 8, we'd be standardizing what we typically do anyway.  When we omit them, we end up with issues opened ("I wish this resource implemented `where`!).  When we add them on an individual basis, it involves a lot of tedious copying, and the errors that entails.

3 - `params` - has a big presence in the wild.  I'd like to see it called 'raw_data', and then we update our examples and training materials accordingly.  `params` is a word that reflects input, not a data accessor.

6 is, I think, a bug or misfeature.  It doesn't make sense that you can filter on any field in method-form `where`, but only selected fields in block-form where.

7 is just for readability.

```
  users.where(last_name: /Naughty/) do
    they { should_not exist }
  end
```

9 if for debugging, and is not a big deal.

clintoncwolfe:
Refs #2779 #3032 #2955 #2420

clintoncwolfe:
I've determined that 6 and 8 are likely impossible - we don't know the fields in the raw data until we get the raw data, which is after the table methods are generated.  We could postpone table method generation until after the raw data fetch, but that would mean attaching new methods to an instance, which would be a whole different approach.

clintoncwolfe:
All that is doable was done :-)

PR #3047, opened 2018-05-14
Refactor: Perform internal rename and add comments to FilterTable

clintoncwolfe:
This PR aims to clarify the functioning of FilterTable by performing an internal renaming of local variables, instance variables, one module name, and a handful of methods (all of which also receive backwards-compatible aliases or explicit method definitions).  Additionally, explanatory comments were added in some of the key locations.

In a few places, the code itself is re-worked for clarity.

This effort focused on the classes FilterTable::Factory and FilterTable::Table.  I didn't look at FilterTable::Trace very much because it only has one point of contact with the rest of the code, and we rarely have to dive into it - it is a presentation-layer class.

I've run the unit tests and functional tests, both pass.

Refs #3041


jquick (approved):
Very nice cleanup! Thanks for this @clintoncwolfe.

miah (approved):
Thank you for the attention to all these details in cleaning up FilterTable. This is much more readable!

PR #3056, opened 2018-05-16
Rename Plugin::name to Plugin::register

### LINKED ISSUE #3055: Rename `name` method in lib/utils/plugin_registry
### 🎛 Description

See https://github.com/pry/pry/issues/1535

`binding.pry` fails if `lib/utils/plugin_registry` is loaded, because it defines a `self.name` method

2.1.72 from gem

### 💁 Possible Solutions

 * Because the method's purpose is "Register a new plugin by name" it might be better to rename it to `self.register` OR
 * `byebug` seems to work where pry doesn't here. It could be kept as a known issue that a contributor needs to use byebug in some cases. 

### 💻 Stacktrace

```
ArgumentError: wrong number of arguments (given 0, expected 1) 
from /home/user/.rvm/gems/ruby-2.4.3/gems/inspec-2.1.72/lib/utils/plugin_registry.rb:49:in 'name' 
[9] pry(unknown)> wtf 
Exception: ArgumentError: wrong number of arguments (given 0, expected 1) 
0: /home/user/.rvm/gems/ruby-2.4.3/gems/inspec-2.1.72/lib/utils/plugin_registry.rb:49:in 'name'
1: /home/user/.rvm/gems/ruby-2.4.3/gems/pry-0.11.3/lib/pry/wrapped_module.rb:150:in 'method_missing'
2: /home/user/.rvm/gems/ruby-2.4.3/gems/pry-0.11.3/lib/pry/wrapped_module.rb:100:in 'nonblank_name' 
3: /home/user/.rvm/gems/ruby-2.4.3/gems/pry-0.11.3/lib/pry/wrapped_module.rb:87:in 'method_prefix' 
4: /home/user/.rvm/gems/ruby-2.4.3/gems/pry-0.11.3/lib/pry/method.rb:271:in 'name_with_owner' 
```

james-stocks:
fixes #3055

The current method name `name` crashes pry (https://github.com/pry/pry/issues/1535)
Although that should be addressed in pry; `register` is a better self-describing name for the method anyway.

### DIFF
--- a/lib/bundles/inspec-compliance/target.rb
+++ b/lib/bundles/inspec-compliance/target.rb
@@ -11,7 +11,7 @@ require 'inspec/errors'
 # similar to `inspec exec http://localhost:2134/owners/%base%/compliance/%ssh%/tar --user %token%`
 module Compliance
   class Fetcher < Fetchers::Url
-    name 'compliance'
+    register 'compliance'
     priority 500
     def self.resolve(target) # rubocop:disable PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/AbcSize
       uri = if target.is_a?(String) && URI(target).scheme == 'compliance'
--- a/lib/bundles/inspec-supermarket/target.rb
+++ b/lib/bundles/inspec-supermarket/target.rb
@@ -9,7 +9,7 @@ require 'fetchers/url'
 # InSpec Target Helper for Supermarket
 module Supermarket
   class Fetcher < Inspec.fetcher(1)
-    name 'supermarket'
+    register 'supermarket'
     priority 500
 
     def self.resolve(target, opts = {})
--- a/lib/fetchers/git.rb
+++ b/lib/fetchers/git.rb
@@ -25,7 +25,7 @@ module Fetchers
   # omnibus source for hints.
   #
   class Git < Inspec.fetcher(1)
-    name 'git'
+    register 'git'
     priority 200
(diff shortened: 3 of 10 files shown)


james-stocks:
I assume this is an internal method and no work needs to be done to manage deprecating the old method name?

jquick:
Hey @james-stocks. This is a great idea but I am a bit torn about it. We use `name` for almost all our other dsl's like resource/transports so it would still flow better here imo. Also if anyone is running custom fetchers we will bust them. Not sure if this is happening but something to think about. I'll bring this up to the team and see what we want to do.

james-stocks:
@jquick thanks. If this change is worrisome then it can perhaps be left as a known issue that pry cannot be used with PluginRegistry

At a quick test, it appears that `byebug` is usable here - I can break into byebug at the point where pry would throw the error

jquick:
We talked this over as a team today and have a few blockers on it:
1) We do conform to 'name' as our internal DSL for transports/resources already so we would like to keep it uniform unless we change them all.
2) We don't want to change our code to supplement Pry or another external tool as that should not be the deciding factor.

For now we have updated inspec to use byebug so that should be a work around. If we do decide to change all the DLS's in the future we will reference this.

Thanks for bringing this up!

james-stocks:
Thanks for giving this consideration @jquick - that decision sounds wise to me!

PR #3061, opened 2018-05-16
Adds middleware dependecy for Azure

dmccown:
We are adding this depdency until we can integrate our connection into Train.

Signed-off-by: David McCown <dmccown@chef.io>


jquick (approved):
Thanks @dmccown !

miah (approved):

PR #3063, opened 2018-05-17
Create separate hab package for use under Automate

clintoncwolfe:
This PR make a separate Habitat package, for use when `inspec` is being installed on an Automate server.  The primary changes are that it changes the licensing, and also installs `inspec-scap`, a plugin for profile interpretation.

The Expeditor configuration is likely borken at this point, as we need to handle the git clone of the inspec-scap repo safely (we currently rely on an env var, which we're not setting).


clintoncwolfe:
I'm closing this for now - not clear if this is teh approach we're taking, and we need to clear up some access issues.

PR #3074, opened 2018-05-23
Adding docker plugin support

frezbo:


frezbo:
@jquick Any plans to get this reviewed, I completely forgot about this.

jquick (commented):
This is really nice @frezbo, great improvement to be able to check plugins on docker. You think you could whip up a unit test like:
https://github.com/inspec/inspec/blob/master/test/unit/resources/docker_container_test.rb https://github.com/inspec/inspec/blob/master/test/unit/resources/docker_service_test.rb 

for this new case?

frezbo:
I will add the tests and test further, I might be able to finish off next week.

frezbo:
@jquick  I have two unrelated changes because of my squash. I hope that's ok

jquick:
Hey @frezbo I think you need to rebase your branch, pulling in some changelog stuff.

frezbo:
@jquick fixed. It was due to my master being incorrectly rebased.

jquick (approved):
Thanks @frezbo !!  This is a very nice improvement.

docs/resources/docker.md.erb
```diff
@@ -148,6 +148,17 @@ Or execute the profile directly via URL:
       its('sizes') { should_not include "1.41 GB" }
     end
 
+### plugins
+
+`plugins` returns information about Docker plugins as returned by [docker plugin ls](https://docs.docker.com/engine/reference/commandline/plugin/).
+
+    describe docker_plugin('rexray/ebs') do
```
clintoncwolfe: The example is of using the `docker_plugin` resource, but here you're documenting using the `plugins` property of the `docker` resource.

docs/resources/docker_plugin.md.erb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+<br>
+
+## Property Examples
+
+### id
+
+The `id` property returns the full plugin id:
+
```
clintoncwolfe, line 40: Please give full examples, including the `describe` block.

docs/resources/docker_plugin.md.erb
```diff
@@ -0,0 +1,72 @@
...
+The resource allows you to pass in an plugin id:
+
+    describe docker_plugin(id: plugin_id) do
+      ...
+    end
+
+<br>
+
+## Property Examples
```
clintoncwolfe: This section should be titled "Properties"

docs/resources/docker_plugin.md.erb
```diff
@@ -0,0 +1,72 @@
...
+
+    its('id') { should eq '0ac30b93ad40' }
+
+### version
+
+The `version` property tests the value of plugin version:
+
+    its('version') { should eq '0.11.0' }
+
+### Test a Docker plugin
```
clintoncwolfe: This should be in a main-level section named "Examples" - this isn't a property.

docs/resources/docker_plugin.md.erb
```diff
@@ -0,0 +1,72 @@
...
+      its('id') { should_not eq '0ac30b93ad40' }
+      its('version') { should eq '0.11.1' }
+      it { should be_enabled }
+    end
+
+<br>
+
+## Matchers
+
+For a full list of available matchers, please visit our [matchers page](https://www.inspec.io/docs/reference/matchers/).
```
clintoncwolfe: Please refer to this as the 'Universal Matchers' page.

docs/resources/docker_plugin.md.erb
```diff
@@ -0,0 +1,72 @@
...
+
+### exist
+
+The `exist` matcher tests if the plugin is available on the node:
+
+    it { should exist }
+
+### enabled
+
+The `be_enabled` matches tests if the plugin is enabled
```
clintoncwolfe: Please provide an example.

lib/resources/docker.rb
```diff
@@ -232,5 +257,17 @@ def parse_images
...
+
+    def parse_plugins
+      plugins = inspec.command('docker plugin ls --format \'{"id": {{json .ID}}, "name": "{{ with split .Name ":"}}{{index . 0}}{{end}}", "version": "{{ with split .Name ":"}}{{index . 1}}{{end}}", "enabled": {{json .Enabled}} }\'').stdout
+      c_plugins = []
+      plugins.each_line { |entry|
+        c_plugins.push(JSON.parse(entry))
+      }
+      c_plugins
+    rescue JSON::ParserError => _e
+      warn 'Could not parse `docker plugin ls` output'
```
clintoncwolfe, line 260: This should throw a ResourceFailed exception - this will otherwise risk giving bad results.

lib/resources/docker.rb
```diff
@@ -56,6 +56,23 @@ def initialize(images)
     end
   end
 
+  class DockerPluginFilter
+    filter = FilterTable.create
+    filter.add_accessor(:where)
```
clintoncwolfe: Please remove the 'where', 'entries', and 'exists?' lines; those are now provided automatically.

lib/resources/docker_plugin.rb
```diff
@@ -0,0 +1,65 @@
+# encoding: utf-8
+#
+# Copyright 2017, Christoph Hartmann
```
clintoncwolfe: Please remove incorrect copyright line.

test/unit/resources/docker_plugin_test.rb
```diff
@@ -0,0 +1,28 @@
+# encoding: utf-8
+# author: Noel Georgi
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::DockerContainer' do
```
clintoncwolfe, line 7: We need two more test cases:
* A case in which docker is installed but there are zero plugins
* A case in which docker is not installed

test/unit/resources/docker_test.rb
```diff
@@ -24,6 +24,13 @@
       _(resource.services.images).must_equal ["foo/image:1.0", "foo/image:1.1", "bar:latest", "bar:latest"]
     end
 
+    it 'check docker plugins parsing' do
```
clintoncwolfe, line 27: A case with zero plugins is needed here.

clintoncwolfe (dismissed):
Great work @frezbo !  I have some documentation feedback and a couple of testing concerns, but the implementation is sound.  Thanks!

lib/resources/docker.rb
frezbo: This cannot be handled as we are only getting stdout (https://github.com/inspec/inspec/pull/3074/files#diff-a405d35eca4d46fa85c19029d6f3b79bR259) and any errors always returns a valid JSON. This would need a lot of code change in the docker resource to handle properly.

test/unit/resources/docker_plugin_test.rb
frezbo: I have tried to add a test where no plugins exist, not sure if it's right. Can you help me here?

test/unit/resources/docker_test.rb
frezbo: I do not see any existing examples, and adding this code would be out of scope of this ticket.

frezbo:
@clintoncwolfe I have updated the README and some part of code. Please do see the inline comments and let me know your thoughts

test/unit/resources/docker_test.rb
clintoncwolfe: Existing examples or not, we need to ensure that InSpec is stable.  It is possible to have zero plugins installed.  Prove that this feature works properly under that condition (I believe it does, but good software isn't built on beliefs :-) )  

Testing your a feature on reasonable, foreseeable edge cases is indeed in scope.

frezbo:
@clintoncwolfe I could not find your last comment. Makes sense. Will add a test while I'm free. Kudos on keeping up the inspec code quality.

clintoncwolfe:
Ahh, I deleted my comment about testing edge cases, because I realized that the test harness makes that pretty hard to do, and I agreed that making the test harness easier is out of scope :-)

frezbo:
:+1: , anything else from my side needed to get this merged.? @clintoncwolfe

docs/resources/docker.md.erb
```diff
@@ -152,11 +152,11 @@ Or execute the profile directly via URL:
 
 `plugins` returns information about Docker plugins as returned by [docker plugin ls](https://docs.docker.com/engine/reference/commandline/plugin/).
 
-    describe docker_plugin('rexray/ebs') do
-      it { should exist }
-      its('id') { should_not eq '0ac30b93ad40' }
-      its('version') { should eq '0.11.1' }
-      it { should be_enabled }
+    describe docker.plugins do
+      its('names') { should include ["store/weaveworks/net-plugin", "docker4x/cloudstor"] }
```
clintoncwolfe: I'm curious, did you see 'compare using include and an array somewhere'?  Where?

test/unit/resources/docker_plugin_test.rb
clintoncwolfe: @jerryaldrichiii , on #3151 , you just added unit test code to handle a mock command when it does not exist.  I have no experience with that.  Here, we'd like to add two more cases - one where `docker plugin ls` generates no output, and one where the binary is not installed.  Could you suggest anything?

docs/resources/docker.md.erb
frezbo: @clintoncwolfe one example here: https://www.inspec.io/docs/reference/resources/limits_conf/ 
There are multiple examples in `limits_conf`, `auditd`, `etc_hosts_allow` and `etc_hosts_deny` (did a grep on the codebase). I'm not sure if I have used such a matcher in my tests. Yes good catch, the example here is wrong. I will update.
```
inspec> describe docker.images do
inspec>   its('ids') { should include ['sha256:6c64e61818d4d683aa975e4754d6fd5c4c252b2b1ac0a5d7734017e8f6acc3be', 'sha256:86d8fd396ed6f344bfbd80cd32c884981365c77eae64d3251c97c44cd9b4a8bf'] }  
inspec> end  

Profile: inspec-shell
Version: (not specified)

  #<Inspec::Resources::DockerImageFilter:0x0000557d5548c128>
     ×  ids should include ["sha256:6c64e61818d4d683aa975e4754d6fd5c4c252b2b1ac0a5d7734017e8f6acc3be", "sha256:86d8fd396ed6f344bfbd80cd32c884981365c77eae64d3251c97c44cd9b4a8bf"]
     expected ["sha256:6c64e61818d4d683aa975e4754d6fd5c4c252b2b1ac0a5d7734017e8f6acc3be", "sha256:86d8fd396ed6f344b...4d8b3b2df83d2695d477636", "sha256:405c277e14b2422035d4d783a48c503c914dc62ba56b160e9ca862239eb46380"] to include ["sha256:6c64e61818d4d683aa975e4754d6fd5c4c252b2b1ac0a5d7734017e8f6acc3be", "sha256:86d8fd396ed6f344bfbd80cd32c884981365c77eae64d3251c97c44cd9b4a8bf"]

inspec> 
```
I'm not sure about the other resources

jerryaldrichiii:
@clintoncwolfe  You could mock empty output of `docker plugin ls` using `empty.call` see:

https://github.com/inspec/inspec/blob/master/test/helper.rb#L204
https://github.com/inspec/inspec/blob/master/test/helper.rb#L496

As far as mocking the command not existing you could do the same, but call `cmd_exit_1.call` see:

https://github.com/inspec/inspec/blob/master/test/helper.rb#L208

Something like: `'bash -c \'type "docker"\'' => cmd_exit_1.call`,

Does that help?

docs/resources/docker.md.erb
clintoncwolfe: Thanks!  refs #3162

frezbo:
@clintoncwolfe I will need extra help to complete this, the testing framework is a little bit complex for me to apprehend. I can't make sense out of it.

jquick:
This has been refined quite a bit. I am going to have another round of quick reviews on this.

lib/resources/docker.rb
jquick: I think we can just do `fail_resource('Could not parse `docker plugin ls` output')` instead of your warn.

test/unit/resources/docker_test.rb
jerryaldrichiii: I think this is in the scope of this ticket, but given it's age I don't want to hold up the PR over it. We can add it later.

Thank you so much for your patience and work @frezbo!

lib/resources/docker_plugin.rb
```diff
@@ -0,0 +1,63 @@
+# encoding: utf-8
+
+module Inspec::Resources
+  class DockerPlugin < Inspec.resource(1)
+    name 'docker_plugin'
+    supports platform: 'unix'
+    desc ''
```
jerryaldrichiii: Could we add a short description here?

lib/resources/docker.rb
jerryaldrichiii: I think we should do ^ also.

If you can't parse the JSON then we should fail the resource. Granted, I see your point if we get a legitimate error in valid JSON.

jerryaldrichiii (dismissed):
So close! Only a few minor changes.

miah (approved):
+1 after @jerryaldrichiii comments are resolved.

frezbo:
let this go in the next release, I could fix the small issues by then @jquick @jerryaldrichiii

jquick:
Thanks for your work on this @frezbo! Sorry for the long wait on it.

PR #3080, opened 2018-05-26
Fixing wrong attribute reference for rspec output

frezbo:


jquick (approved):
Thanks!

miah (approved):
Thanks @frezbo

PR #3081, opened 2018-05-26
Adding YAML reporter

frezbo:


frezbo:
I can close this in favour of #3085 , as then the json or `runner.report` can be parsed with YAML. Let me know your comments

jquick (changes requested):
I love the addition of YAML but we seem to be copying a lot of the building blocks from the json report. Can we just subclass/create this(json) class and then call to_yaml on the object?

frezbo:
Yes I agree with you @jquick . I was holding off until #3085 is addressed, what do you think.

jquick:
Yeah #3085 looks fine to me!

frezbo:
@jquick  I have updated the PR, let me know your comments, I will update the docs then

jquick:
Looking really nice! You mind putting a basic render unit test in? Just something like: https://github.com/inspec/inspec/blob/master/test/unit/reporters/junit_test.rb
Should be fine for now.

frezbo:
@jquick I think this should be good. The `lint ` task is failing though. I can't seem to think of a solution. PTAL.

jquick:
You can just disable the cop for this one if you want. We are passing the complexity to the JSON reporter. Else you could move some to a init block if you wanted.

frezbo:
@jquick  The PR should be good. I have done these changes in the last commit:
 - Fixes the YAML reporter UT
 - Adds the report method to YAML reporter to support code example as in #3085
 - Disables the cyclomatic complexity Metric for reporter

I was wondering if we should have a namespace under `test` for the `reporters`/`mock` tests, I completely missed the failing UT for the YAML and had to add that in Rake to test it locally. What do you think.

jquick (approved):
Thanks @frezbo, this looks good to me.

Yeah the mock folder is just a storage place but a namespace would be a good idea if applicable. Though I am not sure if we want actual code there or just mock src files.

jquick:
AppVeyor check is good https://ci.appveyor.com/project/chef/inspec/build/1.0.6379. This PR passes all checks.

miah (approved):
Awesome! Thanks @frezbo

Issue #2370, opened 2017-12-05
RFC: Lazy column fetching for FilterTable-based resources


clintoncwolfe:
Over at [inspec-aws issue #100](https://github.com/chef/inspec-aws/issues/100), we've been discussing the need for and advantages of being able to "lazy load" columns in the underlying dataset in a FIlterTable-based resource.

As a non-AWS example, the pattern we keep seeing is one in which we have a generic data fetching operation (say, listing running processes) with a property which must be populated by a potentially expensive secondary call (say, listing open files for a process).  In some cases we might be able to get the column value for all rows; in other cases, we are obliged to make one call per-row.

In inspec-aws land (or any other resource pack that involves remote API calls), being obliged to populate all possible cells in the table can be a large performance hit (see aws_iam_access_keys for an example).  Ideally, we'd prefer to only populate a column if it is being requested.

Several approaches involve adding a Proc (a ruby object that represents code to be executed in the future) to the FilterTable.  Where to add it, and what args the Proc should expect, are matters for discussion.  

Any solution would have to be optional: all existing FilterTable-based resources should not have to be modified.

I'd like to thank @aduric and @chris-redekop for bringing up this issue, all their careful thought into the matter, and their experimental implementations.

aduric:
PR can be found here: https://github.com/chef/inspec/pull/2371

miah:
__What’s happening? Why was this issue closed?__
This issue was closed due to some much needed review of legacy issues or issues that were spawned in older versions of InSpec, i.e. < v3.

__Why do I care?__
You would care about this if this was an issue that you are still seeing and/or feel needs to be addressed in the current version of InSpec.

__What do I need to do?__
If this issue is no longer important, no further action is necessary. However, if you think this is something that should be addressed, please open a new issue and refer to the original issue in the description.

PR #3093, opened 2018-05-30
Add lazy-loading to FilterTable

clintoncwolfe:
This PR implements column-based "lazy-loading" for FilterTable.  For plural resources, this allows them to defer performing time-expensive fetches for columns (fields) that may never be used.  

The facility works for both where-method (`things.where(my_field: my_value)`) and where-block (`things.where { my_arbitrary_code && my_field }`).

A single lazy fetcher may populate multiple columns.  Lazy and traditional population may also be mixed; the populator will not clobber a value already present in the raw data if rows are irregular.

Fetchers have access to the row they are working on, as well as any provided filter value, and a reference to the entire FilterTable object, which can provide deep context.

If an exception is thrown in a lazy fetcher, it is captured and the control is failed as a control source code failure.

An example implementation is provided on `aws_iam_users`, whose performance is greatly improved.

Refs #2371 #2370 

Many, many thanks to @aduric and @chris-redekop for their initial proposal on this, and their prototype implementation.  

Attn: @tmonk42


jquick (approved):
Nice enhancement! Thanks @clintoncwolfe.

jquick:
Closing/reopening to refresh tests

miah (approved):
This will be extremely useful. Thank you =)

Issue #3067, opened 2018-05-17
inspec vendor not working for git dependencies


chris-rock:
### 🎛 Description

Create a new `inspec.yml` that looks like:
```
name: gcp-compliance-example
title: InSpec Profile
maintainer: The Authors
copyright: The Authors
copyright_email: you@example.com
license: Apache-2.0
summary: An InSpec Compliance Profile
version: 0.1.0

depends:
  - name: gcp
    git: git@github.com:inspec/inspec-gcp.git 
```

Then run `inspec vendor .`:
```
inspec vendor . 
Dependencies for profile . successfully vendored to vendor
```

Shows that the directory structure is not correct:

```
tree vendor -L 2
vendor
└── 836da0196e894c980153f729d7a1f476ed7e43b8
    └── d20180517-36416-1avoz2b

2 directories, 0 files
```

The subdirectory `d20180517-36416-1avoz2b` should not exist and everything should be in `836da0196e894c980153f729d7a1f476ed7e43b8` 

### 🌍 InSpec and Platform Version

2.1.80

### 🤔 Replication Case

See above

### 💁 Possible Solutions

This is probably an issue in our git fetcher

### 💻 Stacktrace

n/a

PR #3097, opened 2018-05-31
Fix the git fetcher to vendor correctly

jquick:
The current git fetcher copies the folder that was checked out instead of the checkout internals. This breaks the dependency reader as it does not expect a imbedded folder. 

Fixes https://github.com/inspec/inspec/issues/3067

### DIFF
--- a/test/functional/inspec_vendor_test.rb
+++ b/test/functional/inspec_vendor_test.rb
@@ -34,6 +34,25 @@ describe 'example inheritance profile' do
     end
   end
 
+  it 'can vendor profile dependencies from git' do
+    git_depends_path = File.join(profile_path, 'git-depends')
+
+    Dir.mktmpdir do |tmpdir|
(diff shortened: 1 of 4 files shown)


test/functional/inspec_vendor_test.rb
```diff
@@ -34,6 +34,25 @@
     end
   end
 
+  it 'can vendor profile dependencies from git' do
+    git_depends_path = File.join(profile_path, 'git-depends')
```
clintoncwolfe, line 38: *ahem* I think you missed a commit - I don't see a test fixture profile for this, and the functional tests fail.

clintoncwolfe (changes requested):
Missing a test profile mock.

test/functional/inspec_vendor_test.rb
jquick: :D thanks forgot to add it in

clintoncwolfe (approved):
That got it.  Thanks @jquick

PR #3104, opened 2018-06-01
Add common methods to FilterTable

### LINKED ISSUE #3103: Install common FilterTable methods by default
### 🎛 Description

🙋 feature request

Nearly every resource that uses FilterTable does this series of calls:
```
filter = FilterTable.create
         .add_accessor(:where)
         .add_accessor(:entries)
         .add(:exist?) { |t| !t.entries.empty? }
         # Remaining .add calls are specific to the resource - not boilerplate
```

When we don't install one of those methods it always comes back to haunt us, usually as a missing feature issue.

Let's install them by default, without requiring they be specified in the resource code.  I also suggest this as a standard method:

```
         .add(:count) { |t| t.params.count }
```

Note that we can replace the call to `entries` with `params` in the implementation of `exist?` - they both examine the raw data, but `entries` does extra unneeded work to inflate the rows into a special row struct, while we only really care about the count of rows.

Some grepping around in the 2.1.83 source found only one case of a custom `where` - on `etc_groups`, which does not use FilterTable anyway (see #2983).  So, I don't believe we need to watch out for clobbering an existing method.

Refs #3041 

Refs #2779

2.1.83

### 💁 Possible Solutions

It's pretty straightforward to just add the desired methods to the FilterTable::Factor constructor.

### LINKED ISSUE #3110: Ensure that FilterTable entries structs do not expose internals
### 🎛 Description

🐛 bug report

Ensure that block-valued properties (which do not sensibly have a per-row value) do not appear in the structs returned by entries. One example is exist?.

For example, if you have a resource using FilterTable like this:
```ruby
filter_table_config = FilterTable.create
                        .add_accessor(:entries)
                        .add(:exist?) { |table| ! table.entries.empty? }
                        .add(:colors, field: :color)
```

If you examine `entries`, you'll see something odd:
```ruby
my_resource.entries => [
  Struct({ exist?: nil, color: :red }),
  Struct({ exist?: nil, color: :blue }),
]
```

The key `exist?` is spurious - its value will always be nil, and its presence adds no meaning.  All properties added using `add` will display this behavior.

Expected behavior is for the block-valued keys to be absent.

Refs #3041 

2.1.83

### 🤔 Replication Case

See above

### 💁 Possible Solutions

When creating the Struct type for new entries, skip any 'connector' fields that are block-valued.

### 💻 Stacktrace

None.

clintoncwolfe:
Fixes #3103 
Fixes #3110 

As described on #3103, this PR modifies the behavior of FilterTable to automatically provide `where`, `entries`, `raw_data`, `exists?`, and `count`.  You no longer have to call `add_accessor` or `add` to set them up.

In some places, we had tests that were examining the contents of the `entries` table, which would now have at least `:count` and `:exist?` in it, due to an existing bug, #3110.  As those were spurious, this PR removes those entries, fixes one test on `shadow`, and adds at least one field to `aws_iam_groups`.

I think further modifications may be needed to azure_generic_resource, as it does some unusual things, adding additional `add_accessor` calls.


jquick (approved):
Thanks @clintoncwolfe !

jquick:
Looks like tests are failing due to the new fields `exist?, etc`

miah (approved):
Solid improvements. Thank you @clintoncwolfe =)

clintoncwolfe:
Yes, I pushed this late last night without running the unit tests.  It definitely has a regression around `:exists?`

 Diagnosed as #3110 - I'll fix on this PR.

Issue #3072, opened 2018-05-18
Package resource inexplicably fails to find certain installed packages on Windows Server 2016


jquick:
### 🎛 Description

🐛 bug report 

Package resource inexplicably fails to find certain installed packages on Windows Server 2016

**Expected Results:** 
e.g. Microsoft SQL Server 2012 Command Line Utilities is found

**Actual Results:** 
e.g. Microsoft SQL Server 2012 Command Line Utilities is not found

**Repro Steps (if possible):**
 Spin up a Windows Server 2016 instance, install Microsoft SQL Server 2012 Command Line Utilities, develop simple Inspec test to check if it is installed

**Request for Developers:**
Product Fix

**Example test:**
```
control 'Microsoft SQL Server 2012 Command Line Utilities' do
  describe package('Microsoft SQL Server 2012 Command Line Utilities') do
    it {should be_installed}
  end
end

Further proofing:

PS C:\Users\Administrator> Get-ItemProperty (@("HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall*", "HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall*", "HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall*", "HKCU:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall*") | Where-Object { Test-Path $_ }) |Select-Object DisplayName

DisplayName

Windows Driver Package - Amazon Inc. (AWSNVMe) SCSIAdapter  (11/16/2017 1.0.0.105)
Windows Driver Package - Amazon Inc. (AWSNVMe) SCSIAdapter  (01/11/2018 1.0.0.146)
aws-cfn-bootstrap
AWS PV Drivers
InSpec v2.1.59
aws-cfn-bootstrap
Microsoft SQL Server 2012 Native Client
aws-cfn-bootstrap
aws-cfn-bootstrap
aws-cfn-bootstrap
aws-cfn-bootstrap
Microsoft SQL Server 2012 Command Line Utilities
aws-cfn-bootstrap
Amazon SSM Agent
aws-cfn-bootstrap
Mozilla Firefox 59.0.3 (x86 en-US)
Mozilla Maintenance Service
Amazon SSM Agent
AWS Tools for Windows

[...]

PR #3106, opened 2018-06-01
Detect windows packages with trailing/leading spaces

jquick:
The current windows package matcher does a hard string comparison. This did not match when packages had a trailing space in the registry. This change updates the matcher to be a regex that allows leading/trailing spaces. I have confirmed that this `-match` flag is allowed in all powershell versions.

Fixes https://github.com/inspec/inspec/issues/3072


clintoncwolfe (approved):
Thanks @jquick

Issue #3107, opened 2018-06-01
Allow custom resources to access each other


jquick:
### 🎛 Description

Currently when loading up a custom resource it will be bound with the `inspec.backend` generated on startup which only includes core resources. We need a way to either create a on demand for all resources called from `backend` or rebuild when loading custom libraries.

PR #3108, opened 2018-06-01
Allow custom resources to access all other resources

jquick:
Currently custom resources can only access core resources through the use of `inspec.*`. This change loads in any other custom resources so you can access them the same way.

Fixes https://github.com/inspec/inspec/issues/3107


examples/custom-resource/controls/example.rb
```diff
@@ -0,0 +1,6 @@
+# encoding: utf-8
+
+describe gordon do
+  its('example') { should_not eq 'foobar' }
```
clintoncwolfe: My main concern is to use meaningful names.  Gordon here is meant to be Commissioner Gordon, from Batman.  What properties might he be dealing with?

its('crime_rate') { should be < 2 }
it { should have_a_fabulous_mustache }

I don't know what the 'example' property is, or why its value should or should not be  'foobar'.

examples/custom-resource/libraries/foobar.rb
```diff
@@ -0,0 +1,14 @@
+class Foobar < Inspec.resource(1)
```
clintoncwolfe: Perhaps Batsignal?

examples/custom-resource/libraries/foobar.rb
```diff
@@ -0,0 +1,14 @@
...
+  name 'foobar'
+
+  example "
+      describe foobar do
+        its('info') { should eq 'foobar' }
+      end
+  "
+
+  def info
```
clintoncwolfe: Here, I'd suggest something like `whoami_output` or something.  Make it clearer that we're doing a neat trick - calling a core resource from a custom one.

examples/custom-resource/libraries/gordon.rb
```diff
@@ -0,0 +1,15 @@
+class Gordon < Inspec.resource(1)
+  name 'gordon'
+
+  example "
+    describe gordon do
+      its('example') { should eq 'foobar' }
+    end
+  "
+
+  def example
```
clintoncwolfe: Likewise, maybe give this a name that makes it clear we're calling a custom resource from another custom resource.

Ooh, you could have a Batman and Bruce Wayne resource, and bruce implements `full_name` (which calls whoami) and Batman implements `secret_identity` (which calls full name on Batman)

clintoncwolfe: Also: defining a method named `example` is a bit high risk - `example` is part of the resource DSL!  No collision in this case, but the unfamiliar user (for whom this code is intended) would be easily confused.

clintoncwolfe (dismissed):
The implementation code is great, no complaints.

Just some naming issues in the example code.

jquick:
Thanks @clintoncwolfe - Updated the example to be a bit more understandable.

miah (approved):
APPROVE

clintoncwolfe (approved):
Looks good, thanks @jquick

PR #3115, opened 2018-06-06
[ON HOLD] Provide they / their aliases for it / its

### LINKED ISSUE #3109: Provide 'they' and 'their' aliases for plural resources
### 🎛 Description

🙋 feature request

When working with plural resources (those that operate on groups of platform objects, typically filtering them and auditing their properties in bulk), we currently use this sort of expression:

```ruby
describe cars.where(color: :blue) do
  it { should exist }
  its('licence_plates') { should include 'KOOL KAT' }
end
```

It's perfectly sensible for singular resources (imagine describing a single car: its license plate, it exists, etc).  But it reads awkwardly for plurals, and reflects an implementation detail, that `it`/`its` refer to the test subject instance of an RSpec Example Group.

It would be great if we could simply alias those, `it` => `they` and `its` => `their`:
```ruby
describe cars.where(color: :blue) do
  they { should exist }
  their('licence_plates') { should include 'KOOL KAT' }
end
```

Refs #3041 

2.1.83

### 💁 Possible Solutions

Poke at `rspec-its`, likely re-open some classes to inject an alias.

clintoncwolfe:
Fixes #3109 

Simply aliases `they` to `it` and `their` to `its`, allowing us to say:

```ruby
describe things do
  they { should exist }
  their('stuff') { should include 'bananas' }
end
```

This isn't directly impacting FilterTable, but it is useful for all plural resources.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -44,56 +44,58 @@ end
   * `should_not`  is the _negative condition_
   * `be_ridiculous`  is the _matcher_
 
-## Advanced Syntax
+## Plural Syntax
 
 ```
-describe foos('/path/to/foo.txt', ssl_verify: true).where { names == 'blah' } do
+describe foos('/path/to/foo.txt', ssl_verify: true).where { name == 'blah' } do
     its('jared') { should cmp >= 123 }
     its('jared.sort.first.monkey') { should be `loud` }
     its(['jared', 'monkey.with.dots']) { should be `loud` }
 end
 ```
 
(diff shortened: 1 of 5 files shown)


chris-rock:
I think this requires some additional discussion. I love the idea to make the single/multiple resource approach more clear. At this point I not sure how this specific PR is helping users though. Please let'ss have a chat early next week to discuss this through.

clintoncwolfe:
Withdrawn per internal discussion.

PR #3117, opened 2018-06-06
Update core resources with filtertable API changes

clintoncwolfe:
This updates all core resources to:
 * rename `add` to one of the three aliases, as appropriate (`register_column`, `register_custom_matcher`, `register_custom_property`)
 * remove all but one call to `add_accessor` (these are now provided automatically)
 * rename `connect` to `install_filter_methods_on_resource`

There are a lot of lines changed, but they are all simple.  Main motivation for this is so that future code browsers / copiers will get on the right foot.

### DIFF
--- a/lib/resources/aide_conf.rb
+++ b/lib/resources/aide_conf.rb
@@ -44,12 +44,10 @@ module Inspec::Resources
     end
 
     filter = FilterTable.create
-    filter.add_accessor(:where)
-          .add_accessor(:entries)
-          .add(:selection_lines, field: 'selection_line')
-          .add(:rules,           field: 'rules')
+    filter.register_column(:selection_lines, field: 'selection_line')
+          .register_column(:rules,           field: 'rules')
 
-    filter.connect(self, :params)
+    filter.install_filter_methods_on_resource(self, :params)
 
     private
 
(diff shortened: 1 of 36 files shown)


clintoncwolfe:
Temporarily withdrawing this - I'm going to separate out the deletion of the `exists?` matcher onto a separate PR.

clintoncwolfe:
Reverted removal of `exists?`, re-opening

jerryaldrichiii (changes requested):
Looks great, only feedback is style related.

Not sure what the proper term is, but on lining up the `field`s, should it be two spaces after the longest item? Seems there is some extra space in a few of these. It also seems like some are not aligned.

Let me know if this is not really an issue.

clintoncwolfe:
Well, rubocop passes, which has always been my gatekeeper of style.  Do you feel we should update the rubocop rules?

miah (approved):
Thank you for taking the time to clean up existing code after changing filtertable =)

jquick (approved):
Very nice cleanup! Thanks @clintoncwolfe

jerryaldrichiii:
I think rubocop is a good start but cannot be the only gatekeeper

...but since I have no argument save for subjective ones:

jerryaldrichiii (approved):

PR #3130, opened 2018-06-12
Hard code ruby path in binstub to avoid cross-ruby contamination

scotthain:
In order to avoid other rubies (and since we control the inspec ruby path), we want to hardcode our ruby here so we don't use env and end up with an unexpected ruby

Signed-off-by: Scott Hain <shain@chef.io>

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -15,6 +15,7 @@ pkg_deps=(
   core/libxslt
   core/net-tools
   core/ruby
+  core/bundler
 )
 pkg_build_deps=(
   core/bundler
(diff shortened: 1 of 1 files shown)


schisamo (approved):

scotthain:
This does not appear to have all the fixes needed to make it work... I'll dig in more.

clintoncwolfe:
This is targeting the 1.x release branch - master / 2.x is current. Is that intentional?

scotthain:
@clintoncwolfe yes - the code in 2.x has been rewritten and I believe it works correctly. (We are stuck pinning to 1.x until some other issues are resolved...)

jquick (approved):
Thanks @scotthain

jquick:
@scotthain, you mind fixing that one test? you should just be able to remove the line from:

jquick:
Ill fix the 1.stable test in a follow up PR.

PR #3132, opened 2018-06-13
deprecate azure_generic_resource

### LINKED ISSUE #3131: Deprecate Azure Generic Resources
### 🎛 Description

In InSpec we want to ensure by design that we build a predictable system. That means need need to minimize runtime issues and offer a stable resource definition. The `azure_generic_resource` proofed to be not stable and do not recommend its use.

2.x

### 🤔 Replication Case

See https://github.com/inspec/inspec/issues/3096

### 💁 Possible Solutions

Do not use the `azure_generic_resource`. Deprecate the resource.

### 💻 Stacktrace

n/a

chris-rock:
fixes #3131

### DIFF
--- a/lib/resources/azure/azure_generic_resource.rb
+++ b/lib/resources/azure/azure_generic_resource.rb
@@ -16,7 +16,7 @@ module Inspec::Resources
     attr_accessor :filter, :total, :counts, :name, :type, :location, :probes
 
     def initialize(opts = {})
-      warn '[DEPRECATED] use specific azure resources instead. See https://github.com/inspec/inspec/issues/3131'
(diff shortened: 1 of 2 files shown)


lib/resources/azure/azure_generic_resource.rb
```diff
@@ -16,6 +16,8 @@ class AzureGenericResource < AzureResourceBase
     attr_accessor :filter, :total, :counts, :name, :type, :location, :probes
 
     def initialize(opts = {})
+      warn '[DEPRECATED] use specific azure resources instead. See https://github.com/inspec/inspec/issues/3131'
```
clintoncwolfe: Might want to mention that 'azure_generic_resource' was used - otherwise this might be a confusing message.

clintoncwolfe (approved):

jquick (approved):
Thanks!

Issue #3123, opened 2018-06-08
apache_conf resource needs better parsing support of apache configuration files


cchaffee:
@jerryaldrichiii From the chats in the inspec channel yesterday.

### 🎛 Description

**Choose one:**  🙋 feature request

`apache_conf` inspec resource needs to handle parsing keys and values of the `httpd.conf` better.

### 🌍 InSpec and Platform Version

Inspec version 2.1.43
Platform: CentOS 7

### 🤔 Replication Case

Steps to replicate:
1. Grab a default created `httpd.conf` (or use the one attached in the tarball)
2. Open an `inspec shell` with the file locally or on a remote target
```
You are currently running on:

    Name:      centos
    Families:  redhat, linux, unix, os
    Release:   7.4.1708
    Arch:      x86_64

inspec> test = apache_conf("/etc/httpd/conf/httpd.conf")
=> Apache Config /etc/httpd/conf/httpd.conf
inspec> test.DocumentRoot
=> ["\"/var/www/html\""]
```

The value of the `DocumentRoot` key in the configuration file is adding extra forward slashes and quotations. This makes it difficult to use in inspec tests. Currently the file used has the key and value like so:

`DocumentRoot "/var/www/html"`

It should look more like `["/var/www/html"]`. This is one such example. Another is that the `apache_conf` resource does not support parsing the XML like tags that exist in the `httpd.conf` file.

Example:

```
<Directory "/var/www">
    AllowOverride None
    # Allow open access:
    Require all granted
</Directory>
```
Cannot parse this properly to use in tests utilizing the `apache_conf` resource. Need to have a way to specify which "Directory" block we want to look at, and then the underlying keys/values in that block (such as `AllowOverride` in this case).

aaronlippold:
@jquick @jerryaldrichiii and folks - we touched on this a while back if I recall. Thoughts?

jerryaldrichiii:
I don't recall. That being said it looks like I have enough info here to try and reproduce.

Could you re-upload the httpd conf? I can't seem to extract it. If it's just a single text file then you could upload it as plain text or use a Gist too.

cchaffee:
https://gist.github.com/cchaffee/4b2504ef0cc20dc0aae8658fb3ab3244

Gist with source code of httpd.conf

jerryaldrichiii:
Thank you @cchaffee! I'll give this at the start of next week.

jerryaldrichiii:
[...]

PR #3142, opened 2018-06-14
apache_conf resource: Strip quotes from values

jerryaldrichiii:
This removes quotes from values captured by SimpleConfig.

Example:

`['\"/etc/apache2\"']`

becomes

`['/etc/apache2']`

This solves the bug portion of #3123

### DIFF
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -87,8 +87,10 @@ module Inspec::Resources
         ).params
 
         # Capture any characters between quotes that are not escaped in values
-        params.values.map! do |x|
-          x.map! { |y| y[/(?<=["|'])(?:\\.|[^"'\\])*(?=["|'])/] || y }
+        params.values.map! do |value|
+          value.map! do |sub_value|
+            sub_value[/(?<=["|'])(?:\\.|[^"'\\])*(?=["|'])/] || sub_value
+          end
         end
 
         @params.merge!(params)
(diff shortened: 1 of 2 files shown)


aaronlippold:
👍 thanks for the quick turn around on this?

jquick:
Would we ever have legitimate spaces in the quotes for any parameters?

aaronlippold:
Not that I recall but I am not a totally an apache conf expert

jerryaldrichiii:
@jquick has a point though. I'll update the regex to capture any chars between quotes (even escaped quotes).

miah (approved):

lib/resources/apache_conf.rb
```diff
@@ -85,6 +85,12 @@ def read_content
           assignment_regex: /^\s*(\S+)\s+((?=.*\s+$).*?|.*)\s*$/,
           multiple_values: true,
         ).params
+
+        # Capture any characters between quotes that are not escaped in values
+        params.values.map! do |x|
```
clintoncwolfe: Could we not use `x` and `y` here?

clintoncwolfe (changes requested):
Keyboards now come with **multiple** letters!

lib/resources/apache_conf.rb
jerryaldrichiii: Sure can, I changed it to `value` and `sub_value`.

Let me know if you have better names...I'm bad with naming things.

clintoncwolfe (approved):
Thanks @jerryaldrichiii

PR #3151, opened 2018-06-15
auditd resource: Add handling for sudo/no command

### LINKED ISSUE #3021: `auditd` resource should not ostensibly work without --sudo
### 🎛 Description

If you try to work with the `auditd` resource without sudo permissions; no errors occur and the resource appears to indicate there are no audit rules running.
This is because the resource [only looks at stdout](https://github.com/chef/inspec/blob/master/lib/resources/auditd.rb#L33) and not stderr or the exit code of `auditctl -l`

```
    def initialize
      @content = inspec.command('/sbin/auditctl -l').stdout.chomp
```

2.1.54 from bundle install

### 💁 Possible Solutions

 * Raise an error if using this resource without root permissions

AND/OR

 * have `initialize` check stderr and exit code of `auditctl` and raise if they are not normal, instead of assuming stdout will contain the audit rules

jerryaldrichiii:
This adds error handling for the cases where the `auditctl` command does not exist or does not have the correct permissions.

This closes https://github.com/inspec/inspec/issues/3021


miah (approved):

test/helper.rb
```diff
@@ -218,6 +218,7 @@ def md.directory?
       'bash -c \'type "/test/path/pip"\'' => empty.call,
       'bash -c \'type "Rscript"\'' => empty.call,
       'bash -c \'type "perl"\'' => empty.call,
+      'bash -c \'type "/sbin/auditctl"\'' => empty.call,
```
clintoncwolfe, line 221: Does this actually exercise the test?

clintoncwolfe (commented):
Just a question about testing, otherwise looks fine.

test/helper.rb
jerryaldrichiii: I'm not quite sure what you're asking. This just allows `inspec.command('/sbin/auditctl').exist?` to pass.

clintoncwolfe: 👍

clintoncwolfe (approved):
shiiiiiip ittttt

jquick (approved):
Thanks @jerryaldrichiii

lib/resources/auditd.rb
```diff
@@ -30,11 +30,26 @@ class AuditDaemon < Inspec.resource(1)
...
+        raise Inspec::Exceptions::ResourceFailed,
+              'Command `/sbin/auditctl` does not exist'
+      end
+
+      auditctl_cmd = '/sbin/auditctl -l'
+      result = inspec.command(auditctl_cmd)
+
+      if result.exit_status != 0
+        raise Inspec::Exceptions::ResourceFailed,
+              "Command `#{auditctl_cmd}` failed with error: #{result.stderr}"
```
aaronlippold, line 43: Did we want to pass up a note to the user here that they cannot use this resources without escalated privileges? ( e.g. root or sudo )

jerryaldrichiii: I originally had thought that, but then I found out that the stderr when that is the case is informative enough. 

```
[vagrant@localhost ~]$ auditctl -l
You must be root to run this program.
```

PR #3152, opened 2018-06-15
Overhaul Glossary

clintoncwolfe:
This PR makes three major changes to the glossary on the docs site:
1. Re-works the visual glossary example to be a concrete example (car/cars) instead of the 'foo' resource with 'jared' property.
2. Re-names the "Advanced usage" section of the visual glossary to just be the plural example section.  We might add an advanced usage section later.
3. Largest change by linecount was adding a full text glossary with extensive terms, prose, examples, and cross-linking.  

A big side-effect of this PR is that now we'll be able to link to individual glossary terms.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -144,7 +144,7 @@ A [resource](#resource) that is _not_ included with InSpec. It may be a resource
 ### describe
 ### describe block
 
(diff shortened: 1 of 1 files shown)


kagarmoe (approved):
Tinkered with your wording.

jquick (approved):
Thanks @clintoncwolfe

PR #3156, opened 2018-06-19
Overhaul shadow docs

### LINKED ISSUE #496: Add documentation for shadow resource

clintoncwolfe:
Rewrites the shadow docs to clarify use of filter criteria, difference between that and a property, the fact that all the properties are lists, and to clarify that two of the fields are relative to the Unix epoch.

Fixes #496
Attn: @james-stocks

### DIFF
--- a/docs/resources/shadow.md.erb
+++ b/docs/resources/shadow.md.erb
@@ -5,148 +5,255 @@ platform: linux
 
 # shadow
 
-Use the `shadow` InSpec audit resource to test the contents of `/etc/shadow`, which contains password details that are only readable by the `root` user. The format for `/etc/shadow` includes:
+Use the `shadow` InSpec audit resource to test the contents of `/etc/shadow`, which contains password details that are only readable by the `root` user. `shadow` is a [plural resource](https://www.inspec.io/docs/reference/glossary/#plural_resource), which means it functions by performing searches across multiple entries in the shadow file.
+
+ The format for `/etc/shadow` includes:
 
 * A username
 * The hashed password for that user
-* The last time a password was changed
+* The last date a password was changed, as the number of days since Jan 1 1970
 * The minimum number of days a password must exist, before it may be changed
 * The maximum number of days after which a password must be changed
 * The number of days a user is warned about an expiring password
 * The number of days a user must be inactive before the user account is disabled
-* The number of days a user account has been disabled
+* The date on which a user account was disabled, as the number of days since Jan 1 1970
 
 These entries are defined as a colon-delimited row in the file, one row per user:
 
     dannos:Gb7crrO5CDF.:10063:0:99999:7:::
 
+The `shadow` resource understands this format, allows you to search on the fields, and exposes the selected users' properties.
+
 <br>
 
-## Syntax
+## Resource Parameters
 
-A `shadow` resource block declares user properties to be tested:
+The `shadow` resource takes one optional parameter: the path to the shadow file.  If omitted, `/etc/shadow` is assumed.
 
+    # Expect a file to exist at the default location and have 32 users
     describe shadow do
-      its('users') { should_not include 'forbidden_user' }
+      its('count') { should eq 32 }
     end
 
-Properties can be used as a single query:
-
-    describe shadow.user('root') do
-      its('count') { should eq 1 }
+    # Use a custom location
+    describe shadow('/etc/my-custom-place/shadow') do
+      its('count') { should eq 32 }
     end
-
-Use the `.where` method to find properties that match a value:
-
-    describe shadow.where { min_days == '0' } do
-      its ('users') { should include 'nfs' }
-    end
-
-    describe shadow.where { password =~ /[x|!|*]/ } do
-      its('count') { should eq 0 }
-    end
-
-The following properties are available:
-
-* `users`
-* `passwords`
-* `last_changes`
-* `min_days`
-* `max_days`
-* `warn_days`
-* `inactive_days`
-* `expiry_dates`
-* `reserved`
-
-<br>
 
 ## Examples
 
-The following examples show how to use this InSpec audit resource.
+A `shadow` resource block uses `where` to filter entries from the shadow file. If `where` is omitted, all entries are selected.
 
-### Test for a forbidden user
-
+    # Select all users. Among them, there should not be a user with the name 'forbidden_user'.
     describe shadow do
       its('users') { should_not include 'forbidden_user' }
     end
 
-### Test that a user appears one time
-
-    describe shadow.users('bin') do
-      its('passwords') { should cmp 'x' }
+    # Ensure there is only one user named 'root' (Select all with name 'root', then count them).
+    describe shadow.where(user: 'root') do
       its('count') { should eq 1 }
     end
+
+You can use `where` to match any of the [filter criteria](#filter_criteria) supported. `where` has a method form for simple equality, and a block form for more complex queries.
+
+    # Method form, simple
+    # Select just the root user (direct equality)
+    describe shadow.where(user: 'root') do
+      its ('count') { should eq 1 }
+    end
+
+    # Method form, with a regex
+    # Select all users whose names begin with smb
+    describe shadow.where(user: /^smb/) do
+      its ('count') { should eq 2 }
+    end
+
+    # Block form
+    # Select users whose passwords have expired
+    describe shadow.where { expiry_date > 0 } do
+      # This is more clear a a test
+      its('count') { should eq 0 }
+      # But if the count test fails, this test gives clearer diagnostics
+      its('users') { should be_empty }
+    end
+
 
 <br>
 
+## Filter Criteria
+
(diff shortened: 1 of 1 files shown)


docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,255 @@ platform: linux
...
+    # Select all users whose names begin with smb
+    describe shadow.where(user: /^smb/) do
+      its ('count') { should eq 2 }
     end
 
-The following properties are available:
+    # Block form
+    # Select users whose passwords have expired
+    describe shadow.where { expiry_date > 0 } do
+      # This is more clear a a test
```
james-stocks: Suggest `Directly assert that there are 0 such users`

docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,255 @@ platform: linux
...
+      its('count') { should eq 0 }
+      # But if the count test fails, this test gives clearer diagnostics
```
james-stocks: Suggest `# But if the count test fails, this test will output the users that are causing the failure`

docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,255 @@ platform: linux
...
+    describe shadow.where { warn_days != 7 } do
+      its('users') { should be_empty }
+    end
+
+### inactive_days
+
+An integer reflecting the number of days that must pass before a user who has not logged in will be disabled.
+
+
+    # Ensure everyone has a stale policy of no more thn 14 days.
```
james-stocks: Typo `thn`

docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,255 @@ platform: linux
...
+    describe shadow.where { expiry_date.nil? } do
+      its('users') { should be_empty }
+    end
+
+    # Ensure no one is disabled for more than 14 days
+    describe shadow.where { !expiry_date.nil? && expiry_date - Date.new(1970,1,1) > 14} do
+      its('users') { should be_empty }
+    end
 
 ## Properties
```
james-stocks: Would it be helpful to move this `Properties` section before the `Filter Criteria` section?

The example on line 93 uses `password.uniq` and treats the result as an array before this section explains that concept

docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,255 @@ platform: linux
...
 
 ### max\_days
 
-The `max_days` property tests the maximum number of days after which a password must be changed:
+A list of integers reflecting the maximum number of days after which the password must be changed for each user matching the filter.
 
-    its('max_days') { should eq 90 }
+    # Make sure there is no policy allowing longer than 90 days
+    describe shadow do
+      its('max_days.max') { should be < 90 }
```
james-stocks, line 129: If `max_days` is optional then some values might be `nil` and `max_days.max` would throw something like `ArgumentError: comparison of NilClass with 12 failed`

I guess `where { !max_days.nil? } ` is the right way to guard against that. If so, it might be better to leave the example as it currently is rather than mix the `where` concept into these straightforward examples

james-stocks:
Awesome @clintoncwolfe !
I left some minor feedback comments

docs/resources/shadow.md.erb
clintoncwolfe: OK, let's try that.

docs/resources/shadow.md.erb
clintoncwolfe: So, two thoughts on this....
* `where` is not intended to be an advanced concept, or one unique to `shadow`.  All plural resources support it.  I don't want to be skittish about adding it where it makes a compelling example.
* That said, if a column can have nils, a simpler thing is to set the column to be `style: :simple`.  That calls uniq, flatten, and compact (compact discards nils).  That changes the semantics of the property a bit, but I think that's more useful here.

docs/resources/shadow.md.erb
```diff
@@ -5,148 +5,253 @@ platform: linux
...
+An integer reflecting the number of days that must pass before a user who has not logged in will be disabled.
+
+    # Ensure everyone has a stale policy of no more than 14 days.
+    describe shadow.where { inactive_days.nil? || inactive_days > 14 } do
+      its('users') { should be_empty }
+    end
+
+### expiry_date
+
+An integer reflecting the number of days since Jan 1, 1970 on which the user was disabled. The value `nil` means the account was disabled because of an expired password.
```
kagarmoe: Please check if this is correct.

docs/resources/shadow.md.erb
james-stocks: I don't understand the context for "set the column to be `style: :simple`" - if that's a universal change to how filter tables are used, there might be some cases where it's not semantically correct (here we can discard `nils` as those entries will be compliant, but for other controls `nil` values might be non-compliant)

For your first thought - updating the example to use `where` would be good. I partly think it's OK for a documentation example to not be safe for all edge cases if that lets it focus on the one thing its trying to illustrate to the reader. But at the same time you're right we shouldn't be skittish and should maybe just give examples that we'd consider fit-for-purpose ourselves

clintoncwolfe: Just to close the loop on the `style: :simple` question, it's not FIlterTable-wide - it's a setting we can change on a column-by-column basis. The [color property of the filtertable docs](https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md#how-do-i-declare-my-interaction-with-filtertable) demonstrates it, with details [here](https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md#a-colors-method-that-will-return-a-flattened-and-uniqd-array-of-values) about how its behavior changes.

miah (approved):

docs/resources/shadow.md.erb
```diff
@@ -238,7 +238,7 @@ An integer reflecting the number of days that must pass before a user who has no
 
 ### expiry_date
 
-An integer reflecting the number of days since Jan 1, 1970 on which the user was disabled. The value `nil` means the account was disabled because of an expired password.
+An integer reflecting the number of days since Jan 1, 1970 on which the user was disabled. The value `nil` means the account has not been disabled.
```
kagarmoe: "has not been disabled" is both passive & subjunctive. Can you find a solution using the present tense and the active voice.

docs/resources/shadow.md.erb
```diff
@@ -5,7 +5,7 @@ platform: linux
 
 # shadow
 
-Use the `shadow` InSpec audit resource to test the contents of `/etc/shadow`, which contains password details that are readable only by the `root` user. `shadow` is a [plural resource](https://www.inspec.io/docs/reference/glossary/#plural_resource). It functions by performing searches across multiple entries in the shadow file.
+Use the `shadow` InSpec audit resource to test the contents of `/etc/shadow`, which contains password details that are readable only by the `root` user. `shadow` is a [plural resource](https://www.inspec.io/docs/reference/glossary/#plural_resource). Like all plural resources, it functions by performing searches across multiple entries in the shadow file.
```
kagarmoe: This sentence means that "All plural resources search across multiple files _in the shadow directory_."

docs/resources/shadow.md.erb
```diff
@@ -81,7 +81,7 @@ Use `where` to match any of the supported [filter criteria](#filter_criteria). `
 
 ## Properties
 
-As a [plural resource](https://www.inspec.io/docs/reference/glossary/#plural_resource), all of `shadow`'s properties return lists (that is, Ruby arrays). `include` and `be_empty` are two useful matchers when working with lists. You can also perform manipulation of the lists, such as calling `uniq`, `sort`, `count`, `first`, `last`, `min`, and `max`.
+As a [plural resource](https://www.inspec.io/docs/reference/glossary/#plural_resource), all of `shadow`'s properties return lists (that is, Ruby Arrays). `include` and `be_empty` are two useful matchers when working with lists. You can also perform manipulation of the lists, such as calling `uniq`, `sort`, `count`, `first`, `last`, `min`, and `max`.
```
kagarmoe: Data types are not proper nouns. I won't die on this hill, though.

docs/resources/shadow.md.erb
clintoncwolfe: It's not a directory, it is a file.

docs/resources/shadow.md.erb
clintoncwolfe: Ruby classes are proper nouns. Here I'm referring to the Ruby [Array](http://ruby-doc.org/core-2.4.2/Array.html) class.

docs/resources/shadow.md.erb
kagarmoe: Do all plural resources perform searches across multiple entries _in the shadow file_?

docs/resources/shadow.md.erb
kagarmoe: http://web.mit.edu/course/21/21.guide/capitals.htm , see 
_General Guidelines for Capitalizing Scientific Terms_. But as I said, this really isn't my hill.

jquick (approved):
Thanks @clintoncwolfe

PR #3157, opened 2018-06-19
Add functional tests for nested attributes

clintoncwolfe:
On #2991 , we have a PR in which a contributor aimed to simplify access to hierarchical attributes (that is, arrays of arrays, hashes of arrays, etc, in YAML).  Discussion went towards adding `[]` to Attribute, so it would appear natural.  On #2991, unit tests were added.  However, no functional tests were on hand - so the YAML parser and Attribute were never tested directly.

It turns out that without #2991, Attribute already works for arrays of arrays, hashes of hashes, hashes of arrays, but not well for arrays of hashes (though that may be a flaw in rspec-its).

Additionally, `attribute(...)` was found to have two peculiar properties:
* it can only be used in profile context (not within a control, or a describe block)
* it must key all attribute lookups using a string, and silently misses of a Symbol is provided

Both of those could be improved.

This PR simply adds functional tests to exercise attributes.

### DIFF
--- a/test/unit/mock/profiles/attributes/controls/flat.rb
+++ b/test/unit/mock/profiles/attributes/controls/flat.rb
@@ -1,0 +1,29 @@
+expecteds = {
+  an_integer: 1,
+  a_quoted_string: 'Should not have quotes',
+  an_unquoted_string: 'Should not have quotes',
+  lowercase_true: true,
+  titlecase_true: true,
+  titlecase_false: false,
+}
+tests = expecteds.keys.map do |test_name|
+  {
+    name: test_name,
+    expected: expecteds[test_name],
+    attr_via_string: attribute(test_name.to_s, default: "#{test_name}_default"),
+    attr_via_symbol: attribute(test_name, default: "#{test_name}_default"),
+  }
+end
+
+control 'flat' do
+  tests.each do |info|
+    describe "#{info[:name]} using string key" do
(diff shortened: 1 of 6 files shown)


jerryaldrichiii (approved):
Yay! More tests! Safety++

test/unit/mock/profiles/attributes/controls/flat.rb
```diff
@@ -0,0 +1,29 @@
...
+  }
+end
+
+control 'flat' do
+  tests.each do |info|
+    describe "#{info[:name]} using string key" do
+      subject { info[:attr_via_string] }
+      it { should eq info[:expected] }
+    end
+    # describe "#{info[:name]} using symbol key" do
```
jerryaldrichiii: Any reason why this is left in?

clintoncwolfe: Yes, on #2991 , there was some discussion about wanting to be able to use symbols as hash accessors.  That's much more rubyish.  Currently the keys are always Strings.  

We can remove that commented-out test if desired.

jerryaldrichiii: I'm for removing it, but I'll leave that up to you.

jquick (approved):
Thanks @clintoncwolfe

Issue #3135, opened 2018-06-13
Inspec exec report for inherited profiles with different name


alexpop:
### 🎛 Description

🐛 bug report

The inspec report for a wrapper profile with the following `inspec.yml`:
```
name: cis-rhel7-level2-server-wrapper
...
depends:
  - name: cis-rhel7-level2-server-v2.1.1
    compliance: admin/cis-rhel7-level2-server
```

will have this content:
```
{
  "profiles": [
    {
      "name": "cis-rhel7-level2-server-wrapper",
      "depends": [
        {
          "name": "cis-rhel7-level2-server-v2.1.1",
          "compliance": "admin/cis-rhel7-level2-server"
        }]
      ...
    },
    {
      "name": "cis-rhel7-level2-server",
      "version": "2.1.1-18",
      ...
    }
  ]
}
```

without an easy way to link depends name `cis-rhel7-level2-server-v2.1.1` with inherited profile name `cis-rhel7-level2-server`
### 🌍 InSpec and Platform Version

```
inspec version 2.1.81
```

### 💁 Possible Solutions

Add a `profile_name` or similar field in depends with the actual name of the profile: `cis-rhel7-level2-server`

clintoncwolfe:
@jquick - is this an example of the "inheritance json report" issue?

jquick:
This will be fixed by the inheritance redo in 3.0 but is a smaller subset of an issue. For now we are going to add another field into the json report so A2 can match child profiles accordingly.

jquick:
This actually may be a little difficult. The profile has no idea what its upstream name is or its location. So we can do some fuzzy matching but I don't see an easy 1:1 map here. May end up having to inject them during the loading.

PR #3164, opened 2018-06-22
Set parent_profile field on child profiles (json report)

jquick:

This change adds a optional `parent_profile` filed to the profile info section. This is used when a profile depends on others and they are loaded. This addition will make it possible to link child profiles to parent profiles when digesting the json report.

Fixes https://github.com/inspec/inspec/issues/3135

### DIFF
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -14,9 +14,9 @@ module Inspec
     # @param cwd [String] Current working directory for relative path includes
     # @param vendor_path [String] Path to the vendor directory
     #
-    def self.from_lockfile(lockfile, cwd, cache, backend, opts = {})
+    def self.from_lockfile(lockfile, cwd, cache, backend, opts = {}, parent_profile = nil)
       dep_tree = lockfile.deps.map do |dep|
-        Inspec::Requirement.from_lock_entry(dep, cwd, cache, backend, opts)
+        Inspec::Requirement.from_lock_entry(dep, cwd, cache, backend, opts, parent_profile)
       end
 
       dep_list = flatten_dep_tree(dep_tree)
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -14,9 +14,11 @@ module Inspec
     # @param cwd [String] Current working directory for relative path includes
     # @param vendor_path [String] Path to the vendor directory
     #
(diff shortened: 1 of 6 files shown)


lib/inspec/dependencies/dependency_set.rb
```diff
@@ -14,9 +14,9 @@ class DependencySet
     # @param cwd [String] Current working directory for relative path includes
     # @param vendor_path [String] Path to the vendor directory
     #
-    def self.from_lockfile(lockfile, cwd, cache, backend, opts = {})
+    def self.from_lockfile(lockfile, cwd, cache, backend, opts = {}, parent_profile = nil)
```
clintoncwolfe: Placing a positional arg after the hash arg is unrubyish - usually the last literals are swept into an implicit hash.  I'm not sure if / how this would work.  Can `parent_profile` appear before `opts`, or be made one of the opts?

lib/inspec/dependencies/dependency_set.rb
```diff
@@ -14,9 +14,9 @@ class DependencySet
...
       dep_tree = lockfile.deps.map do |dep|
-        Inspec::Requirement.from_lock_entry(dep, cwd, cache, backend, opts)
+        Inspec::Requirement.from_lock_entry(dep, cwd, cache, backend, opts, parent_profile)
```
clintoncwolfe: Same concern - positional arg vs implicit hash

lib/inspec/dependencies/requirement.rb
```diff
@@ -25,29 +25,31 @@ def self.from_metadata(dep, cache, opts)
           opts.merge(dep))
     end
 
-    def self.from_lock_entry(entry, cwd, cache, backend, opts = {})
+    def self.from_lock_entry(entry, cwd, cache, backend, opts = {}, parent_profile = nil)
```
clintoncwolfe: Same concern - positional arg vs implicit hash

clintoncwolfe (changes requested):
Some concern about the arg list in a couple of places.  And CI is unhappy.  Otherwise OK by me!

lib/inspec/dependencies/dependency_set.rb
jquick: Yeah, my original stab at this put it in options but thats really for target options and gets pulled into the profile as such. It felt ugly putting stuff in there for transport and then removing it later as it does not belong. I'll rearrange the args to have the hash last.

miah (approved):

clintoncwolfe (approved):
Looks good, thanks!

Issue #2106, opened 2017-08-30
be_linked_to resolves symlinks too deep


zbikmarc:
### Description
`be_linked_to` of `file` resource resolves links "too deep".
It may be similar or same as https://github.com/chef/inspec/issues/974

Image situation, that we have application residing in directory which name contains version, like `/opt/myApp-1.0`. We created non-versioned/latest symlink to this directory `/opt/myApp-1.0 -> /opt/myApp`. And in this directory there are few other directories, one of them is`config`. Now we want this `config` directory (and only this) also be available in `/apps/myApp/config`. And we want to check if `/apps/myApp/config` really points to `/opt/myApp/config`, so non-versioned directory of our application.

```
# tree /opt
/opt/
(... omitted ...)
├── myApp -> /opt/myApp-1.0
└── myApp-1.0
    └── config

# tree /apps
/apps/
└── myApp
    └── config -> /opt/myApp/config

# ls -ld /apps/myApp
drwxr-xr-x. 2 root root 20 Aug 30 09:02 /apps/myApp
# ls -ld /apps/myApp/config
lrwxrwxrwx. 1 root root 17 Aug 30 09:02 /apps/myApp/config -> /opt/myApp/config
```

```
irb(main):001:0> File.symlink?('/apps/myApp/config')
=> true
irb(main):002:0> File.readlink('/apps/myApp/config')
=> "/opt/myApp/config"
irb(main):003:0> File.realpath('/apps/myApp/config')
=> "/opt/myApp-1.0/config"
```

inSpec test file:
```ruby
control "check-symlink-1.0" do
  impact 1.0
  title "Check symlinks"

  describe file('/apps/myApp/config') do
    it { should be_symlink }
    it { should be_linked_to '/opt/myApp/config' }     # this should pass
    its(:link_path) { should eq '/opt/myApp/config' }  # this is only for "showcase" purposes
  end
end
```

Result of inSpec test run:
```
Profile: tests from symlink_bug.rb
Version: (not specified)
Target:  local://

  ×  check-symlink-1.0: Check symlinks (2 failed)
     ✔  File /apps/myApp/config should be symlink
     ×  File /apps/myApp/config should be linked to "/opt/myApp/config"
     expected `File /apps/myApp/config.linked_to?("/opt/myApp/config")` to return true, got false
     ×  File /apps/myApp/config link_path should eq "/opt/myApp/config"

     expected: "/opt/myApp/config"
          got: "/opt/myApp-1.0/config"

[...]

PR #3168, opened 2018-06-23
Add support for shallow link paths

### LINKED ISSUE #3100: file resource link_path always traces link to the target
### 🎛 Description

bug report
could be related to #3099
situation:
```
# ls -l /another_link /link
lrwxrwxrwx. 1 root root  4 May 31 17:47 /another_link -> /usr
lrwxrwxrwx. 1 root root 13 May 31 17:47 /link -> /another_link
```
I need to check that /link points to /another_link and I don't care where /another_link points, but it looks like impossible, because:
```
inspec> file('/link').link_path
=> "/usr"
```

inspec --version
2.1.72

### 🤔 Replication Case

see above

ColinHebert:
This should wait on inspec/train#309

Fixes #3100 and #2106

### DIFF
--- a/docs/resources/file.md.erb
+++ b/docs/resources/file.md.erb
@@ -33,7 +33,7 @@ content, size, basename, path, owner, group, type
 
 ### Unix/Linux Properties
 
-symlink, mode, link_path, mtime, size, selinux\_label, md5sum, sha256sum, path, source, source\_path, uid, gid
+symlink, mode, link_path, shallow_link_path, mtime, size, selinux\_label, md5sum, sha256sum, path, source, source\_path, uid, gid
 
 ### Windows Properties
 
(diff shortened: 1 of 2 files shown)


docs/resources/file.md.erb
```diff
@@ -74,10 +74,17 @@ The following examples show how to use this InSpec audit resource.
...
 
 The `link_path` property tests if the file exists at the specified path. If the file is a symlink,
-InSpec will resolve the symlink and return the ultimate linked file.
+InSpec will resolve the symlink recursively and return the ultimate linked file.
 
     its('link_path') { should eq '/some/path/to/file' }
 
+### shallow_link_path
+
+The `link_path` property tests if the file exists at the specified path. If the file is a symlink,
```
clintoncwolfe: I feel this is an unclear explanation.  A [property](https://www.inspec.io/docs/reference/glossary/#property) doesn't test anything, it returns a value.  Here, you might say, "The `shallow_link_path` property returns the path that the file refers to, only resolving it once (that is, it performs a `readlink` operation).  If the file is not a symlink, `nil` is returned."

clintoncwolfe (changes requested):
Just a documentation clarification - otherwise looks great, thanks @ColinHebert !

docs/resources/file.md.erb
ColinHebert: Fair, updating to reflect that text

clintoncwolfe (approved):
Thanks - looks good @ColinHebert!

jquick (approved):
Thanks @ColinHebert

Issue #3169, opened 2018-06-25
Resource oracledb_session broken if query result contains dozens of records


voroniys:
### 🎛 Description

*🐛 bug report 

If result of the query contains many records html parser is broken due to repeating header entry

### 🌍 InSpec and Platform Version

2.2.16

### 🤔 Replication Case

Do a query that returns dozens of rows

### 💁 Possible Solutions

In the options for sqlplus option SET PAGESIZE 32000 needs to be added

Nik08:
This issue seems to have been resolved in 2018 with [#3170](https://github.com/inspec/inspec/pull/3170). But I verified this again myself and this seems to be working.

I tried this:

```
sql = oracledb_session(user: 'USERNAME', password: 'PASS', service: 'XE', host: 'HOST')
describe sql.query("SELECT UPPER(VALUE) AS VALUE FROM V$PARAMETER").rows do
  its('count') { should eq 434 }
end
```

Thanks!

PR #3170, opened 2018-06-25
A number of bug fixes and new features for oracledb_session resource

voroniys:
Started with fixing the issue #3169 and finished with extended resource, which covers my real life needs in checking Oracle query results.
The bug fixes includes:
* If user specify sqlplus_bin always use it, even if sql exists 
* reformat command for execution to make sudo working. So I can specify:
sqlplus_bin: 'su - oracle sqlplus'.
The new features include:
* methods 'rows' and 'column' which allows easier processing of multi-rows results
* make possible to execute queries as sysdba/sysoper/sysasm

All fixes and features are in separate commits to make possible partial merge.

### DIFF
--- a/docs/resources/oracledb_session.md.erb
+++ b/docs/resources/oracledb_session.md.erb
@@ -25,6 +25,11 @@ where
 
 <br>
 
+## Resource Properties
+* rows the query result as array of hashes
+* row(id)  selected row from query result
+* column(name) array with values from selected column
+ 
 ## Examples
 
 The following examples show how to use this InSpec audit resource.
@@ -43,6 +48,14 @@ The following examples show how to use this InSpec audit resource.
 
     describe sql.query('SELECT NAME FROM v$database;').row(0).column('name') do
       its('value') { should cmp 'ORCL' }
+    end
+
+### Test for table contains a specified value in any row for the given column name
+
+    sql = oracledb_session(user: 'my_user', pass: 'password')
+
+    describe sql.query('SELECT * FROM my_table;').column('my_column') do
+      it { should include 'my_value' }
     end
 
 <br>
--- a/docs/resources/oracledb_session.md.erb
+++ b/docs/resources/oracledb_session.md.erb
@@ -28,7 +28,7 @@ where
 
 ## Resource Properties
 * rows the query result as array of hashes
-* row(id)  selected row from query result
+* row(number)  selected row from query result, where number is just a row number in the query result
 * column(name) array with values from selected column
  
 ## Examples
--- a/docs/resources/oracledb_session.md.erb
+++ b/docs/resources/oracledb_session.md.erb
@@ -20,7 +20,7 @@ A `oracledb_session` resource block declares the username and password to use fo
 where
 
 * `oracledb_session` declares a username and password with permission to run the query (required), and an optional parameters for host (default: `localhost`), SID (default: `nil`, which uses the default SID, and path to the sqlplus binary (default: `sqlplus`).
-* it is possible to run queries as sysdba/sysoper by setting password to value 'as sysdba', see examples
+* it is possible to run queries as sysdba/sysoper by using `as_db_role option`, see examples
 * `query('QUERY')` contains the query to be run
 * `its('value') { should eq('') }` compares the results of the query against the expected result in the test
 
@@ -53,7 +53,7 @@ The following examples show how to use this InSpec audit resource.
 
 ### Test for table contains a specified value in any row for the given column name
 
-    sql = oracledb_session(user: 'my_user', pass: 'password')
+    sql = oracledb_session(user: 'my_user', pass: 'password', service: 'MYSID')
 
     describe sql.query('SELECT * FROM my_table;').column('my_column') do
       it { should include 'my_value' }
@@ -62,11 +62,12 @@ The following examples show how to use this InSpec audit resource.
 ### Test tablespace exists as sysdba
     The check will change user (with su) to specified user and run 'sqlplus / as sysdba' (sysoper, sysasm) 
 
-    sql = oracledb_session(user: 'oracle', pass: 'as sysdba', service: 'MYSID')
+    sql = oracledb_session(as_os_user: 'oracle', as_db_role: 'sysdba', service: 'MYSID')
 
     describe sql.query('SELECT tablespace_name AS name FROM dba_tablespaces;').column('name') do
       it { should include 'MYTABLESPACE' }
     end
+    NOTE: option `as_os_user` available only on unix-like systems and not supported on Windows. Also this option requires that you are running inspec as `root` or with `--sudo`
 
 ### Test number of rows in the query result
 
--- a/docs/resources/oracledb_session.md.erb
+++ b/docs/resources/oracledb_session.md.erb
@@ -20,6 +20,7 @@ A `oracledb_session` resource block declares the username and password to use fo
 where
 
 * `oracledb_session` declares a username and password with permission to run the query (required), and an optional parameters for host (default: `localhost`), SID (default: `nil`, which uses the default SID, and path to the sqlplus binary (default: `sqlplus`).
+* it is possible to run queries as sydba/sysoper by set password to value 'as sysdba', see examples
 * `query('QUERY')` contains the query to be run
 * `its('value') { should eq('') }` compares the results of the query against the expected result in the test
 
@@ -56,6 +57,15 @@ The following examples show how to use this InSpec audit resource.
 
     describe sql.query('SELECT * FROM my_table;').column('my_column') do
       it { should include 'my_value' }
+    end
+
+### Test tablespace exists as sysdba
+    The check will change user (with su) to specified user and run 'sqlplus / as sysdba' (sysoper, sysasm) 
+
+    sql = oracledb_session(user: 'oracle', pass: 'as sysdba', service: 'MYSID')
+
+    describe sql.query('SELECT tablespace_name AS name FROM dba_tablespaces;').column('name') do
+      it { should include 'MYTABLESPACE' }
     end
 
 <br>
--- a/docs/resources/oracledb_session.md.erb
+++ b/docs/resources/oracledb_session.md.erb
@@ -26,7 +26,7 @@ where
 
 <br>
 
-## Resource Properties
+## oracledb_session(...).query method Properties
 * rows the query result as array of hashes
 * row(number)  selected row from query result, where number is just a row number in the query result
 * column(name) array with values from selected column
@@ -68,6 +68,23 @@ The following examples show how to use this InSpec audit resource.
       it { should include 'MYTABLESPACE' }
     end
 
+### Test number of rows in the query result
+
+    sql = oracledb_session(user: 'my_user', pass: 'password')
+
+    describe sql.query('SELECT * FROM my_table;').rows do
+      its('count') { should eq 20 }
+    end
+
+### Use data out of (remote) DB query to build other tests
+
+    sql = oracledb_session(user: 'my_user', pass: 'password', host: 'my.remote.db', service: 'MYSID')
+
+    sql.query('SELECT * FROM files;').rows.each do |file_row|
+      describe file(file_row['path']) do
+	its('owner') { should eq file_row['owner']}
+      end
+    end
 <br>
 
 ## Matchers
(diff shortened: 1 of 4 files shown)


docs/resources/oracledb_session.md.erb
```diff
@@ -20,11 +20,17 @@ A `oracledb_session` resource block declares the username and password to use fo
 where
 
 * `oracledb_session` declares a username and password with permission to run the query (required), and an optional parameters for host (default: `localhost`), SID (default: `nil`, which uses the default SID, and path to the sqlplus binary (default: `sqlplus`).
+* it is possible to run queries as sydba/sysoper by set password to value 'as sysdba', see examples
```
clintoncwolfe: typo: sydba, you want sySdba
grammo: "by set password" => " by  setting password"

docs/resources/oracledb_session.md.erb
```diff
@@ -45,6 +51,23 @@ The following examples show how to use this InSpec audit resource.
...
+    sql = oracledb_session(user: 'my_user', pass: 'password')
+
+    describe sql.query('SELECT * FROM my_table;').column('my_column') do
+      it { should include 'my_value' }
+    end
+
+### Test tablespace exists as sysdba
+    The check will change user (with su) to specified user and run 'sqlplus / as sysdba' (sysoper, sysasm) 
+
+    sql = oracledb_session(user: 'oracle', pass: 'as sysdba', service: 'MYSID')
```
clintoncwolfe: Hrm, I don't like that design - hiding the "as user" functionality in the password option.  Why not just add a new option?

lib/resources/oracledb_session.rb
```diff
@@ -49,19 +49,23 @@ def query(q)
...
-        opts = "SET MARKUP HTML ON\nSET FEEDBACK OFF"
+        opts = "SET MARKUP HTML ON\nSET PAGESIZE 32000\nSET FEEDBACK OFF"
         p = :parse_html_result
       end
 
       query = verify_query(escaped_query)
       query += ';' unless query.end_with?(';')
-      command = %{echo "#{opts}\n#{query}\nEXIT" | #{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service}}
+      if /as (sysdba|sysoper|sysasm)/i === @password
+        command = %{su - #{user} -c "env ORACLE_SID=#{@service} #{bin} / #{@password} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC"}
```
clintoncwolfe: I'm not familiar with oracle tools, so I searched a bit as this seemed like an odd construction.  I found [this](https://asktom.oracle.com/pls/apex/f?p=100:11:0::::P11_QUESTION_ID:409364600346264472); it makes me think that the two branches of this conditional might rely on different credential resolution systems; that would seem to be an unexpected side effect.

docs/resources/oracledb_session.md.erb
```diff
@@ -20,11 +20,17 @@ A `oracledb_session` resource block declares the username and password to use fo
...
 * `oracledb_session` declares a username and password with permission to run the query (required), and an optional parameters for host (default: `localhost`), SID (default: `nil`, which uses the default SID, and path to the sqlplus binary (default: `sqlplus`).
+* it is possible to run queries as sydba/sysoper by set password to value 'as sysdba', see examples
 * `query('QUERY')` contains the query to be run
 * `its('value') { should eq('') }` compares the results of the query against the expected result in the test
 
 <br>
 
+## Resource Properties
+* rows the query result as array of hashes
+* row(id)  selected row from query result
```
clintoncwolfe: I think (especially in a database context) the term `id` is misleading.  From the implementation, this appears to be simply the zero-based index into the result array.  I think we should clarify that.

lib/utils/database_helpers.rb
```diff
@@ -48,10 +48,22 @@ def successful?
...
+    end
+
     def row(id)
       SQLRow.new(self, @results[id])
     end
 
+    def column(column)
+      result = []
+      @results.each do |row|
+        result << row[column] if row.has_key?(column)
```
clintoncwolfe: Why the conditional here?  RDBMS query results are always regular; column presence can't vary by row.

docs/resources/oracledb_session.md.erb
```diff
@@ -45,6 +51,23 @@ The following examples show how to use this InSpec audit resource.
...
+
+### Test tablespace exists as sysdba
+    The check will change user (with su) to specified user and run 'sqlplus / as sysdba' (sysoper, sysasm) 
+
+    sql = oracledb_session(user: 'oracle', pass: 'as sysdba', service: 'MYSID')
+
+    describe sql.query('SELECT tablespace_name AS name FROM dba_tablespaces;').column('name') do
+      it { should include 'MYTABLESPACE' }
+    end
+
```
clintoncwolfe: Could you include some examples using the `rows` functionality?

docs/resources/oracledb_session.md.erb
clintoncwolfe: Also note that `rows` returns a simple Array of Hashes, while `row[index]` returns a SQLRow object, which is a much wierder thing than a hash.  It seems odd that they are different.

If you want to ease column / cell access, you might implement `[]` on the SQLRows and SQLColumn Classes.

clintoncwolfe (changes requested):
I like some of the ideas here, but I think we need to clarify things a bit and make sure we don't confuse things.  I'd also like someone with more Oracle experience than I to poke at this.

docs/resources/oracledb_session.md.erb
voroniys: Hi,
thanks for quick review - very much appreciated. The typos in docs already fixed.
Here I'll try to explain why I've done it that way.
My first solution was a new option, but later on I came to necessity to set also user to for su (it is not always 'oracle', sometimes I need user 'grid' for instance) as well as it is not always 'sysdba', sometimes I need 'sysoper'. It was possible to create two new properties (su_user and as), but I decided to use existing ones. For sqlplus normal connect format user/password is used, for connect AS it is 'sqlplus /as sysoper', so words "as sysoper" taking place of password. And when I decided to use password option, using user option to specify su user was somehow obvious :)

I'm completely not against extra options, just trying to not create to much of them. If you think that extra options are better way, what you think about options `asuser` and `asrole ?`

lib/resources/oracledb_session.rb
voroniys: In article you mentioned they said:
/ as sysdba - connect to LOCAL instance as specified in the environment variables ORACLE_HOME + ORACLE_SID

sys/a@fstest as sysdba - connect OVER A NETWORK to a remote database, using the credentials sys/a

I'm using the first way - connect to the local instance. That is why I need to do su to some user and provide ORACLE_SID in environment (for the case I have more than one). In this case I don't need to provide any sensitive data, like real sysdba credentials. But this is possible only as a particular user (oracle, grid, etc).

lib/utils/database_helpers.rb
voroniys: Yes, it is just an empty glass for the case I don't wanna drink :)
To be serious - if provided column name  does not exists it will cause a ruby error, with if it will result in just an empty array.

lib/resources/oracledb_session.rb
frezbo: Yes I agree, oracle has been always very hard to test. I had to resort to login shell to get the environmental vars sourced. This might fail in the case the `oracle` user has no sudo access and cannot use `su` we might need to handle such conditions too.

frezbo (changes requested):

voroniys:
@frezbo my point is that inspec is normally speaking running either as root or as user with sudo rights and --sudo option for inspec. If it is executed as plain user the check will of course fail. And frankly speaking I don't have good ideas how to cower it. Can you suggest something?

frezbo:
@voroniys you are right, there are open issues with inspec to fail if not running as sudo.

docs/resources/oracledb_session.md.erb
voroniys: I'll answer here on two comments about rows method. I have not added any examples for rows because my usage of it is not really clean inspec. I'm using it in ruby code to build another checks based on SQL query result, i.e. when I need to use data out of (remote) DB for my checks.
For example something like this:
```ruby
sql.query('SELECT * FROM files;').rows.each do |file_row|
  describe file(file_row['path']) do
    its('owner') { should eq file_row['owner']}
  end
end
```
If you think that such unconventional example may be added to documentation I'll do it. 

This is also answer to why rows return plain array of hashes - in construction above hashes are easier to use and offers more possibilities, that array of SQLRows.

docs/resources/oracledb_session.md.erb
clintoncwolfe: I do like adding more options to cover the functionality.  For someone familiar with Oracle, `password: 'as sysdba'` is probably not surprising, but to anyone else, it likely would be.  For example, an auditor might think you are using the string 'as sysdba' as a password, which would be problematic.

With `asuser` and `asrole`, two thoughts: we usually* use underscores to separate words, and it's also not clear that user is refering to the OS user and role to the database.  What about `as_os_user` and `as_db_role`?

docs/resources/oracledb_session.md.erb
clintoncwolfe: I think that's a perfectly valid example.  You might offer it as a second, more advanced example.  As a first example, you might do:
```
describe sql.query('SELECT * FROM files;').rows
  its('count') { should eq 20 }
end
```

lib/utils/database_helpers.rb
clintoncwolfe: I think you might want to fail the resource if the column doesn't exist.  That would almost certainly indicate an error in the query, which the user should know about and fix.  An empty array could be interpreted as a normal empty result.

docs/resources/oracledb_session.md.erb
voroniys: Looked at the code I've found one more reason, why I've used user and password - they are like a mandatory parameters:
```ruby
      return skip_resource "Can't run Oracle checks without authentication" if @user.nil? || @password.nil?
```
This needs to be changed as well than. I'll change the code in favor of 2 new options with names as you suggested.

docs/resources/oracledb_session.md.erb
voroniys: To check the number of rows in result you can do:
```ruby
describe sql.query('SELECT * FROM files;') do
  its('size') { should eq 20 }
end
```
But I surely would agree, that using rows and count make it much more readable.

docs/resources/oracledb_session.md.erb
voroniys: The code change for 2 new options is pushed. I believe all issues are solved now?

lib/resources/oracledb_session.rb
```diff
@@ -34,11 +35,16 @@ def initialize(opts = {})
...
+      return skip_resource "Option 'as_os_user' not available in Windows" if inspec.os.windows? && opts[:as_os_user]
+      @su_user = opts[:as_os_user]
+      @db_role = opts[:as_db_role]
+
       # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
       @sqlcl_bin = 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
       @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'
 
-      return skip_resource "Can't run Oracle checks without authentication" if @user.nil? || @password.nil?
+      return skip_resource "Can't run Oracle checks without authentication" if @su_user.nil? && (@user.nil? || @password.nil?)
```
clintoncwolfe: This should probably be `fail_resource`.  People ignore skip_resource, because they write cross-platform profiles.  But omitting the user and password is likely a control authoring error.

voroniys: This is from original code of the resource, but to my mind you are right, so I'll change it.

voroniys:
Could it be merged now?

voroniys:
Hi,

would you please to give a status update? IMHO all requested changes are done, the failed check is hickup of Travis and has nothing to do with my code. Can it be merged now? Please update the status.

miah (approved):

clintoncwolfe (approved):

PR #3175, opened 2018-06-26
Prevent Slashes in profile names

miah:
When a profile is created with init, the last item after a / is the
profile name. eg "with/slash" would result in a profile created in the
"with" directory named "slash"

Add test for inspec init, and updated other for new output.

Clean up profiles created during testing and place them in temporary
directories.

Fixes chef#1783

### DIFF
--- a/lib/bundles/inspec-init/cli.rb
+++ b/lib/bundles/inspec-init/cli.rb
@@ -1,5 +1,4 @@
 # encoding: utf-8
-# author: Christoph Hartmann
 
 require 'pathname'
 
@@ -42,8 +41,13 @@ module Init
       base_dir = File.join(dir, 'templates', type)
       # prepare glob for all subdirectories and files
       template = File.join(base_dir, '**', '{*,.*}')
-      # generate target path
-      target = Pathname.new(Dir.pwd).join(attributes[:name])
+      # Use the name attribute to define the path to the profile.
+      profile_path = attributes[:name]
+      # Use slashes (\, /) to split up the name into an Array then use the last entry
+      # to reset the name of the profile.
+      attributes[:name] = attributes[:name].split(%r{\\|\/}).last
+      # Generate the full target path on disk
+      target = Pathname.new(Dir.pwd).join(profile_path)
       puts "Create new #{type} at #{mark_text(target)}"
 
       # check that the directory does not exist
--- a/test/functional/inspec_init_test.rb
+++ b/test/functional/inspec_init_test.rb
@@ -1,0 +1,23 @@
+require 'functional/helper'
+require 'fileutils'
+require 'tmpdir'
+require 'yaml'
+
+describe 'inspec init' do
+  include FunctionalHelper
+
+  tmpdir = Dir.tmpdir
+
+  describe 'inspec init profile with/slash' do
+    it 'names profile with string after last slash' do
+      slash_profile = "#{tmpdir}/inspecwith/slash"
+      out = inspec("init profile #{slash_profile}")
+      out.exit_status.must_equal 0
+      File.exist?(slash_profile).must_equal true
+      profile = YAML.load_file("#{slash_profile}/inspec.yml")
+      profile['name'].must_equal 'slash'
+    end
+  end
+
+  Dir.glob("#{tmpdir}/inspecwith*").each {|i| FileUtils.remove_entry_secure(i) }
+end
--- a/test/unit/profiles/profile_test.rb
+++ b/test/unit/profiles/profile_test.rb
@@ -1,6 +1,4 @@
 # encoding: utf-8
-# author: Christoph Hartmann
-# author: Dominik Richter
 
 require 'helper'
 require 'inspec/profile_context'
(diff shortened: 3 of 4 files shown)


test/functional/inspec_init_test.rb
```diff
@@ -0,0 +1,24 @@
...
+require 'tmpdir'
+
+describe 'inspec init' do
+  include FunctionalHelper
+
+  tmpdir = Dir.tmpdir
+  puts tmpdir
+
+  describe 'inspec init profile with/slash' do
+    it 'prevents profile names without valid characters' do
```
clintoncwolfe: This test isn't clear to me.  I thought the intent was to allow slashes in the arg, but only use the last path component.  The test name says we are preventing it, but the this seems to explicitly allow it, checking for exit code 0.

test/functional/inspec_init_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      out = inspec("init profile #{tmpdir}/inspecwith/slash")
+      out.exit_status.must_equal 0
```
clintoncwolfe: We should also check to see where the profile was created (two subdirs down) and what its name is (`slash` only).

clintoncwolfe (changes requested):
A couple of tweaks to the functional test, otherwise looks good!

test/functional/inspec_init_test.rb
miah: Thanks I'll resolve these!

miah:
I've made some updates to the test. Let me know what you think @clintoncwolfe

clintoncwolfe (approved):
Many thanks @miah, looks good!

lib/bundles/inspec-init/cli.rb
```diff
@@ -42,8 +41,13 @@ def generator(type, attributes = {}, options = {}) # rubocop:disable Metrics/Abc
       base_dir = File.join(dir, 'templates', type)
       # prepare glob for all subdirectories and files
       template = File.join(base_dir, '**', '{*,.*}')
-      # generate target path
-      target = Pathname.new(Dir.pwd).join(attributes[:name])
+      # Use the name attribute define the path to the profile.
```
jerryaldrichiii: Could we change to `to define the path`?

lib/bundles/inspec-init/cli.rb
```diff
@@ -42,8 +41,13 @@ def generator(type, attributes = {}, options = {}) # rubocop:disable Metrics/Abc
...
+      profile_path = attributes[:name]
+      # Use slashes (\, /) to split up the name into an Array then uses the last entry
```
jerryaldrichiii: `then use`?

test/functional/inspec_init_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'functional/helper'
+require 'fileutils'
+require 'tmpdir'
+require 'yaml'
+
+describe 'inspec init' do
+  include FunctionalHelper
+
+  tmpdir = Dir.tmpdir
+  puts tmpdir
```
jerryaldrichiii: Do we want this puts here?

test/functional/inspec_init_test.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  describe 'inspec init profile with/slash' do
+    it 'Names profile with string after last slash' do
```
jerryaldrichiii: Usually we use lower case here i.e: `it 'names....` but I won't lose sleep if you don't change it :smile:

test/functional/inspec_init_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      slash_profile = "#{tmpdir}/inspecwith/slash"
+      out = inspec("init profile #{slash_profile}")
+      out.exit_status.must_equal 0
+      File.exist?(slash_profile).must_equal true
+      profile = YAML.load_file("#{slash_profile}/inspec.yml")
+      profile["name"].must_equal 'slash'
```
jerryaldrichiii: Single quote since no interpolation?

NOTE: rubcop doesn't cover tests so...I can't fault you if you leave it.

test/unit/profiles/profile_test.rb
```diff
@@ -337,19 +335,20 @@
     describe 'a profile with a slash in the name' do
       let(:profile_path) { 'slash-in-name/not-allowed' } # Slashes allowed here
       let(:profile_name) { 'slash-in-name/not-allowed' }   # But not here
-      it 'issues a deprecation warning' do
+      it 'issues a error' do
```
jerryaldrichiii: `an error`?

jerryaldrichiii (changes requested):
Great work @miah only a few style/grammatical nit picks.

miah:
Thanks Jerry, will resolve 😄

miah:
Thanks @jerryaldrichiii these should be cleaned up now =)

jerryaldrichiii (approved):

larryebaum:
@miah Any idea why this is still an issue? This old issue maps back to this and was raised, I suspect, with planning for 2018 InSpec workshop for ChefConf.. I'm prepping for the 2019 workshop and it seems the same issue exists...
https://github.com/chef/inspec/issues/2856

larryebaum:
I think nevermind/disregard above... will revise if needed...

Issue #1825, opened 2017-05-17
Unique exit code for control/profile failures


adamleff:
I'd like to be able to differentiate between a "failed InSpec run" (where InSpec failed to execute profiles due to a Ruby issue, SSL issue fetching a profile, etc.) and a "successful InSpec run that encountered a profile/control failure". This will allow tools like the habitat profile creator to present more useful error messages to the user without parsing JSON.

Proposal:
 * 0: successful run, no control failures
 * 100: successful run, critical failures found
 * 101: successful run, major failures found
 * 102: successful run, minor failures found
 * anything else: unsuccessful run

elliott-davis:
![tenor-91893169](https://cloud.githubusercontent.com/assets/1130349/26169099/1b24d790-3b02-11e7-9b2a-e7f9756e5ea9.gif)

chris-rock:
That is a great idea @adamleff

trevor-vaughan:
Just ran into this when running automated tests, very much a :+1:

adamleff:
NOTE TO IMPLEMENTER: since this is a breaking change, the PR should target the `release-2.0` branch, NOT `master`

arlimus:
great stuff!!

adamleff:
This should be ready to implement once @jquick completes #2415

chris-rock:
The exit codes we settled on are:

    0 run ok + all passed:
    100 run okay + failures
    101 run okay + skipped only

PR #3178, opened 2018-06-28
Document exit codes for 'inspec exec' and add --no-distinct-exit option

### LINKED ISSUE #2896: Skipped tests shouldn't result in a non-zero exit value since it breaks logic tests
### 🎛 Description

🐛 bug report

When a control is skipped due to an only_if condition it returns a 101 exit code. While this is planned behavior due to #1825, it makes it difficult to do simple boolean logic such as `inspec exec || echo "Failed"`.

In my use case, I'm using InSpec as part of a Packer build and it's causing all builds to fail since certain controls only are valid per platform (i.e. Docker vs AWS AMI)

```
$ inspec --version
2.1.10
```

Running on OSX against an Amazon Linux target

### 🤔 Replication Case

Run the https://github.com/dev-sec/linux-baseline InSpec profile against a docker target and check the return code. It will be 101 (i.e. non-zero, breaking boolean comparisons) 

### 💁 Possible Solutions

Revert the 101 exit code in https://github.com/chef/inspec/blob/master/lib/inspec/runner_rspec.rb#L91-L92

### LINKED ISSUE #3167: Inspec documentation missing exit code status
### 🎛 Description

**Choose one:**  🐛 bug report

Inspec documentation missing exit code status as per this: https://github.com/inspec/inspec/issues/1825 ?

### 💁 Possible Solutions

PR's to README

clintoncwolfe:
Fixes #2896 
Fixes #3167 
Refs #1825 

* Adds a 'long description' to the Thor configuration for the `inspec exec` command, which includes the exit code behavior.  This makes the CLI usage info include it.  
* Alters the website CLI doc generator to use the long description if present.
* Adds a `--distinct-exit` CLI option to `exec`, enabled by default.  If disabled, changes behavior to exit code 1 on test or any other failure, exit 0 on normal or skip-only runs.

To test the website generator, `cd www && bundle update && bundle rake docs:cli`
Excerpt from generated website docs:

---

## exec

Run all test files at the specified path.

Loads the given profile(s) and fetches their dependencies if needed.  Then connects to the target and executes any controls contained in the profiles.  One or more reporters are used to generate output.  If all tests passed (no fails, no skips) exit code 0 is returned.  If some tests skipped but none failed, exit code 101 is returned. If at least one test failed, exit code 100 is returned.  If inspec failed for any other reason, exit code 1 is returned.

### Syntax

This subcommand has the following syntax:

---

Excerpt from the CLI usage, from `inspec help exec`:

---
...
  l, [--log-level=LOG_LEVEL]                       # Set the log level: info (default), debug, warn, error
      [--log-location=LOG_LOCATION]                # Location to send diagnostic log messages to. (default: STDOUT or STDERR)
      [--diagnose], [--no-diagnose]                # Show diagnostics (versions, configurations)

Description:
  Loads the given profile(s) and fetches their dependencies if needed. Then connects to the target and executes
  any controls contained in the profiles. One or more reporters are used to generate output. If all tests
  passed (no fails, no skips) exit code 0 is returned. If some tests skipped but none failed, exit code 101 is
  returned. If at least one test failed, exit code 100 is returned. If inspec failed for any other reason, exit
  code 1 is returned.

---

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -83,6 +83,8 @@ module Inspec
         desc: 'Allow caching for backend command output. (default: true)'
       option :show_progress, type: :boolean,
         desc: 'Show progress while executing tests.'
+      option :distinct_exit, type: :boolean, default: true,
+        desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
     end
 
     def self.default_options
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -157,7 +157,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
 
   desc 'exec PATHS', 'run all test files at the specified PATH.'
   long_desc <<~EOT
-    Loads the given profile(s) and fetches their dependencies if needed.  Then connects to the target and executes any controls contained in the profiles.  One or more reporters are used to generate output.  If all tests passed (no fails, no skips) exit code 0 is returned.  If some tests skipped but none failed, exit code 100 is returned. If at least one test failed, exit code 101 is returned.  If inspec failed for any other reason, exit code 1 is returned.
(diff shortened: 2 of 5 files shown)


frezbo:
Thanks @clintoncwolfe for the quick turnaround. Awesome!!!

jquick (approved):
Thanks @clintoncwolfe !

Issue #3128, opened 2018-06-12
RFC: Consider defining a "maturity path" and stability metadata field


clintoncwolfe:
### 🎛 Description

🙋 feature request

We'd like to be able to express the stability of individual resources (and ideally, individual features) in a robust way.  This was [proposed before](https://github.com/inspec/inspec/blob/47c6427082d5622b77bb0769229cd8fab9dec62c/docs/README.md#stability-index) in the earlier days of InSpec.  With cloud platform support, we see a tension between wanting to offer stable, secure functionality; and using API introspection to rapidly gain broad coverage of cloud platforms.  This has come up a few times; most recently on the [inspec-gcp project issue 25](https://github.com/inspec/inspec-gcp/issues/25).

One middle ground might be to allow resources to declare their level of stability.

* **experimental**  Not to be relied on.  Missing docs and or tests.  May use runtime introspection of remote APIs to increase feature coverage, which makes them inherently unstable. May be listed on the website but otherwise undocumented.  Must read source code to understand.
* **beta** Not to be relied on but improving.  May use build-time API introspection.  Incomplete docs but getting there.  Some tests.
* **stable** Reliable.  Fully documented and tested.  Future work focuses on maintaining compatibility while expanding functionality.
* **locked** Reliable but closed to new features.  Accepting bugfixes.
* **deprecated** Will be removed in next major release.  Do not use.

These are just example names.  Other names or criteria are welcome to be defined.

Additional behavior:
* warn if experimental resources used.  Add CLI option to suppress warning.
* warn if deprecated resources used.  Add CLI option to suppress warning.
* Make `inspec check` aware of stability.
* docs / examples / toolkit for making experimental-level resources for cloud platforms

### 🌍 InSpec and Platform Version

2.2.10

Attn: @arlimus @jquick @chris-rock @trickyearlobe

Aha! Link: https://chef.aha.io/features/SH-2321

chris-rock:
I really like the idea of adding the stability index to resources. I think we also need to define a clear path for resource to get into inspec core. e.g. only stable and locked resources should be in core, all other resources should be in resource packs. That allows us to balance velocity/features with stability.

clintoncwolfe:
I think there is a place in inspec core for experimental resources / features, especially if they are disabled by default or trigger warnings when used.  I see several motivations for this:
[...]

PR #3184, opened 2018-06-29
Ability to render contrib resource pack docs on main website

clintoncwolfe:
We've having a lot of discussions about how to best handle resource packs lately.  We want innovation to happen, but to keep the core of inspec stable; one solution to this is encourage the use of [resource packs](https://www.inspec.io/docs/reference/glossary/#resource-pack) as first-class citizens.

One major pain point around that is publicity.  If we at Chef place innovative functionality in a resource pack, how can people find it?

A long term solution might involve a public service like Supermarket.  A short term solution might be to render the docs of the resource packs (selectively) on the main inspec.io website.

This PR:

* Introduces a YAML config file at `contrib/contrib.yaml` which lists git locations of resource packs to fetch during doc build time
* Adds Rake tasks, which can fetch the resource packs, and copy out the docs to core.  _This functionality may also be useful in future work easing the testing cycle for resource packs._
* Overhauls the Rake code that generates the [Resource Overview docs page](https://www.inspec.io/docs/reference/resources/), to allow it to list resources in a more generic way.

Refs #3128 
Attn: @dmccown @davymcaleer @chris-rock 

Here are a few screenshots, with the REST-based Azure resource pack added as a contrib.  Note that the existing Azure resources and the the ones in the resource pack are co-habitating nicely.

Presumably we'd have PRS on the contrib resource pack to add a note to the docs indicating installation instructions (that is, a profile dependency).

screen shot 2018-06-28 at 19 52 23
screen shot 2018-06-28 at 19 52 43
screen shot 2018-06-28 at 20 01 40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ terraform.tfstate*
 terraform.tfstate.backup
 inspec-azure.plan
 inspec-aws-*.plan
+contrib/*
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ group :tools do
   gem 'pry', '~> 0.10'
   gem 'rb-readline'
   gem 'license_finder'
+  gem 'git', '~> 1.4'
 end
 
 # gems for Maintainers.md generation
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,6 @@ require 'passgen'
 require 'train'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
-require_relative 'tasks/contrib'
 
 Bundler::GemHelper.install_tasks name: 'inspec'
 
@@ -26,6 +25,13 @@ begin
   require_relative 'tasks/docs'
 rescue LoadError
   puts 'docs tasks are unavailable because the ruby-progressbar gem is not available.'
+end
+
+begin
+  require 'git'
+  require_relative 'tasks/contrib'
+rescue LoadError
+  puts 'contrib tasks are unavailable because the git gem is not available.'
 end
 
 # Rubocop
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,7 @@ require 'passgen'
 require 'train'
 require_relative 'tasks/maintainers'
 require_relative 'tasks/spdx'
+require_relative 'tasks/contrib'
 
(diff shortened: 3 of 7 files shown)


chris-rock:
This is a great improvement @clintoncwolfe and brings our resource packs into the core docs! Do we have any requirements on the resource packs that we should bring in those as soon as possible? How do we make clear to the user that he needs to change the inspec.yml a bit? This discussion should not block this PR, I just try to think about all the stuff that we need for a great user experience.

clintoncwolfe:
Regarding which resource packs to include, the main inspec project controls that through the use of a YAML file.  So, we can choose to include none, or certain ones we select.  I'm not proposing any policy rules or conditions for including resource packs; that's something we should clearly discuss.

As for notes about installing via inspec.yml, my thought was to open a PR on each of the resource packs, adding a section to each resource's documentation, such as:

```
## Availability

### Installation

This resource is available in the `inspec-foo` resource pack.  To use it, add the following to your `inspec.yml` in your top-level profile:

\```
depends:
  inspec-foo:
    git: https://github.com/some-project/inspec-foo.git
\```

### Version

This resource first became available in v1.2.3 of the inspec-foo resource pack.
```

clintoncwolfe:
Latest changeset accepts doc files that end in `.md` as well as `.md.erb`.  Refs https://github.com/inspec/inspec-gcp/pull/35

screen shot 2018-07-05 at 21 49 22
screen shot 2018-07-05 at 21 49 36

clintoncwolfe:
Additionally, the 'Azure Resources' button now wraps if we have more than 3 sections.  Modified renderer to omit the word 'resources' unless it is the OS group.

screen shot 2018-07-05 at 21 48 32

clintoncwolfe:
I've opened https://github.com/inspec/inspec-azure/pull/73 to add an Availability section to the Azure docs.  Here's what it looks like, rendered.
screen shot 2018-07-05 at 22 36 18

jquick (approved):
Very nice @clintoncwolfe.

miah (approved):
Awesome work =)

jerryaldrichiii (approved):
Looks good to me! Thanks @clintoncwolfe!

PR #3186, opened 2018-06-29
Update the node platform issues to warn severity

jquick:

These issues will not block a run currently so its confusing to the user to show `ERROR` in the output. Updating these to `WARN` calls instead.


clintoncwolfe (approved):
Short and sweet - thanks @jquick!

PR #3195, opened 2018-07-03
cli: Add `--insecure` option for `exec` and `shell`

jerryaldrichiii:
This is used for the `vmware://` transport but could be used for others in the future.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -13,7 +13,7 @@ module Inspec
       true
     end
 
-    def self.target_options
+    def self.target_options # rubocop:disable MethodLength
       option :target, aliases: :t, type: :string,
         desc: 'Simple targeting option using URIs, e.g. ssh://user:pass@host:port'
       option :backend, aliases: :b, type: :string,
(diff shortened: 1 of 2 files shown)


miah (approved):

jerryaldrichiii:
@miah, added `shell` as well. Let me know if you still approve :smile:

lib/inspec/base_cli.rb
```diff
@@ -85,6 +85,8 @@ def self.exec_options
         desc: 'Show progress while executing tests.'
       option :distinct_exit, type: :boolean, default: true,
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
+      option :insecure, type: :boolean, default: false,
```
jquick: Do you think this would be better in the "target_options" above instead? I think that will lay out better in the help menu as this only pertains to the target.

lib/inspec/cli.rb
```diff
@@ -209,6 +209,8 @@ def detect
     desc: 'A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell'
   option :distinct_exit, type: :boolean, default: true,
     desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
+  option :insecure, type: :boolean, default: false,
```
jquick: You can remove this then, as shell calls target_options a little above this.

jquick (commented):
Pretty nice @jerryaldrichiii. Some small suggestions.

jerryaldrichiii:
Thanks @jquick, I didn't see target options! Should be good now.

jquick (approved):
Thanks @jerryaldrichiii

PR #3198, opened 2018-07-05
Fix vendor functional test to not validate a repo hash that can change.

miah:
This functional test was validating that a randomly named directory 
created by mktemp existed. We are now checking that _a_ directory exists
where it should but not validating the name. We are also now checking
that the created directory has around the right number of files.


jquick (approved):
Thanks @miah

PR #3200, opened 2018-07-09
Add habitat plan for windows

TheLunaticScripter:

### DIFF
--- a/habitat/plan.ps1
+++ b/habitat/plan.ps1
@@ -1,0 +1,31 @@
+$pkg_name="inspec"
+$pkg_origin="chef"
+$pkg_version=$(Get-Content "$PLAN_CONTEXT/../VERSION")
+$pkg_description="InSpec is an open-source testing framework for infrastructure
+  with a human- and machine-readable language for specifying compliance,
+  security and policy requirements."
+$pkg_upstream_url="https://www.inspec.io/"
+$pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+$pkg_license=('Apache-2.0')
+$pkg_source="https://packages.chef.io/files/stable/inspec/${pkg_version}/windows/2016/inspec-${pkg_version}-1-x64.msi"
+$pkg_shasum="fcdf17b29bc94f2eff1ed59fbb8d224f05df6552e0ff85e1eb71fb4e3b7db24d"
+$pkg_build_deps=@("core/lessmsi")
+$pkg_bin_dirs=@(
+    "inspec/bin",
+    "inspec/embedded/bin"
+)
+
+function Invoke-Unpack {
+  lessmsi x (Resolve-Path "$HAB_CACHE_SRC_PATH/$pkg_filename").Path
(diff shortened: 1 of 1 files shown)


clintoncwolfe:
Not opposed to this, just want to verify that someone needs this?  I'm curious about your use case.

TheLunaticScripter:
Yep so we are working on running InSpec profiles to report to Automate on windows using InSpec as a dependency. I had added this into the Habitat core-plans but was redirected to add it here instead.

habitat/plan.ps1
```diff
@@ -0,0 +1,31 @@
...
+$pkg_origin="chef"
+$pkg_version=$(Get-Content "$PLAN_CONTEXT/../VERSION")
+$pkg_description="InSpec is an open-source testing framework for infrastructure
+  with a human- and machine-readable language for specifying compliance,
+  security and policy requirements."
+$pkg_upstream_url="https://www.inspec.io/"
+$pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+$pkg_license=('Apache-2.0')
+$pkg_source="https://packages.chef.io/files/stable/inspec/${pkg_version}/windows/2016/inspec-${pkg_version}-1-x64.msi"
+$pkg_shasum="fcdf17b29bc94f2eff1ed59fbb8d224f05df6552e0ff85e1eb71fb4e3b7db24d"
```
clintoncwolfe, line 11: If the package source is parameterized with the version, the shasum should be as well - this value will only be right once (we hope!)

clintoncwolfe (changes requested):
I think we need to either parameterize or remove the shasum.  Otherwise seems straightforward - just repacking an MSI.  Thanks!

habitat/plan.ps1
TheLunaticScripter: Yep investigating how to do this properly. I think it needs to be put in a file someplace after the omnibus builds the msi just not sure yet how to do that.

jquick:
Hey @TheLunaticScripter - Looks like this has gone a bit stale. Should we close this for now until you have time to vet out the options here?

habitat/plan.ps1
```diff
@@ -0,0 +1,31 @@
+$pkg_name="inspec"
+$pkg_origin="chef"
+$pkg_version=$(Get-Content "$PLAN_CONTEXT/../VERSION")
+$pkg_description="InSpec is an open-source testing framework for infrastructure
+  with a human- and machine-readable language for specifying compliance,
+  security and policy requirements."
+$pkg_upstream_url="https://www.inspec.io/"
+$pkg_maintainer="The Habitat Maintainers <humans@habitat.sh>"
+$pkg_license=('Apache-2.0')
+$pkg_source="https://packages.chef.io/files/stable/inspec/${pkg_version}/windows/2016/inspec-${pkg_version}-1-x64.msi"
```
jquick, line 10: Lets move this to unstable and we can bump it to stable during our release process. We will build a unstable version along with the linux binaries on every PR merge.

jquick:
Hey @TheLunaticScripter - Were you able to figure out anything on this?

TheLunaticScripter:
Still working on if it is possible to populate the checksum in `Invoke-Before`.

jquick:
Closing this for now. Please re-open when you have time/information to go forward.

PR #3202, opened 2018-07-09
Refactor 'inspec init profile' into a reusable component.

mattray:
base_cli.rb had several methods used internally, these are exposed so
lib/bundles/inspec-init/profile.rb can act as a library for anything
that needs to create new Inspec profiles programatically (ie. inspec-iggy)

Additional PRs will add CLI options similar to "chef generate cookbook"

### DIFF
--- a/lib/bundles/inspec-init/cli.rb
+++ b/lib/bundles/inspec-init/cli.rb
@@ -15,7 +15,7 @@ module Init
       namespace
     end
 
-    # read template directoy
+    # read template directory
     template_dir = File.join(File.dirname(__FILE__), 'templates')
     Dir.glob(File.join(template_dir, '*')) do |template|
       relative = Pathname.new(template).relative_path_from(Pathname.new(template_dir))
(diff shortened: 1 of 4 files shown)


lib/bundles/inspec-init/profile.rb
```diff
@@ -0,0 +1,69 @@
+module Init
+  class Profile
```
clintoncwolfe, line 2: I'd suggest the name `ProfileGenerator`, as it is more in line with the meaning.  Or `ProfileTemplater`.

lib/bundles/inspec-init/cli.rb
```diff
@@ -15,85 +15,19 @@ def self.subcommand_prefix
...
       relative = Pathname.new(template).relative_path_from(Pathname.new(template_dir))
 
       # register command for the template
       desc "#{relative} NAME", "Create a new #{relative}"
       option :overwrite, type: :boolean, default: false,
-         desc: 'Overwrites existing directory'
+        desc: 'Overwrites existing directory'
       define_method relative.to_s.to_sym do |name|
-        generator(relative.to_s, { name: name }, options)
+        Init::Profile.generator(relative.to_s, { name: name }, options)
```
clintoncwolfe, line 28: I'm not sure this make sense to become a class method

lib/inspec/base_cli.rb
```diff
@@ -364,15 +364,15 @@ def configure_logger(o)
       o[:logger].level = get_log_level(o.log_level)
     end
 
-    def mark_text(text)
+    def self.mark_text(text)
```
clintoncwolfe, line 367: OK, I think what you want is "I want my base CLI class to help me with formatting".  Make these public, instance methods, not "attempted private" class methods. 

Then, have ProfileGenerator inherit from here, or we make a real UI object.

clintoncwolfe (changes requested):
This is a great idea.  Some coding issues we need to address, though.  First, we need the rubocop issues fixed; then the change in scope of the formatters needs to be reverted, or many functional tests will break.  I may adopt this if needed.

aaronlippold:
I think this will be a very valuable addition small though it is:-)

clintoncwolfe:
Adopted on #3214

PR #3207, opened 2018-07-10
command resource: Allow redacting `#to_s`

### LINKED ISSUE #3185: mysql_session password sensitivity
### 🎛 Description🙋 feature request

Is it possible to make it so that the mysql_session password is redacted when the name of the test is displayed?

```
✔  Command mysql -uroot -p<NOT_REDACTED> -h localhost  -s -e "show databases" exit_status should eq 0
```

jerryaldrichiii:
This allows the user to redact command arguments when using the `command` resource.

This also changes how the `#to_s` is rendered (adds a `:` and some backticks) let me know if that is too drastic of a change.

Examples:

```
# With capture groups
options = { redact_command: /(-p ).*( -d)/ }
describe command('myapp -p secretpassword -d no_redact', options) do
end

# Result (capture groups used)
Command: `myapp -p REDACTED -d no_redact`
          ✔  exit_status should cmp == 0

# Without capture groups
options = { redact_command: /-p .* -d/ }
describe command('myapp -p secretpassword -d no_redact', options) do
end

# Result (no capture groups used)
Command: `myapp REDACTED no_redact`
          ✔  exit_status should cmp == 0
```

This fixes https://github.com/inspec/inspec/issues/3185

### DIFF
--- a/docs/resources/command.md.erb
+++ b/docs/resources/command.md.erb
@@ -124,19 +124,25 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
       end
     end
 
-<br>
-
 ### Redacting Sensitive Commands
 
-By default the command that is ran is shown in the InSpec output. This can be problematic if the command contains sensitive arguments such as a password.
+By default the command that is ran is shown in the InSpec output. This can be problematic if the command contains sensitive arguments such as a password. These sensitive parts can be redacted by passing in `redact_regex` and a regular expression to redact. Optionally, you can use 2 capture groups to fine tune what is redacted.
 
-These sensitive parts can be redacted by passing in `redact_command` and a regular expression to redact. Optionally, you can use 2 capture groups to fine tune what is redacted.
+The following examples show how to use `redact_regex`:
 
-The following examples show how to use `redact_command` with the command resource:
+    # Example without capture groups
+    describe command('myapp -p secretpassword -d no_redact', redact_regex: /-p .* -d/) do
+      its('exit_status') { should cmp 0 }
+    end
 
-    # With capture groups
-    options = { redact_command: /(-p ).*( -d)/ }
-    describe command('myapp -p secretpassword -d no_redact', options) do
+    # Result (no capture groups used)
+    Command: `myapp REDACTED no_redact`
+       ✔  exit_status should cmp == 0
+
+    # Example with capture groups
+    # Each set of parenthesis is a capture group.
+    # Anything in the two capture groups will not be 'REDACTED'
+    describe command('myapp -p secretpassword -d no_redact', redact_regex: /(-p ).*( -d)/) do
       its('exit_status') { should cmp 0 }
     end
 
@@ -144,15 +150,9 @@ The following examples show how to use `redact_command` with the command resourc
     Command: `myapp -p REDACTED -d no_redact`
        ✔  exit_status should cmp == 0
 
-    # Without capture groups
-    options = { redact_command: /-p .* -d/ }
-    describe command('myapp -p secretpassword -d no_redact', options) do
-      its('exit_status') { should cmp 0 }
-    end
+> For more info/help on regular expressions, we recommend [RegExr](https://regexr.com/)
 
-    # Result (no capture groups used)
-    Command: `myapp REDACTED no_redact`
-       ✔  exit_status should cmp == 0
+<br>
 
 ## Matchers
 
--- a/docs/resources/command.md.erb
+++ b/docs/resources/command.md.erb
@@ -126,6 +126,34 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
 
 <br>
 
+### Redacting Sensitive Commands
+
+By default the command that is ran is shown in the InSpec output. This can be problematic if the command contains sensitive arguments such as a password.
+
+These sensitive parts can be redacted by passing in `redact_command` and a regular expression to redact. Optionally, you can use 2 capture groups to fine tune what is redacted.
+
+The following examples show how to use `redact_command` with the command resource:
+
+    # With capture groups
+    options = { redact_command: /(-p ).*( -d)/ }
+    describe command('myapp -p secretpassword -d no_redact', options) do
+      its('exit_status') { should cmp 0 }
+    end
+
+    # Result (capture groups used)
+    Command: `myapp -p REDACTED -d no_redact`
+       ✔  exit_status should cmp == 0
+
+    # Without capture groups
+    options = { redact_command: /-p .* -d/ }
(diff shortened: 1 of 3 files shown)


clintoncwolfe:
Close/reopen to trigger re-run of travis job (another APT ephemeral failure)

docs/resources/command.md.erb
```diff
@@ -126,6 +126,34 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
 
 <br>
 
+### Redacting Sensitive Commands
+
+By default the command that is ran is shown in the InSpec output. This can be problematic if the command contains sensitive arguments such as a password.
+
+These sensitive parts can be redacted by passing in `redact_command` and a regular expression to redact. Optionally, you can use 2 capture groups to fine tune what is redacted.
+
+The following examples show how to use `redact_command` with the command resource:
```
clintoncwolfe: You might want to omit the 'with the command resource' - it might be confusing (anything you do here is with the command resource).

docs/resources/command.md.erb
```diff
@@ -126,6 +126,34 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
...
+
+    # With capture groups
+    options = { redact_command: /(-p ).*( -d)/ }
```
clintoncwolfe: I might suggest not putting the options in a separate variable.  It's no big deal for ruby programmers, but non-developers might see that as part of the syntax.  You might inline it just to keep it clearer.

docs/resources/command.md.erb
```diff
@@ -126,6 +126,34 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
...
+    Command: `myapp -p REDACTED -d no_redact`
+       ✔  exit_status should cmp == 0
+
+    # Without capture groups
+    options = { redact_command: /-p .* -d/ }
+    describe command('myapp -p secretpassword -d no_redact', options) do
+      its('exit_status') { should cmp 0 }
+    end
+
+    # Result (no capture groups used)
```
clintoncwolfe: If possible you might show the simpler approach (no capture groups) first.  
If you feel that the no-capture approach is ugly or less  useful, we don't need to document it - we could present with-capture-groups as the only way.

docs/resources/command.md.erb
```diff
@@ -126,6 +126,34 @@ Wix includes several tools -- such as `candle` (preprocesses and compiles source
 
 <br>
 
+### Redacting Sensitive Commands
+
+By default the command that is ran is shown in the InSpec output. This can be problematic if the command contains sensitive arguments such as a password.
+
+These sensitive parts can be redacted by passing in `redact_command` and a regular expression to redact. Optionally, you can use 2 capture groups to fine tune what is redacted.
```
clintoncwolfe: For my $0.02, I think a short option name (`redact`, `redact_re`, `redact_regex`, `redact_output`) might be preferable.  You're already working with a `command` resource; that as a suffix might not add much information.

clintoncwolfe: I think asking people to know what regexes are is one big step, and asking them to properly use capture groups is another big step - especially when they don't see how the implementation is using the capture groups.

We could link to a regexp playground app like [regexr](https://regexr.com).

There are several other ways to handle this - thoughts?

test/unit/resources/command_test.rb
```diff
@@ -35,4 +35,22 @@ def resource(y)
   it 'raises when called with nil as a command' do
     proc { resource(nil).result }.must_raise StandardError
   end
+
+  it 'fails the resource if `redact_command` is not a regular expression' do
+    result = resource('env', redact_command: 'string')
+    result.resource_failed?.must_equal true
+    result.resource_exception_message.must_match /must be a regular expression/
```
clintoncwolfe: I see you like regexes, so I put the string 'regular expression' in your regular expression

clintoncwolfe (changes requested):
I'd like to see the param name adjusted a bit (which I saw you asked for feedback on that), and maybe a bit more refining on the docs.  No serious complaints, easy to flip this to approved, just a bit of discussion.

docs/resources/command.md.erb
jerryaldrichiii: Yup, good call. Fixing.

docs/resources/command.md.erb
jerryaldrichiii: Hmmm...it was getting a bit long, but I agree. I'd rather have length over complexity.

docs/resources/command.md.erb
jerryaldrichiii: It's a bit ugly, but not dangerously so. I'll show it first since it is simpler.

docs/resources/command.md.erb
jerryaldrichiii: I wanted something to show we are just redacting the command that is printed. As opposed to stdout/stderr. You're right though. I think I'll go with `redact_regex` and rely on documentation.

jerryaldrichiii: I thought the same, but didn't think about linking regexr. I think that is a solid idea. Will add.

docs/resources/command.md.erb
jerryaldrichiii: Well...it renders really bad when I do it that way due to the line wrapping...I'm torn.

clintoncwolfe: Also, using `.*` is a bit dangerous - IIRC, star is greedy, so it will keep matching until the next atom matches. So if your redact is `/-p .*-d/` and you have `-p what --debug -d`, that will match out to `-p REDACTED -d`

I think you might want the minimal version, `.*?`.

clintoncwolfe (changes requested):

clintoncwolfe (approved):
thanks @jerryaldrichiii !

jquick (approved):
Very nice! Thanks @jerryaldrichiii

kplimack:
Thanks @jerryaldrichiii

kplimack:
@jerryaldrichiii any chance of backporting this to inspec 1.x and releasing a gem?

unfortunately im stuck with an old version of chef that i cannot upgrade due to bugs/incompatibility with newer versions chef 

```
>>>> Caused by Gem::ConflictError: Unable to activate google-api-client-0.19.8, because addressable-2.4.0 conflicts with addressable (>= 2.5.1, ~> 2.5)
/opt/chef/embedded/lib/ruby/site_ruby/2.3.0/rubygems/specification.rb:2290:in `raise_if_conflicts'
/opt/chef/embedded/lib/ruby/site_ruby/2.3.0/rubygems/specification.rb:1410:in `activate'
```

PR #3213, opened 2018-07-13
Add new resource: aws_ecs_cluster

meringu:
Hi! I want to be able to test the workloads on my ECS clusters, I want to check they are doing things that we require, such as setting memory and CPU limits, and not using host networking etc...

To do this I would like to implement some new resources, and their plurals:
- aws_ecs_cluster
- aws_ecs_task
- aws_ecs_task_definition

I thought I'd start with the simplest one first, as this is my first inspec resource. If this gets accepted, I'd be happy to start work on the other resources.

I copied most of this from the `aws_elb` resource as my starting point.

Let me know if there is anything else I need to add. I wasn't sure if I should do anything about documentation.

Thanks and kind regards,

Henry


jquick (approved):
Thanks so much for this @meringu! This looks great and a nice addition.

chris-rock (approved):
Great addition @meringu

PR #3214, opened 2018-07-13
Refactor 'inspec init profile' into a reusable component.

clintoncwolfe:
This PR is an adoption of #3202 from @mattray.

Refactors Init::Profile into a reusable class named Init::Renderer, and alters its interface from being class-based to instance-based.

Exposes several output formatting methods on BaseCLI for use by subclasses.

Variable name clarification pass on Init::Renderer.

---
base_cli.rb had several methods used internally, these are exposed so
lib/bundles/inspec-init/profile.rb can act as a library for anything
that needs to create new Inspec profiles programatically (ie. inspec-iggy)

Additional PRs will add CLI options similar to "chef generate cookbook"

### DIFF
--- a/lib/bundles/inspec-init.rb
+++ b/lib/bundles/inspec-init.rb
@@ -5,4 +5,8 @@
 libdir = File.dirname(__FILE__)
 $LOAD_PATH.unshift(libdir) unless $LOAD_PATH.include?(libdir)
 
+module Init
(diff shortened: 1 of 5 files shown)


jquick (approved):
Nice refactor! Thanks @clintoncwolfe

jerryaldrichiii (approved):
Great work @clintoncwolfe!

PR #3215, opened 2018-07-16
Add Alpine package provider

damacus:
Adds Alpine package provider.

- [x] Add Alpine tests
- [x] Stub apk grep command for alpine
- [x] Resolve (disable for now) rubocop ABC/CyclomaticComplexity/PerceivedComplexity

### DIFF
--- a/lib/resources/package.rb
+++ b/lib/resources/package.rb
@@ -19,8 +19,10 @@ module Inspec::Resources
         its('version') { should eq 1.9.5 }
       end
     "
-
-    def initialize(package_name, opts = {}) # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def initialize(package_name, opts = {})
       @package_name = package_name
       @name = @package_name
       @cache = nil
@@ -44,12 +46,17 @@ module Inspec::Resources
         @pkgman = SolarisPkg.new(inspec)
       elsif ['hpux'].include?(os[:family])
         @pkgman = HpuxPkg.new(inspec)
+      elsif ['alpine'].include?(os[:name])
+        @pkgman = AlpinePkg.new(inspec)
       else
         raise Inspec::Exceptions::ResourceSkipped, 'The `package` resource is not supported on your OS yet.'
       end
 
       evaluate_missing_requirements
     end
+    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
 
     # returns true if the package is installed
     def installed?(_provider = nil, _version = nil)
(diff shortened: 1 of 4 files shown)


lib/resources/package.rb
```diff
@@ -44,6 +44,8 @@ def initialize(package_name, opts = {}) # rubocop:disable Metrics/AbcSize
         @pkgman = SolarisPkg.new(inspec)
       elsif ['hpux'].include?(os[:family])
         @pkgman = HpuxPkg.new(inspec)
+      elsif os.alpine?
```
chris-rock: I would use `['alpine'].include?(os[:name])`, we do not need a new helper for alpine

docs/resources/os.md.erb
```diff
@@ -70,6 +70,7 @@ The `os` audit resource includes a collection of helpers that enable more granul
 * `suse?`
 * `unix?`
 * `windows?`
+* `alpine?`
```
chris-rock: I think we do not need alpine here

test/unit/resources/package_test.rb
```diff
@@ -152,6 +152,15 @@
...
 
+  # alpine
+  it 'verify Alpine package processing' do
+    resource = MockLoader.new(:alpine).load_resource('package', 'git')
+    pkg = { name: 'git', installed: true, version: '2.15.0-r1', type: 'pkg' }
+    _(resource.installed?).must_equal true
+    _(resource.version).must_equal '2.15.0-r1'
+    _(resource.info).must_equal pkg
+  end
+
```
chris-rock, line 163: Great work, you need to add the command output for `apk info -vv --no-network | grep git` in https://github.com/inspec/inspec/tree/master/test/unit/mock/cmd and the command registration similar to https://github.com/inspec/inspec/blob/master/test/helper.rb#L244

chris-rock (dismissed):
Great work @damacus I added some comments

test/unit/resources/package_test.rb
damacus: Thank you, I was wondering how/where to stub and mock that data.

damacus:
Looks like the functional test suite is failing 🤔

damacus:
@chris-rock I think this is ready to merge now :+1:

jquick (approved):
Looks good! Thanks for this @damacus

jerryaldrichiii (approved):
Great work @damacus. Super valuable for the community!

lib/resources/package.rb
```diff
@@ -19,8 +19,10 @@ class Package < Inspec.resource(1)
         its('version') { should eq 1.9.5 }
       end
     "
-
-    def initialize(package_name, opts = {}) # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def initialize(package_name, opts = {})
```
jquick, line 25: In the future we usually just cascade the rubocop disables on the method call. That way we don't have to enable after. Example:
```
def initialize(package_name, opts = {}) # rubocop:disable Metrics/AbcSize Metrics/CyclomaticComplexity  Metrics/PerceivedComplexity
...
end
```

damacus:
Awesome thanks for letting me know about that rubocop rule. I might have a
think about how to refactor that method in the future

miah:
Thanks for the contribution @damacus =)

Issue #3009, opened 2018-04-30
Enable Inspec profile caching with Chef Automate


srb3:
### 🎛 Description

feature request

Currently when inspec runs via the audit cookbook and the ChefServer::Fetcher, profiles from Chef Automate are pulled down every time even if they are present in the local cache `/root/.inspec/cache`.

Inspec seems to have the ability to check if the profile needs to be pulled down again via the cached_fetcher: 
https://github.com/chef/inspec/blob/master/lib/inspec/cached_fetcher.rb

however this requires that the target hash contains the sha256 sum of the profile in question, so it can be checked against the sha256 sum of the cached profile. 

if the target hash does not have the sha256 key then the download_archive_to_temp method is called in the ChefServer::Fetcher implementation.
https://github.com/chef-cookbooks/audit/blob/master/files/default/vendor/chef-server/fetcher.rb#L91-L104
 
It would be nice if Inspec could grab the sha256 sum of the profile from Chef Automate and decide if the profile needs to be download or not.

This would help ease the load on large scale Automate / Inspec setups.

### 🌍 InSpec and Platform Version
all

### 🤔 Replication Case

One Automate server with a profile available
One target node running Chef Client with the Audit Cookbook and the following attributes set:
```
{
  "audit": {
    "reporter": "chef-server-automate",
    "fetcher": "chef-server",
    "profiles": [
      {
        "name": "linux-baseline",
        "compliance": "admin/linux-baseline"
      }
    ]
  }
```
log into the target node in one terminal and run:
` while true; do if [ -f /tmp/chef-rest* ]; then tar xzf /tmp/chef-rest*; break; fi; done`
and in a second terminal log into the target node and run `chef-client`

### 💁 Possible Solutions

To enable this, work would probably be needed on the Inspec code and on the Automate compliance-service code.

Profile caching can be enabled by modifying the audit cookbook. But I think the solution 
would be better solved from within Inspec and Automate.

example of the changes needed in the audit cookbook to enable profile caching
https://github.com/chef-cookbooks/audit/compare/master...devoptimist:profile_caching

PR #3221, opened 2018-07-16
Cached profiles with Compliance Fetcher

itmustbejj:
This PR relies on https://github.com/chef/a2/pull/3070 to leverage the sha256 result returned from `/compliance/profiles/search` that is currently being called in `Compliance::API.exist?`. 

Fixes https://github.com/inspec/inspec/issues/3009
Fixes https://github.com/inspec/inspec/issues/3311


lib/fetchers/url.rb
```diff
@@ -91,15 +91,25 @@ def self.transform(target)
...
+        @upstream_sha256 = nil
+      end
       @insecure = opts['insecure']
       @token = opts['token']
       @config = opts
       @archive_path = nil
       @temp_archive_path = nil
     end
 
+    def upstream_sha256
```
jerryaldrichiii: You could make this a `attr_reader`:

https://github.com/itmustbejj/inspec/blob/e7ff9dd83ef01e5154ec074183d759bed2ea4b83/lib/fetchers/url.rb#L92

lib/fetchers/url.rb
```diff
@@ -91,15 +91,25 @@ def self.transform(target)
 
     attr_reader :files, :archive_path
 
-    def initialize(url, opts)
-      @target = url
+    def initialize(target, opts)
+      if target.is_a?(Hash) && target.key?(:url)
```
jerryaldrichiii: I think you can shorten the if statements to the below if you want (this will also handle if target isn't a String or Hash):
```
@target = target.is_a?(Hash) ? target[:url] : target 
@upstream_sha256 = target.is_a(Hash) ? target[:sha256] : nil
```

lib/bundles/inspec-compliance/target.rb
```diff
@@ -53,8 +53,15 @@ def self.resolve(target) # rubocop:disable PerceivedComplexity, Metrics/Cyclomat
 
         # verifies that the target e.g base/ssh exists
         profile = Compliance::API.sanitize_profile_name(uri)
-        if !Compliance::API.exist?(config, profile)
+        # Call profiles directly instead of exist? to capture the results
+        # so we can access the upstream sha256 from the results.
+        msg, profile_result = Compliance::API.profiles(config, profile)
```
jerryaldrichiii: Do you use `msg`? If not you can jsut do `_, profile_result`

lib/bundles/inspec-compliance/api.rb
```diff
@@ -101,25 +110,10 @@ def self.version(config)
...
-          profile_owner = p['owner_id'] || p['owner']
-          profile_owner == owner &&
-            p['name'] == id &&
-            (ver.nil? || p['version'] == ver)
-        end
-      else
-        false
-      end
+      _msg, profiles = Compliance::API.profiles(config, profile)
+      return profiles.empty? ? false : true
```
jerryaldrichiii: No need for a `return` here since it is the last thing called in the function. You can do:

```
!profiles.empty?
```

itmustbejj:
There is a single unit test in appveyor that is failing because the unit test does not seem to respect my stub_request in 2.3.3, even though it appears to work fine in travis on 2.3.6. Here is a diff of my stub_request, vs the stub_request appveyor's failed unit test is recommending:
```diff /tmp/my_stub /tmp/appveyor_stub
1,4c1,3
<       stub_request(:get, 'https://myautomate/profiles/admin').with(headers:           { 'Accept'=>'*/*',
<           'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
<           'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby',
<           'X-Data-Collector-Token'=>'' }).to_return(status: 200, body: profiles_response.to_json, headers: {})
---
> stub_request(:get, "https://myautomate/profiles/admin").
>   with(:headers => {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby', 'X-Data-Collector-Token'=>''}).
>   to_return(:status => 200, :body => "", :headers => {})```

The only difference is that I am returning a `body: profiles_response.to_json` and appveyor is recommending `body: ""`, which would make the test pointless. Is this safe to ignore since it is working on ruby 2.3.6?

**Comment by itmustbejj** [comment]
It looks like the one test that is failing in appveyor is on ruby 2.3.3 where webmock doesn't seem to recognize my "to_return" method.
```
  1) Error:
Compliance::API::exist?#test_0001_works with profiles returned by Automate:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET https://myautomate/profiles/admin with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby', 'X-Data-Collector-Token'=>''}

You can stub this request with the following snippet:

stub_request(:get, "https://myautomate/profiles/admin").
  with(:headers => {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby', 'X-Data-Collector-Token'=>''}).
  to_return(:status => 200, :body => "", :headers => {})

registered request stubs:

stub_request(:get, "https://myautomate/profiles/admin").
  with(:headers => {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby', 'X-Data-Collector-Token'=>''})
```
Here is my stub_request I define in the test.
```
      stub_request(:get, 'https://myautomate/profiles/admin')
        .with(headers: { 'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3
', 'Chef-Delivery-Enterprise'=>'automate', 'User-Agent'=>'Ruby', 'X-Data-Collector-Token'=>'' })
        .to_return(status: 200, body: profiles_response.to_json, headers: {})
```

The same stub_request is respected in every other ruby version suite, including ruby 2.3.6 in travis, which is using the same version of webmock (2.3.2) as the ruby 2.3.3 suite with the failing test on AppVeyor. I would think it's fine to ignore this failure since it passes on the latest patch version of ruby 2.3.x.

File: lib/bundles/inspec-compliance/target.rb [hunk 5]
```diff
@@ -13,7 +13,48 @@ module Compliance
...
+    def sha256
+      upstream_sha256.empty? ? super : upstream_sha256
+    end
+
+    def self.check_compliance_token(config)
+      if config['token'].nil? && config['refresh_token'].nil?
+        if config['server_type'] == 'automate'
+          server = 'automate'
+          msg = 'inspec compliance login https://your_automate_server --user USER --ent ENT --dctoken DCTOKEN or --token USERTOKEN'
+        elsif config['server_type'] == 'automate2'
```
chris-rock, line 38: Will this work with all version of Automate 2 or just the latest one?

lib/bundles/inspec-compliance/target.rb
```diff
@@ -27,34 +68,19 @@ def self.resolve(target) # rubocop:disable PerceivedComplexity, Metrics/Cyclomat
...
-        if !Compliance::API.exist?(config, profile)
+        # Call profiles directly instead of exist? to capture the results
+        # so we can access the upstream sha256 from the results.
+        _msg, profile_result = Compliance::API.profiles(config, profile)
+        if profile_result.empty?
           raise Inspec::FetcherFailure, "The compliance profile #{profile} was not found on the configured compliance server"
+        else
+          # If version wasn't specified the results are version sorted, so grab the first result.
+          # If version was specified, it will be the first and only result.
+          profile_info = profile_result[0]
```
chris-rock: What happens if we have more then one result. Seems unreliable to depend on the first entry since an order from the server is not guaranteed. We do not want to have different profiles running for the same identifier.

chris-rock (commented):
@itmustbejj Great idea to improve that!

lib/bundles/inspec-compliance/target.rb
```diff
@@ -13,7 +13,48 @@ module Compliance
...
+        @target = target[:url]
+        @upstream_sha256 = target[:sha256]
+      elsif target.is_a?(String)
+        @target = target
+        @upstream_sha256 = ''
+      end
+    end
+
+    def sha256
+      upstream_sha256.empty? ? super : upstream_sha256
```
itmustbejj: After talking to @chris-rock it sounds like we need some logic here to only use the upstream_sha256 for lockfile versions > 1 in order to not break vendored profiles with a lockfile using the old sha256 method.

lib/bundles/inspec-compliance/target.rb
itmustbejj: Yes, it should work with all versions of Automate 2. The logic for check_compliance_token is unchanged. I took it out of the resolve method because Metrics/AbcSize pointed to the resolve method having too many lines. Since it was a discrete piece of logic, I broke it out into its own method. You can compare it to the old code here: https://github.com/inspec/inspec/pull/3221/files#diff-e975c4d3a237ff3f9a0f22cc54abab4aL32

lib/bundles/inspec-compliance/target.rb
itmustbejj: This is a fair question, especially for legacy Compliance or Automate 1. For Automate 2, the sql results for namespaceProfilesAll and namespaceProfilesByName order the query results by profile name and then descending version. However, since we are using an unversioned api endpoint, there's no guarantee the behavior would not change. I could sort the profiles before returning them here: https://github.com/itmustbejj/inspec/blob/60d3a3746b3a03da4354f715f85ebf5cb3ddc658/lib/bundles/inspec-compliance/api.rb#L81

A2 code sorting profiles for reference:
https://github.com/chef/a2/blob/master/components/compliance-service/profiles/db/store.go#L498-L502
https://github.com/chef/a2/blob/master/components/compliance-service/profiles/db/store.go#L464
https://github.com/chef/a2/blob/master/components/compliance-service/profiles/db/store.go#L481

lib/bundles/inspec-compliance/target.rb
```diff
@@ -13,48 +13,80 @@ module Compliance
...
+    end
+
+    def self.resolve(target)
+      uri = get_target_uri(target)
       return nil if uri.nil?
 
       # we have detailed information available in our lockfile, no need to ask the server
       if target.respond_to?(:key?) && target.key?(:url)
         profile_fetch_url = target[:url]
+        profile_checksum = target[:sha256]
```
itmustbejj: @chris-rock Regarding, backwards compatibility of lockfiles generated using the previous cache checksum method, as the comment here states, if the lockfile is present the profile will never be requested from the server (bypassing any possibility of using the new sha256 returned from the server). Instead, the checksum from the lockfile is passed, maintaining existing behavior for lockfiles. If you are using a version of inspec with these changes to re-vendor with --overwrite, the new sha256 will be used, and the old vendored dependency will be removed.

lib/bundles/inspec-compliance/target.rb
itmustbejj: This was addressed in another comment [here](https://github.com/inspec/inspec/pull/3221/files#r211676981)

lib/bundles/inspec-compliance/target.rb
jquick: This sounds reasonable. What about when you have a lock file but the cache has been removed? I think it may try to download and then compare but please confirm.

itmustbejj: With my latest commit, it will download the profile and use the checksum in the lockfile when the profile is not cached/vendored. I verified this works on A1 and A2.

lib/bundles/inspec-compliance/api.rb
```diff
@@ -69,6 +74,10 @@ def self.profiles(config) # rubocop:disable PerceivedComplexity, Metrics/Cycloma
             e
           }
         end
+        # filter by name and version if they were specified in profile_filter
+        mapped_profiles.select! { |p|
```
jquick: I think this should be a do/end if its a multiline block.

lib/bundles/inspec-compliance/target.rb
```diff
@@ -13,50 +13,80 @@ module Compliance
...
+        # so we can access the upstream sha256 from the results.
+        _msg, profile_result = Compliance::API.profiles(config, profile)
+        if profile_result.empty?
           raise Inspec::FetcherFailure, "The compliance profile #{profile} was not found on the configured compliance server"
+        else
+          # Guarantee sorting by verison and grab the latest.
+          # If version was specified, it will be the first and only result.
+          # Note we are calling the sha256 as a string, not a symbol since
+          # it was returned as json from the Compliance API.
+          profile_checksum = profile_result.sort { |x| Gem::Version(x['version']) }[0]['sha256']
```
jquick: Will this break if they version a profile like `2.1-2` ? I think we have this same issue on A2 though were we only support SemVer. I'll create a ticket to update our `inspec check` to confirm that.

jquick (approved):
Looks good to me, just some small notes. Nothing blocking. Nice addition @itmustbejj !

jquick:
@itmustbejj - Looks like you have a failing windows test. I tried retrying but it looks like it may be due to the compliance changes. https://ci.appveyor.com/project/chef/inspec/build/1.0.6918/job/lfr6hs6w73qac9hc

lib/bundles/inspec-compliance/api.rb
```diff
@@ -101,25 +110,10 @@ def self.version(config)
       parsed
     end
 
-    # verifies that a profile
+    # verifies that a profile exists
```
jerryaldrichiii, line 113: Heh, love this!

lib/bundles/inspec-compliance/api.rb
```diff
@@ -101,25 +110,10 @@ def self.version(config)
...
-          profile_owner = p['owner_id'] || p['owner']
-          profile_owner == owner &&
-            p['name'] == id &&
-            (ver.nil? || p['version'] == ver)
-        end
-      else
-        false
-      end
+      _msg, profiles = Compliance::API.profiles(config, profile)
+      profiles.empty? ? false : true
```
jerryaldrichiii: I think `!profiles.empty?` is easier to read, I'll leave it up to you though.

jerryaldrichiii (approved):
Looks good to me! Only a few style issues but I won't hold the PR over it.

lib/bundles/inspec-compliance/api.rb
itmustbejj: @jerryaldrichiii This is a ternary operator to determine the return value (false/true) depending on the result of `profiles.empty?`.

It's equivalent of:
```
if profiles.empty?
  false
else
  true
end
```

lib/bundles/inspec-compliance/target.rb
itmustbejj: Yes it supports prerelease identifiers like that.

lib/bundles/inspec-compliance/api.rb
jerryaldrichiii: Yup, `profiles.empty?` returns a boolean (the ruby convention is all `?` methods return booleans). So what you are doing with the ternary is inverting it, which is the same as `!profiles.empty?`

```
[1] pry(main)> [].empty?
=> true
[2] pry(main)> ![].empty?
=> false
[3] pry(main)> [].empty? ? false : true
=> false
```

itmustbejj: Thanks, I see your point now. I'm happy to make that change.

PR #3224, opened 2018-07-17
windows_feature resource: Add DISM support

### LINKED ISSUE #3141: windows_feature resource fails on windows client OS types
### 🎛 Description

🐛 bug report

After installing a Windows feature on a client OS like Windows 10 the windows_feature resource in a test always fails due to the `Get-WindowsFeature` cmdlet not being available on client operating systems. 

https://github.com/inspec/inspec/blob/7b23fa479c516beaa0bd81d209ecf63133af2488/lib/resources/windows_feature.rb#L55  
Even if you install that module onto a client operating system you'll get this error:

`get-windowsfeature : The target of the specified cmdlet cannot be a Windows client-based operating system.`

Running InSpec 1.51.21 on Windows 10

### 🤔 Replication Case

1. Install a new feature (e.g. `dism /online /enable-feature /all /featurename:telnetclient`) on Windows 10.
2. Run an InSpec test to test for the existence of this new feature like so
```
control 'telnet client enabled' do
  describe windows_feature('TelnetClient') do
    it { should be_installed }
  end
end
```
3. Test should fail

### 💁 Possible Solutions

My suggestion would be using DISM to check for the existence of the feature (e.g. `dism /online /get-featureinfo /featurename:TelnetClient`). DISM is a tool I know is installed on both client and server Windows operating systems. Also DISM will return 0 if the feature exists and something else if it doesn't which could help make the windows_feature resource a bit cleaner.

jerryaldrichiii:
This modifies the `windows_feature` resource to fallback to DISM when the `Get-WindowsFeature` command is not available.

This closes #3141

### DIFF
--- a/lib/resources/windows_feature.rb
+++ b/lib/resources/windows_feature.rb
@@ -29,9 +29,6 @@ module Inspec::Resources
       @feature = feature
       @method = method
       @cache = nil
-
-      # verify that this resource is only supported on Windows
-      return skip_resource 'The `windows_feature` resource is not supported on your OS.' if !inspec.os.windows?
     end
 
     # returns true if the package is installed
--- a/lib/resources/windows_feature.rb
+++ b/lib/resources/windows_feature.rb
@@ -1,40 +1,24 @@
 # encoding: utf-8
 
-# check for a Windows feature
-# Usage:
-# describe windows_feature('DHCP Server') do
-#   it{ should be_installed }
-# end
-#
-# deprecated serverspec syntax:
-# describe windows_feature('IIS-Webserver') do
-#   it{ should be_installed.by("dism") }
-# end
-#
-# describe windows_feature('Web-Webserver') do
-#   it{ should be_installed.by("powershell") }
-# end
-#
-# This implementation uses the Get-WindowsFeature commandlet:
-# Get-WindowsFeature | Where-Object {$_.Name -eq 'XPS Viewer' -or $_.DisplayName -eq 'XPS Viewe
-# r'} | Select-Object -Property Name,DisplayName,Description,Installed,InstallState | ConvertTo-Json
-# {
-#     "Name":  "XPS-Viewer",
-#     "DisplayName":  "XPS Viewer",
-#     "Description":  "The XPS Viewer is used to read, set permissions for, and digitally sign XPS documents.",
-#     "Installed":  false,
-#     "InstallState":  0
-# }
 module Inspec::Resources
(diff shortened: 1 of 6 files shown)


lib/resources/windows_feature.rb
```diff
@@ -1,40 +1,24 @@
...
 
-# check for a Windows feature
-# Usage:
-# describe windows_feature('DHCP Server') do
-#   it{ should be_installed }
-# end
-#
-# deprecated serverspec syntax:
-# describe windows_feature('IIS-Webserver') do
-#   it{ should be_installed.by("dism") }
```
TheLunaticScripter, line 11: So I don't think this can be done but it would be nice to have the option to select what command runs the check, ie dism powershell or something else. The main reasoning would be that the names of features are slightly different and it would be a better user story if the user can control that.

TheLunaticScripter (approved):

TheLunaticScripter (approved):

jerryaldrichiii:
I agree @TheLunaticScripter, I pushed another commit up that adds this feature. Let me know what you think.

TheLunaticScripter (approved):
Awesome

jquick (commented):
This looks pretty good to me. Confirm for me that we are not going to bust the legacy
```
it{ should be_installed.by("dism") }
```
 format. I dont see the support code for it but I want to make sure im not crazy.

jerryaldrichiii:
@jquick Should be good, we never supported `.by`. Those examples were for ServerSpec.

jquick:
ah missed the `deprecated serverspec syntax:` we should be good

jquick (approved):
Thanks @jerryaldrichiii

lib/resources/windows_feature.rb
```diff
@@ -1,84 +1,126 @@
...
+      end
+    EOX
+
+    def initialize(feature, method = nil)
       @feature = feature
+      @method = method
       @cache = nil
 
       # verify that this resource is only supported on Windows
       return skip_resource 'The `windows_feature` resource is not supported on your OS.' if !inspec.os.windows?
```
miah: I don't think we need this as we already have `supports platform: 'windows'` on line 6.

miah (approved):
This looks good apart from my one point about platform.

lib/resources/windows_feature.rb
jerryaldrichiii: Agreed! Bit out of scope but I'm always willing to clean up old code. Fixed in latest commit.

Issue #1281, opened 2016-11-06
registry_key fails when item has a '.' in name


pburkholder:
### Description

Checks like this where the item, 'explorer.exe' has a dot ('.') in the name, always results in a NoMethodError:

```
      describe registry_key({
        hive: 'HKLM',
        key:  'Software\Policies\Microsoft\Internet Explorer\Main\FeatureControl\FEATURE_RESTRICT_ACTIVEXINSTALL',
      }) do
        its('explorer.exe') { should eq 1 }
      end
```

results in

```
 1) Registry Key HKLM\oftware\Policies\Microsoft\Internet Explorer\Main\FeatureControl\FEATURE_RESTRICT_ACTIVEXINSTALL explorer.exe
     Failure/Error: inner_subject.send(attr)

     NoMethodError:
       undefined method `exe' for nil:NilClass
```

### InSpec and Platform Version

1.4.1

### Replication Case

Tried this in two 2012R2 instances...

```
PS C:\Users\pburkholder> inspec shell
Welcome to the interactive InSpec Shell
To find out how to use it, type: ☺☻help☺☻
inspec> describe registry_key({
inspec>     hive: 'HKLM',
inspec>     key:  'Software\Policies\Microsoft\Internet Explorer\Main\FeatureControl\FEATURE_RESTRICT_ACTIVEXINSTALL',

inspec> }) do
inspec>   its('explorer.exe') { should eq 1 }
inspec> end

  Registry Key
     ✖  undefined method `exe' for nil:NilClass

Summary: 0 successful, 1 failures, 0 skipped
```

Try the above example...

### Possible Solutions

If you have already ideas how to solve the issue, add them here.

### Stacktrace

Please include the stacktrace output or link to a gist of it, if there is one.

username-is-already-taken2:
Hi there

I thought I would take a punt last night and see if I could find where the problem lies (I'm still find my way with ruby/pry)

the registry_key resource looks to be OK IMHO, the values looks to gathered correctly and the def registry_property_value does evaluate correctly if it has the right property :)

but at this point the property is now an array `["explorer", "exe"]
`
digging a bit further it looks like its being done here

C:\opscode\chefdk\embedded\lib\ruby\gems\2.3.0\gems\rspec-its-1.2.0\lib\rspec\its.rb

```
  def its(attribute, *options, &block)
      its_caller = caller.select {|file_line| file_line !~ %r(/lib/rspec/its) }
[...]

PR #3227, opened 2018-07-17
Generate describe code for an array of strings

james-stocks:
Context:
When testing a Windows registry key with a period character in it e.g. `explorer.exe` it is not possible to use `its("explorer.exe")` because the period would be interpreted as method chaining.
In this case, you must instead use `its(["explorer", "exe"])`
See https://github.com/inspec/inspec/issues/1281

This commit fixes `to_ruby`in `Inspec::Describe` so that it produces an array in the generated Inspec code instead of a string.

### DIFF
--- a/lib/inspec/objects/describe.rb
+++ b/lib/inspec/objects/describe.rb
@@ -11,7 +11,14 @@ module Inspec
       end
 
       def to_ruby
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @james-stocks. Nice addition.

jerryaldrichiii (approved):
Looks good to me, thanks @james-stocks!

Issue #3158, opened 2018-06-20
RFC: Skip Improvements


jquick:
### Current state

"Skip" is currently used to identify cases where a control cannot run:

1. The user doesn't want it to run when inheriting the profile.
    Example: Use your colleague's AWS profile but skip all controls that test S3.
2. The control isn't run when certain conditions are met.
    For example don't test MySQL if it isn't installed.
3. A control is skipped if it uses resources that are not supported on the target system.
    For example looking for a "package" when testing an API.

### Direction

Q: What can skip communicate?
A: It is a way to tell in reporting that a test wasn't run for an accepted reason.

Skip means: we don't execute a test we intended to run and let the user know we had to skip it.

Skip doesn't require the user to take any action. We let you know that the test was included, it was considered, but it was no executed for an intended reason. In this sense it's different from passed/failed results (indicating status) and failures (where a test that was run failed to execute).

### Design

To tackle skip, we'd establish:

1. A core use-case for skip is `only_if` in controls or files. Skip tests that don't match this condition and report that they were skipped. (same behavior as before)

2. A non-case are inherited profiles that aren't supported. We see this with combined profile layers. Until now, skip has served the value of outlining tests that were not run because the profile isn't supported. We don't want to use skip for this condition and instead hand it over to the profile inheritance and layering mechanism (and separate reporting). The user knows these tests were skipped, without reporting all of them. (new behavior: don't report these skipped tests)

3. A non-case is skipping controls whose resources are not supported. This is a change from previous convention, which turns some controls from being skipped into failing because they aren't supported. Fail explicitly with the condition that wasn't met. For example: instead of reporting `Config.Value.X != Expectation` we report that `Config is not supported on this system`.

4. A non-case are accepted tests or other exceptions. We already covered `only_if` exceptions and inheritance-based exceptions in 1+2. For acceptance, we want the test to be executed, i.e. this is a different domain than skip, since it covers tests that were executed and only require different reporting. (same behavior, will be covered in a separate RFC) (also see https://github.com/inspec/inspec/issues/3017)

Implicit resource-skips would now be turned into explicit user-defined skips (`only_if`, reported per control) or profile-level skips (`inspec.yml`, reported on the profile). If the user applies a resource on a target that doesn't support it and doesn't limit the profile's supported platforms, it will result in an error i.e. a mandatory user-action is required.

In Summary: turn some implicit skips into explicit actions. Remove other skips that were reported that don't provide value.

### Action Items

[...]

PR #3231, opened 2018-07-17
Ensure resources fail that target something that isn't supported

miah:
This PR Implements one task from our skip overhaul where we will now fail rather than skip on a unsupported platform. 

ref: https://github.com/inspec/inspec/issues/3158


jerryaldrichiii (approved):
Looks good to me, thanks @miah!

jquick (approved):
Nice work @miah

PR #3239, opened 2018-07-20
Add extra fault checking to git fetcher

james-stocks:
Raise with the stderr from `git ls-remote` if stderr is not empty.
This is useful when inspec runs in CI and you do not have direct control/troublshooting options.


jerryaldrichiii (approved):
Awesome improvement! Thank you @james-stocks.

jquick (approved):
Thanks!

PR #3242, opened 2018-07-20
cli: Downcase supermarket tool name to match URL

jerryaldrichiii:
This downcases the user provided tool name. Without this fetching the profile will fail because the Supermarket API downcases in the URL.

### DIFF
--- a/lib/bundles/inspec-supermarket/api.rb
+++ b/lib/bundles/inspec-supermarket/api.rb
@@ -48,7 +48,9 @@ module Supermarket
     # compares a profile with the supermarket tool info
     def self.same?(profile, supermarket_tool, supermarket_url = SUPERMARKET_URL)
       tool_owner, tool_name = profile_name(profile)
-      tool = "#{supermarket_url}/api/v1/tools/#{tool_name}"
+
+      # Tool name in Supermarket URL is downcased so we need to downcase
+      tool = "#{supermarket_url}/api/v1/tools/#{tool_name.downcase}"
       supermarket_tool['tool_owner'] == tool_owner && supermarket_tool['tool'] == tool
     end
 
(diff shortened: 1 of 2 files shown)


jerryaldrichiii:
Supermarket does more to slug the URL than we do here. To match this behavior we would have to add the `ActiveSupport` gem and also figure out how to make Thor deal with things such as spaces...both are outside the scope of what this PR covers in my opinion. A simple downcase will cover most issues.

See:

https://github.com/chef/supermarket/blob/7670960844bd0c2c4a71d6bd5b4114f1dc8b99c1/src/supermarket/app/models/category.rb#L28

http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize

lib/bundles/inspec-supermarket/api.rb
```diff
@@ -48,7 +48,9 @@ def self.info(profile, supermarket_url = SUPERMARKET_URL)
     # compares a profile with the supermarket tool info
     def self.same?(profile, supermarket_tool, supermarket_url = SUPERMARKET_URL)
       tool_owner, tool_name = profile_name(profile)
-      tool = "#{supermarket_url}/api/v1/tools/#{tool_name}"
+
+      # Tool name in Supermarket URL is downcased so we need to downcase
+      tool = "#{supermarket_url}/api/v1/tools/#{tool_name.downcase}"
```
chris-rock: Are there cases where `tool_name` is nil?

jerryaldrichiii: That comes from user input (parsing `supermarket:://tool_owner/tool_name`) so in theory it could be `nil`.

jerryaldrichiii:
@chris-rock I added another `.downcase` and handling for missing tool name. Let me know what you think.

jquick (approved):
Thanks @jerryaldrichiii

PR #3243, opened 2018-07-21
Since /proc/xen is an empty dir in Amazon Linux, inspec falsely detects docker instances as platform='xen'

woneill:
This was already noted in the comments, albeit for EL6. Since the code already checks `/proc/xen/capabilities` to determine if it's a Xen host vs Xen guest, use that more specific test to `set virtualization_data[:system] = 'xen'`


jquick (approved):
Thanks @woneill ! Nice enhancement.

miah (approved):

jerryaldrichiii (approved):
Nice improvement @woneill!

PR #3248, opened 2018-07-24
Correct docs for Windows registry_key resource

james-stocks:
Last update was mistaken about the workaround that should be used.
This commit corrects that mistake and provides some example output from `inspec shell`.


jquick (approved):
Thanks @james-stocks

chris-rock (approved):
Thank you @james-stocks

Issue #3183, opened 2018-06-28
Merge inherited profiles for A2 report


jquick:
### 🎛 Description

🙋 feature request?

We need to merge the json report to have the parent profiles contain all control and result information from the children. We should also populate the parent depends section with the sha256 and profile inheritance details. 

### 💁 Possible Solutions

Take the current json report and merge upstream on profiles based on `parent_profile` tags. Copy over all control and result information from children if they are not set on the parent.

clintoncwolfe:
#3203 sort of relates here, in spirit, sort of.

PR #3261, opened 2018-07-30
Add a merged json report for A2

jquick:
This PR creates a new json report which merges the profiles upstream into their parents. This will offload the work from A2 and allow us to send a smaller report up.

Fixes https://github.com/inspec/inspec/issues/3183


aaronlippold:
@jquick which use case are we hitting here? Just so I can go test in the right context?

jquick:
Basically this report combines all child profiles into their parents which is what A2 requires for its reporting structure. So instead of having:

```
report a
report b
report c -> child of b
report d -> child of c
```
We would now have 

```
report a
report b (with all results/controls/depends from c/d populated)
```

aaronlippold:
Hi,

Sounds good, what's the value of having `report a` in this case? For the
inherited profile which is a daisy-chain of profiles which creates an
`overlay` of an app/system stack - would `report b` not be what we really
want in the end? So my basic question as an auditor would be can we
streamline the data even more and drop `report a`? I must be missing a
detail in here :)

Thanks,

Aaron
--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

jquick:
Yes, in a normal case you are correct. You would only have multiple top level profiles if you are running multiple profiles at the same time.

chris-rock (approved):
Thank you @jquick

PR #3263, opened 2018-07-31
Enable inspec archive, check, and json to run as unpriveleged user

phiggins:
Hello,

I'm trying to run inspec archive, check, and json with an unprivileged user and it's failing because inspec expects to be able to write to a directory it doesn't have permission to. Here is an example of what I'm attempting and the error message I get:

```
vagrant@ubuntu-xenial:~/inspec$ sudo -u nobody ~/inspec/bin/inspec check ~/naughty_profile-0.1.0.tar.gz
/usr/lib/ruby/2.3.0/fileutils.rb:253:in `mkdir': Permission denied @ dir_s_mkdir - /home/vagrant/.inspec (Errno::EACCES)
        from /usr/lib/ruby/2.3.0/fileutils.rb:253:in `fu_mkdir'
        from /usr/lib/ruby/2.3.0/fileutils.rb:227:in `block (2 levels) in mkdir_p'
        from /usr/lib/ruby/2.3.0/fileutils.rb:225:in `reverse_each'
        from /usr/lib/ruby/2.3.0/fileutils.rb:225:in `block in mkdir_p'
        from /usr/lib/ruby/2.3.0/fileutils.rb:211:in `each'
        from /usr/lib/ruby/2.3.0/fileutils.rb:211:in `mkdir_p'
        from /home/vagrant/inspec/lib/inspec/dependencies/cache.rb:22:in `initialize'
        from /home/vagrant/inspec/lib/inspec/profile.rb:76:in `new'
        from /home/vagrant/inspec/lib/inspec/profile.rb:76:in `for_target'
        from /home/vagrant/inspec/lib/inspec/cli.rb:75:in `check'
        from /var/lib/gems/2.3.0/gems/thor-0.20.0/lib/thor/command.rb:27:in `run'
        from /var/lib/gems/2.3.0/gems/thor-0.20.0/lib/thor/invocation.rb:126:in `invoke_command'
        from /var/lib/gems/2.3.0/gems/thor-0.20.0/lib/thor.rb:387:in `dispatch'
        from /var/lib/gems/2.3.0/gems/thor-0.20.0/lib/thor/base.rb:466:in `start'
        from /home/vagrant/inspec/bin/inspec:12:in `<main>'
```

The exec command had an option, `--vendor-cache`, that allowed specifying where inspec should store things needed during processing, so I made that code work properly for the other commands.

On this branch the above example succeeds:

```
vagrant@ubuntu-xenial:~/inspec$ sudo -u nobody ./bin/inspec check ~/naughty_profile-0.1.0.tar.gz --vendor-cache=/tmp
Location:    /home/vagrant/naughty_profile-0.1.0.tar.gz
Profile:     naughty_profile
Controls:    0
Timestamp:   2018-07-31T01:38:20+00:00
Valid:       true

  !  No controls or tests were defined.

Summary:     0 errors, 1 warnings
```

I did not add any tests for the changed commands because I did not see any similar ones for the exec command, so I assumed these wouldn't be necessary.

I added the second commit because adding one line of code to the check method pushed it over the line length limit of the linter. I'd be happy to submit that code as a separate PR.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -67,6 +67,8 @@ module Inspec
     def self.profile_options
       option :profiles_path, type: :string,
         desc: 'Folder which contains referenced profiles.'
+      option :vendor_cache, type: :string,
+        desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
     end
 
     def self.exec_options
@@ -83,8 +85,6 @@ module Inspec
         desc: 'Use colors in output.'
       option :attrs, type: :array,
         desc: 'Load attributes file (experimental)'
-      option :vendor_cache, type: :string,
-        desc: 'Use the given path for caching dependencies. (default: ~/.inspec/cache)'
       option :create_lockfile, type: :boolean,
         desc: 'Write out a lockfile based on this execution (unless one already exists)'
       option :backend_cache, type: :boolean,
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -37,6 +37,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     o[:ignore_supports] = true
     o[:backend] = Inspec::Backend.create(target: 'mock://')
     o[:check_mode] = true
+    o[:vendor_cache] = Inspec::Cache.new(o[:vendor_cache])
 
     profile = Inspec::Profile.for_target(target, o)
     info = profile.info
@@ -70,6 +71,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     o[:ignore_supports] = true # we check for integrity only
     o[:backend] = Inspec::Backend.create(target: 'mock://')
     o[:check_mode] = true
+    o[:vendor_cache] = Inspec::Cache.new(o[:vendor_cache])
 
     # run check
     profile = Inspec::Profile.for_target(path, o)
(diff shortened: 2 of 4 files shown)


jquick:
I'm wondering if we could run into an issue with removing the `ignore_supports`. If we try to check a profile on a unsupported OS. I think we still want to skip those checks.

phiggins:
The `ignore_supports` flag appeared to never get set, but you probably know this code better than I do. :) I'd be happy to remove it.

frezbo:
@phiggins 
 - any particular reason to run as user nobody?
 - Why not run in a shell of `nobody` user? (I assume your `nobody` has a shell)

chris-rock:
@frezbo How would you like to see this working?

frezbo:
just wanted to know the use case

chris-rock:
We want to be able to run inspec as a non-priviledged user. This is helpful if you want to do a remote scan without the need to access the local system.

jquick:
Thanks @phiggins ! The `ignore_supports` does not even use the runner so its legacy code that is not needed. I think after a quick functional test for this we should be good to go.

phiggins:
> any particular reason to run as user nobody?

I used the nobody user because it should be usable crossplatform and avoided the necessity of having to create a new user for this purpose. Is there any particular reason not to?

> Why not run in a shell of nobody user? (I assume your nobody has a shell)

In the application I'm working on, it is exec'ing inspec and parsing the result, I am not directly executing it on the commandline. Is there an advantage to using a shell as nobody?

Sorry to give such terse answers, I currently have one hand in a cast and I'm not sure I understand the context of your questions.

phiggins:
@jquick I added a functional test. WDYT?

frezbo:
@phiggins @chris-rock the explanation by phiggins makes perfect sense, good work.

jquick (approved):
Thanks @phiggins

chris-rock (approved):
Thank you @phiggins for this improvement

Issue #1900, opened 2017-06-07
Make 'only_if' a bit more verbose by default and or to take a 'msg' parameter


aaronlippold:
### Description
Feature Request: 

In many cases, especially when using 'describe.one`, I would like to have a more specific message from only_if as to which only_if case was triggered and why. This would help provide better and more concise reporting when we have complex profiles.

I have seen at least three or four questions from users as to how to put a "message" statement in a control and have run into the situation myself.

Examples:

* OS rules that will skip because they are scanning a docker target, i.e. skipped kernel modules tests 
* Skipped because 'sshd' is not installed and or enabled etc.
* Rule skipped because of missing file
* Control skipped because service is not enabled

### InSpec and Platform Version
All

### Possible Solutions
```
only_if("my message if skipped")
  { <<CONDITION>> }
end
```

aaronlippold:
The default behavior could also be improved by just reporting what condition failed if there isn't a message supplied. i.e. `control skipped: package 'httpd' not installed` and or `control skipped: command 'psql' not installed` etc.

aaronlippold:
i.e. utilizing the matchers and function to construct the message

aaronlippold:
for example, 

```
only_if
  { inspec.package('httpd').installed? and inspec.service('httpd').installed? }
end
```
produces:
---
Control Skipped: The `httpd` package was not installed.
or 'the httpd service is not enabled.

PR #3267, opened 2018-08-01
Update `only_if` to allow user specified messages.

miah:
Updates only_if to allow a user to specify a message to say why it was skipped. If the user does not supply a message the default
message is used.

```
control 'nutcracker-connect-redis-001' do
  impact 1.0
  title 'Check if nutcracker can pass commands to redis'
  desc 'execute redis-cli set key command, to check connectivity of the service'

  only_if('redis is not installed.') do
    command('redis-cli').exist?
  end

  describe command('redis-cli SET test_inspec "HELLO"') do
    its(:stdout) { should match(/OK/) }
  end
end
```

```
  ↺  nutcracker-connect-redis-001: Check if nutcracker can pass commands to redis
     ↺  Skipped control due to only_if condition: redis is not installed.
```

Ref: https://github.com/inspec/inspec/issues/3158

Fixes: https://github.com/inspec/inspec/issues/1900

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -174,8 +174,9 @@ module Inspec
       rule.instance_variable_get(:@__skip_rule)
     end
 
-    def self.set_skip_rule(rule, value)
-      rule.instance_variable_set(:@__skip_rule, { result: value })
+    def self.set_skip_rule(rule, value, message = nil)
+      rule.instance_variable_set(:@__skip_rule,
+                                 { result: value, message: message })
     end
 
     def self.merge_count(rule)
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -117,7 +117,7 @@ module Inspec
     #
     # @param [Type] &block returns true if tests are added, false otherwise
     # @return [nil]
-    def only_if(message=nil)
+    def only_if(message = nil)
       return unless block_given?
       return if @__skip_only_if_eval == true
 
@@ -175,7 +175,7 @@ module Inspec
     end
 
     def self.set_skip_rule(rule, value)
-      rule.instance_variable_set(:@__skip_rule, value)
+      rule.instance_variable_set(:@__skip_rule, { result: value })
     end
 
     def self.merge_count(rule)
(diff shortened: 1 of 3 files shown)


jquick (commented):
Really nice enhancement on this. I think we will also need to update: https://github.com/inspec/inspec/blob/ca1746debb0a3a1c72394e63b991ad8c111eb478/lib/inspec/rule.rb#L177 to set the `skip_rule[:result]` and maybe add a test for this case.

aaronlippold:
Fixes: https://github.com/inspec/inspec/issues/1900 - please add to commit message on merge so #1900 get closed.

lib/inspec/rule.rb
```diff
@@ -183,8 +185,12 @@ def self.merge_count(rule)
 
     def self.prepare_checks(rule)
       msg = skip_status(rule)
-      return checks(rule) unless msg
-      msg = 'Skipped control due to only_if condition.' if msg == true
+      return checks(rule) unless msg[:result].eql?(true)
+      if msg[:result].eql?(true) && msg[:message].nil?
```
jerryaldrichiii: You can remove the first criterion since we cannot reach here if `msg[:result]` is false. I'd also modify it a bit by changing `msg` to `skip_check` (or some better name...) then do the following:

```ruby
if skip_check[:message]
  msg = "Skipped control due to only_if condition: #{msg[:message]}"
else
  msg = "Skipped control due to only_if condition."
end

**Review by jerryaldrichiii — CHANGES_REQUESTED** [review]
Looks great @miah only one small suggestion.

File: test/unit/profiles/profile_context_test.rb [hunk 2]
```diff
@@ -117,6 +117,7 @@ def load(call)
     describe 'global only_if' do
       let(:if_true) { "only_if { true }\n" }
```
clintoncwolfe: For my part, having all of these brief `let`s containing short Strings is a bit confusing; if the lets were inlined, I think it would read clearer.  But that was already there.

clintoncwolfe (approved):
Code looks good; tests are good on coverage but read a bit unclear too me.  Not a blocker. Thanks, @miah!

jerryaldrichiii (approved):

Issue #3269, opened 2018-08-06
aws_route_table expect 8 hexadecimal characters Route Table ID


kchistova:
**### Bug report.**

**inspec version:**
2.2.41

**Description:**
InSpec doesn't handle Route Table ID which contain 17 characters and expect just 8

**Case:**
 describe aws_route_table('rtb-12345a1234567b12c') do
    it { should exist }
Expected result:
 ✔  Route Table rtb-12345a1234567b12c should exist
Actual result:
×  Route Table
     aws_route_table Route Table ID must be in the format "rtb-" followed by 8 hexadecimal characters.

jerryaldrichiii:
Should this support both 8 and 17 characters? Or, is 8 never correct?

jerryaldrichiii:
Looks like both 8 and 17 are valid according to [this](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/resource-ids.html)

>A resource ID takes the form of a resource identifier (such as snap for a snapshot) followed by a hyphen and a unique combination of letters and numbers. Starting in January 2016, we're gradually introducing longer length IDs for Amazon EC2 and Amazon EBS resource types. The length of the alphanumeric character combination was in an 8-character format; the new IDs are in a 17-character format, for example, i-1234567890abcdef0 for an instance ID.

I think we should go through and add support for 8 OR 17 everywhere.

kchistova:
Yes, sorry!

It still support 8 now and I think it should support both

mahsoud:
#3277 failed with weird CI error? Any advice on how I can help, would love to see this fix

PR #3270, opened 2018-08-06
Change number of characters in route_table_id

kchistova:
Reason: issue #3269
https://github.com/inspec/inspec/issues/3269


jerryaldrichiii:
Great first stab @kchistova! Given https://github.com/inspec/inspec/issues/3269#issuecomment-410795454 I think this should support both 8 or 17 character versions of the ID though.

PR #3279, opened 2018-08-08
update docs to use new platform attributes in supports

chris-rock:

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -81,12 +81,14 @@ $ inspec check examples/profile
 
 # Platform Support
 
-Use the `supports` setting in the `inspec.yml` file to specify one (or more) platforms for which a profile is targeting. The list of supported platforms may contain simple names, names and versions, or detailed flags, and may be combined arbitrarily. For example, to target anything running Debian Linux:
+Use the `supports` setting in the `inspec.yml` file to specify one (or more) platforms for which a profile is targeting. The list of supported platforms may contain simple names, names and versions, or detailed flags, and may be combined arbitrarily. 
+
+For example, to target anything running Debian Linux:
 
 ```YAML
 name: ssh
 supports:
-  - os-name: debian
+  - platform-name: debian
 ```
 
 and to target only Ubuntu version 14.04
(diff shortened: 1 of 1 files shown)


docs/profiles.md
```diff
@@ -119,13 +121,17 @@ and to target all of these examples in a single `inspec.yml` file:
...
-  - os-name: ubuntu
+  - platform-name: debian
+  - platform-name: ubuntu
     release: 14.04
-  - os-family: redhat
+  - platform-family: redhat
   - platform: aws
 ```
 
+With InSpec 2.0, we introduced the `platform` attribute, which allows to restrict the execution to `os`, `aws`, `azure` or `gcp`
```
clintoncwolfe: Do we actually have an attribute literally named `platform`?  This might be confusing

docs/profiles.md
```diff
@@ -81,20 +81,22 @@ $ inspec check examples/profile
 
 # Platform Support
 
-Use the `supports` setting in the `inspec.yml` file to specify one (or more) platforms for which a profile is targeting. The list of supported platforms may contain simple names, names and versions, or detailed flags, and may be combined arbitrarily. For example, to target anything running Debian Linux:
+Use the `supports` setting in the `inspec.yml` file to specify one (or more) platforms for which a profile is targeting. The list of supported platforms may contain simple names, names and versions, or detailed flags, and may be combined arbitrarily.
```
clintoncwolfe: Could we have an explicit list of the keys that are allowed here? ` may contain simple names, names and versions, or detailed flags` does not tell what is allowed and what isn't.  Examples are good, but don't tell you the complete picture.

clintoncwolfe (dismissed):
Thanks, @chris-rock ! Just some clarifications

docs/profiles.md
jquick: Yes we do have `platform` which under the hood is if platform-name OR platform-family are true. I'll re-write this to be a bit more clear.

chris-rock:
Thank you @clintoncwolfe for the questions and @jquick for the clarification

clintoncwolfe (approved):
Thanks!

PR #3284, opened 2018-08-10
Yaml global attributes

jquick:
This PR is still in works. I am looking for functional and big picture feedback. Main changes:

1) Convert attributes to a hash on the profile context
2) Allow loading of attributes from yml when loading a profile (child or parent)
3) Add type/require validation to attributes

Functional tests are completed, just missing some object level unit tests at this time.


aaronlippold:
Everything seems reasonable so far, one thing to keep in mind right at the beginning is how will the code handle it when we do a dependent profile, over-ridding via the cli - especially when thinking on load priority - and finally will we stock over-ride values when the same key exists in the chain or will we combine the data or further or perhaps even more flexibility if we  have a last-in-wins default we could add a flag to an attribute to allow it to tell us what the child-attribute wants to be: 'ignore-in-favor of parent', 'override-parent' or 'join-parent'. 

Other things to keep in mind:
* I would update (2) to say `Allow loading of attributes from yml on loading a profile or via a flag on the CLI: --attrib='my_attribe':"value"`,'my_attrib2':"${ENV_VAR}"` or simular: trying to cover the dynamic or secret password, user, or dynamic_pipeline created value issue while we are here :)  
 
* what happens if a parent attribute is type String and the child attribute is type Int, or type single element array of type x or ...  

* what happens if the validation of the child or parent differs - who wins or is the default to make sure it meets both ( which could be fine and offer some interesting logical options )

jquick:
Thanks @aaronlippold. For the `--attrib` addition, while nice is out of scope for this specific PR. The inheritance part is a great thing to bring up. Currently attributes are tied to a profile context. That is to say, attributes set on a child will interact and run for any child controls pulled in. Likewise on a parent any attributes set, even the same named as a child, will act on any parent controls. To override you can simply override the child control in the parent and it will use the parents attributes.

test/unit/mock/profiles/global_attributes/controls/attributes.rb
```diff
@@ -0,0 +1,39 @@
+val_user = attribute('user', default: 'alice', description: 'An identification for the user')
+val_user_override = attribute('val_user_override', default: 'alice', description: 'An identification for the user')
+
+describe val_user do
```
clintoncwolfe, line 4: I might suggest you use string names that describe a test here, and then set subject in the describe block.
```
describe 'reading an attribute in a file-level definition with a default value and a value in metadata' do
  subject(val_user)
  it { should cmp 'alice' }
end
```
That may add or remove clarity, depending on your perspective :)

test/unit/mock/profiles/global_attributes/controls/attributes.rb
```diff
@@ -0,0 +1,39 @@
...
+  it { should cmp 'alice' }
+end
+
+describe val_user_override do
+  it { should cmp 'bob' }
+end
+
+control 'test val_numeric' do
```
clintoncwolfe, line 12: Here I definitely think you should use a descriptive control name. If I see a list of tests with descriptive names, it's easy to see which combinations we aren't testing.

test/unit/mock/profiles/global_attributes/controls/attributes2.rb
```diff
@@ -0,0 +1,23 @@
+describe attribute('val_string') do
```
clintoncwolfe, line 1: I suggest renaming this file to be 'attribute_validation.rb'

test/unit/mock/profiles/global_attributes/controls/attributes.rb
```diff
@@ -0,0 +1,39 @@
+val_user = attribute('user', default: 'alice', description: 'An identification for the user')
```
clintoncwolfe, line 1: I suggest renaming this file to be something alike 'attribute_values.rb' - it's focused on testing how / when values get set

test/unit/mock/profiles/global_attributes/inspec.lock
```diff
@@ -0,0 +1,8 @@
+---
+lockfile_version: 1
+depends:
+- name: myprofile1
+  resolved_source:
+    url: https://s3-eu-west-1.amazonaws.com/apop-bucket/profiles/myprofile1-1.0.0.tar.gz
```
clintoncwolfe, line 6: Is this a stable location for a test fixture?

clintoncwolfe (changes requested):
Just some naming suggestions in the functional tests - I have not reviewed the main code yet.

test/unit/mock/profiles/global_attributes/inspec.lock
jquick: The location is does not actually matter as we are just pulling from the cache. I'll rename this to be more testlike though.

jquick:
This PR is going to have some significant changes in the next few days with some new decisions being made. Closing this for now and will reopen a fresh one.

PR #3285, opened 2018-08-10
Allow the jsonAutomate report to be executed from cli

jquick:

This change allows the A2 team or other reports to test the JsonMerged report used in A2. This change also allows us to easily replace the report type in the Audit cookbook.

Please not the JsonMerged already has a full test suite and this change is just allowing it to be accessed via the reporter cli flag.

Note this is intentionally undocumented for now.

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -168,7 +168,7 @@ module Inspec
         'documentation',
         'html',
         'json',
-        'json-merged',
+        'json-automate',
         'json-min',
         'json-rspec',
         'junit',
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -168,6 +168,7 @@ module Inspec
         'documentation',
         'html',
         'json',
+        'json-merged',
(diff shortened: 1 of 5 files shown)


chris-rock:
I am wondering if we should call this `json-3` to indicate that this is becoming the default json reporter for inspec 3. Then we can also introduce `json-2` to make the current json reporter available with InSpec 3 too. If users will rely on the old behavior, they just need to switch to the reporter.
@arlimus what do you think?

jquick:
After talking this over we will update the name to `json-automate` as we may have additional automate info in the report.

chris-rock:
Please add comment, that this is for Chef internal and we reserve the right to introduce breaking changes to that reporter at any point. We also are not going to document the reporter in our public docs.

alexpop:
I pushed a small branch to the `audit` cookbook in prep for this change when it goes in:

https://github.com/chef-cookbooks/audit/compare/ap/chef-automate-reporter

jerryaldrichiii (approved):

miah (approved):

PR #3294, opened 2018-08-16
Allow apache_conf include files to be empty

### LINKED ISSUE #3232: apache_conf is skipped due to empty include file
### 🎛 Description

🐛 bug report

When the Apache config includes a file that is empty the control is skipped.

2.2.38

Platform:

Name:      cloudlinux
Release:   7.5
Arch:      x86_64

### 🤔 Replication Case

touch /tmp/empty.conf
echo 'Include /tmp/empty.conf' >/tmp/apache.conf
inspec shell
inspec> describe apache_conf('/tmp/apache.conf') do
inspec>   its('Include') { should exist }
inspec> end

### 💁 Possible Solutions

Update the call in read_file to read_file_content(path, true)

KrisShannon:
Fixes #3232 

In the `read_file` method we can call `read_file_content` with the second
argument `true` to avoid skipping on an empty file.

We will still skip the control if the main configuration file is empty
as there is still an explicit call to `read_file_content` without this
argument in the `read_content` method.

### DIFF
--- a/lib/resources/apache_conf.rb
+++ b/lib/resources/apache_conf.rb
@@ -125,7 +125,7 @@ module Inspec::Resources
     end
 
     def read_file(path)
-      @files_contents[path] ||= read_file_content(path)
+      @files_contents[path] ||= read_file_content(path, true)
     end
 
     def conf_dir


jquick (approved):
Thanks @KrisShannon

PR #3308, opened 2018-08-20
WIP - Plugins installer

clintoncwolfe:
Temporary PR to verify TravisCI behavior re: Inspec.config_dir


clintoncwolfe:
Just opened for a ... reasons that escape me.  Closing, we don't do WIPs.

PR #3313, opened 2018-08-21
Infer `--sudo` when `--sudo-password` is used

jerryaldrichiii:
This does the following:
  - Adds `--sudo` if using `--sudo-password`
  - Warns the user if using `--sudo-password` without `--sudo`
  - Adds unit tests for `Inspec::BaseCLI#opts`

This closes #3313

### DIFF
--- a/test/unit/base_cli_test.rb
+++ b/test/unit/base_cli_test.rb
@@ -7,6 +7,36 @@ require 'thor'
 describe 'BaseCLI' do
   let(:cli) { Inspec::BaseCLI.new }
 
+  describe 'opts' do
+    it 'raises if `--password/--sudo-password` are used without value' do
+      default_options = { mock: { sudo_password: -1 } }
+      Inspec::BaseCLI.stubs(:default_options).returns(default_options)
+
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @jerryaldrichiii

test/unit/base_cli_test.rb
```diff
@@ -7,6 +7,36 @@
...
+    it 'raises if `--password/--sudo-password` are used without value' do
+      default_options = { mock: { sudo_password: -1 } }
+      Inspec::BaseCLI.stubs(:default_options).returns(default_options)
+
+      e = proc { cli.send(:opts, :mock) }.must_raise(ArgumentError)
+      e.message.must_match /Please provide a value for --sudo-password/
+    end
+
+    it 'assumes `--sudo` if `--sudo-password` is used without it' do
+      default_options = { mock: { sudo_password: 'p@ssw0rd' } }
```
miah, line 20: yep, no matter how many times you type p@ssw0rd, it will show to us as ********

jerryaldrichiii: Gah! I missed my chance to use hunter2.

Issue #3176, opened 2018-06-26
RFC unified attributes


arlimus:
### Description

Currently attributes are defined on a per-file basis:

```ruby
> controls/a.rb
user = attribute('user', required: true)
...
```

For attributes that are shared across files, this leads to a lot of duplication and potential overrides:

```ruby
> controls/b.rb
user = attribute('user', required: true)
...
```

### Goals

- define attributes in one place only and make them accessible across all inspec files
- prevent accidental attribute overrides 

### Design

`inspec.yml` will define the attributes:

```yaml
attributes:
- name: NAME
  required: [false]/true
  type: [any]/string/numeric/boolean/regex/array/hash
  default: []
```

Attributes are then used in regular control files:

```
attribute(NAME)
```

**Full example:**

`inspec.yml` contains:

```yaml
attributes:
- name: port
  type: int
  default: 443
```

`controls/example.rb` contains:

```ruby
control ssl(port: attribute('port')) do
  it { should be_enabled }
end
```

### Action Items
1) Implement new yml attribute definition
 - include basic validation (regex)
2) Deprecate the inline control attribute definition

### References

https://github.com/inspec/inspec/issues/719

aaronlippold:
I think this is a great solution. The one thing I might add is pattern
definitions so that we can do some sort of input validation and or default
validation. Is perhaps a string or a number has to match a particular
pattern that would be very useful to enforce standards

clintoncwolfe:
I like validation, too.  In the past, I've worked with validators that work like this:
validation: regex/array/proc

The array would be an enumeration of valid values.

We can't really have a Proc in YAML, so that's probably out.  We could reference a method name in a library file.  That all sounds like a nice later enhancement.

aaronlippold:
the validation would be in `inspec check` process

I am not sure how we would define the patterns outside regex or accepted values in an array or hash or whatnot.

aaronlippold:
@arlimus et all, one thing we would like to add to this is the long awaited ability to pass attribute values ( such as `user` or `password` ) via the CLI directly. ( i.e. not using a file ) so we can ease the integration into the CI/CD pipeline. We have the ability now via a json file but being able to pass single attribute values via the cli would be very useful. We do this in kitchen-inspec now but don't have support in the CLI for it.

aaronlippold:
[...]

PR #3318, opened 2018-08-22
Implement InSpec global attributes

jquick:
This PR implements the global attributes for InSpec. This allows you to set attributes in profile's `inspec.yml` to be used in any control. You also now have the option to return a created attribute anywhere in a inspec test code. This includes outside of a control, inside a control block, inside a describe block, or inside a test case.

Fixes https://github.com/inspec/inspec/issues/3176


aaronlippold (commented):
This is looking good to me :) 👍

aaronlippold:
I am so excited to start playing with this ... can I do it now?

ojongerius:
Nice work! Any idea on timelines of this getting merged & released?

aaronlippold:
@jquick did you document the `secrets_override`? i.e. can we label an attribute `secret` and it won't print out on the CLI?

jquick:
The plan is to get this in this week. The `secrets_override` is internal only and is only used from the `--attrs` file as a passthrough. In a future PR we may be able to tap into the:

https://github.com/inspec/inspec/blob/master/lib/inspec/formatters/base.rb#L46 

With something like that though.

aaronlippold:
, hi, cool sounds great yes that probably would be good just to avoid
confusion in long-term maintenance. Given I immediately thought that was in
reference to the secrets capability of the describe

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
 
     it 'returns the user-configured default value if no value is assigned (false)' do
       attribute = Inspec::Attribute.new('test_attribute', default: false)
       attribute.value.must_equal false
     end
   end
+
+  describe 'validate required method' do
+    it 'does not error if a default is set' do
+      attribute = Inspec::Attribute.new('test_attribute', default: 'default_value', require: true)
```
clintoncwolfe: `required` reads better to me than `require`, but it's not a blocker.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      attribute.value.must_equal 'default_value'
+    end
+
+    it 'does not error if a secrets value is set' do
+      attribute = Inspec::Attribute.new('test_attribute', secrets_override: 'default_value', required: true)
+      attribute.value.must_equal 'default_value'
+    end
+
+    it 'returns an error is no value is set' do
```
clintoncwolfe: s/is/if/

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
     end
   end
+
+  describe 'validate required method' do
+    it 'does not error if a default is set' do
+      attribute = Inspec::Attribute.new('test_attribute', default: 'default_value', require: true)
+      attribute.value.must_equal 'default_value'
+    end
+
+    it 'does not error if a secrets value is set' do
```
clintoncwolfe: I'd like to see a test in which both `default_value` and `secrets_override` are provided, specifying which takes precedence.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      opts[:type] = ':string'
+      attribute.send(:validate_value_type, 'string').must_be_nil
+    end
+
+    it 'validates a integer type' do
+      opts[:type] = 'integer'
+      attribute.send(:validate_value_type, 123).must_be_nil
+    end
+
+    it 'validates a float type' do
```
clintoncwolfe: I seem to recall us discussing as a team that the terms `float`, `double`, `single`, etc are developer-centric and may be unfamiliar to inspec users, and that we were going to use the term `numeric` to describe a non-integral value.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      e.message.must_match /Attribute 'test_attribute' is required and does not have a value./
+    end
+  end
+
+  describe 'validate value type method' do
+    let(:opts) { {} }
+    let(:attribute) { Inspec::Attribute.new('test_attribute', opts) }
+
+    it 'validates a string type' do
+      opts[:type] = ':string'
```
clintoncwolfe: I'm confused a bit here: some of these test cases specify the type as ':string' (a literal string beginning with a colon character) and others do not.  Are they expected to be Symbols?  Why do some have the colon prefix and others do not?

lib/inspec/attributes.rb
```diff
@@ -0,0 +1,52 @@
+require 'forwardable'
+require 'singleton'
+require 'inspec/objects/attribute'
+
+module Inspec
+  class Attributes
```
clintoncwolfe: I feel this class should be renamed: `AttributeCollection`, `AttributeRegistry`, or similar.  Having  Attribute and Attributes classes might be confusing.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+      invalid_type = false
+      if type == 'Regexp'
+        invalid_type = true if !value.is_a?(String) || (!value.start_with?('/') || !value.end_with?('/'))
+      elsif type == 'Boolean'
+        invalid_type = true if ![true, false].include?(value)
+      elsif instance_eval("value.is_a?(#{type})") == false
+        invalid_type = true
+      end
+
+      raise Inspec::AttributeError, "Attribute '#{@name}' with value '#{value}' does not validate to type '#{type}'." if invalid_type == true
```
clintoncwolfe: Create a more specific exception class for this.  Failing validation is a pretty common and specific case.
```
class Inspec::AttributeValidationError < Inspec::AttributeError; end
```

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+    end
+
```
clintoncwolfe: Re-enable rubocop checks after disabling.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+        end
+      else
+        value
+      end
+    end
+
+    def validate_value_type(value) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      type = clean_type(@opts[:type])
+      return if type == 'Any'
+      raise Inspec::AttributeError, "Type '#{type}' is not a valid attribute type" if !VALID_TYPES.include?(type)
```
clintoncwolfe: I'd recommend a more specific exception class here - perhaps AttributeSpecificationError.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+      value = convert_string_to_numeric(type, value) if NUMERIC_TYPES.include?(type) && value.is_a?(String)
+
+      invalid_type = false
+      if type == 'Regexp'
+        invalid_type = true if !value.is_a?(String) || (!value.start_with?('/') || !value.end_with?('/'))
```
clintoncwolfe: in addition to checking the `/` delimiters, you should verify it is a valid Regexp using `Regexp.try_convert`.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+      elsif type == 'Boolean'
+        invalid_type = true if ![true, false].include?(value)
+      elsif instance_eval("value.is_a?(#{type})") == false
```
clintoncwolfe: instance_eval should only be used as a last resort, or when applying DSLs.  Here, you can solve the problem you have - converting a type-specifying String into a const naming a class - in a cleaner, safer way:
```
value.is_a?(Module.const_get(type))
```

You don't even have to convert to  Symbol or anything:
```
irb(main):003:0> [].is_a?(Module.const_get('Array'))
=> true
```

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      attribute.send(:validate_value_type, 'string').must_be_nil
```
clintoncwolfe: Are we specifying that the behavior of `validate_value_type` is to return nil when the value is valid?  From the implementation, it appears that the behavior is to be unexceptional on valid, and raise an exception on any invalid case.  If that is the intended behavior, then these tests are overspecified.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      opts[:type] = 'integer'
+      attribute.send(:validate_value_type, 123).must_be_nil
+    end
+
+    it 'validates a float type' do
+      opts[:type] = ':float'
+      attribute.send(:validate_value_type, 123.33).must_be_nil
+    end
+
+    it 'validates a regex type' do
```
clintoncwolfe: I'd like to see a test that tries a properly fenced but invalid Regex, like `/(.+/` (no terminating paren).

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+
+    it 'validates a boolean type' do
+      opts[:type] = ':any'
+      attribute.send(:validate_value_type, false).must_be_nil
+      attribute.send(:validate_value_type, true).must_be_nil
+      attribute.send(:validate_value_type, 1).must_be_nil
+      attribute.send(:validate_value_type, 'bob').must_be_nil
+    end
+
+    it 'returns an error if a invalid type is set' do
```
clintoncwolfe: I'd want negative cases for each type, not just string.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      opts[:type] = ':string'
+      e = proc { attribute.send(:validate_value_type, 123) }.must_raise(Inspec::AttributeError)
```
clintoncwolfe: Using a custom error class here for validation failures would let you have accessors on the exception for things like expected type, actual value seen, etc, which simplify the UI and testing.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+
+    def validate_required(value)
+      if (!@opts.key?(:defualt) && value.nil?) || (@opts[:default].nil? && value.nil?)
+        raise Inspec::AttributeError, "Attribute '#{@name}' is required and does not have a value."
+      end
+    end
+
+    def clean_type(type)
+      type = type.gsub(/^:/, '').capitalize
+      case type
```
clintoncwolfe: Consider re-packing this into a Hash of abbreviations => normalized types.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+    private
+
+    def validate_required(value)
+      if (!@opts.key?(:defualt) && value.nil?) || (@opts[:default].nil? && value.nil?)
+        raise Inspec::AttributeError, "Attribute '#{@name}' is required and does not have a value."
+      end
+    end
+
+    def clean_type(type)
+      type = type.gsub(/^:/, '').capitalize
```
clintoncwolfe: A comment explaining why allowing an optional colon in a String is allowed would be helpful here.  

If a type specification can begin with a colon, might you recieve a Symbol here?  In which case you might want to `.to_s` prior to working with the type spec.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+        'Integer'
+      when 'Regex'
+        'Regexp'
+      else
+        type
+      end
+    end
+
+    def convert_string_to_numeric(type, value)
+      if value =~ /\A[-+]?[0-9]*\.?[0-9]+\Z/
```
clintoncwolfe: We probably should not be in the business of parsing numerics.  Use Integer(value,10) and Float(value), both of which will throw ArgumentErrors if invalid, and know all the tricks like scientific notation, etc.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      e.message.must_match /Attribute 'test_attribute' with value '123' does not validate to type 'String'./
+    end
+  end
+
+  describe 'clean type method' do
```
clintoncwolfe: Need a negative test: what if you pass something unrecognized as the type?

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,115 @@
...
+      attribute.send(:clean_type, ':int').must_equal 'Integer'
+    end
+
+    it 'returns the same value if there is nothing to clean' do
+      attribute.send(:clean_type, 'String').must_equal 'String'
+    end
+  end
+
+  describe 'convert string to numeric method' do
+    it 'converts value to a int' do
```
clintoncwolfe: Need negative tests for all of these: how does it handle garbage input?

Also more edge cases.  Zeros.  Strings that begin with zero.  Negative numbers.  Big numbers.  A fraction (is that accepted? decide!)

lib/inspec.rb
```diff
@@ -15,6 +15,8 @@
 require 'inspec/shell'
 require 'inspec/formatters'
 require 'inspec/reporters'
+require 'inspec/attributes'
```
clintoncwolfe: This PR significantly changes the UX around attributes, but includes no documentation changes.

lib/inspec/objects/attribute.rb
```diff
@@ -28,16 +44,27 @@ def respond_to_missing?(_, _)
     def initialize(name, options = {})
       @name = name
       @opts = options
-      @value = nil
+      validate_value_type(default) if @opts.key?(:type) && @opts.key?(:default)
+      if @opts.key?(:secrets_override)
```
clintoncwolfe: I don't know what this option means, but I'd like to know, and than discuss options for other names.  Attributes have no integration with secrets management (though you could use attributes to get that effect).  Using the word 'secrets' is thus problematic.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
     def to_s
       "Attribute #{@name} with #{@value}"
     end
+
+    private
+
+    def validate_required(value)
+      if (!@opts.key?(:defualt) && value.nil?) || (@opts[:default].nil? && value.nil?)
```
clintoncwolfe: A comment here clarifying that "At this point, any secrets_override value should already be in place" would have helped my understanding.

lib/inspec/objects/attribute.rb
```diff
@@ -71,5 +98,59 @@ def to_ruby
...
+        raise Inspec::AttributeError, "Attribute '#{@name}' is required and does not have a value."
```
clintoncwolfe: A specific exception here would be good, if you're into that sort of thing.

lib/inspec/attributes.rb
```diff
@@ -0,0 +1,52 @@
...
+    def_delegator :list, :select
+
+    # this allows access to the list hash without having to call instance
+    def self.[](profile)
+      instance[profile]
+    end
+
+    def initialize
+      @list = {}
+      @profile_overrides = {}
```
clintoncwolfe: I'd like to see some comments or docs somewhere (here is fine) explaining the intent of "profile override".

test/unit/attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  describe 'creating a profile attribute' do
+    it 'creates a attribute without options' do
+      attributes.find_or_create('test_attribute', 'profile')
+      # confirm we get the dummy default
+      attributes.find_or_create('test_attribute', 'profile').value.class.must_equal Inspec::Attribute::DEFAULT_ATTRIBUTE
+    end
+
+    it 'creates a attribute without options' do
```
clintoncwolfe: Copypasta - something like 'it creates an attribute with a default value'

test/unit/attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+describe Inspec::Attributes do
+  let(:attributes) { Inspec::Attributes.instance }
+
+  def setup
+    Inspec::Attributes.instance.__reset
+  end
+
+  describe 'creating a profile attribute' do
+    it 'creates a attribute without options' do
+      attributes.find_or_create('test_attribute', 'profile')
```
clintoncwolfe: On first reading, I thought the second arg might be indicating a "scope" for the attribute (e.g., profile, control, describe, test).  But instead I think this is just the name of a dummy profile.  I suggest you use a different word - 'my-profile', 'test-profile', etc.  Please don't use 'foo' :)

lib/inspec/attributes.rb
```diff
@@ -0,0 +1,52 @@
...
+        list[profile] = {} unless list.key?(profile)
+        list[profile][name] = Inspec::Attribute.new(name, options)
+      end
+    end
+
+    def register_profile_override(name, override)
+      @profile_overrides[name] = override
+    end
+
+    def select_profile(profile)
```
clintoncwolfe: I found this name confusing - on first reference, I thought this meant that Attributes maintained a particular "profile context", and a consumer of the class would need to set the context externally.

Instead, it seems this simply lists things.  I'd recommend renaming this to `list_attributes_for_profile` or similar.

lib/inspec/attributes.rb
```diff
@@ -0,0 +1,52 @@
...
+    def_delegator :list, :key?, :profile_exist?
+    def_delegator :list, :select
+
+    # this allows access to the list hash without having to call instance
+    def self.[](profile)
+      instance[profile]
+    end
+
+    def initialize
+      @list = {}
```
clintoncwolfe: A comment mentioning the expected data structure would be great.  What is this a hash of?  What keys?

lib/inspec/attributes.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      if list.key?(profile) && list[profile][name] && options.empty?
+        list[profile][name]
+      else
+        list[profile] = {} unless list.key?(profile)
+        list[profile][name] = Inspec::Attribute.new(name, options)
+      end
+    end
+
+    def register_profile_override(name, override)
```
clintoncwolfe: A comment giving context and intention would be good; I *think* this is used when a user re-names a dependent profile in their inspec.yaml.  Consider `register_profile_alias` - that indicates that the key change is the name.

test/unit/attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    it 'creates a attribute without options' do
+      attributes.find_or_create('color', 'profile', default: 'silver')
+      attributes.find_or_create('color', 'profile').value.must_equal 'silver'
+    end
+  end
+
+  describe 'creating a profile with a name override' do
+    it 'creates a attribute without options' do
```
clintoncwolfe: A better name for this test might be 'uses the profile override name to access default values"

test/unit/attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+      attributes.register_profile_override('old_profile', 'new_profile')
+      attributes.find_or_create('color', 'new_profile', default: 'blue')
+      attributes.find_or_create('color', 'new_profile').value.must_equal 'blue'
```
clintoncwolfe: I think you might also want a plain `find` and a plain `create` method.  The intent here is to show if you create an attribute, then you can find it using two different profile names.  Not a blocker.

test/unit/attributes.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  describe 'creating a profile and select it' do
+    it 'creates a profile with attributes' do
+      attributes.find_or_create('color', 'profile', default: 'silver')
+      attributes.find_or_create('color2', 'profile', default: 'blue')
+      attributes.find_or_create('color3', 'profile', default: 'green')
+      attributes.select_profile('profile').size.must_equal 3
+    end
+  end
+end
```
clintoncwolfe: Since Attributes is a data collection, I'd like to see more tests around what it is returning: are they they right class?  Can you feed it garbage, will it reject things?

test/unit/mock/profiles/global_attributes/README.md
```diff
@@ -0,0 +1,3 @@
+# Example InSpec Profile
+
+This example shows the implementation of an InSpec profile.
```
clintoncwolfe: Some details about why this text fixture profile is unique would be helpful to future developers.

test/unit/mock/profiles/global_attributes/controls/attributes.rb
```diff
@@ -0,0 +1,47 @@
+val_user = attribute('user', default: 'alice', description: 'An identification for the user')
```
clintoncwolfe: Suggest renaming this file to 'attribute_read_scoping'.

test/unit/mock/profiles/global_attributes/controls/attributes2.rb
```diff
@@ -0,0 +1,42 @@
+describe 'test the val_string attribute set in the global inspec.yml' do
```
clintoncwolfe: Consider re-organizing your tests.  You have some tests here that deal with validation, and that's good.  Some of those tests overlap with verifying that you can read values from the inspec.yaml, which I should should be tested separately.  Finally, you have controls that test profile inheritance, which is great, but should be in a separate control file.

test/unit/mock/profiles/global_attributes/inspec.yml
```diff
@@ -0,0 +1,47 @@
...
+    - a
+    - b
+    - c
+- name: val_hash
+  type: hash
+  default:
+    a: true
+    b: false
+    c: '123'
+- name: val_no_default
```
clintoncwolfe, line 41: I'd like to see one with a type, but no default.

test/functional/attributes_test.rb
```diff
@@ -20,4 +20,17 @@
       out.exit_status.must_equal 0
     end
   end
-end
\ No newline at end of file
+
+  describe 'global attributes' do
```
clintoncwolfe: I'd like to see a few failed runs.  Behavior when validation fails? When an attribute specification is nonsensical (e.g. type: banana)?  Do we get a graceful exit, or a stacktrace?  Does the error message indicate the problem?

lib/inspec/rspec_example_groups.rb
```diff
@@ -0,0 +1,11 @@
+require 'inspec/attributes'
+require 'rspec/core/example_group'
+
+# This file allows you to add ExampleGroups to be used in rspec tests
+#
+class RSpec::Core::ExampleGroup
+  def attribute(name, options = {})
```
clintoncwolfe: A comment indicating why this is here would be helpful.

lib/inspec/rspec_example_groups.rb
```diff
@@ -0,0 +1,11 @@
+require 'inspec/attributes'
```
clintoncwolfe: Consider renaming this to be 'rspec_extensions' or similar.

Consider consolidating any other rspec modifications (whether extensions using their DSL (like this), or outright monkeypatching) into this file, if needed.

lib/inspec/profile_context.rb
```diff
@@ -28,12 +28,15 @@ def initialize(profile_id, backend, conf)
...
       @conf = conf.dup
+      @profile_name = @conf['profile'].profile_name || @profile_id if @conf['profile']
       @skip_only_if_eval = @conf['check_mode']
       @rules = {}
       @control_subcontexts = []
       @lib_subcontexts = []
       @require_loader = ::Inspec::RequireLoader.new
-      @attributes = []
+      @global_attributes = Inspec::Attributes.instance
+      @global_attributes.register_profile_override(@profile_id, @profile_name) if @profile_id != @profile_name
```
clintoncwolfe: No tests for this code.

lib/inspec/errors.rb
```diff
@@ -11,4 +11,5 @@ class UnsatisfiedVersionSpecification < Error; end
   class DuplicateDep < Error; end
   class FetcherFailure < Error; end
   class ReporterError < Error; end
+  class AttributeError < Error; end
```
clintoncwolfe: As mentioned elsewhere, a richer palette of errors here might be helpful.

lib/inspec/control_eval_context.rb
```diff
@@ -19,11 +19,16 @@ class ControlEvalContext
...
-    def self.rule_context(resources_dsl)
+    def self.rule_context(resources_dsl, profile_context_owner)
       require 'rspec/core/dsl'
       Class.new(Inspec::Rule) do
         include RSpec::Core::DSL
         with_resource_dsl resources_dsl
+
+        # allow attributes to be accessed within control blocks
+        define_method :attribute do |name, options = {}|
+          profile_context_owner.find_or_create_attribute(name, options)
```
clintoncwolfe: Find or create here makes me squeamish.  In a control, is it proper to _create_ an attribute? Shouldn't you only be reading values of already created attributes?

lib/inspec/control_eval_context.rb
```diff
@@ -137,8 +142,8 @@ def to_s
         end
 
         # method for attributes; import attribute handling
-        define_method :attribute do |name, options|
-          profile_context_owner.register_attribute(name, options)
+        define_method :attribute do |name, options = {}|
+          profile_context_owner.find_or_create_attribute(name, options)
```
clintoncwolfe: Same concern as above; in a test block, I'd imagine that you only want to access existing attributes.  If you're able to autovivify attributes, that will hide typos, etc.

clintoncwolfe (dismissed):

lib/inspec/objects/attribute.rb
jquick: This should only be a method scope disable.

jquick:
@clintoncwolfe - Thanks for the feedback. I think this is ready for another pass.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+      attribute.value.must_be_nil
     end
 
     it 'returns the user-configured default value if no value is assigned (false)' do
       attribute = Inspec::Attribute.new('test_attribute', default: false)
       attribute.value.must_equal false
     end
+
+    it 'returns value if overriding the default' do
+      attribute = Inspec::Attribute.new('test_attribute', required: 'default_value')
```
clintoncwolfe: `required: 'default_value'` ?

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+      attribute.value = 'test_value'
+      attribute.value.must_equal 'test_value'
+    end
+  end
```
clintoncwolfe: Is secrets_override removed?

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+    let(:attribute) { Inspec::Attribute.new('test_attribute', opts) }
+
+    it 'validates a string type' do
+      opts[:type] = 'string'
+      attribute.send(:validate_value_type, 'string')
+    end
+
+    it 'returns an error if a invalid string is set' do
+      opts[:type] = 'string'
+      e = proc { attribute.send(:validate_value_type, 123) }.must_raise(Inspec::AttributeValidationError)
```
clintoncwolfe: Opinion: 
```
ex = assert_raises(ExceptionClass) { dangerous_thing() }
```
reads waayyy better to me.  Not asking you to change away from must_*, just sayin' :)

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+      e.message.must_match /Attribute 'test_attribute' with value 'invalid' does not validate to type 'Hash'./
+    end
+
+    it 'validates a boolean type' do
+      opts[:type] = 'boolean'
+      attribute.send(:validate_value_type, false)
+      attribute.send(:validate_value_type, true)
+    end
+
+    it 'returns an error if a invalid hash is set' do
```
clintoncwolfe: Copypasta in test title

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+      opts[:type] = 'boolean'
+      e = proc { attribute.send(:validate_value_type, 'not_true') }.must_raise(Inspec::AttributeValidationError)
+      e.message.must_match /Attribute 'test_attribute' with value 'not_true' does not validate to type 'Boolean'./
+    end
+
+    it 'validates a boolean type' do
```
clintoncwolfe: copypasta in test title

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+      attribute.send(:validate_value_type, false)
+      attribute.send(:validate_value_type, true)
+      attribute.send(:validate_value_type, 1)
+      attribute.send(:validate_value_type, 'bob')
+    end
+  end
+
+  describe 'validate type method' do
+    it 'converts int to integer' do
+      attribute.send(:validate_type, 'regex').must_equal 'Regexp'
```
clintoncwolfe: Test title says integer, but we test regexes?

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+    end
+
+    it 'returns the same value if there is nothing to clean' do
+      attribute.send(:validate_type, 'String').must_equal 'String'
+    end
+
+    it 'returns an error if a invalid type is sent' do
+      e = proc { attribute.send(:validate_type, 'dressing') }.must_raise(Inspec::AttributeTypeError)
+      e.message.must_match /Type 'Dressing' is not a valid attribute type./
```
clintoncwolfe: It should be provided 'on the side'. :)

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+    end
+  end
+
+  describe 'valid_regexp? method' do
+    it 'vaildates a string regex' do
```
clintoncwolfe: validates, not vaildates.  Several copypastas here.

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+  describe 'valid_numeric? method' do
+    it 'vaildates a string number' do
+      attribute.send(:valid_numeric?, '123').must_equal true
+    end
+
+    it 'vaildates a float number' do
+      attribute.send(:valid_numeric?, 44.55).must_equal true
+    end
+
+    it 'validats a wong padded number' do
```
clintoncwolfe: wong?

test/unit/attribute_registry_test.rb
```diff
@@ -0,0 +1,71 @@
...
+      attributes.find_attribute('color', 'dummy_profile').value.must_equal 'black'
+    end
+
+    it 'errors when trying to find a attribute on a unknown profile' do
+      attribute = attributes.register_attribute('color', 'dummy_profile')
+      e = proc { attributes.find_attribute('color', 'unknown_profile') }.must_raise(Inspec::AttributeRegistryUnknownProfile) 
+      e.message.must_match "Profile 'unknown_profile' does not have any attributes"
+    end
+
+    it 'errors when trying to find a unkown attribute on a known profile' do
```
clintoncwolfe: unkNown

test/unit/attribute_registry_test.rb
```diff
@@ -0,0 +1,71 @@
...
+    it 'find a attribute which exist' do
+      attribute = attributes.register_attribute('color', 'dummy_profile')
+      attribute.value = 'black'
+
+      attributes.find_attribute('color', 'dummy_profile').value.must_equal 'black'
+    end
+
+    it 'errors when trying to find a attribute on a unknown profile' do
+      attribute = attributes.register_attribute('color', 'dummy_profile')
+      e = proc { attributes.find_attribute('color', 'unknown_profile') }.must_raise(Inspec::AttributeRegistryUnknownProfile)
```
clintoncwolfe: moderately fugly name, there

test/functional/attributes_test.rb
```diff
@@ -0,0 +1,56 @@
...
+      out.stdout.must_include '20 successful'
+      out.exit_status.must_equal 0
+    end
+
+    it "errors with invalid attribute types" do
+      cmd = 'exec '
+      cmd += File.join(profile_path, 'invalid_attributes')
+      cmd += ' --no-create-lockfile'
+      out = inspec(cmd)
+      out.stderr.must_include "Type 'Color' is not a valid attribute type.\n"
```
clintoncwolfe: Might include a stderr.must_not_include('some string that would appear in a stacktrace')

lib/inspec/rspec_extensions.rb
```diff
@@ -0,0 +1,12 @@
+require 'inspec/attribute_registry'
+require 'rspec/core/example_group'
+
+# This file allows you to add ExampleGroups to be used in rspec tests
+#
+class RSpec::Core::ExampleGroup
+  # this allows us to use the attribute method in inspec tests
```
clintoncwolfe: Comment at the level of intent, not implementation.  Yes, this will allow us to call a method you define on the next line :).

"This DSL method allows us to access the values of attributes within InSpec tests"

lib/inspec/control_eval_context.rb
```diff
@@ -19,11 +19,16 @@ class ControlEvalContext
...
     # @param [ResourcesDSL] resources_dsl which has all resources to attach
     # @return [RuleContext] the inner context of rules
-    def self.rule_context(resources_dsl)
+    def self.rule_context(resources_dsl, profile_id)
       require 'rspec/core/dsl'
       Class.new(Inspec::Rule) do
         include RSpec::Core::DSL
         with_resource_dsl resources_dsl
+
+        # allow attributes to be accessed within control blocks
```
clintoncwolfe, line 28: :100: for intent-level comment

test/unit/objects/attribute_test.rb
```diff
@@ -36,12 +36,175 @@
...
+
+    it 'does not error if a value is specified' do
+      attribute = Inspec::Attribute.new('test_attribute', required: true)
+      attribute.value = 'test_value'
+      attribute.value.must_equal 'test_value'
+    end
+
+    it 'returns an error if no value is set' do
+      attribute = Inspec::Attribute.new('test_attribute', required: true)
+      e = proc { attribute.value }.must_raise(Inspec::AttributeError)
```
clintoncwolfe: This will pass because
```
class AttributeRequiredError < AttributeError
```
but I think you mean to be looking for the more specific class.

lib/inspec/errors.rb
```diff
@@ -11,4 +11,13 @@ class UnsatisfiedVersionSpecification < Error; end
   class DuplicateDep < Error; end
   class FetcherFailure < Error; end
   class ReporterError < Error; end
+
+  class AttributeError < Error; end
```
clintoncwolfe: The namespaces here are bit ugly - you might consider building a tree under Inspec::Attribute::Error, and then Inspec::Attribute::RequiredError, or similar.  Exceptions should usually end with the word Error, and be grouped together in a namespace tree.  AttributeRegistry might get its own branch.

clintoncwolfe: Also, consider adding attr_accessors to some of these - most of them would do well to have a slot for the attribute name involved, or the value provided, or the profile name.  Or, YAGNI, and string matching against the error message is good enough.

docs/profiles.md
```diff
@@ -336,27 +337,80 @@ profile `my_dep` using the name `my_res2`.
 
 # Profile Attributes
 
-Attributes may be used in profiles to define secrets, such as user names and passwords, that should not otherwise be stored in plain-text in a cookbook. First specify a variable in the control for each secret, then add the secret to a YAML file located on the local machine, and then run `inspec exec` and specify the path to that Yaml file using the `--attrs` attribute.
+Attributes may be used in profiles to define secrets, such as user names and passwords, that should not otherwise be stored in plain-text in a cookbook. Attributes may be set for the whole profile in the `inspec.yml`.
```
clintoncwolfe: I know you're just building on what is already there, but this is too limiting of a description what attributes are for - they provide no facilities for secrets, but are frequently used to parameterize a profile for use in different environments / targets.

docs/profiles.md
```diff
@@ -336,27 +337,80 @@ profile `my_dep` using the name `my_res2`.
...
 
-For example, a control:
+Attributes may contain the following options:
 
+* Use `default` to set a default value for the attribute.
+* Use `type` to restrict an attribute to a specific type (any, string, numeric, array, hash, boolean, regex).
+* Use `required` to ensure the attribute has a default value or a value from a secrets file.
```
clintoncwolfe: s/ensure/insist/ ? mandate?

clintoncwolfe: "Secrets file?"

docs/profiles.md
```diff
@@ -336,27 +337,80 @@ profile `my_dep` using the name `my_res2`.
...
+    it { should eq 'bob' }
+  end
+
+  describe current_user do
+    it { should eq attribute('user') }
+  end
+end
+```
 
+For sensitive data it is recomended to use a secrets YAML file located on the local machine to populate the values of attributes. A secrets file will always overwrite a attributes default value. To use the secrets file run `inspec exec` and specify the path to that Yaml file using the `--attrs` attribute.
```
clintoncwolfe: If we are going to explain this use case, I think we should also explain the parameterization use case (which is different in motivation, but identical in mechanics).

lib/inspec/attribute_registry.rb
```diff
@@ -0,0 +1,71 @@
...
+    include Singleton
+    extend Forwardable
+
+    attr_reader :list
+    def_delegator :list, :each
+    def_delegator :list, :[]
+    def_delegator :list, :key?, :profile_exist?
+    def_delegator :list, :select
+
+    def self.find_attribute(name, profile)
```
clintoncwolfe: hrrrrmmmm  If this is a singleton, I'd expect to call these via 

```
Inspec::AttributeRegistry.instance.find_attribute(...)```

I suppose the class methods are just a convenience? Might add a comment to that effect.

File: lib/inspec/objects/attribute.rb [hunk 65]
```diff
@@ -71,5 +85,62 @@ def to_ruby
...
+      elsif type == 'Numeric'
+        invalid_type = true if !valid_numeric?(value)
+      elsif type == 'Boolean'
+        invalid_type = true if ![true, false].include?(value)
+      elsif value.is_a?(Module.const_get(type)) == false
+        invalid_type = true
+      end
+
+      raise Inspec::AttributeValidationError, "Attribute '#{@name}' with value '#{value}' does not validate to type '#{type}'." if invalid_type == true
+    end
```
clintoncwolfe: You *must* re-enable rubocop checks after a block.  Rubocop variances are FILE-scoped, not block-scoped.  This causes much gnashing of teeth and rending of clothes; see https://github.com/rubocop-hq/rubocop/issues/1690#issuecomment-286347494

clintoncwolfe (dismissed):
Great work! Mostly very minor typos, and copypasta.  Great first pass on docs, I had a bit more feedback there.  I think it will be mergeable tomorrow without a problem.

test/unit/objects/attribute_test.rb
jquick: Correct, it was not needed and confusing.

jquick:
@clintoncwolfe - Thanks for the feedback. This should be set.

clintoncwolfe (approved):
All set! Thanks so much for a great contribution, @jquick !

lib/inspec/control_eval_context.rb
```diff
@@ -36,9 +41,9 @@ def self.rule_context(resources_dsl)
     # @param outer_dsl [OuterDSLClass]
     # @return [ProfileContextClass]
     def self.create(profile_context, resources_dsl) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-      rule_class = rule_context(resources_dsl)
       profile_context_owner = profile_context
       profile_id = profile_context.profile_id
+      rule_class = rule_context(resources_dsl, profile_id)
 
       Class.new do # rubocop:disable Metrics/BlockLength
```
miah, line 48: Would it be possible to give this class a name?

jquick: We could probably pull this out in a refactor and move away from the dynamic. I think its out of scope for this PR since its not messing with that class though. Will put it in techDebt

miah (approved):
This is awesome. Thanks @jquick !

lib/inspec/control_eval_context.rb
miah: Thanks =)

aaronlippold:
👍 👏 @jquick et all great work! Thank you for all!!

Issue #3319, opened 2018-08-22
Allow target-id passthrough


jquick:
### 🎛 Description

🙋 feature request?

We have decided to move away from the Train::UUID generation in favor of a passthrough system. This means you will have to pass in a unique id for your target. When you run a inspec command and it will show up on the report at the end. 

We will want to allow both `cli` 
```
inspec exec my_profile --reporter json --target-id 1d3e399f-4d71-4863-ac54-84d437fbc444
```

and `--json-config`
```
inspec exec my_profile --reporter json --json-config config.json

{
    "target_id": "1d3e399f-4d71-4863-ac54-84d437fbc444"
}
```

Along with this we will be removing the call to generate a UUID from Train.

PR #3320, opened 2018-08-22
Allow target-id passthrough

jquick:

This adds a new CLI option to pass through a UUID. This will show up on the output reports accordingly. Both `--target-id` and `--json-config` are supported. 

Fixes https://github.com/inspec/inspec/issues/3319

### DIFF
--- a/test/functional/inspec_exec_test.rb
+++ b/test/functional/inspec_exec_test.rb
@@ -46,6 +46,15 @@ Test Summary: 0 successful, 0 failures, 0 skipped
     File.stat("#{outpath}/foo/bar/test.json").size.must_be :>, 0
   end
 
+  it 'can execute the profile with a uuid passthrough' do
+    outpath = Dir.tmpdir
+    out = inspec("exec #{example_profile} --no-create-lockfile --uuid 1d3e399f-4d71-4863-ac54-84d437fbc444")
+    out.stderr.must_equal ''
+    out.exit_status.must_equal 101
+    stdout = out.stdout.force_encoding(Encoding::UTF_8)
+    stdout.must_include "UUID:    1d3e399f-4d71-4863-ac54-84d437fbc444"
+  end
+
   it 'executes a metadata-only profile' do
     out = inspec('exec ' + File.join(profile_path, 'complete-metadata') + ' --no-create-lockfile')
     out.stderr.must_equal ''
--- a/test/functional/inspec_exec_test.rb
+++ b/test/functional/inspec_exec_test.rb
@@ -52,7 +52,7 @@ Test Summary: 0 successful, 0 failures, 0 skipped
     out.stderr.must_equal ''
     out.exit_status.must_equal 101
     stdout = out.stdout.force_encoding(Encoding::UTF_8)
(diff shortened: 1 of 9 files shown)


jquick:
Talked with @arlimus. We may want to change this to `transport-id`/`target-id` or something more generic so we are not forcing people to use a `UUID`. If we did this we would keep the private `node_uuid` for the A2 reporter.

aaronlippold:
Hi, I agree it should be generic but I am not sure 'transmission-id`
communicates well or at least clearly. I would suggest that having `uuid`
as part of the automatically generated run / results data is still
something we should have. Users can or not use it as they like but it does
help fill a data and alignment gap and provide a standardized identifier.
--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

test/unit/base_cli_test.rb
```diff
@@ -123,6 +123,13 @@
       parsed.must_equal assert
     end
 
+    it 'parses cli report and attaches target_id' do
+      opts = { 'reporter' => ['cli'], 'target_id' => '1d3e399f-4d71-4863-ac54-84d437fbc444' }
+      parsed = Inspec::BaseCLI.parse_reporters(opts)
+      assert = {"reporter"=>{"cli"=>{"stdout"=>true, "target_id"=>"1d3e399f-4d71-4863-ac54-84d437fbc444"}}, "target_id"=>"1d3e399f-4d71-4863-ac54-84d437fbc444"}
```
clintoncwolfe: I think "expected_value" might be a better word than 'assert', here.

clintoncwolfe: I think "expected_value" might be a better word than 'assert', here.

test/functional/inspec_exec_test.rb
```diff
@@ -46,6 +46,15 @@
     File.stat("#{outpath}/foo/bar/test.json").size.must_be :>, 0
   end
 
+  it 'can execute the profile with a target_id passthrough' do
+    outpath = Dir.tmpdir
+    out = inspec("exec #{example_profile} --no-create-lockfile --target-id 1d3e399f-4d71-4863-ac54-84d437fbc444")
+    out.stderr.must_equal ''
+    out.exit_status.must_equal 101
+    stdout = out.stdout.force_encoding(Encoding::UTF_8)
+    stdout.must_include "ID:      1d3e399f-4d71-4863-ac54-84d437fbc444"
```
clintoncwolfe: Minor gripe, but if we're asking for the user to pass this using the `--target-id` option, I think we should consistently refer to it as a *target* id.  We may have other ID's in the future - report IDs, etc.

clintoncwolfe (approved):
Approved with one minor gripe, and one pedantic gripe.  Great work!

test/functional/inspec_exec_test.rb
jquick: Actually I think I may leave this for now. The other formatting options look bad to me:

jerryaldrichiii (approved):
Looks great!

PR #3321, opened 2018-08-22
Fix for custom controls

frezbo:
We have a custom control named lets say `custom_control` which we used in our  inspec code. It is created by extending the `::Inspec::DSL` module. After https://github.com/inspec/inspec/pull/3267 was merged it broke our pipelines. I identified that https://github.com/inspec/inspec/commit/3a9ed68c729311c5b06c2ddbdb28f2d2cd41e0fa#diff-7116ea72e5d980e406887620c1facc6cR179 has changed and was returning a hash now instead of a boolean. So in our custom control it was returning something like `{:result=>{:result=>true, :message=>nil}, :message=>nil}` instead of `{:result=>true, :message=>nil}`. I have fixed this by reverting this to just the value. I tested with a standard control and our custom control and it seems to be working. @jerryaldrichiii @miah @jquick What do you think?

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -228,7 +228,7 @@ module Inspec
       sc = checks(src)
       dst.instance_variable_set(:@__checks, sc) unless sc.empty?
       sr = skip_status(src)
-      set_skip_rule(dst, sr) unless sr.nil?
+      set_skip_rule(dst, sr[:result] || sr) unless sr.nil?
 
       # Save merge history
       dst.instance_variable_set(:@__merge_count, merge_count(dst) + 1)


jerryaldrichiii:
I'm not following the logic of reverting in InSpec @frezbo. We added that so that users can set a message so they can see more info other than "skipped due to only_if". I'm not sure what path is leading to `value` being a Hash though. I don't see a change that would cause that behavior.

Also, regarding extending `::Inspec::DSL` I'm not sure if we support that, but I'll bring it up to the team. If so, we might revert this PR and add the feature in InSpec 3.0 since it would be a breaking change.

frezbo:
@jerryaldrichiii Yes you are right about `value`. I have been looking at code wrong. Guess I need a coffee. It's real late here :D. Anyway I have now updated PR with only the code changes required to show the skip message. Other changes seemed un-necessary. I have run the `functional` test suite also locally and seems all are passing.

frezbo:
Closing this PR as the bug is in the extended DSL resource. The resource was getting instantiated twice.

Issue #3322, opened 2018-08-23
Inspec merger fails with `only_if` after v2.2.61


frezbo:
### 🎛 Description

🐛 bug report

Inspec merger at: https://github.com/inspec/inspec/blob/master/lib/inspec/rule.rb#L207 merger controls with same name. After https://github.com/inspec/inspec/pull/3267, if we have two controls with same name the `rule.instance_variable_get(:@__skip_rule)` get updated at each merge, thus further failing the any `only_if`'s in the controls. This will also create looped data for the `skip`. Ref: https://github.com/inspec/inspec/pull/3321

### 🌍 InSpec and Platform Version

Inspec version above `v2.2.61`

Caused by: https://github.com/inspec/inspec/pull/3267

### 🤔 Replication Case

**Inspec execution o/p**

```
inspec (aws:rean-gov-sd)(kc)(git:master)$ git checkout master 
Already on 'master'
Your branch is up to date with 'origin/master'.
inspec (aws:rean-gov-sd)(kc)(git:master)$ git checkout v2.2.61
Note: checking out 'v2.2.61'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at a83ae6ac Bump version to 2.2.61 by Chef Expeditor
inspec (aws:rean-gov-sd)(kc)(git:detached)$ bundle exec bin/inspec exec /home/frezbo/git/work/ruby/inspec_test/t.rb 

Profile: tests from /home/frezbo/git/work/ruby/inspec_test/t.rb (tests from .home.frezbo.git.work.ruby.inspec_test.t.rb)
Version: (not specified)
Target:  local://

  ↺  dummy: Operating System Detection
     ↺  Skipped control due to only_if condition.


Profile Summary: 0 successful controls, 0 control failures, 1 control skipped
Test Summary: 0 successful, 0 failures, 1 skipped
inspec (aws:rean-gov-sd)(kc)(git:detached)$ git checkout v2.2.62
Previous HEAD position was a83ae6ac Bump version to 2.2.61 by Chef Expeditor
HEAD is now at f18bcc34 Bump version to 2.2.62 by Chef Expeditor
[...]

PR #3323, opened 2018-08-23
Fix skip hash being passed instead of boolean value

frezbo:
Fixes: https://github.com/inspec/inspec/issues/3322


frezbo:
@jerryaldrichiii Seems I was right about the PR yesterday, took me some time to figure out the issue.

jquick (approved):
Thanks for figuring out this but @frezbo !

jquick:
This is a regression from the skip changes. I will add some testing for this after this is merged in.

jerryaldrichiii (approved):
Great catch @frezbo!

PR #3332, opened 2018-08-28
Remove skip from inherited profiles from showing up in reporting (breaking change)

miah:
I've made a few changes to allow inherited profiles to be skipped when they are loaded on unsupported platforms. We will now also output a WARN to STDERR detailing which profile was skipped:

```
[2018-08-28T14:36:37-07:00] WARN: Skipping profile: 'windows-only' on unsupported platform: 'mac_os_x'.
```

Please let me know how this implementation looks and if I missed anything.

ref: https://github.com/inspec/inspec/issues/3158

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -548,10 +548,12 @@ module Inspec
       params[:controls] = controls = {}
       params[:groups] = groups = {}
       prefix = @source_reader.target.prefix || ''
-      tests.each do |rule|
-        next if rule.nil?
-        f = load_rule_filepath(prefix, rule)
-        load_rule(rule, f, controls, groups)
+      unless tests.nil?
+        tests.each do |rule|
+          next if rule.nil?
+          f = load_rule_filepath(prefix, rule)
+          load_rule(rule, f, controls, groups)
+        end
       end
       params[:attributes] = @runner_context.attributes
       params
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -157,10 +157,16 @@ module Inspec
       supports_platform? && supports_runtime?
     end
 
+    # We need to check if we're using a Mock'd backend for tests to function.
+    # @returns [TrueClass, FalseClass]
     def supports_platform?
       if @supports_platform.nil?
         @supports_platform = metadata.supports_platform?(@backend)
       end
+      if @backend.backend.class.to_s == 'Train::Transports::Mock::Connection'
+        @supports_platform = true
+      end
+
       @supports_platform
     end
 
@@ -176,10 +182,11 @@ module Inspec
     end
 
     def collect_tests(include_list = @controls)
-      if !@tests_collected
+      unless @tests_collected
         locked_dependencies.each(&:collect_tests)
 
         tests.each do |path, content|
+          next unless supports_platform?
           next if content.nil? || content.empty?
           abs_path = source_reader.target.abs_path(path)
           @runner_context.load_control_file(content, abs_path, nil)
@@ -221,6 +228,7 @@ module Inspec
       return @runner_context if @libraries_loaded
 
       locked_dependencies.each do |d|
+        next unless d.supports_platform?
         c = d.load_libraries
         @runner_context.add_resources(c)
       end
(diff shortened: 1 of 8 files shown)


frezbo:
what about profiles that supports, let's say `windows` and `linux`, does it get skipped, let me see If I can test this PR, might have to allocate some free time.

jquick:
@frezbo as long as one of the supports lines returns true it should still run. So in your case as long as you are on `windows` OR `linux` you will not skip.

lib/inspec/profile.rb
```diff
@@ -162,10 +168,11 @@ def params
     end
 
     def collect_tests(include_list = @controls)
-      if !@tests_collected
+      unless @tests_collected
         locked_dependencies.each(&:collect_tests)
 
         tests.each do |path, content|
+          next unless supports_platform?
```
jquick: I think you could put this before the tests array like:
```
return unless supports_platform?

tests.each do |path, content|
...
```
This will keep it from being called for every single test

jquick (commented):
Looking really good @miah !

lib/inspec/profile.rb
miah: Oh right. Thank you =)

test/unit/profiles/profile_test.rb
```diff
@@ -99,6 +99,15 @@
     end
   end
 
+  describe 'skips loading on unsupported platform' do
+    let(:profile_id) { 'windows-only' }
```
clintoncwolfe, line 114: If you only use it once, my preference would be to just inline this string, and not use `let`.  Not a blocker.

clintoncwolfe, line 114: I think we should add a test reflecting @frezbo's concern - a run on a linux platform should not skip a profile that supports windows and linux.  That may already be covered in another test - if so, link it and I'll be quiet :)

test/functional/inspec_exec_test.rb
```diff
@@ -301,7 +301,8 @@
   describe 'when a dependency does not support our backend platform' do
     it 'skips the controls from that profile' do
       out = inspec("exec #{File.join(profile_path, 'profile-support-skip')} --no-create-lockfile")
-      out.stdout.force_encoding(Encoding::UTF_8).must_include "Profile Summary: 0 successful controls, 0 control failures, \e[38;5;247m2 controls skipped\e[0m\n"
+      out.stdout.force_encoding(Encoding::UTF_8).must_include "WARN: Skipping profile: 'windows-only' on unsupported platform"
```
clintoncwolfe: Personally, I prefer error messages that are telling me I did something wrong, to give hints as to how to do it right; here, you could add info as to the current platform (platform lineage?), and a list of the supported platforms that the profile supports.

EDIT: ahh, I see you do include the backend platform.  Mentioning the profile platform might be nice, though.  Not a blocker.

lib/inspec/profile.rb
```diff
@@ -162,7 +168,8 @@ def params
     end
 
     def collect_tests(include_list = @controls)
-      if !@tests_collected
+      unless @tests_collected
```
clintoncwolfe, line 185: Hrm.  How would we get here, if we're on an unsupported platform?  I thought the code in runner.rb would catch the unsupported platform, issue a warning, and skip adding the profile to the test collector.  Is there another way to get here?

clintoncwolfe (commented):
No major blockers, just some clarifications and one desired extra test.  Looks great overall - great contribution!

test/unit/profiles/profile_test.rb
miah: I'll add a example profile to our mocks that supports bsd, linux, and windows.

jerryaldrichiii (approved):
Looks good to me!

lib/inspec/profile.rb
jquick: I think we need this as the runner only passes in the parent but this will call itself for each dependent.

jquick (approved):
Thanks @miah

jquick:
Merging this, we can test off inspec-3.0 branch if needed.

Issue #3144, opened 2018-06-14
Harmonize platform supports output


chris-rock:
### 🎛 Description

When we run `inspec json examples/profile` we get the following output

```
{
  "name": "profile",
  "title": "InSpec Example Profile",
  "maintainer": "Chef Software, Inc.",
  "copyright": "Chef Software, Inc.",
  "copyright_email": "support@chef.io",
  "license": "Apache-2.0",
  "summary": "Demonstrates the use of InSpec Compliance Profile",
  "version": "1.0.0",
  "supports": [
    {
      "os-family": "unix"
    }
  ],
```

Going forward, we should map `os-family` to `platform-familiy` and deprecate the use of `os-family`. We should not remove it though, since it would break users. The specification of all fields is included in section `The platform resource` of https://github.com/inspec/inspec/issues/1661.

https://github.com/inspec/inspec/issues/2488 is an related issue

### 🌍 InSpec and Platform Version

InSpec 2.x

### 🤔 Replication Case

see above

### 💁 Possible Solutions

map `os-*` to `platform-*`

### 💻 Stacktrace

n/a

PR #3333, opened 2018-08-29
Convert legacy supports to their platform counterparts

jquick:

This change will update `os-name` and `os-family` when outputting to reports or json profile extract.

Fixes https://github.com/inspec/inspec/issues/3144


miah (approved):
Thanks @jquick

PR #3335, opened 2018-08-29
Fix the brew command to install inspec

tas50:
InSpec was pulled from the casks a while ago. We have our own setup now,
but the command to pull from there is a bit different. Also OSX should
be macOS.


jquick (approved):
Thanks @tas50

Issue #3315, opened 2018-08-22
Broken Link: https://www.inspec.io/docs/reference/resources/azure_generic_resource/


kmf:
The Link to 

`Inspec Integration Tests for Azure Generic Resources` is broken

PR #3336, opened 2018-08-29
Fix documentation link for azure integration tests.

miah:
Fixes https://github.com/inspec/inspec/issues/3315


jquick (approved):
Thanks @miah

PR #3354, opened 2018-09-05
Allow basic retries

jorisd:
This would cover case where we get failure like this : 

RuntimeError:
       Failed to open TCP connection to login.microsoftonline.com:443 (getaddrinfo: Temporary failure in name resolution)

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -169,9 +169,19 @@ class Inspec::InspecCLI < Inspec::BaseCLI
 
     runner = Inspec::Runner.new(o)
     targets.each { |target| runner.add_target(target) }
-
-    exit runner.run
-  rescue ArgumentError, RuntimeError, Train::UserError => e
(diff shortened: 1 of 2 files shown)


jquick:
Hey @jorisd, thanks for your first contribution!

Im not sure if this belongs here in the global runner for all transports. I wonder if we should put something in the Train TCP/SSH connection instead. I will talk with the team and see what we want to do here.

clintoncwolfe:
I agree with @jquick - this is a greatly needed feature, but this is too broad of a place to put it.  It could be an option on individual resources (for example, a postgresql test might want a special retry value or timeout value); or on specific transports.  I'd also want to make the exception handling much more specific, both in terms of what is caught, and what we re-raise.

jorisd:
Hi @jquick @clintoncwolfe , thanks for the feedback, that makes sense.

I will close this PR and take your comments into account to improve that.

Issue #3357, opened 2018-09-06
Inspec Archive produces different checksums on each run


frezbo:
### 🎛 Description

`inspec archive` produces different checksum on each run even when the profile has not changed.

**Choose one:** 🐛 bug report

### 🌍 InSpec and Platform Version

`2.2.78`

### 🤔 Replication Case

```
~ (aws:rean-product)(kc)$ for i in {1..5}; do inspec archive --overwrite /home/frezbo/git/work/inspec/dummy_profile && sha256sum dummy_profile.tar.gz; done
I, [2018-09-06T00:20:17.986967 #11270]  INFO -- : Checking profile in /home/frezbo/git/work/inspec/dummy_profile
I, [2018-09-06T00:20:17.987225 #11270]  INFO -- : Metadata OK.
I, [2018-09-06T00:20:20.372446 #11270]  INFO -- : Found 240 controls.
I, [2018-09-06T00:20:20.372982 #11270]  INFO -- : Control definitions OK.
I, [2018-09-06T00:20:20.373279 #11270]  INFO -- : Generate archive /home/frezbo/dummy_profile.tar.gz.
I, [2018-09-06T00:20:20.429334 #11270]  INFO -- : Finished archive generation.
cf4d60cba255f903d3feefa18a9ed4c979a8c8fffc675b5f76c10463d65c4a60  dummy_profile.tar.gz
I, [2018-09-06T00:20:22.517429 #11279]  INFO -- : Checking profile in /home/frezbo/git/work/inspec/dummy_profile
I, [2018-09-06T00:20:22.517674 #11279]  INFO -- : Metadata OK.
I, [2018-09-06T00:20:25.463056 #11279]  INFO -- : Found 240 controls.
I, [2018-09-06T00:20:25.463647 #11279]  INFO -- : Control definitions OK.
I, [2018-09-06T00:20:25.463921 #11279]  INFO -- : Generate archive /home/frezbo/dummy_profile.tar.gz.
I, [2018-09-06T00:20:25.536955 #11279]  INFO -- : Finished archive generation.
0a079f6aeb5157cea22c648c685557e52e498e70b9006a07744b03dd04a4e40d  dummy_profile.tar.gz
I, [2018-09-06T00:20:27.291843 #11290]  INFO -- : Checking profile in /home/frezbo/git/work/inspec/dummy_profile
I, [2018-09-06T00:20:27.292028 #11290]  INFO -- : Metadata OK.
[...]

PR #3358, opened 2018-09-06
Remove the `.git` directory after vendor

frezbo:
Fixes: https://github.com/inspec/inspec/issues/3357

### DIFF
--- a/lib/fetchers/git.rb
+++ b/lib/fetchers/git.rb
@@ -50,6 +50,7 @@ module Fetchers
         Dir.mktmpdir do |tmpdir|
           checkout(tmpdir)
           Inspec::Log.debug("Checkout of #{resolved_ref} successful. Moving checkout to #{dir}")
+          FileUtils.rm_rf("#{tmpdir}/.git")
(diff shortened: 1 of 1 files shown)


clintoncwolfe (changes requested):
Thanks so much for this contribution!  I agree we don't want to ship the .git directory in a vendored profile.  However, I think this is too broad.  It will remove the .git directory on any command, not just `vendor` (though we could make it conditional on that).  Deleting the .git directory may cause problems with repeated runs - see line 47-48, for example.

I'd at least like to see a functional test in which a git fetcher is repeatedly run, to make sure this code doesn't cause problems.

An alternate solution would place this deletion code directly in the vendoring, and perform the deletion after the copy.... though that would likely break checksumming, which was the point in the first place.

frezbo:
I will hold on to this until @jquick comes back with more findings.

jquick:
I wont have time to dedicate to this for awhile. Lets do some brain storming on the linked issue and find the best solution here. Closing for now.

Issue #562, opened 2016-03-19
Use text values for `severity`


chris-rock:
As proposed by @aaronlippold we use human-readable names for `impact`. At the moment a control impact is defined by as following:

```
  control 'my-test-1' do
    impact 1.0
    title '...'
  end
```

It would be nice to use words for it like:
- High, Medium, Low, Critical
- crit, high, med, low

We need to ensure:
- all human-readable numbers are mapped to a number
- the number scheme is well defined (e.g. CVSS score as we use it at the moment)

trevor-vaughan:
I would like to note that having a numeric impact allows for calculation and this is a *good thing*.

The tag for severity should suffice.

chris-rock:
We should use the CVSS 3.0 mapping in inspec https://www.first.org/cvss/specification-document

<img width="821" alt="image" src="https://user-images.githubusercontent.com/1178413/43704538-30d0b722-9960-11e8-9b4c-9a4657bda7b4.png">

trevor-vaughan:
Should there be general discussion around how to adjust the impact based on the environment?

Also, is it currently possible to adjust the impact programmatically?

For instance, I would like to be able to lower the impact of some items if `selinux` is enforcing and active on a daemon.

chris-rock:
You can inherit a control and adapt its impact. The challenge with adjusting the impact per environment is that its reporting may be difficult, because right now we assume that the impact is static for a control.

If I understand you correctly, you essentially want to return a score based on your  findings. Is that correct?

trevor-vaughan:
Yes, this is exactly what I want.

I also want to be able to adjust the score on individual items based on
inspection of the environment at run time to alleviate work on the system
maintainers.

Trevor

PR #3359, opened 2018-09-06
Add string impact options for controls

jquick:
This PR adds the option to specify:
```
none
low
medium
high
critical
```

For your impact section in controls. This will map to their CVSS equivalent number values. This is to help with readability on controls.

Fixes https://github.com/inspec/inspec/issues/562


miah (approved):

test/unit/impact_test.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    it 'returns true for float string' do
+      impact.is_number?('10.99').must_equal true
+    end
+
+    it 'returns false for char string' do
+      impact.is_number?('abc').must_equal false
+    end
+  end
+
```
jerryaldrichiii: Extra newline, but won't hold up the PR for it.

/me shakes fist at tests not being linted

lib/inspec/impact.rb
```diff
@@ -0,0 +1,34 @@
...
+  def self.is_number?(value)
+    Float(value)
+    true
+  rescue
+    false
+  end
+
+  def self.string_from_impact(value)
+    value = value.to_f
+    raise Inspec::ImpactError, "'#{value}' is not a valid impact score [0.0 - 1]." if value < 0 || value > 1
```
jerryaldrichiii: Should this be `[0.0 - 1.0]`?

lib/inspec/impact.rb
```diff
@@ -0,0 +1,34 @@
...
+  }.freeze
+
+  def self.impact_from_string(value)
+    # return if its a number
+    return value if is_number?(value)
+    raise Inspec::ImpactError, "'#{value}' is not a valid impact name [none, low, medium, high, critical]." unless IMPACT_SCORES.key?(value.downcase)
+    IMPACT_SCORES[value]
+  end
+
+  def self.is_number?(value)
```
jerryaldrichiii: You could probably shorten this to:

```
value.is_a?(Integer) || value.is_a?(Float)
```

jerryaldrichiii: Oh...I see...looks like you are attempting to cast as a Float then relying on that not erroring? I'm not following.

lib/inspec/impact.rb
```diff
@@ -0,0 +1,34 @@
...
+    'low'      => 0.01,
+    'medium'   => 0.4,
+    'high'     => 0.7,
+    'critical' => 0.9,
+  }.freeze
+
+  def self.impact_from_string(value)
+    # return if its a number
+    return value if is_number?(value)
+    raise Inspec::ImpactError, "'#{value}' is not a valid impact name [none, low, medium, high, critical]." unless IMPACT_SCORES.key?(value.downcase)
```
jerryaldrichiii: I'd like to see something like:

```
'#{value}' is not a valid impact name. Options: none, low, medium, high, critical
```

jerryaldrichiii (commented):
Couple of suggestions/questions. Won't hold up the PR though.

lib/inspec/impact.rb
jquick: Right, if we can cast the string as a Float then its some sort of number and we want to leave it alone.

jquick: Take a look at the test to see how I am using this.

jerryaldrichiii: Hmm...I understand now. Thanks!

aaronlippold:
Hi,

Low is not cast to the correct value is it? 'low' => 0.01, should be 'low'
=> 0.1 right?

Also, what about the values in between? .2, .3, .5, .6 ...

Yours,

Aaron
--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

jquick:
Hey @aaronlippold . Low should be correct at `.01` please take a look at https://www.first.org/cvss/specification-document where it shows the breakdown. values in between will fall into the bracket. So .2/.3 would be low, etc. You can see this here https://github.com/inspec/inspec/blob/c1dfd08d759ac14ac6f0100c765bdc40ec9e2aaa/lib/inspec/impact.rb#L31

aaronlippold:
Hi,

OK I see why you are using .01 i.e. to provide the lower bound... got it.
--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

jerryaldrichiii (approved):

PR #3360, opened 2018-09-06
handle errors from automate report and display them to the user

chris-rock:
I misconfigured the automate reporting url and it took me 30 min to figure out why it was not reported to Automate. I added a check that verifies that the report post was successful.


jquick (approved):
Thanks @chris-rock

PR #3362, opened 2018-09-06
✓ adds additional checks for vendored profiles

chris-rock:
This PR adds additional checks to verify if:
- inspec.yml and inspec.lock dependencies are out-of-sync
- complains about a missing lock file if we have dependencies defined

This also mitigates the issue where users are trying to upload un-vendored profiles to Automate. If you agree with the approach, I am happy to add functional tests for this PR.

* [x] functional tests


arlimus:
Love the direction of this :+1: prevents some accidental misuse

test/functional/inspec_check_test.rb
```diff
@@ -65,4 +65,29 @@
...
+      out = inspec('check ' + File.join(profile_path, 'profile-lock-insync'))
+      out.exit_status.must_equal 0
+    end
+
+    it 'can check a profile where lock file and inspec.yml are in not synnc' do
+      out = inspec('check ' + File.join(profile_path, 'profile-lock-outofsync'))
+      out.exit_status.must_equal 1
+      out.stdout.must_include 'inspec.yml and inspec.lock are out-of-sync. Please re-vendor with `inspec vendor`.'
+      out.stdout.must_include 'Cannot find linux-baseline in lockfile. Please re-vendor with `inspec vendor`.'
+    end
```
clintoncwolfe, line 91: :100: great functional tests.  After reading the PR description, the next thing I read is the tests, to see if the author is proving what they say they are doing.

And 💯 on having good test fixture profiles.

lib/inspec/profile.rb
```diff
@@ -316,6 +316,25 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
       m_unsupported.each { |u| warn.call(meta_path, 0, 0, nil, "doesn't support: #{u}") }
       @logger.info 'Metadata OK.' if m_errors.empty? && m_unsupported.empty?
 
+      # verify that a lockfile is present if we have dependencies
+      if !metadata.dependencies.empty?
+        error.call(meta_path, 0, 0, nil, 'Profile need to be vendored with `inspec vendor`.') if !lockfile_exists?
```
clintoncwolfe: Grammar error here:

'Your profile needs to be vendored with `inspec vendor`.'

(need vs needs, and missing article).

Also impacts the functional test.

lib/inspec/profile.rb
```diff
@@ -316,6 +316,25 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
       m_unsupported.each { |u| warn.call(meta_path, 0, 0, nil, "doesn't support: #{u}") }
       @logger.info 'Metadata OK.' if m_errors.empty? && m_unsupported.empty?
 
+      # verify that a lockfile is present if we have dependencies
```
clintoncwolfe: I like the implementation, but I think it is in the wrong class.  We have a class for the lockfile, and one for the metadata.  Why is it Profile's job to validate them?  I could see `metadata.check_for_stale_lockfile(lockfile)` or `lockfile.is_current?(metadata)` or similar.  

It's a matter of opinion; but Profile does too many things already, I think.

lib/inspec/profile.rb
```diff
@@ -316,6 +316,25 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
...
+      end
+
+      if lockfile_exists?
+        # verify if metadata and lockfile are out of sync
+        if lockfile.deps.size != metadata.dependencies.size
+          error.call(meta_path, 0, 0, nil, 'inspec.yml and inspec.lock are out-of-sync. Please re-vendor with `inspec vendor`.')
+        end
+
+        metadata.dependencies.each { |dep|
+          # TODO: should we also verify that the soure is the same?
```
clintoncwolfe: And what about version constraints?

clintoncwolfe (dismissed):
One small grammo, and a preference to move the implementation to another class; but otherwise straightforward, with excellent testing.  Great contribution!

lib/inspec/profile.rb
chris-rock: I was thinking about this as well. I was thinking that we may need to externalize all checks instead of keeping them in profiles. I am not sure that we should put the check logic into lockfile or metadata either since checks should be completely independent. Maybe checks should be an inspec profile that checks inspec profiles?

lib/inspec/profile.rb
chris-rock: Indeed. I left that open since it is more complicated. We should cover it, but I am not able to cover it as part of this PR.

jquick (approved):
Thanks @chris-rock

miah (approved):
Thanks @chris-rock !

jerryaldrichiii (approved):

lib/inspec/profile.rb
```diff
@@ -345,6 +346,28 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
...
+
+        if lockfile_exists?
+          # verify if metadata and lockfile are out of sync
+          if lockfile.deps.size != metadata.dependencies.size
+            error.call(meta_path, 0, 0, nil, 'inspec.yml and inspec.lock are out-of-sync. Please re-vendor with `inspec vendor`.')
+          end
+
+          metadata.dependencies.each { |dep|
+            # TODO: should we also verify that the soure is the same?
+            if !lockfile.deps.map { |x| x[:name] }.include? dep[:name]
```
jquick, line 396: Currently this breaks check for profiles where the inspec.yml does not specify a name like: https://github.com/inspec/inspec/blob/master/examples/meta-profile/inspec.yml#L11

I am going to only enable this check for defined names in the depends.

PR #3366, opened 2018-09-07
do not show success message since its confusing

chris-rock:
Turns the output from
screen shot 2018-09-07 at 12 35 45 pm
to
screen shot 2018-09-07 at 12 35 57 pm


jquick (approved):
Thanks @chris-rock

clintoncwolfe (approved):
Looks good, thanks!

PR #3370, opened 2018-09-07
ensure we use the mock backend when we upload profiles

chris-rock:
The compliance upload has not used the mock backend. This is especially recognized due to long run-times on Windows if wmi resources have been used. We should never execute a profile against the local target for upload.

It also catches an issue where generated profiles are not deleted if they've been generated


jquick (approved):
Nice addition.

clintoncwolfe (approved):
Refs #3373.  Code seems reasonable, but currently untestable.

PR #3371, opened 2018-09-07
Further templatize the inspec.yml with additional profile options

mattray:

### DIFF
--- a/lib/bundles/inspec-init/cli.rb
+++ b/lib/bundles/inspec-init/cli.rb
@@ -25,8 +25,40 @@ module Init
 
       # register command for the template
       desc "#{template_name} NAME", "Create a new #{template_name}"
-      option :overwrite, type: :boolean, default: false,
-        desc: 'Overwrites existing directory'
+
+      option :copyright,
+             desc: 'Name of the copyright holder',
+             default: 'The Authors'
+
+      option :email,
+             desc: 'Email address of the author',
+             default: 'you@example.com'
+
+      option :license,
+             desc: 'License for the profile',
+             default: 'Apache-2.0'
+
+      option :maintainer,
+             desc: 'Name of the copyright holder',
+             default: 'The Authors'
+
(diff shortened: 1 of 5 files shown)


chris-rock:
@mattray Can you share the usecase for that behaviour?

mattray:
1) When you do a "chef generate cookbook" there are many CLI options available to pre-populate the metadata.rb, "inspec init profile" should have similar functionality.
2) inspec-iggy is going to generate full compliance profiles, I'd rather use the existing profile ERBs from "inspec init" than come up with a duplicate set of templates.

chris-rock:
I agree that it makes sense for Chef since a ruby file will be generated. I am not sure about InSpec since yml can easily be edited and generated from nearly every language. Would love to get other opinions.

For iggy, I would prefer if we continue to use InSpec objects located in https://github.com/inspec/inspec/tree/master/lib/inspec/objects. We do not have a profile and a metadata object yet. I would prefer if we are creating those and rewrite the generator to use inspec objects instead. I would advise against going down the route to reuse the `inspec init` command in iggy.

mattray:
I submitted a PR to make the 'inspec init profile' a reusable component in July and @clintoncwolfe rewrote it in https://github.com/inspec/inspec/commit/02cb93ec2354fd5e678f939024b2c2b76da0d257. I was going to continue to use the InSpec objects for the controls, are there immediate plans to add profiles and metadata objects? I can pull the code back into inspec-iggy if necessary.

mattray:
I've gone ahead and implemented it without inspec-init in inspec-iggy, so I no longer will be using this functionality. A proper InSpec object would be nice, it would allow some validation. If you want the functionality for "inspec init profile" you may merge it, if not go ahead and close it.

jquick:
Yeah, I feel like this may be CLI bloat. I think keeping it as YML for now sounds good. Thanks for the suggestion and concept!

PR #3374, opened 2018-09-09
Fixing AWS integration tests.

MartinLogan:
Fixed some issues in the AWS default integration tests.

These should be obvious, but if there are any questions let me know.


jquick (approved):
Thanks for the fix @MartinLogan

miah (approved):

PR #3383, opened 2018-09-11
Added db_name flag

kdoores:
Signed-off-by: Kayleigh <kayleigh.doores@gmail.com>

By default SQLCMD queries against the default dabase that is assigned to that users login.  Adding the `-d` flag allows for specifying a database outside of the default scope, that a user has access to.  I added `db_name` as a flag for the resource, and an example so that users know that it's there.

### DIFF
--- a/docs/resources/mssql_session.md.erb
+++ b/docs/resources/mssql_session.md.erb
@@ -62,6 +62,14 @@ The following examples show how to use this InSpec audit resource.
     describe sql.query("SELECT SERVERPROPERTY('ProductVersion') as result").row(0).column('result') do
       its("value") { should cmp > '12.00.4457' }
     end
+  
+### Test a specific database
+
+    sql = mssql_session(user: 'my_user', password: 'password', db_name: 'test')
+
+    describe sql.query("SELECT Name AS result FROM Product WHERE ProductID == 1").row(0).column('result') do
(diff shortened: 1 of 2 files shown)


jquick:
Hey @kdoores ! Thanks for this PR. This is a nice addition. We will need to fix the Rubocop issues before we can move forward with this. You can find them on the failing tests above. Let me know if you have any issues!

kdoores:
Signed-off-by: Kayleigh kayleigh.doores@gmail.com

I removed the trailing whitespace and made the `unless` statement more readable and in in better position to match other `unless` statements of similar nature.  However the complexity max needs to be bumped to resolve that issue.  I tried sending empty strings but if you specify the `-d` flag, it wants a database.  It will no longer default to the users default database.  Therefore any type of if statement is necessary in that position to determine whether the `-d` flag should be used or not.  Is that change to the rubocop.yml file something you want me to do?  Or one you you guys that maintain the repo?

jquick:
Putting a:
`def query(q) # rubocop:disable Metrics/PerceivedComplexity`
is fine for now. We will refactor that later. Thanks so much for your help!

jquick (approved):
Thanks for this addition @kdoores !

miah (approved):
Thanks @kdoores !

PR #3384, opened 2018-09-12
Plugins: Add support for 'bundles' migration

clintoncwolfe:
As part of the Plugins API v2 effort, we'll be gradually moving the 'bundles' into a 'plugins' area, along with their tests; and we'll be upgrading them to use the Plugin V2 API.

This PR does three things:
* Configures the Loader to look in lib/plugins for v2 plugins
* Adds some basic test helper code for use by 'core' plugins
* Fixes a small handful of Loader issues discovered while developing the first plugin to use the `plugins` directory, the `inspec-plugin-manager-cli`.


jquick (approved):
Thanks @clintoncwolfe ! Really nice foundation for our core plugins.

miah (approved):
Thanks @clintoncwolfe! Looks solid =)

PR #3385, opened 2018-09-12
Add windows functional tests

jquick:

This PR adds some functional testing to windows. Currently only a few functional tests are supported but will be adding others.

Some tests were altered with different expected output for windows due to skips or platform specific testing. These did not impact what the tests were actually checking for.


miah (approved):
Thanks @jquick !

Rakefile
```diff
@@ -75,6 +75,23 @@ namespace :test do
     t.ruby_opts = ['--dev'] if defined?(JRUBY_VERSION)
   end
 
+  # Functional tests on windows are a WIP. Currently some
```
jerryaldrichiii: Could we capitalize Windows and add a period to the next line?

jerryaldrichiii: Might also change this to a `TODO: Ensure all functional tests are ran on Windows`?

test/functional/helper.rb
```diff
@@ -38,17 +39,47 @@ module FunctionalHelper
...
+  end
+
   def inspec(commandline, prefix = nil)
-    CMD.run_command("#{prefix} #{exec_inspec} #{commandline}")
+    if is_windows?
+      result = CMD.run_command("cmd /C \"#{prefix} bundle exec ruby #{exec_inspec} #{commandline}\"")
+      result.stdout.encode!(universal_newline: true)
+      result.stderr.encode!(universal_newline: true)
+      convert_windows_output(result.stdout)
+      # remove the CLIXML header trash in windows
```
jerryaldrichiii, line 62: Hah!

test/functional/inspec_exec_test.rb
```diff
@@ -12,14 +12,20 @@
...
     stdout = out.stdout.force_encoding(Encoding::UTF_8)
-    stdout.must_include "\e[38;5;41m  ✔  ssh-1: Allow only SSH Protocol 2\e[0m\n"
     stdout.must_include "\e[38;5;41m  ✔  tmp-1.0: Create /tmp directory\e[0m\n"
     stdout.must_include "
 \e[38;5;247m  ↺  gordon-1.0: Verify the version number of Gordon (1 skipped)\e[0m
 \e[38;5;247m     ↺  Can't find file `/tmp/gordon/config.yaml`\e[0m
 "
-    stdout.must_include "\nProfile Summary: \e[38;5;41m2 successful controls\e[0m, 0 control failures, \e[38;5;247m1 control skipped\e[0m\n"
-    stdout.must_include "\nTest Summary: \e[38;5;41m4 successful\e[0m, 0 failures, \e[38;5;247m1 skipped\e[0m\n"
+    if is_windows?
```
jerryaldrichiii, line 20: This is out of scope...but I wonder if we should use `--json` and check things that way. Do we care about the formatting colors and such?

jerryaldrichiii (approved):
So much yes! Only one question regarding checking formatting, but I won't hold up the PR over it.

test/functional/inspec_exec_test.rb
jquick: I think its valid to check both cli and json. The mapping of colors/symbols should make it easy though.

PR #3386, opened 2018-09-12
Populate code for inspec json inheritance

jquick:
Fixes https://github.com/chef/a2/issues/3622

This fix allows the `inspec json` command to pull code info from child profiles when needed.

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -229,7 +229,7 @@ module Inspec
       info(load_params.dup)
     end
 
-    def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity
+    def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       # add information about the controls
       res[:controls] = res[:controls].map do |id, rule|
         next if id.to_s.empty?
@@ -242,9 +242,9 @@ module Inspec
 
         # if the code field is empty try and pull info from dependencies
         if data[:code].empty?
-          locked_dependencies.dep_list.each do |name, dep|
+          locked_dependencies.dep_list.each do |_name, dep|
             profile = dep.profile
-            if profile.runner_context&.current_load.values.first == data[:source_location][:ref]
+            if profile.runner_context&.current_load&.values&.first == data[:source_location][:ref]
               data[:code] = Inspec::MethodSource.code_at(data[:source_location], profile.source_reader)
             end
           end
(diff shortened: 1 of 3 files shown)


miah (approved):
Thanks @jquick

lib/inspec/profile.rb
```diff
@@ -263,6 +272,7 @@ def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity
 
       res
     end
+    # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
jerryaldrichiii: I don't think we need this line, when putting it after a method call it is restricted to only that method.

lib/inspec/profile.rb
```diff
@@ -239,6 +239,15 @@ def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity
         data[:impact] = 1.0 if data[:impact] > 1.0
         data[:impact] = 0.0 if data[:impact] < 0.0
         data[:id] = id
+
+        # if the code field is empty try and pull info from dependencies
+        if data[:code].empty? && parent_profile.nil?
+          locked_dependencies.dep_list.each do |_name, dep|
```
jerryaldrichiii, line 245: Could this be extracted to a method to avoid `Metrics/PerceivedComplexity`?

lib/inspec/profile.rb
```diff
@@ -239,6 +239,15 @@ def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity
...
+            profile = dep.profile
+            code = Inspec::MethodSource.code_at(data[:source_location], profile.source_reader)
+            data[:code] = code unless code.nil? || code.empty?
```
jerryaldrichiii, line 248: I may just not be following, but won't this get overwritten by the last entry in `dep_list`? Is that the intended behavior?

jerryaldrichiii (dismissed):
Looks great, only a few questions, and a request to remove one line.

lib/inspec/profile.rb
jquick: Turns out you can't do it on a method, and its file based.

lib/inspec/profile.rb
jquick: Yeah we need a break

lib/inspec/profile.rb
jquick: Tried that, even putting a method call to a new one will blow it. We will need to redo this method in full to break it up more. Due to the nature of this bandaid hot fix we will put it under techDebt.

jerryaldrichiii: Understood. No judgment here!

jerryaldrichiii (approved):

PR #3387, opened 2018-09-12
Revert uuid change from A2 report

jquick:

This is needed before the release tomorrow as some A2 reports are still depending on it.

Revert the UUID fallback from https://github.com/inspec/inspec/pull/3320


miah (approved):
Thanks @jquick !

jerryaldrichiii (approved):

PR #3391, opened 2018-09-13
Prevent logs from showing up when running inspec json

jquick:

Currently our functional tests are broken for `inspec json` depending on the order of the tests. This fix pushes `Inspec::Log` to STDERR when running `inspec json`


jerryaldrichiii (approved):
Little sad about the phrasing of the `it` block, but won't hold the PR over it. I would like to see:

```
it 'can execute a profile with warn calls and parse STDOUT as valid JSON'
```

Or something similar.

PR #3392, opened 2018-09-13
Fix the compliance target error checks

jquick:

This makes sure `@upstream_sha256` is always set along with making the `check_compliance_token` error have a `uri` to specify in the error message.

Confirmed this fixes the upstream Audit issues.


jerryaldrichiii (approved):
Great catch!

PR #3399, opened 2018-09-16
Fix etc_fstab mount_options split

pielu:
Most often the _mounts_options_ part in _/etc/fstab_ will be simple _defaults_

### DIFF
--- a/lib/resources/etc_fstab.rb
+++ b/lib/resources/etc_fstab.rb
@@ -79,7 +79,7 @@ module Inspec::Resources
         'device_name'         => attributes[0],
         'mount_point'         => attributes[1],
         'file_system_type'    => attributes[2],
-        'mount_options'       => attributes[3].split(','),
+        'mount_options'       => attributes[3] =~ /,/ ? attributes[3].split(',') : attributes[3],
         'dump_options'        => attributes[4].to_i,
         'file_system_options' => attributes[5].to_i,
       }


clintoncwolfe:
This PR is simple, but I think it is not something we want.  The issue is with data types.  Currently, the mount options will always be split into an array, so you can rely on that data type.  It might be a single-element array `['defaults']`, but it is an array nonetheless, and if you care about it being 'defaults', you can check for that.

```
 defaults     => ['defaults'] (Array)
 ro           => ['ro']       (Array)
 defaults,ro  => ['defaults', 'ro'] (Array)
```

This PR would modify the behavior so that if no commas are in the raw mount options information, it would return a String instead; so you would get:

```
 defaults     => 'defaults' (String)
 ro           => 'ro'       (String)
 defaults,ro  => ['defaults', 'ro'] (Array)
```

I think that would be harder to deal with.  Thoughts, @jquick?

magdamag:
@clintoncwolfe I have just realised that, please disregard the PR

jquick:
I agree with @clintoncwolfe. Keeping the types the same would be ideal. Thanks for your effort on this @pielu. @magdamag do you happen to be the same person using a different account?

pielu:
No, of course not. I wasn't using my computer and didn't check credentials. Ugh. I've asked to remove previous comment. It's all good now. Again, apologies  for misdirected PR...

clintoncwolfe:
Closing per discussion.  Thanks so much for your idea - even if it wasn't accepted, it shows your enthusiasm, and we welcome future contributions from you!

PR #3400, opened 2018-09-17
Plugins: Example CLI Plugin, a Resource Lister

clintoncwolfe:
Included in this PR is an example of Plugins v2 CliCommand plugin. It is very simple - it simply lists the names of the resources that InSpec knows about.

The plugin is exhaustively commented for pedagogic purposes.

If we wish to include information on how to test a plugin using core facilities, we can do that after #3384 is merged, which provides such facilities.

If we wish to publish this to rubygems, I see no harm in that.


clintoncwolfe:
Temporarily closing this until test examples are added.

PR #3401, opened 2018-09-17
Add json-automate to the report method

jquick:

This allows the `json-automate` report be access via ad-hoc runners calling report. This is mainly for the Audit cookbook.


miah (approved):
Thanks @jquick

PR #3402, opened 2018-09-17
load inspec plugin help if no argument is given

chris-rock:
Before this PR, everything was working for `inspec help` but not pure `inspec`. `inspec` also shows the help message.

depends on #3384 and need to be rebased, once its merged


jquick (approved):
Thanks @chris-rock

PR #3398, opened 2018-09-15
adding `versions` to the `gem` resource

majormoses:
This is useful when you have multiple versions of the same gem installed. It can be leveraged like so:
```
describe gem('rest-client') do
  its('versions') { should include /1.8\.\d+/ }
  its('versions') { should include /2.0\.\d+/ }
  its('versions.count') { should_be eq 2 }
end
```

Signed-off-by: Ben Abrams <me@benabrams.it>

PR #3403, opened 2018-09-18
adding a property of `file_count` to `directory` resource

majormoses:
This is helpful to support testing if a directory is empty, not empty or even in some cases matching a specific number.

While refactoring some chef changes I wanted to make sure that a particular directory includes no files as the client should not need them. They are needed on the server and the directory is automatically created on both client and server.

Signed-off-by: Ben Abrams <me@benabrams.it>

### DIFF
--- a/lib/resources/directory.rb
+++ b/lib/resources/directory.rb
@@ -21,5 +21,11 @@ module Inspec::Resources
     def to_s
       "Directory #{source_path}"
     end
+
+    def file_count
+      return unless file.exist?
+      # intentionally used entries as we don't want to recurse/glob
+      Dir.entries(file).reject { |f| f == '.' || f == '..' }.count
+    end
   end
 end
--- a/docs/resources/directory.md.erb
+++ b/docs/resources/directory.md.erb
@@ -33,7 +33,37 @@ A `directory` resource block declares the location of the directory to be tested
 
 All of the properties available to `file` may be used with `directory`.
 
+### file_count
+
+The `file_count` property returns an integer of the number of files directly under the directory. It does not recurse directories and removes `.` and `..` entries. This is helpful for both verifying a directory is empty or if there are files present.
+
+  # empty directory
+  its('file_count') { should eq 0 }
+  # one or more files present
+  its('file_count') { should be > 0 }
+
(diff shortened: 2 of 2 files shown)


majormoses:
Looks like the same transient issue as I encountered here https://github.com/inspec/inspec/pull/3398#issuecomment-421523635 if you can kick appveyor again and see if it resolves that would be great.

clintoncwolfe (dismissed):
So straightforward, I don't think it even needs a test.  And you included docs!  Thanks!

lib/resources/directory.rb
```diff
@@ -21,5 +21,11 @@ def exist?
     def to_s
       "Directory #{source_path}"
     end
+
+    def file_count
+      return unless file.exist?
+      # intentionally used entries as we don't want to recurse/glob
+      Dir.entries(file).reject { |f| f == '.' || f == '..' }.count
```
jquick, line 28: Using a Dir.entries will only work on the local transport. Even then I am getting errors using this like:
```
(master)> _inspec exec file_count.rb

Profile: tests from file_count.rb (tests from file_count.rb)
Version: (not specified)
Target:  local://

  Directory /tmp
     ✔  should exist
     ✔  should be directory
     ×  file_count
     no implicit conversion of Train::File::Local::Unix into String
```
Changing it to `file.path` fixes the issue for local but will cause unexpected results when using a remote target. I think we need to add more logic into train if we want to do this: https://github.com/inspec/train/tree/master/lib/train/file. Adding a Directory class in Train would be a good start i think.

jquick (changes requested):
Thanks for the PR @majormoses. Unfortunately I think we need to revisit this implementation and make sure it works for both local and remote transports.

lib/resources/directory.rb
clintoncwolfe: Goooood catch, @jquick.

jquick:
Since this will require some pre-work in inspec/train I am closing this PR for now. Once we have the proper setup in Train we can then implement it here. Thanks again for your help and involvement!

majormoses:
@jquick sorry for the late response and good catch, is there an issue to add the required bits to train?

jquick:
Hey @majormoses - I created https://github.com/inspec/train/issues/360 to track this.

PR #3405, opened 2018-09-18
New resource to work with Windows security identifiers (SIDs)

james-stocks:
We're currently making use of this resource as a custom resource in our profiles; I would like to propose adding it into Inspec.

Example control that this new resource helps with when testing `security_policy`:

```
permitted_users = ['S-1-5-9', security_identifier(group: 'Administrators').sid, security_identifier(user: 'Administrator').sid]
  filtered_principals = security_policy.SeNetworkLogonRight - permitted_users
  filtered_principals.each do |suspect_principal|
    describe(suspect_principal) do
      it { should_not match /.*/ }
    end
  end
```

_Why bother with this resource instead of stating the SIDs as literal strings?_
For one, some SIDs are dynamic (include the domain ID). Secondly, some automated compliance content will use user and group names when, in practice, the SID is needed. To automatically consume and execute this content, we need an automatic way to convert an entity name to a SID.

_Why bother needing the user to specify user, group or unspecified_
It is possible to create users and groups and other entity types on Windows with the same name. It may cause inaccurate results and leave a slight gap in security if the name of the entity being tested is ambiguous.
The `unspecified` type is needed because some automated compliance content will name entities that should be tested without indicating what type of entity they are. To keep this automated, we need a way to try to determine a SID when we're given an entity name but not the entity type. Ideally, `unspecified` should not be used.


test/integration/default/controls/security_identifier_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+describe security_identifier({ user: 'Administrator' }) do
+  its('sid') { should match %r{S-1-5-21.+-500} }
+end
+
+describe security_identifier({ unspecified: 'Administrator' }) do
+  its('sid') { should match %r{S-1-5-21.+-500} }
+end
+
+# If not appropriate, sid just returns the name back
```
clintoncwolfe: Hrm, this would be atypical behavior compared to other InSpec core resources. When a resource performs a search, the usual way of describing a "miss" is:

* set `.exists?` to false
* set all properties to `nil`

test/integration/default/controls/security_identifier_spec.rb
```diff
@@ -0,0 +1,32 @@
+# encoding: utf-8
+
+unless os.windows?
+  STDERR.puts "\033[1;33mTODO: Not running #{__FILE__} because we are not on Windows.\033[0m"
+  return
+end
+
```
clintoncwolfe, line 7: I think you should have an `.exist?` property.  That is more typical for resources.

test/integration/default/controls/security_identifier_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+describe security_identifier({ user: 'Administrator' }) do
+  its('sid') { should match %r{S-1-5-21.+-500} }
+end
+
+describe security_identifier({ unspecified: 'Administrator' }) do
```
clintoncwolfe, line 13: Great start on the tests!

test/unit/resources/security_identifier_test.rb
```diff
@@ -0,0 +1,11 @@
+# encoding: utf-8
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::SecurityIdentifier' do
+  it 'instantiates' do
+    resource = MockLoader.new(:windows).load_resource('security_identifier', { user => 'Administrator' })
+    _(resource).wont_be_nil
+  end
```
clintoncwolfe, line 17: Since you accept a variety of search keys (user, group, unspecified) you should probably test that all 3 of those are accepted.  Then you should try passing an unrecognized search key, and verify that an ArgumentError or similar is thrown.

test/unit/resources/security_identifier_test.rb
```diff
@@ -0,0 +1,11 @@
+# encoding: utf-8
+
+require 'helper'
+require 'inspec/resource'
+
+describe 'Inspec::Resources::SecurityIdentifier' do
+  it 'instantiates' do
```
clintoncwolfe: I would prefer that this test be expanded to also test actual searches, and their results.  For example, you could have a mock that knows about a user named Alice, and verify that you can find Alice.  We're not talking to an actual Windows machine here, but the goal would be to show that you can interpret the output of the machine properly; and you can add data (like Alice) that would not be on an integration test fixture.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+module Inspec::Resources
+  class SecurityIdentifier < Inspec.resource(1)
+    name 'security_identifier'
+    supports platform: 'windows'
+    desc 'Resource that returns a Security Identifier for a given entity name in Windows. Because different entities can have the same name (e.g. a user and group can both be called \'devops\') the resource requires the type of the entity (:user, :group) to be stated to avoid an ambiguous test. If the entity type is unknown (e.g. when working with SCAP content that names an entity but does not declare its type) you may give the type as :unspecified to explicitly state that you need the resource to try to determine a SID when the entity type is not known.'
```
clintoncwolfe: This should be moved to docs/resources/security_identifier.md.erb , and expanded to have the section headers typical of a resource doc.  Thanks!

You can leave a one-line description here.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+    example "
```
clintoncwolfe: Consider using heredoc markers, which will automatically escape quotes.  That preserves the "copy-paste" ability that a lot of users want.

```
example <<~EOD
  describe security_policy do
    its("SeRemoteInteractiveLogonRight") { should_not include security_identifier(group: 'Guests') }
   end
EOD
```

clintoncwolfe: While I understand this is primarily a data lookup resource, you should still show it in use alone.  Having two examples - one in which it is used alone, and one 'advanced example' in which it is used in conjunction with a `security_policy` would be clearer to show what this resource does.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+      describe security_policy do
+        its(\"SeRemoteInteractiveLogonRight\") { should_not include security_identifier(group: 'Guests') }
+      end
+    "
+
+    def initialize(opts = {})
+      supported_opt_keys = [:user, :group, :unspecified]
+      raise "Invalid security_identifier param '#{opts}'. Please pass a hash with these supported keys: #{supported_opt_keys}" unless opts.respond_to?(:keys)
```
clintoncwolfe: Please raise ArgumentError exceptions when an arg is invalid.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+        @type = :unspecified
+        @name = opts[:unspecified]
+      end
+      raise 'Specify one of :user :group or :unspecified for security_identifier' unless @name
+      @sids = nil
+    end
+
+    def sid
+      fetch_sids unless @sids
+      @sids[@name] || @name
```
clintoncwolfe: If you can't find something, return nil, and have an `exist?` property.  That's the usual interface.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+          sid_data = wmi_results_array("wmic useraccount where 'Name=\"#{@name}\"' get Name\",\"SID /format:csv")
+        end
+      else
+        raise "Unhandled entity type '#{@type}'"
+      end
+      sid_data.each { |sid| @sids[sid[1]] = sid[2] }
+    end
+
+    def wmi_results_array(query)
+      inspec.command(query).stdout.strip.split("\r\n\r\n")[1..-1].map { |entry| entry.split(',') }
```
clintoncwolfe: Do you need the `\r`?  `\n` is typically the OS-Specific newline, so it should handle that on its own.

clintoncwolfe: It would be kinder to your fellow developers to leave a comment with an example of the CLI data you are parsing, and the data structure you intend to return.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+        if sid_data.empty?
+          sid_data = wmi_results_array("wmic useraccount where 'Name=\"#{@name}\"' get Name\",\"SID /format:csv")
+        end
+      else
+        raise "Unhandled entity type '#{@type}'"
+      end
+      sid_data.each { |sid| @sids[sid[1]] = sid[2] }
+    end
+
+    def wmi_results_array(query)
```
clintoncwolfe: You can probably drop the suffix `array`.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,66 @@
...
+      @sids[@name] || @name
+    end
+
+    private
+
+    def fetch_sids
+      @sids = {}
+      case @type
+      when :group
+        sid_data = wmi_results_array("wmic group where 'Name=\"#{@name}\"' get Name\",\"SID /format:csv")
```
clintoncwolfe: It looks like only the `group`/`useraccount` part actually varies in each of these invocations.  Consider moving the construction of the command line into `wmi_results_array`, so you only are obliged to build the complex command line in one place, not 4.

clintoncwolfe (changes requested):
Thanks for the contribution!  I agree, this would be great data lookup resource to have, and I think you have made a great start.  

I'd ask that it handle search misses differently, move the docs to a markdown file, and improve test coverage.  It's a great start, and can easily be matured into a strong resource.  Thanks again!

PS - Looks like you have a failing unit test - see https://travis-ci.org/inspec/inspec/jobs/430099348#L1237

test/integration/default/controls/security_identifier_spec.rb
james-stocks: When you suggest an `.exist?` property I'm assuming this is tested with `it { should exist }` , right?

clintoncwolfe: Yes, that's right.

james-stocks:
Hi @clintoncwolfe thanks for the awesome feedback.
I believe I've addressed it all now; can you have another look at this PR?

One thing I need help with is specifying the release version in the documentation page - I can quickly update that just before this gets merged; or is there some automation to set the right version?

Thanks

docs/resources/security_identifier.md.erb
```diff
@@ -0,0 +1,63 @@
...
+
+## Availability
+
+### Installation
+
+This resource is distributed along with InSpec itself. You can use it automatically.
+
+### Version
+
+This resource first became available in vX.Y.Z of InSpec.
```
james-stocks: Can I just put the next version number for Inspec into this line?

test/unit/resources/security_identifier_test.rb
james-stocks: Have added tests for these

docs/resources/security_identifier.md.erb
clintoncwolfe: Simply omit the Version section for now.

docs/resources/security_identifier.md.erb
```diff
@@ -0,0 +1,63 @@
...
+A `security_identifier` resource should specify the name and type of the entity to test the SID for:
+
+    describe security_identifier(group: 'Everyone') do
+      its('sid') { should eq 'S-1-1-0' }
+    end
+
+where
+
+* `group:` specifies that `'Everyone'` should be a group. `user:` can be used to specify a user account.
+** It is necessary to declare the type of the entity because Windows allows users, groups and other entities to share names. If you really need to not specify the type, `unsupported:` can be used. This will attempt to match the name to a group and then a useraccount. This may take longer to execute and comes with the risk of Inspec matching the name to an unintended type of entity.
```
clintoncwolfe: I think you mean 'unspecified', not 'unsupported'

docs/resources/security_identifier.md.erb
```diff
@@ -0,0 +1,63 @@
...
+
+### Installation
+
+This resource is distributed along with InSpec itself. You can use it automatically.
+
+### Version
+
+This resource first became available in vX.Y.Z of InSpec.
+
+## Syntax
```
clintoncwolfe: Prefer 'Resource Parameters' as a header instead of 'Syntax'.  We're inconsistent about this :/

docs/resources/security_identifier.md.erb
clintoncwolfe: When referred to as a product, we try to refer to it as InSpec (note caps).  As a CLI or gem, use 'inspec'.

docs/resources/security_identifier.md.erb
```diff
@@ -0,0 +1,63 @@
...
+### Verify that a SID is the expected value
+
+    describe security_identifier(group: 'Everyone') do
+      its('sid') { should eq 'S-1-1-0' }
+    end
+
+### Use in conjunction with the security_policy resource to specify the entity to test for in the audit policy
+
+    describe security_policy do
+      its(\"SeRemoteInteractiveLogonRight\") { should_not include security_identifier(group: 'Guests') }
```
clintoncwolfe: I don't think the escapes on the quotes are needed here; and will be syntax errors if someone copy-pastes this.

docs/resources/security_identifier.md.erb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+<br>
+
+## Matchers
```
clintoncwolfe, line 69: You should list `exist` here.

clintoncwolfe, line 69: You are missing a section, 'Properties', in which you explain anything you can pass to `its`.  At least `sid`.

lib/resources/security_identifier.rb
```diff
@@ -0,0 +1,84 @@
...
+        query += 'group'
+      when :user
+        query += 'useraccount'
+      end
+      query += " where 'Name=\"#{@name}\"' get Name\",\"SID /format:csv"
+      # Example output:
+      #     inspec> command("wmic useraccount where 'Name=\"Administrator\"' get Name\",\"SID /format:csv").stdout
+      #     => "\r\n\r\nNode,Name,SID\r\n\r\nComputer1,Administrator,S-1-5-21-650485088-1194226989-968533923-500\r\n\r\n"
+      # Remove the \r characters, split on \n\n, ignore the CSV header row
+      inspec.command(query).stdout.strip.tr("\r", '').split("\n\n")[1..-1].map { |entry| entry.split(',') }
```
clintoncwolfe, line 81: Very good comments!

clintoncwolfe (changes requested):

clintoncwolfe:
Looking very good! Only changes needed are all in the docs.

docs/resources/security_identifier.md.erb
james-stocks: Added

james-stocks: Thanks, added

james-stocks:
Thanks @clintoncwolfe - I made the docs improvements you pointed out. I also added a link to the Microsoft page that outlines Security Identifiers and changed the resource doc wording to match their terminology better (i.e. call a thing a `trustee` and not a `entity`)

clintoncwolfe (approved):

jquick (approved):
Looks good. Thanks @james-stocks

jquick:
closing/reopening to trigger windows tests

james-stocks:
Closing to refresh appveyor

PR #3407, opened 2018-09-18
Move inspec init to v2 plugins

jquick:

This moves the core inspec init plugin to the new v2 plugins.


lib/plugins/shared/core_plugin_test_helper.rb
```diff
@@ -33,9 +33,9 @@ module CorePluginFunctionalHelper
   require 'train'
   TRAIN_CONNECTION = Train.create('local', command_runner: :generic).connection
 
-  def run_inspec_process(command_line, env = {})
-    env_prefix = env.to_a.map { |assignment| "#{assignment[0]}=#{assignment[1]}" }.join(' ')
-    TRAIN_CONNECTION.run_command("#{env_prefix} #{inspec_bin_path} #{command_line}")
+  def run_inspec_process(command_line, prefix = nil)
```
clintoncwolfe: I think if you want to have its behavior change like this, we should change the signature to this:

```
def run_inspec_process(command_line, opts = {})
  # Handle opts[:env] hash if present
  # Handle opts[:prefix] string if present
end
```

That allow for future growth much better.  For example, it would be great to have a :json_output mode that automatically adds a json formatter to the command and parses the output back to ruby, so you can examine run output in detail.  Or a :timeout option, etc.

If you agree with that idea (of treating the remain args as an opts hash), please make the change on an isolated commit, so I can cherry-pick it.

clintoncwolfe (dismissed):
(pending clean linter and test results)

This looks good!  One concern about run_inspec_process, which I think we should discuss.

lib/plugins/shared/core_plugin_test_helper.rb
jquick: will do, i like this idea. Will push it up

jquick: commit https://github.com/inspec/inspec/pull/3407/commits/7f828ca168c85d6414dfd16d81f03f6c64a028a2

jerryaldrichiii (approved):

miah (approved):

PR #3408, opened 2018-09-18
Fix json automate tests and render call

jquick:

The `json-automate` reporter's test were not firing because the test did not end with `_test`. This let some bugs out in the last audit fix. This PR fixes the tests along with the current render issue.


jerryaldrichiii (approved):

miah (approved):
Thanks @jquick

vjeffrey (approved):
thank you!

Issue #3174, opened 2018-06-26
Clarify what should happen when a resource is unable to run a test


clintoncwolfe:
### 🎛 Description

🙋 feature request

When a resource is unable to perform a test, what should the outcome of the test be?

Examples of being unable to perform a test:

* `http` resource requires `curl`, but it is not installed
* timeout while trying to connect to a database resource
* need to obtain information only available to `root`, but not running with sudo
* the test relies on examining a configuration file, but it does not exist (for example, `shadow` and `/etc/shadow`)

This is considered distinct from running a resource on an unsupported platform, which is detect by inspec before the test is attempted, and reported as a `skip`.

#### Expected Behavior

We currently handle test failures in some places by throwing an exception, or calling a helper method that throws the exception.  This bubbles up to be treated as a failed test.

A similar approach could be taken for `unable` tests, in terms of throwing an exception via a helper.

Overall, consensus within the InSpec team was that this should probably be handled as a type of failure; if it were treated as a type of skip, it would likely be ignored.

Refs #3017 #3158 

### 🌍 InSpec and Platform Version

2.2.x

wardwouts:
I agree that tests that cannot be correctly executed should result in some sort of failure. To me the current behaviour where if I run this simple test, as a non-root user:

```
control "Tests reading in protected file" do
  title "Not allowed to read"
  desc "A file that cannot be read should not let a test succeed."
  describe file('/etc/shadow') do
    its('content') { should_not match /testvalue/ }
  end
end
```


the result is a pass, is obviously wrong.

ghost:
I also agree that tests that cannot be correctly executed should result in some sort of a failure. Having test that did not even run report as succesfull is not the way to go.

majormoses:
This caused a production issue that was meant to be caught by the tests I wrote, if the tests can't run it should report back so via an exit code indicating all is not well. There are maybe use cases where they should be skipped because the tests are flakey themselves but not something as obvious I dont have the deps to run the check.

aaronlippold:
So are we actively working the above list ?

majormoses:
Well I tried to open a pull request to start working on it but it was closed out for further discussion.

majormoses:
Is there any progress on being able to move forward on this? I am happy to help but don't want to keep opening pull requests if we want to wait for more feedback. I think the overwhelming majority supports this behavior change and without it inspec is fundamentally not reliable for test automation or compliance as you can't trust the tests.

[...]

PR #3413, opened 2018-09-18
make `inspec` fail file resources when they are not found.

### LINKED ISSUE #3412: Inspec skipping tests when file based resources do not exist
### 🎛 Description

We experienced an issue that was supposed to be caught by testing and was not because it was marked as skipped and skipped does not actually report as failures. This is a breaking change and would need to happen in the next major release.

OS: Ubuntu 18.04

```
$ bundle info inspec
  * inspec (2.2.102)
    Summary: Infrastructure and compliance testing.
    Homepage: https://github.com/chef/inspec
    Path: /home/babrams/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/inspec-2.2.102
```

### 🤔 Replication Case

```
inspec> describe json('/tmp/dummy') do
inspec>   it { should exist } 
inspec> end  

Profile: inspec-shell
Version: (not specified)

  JSON /tmp/dummy
     ↺  Can't find file: /tmp/dummy

```

### 💁 Possible Solutions

https://github.com/inspec/inspec/blob/v2.2.111/lib/utils/file_reader.rb#L10-L12 I think this needs to change to something like this:
```
if !file.file?
  raise Inspec::Exceptions::ResourceFailed, "Can't find file: #{path}"
end
```

majormoses:
This is a breaking change and should make it into the next major release.

fixes #3412 by making tests fail when they rightfully should.

Signed-off-by: Ben Abrams <me@benabrams.it>

### DIFF
--- a/lib/utils/file_reader.rb
+++ b/lib/utils/file_reader.rb
@@ -8,16 +8,16 @@ module FileReader
     # ResourceFailed during a major version bump.
     file = inspec.file(path)
     if !file.file?
-      raise Inspec::Exceptions::ResourceSkipped, "Can't find file: #{path}"
+      raise Inspec::Exceptions::ResourceFailed, "Can't find file: #{path}"
     end
 
     raw_content = file.content
     if raw_content.nil?
-      raise Inspec::Exceptions::ResourceSkipped, "Can't read file: #{path}"
+      raise Inspec::Exceptions::ResourceFailed, "Can't read file: #{path}"
     end
 
     if !allow_empty && raw_content.empty?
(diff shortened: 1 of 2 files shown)


clintoncwolfe:
Refs #3174

clintoncwolfe:
Thanks for the contribution! However, I fear this change would have very wide impact, as file_reader is used in many places.  I think #3174, the general case of what to do when a resource cannot run, should be addressed, instead of just files being missing.

jquick:
Closing this as we need to talk about the solution as a community here https://github.com/inspec/inspec/issues/3174.

PR #3420, opened 2018-09-20
Remove load locks for cloud resources

jquick:

If you install InSpec via a gem (without bundle) you will not load aws/azure at the time of resource load (the transport init happens after this). Thus all resources will fail like:
```
×  aws-1: Checks the machine is running
     ×  Control Source Code Error myaws/controls/example.rb:6
     undefined method `aws_ec2_instance' for #<#<Class:0x00007fa1f8cf7b40>:0x00007fa1f8cf5ea8>
```

Using a bundle, omni, or hab(A2) will get around this issue as it will auto load all the gems in your bundle during a bundle exec.

Removing the load locks fixes the issue.


jerryaldrichiii (approved):
Great catch @jquick!

miah (approved):

PR #3424, opened 2018-09-20
Add support for multiple descriptions for controls

### LINKED ISSUE #1695: RFC: Support multiple descriptions per control
### Description

InSpec has already a great support for tags, references and description of controls.  This works great and looks like:

```
control "V-72841" do
  title "PostgreSQL must be configured to prohibit or restrict the use of
        organization-defined functions, ports, protocols, and/or services, as
        defined in the PPSM CAL and vulnerability assessments."
  desc  "In order to prevent unauthorized connection of devices, unauthorized
        transfer of information, or unauthorized tunneling (i.e., embedding of
       ...
       "
```

Sometimes you need additional description e.g. for a specific check or remediation. Right now, those blocks need to be part of the `desc` block, since tags are not suited for long text.

```
  desc "Check": "As the database administrator, run the following SQL:
  $ psql -c \"SHOW port\"
  If the currently defined port configuration is deemed prohibited, this is a
  finding."

  desc "Fix": "Note: The following instructions use the PGDATA environment
  variable. See supplementary content APPENDIX-F for instructions on configuring
  PGDATA.
```

I propose we extend the current `desc` for `control` to allow

```
desc "Section", "Long text"
```

You can still set the default desc by only passing one parameter:
```
control 'my control' do
  desc "A brief description of the control"
  desc "Section", "Long text"

  describe ...
end
```

When overriding a control from a parent the parent description fields will overwrite the child's:
Child control:
 
```
control 'ssh-01' do
  desc "Test to make sure admin password is not admin"
  desc "reasoning", "If this does not work, bad things happen"

  describe ...
end
```

Parent override:

```
include_controls 'child_profile' do
  control 'ssh-01' do
    desc "reasoning", "If this does not work, pull the fire alarm"
    desc "fix", "change the admin password!"
  end
end
```

you should end up with:
```
  desc "Test to make sure admin password is not admin"
  desc  "reasoning", "If this does not work, pull the fire alarm"
  desc "fix", "change the admin password!"
```

For output we will want to add a section to the json schema:
```
"desc:" : "the default description"
^ will stay the same and just be the default description set

"descriptions": [
    { "type": "reason", "data": "why ..." },
^ is a new fields and will be an ARRAY of descriptions
```

To this end we will also need to add descriptions to the inspec/schema fields.

1.20.0

jerryaldrichiii:
This adds the ability to specify multiple descriptions in controls.

Example:

```ruby
control 'my-control' do
  impact 1.0
  title 'My control'
  desc 'A default description'
  desc 'rational', 'I need an example'
  describe file('/tmp') do
    it { should be_directory }
  end
end
```

Many thanks to @jquick for helping me with the unit tests.

This closes https://github.com/inspec/inspec/issues/1695.


aaronlippold:
I thought we were going with :

`desc 'fix': "do the right thing" `

to keep more in line with our `tag 'joe': ['billy','bob']` formatting style ...

jquick:
We want to move away from having people understand how hashes work for non-programmers. To this end we are using parameter passing instead of hash format.

lib/inspec/rule.rb
```diff
@@ -89,9 +89,18 @@ def title(v = nil)
...
+    def desc(v = nil, data = nil)
+      return @descriptions[:default] if v.nil?
+      if data.nil?
+        @descriptions[:default] = unindent(v)
+      else
+        @descriptions[v.to_sym] = unindent(data)
+      end
+    end
+
+    def descriptions(desc_array = nil)
```
jquick: What is this for? I assume it would be for:
```
control 'my_control' do
  descriptions(default: 'test', status: 'success')
  describe ...
end
```
I am not sure this is needed at all. If its for testing id make it a __descriptions. We don't want to confuse people with two formats to do the same thing.

test/unit/mock/profiles/myprofile1/inspec.yml
```diff
@@ -0,0 +1,8 @@
+name: myprofile1
```
jquick: I don't think you need this profile. The wrapper-override is vendored.

jquick (commented):
Looking pretty good. Needs documentation also.

test/functional/inspec_exec_test.rb
```diff
@@ -446,16 +446,27 @@
       controls.count.must_equal 2
 
       # check for json override
-      assert = "  control 'pro1-con2' do\n    impact 0.999\n    title 'Profile 1 - Control 2-updated'\n    desc 'Profile 1 - Control 2 description-updated'\n    tag 'password-updated'\n    ref 'Section 3.5.2.1', url: 'https://example.com'\n    describe file('/etc/passwd') do\n      it { should exist }\n    end\n  end\n"
+      assert = "  control 'pro1-con2' do\n    impact 0.999\n    title 'Profile 1 - Control 2-updated'\n    desc 'Profile 1 - Control 2 description-updated'\n    desc 'overwrite me', 'it is overwritten'\n    desc 'append this', 'it is appended'\n    tag 'password-updated'\n    ref 'Section 3.5.2.1', url: 'https://example.com'\n    describe file('/etc/passwd') do\n      it { should exist }\n    end\n  end\n"
```
clintoncwolfe: Consider calling this 'expected_value'.  Assert is a verb, and this is not an assertion 😄

test/functional/inspec_exec_test.rb
```diff
@@ -446,16 +446,27 @@
...
+        },
+      ])
       override['title'].must_equal "Profile 1 - Control 2-updated"
       tags_assert = {"password"=>nil, "password-updated"=>nil}
       override['tags'].must_equal tags_assert
       child_profile['parent_profile'].must_equal 'wrapper-override'
 
       # check for original code on child profile
-      assert = "control 'pro1-con2' do\n  impact 0.9\n  title 'Profile 1 - Control 2'\n  desc 'Profile 1 - Control 2 description'\n  tag 'password'\n  describe file('/etc/passwdddddddddd') do\n    it { should exist }\n  end\nend\n"
+      assert = "control 'pro1-con2' do\n  impact 0.9\n  title 'Profile 1 - Control 2'\n  desc 'Profile 1 - Control 2 description'\n  desc 'overwrite me', 'overwrite this'\n  tag 'password'\n  describe file('/etc/passwdddddddddd') do\n    it { should exist }\n  end\nend\n"
```
clintoncwolfe: expected_value ?

test/functional/inspec_exec_test.rb
clintoncwolfe: Consider storing the expected value as a string with expanded newlines, and then escaping them in code.  That makes the expected value much more readable. Not a blocker.

test/unit/mock/profiles/wrapper-override/controls/defaut.rb
```diff
@@ -7,6 +7,8 @@
     impact 0.999
     title 'Profile 1 - Control 2-updated'
     desc 'Profile 1 - Control 2 description-updated'
+    desc 'overwrite me', 'it is overwritten'
+    desc 'append this', 'it is appended'
```
clintoncwolfe: I think 'append' is not what you mean here. If you have something that is overwriting, and something that is appending, one expects both of their child values to pre-exist, and the two to differ in the way they handle the pre-existing value.  

I think something like 'new entry' might be better.

test/unit/mock/profiles/myprofile1/README.md
```diff
@@ -0,0 +1,4 @@
+# Demo Compliance Profile
```
clintoncwolfe: As Jared said, I don't think we need this.  If we do, please give it a meaningful name, not 'myprofile1'. 'override-child'?

lib/inspec/schema.rb
```diff
@@ -84,6 +84,7 @@ class Schema
         'id' => { 'type' => 'string' },
         'title' => { 'type' => %w{string null} },
         'desc' => { 'type' => %w{string null} },
+        'descriptions' => { 'type' => %w{array} },
```
clintoncwolfe, line 87: Personally, I would prefer 'desc' and 'desc_sections'.  Having desc and descriptions is confusing - it does not communicate that the latter is 'unassembled'.

If `descriptions` is specified by the RFC, welp, I guess that's what we  want. 🤷‍♂️

test/functional/inspec_exec_json_test.rb
```diff
@@ -117,6 +117,9 @@
         "id" => "tmp-1.0",
         "title" => "Create /tmp directory",
         "desc" => "An optional description...",
+        "descriptions" => [
+          { 'type' => 'default', 'data' => 'An optional description...' }
```
clintoncwolfe: This probably isn't a big deal, but maybe the attribute 'type' should be reconsidered.  I don't think we can assert that the user will think of each call to `desc arg1, arg2` as specifying a 'type of description'.  I think users will use this feature in a number of ways.  'name' or 'label' reflects what we know about the user's intent - they wanted to separate out and label some chunk of the description for some reason. We can't assume it was due to any particular purpose, like assigning a type of description.

Just a thought.

test/functional/inspec_exec_json_test.rb
```diff
@@ -117,6 +117,9 @@
...
+        ],
```
clintoncwolfe: It would be great if we could have more testing here: a test in which `inspec json` is run against a profile that defines multiple descriptions for a control, and that they appear in the right places in the schema.

(inspec_exec_test.rb does test the reporter dump of that; but inspec json is an exporter)

test/unit/dsl/control_test.rb
```diff
@@ -22,22 +22,42 @@ def load(content)
...
   it 'adds a description' do
-    load("desc #{rand_string.inspect}")[:desc].must_equal rand_string
+    load("desc #{rand_string.inspect}")[:descriptions][:default].must_equal rand_string
+  end
+
+  it 'allows multiple desc with keys and values' do
+    lines = <<-DESC_EXAMPLES
+      desc 'this is default'
+      desc 'foo', 'this is foo'
+      desc 'a space', 'this has a space'
```
clintoncwolfe: Consider 'the label has a space'

test/unit/dsl/control_test.rb
```diff
@@ -22,22 +22,42 @@ def load(content)
...
+      desc 'newline', 'this has
```
clintoncwolfe: Consider 'the value has a newline'

test/unit/dsl/control_test.rb
```diff
@@ -22,22 +22,42 @@ def load(content)
...
 
   it 'adds a description' do
-    load("desc #{rand_string.inspect}")[:desc].must_equal rand_string
+    load("desc #{rand_string.inspect}")[:descriptions][:default].must_equal rand_string
+  end
+
+  it 'allows multiple desc with keys and values' do
+    lines = <<-DESC_EXAMPLES
+      desc 'this is default'
+      desc 'foo', 'this is foo'
```
clintoncwolfe: Consider 'basic' instead of 'foo'. Personally, 'foo' reads unprofessional to me.

test/unit/dsl/control_test.rb
```diff
@@ -22,22 +22,42 @@ def load(content)
...
+      desc 'heredoc', <<-EOF
+        This is a heredoc
+      EOF
+    DESC_EXAMPLES
+
+    result = load(lines)
+    result[:descriptions][:default].must_equal('this is default')
+    result[:descriptions][:foo].must_equal('this is foo')
+    result[:descriptions][:'a space'].must_equal('this has a space')
+    result[:descriptions][:newline].must_match(/this has\n.*a newline/)
```
clintoncwolfe: Consider tightening regex to '\s*' instead of '.*'

test/unit/dsl/objects_test.rb
```diff
@@ -274,13 +274,20 @@
...
+        rational: 'It is needed to save the planet',
+        'more info': 'Insert clever joke here',
+      }
       control.refs = ['simple ref', {ref: 'title', url: 'my url'}]
       control.impact = 1.0
       control.to_ruby.must_equal '
 control "sample.control.id" do
   title "Sample Control Important Title"
   desc  "The most critical control the world has ever seen"
+  desc  "rational", "It is needed to save the planet"
```
clintoncwolfe: rationale?

lib/inspec/rule.rb
jerryaldrichiii: Oh, that was from when I was toying with the idea of `descriptions` instead of `desc`. I'll remove it.

test/unit/mock/profiles/myprofile1/inspec.yml
jerryaldrichiii: Yup, good catch!

lib/inspec/objects/control.rb
```diff
@@ -18,13 +19,27 @@ def add_tag(t)
...
+        tests: tests.map(&:to_hash),
+        tags: tags.map(&:to_hash),
+      }
     end
 
     def to_ruby # rubocop:disable Metrics/AbcSize
       res = ["control #{id.inspect} do"]
       res.push "  title #{title.inspect}" unless title.to_s.empty?
-      res.push "  desc  #{prettyprint_text(desc, 2)}" unless desc.to_s.empty?
+      descriptions&.each do |k, v|
```
clintoncwolfe: You can drop the '&' - we know descriptions has a Hash value because it is initialized.

clintoncwolfe: It's kinder to your fellow developers to use meaningful variable names.  Are the keys here labels?  Are the values text chunks?  Use words that reflect that.  Thanks!

lib/inspec/reporters/json.rb
```diff
@@ -60,7 +60,8 @@ def profile_controls(profile)
         control = {
           id: c[:id],
           title: c[:title],
-          desc: c[:desc],
+          desc: c&.dig(:descriptions, :default),
```
clintoncwolfe: Hey, you don't have to use both '&.' and 'dig'. We know c is defined, and a hashlike thing, so you can certainly call `.dig` .  Dig will handle any nil steps on its own.

lib/inspec/reporters/json.rb
```diff
@@ -116,5 +117,12 @@ def profiles
       end
       profiles
     end
+
+    def convert_descriptions(data)
+      return [] if data.nil?
+      results = []
+      data.each { |x| results.push({type: x[0].to_s, data: x[1]} ) }
```
clintoncwolfe: For your fellow developers, please use meaningful variable names.  It also helps show your intent and clarity of thought.  Thanks!

clintoncwolfe (dismissed):
Thanks so much Jerry!  Mainly changes for clarity, a a bit more test coverage. Looks good overall in terms of approach.  Thanks!

lib/inspec/schema.rb
aaronlippold: I agree `desc_sections` follows more naturally. `desc_sections.each do |section|`

jquick: I think the reason behind `descriptions` is that we want to deprecate `desc`(schema output) and eventually remove it. I agree though during transition it will be a bit confusing.

test/functional/inspec_exec_json_test.rb
jquick: Yeah I like `name` or `label`. I think `type` is to general for this case. Whatever one you think @jerryaldrichiii should be fine. A2 can adapt.

test/functional/inspec_exec_test.rb
jerryaldrichiii: Yup, both of those were to match style of the other tests. Changing `assert` to `expected-value` seems easy enough though, I'll do that.

test/unit/mock/profiles/wrapper-override/controls/defaut.rb
jerryaldrichiii: Append is what I mean here. This appends `'append this'` to the list of descriptions. Where as `'overwrite me'` overwrites the one in the child. I see your point about the ambiguity of append though. Will change to `new_entry` and modify the description to be more verbose.

test/unit/dsl/objects_test.rb
jerryaldrichiii: Herp derp. Will fix.

lib/inspec/objects/control.rb
jerryaldrichiii: Yup, great catch!

jerryaldrichiii: I like that, I'll make them more verbose.

lib/inspec/reporters/json.rb
jerryaldrichiii: Yup, good point. Will fix.

test/functional/inspec_exec_json_test.rb
jerryaldrichiii: I prefer, label. I'll go with that.

clintoncwolfe (approved):
Thanks, Jerry!

jquick (approved):
Looks good to me! Thanks @jerryaldrichiii

PR #3425, opened 2018-09-21
Support the Busybox variant of netstat in the port resource

### LINKED ISSUE #2890: InSpec won't admit that port 5432 is listening inside a PostgreSQL container
# Background
I'm trying to test a [standard PostgreSQL container][1] with InSpec 2.1.10, but I cannot get a simple port check to return true, even though Docker, netstat, and psql all report that the port is properly mapped. While it's possible this is a bug in InSpec, it's also possible that the problem exists between the keyboard and the chair.

# Repeatable Tests
## My Spec

    # port_test.rb
    describe port(5432) do
      it { should be_listening }
    end

## My Testing Commands

    container=$(docker run -d -P --rm postgres:alpine)
    inspec exec port_test.rb -t docker://$container

## But It Should Work!
Everything but InSpec reports that things are working as expected.

    $ docker port $container
    5432/tcp -> 0.0.0.0:32770

    # from inside container using netstat
    $ docker exec $container netstat -lntp
    Active Internet connections (only servers)
    Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
    tcp        0      0 0.0.0.0:5432            0.0.0.0:*               LISTEN      -
    tcp        0      0 :::5432                 :::*                    LISTEN      -

    # on macOS Docker host; flags are the macOS equivalent to `-lntp`
    $ netstat -anv | fgrep $(docker port $container | cut -d: -f2)
    tcp6       0      0  ::1.32770              *.*                    LISTEN      131072 131072    709      0
    tcp4       0      0  *.32770                *.*                    LISTEN      131072 131072    709      0

    # connecting to database in container from host
    $ psql -V \
           -h localhost \
           -p $(docker port $container | cut -d: -f2)
    psql (PostgreSQL) 10.3

  [1]: https://hub.docker.com/_/postgres/
  [2]: https://www.inspec.io/

hartmantis:
This would fix #2890.

The Busybox variant of netstat doesn't have a User or Inode column in the
output, resulting in issues with the port resource in Habitat or Alpine Linux
when the iproute2 package isn't installed and the ss command isn't there.

Signed-off-by: Jonathan Hartman <j@hartman.io>

### DIFF
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -473,22 +473,26 @@ module Inspec::Resources
 
     def parse_netstat_line(line)
       # parse each line
-      # 1 - Proto, 2 - Recv-Q, 3 - Send-Q, 4 - Local Address, 5 - Foreign Address, 6 - State, 7 - Inode, 8 - PID/Program name
-      parsed = /^(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)?\s+(\S+)\s+(\S+)\s+(\S+)/.match(line)
+      # 1 - Proto, 2 - Recv-Q, 3 - Send-Q, 4 - Local Address, 5 - Foreign Address, 6 - State, 7 - User, 8 - Inode, 9 - PID/Program name
+      # * UDP lines have an empty State column and the Busybox variant lacks
+      # the User and Inode columns.
+      reg =  /^(?<proto>\S+)\s+(\S+)\s+(\S+)\s+(?<local_addr>\S+)\s+(?<foreign_addr>\S+)\s+(\S+)?\s+((\S+)\s+(\S+)\s+)?(?<pid_prog>\S+)/
+      parsed = reg.match(line)
+
       return {} if parsed.nil? || line.match(/^proto/i)
 
       # parse ip4 and ip6 addresses
-      protocol = parsed[1].downcase
+      protocol = parsed[:proto].downcase
 
       # detect protocol if not provided
-      protocol += '6' if parsed[4].count(':') > 1 && %w{tcp udp}.include?(protocol)
+      protocol += '6' if parsed[:local_addr].count(':') > 1 && %w{tcp udp}.include?(protocol)
 
       # extract host and port information
-      host, port = parse_net_address(parsed[4], protocol)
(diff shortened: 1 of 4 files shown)


hartmantis:
I'm trying to figure out the build failures. It's evidently something I did, but all the tests pass locally. :\

hartmantis:
Okay, it's down to that intermittently failing`test_load_mock_plugin_by_gem`, which seems to be happening to other builds as well.

clintoncwolfe:
Thanks for your patience.  We have the (unrelated) plugin test failure fixed on master now, so if you rebase, that should go away.  That said, we can review without a rebase, too.

lib/resources/port.rb
```diff
@@ -473,22 +473,26 @@ def parse_net_address(net_addr, protocol)
 
     def parse_netstat_line(line)
       # parse each line
-      # 1 - Proto, 2 - Recv-Q, 3 - Send-Q, 4 - Local Address, 5 - Foreign Address, 6 - State, 7 - Inode, 8 - PID/Program name
-      parsed = /^(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)?\s+(\S+)\s+(\S+)\s+(\S+)/.match(line)
+      # 1 - Proto, 2 - Recv-Q, 3 - Send-Q, 4 - Local Address, 5 - Foreign Address, 6 - State, 7 - User, 8 - Inode, 9 - PID/Program name
+      # * UDP lines have an empty State column and the Busybox variant lacks
+      # the User and Inode columns.
+      reg =  /^(?<proto>\S+)\s+(\S+)\s+(\S+)\s+(?<local_addr>\S+)\s+(?<foreign_addr>\S+)\s+(\S+)?\s+((\S+)\s+(\S+)\s+)?(?<pid_prog>\S+)/
```
clintoncwolfe, line 479: :100: for using named captures in a long regex!

clintoncwolfe (approved):
This looks great.  Improved regex clarity, the functionality you aimed for, and great test coverage.  A model PR!  Thank you so much for your contribution!

jquick (approved):
Thanks @RoboticCheese

Issue #3429, opened 2018-09-21
Improve Integration Testing with Audit Cookbook


clintoncwolfe:
### 🎛 Description

🐛 bug report

The InSpec project does not currently provide any testing ensuring that integration with the audit cookbook is working.

The audit cookbook is one of the main ways people consume inspec.  Lack of integration testing creates an undelightful experience for our largest community.

Refs #3373.

### 🌍 InSpec and Platform Version

2.2.100

jquick:
The audit cookbook also directly requires some of our plugins. 
https://github.com/chef-cookbooks/audit/blob/master/files/default/handler/audit_report.rb#L122

Adding some travis tests to test audit cookbook and the export would be ideal.

jquick:
This is done.

PR #3431, opened 2018-09-22
Add audit-cookbook integration testing

jquick:

This PR does a few things:
1) It builds the current checked out inspec folder into a gem
2) Update the os_prepare cookbook to install our custom local inspec gem into chef_gem bundle
3) Add audit cookbook to run a profile during execution
4) Add some verifier controls to test the audit json export is working as expected.
5) Add some TestKitchen and Audit attributes to make sure they are passing down as expected.

This will take care of the linux portion of #3429


test/cookbooks/os_prepare/recipes/default.rb
```diff
@@ -4,6 +4,18 @@
 #
 # prepare all operating systems with the required configuration
 
+
+# inject the current inspec gem for use with audit cookbook
+# this is generated via Rake test:integration
+cookbook_file '/root/inspec-core-local.gem' do
```
clintoncwolfe, line 10: This is awesome.  Only thought: could we include the inspec version (or current git commit SHA hash) in the name of the gemfile?  Might help verify we're running on what we think we are.

.kitchen.yml
```diff
@@ -77,10 +79,23 @@ platforms:
...
   attributes:
+    audit:
+      attributes:
+        audit_attribute: 'Attribute Override!'
+      reporter: 'json-file'
+      json_file:
+        location: /tmp/json_export.json
+      profiles:
+        - name: integration
+          url: https://github.com/jquick/profile_with_attributes/archive/master.zip
```
clintoncwolfe: Could we source a profile out of the inspec/inspec repo, instead of jquick's personal repos? :P)

test/integration/default/controls/audit_spec.rb
```diff
@@ -0,0 +1,43 @@
+# encoding: utf-8
```
clintoncwolfe, line 1: It would be great to have (somewhere) a dev doc about testing the audit cookbook.  It could be comments here, but I'd look at docs/dev/ first.

clintoncwolfe (commented):
This is fantastic to have!

test/cookbooks/os_prepare/recipes/default.rb
jquick: Yeah there should be a way to dynamic that. We do also test that the version outputted in the audit json_report matches the verifier version.

.kitchen.yml
jquick: For sure 😆

test/cookbooks/os_prepare/recipes/default.rb
jquick: Dug into this a bit. Looks like we would have to do a dynamic attribute and some searching around. Since it installs with the correct version on install and we validate that in the checks I am ok with this for the first pass.

miah (approved):
Awesome. Thanks @jquick

Rakefile
```diff
@@ -106,10 +106,23 @@ namespace :test do
     sh(Gem.ruby, 'test/docker_test.rb', *tests)
   end
 
-  task :integration do
+  task :integration, [:os] do |task, args|
     concurrency = ENV['CONCURRENCY'] || 1
-    os = ENV['OS'] || ''
-    sh("bundle exec kitchen test -c #{concurrency} #{os}")
+    os = args[:os] || ENV['OS'] || ''
+    ENV['DOCKER'] = 'true' if ENV['docker'].nil?
```
jerryaldrichiii, line 112: This is interesting. Why isn't this `unless`?

Rakefile
```diff
@@ -106,10 +106,23 @@ namespace :test do
...
+    puts "Building current InSpec gem for audit cookbook testing..."
+    output = %x[gem build inspec-core.gemspec]
+    puts output
```
jerryaldrichiii, line 115: Do we need this puts?

test/integration/default/controls/audit_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    its(['platform', 'name']) { should eq platform.name }
+    its(['statistics', 'duration']) { should  > 0 }
+    its('version') { should cmp inspec_version }
+  end
+end
+
+# test kitchen verify attr passthrough
+attr = attribute('verifier_attribute', default: 'none')
+control 'validate verifier attribute override' do
+  describe attr do
```
jerryaldrichiii, line 30: Hmmm...thought `attr` was a reserved word...guess not though.

jerryaldrichiii (approved):
Looks good to me, only a few questions.

Rakefile
jquick: yes so we know what was built on cli output

Rakefile
jquick: yeah that would also be fine as well.

jerryaldrichiii: I don't think I am clear.

Why if `ENV['docker'].nil?` do we set `ENV['DOCKER']` to `true`.

PR #3439, opened 2018-09-25
Grammar correction in error message: use "an" with attribute and unknown

alexpop:


jquick (approved):
Thanks @alexpop

clintoncwolfe (approved):
Thanks, great catch!

PR #3446, opened 2018-09-26
Support finding larger processes on Busybox

hartmantis:
For larger processes, Busybox's ps displays the vsz and rss columns in
megabytes or gigabytes, with no option I've found to override the behavior.

This change updates the process regex to account for that and converts
the values to kilobytes so they can still be cast as integers.

Signed-off-by: Jonathan Hartman <j@hartman.io>


jquick (approved):
Nice addition! Thanks @RoboticCheese

jerryaldrichiii (approved):
Great work! Glad to see you at the Summit as well :smile:

PR #3450, opened 2018-09-27
Modify `cmp` matcher output to use `.inspect`

jerryaldrichiii:
This modifies the output of failed test when using the `cmp` matcher by calling `.inspect` on the actual value passed.

This makes it easier to diagnose failed matches due to characters such as `\n` not being visible.

Current behavior:
```
inspec> describe command('echo demo') do
inspec>   its('stdout') { should cmp 'demo' }
inspec> end

Profile: inspec-shell
Version: (not specified)

  Command: `echo demo`
     ×  stdout should cmp == "demo"

     expected: "demo"
          got: demo

     (compared using `cmp` matcher)

```

New behavior:
```
inspec> describe command('echo demo') do
inspec>   its('stdout') { should cmp 'demo' }
inspec> end

Profile: inspec-shell
Version: (not specified)

  Command: `echo demo`
     ×  stdout should cmp == "demo"

     expected: "demo"
          got: "demo\n"

     (compared using `cmp` matcher)

```

Many thanks to @jazaval for discovering this!

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -294,13 +294,13 @@ RSpec::Matchers.define :cmp do |first_expected| # rubocop:disable Metrics/BlockL
   end
 
   failure_message do |actual|
-    actual = ('0' + actual.to_s(8)).inspect if octal?(@expected)
-    "\n" + format_expectation(false) + "\n     got: #{actual}\n\n(compared using `cmp` matcher)\n"
+    actual = ('0' + actual.to_s(8)) if octal?(@expected)
+    "\n" + format_expectation(false) + "\n     got: #{actual.inspect}\n\n(compared using `cmp` matcher)\n"
   end
 
   failure_message_when_negated do |actual|
     actual = ('0' + actual.to_s(8)).inspect if octal?(@expected)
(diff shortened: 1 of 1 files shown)


jquick (approved):
Nice addition @jerryaldrichiii !

trevor-vaughan:
I think this is great!

Would you be up for printing the class name as well? That would be *really* valuable for telling the difference between `true` and `'true'` for instance.

jerryaldrichiii:
@trevor-vaughan that opens up a larger discussion. I think this is a small enough change that it can go through with minimal impact. We are close with this change though.

See:
```
inspec> describe true do; it { should cmp 'true' }; end

Profile: inspec-shell
Version: (not specified)

  true
     ✔  should cmp == "true"

```
```
inspec> describe 'true' do; it { should cmp true }; end

Profile: inspec-shell
Version: (not specified)

  true
     ×  should cmp == true
     
     expected: true
          got: "true"
     
     (compared using `cmp` matcher)


```

trevor-vaughan:
@jerryaldrichiii Sorry it took me so long to reply. Your explanation makes sense, thanks for providing the example!

miah (approved):
Thanks @jerryaldrichiii

PR #3451, opened 2018-09-28
Update the logic to find local installed gems

jquick:

This change creates a valid VendorSet of all Gems the instance knows about. This format is needed when passing into the `Gem::Resolver.compose_sets`.

This was a rush implementation but wanted to get something others can look at. I will split the class out and clean it up tomorrow.


jquick:
This still needs some work, as its not always blocking installs. Closing for now till I flush it out.

PR #3453, opened 2018-09-28
Update plugin gem install code

jquick:
This change adds some checks around the gem install code and also fixes the resolver bug.

1) From what I can gather the resolve code just confirms we have dependencies available for the install.
2) Added a activation section to confirm we can activate the gem on top of inspec + current plugins.
3) Added some error validation for them and testing.


clintoncwolfe (approved):
Only thing I saw was to loosen the match on the rake conflict.  I manually tested, and it worked well!  Thanks!

miah (approved):
Thanks for the zoom review on this. Looks great =)

PR #3454, opened 2018-09-28
Fix `attribute` with empty hash regression

### LINKED ISSUE #3442: Version 2.2.101 introduced breaking changes to profile attributes
### 🎛 Description

This is a 🐛 bug report.

#3318 changed the implementation of profile attributes to require a definition in the associated profile file. This was a breaking change as attributes previously did not need to be defined before being used.

### 🌍 InSpec Version

2.2.101

### 🤔 Replication Case

Use InSpec 2.2.78 to execute a profile named `test_profile` which uses an attribute named `test_attribute` that is not defined in the profile file and see it succeed. 

Use InSpec 2.2.101 to execute the same profile and see it fail with an error like `Profile 'test_profile from {:path=>"tests/inspec"}' does not have a attribute with name 'test_attribute'`.

Define `test_attribute` in the profile file, replicate the previous step, and see it succeed.

### 💁 Possible Solutions

Implement backward compatible logic so that undefined attributes are still valid until InSpec 3.0.0 is released.

jerryaldrichiii:
This will allows for using the below in controls again:

```ruby
attribute('some_attribute', {})
```

This closes https://github.com/inspec/inspec/issues/3442

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -142,8 +142,8 @@ module Inspec
         end
 
         # method for attributes; import attribute handling
-        define_method :attribute do |name, options = {}|
-          if options.empty?
+        define_method :attribute do |name, options = nil|
+          if options.nil?
             Inspec::AttributeRegistry.find_attribute(name, profile_id).value
           else
             profile_context_owner.register_attribute(name, options)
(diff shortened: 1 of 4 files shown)


jquick:
Thanks @jerryaldrichiii, i think you just need to bump the success number for the attributes_test to fix the unit test.

miah (approved):

clintoncwolfe (approved):
Looks good - thanks!

PR #3463, opened 2018-10-03
Fix distinct_exit cli desc to reflect reality

teknofire:
The description for `distinct_exit` had the exit codes backwards.

From the long desc
> If some tests skipped but none failed, exit code 101 is returned. If at least one test failed, exit code 100 is returned.

Signed-off-by: Will Fisher <wfisher@chef.io>


jerryaldrichiii (approved):
Great catch @teknofire!

jquick (approved):
Thanks @teknofire !

PR #3467, opened 2018-10-05
Add missing tests for groups resource, document members property, and assorted fixes.

### LINKED ISSUE #3422: Add group members as a property for the group resource
### 🎛 Description

 🙋 feature request

For the group resource I would like to test that a user is part of a group from looking at the docs it does not expose the list of members. I could spend some time looking into adding it but I will need help if its gonna support windows. For now I am testing with a block like this on linux:
```
inspec> describe command("awk -F':' '/docker/{print $4}' /etc/group") do
inspec>   its('stdout') { should match(/babrams/) }  
inspec> end  

Profile: inspec-shell
Version: (not specified)

  Command: `awk -F':' '/docker/{print $4}' /etc/group`
     ✔  stdout should match /babrams/

```
We should be able to grab the list of members of the group and expose it as a property to be tested against. 

Briefly describe the issue and the expected behavior. Also, please be aware of our [Code of Conduct](https://github.com/inspec/inspec/blob/master/CODE_OF_CONDUCT.md).

Platform:
    Name:      ubuntu
    Release:   18.04
    Arch:      x86_64

Running versions (multiple projects with multile contexts):
- `2.2.102`
- `2.2.112`

### 💁 Possible Solutions

For linux we could look through `/etc/groups` like so:
```
# docker is just an example group
# and awk is just an example of how to parse we would do it in native ruby
awk -F':' '/docker/{print $4}' /etc/group
```
There may very well be a better option than this I just have not had enough time to look at it.

No idea on windows

miah:
Please let me know if I missed anything. This should improve tests around the group resource and fix missing documentation as noted in #3422 

Changes:
Update existing documentation for group resource.
Add documentation for groups resource.
Update group resource tests to test members property.
Change groups resource members property to have simple style. (this
ensures members is a single array)
remove deprecated have_gid propery.
change `if !` to `unless`
Remove early return from members method. This prevented members from
working correctly on any OS other than Windows.
Add missing tests for the groups resource.

Fixes https://github.com/inspec/inspec/issues/3422

### DIFF
--- a/docs/resources/groups.md.erb
+++ b/docs/resources/groups.md.erb
@@ -1,0 +1,85 @@
+---
+title: About the groups Resource
+platform: os
+---
+
+# group
+
+Use the `groups` InSpec audit resource to test multiple groups on the system.
+
+<br>
+
+## Availability
+
+### Installation
+
+This resource is distributed along with InSpec itself. You can use it automatically.
+
+### Version
+
+This resource first became available in v1.0.0 of InSpec.
+
+## Syntax
+
+A `groups` resource block uses `where` to filter entries from the systems groups. If `where` is omitted, all entries are selected.
+
+    describe groups do
+      its('names') { should eq ['wheel', 'daemon', 'sys', 'adm'] }
+      its('names') { should include 'wheel' }
+    end
+
+    describe groups.where { members =~ /root/ } do
+      its('names') { should eq ['wheel', 'daemon', 'sys', 'adm'] }
+    end
+
+<br>
+
+## Examples
+
+The following examples show how to use this InSpec audit resource.
+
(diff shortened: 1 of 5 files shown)


docs/resources/groups.md.erb
```diff
@@ -0,0 +1,85 @@
+---
+title: About the groups Resource
+platform: os
+---
+
+# group
```
jerryaldrichiii: Should be `groups`, right?

jerryaldrichiii (changes requested):
Just one tiny change.

jquick (approved):
Thanks @miah

jerryaldrichiii (approved):

PR #3468, opened 2018-10-05
Fix archive with required attributes

### LINKED ISSUE #3464: Cannot archive profile when attributes have no default attribute
### 🎛 Description

this a 🐛 bug report

When using unified attributes (I think applies to normal attribute behaviour too, but not tested) without default values, and required, the profile cannot be archived with `inspec archive` 

Use https://github.com/dev-sec/cis-dil-benchmark/pull/55 and change the inspec.yml to match (no default value provided)

```yml
attributes:
  - name: cis_level
    required: true
    description: 'CIS profile level to audit'
    type: numeric
```

Then InSpec returns the following error

```bash
$ inspec archive cis-dil-benchmark
Attribute 'cis_level' is required and does not have a value.
```

InSpec 2.3.8 (3248af1f)

### 🤔 Replication Case

As described above.

### 💁 Possible Solutions

ignore missing attributes when used in archive mode 

### 💻 Stacktrace

n/a

jquick:
This allows profiles to be vendored and archived when they have required attributes.

Fixes #3464

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -89,6 +89,9 @@ module Inspec
     private
 
     def validate_required(value)
+      # skip if we are not doing a exec (archive/vendor)
+      return unless Inspec::BaseCLI.inspec_cli_command == :exec
+
       # value will be set already if a secrets file was passed in
       if (!@opts.key?(:default) && value.nil?) || (@opts[:default].nil? && value.nil?)
         error = Inspec::Attribute::RequiredError.new
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -89,7 +89,7 @@ module Inspec
     private
 
     def validate_required(value)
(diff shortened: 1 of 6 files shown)


miah (approved):

lib/inspec/objects/attribute.rb
```diff
@@ -89,6 +89,9 @@ def to_s
     private
 
     def validate_required(value)
+      # skip if we are not doing a exec (archive/vendor)
```
jerryaldrichiii: Should be `an exec`, I also think the below is easier to understand but I'll leave that to you:

```
# Skip if we are using `inspec exec` (`inspec archive/vendor` is okay)

jerryaldrichiii (changes requested):

lib/inspec/objects/attribute.rb
jerryaldrichiii: Whoops...I got this backwards. You've got it right.

jerryaldrichiii (approved):

PR #3469, opened 2018-10-05
Fix Packages Resource Docs

pwelch:
Some instances of the packages resource in the examples
were referencing package instead of packages.

Signed-off-by: Paul Welch <pwelch@chef.io>


jerryaldrichiii (approved):
Great catch!

jquick (approved):
Thanks @pwelch

PR #3471, opened 2018-10-05
Exclude docs and examples from the gem

tas50:
This greatly slims down the installed size of inspec in both Chef and
Chef-DK. This translates into faster installs which are of importance to
workstation users, but also smaller download sizes / CDN bills.

This slims the installed on disk from 2.9 MB to 2.0 MB (31%) and the gem
drops from 567k to 396k (30%).


miah (approved):
Thanks for the cleanup @tas50!

jerryaldrichiii (approved):
Much cleaner, thanks @tas50!

jquick (approved):
Thanks @tas50, looks good!

PR #3472, opened 2018-10-08
Enable compression for deb/rpm packages

tas50:
In Chef we've seen this shrink package sizes by 22%. Should save us some sweet $$$ on the CDN costs.


miah:
I believe we are still supporting RHEL5, so we'll need to handle that.

miah (approved):
Thanks @tas50 !

jquick (approved):
Thanks @tas50 !!

PR #3475, opened 2018-10-08
Remove 'demo' from website.

miah:
We weren't maintaining this well and it needed lots of updates. It served its purpose but now its time to remove it.


jerryaldrichiii (approved):

jquick (approved):
This looks good. Was able to generate the site and everything worked as expected except with no demo!

PR #3481, opened 2018-10-09
FilterTable: allow Strings or Symbols as fields

clintoncwolfe:
Allow strings or symbols to be used interchangeably in filtertable criteria.  Also, allow resources to use strings as fields for raw data (but please don't do that - look at you, `shadow` resource).

Also adds stuff to an error message to clarify expected types.

Functional tests to cover the repro case included.

Reported by @james-stocks

### DIFF
--- a/lib/utils/filter.rb
+++ b/lib/utils/filter.rb
@@ -106,7 +106,7 @@ module FilterTable
       # If we were provided params, interpret them as criteria to be evaluated
       # against the raw data. Criteria are assumed to be hash keys.
       conditions.each do |raw_field_name, desired_value|
-        raise(ArgumentError, "'#{raw_field_name}' is not a recognized criterion - expected one of #{list_fields.join(', ')}'") unless field?(raw_field_name)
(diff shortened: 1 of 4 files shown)


lib/utils/filter.rb
```diff
@@ -175,7 +175,15 @@ def field?(proposed_field)
       # Currently we only know about a field if it is present in a at least one row of the raw data.
       # If we have no rows in the raw data, assume all fields are acceptable (and rely on failing to match on value, nil)
       return true if raw_data.empty?
-      list_fields.include?(proposed_field) || is_field_lazy?(proposed_field)
+
+      # Most resources have Symbol keys in their raw data.  Some have Strings (looking at you, `shadow`).
```
james-stocks, line 179: I also noticed that if you create your filter table data via "JSON.parse(some_api_response)" you'll get string keys.
I was using ` @data.map! { |entry| Hash[entry.map { |k,v| [k.to_sym, v] }] }` to get the keys to symbols.

james-stocks (approved):
Thanks Clinton!

miah (approved):
Thanks @clintoncwolfe

jerryaldrichiii (approved):

Issue #3466, opened 2018-10-04
null attributes section in inspec.yml produces error in check and exec


aaronlippold:
### 🎛 Description
 🐛 bug report 
v2.3.5

If I create an inspec.yml file with just the 

```
attributes:
``` 
as a place-holder - inspec check gives ...
```
/Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:130:in `register_metadata_attributes': undefined method `each' for nil:NilClass (NoMethodError)
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:125:in `initialize'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:67:in `new'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:67:in `for_path'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:73:in `for_fetcher'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/profile.rb:79:in `for_target'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/inspec-2.3.5/lib/inspec/cli.rb:79:in `check'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/command.rb:27:in `run'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor.rb:387:in `dispatch'
	from /Users/aaronl/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/thor-0.20.0/lib/thor/base.rb:466:in `start'
[...]

PR #3485, opened 2018-10-10
Fix error on empty attributes yaml

jquick:

Fixes https://github.com/inspec/inspec/issues/3466

This makes sure the `attributes` param is an array before trying digest it.

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -132,6 +132,8 @@ module Inspec
           name = attr_dup.delete(:name)
           @runner_context.register_attribute(name, attr_dup)
         end
+      elsif metadata.params.key?(:attributes)
+        Inspec::Log.warn 'Attributes must be defined as an Array. Skipping current definition.'
       end
     end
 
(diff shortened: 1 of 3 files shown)


clintoncwolfe (commented):
Just a concern about maybe issuing a warning.

lib/inspec/profile.rb
```diff
@@ -126,7 +126,7 @@ def initialize(source_reader, options = {})
     end
 
     def register_metadata_attributes
-      if metadata.params.key?(:attributes)
+      if metadata.params.key?(:attributes) && metadata.params[:attributes].is_a?(Array)
```
clintoncwolfe, line 129: (lost comment)

I think we should issue a warning if the `attributes` key exists and is a hash or string.  I can't imagine what a string would mean, but since attributes are key-values, a user might assume a Hash is wanted here.

miah (approved):
Thanks @jquick

lib/inspec/profile.rb
jquick: Agreed, I think it will have to be in a different PR as my capacity is full for 3.0. I will make a issue for it.

jerryaldrichiii:
+1 to @clintoncwolfe's comment, I think we should have a warning and a set of test cases for it.

jerryaldrichiii (approved):

PR #3484, opened 2018-10-10
Adding inspec init profile for GCP.

skpaterson:
An attempt to address https://github.com/inspec/inspec-gcp/issues/61 and give GCP users a nice profile generator. 

Signed-off-by: Stuart Paterson <spaterson@chef.io>

PR #3491, opened 2018-10-11
Init: Make profile templates platform specific

clintoncwolfe:
Refs https://github.com/inspec/inspec-gcp/issues/61
Refs #3484 

Alters `inspec-init` plugin in the following ways:
 * The `profile` subcommand is now implemented explicitly.
 * `profile` now takes a `--platform` argument, which defaults to `os`.  It causes the templater to look in a subdirectory.
 * Help message altered to dynamically list available platforms
 * Errors with code 1 if bad platform specified
 * Default template now has an explicit `supports: os` addition in the metadata, to help give a more useful error message.
 * Functional tests updated to include `--platform` option

If accepted, this PR should be merged prior to #3484


clintoncwolfe:
This is included in #3484

PR #3492, opened 2018-10-11
Update iis_site bindingInformation construction and add tests

### LINKED ISSUE #3490: iis_site: bindings property not populated correctly
### 🎛 Description

🐛 Bug Report!

There's a slight hiccup in the `SiteProvider::iis_site` method for the `iis_site` resource. The way that the binding information string is constructed uses the `<<` operator, which ends up modifying the `protocol` property, which in turn causes a later ternary expression involving the `protocol` property to always return false.

### 🤔 Replication Case

On a Windows Server 2012+ server with IIS installed, create an HTTPS binding on port 443 on the Default Web Site (doesn't have to actually be bound to any certificate). Then create a small Inspec test with the following assertion:

    control 'any-control' do
      describe iis_site('Default Web Site') do
        it { should have_binding("https *:443:") }
        it { should have_binding("https *:443: sslFlags=0") }
      end
    end

The first assertion should fail and the second succeed because HTTPS binding information should contain the `sslFlags` property; instead, the first succeeds and the second fails.

### 💁 Possible Solutions

I will have a PR ready very soon; just wanted to create the issue first to put in the commit message.

mrshanahan:
Signed-off-by: Matt Shanahan <mrshanahan11235@gmail.com>

I have not made any changes to the docs, since this brings the code in line with the documentation. Also added some unit tests to the iis_site resource.

Relevant resource docs: https://github.com/inspec/inspec/blob/master/docs/resources/iis_site.md.erb#L109-L139

Fixes #3490 
EDIT (@clintoncwolfe): Move reference to issue from PR title to description to trigger close-on-merge.

### DIFF
--- a/lib/resources/iis_site.rb
+++ b/lib/resources/iis_site.rb
@@ -94,7 +94,7 @@ module Inspec::Resources
 
     # want to populate everything using one powershell command here and spit it out as json
     def iis_site(name)
-      command = "Get-Website '#{name}' | select-object -Property Name,State,PhysicalPath,bindings,ApplicationPool | ConvertTo-Json"
+      command = "Get-Website '#{name}' | Select-Object -Property Name,State,PhysicalPath,bindings,ApplicationPool | ConvertTo-Json"
       cmd = @inspec.command(command)
 
       begin
@@ -103,11 +103,8 @@ module Inspec::Resources
         return nil
       end
 
-      bindings_array = site['bindings']['Collection'].map { |k, _str|
-        k['protocol'] <<
-          ' ' <<
-          k['bindingInformation'] <<
-          (k['protocol'] == 'https' ? ' sslFlags=' << flags : '')
+      bindings_array = site['bindings']['Collection'].map { |k|
+        "#{k['protocol']} #{k['bindingInformation']}#{k['protocol'] == 'https' ? " sslFlags=#{k['sslFlags']}" : ''}"
       }
 
       # map our values to a hash table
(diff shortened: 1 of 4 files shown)


aaronlippold:
hi, given that you document at least twice that you only support 2012+, would it be a good idea to add that as part of the `platforms` lotic in the `init` method of the resource?

mrshanahan:
@aaronlippold That makes sense! Let me take a look.

miah:
@mrshanahan, @aaronlippold 

We're moving away from doing platform support restrictions in initialize. This should all be handled by the [supports](https://www.inspec.io/docs/reference/dsl_resource/) dsl now.

We could do something like:

```
supports platform: 'windows', release: '6.2.*'
```

Where `release` is the right version number =) Note that you can add multiple supports, so you should be able to cover all the platforms.

aaronlippold:
I agree for the profile for iis. But for the initialization of the iis resource itself shouldn't we still have that fail fast?

mrshanahan:
I'm happy to add the appropriate `supports` statements to narrow down the acceptable list of platforms, but I'm not sure how to find what the available platform names are. Can I select e.g. `windows_server` instead of `windows`? I had some trouble finding the relevant code.

jquick:
@mrshanahan , The windows specific names ones are a bit dynamic and end up being something like `windows_server_2012_r2_standard_evaluation`. I would stick with something like what @miah suggested. The release portion should contain it only to the ones we want.

We use this command to find the windows release https://github.com/inspec/train/blob/master/lib/train/platforms/detect/helpers/os_windows.rb#L6

clintoncwolfe (approved):
Thanks so much! This is a great first contribution, and we really appreciate it.  There is some concern about restricting the releases on which the resource will run, but your PR is independent of that concern; I've opened #3591 to track needing to tighten the platform support, and that clears the way to get this merged in.

Thanks!

jquick (approved):
Thanks @mrshanahan

mrshanahan:
Thank you!

Issue #3488, opened 2018-10-10
Add skip metadata for json-automate report


jquick:
1. if a dependent profile was not executed => flag it as such and report it
2. if the profile the user executed is not supported => flag it as well and report it

For both of these cases we are going to add `skipped` and `skip_reason` fields.

For #1
we were thinking like:
  ```"profiles": [
    {
      "name": "unsupported_dependencies_wrapper",
      "version": "0.1.0",
      "sha256": "932026cd66e7407f67f9f03d874865ce60b77dad1b168acd5dfa54b9c97c2f16",
      "title": "InSpec Profile",
      "depends": [
        {
          "name": "child_profile",
          "path": "../child_profile",
          "status": "skipped",
          "skip_message: "Skipping profile: 'child_profile' on unsupported platform: 'mac_os_x'"
        }
```

For #2 
 ```"profiles": [
    {
      "name": "unsupported_dependencies_wrapper",
      "status": "skipped",
      "skip_message": "Skipping profile: 'child_profile' on unsupported platform: 'mac_os_x'"
  }
```

PR #3495, opened 2018-10-11
Add json metadata for skipped profiles

jquick:
This change adds the new `status` and optional `skip_message` to the json reports. It changes the behavior when a main profile is not supported. Instead of exiting error with a message we will output that we have skipped the profile and continue as expected.

Fixes https://github.com/inspec/inspec/issues/3488

### DIFF
--- a/test/functional/inspec_exec_test.rb
+++ b/test/functional/inspec_exec_test.rb
@@ -161,9 +161,9 @@ Test Summary: 0 successful, 0 failures, 0 skipped
     let(:out) { inspec('exec ' + File.join(profile_path, 'skippy-profile-os') + ' --no-create-lockfile') }
     let(:json) { JSON.load(out.stdout) }
 
-    it 'exits with an error' do
-      out.stderr.must_match(/^This OS\/platform \(.+\) is not supported by this profile.$/)
-      out.exit_status.must_equal 1
+    it 'exits with skip message' do
+      out.stdout.must_include("Skipping profile: 'skippy' on unsupported platform:")
+      out.exit_status.must_equal 0
     end
   end
 
--- a/test/functional/inspec_exec_test.rb
+++ b/test/functional/inspec_exec_test.rb
@@ -163,7 +163,7 @@ Test Summary: 0 successful, 0 failures, 0 skipped
 
     it 'exits with skip message' do
       out.stdout.must_include("Skipping profile: 'skippy' on unsupported platform:")
-      out.exit_status.must_equal 0
+      out.exit_status.must_equal 101
     end
   end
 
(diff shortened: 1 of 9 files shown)


test/unit/mock/profiles/unsupported_dependencies/wrapper-profile/inspec.yml
```diff
@@ -9,5 +9,4 @@ version: 0.1.0
 depends:
   - name: child_profile
     path: ../child_profile
-  - name: child_profile2
-    path: ../child_profile2
+  - path: ../child_profile2
```
jerryaldrichiii, line 12: We lost `name:` here

test/functional/inspec_exec_test.rb
```diff
@@ -161,9 +161,9 @@
     let(:out) { inspec('exec ' + File.join(profile_path, 'skippy-profile-os') + ' --no-create-lockfile') }
     let(:json) { JSON.load(out.stdout) }
 
-    it 'exits with an error' do
-      out.stderr.must_match(/^This OS\/platform \(.+\) is not supported by this profile.$/)
-      out.exit_status.must_equal 1
+    it 'exits with skip message' do
+      out.stdout.must_include("Skipping profile: 'skippy' on unsupported platform:")
+      out.exit_status.must_equal 0
```
jerryaldrichiii: Thought we had a different exit code for skips?

test/unit/mock/profiles/unsupported_dependencies/wrapper-profile/inspec.yml
jquick: yeah it was deliberate to make sure we work if they don't pass a name.

test/functional/inspec_exec_test.rb
jquick: only if we skipped tests not whole profiles.

jerryaldrichiii (approved):

test/functional/inspec_exec_test.rb
clintoncwolfe: I'm not sure if I agree with that.  If you skip a profile, you're almost certainly skipping tests.  I think this should be the "clean run with skip" exit code, either 100 or 101.

jquick: The issue here is if we are skipping tests (with only_if) we actually count them and show that we skipped (n) tests. If we skip the main profile we output an empty report for digest with 0 skips. @arlimus - I think we chatted about this at one point if you could confirm.

miah (approved):
Solid. Thanks @jquick

test/functional/inspec_exec_test.rb
jquick: I think we can make this 101, ill see if i can slide that in.

PR #3498, opened 2018-10-12
Update profiles.md

aaronherman:
Provide more clarity about the depends path attribute.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -178,7 +178,7 @@ InSpec supports a number of dependency sources.
 
 ### path
 
-The `path` setting defines a profile that is located on disk. This setting is typically used during development of profiles and when debugging profiles.
+The `path` setting defines a profile that is located on disk. This setting is typically used during development of profiles and when debugging profiles. The profile path should be somewhere other than the`controls`directory. For example, place them in a directory called `profiles` at the same level as `controls`.
 
 ```YAML
 depends:


jerryaldrichiii:
Thank you for your first PR @aaronherman!

I like the idea you're getting at about keeping the profiles out of `controls/`. However, I would recommend placing it outside the profile they're working on entirely so the other profiles don't get accidentally committed. 

Also, as of the most recent InSpec versions, the profiles that are mentioned via `path:` will be placed in `vendor/` when using `inspec vendor` or `inspec archive` so having them in the same folder could get confusing.

How about?:

> The `path` setting defines a profile that is located on disk. This setting is typically used during development of profiles and when debugging profiles. It is recommended that the referenced profile be placed outside of the profile currently being developed (if you use `inspec archive` or `inspec vendor` then the referenced profile will be placed in the `vendor` directory for you).

jerryaldrichiii:
Also, could you do a Developer Certified Origin sign off on your commit? See: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

TL;DR: `git commit --amend -s`

Or, since this is such a small change you can just add the words `Obvious fix` to your commit message. See: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#chef-obvious-fix-policy

I'll leave it up to you :smile:

PR #3507, opened 2018-10-12
Remove need for UUID detection for Automate report

jerryaldrichiii:
This reverts the reversion here: https://github.com/inspec/inspec/pull/3387


jquick (approved):
Thanks @jerryaldrichiii

miah (approved):

Issue #3443, opened 2018-09-26
Confusing warning message and beahavior for inspec attributes


chris-rock:
### 🎛 Description

This is a discussion about the right behaviour. We have an example profile that requires attributes where `val_user` has a default value, but password has not.

```ruby
# encoding: utf-8
val_user = attribute('user', default: 'alice', description: 'An identification for the user')
val_password = attribute('password', description: 'A value for the password')

describe val_user do
  it { should eq 'bob' }
end

describe val_password do
  it { should eq 'secret' }
end
```

The following happens

```bash
$ inspec exec examples/profile-attribute
[2018-09-26T13:54:17+02:00] WARN: Returning DEFAULT_ATTRIBUTE for 'password'. Use --attrs to provide a value for 'password' or specify a default  value with `attribute('password', default: 'somedefault', ...)`.
[2018-09-26T13:54:17+02:00] WARN: Returning DEFAULT_ATTRIBUTE for 'password'. Use --attrs to provide a value for 'password' or specify a default  value with `attribute('password', default: 'somedefault', ...)`.

Profile: InSpec Profile (profile-attribute)
Version: 0.1.0
Target:  local://

  alice
     ×  should eq "bob"
     
     expected: "bob"
          got: "alice"
     
     (compared using ==)

  #<Inspec::Attribute::DEFAULT_ATTRIBUTE:0x00007f83bdff36b8>
     ↺  #<Inspec::Attribute::DEFAULT_ATTRIBUTE:0x00007f83bdff36b8>

Test Summary: 0 successful, 1 failure, 1 skipped
```

The output raises questions:
- as the user I do not know what a DEFAULT_ATTRIBUTE is
- should it be possible to execute the profile if no default value for the attribute is defined? 

### 🌍 InSpec and Platform Version

2.2.119 (3b6bd808)

### 🤔 Replication Case

run `inspec exec examples/profile-attribute` without attributes

### 💁 Possible Solutions

- implement the `to_s` method for better human-readable name in https://github.com/inspec/inspec/blob/master/lib/inspec/objects/attribute.rb#L17
- block execution if the no default value is provided

If we are going to block the execution, it need to be tied to a major version

chris-rock:
May be related to #3442

chris-rock:
We have a required field for attributes https://github.com/inspec/inspec/blob/master/lib/inspec/objects/attribute.rb#L50, discussed with @jquick briefly and we think a hard stop should only be done if required is set explicitly.

[...]

PR #3508, opened 2018-10-12
Update the text on the generic default attribute

jquick:

Fixes https://github.com/inspec/inspec/issues/3443

New output:
```
(jq/update_fallback_attribute_text)> _inspec exec default_attribute.rb
[2018-10-12T16:40:35-04:00] WARN: Attribute 'password' does not have a value. Use --attrs to provide a value for 'password' or specify a default  value with `attribute('password', default: 'somedefault', ...)`.

Profile: tests from default_attribute.rb (tests from default_attribute.rb)
Version: (not specified)
Target:  local://

  ↺  Test using an attribute without a value: Attribute 'password' does not have a value. Skipping test.
     ↺  Attribute 'password' does not have a value. Skipping test.


Profile Summary: 0 successful controls, 0 control failures, 1 control skipped
```

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -17,11 +17,13 @@ module Inspec
     DEFAULT_ATTRIBUTE = Class.new do
       def initialize(name)
         @name = name
+
+        # output warn message if we are in a exec call
         Inspec::Log.warn(
           "Attribute '#{@name}' does not have a value. "\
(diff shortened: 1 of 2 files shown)


jerryaldrichiii (approved):

clintoncwolfe (approved):
Well done, the to_s approach is a good idea.  I would like to move the `Inspec::BaseCLI.inspec_cli_command` one day, but that's for another day.

PR #3509, opened 2018-10-13
Allow http resource to follow redirects

cattywampus:
By specifying a `max_redirects` attribute, the `http` resource worker will follow any HTTP Redirect response (301, 302, etc...) up to the limit defined by this attribute. For a local worker, exceeding that limit will raise a `FaradayMiddleware::RedirectLimitReached` exception. For a remote worker, the curl command will exit without populating the `status` and `body` properties.

This is a first cut a solution for enabling this feature. Another option I considered was introducing two attributes: `follow_redirects` as a boolean defaulted to `false` and `max_redirects` as an integer defaulted to `3` which is consistent with `faraday` (curl defaults to 50 which seems excessive). In this alternate implementation, users would only have to provide the `follow_redirects` attribute to the resource to enable the behavior without having to think or worry about setting a redirect limit for most use cases. 

However, when you do need to control the limit there are more knobs to turn which could appear as clutter in the resource of the user's profile. Which is why this solution opts for the single `max_redirects` to enable and limit the behavior. I'm not married to either option so I'm curious to hear other opinions in terms of user experience or preference.

Also - I'm having a weird issue where running the unit test for the http resource does not produce consistent results on my local dev workstation. The majority of the time the tests pass but if I try to run the `rake test` task multiple times in a row I'll see this error:

```
# Running:

.............E.......

Finished in 1.165437s, 18.0190 runs/s, 48.0506 assertions/s.

  1) Error:
Inspec::Resources::Http::InSpec::Resources::Http::Worker::Local::request with redirect enabled#test_0002_does not exceed max_redirects:
NameError: uninitialized constant FaradayMiddleware::RedirectLimitReached
Did you mean?  FaradayMiddleware::RedirectLimitReached
    /Users/kwalters/code/cattywampus/inspec/test/unit/resources/http_test.rb:50:in `block (4 levels) in <top (required)>'

21 runs, 56 assertions, 0 failures, 1 errors, 0 skips
```

I've tried to look into why there's this 10-20% chance that the Faraday exception class isn't loaded when the tests run, but I couldn't figure it out...

### DIFF
--- a/docs/resources/http.md.erb
+++ b/docs/resources/http.md.erb
@@ -23,7 +23,7 @@ This resource first became available in v1.10.0 of InSpec.
 
 An `http` resource block declares the configuration settings to be tested:
 
-    describe http('url', auth: {user: 'user', pass: 'test'}, params: {params}, method: 'method', headers: {headers}, data: data, open_timeout: 60, read_timeout: 60, ssl_verify: true) do
+    describe http('url', auth: {user: 'user', pass: 'test'}, params: {params}, method: 'method', headers: {headers}, data: data, open_timeout: 60, read_timeout: 60, ssl_verify: true, max_redirects: 3) do
       its('status') { should eq number }
       its('body') { should eq 'body' }
       its('headers.name') { should eq 'header' }
@@ -40,6 +40,7 @@ where
 * `open_timeout` may be specified for a timeout for opening connections (default to 60)
 * `read_timeout` may be specified for a timeout for reading connections (default to 60)
 * `ssl_verify` may be specified to enable or disable verification of SSL certificates (default to `true`)
+* `max_redirects` may be specified to control how many HTTP Redirects to follow (defaults to `0`)
 
 <br>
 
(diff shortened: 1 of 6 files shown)


docs/resources/http.md.erb
```diff
@@ -40,6 +40,7 @@ where
 * `open_timeout` may be specified for a timeout for opening connections (default to 60)
 * `read_timeout` may be specified for a timeout for reading connections (default to 60)
 * `ssl_verify` may be specified to enable or disable verification of SSL certificates (default to `true`)
+* `max_redirects` may be specified to control how many HTTP Redirects to follow (defaults to `0`)
```
clintoncwolfe, line 43: I like the single-flag approach, it's concise and powerful, yet obvious how to disable redirects.

clintoncwolfe (approved):
This looks great to me - sorry it sat in our PR review box for so long!

miah (approved):
Thank you @cattywampus! I agree with @clintoncwolfe, single flag is great. I also find it strange that curl does _50_ retries wow.

clintoncwolfe (changes requested):
Ach, hang on, looks like we need to update the gemspec - from travis:
```
[2018-10-13T06:47:35+00:00] ERROR: Report handler Chef::Handler::AuditReport raised #<LoadError: cannot load such file -- faraday_middleware>
```

cattywampus:
I'll fix that and push something up shortly.

cattywampus:
@clintoncwolfe I added `faraday-middleware` to the `inspec-core` gemspec. That seemed to fix it.

clintoncwolfe (approved):
Thanks!

PR #3341, opened 2018-08-29
Add HTTP basic auth for URL based inspec deps

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

PR #3338, opened 2018-08-29
Support erb rendering

frezbo:
Fixes: https://github.com/inspec/inspec/issues/3337

Signed-off-by: Noel Georgi <git@frezbo.com>

PR #3510, opened 2018-10-13
fixes bug with running profiles that comes from authenticated URL

frezbo:
This is fixed by generating the lock
file with the username/password data and makign sure we render it with
ERB before running the profile.

### DIFF
--- a/lib/inspec/dependencies/requirement.rb
+++ b/lib/inspec/dependencies/requirement.rb
@@ -80,15 +80,30 @@ module Inspec
       @resolved_source ||= fetcher.resolved_source
     end
 
+    def metadata
+      @metadata ||= Inspec::Metadata.from_unrendered_yaml('mock', File.read(File.join(opts[:cwd], 'inspec.yml')), nil)
+    end
+
     def to_hash
       h = {
         'name' => name,
         'resolved_source' => resolved_source,
         'version_constraints' => version_constraints,
       }
-
       if !dependencies.empty?
         h['dependencies'] = dependencies.map(&:to_hash)
+      end
+      if resolved_source[:url] && opts[:username] && opts[:password]
+        auth_data = metadata.dependencies.detect { |deps| deps[:name] == name }
+        if auth_data
+          resolved_source[:username] = auth_data[:username]
+          resolved_source[:password] = auth_data[:password]
+
+          h['dependencies']&.each do |dep|
+            dep['resolved_source'][:username] = auth_data[:username]
+            dep['resolved_source'][:password] = auth_data[:password]
+          end
+        end
       end
 
       h
--- a/lib/inspec/metadata.rb
+++ b/lib/inspec/metadata.rb
@@ -210,6 +210,13 @@ module Inspec
       finalize(res, profile_id, {}, logger)
     end
 
+    def self.from_unrendered_yaml(ref, content, profile_id, logger = nil)
+      res = Metadata.new(ref, logger)
+      res.params = YAML.load(content)
+      res.content = content
+      finalize(res, profile_id, {}, logger)
+    end
+
     def self.from_ruby(ref, content, profile_id, logger = nil)
       res = Metadata.new(ref, logger)
       res.instance_eval(content, ref, 1)
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -456,7 +456,7 @@ module Inspec
 
     def lockfile
       @lockfile ||= if lockfile_exists?
-                      Inspec::Lockfile.from_content(@source_reader.target.read('inspec.lock'))
+                      Inspec::Lockfile.from_content(ERB.new(@source_reader.target.read('inspec.lock')).result)
                     else
                       generate_lockfile
                     end


frezbo:
We have nested profiles that depends on other profiles. This generates a lock file similar to:

```
---
lockfile_version: 1
depends:
- name: tomcat
  resolved_source:
    url: https://artifactory.prod.com/artifactory/inspec-release/tomcat/0.1.1/tomcat-0.1.1.tar.gz
    sha256: 63b4144d9b32a23545a9dbe5e489f586dcc1177a9526213492ef2b9bbe156011
    username: "<%= ENV['ARTIFACTORY_MASTER_USERNAME'] %>"
    password: "<%= ENV['ARTIFACTORY_MASTER_API_KEY'] %>"
  version_constraints: []
  dependencies:
  - name: linux-security
    resolved_source:
      url: https://artifactory.prod.com/artifactory/inspec-release/linux-security/0.1.1/linux-security-0.1.1.tar.gz
      sha256: f576e26c3e4cc471cfc35850077c9356e9e5b7dbe58aad9ee47819b398c271ec
      username: "<%= ENV['ARTIFACTORY_MASTER_USERNAME'] %>"
      password: "<%= ENV['ARTIFACTORY_MASTER_API_KEY'] %>"
    version_constraints: []
- name: tomcat-compliance
  resolved_source:
    url: https://artifactory.prod.com/artifactory/inspec-release/tomcat-compliance/0.1.1/tomcat-compliance-0.1.1.tar.gz
    sha256: 749d00341ba08d3f84b6e9ccc383696c827c27615aae6b9ec47a56c33a236217
    username: "<%= ENV['AARTIFACTORY_USERNAME'] %>"
    password: "<%= ENV['AARTIFACTORY_API_KEY'] %>"
  version_constraints: []
  dependencies:
  - name: linux-security
    resolved_source:
      url: https://artifactory.prod.com/artifactory/inspec-release/linux-security/0.1.1/linux-security-0.1.1.tar.gz
      sha256: f576e26c3e4cc471cfc35850077c9356e9e5b7dbe58aad9ee47819b398c271ec
      username: "<%= ENV['AARTIFACTORY_USERNAME'] %>"
      password: "<%= ENV['AARTIFACTORY_API_KEY'] %>"
    version_constraints: []
- name: base-os
  resolved_source:
    url: https://artifactory.prod.com/artifactory/inspec-release/base-os/0.4.4/base-os-0.4.4.tar.gz
    sha256: 042cbc025272b9ea15a3a8746c5ef8a03f9af574c841f40a95d0d9afee06ddad
    username: "<%= ENV['ARTIFACTORY_SECONDARY_USERNAME'] %>"
    password: "<%= ENV['ARTIFACTORY_SECONDARY_API_KEY'] %>"
  version_constraints: []
  dependencies:
  - name: linux-security
    resolved_source:
      url: https://artifactory.prod.com/artifactory/inspec-release/linux-security/0.1.1/linux-security-0.1.1.tar.gz
      sha256: f576e26c3e4cc471cfc35850077c9356e9e5b7dbe58aad9ee47819b398c271ec
      username: "<%= ENV['ARTIFACTORY_SECONDARY_USERNAME'] %>"
      password: "<%= ENV['ARTIFACTORY_SECONDARY_API_KEY'] %>"
    version_constraints: []
```

PR that's added initial support for this:
- https://github.com/inspec/inspec/pull/3338
- https://github.com/inspec/inspec/pull/3341

frezbo:
@jerryaldrichiii can this be looked at?

jerryaldrichiii:
It sure can! I'll ping @jquick.

aaronlippold:
:-) hehe

jquick:
Thanks @frezbo. I am a bit confused by this. If we have a lock file it means we have cached the dependency either in the cache or in the vendor. If for some reason this is not the case we would re-download using the settings in the depends section of the `inspec.yml`. Can you specify the stacktrace of the bug? I want to fully understand the issue at hand.

frezbo:
yes I can explain, lets say I was running a profile that has deep dependency on url's with authentication from a clean VM (it does not now have the local `vendor` folder, not the `~/.inspec/cache`). When it tries to resolve the dependencies from external URL it throws `404` (Artifactory throws `404` for un-authenticated requests to the full path) or any other webserver might throw `403` as the authentication info is not present in the `inspec.lock` (`inspec exec` only tries to resolve deps data from `inspec.lock`). So this PR was an effort to fix that.

Moreover having auth info in the dep file (here `inspec.yml` ) is the standard behavior as seen with other tools like **berks** and **bundler**. There we could have the lockfile without any auth info and the local `Gemfile`/`Berksfile` have the auth info. I tried to implment the same with inspec also.

Hope this clarifies the need for this PR. @jquick

jquick:
Hey @frezbo. The team and I sat down and talked this over. We are uncomfortable with this approach currently. The reasons being:
1) We don't want any dynamic data being written to a lock file. They should stay static.
2) We also have a concern with adding potentially sensitive information to another file.
3) We have this data already in the `inspec.yml`.

As a work around for now you can run your profiles with `--no-create-lockfile` which will not lock/vendor the dependencies. This way you can always take it with you. Another option is to `inspec vendor` your profile which will embed the dependencies so they are always accessible. 

If we do end up implementing this it will likely be some sort of merging of options from `inspec.yml` at run time if needed. Thanks again for your proof of concept here and addition!

frezbo:
@jquick Thanks for taking a look at this. I don't want the profiles to be run without locked deps. :D . I have fixed it in our code by calling the `vendor` API as you mentioned. The explanation makes sense.

PR #3520, opened 2018-10-16
Allow up to 17 characters for vpc id's

kchistova:


test/unit/resources/aws_vpc_test.rb
```diff
@@ -17,7 +17,7 @@ def test_empty_params_ok
   end
 
   def test_accepts_vpc_id_as_scalar
-    AwsVpc.new('vpc-12345678')
+    AwsVpc.new('vpc-12345678987654321')
```
miah, line 20: Thanks for the contribution @kchistova. Rather than replacing this test, could we add another? Have one test for 8 character and the other for 17?

miah (changes requested):

aaronlippold:
Should they be separate describe blocks so if one fails we know which one

Issue #3297, opened 2018-08-17
RFC: Improve airgap (conservative network usage) support


clintoncwolfe:
### 🎛 Description

🙋 feature request

As InSpec is often used as a security tool, some users use it in an "airgapped" deployment (isolated from the Internet or non-organizational networks).  Some InSpec features break in an ungraceful way under these circumstances, as the feature timesout while trying to reach a public URL.

Note that InSpec should still try "normal" network attempts to the test target, as well as any hosts mentioned in tests (for example, the hostname of a `postgres` resource).  Failure to connect to those destinations indicates a test failure, not something InSpec should ignore.

Some possible changes:

* Add an --airgap CLI option to disable all non-test-related network communication
* As with all CLI options, allow setting --airgap in `~./inspec/config.json`
* Disable 'do I have the last version' check
* Disable all remote profile fetchers; gracefully error out, mentioning the `vendor` command if we need to fetch a profile
* At the cost of more complexity, we could have a list of permitted fetcher hosts in the config file; this would permit fetching from a privately hosted git server, for example. 
* Modify the upcoming `inspec plugin install` command to insist on local gemfiles, or use a permitted list of gem source servers

Like all RFCs I submit, I'm looking to gauge interest, discover / refine use cases, and otherwise get feedback; nothing I mentioned above is a commitment.

### 🌍 InSpec and Platform Version

2.2.50

clintoncwolfe:
Refs #4337

PR #3522, opened 2018-10-17
Add support for relative paths to the Git fetcher

### LINKED ISSUE #1237: Allow multiple profiles in a git repository
At the moment, InSpec's dependency management allows only one profile per git repository. This allows tags to be used to version the profile in the repo.

This feature request wants the ability to store multiple profiles in a single repository, under a sub-directory or at root.

Reasons that come to my mind:
- per repo pricing
- grouping of rarely updated profiles

---

Berkshelf example:

``` ruby
cookbook "mykb", git: "https://github.com/example/super_cookbooks.git", rel: "base/linux"
```

### LINKED ISSUE #1974: Add sub-directory property to Git fetcher
### Description

In my experience, various users have issues with large numbers of closely related repositories being combined into a reference platform.

See https://github.com/chris-rock/acme-inspec-profile/blob/master/inspec.yml#L10-L21 for a small example.

It would be nice to be able to specify named subdirectories for those projects that wish to use a single repository for InSpec content.

Example:

```yaml
depends:
  - name: my_inspec
    url: https://github.com/me/my_inspec_bundle
    directory: special_sauce
```

This would point `my_inspec` at the `special_sauce` subdirectory of `my_inspec_bundle` and improve the user experience.

### LINKED ISSUE #2167: Support profiles not at root level of a git repository
Please add the ability to specify a relative path when using the `git` fetcher such that InSpec can support using a profile in a repo where the repo is not at the root directory, such as a Chef cookbook used by Test Kitchen, or a repository that contains multiple profiles.

(The above is an updated description of this feature request.  Below is the original description.)

----

This was discussed briefly in https://github.com/chef/kitchen-inspec/issues/66 and determined to be a new issue.

When including a profile from an `inspec.yml` file, InSpec expects the included profile to live in top-level directories in the referenced git repository/tarball/path.

When writing a profile to be used with TestKitchen, specifically with kitchen-inspec, the profile must live in a subdirectory called `test/integration/$suite/inspec`.

Because of this discrepancy, it's impossible to write a profile that can be run with TestKitchen and also exported for use by other profiles.

jerryaldrichiii:
This adds support for specifying a path to a profile in a Git repo.

For example, you could have a Git repo containing multiple profiles and you want to specify a single one like below:

```yaml
depends:
  - name: profile-1
    git: https://github.com/myorg/all-my-profiles.git
    profile_path: profiles/my-specific-profile-1
  - name: profile-2
    git: https://github.com/myorg/all-my-profiles.git
    profile_path: profiles/my-specific-profile-2
```

This closes #2167 
This closes #1974 
This closes #1237

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -212,9 +212,7 @@ depends:
 
 ### git
 
-A `git` setting specifies a profile that is located in a git repository, with optional settings for branch, tag, commit, and version. The source location is translated into a URL upon resolution. This type of dependency supports version constraints via semantic versioning as git tags.
-
-For example:
+A `git` setting specifies a profile that is located in a git repository, with optional settings for branch, tag, commit, version, and profile_path. The source location is translated into a URL upon resolution. This type of dependency supports version constraints via semantic versioning as git tags.
 
 ```YAML
 depends:
@@ -224,6 +222,7 @@ depends:
   tag:     desired_version
   commit:  pinned_commit
   version: semver_via_tags
+  profile_path: optional/path/to/profile
 ```
 
 ### supermarket
--- a/lib/fetchers/git.rb
+++ b/lib/fetchers/git.rb
@@ -38,10 +38,12 @@ module Fetchers
 
     def initialize(remote_url, opts = {})
       @branch = opts[:branch]
+      @profile_directory = nil
       @tag = opts[:tag]
       @ref = opts[:ref]
       @remote_url = remote_url
       @repo_directory = nil
+      @target_profile_path = opts[:profile_path]
     end
 
     def fetch(dir)
@@ -53,23 +55,35 @@ module Fetchers
       else
         Dir.mktmpdir do |tmpdir|
           checkout(tmpdir)
-          Inspec::Log.debug("Checkout of #{resolved_ref} successful. Moving checkout to #{dir}")
-          FileUtils.cp_r(tmpdir + '/.', @repo_directory)
+          if @target_profile_path
+            @profile_directory = dir
+            Inspec::Log.debug("Checkout of #{resolved_ref} successful. " \
+                              "Moving #{@target_profile_path} to #{dir}")
+            target_profile = File.join(tmpdir, @target_profile_path)
+            FileUtils.cp_r(target_profile, dir)
+          else
+            Inspec::Log.debug("Checkout of #{resolved_ref} successful. " \
+                              "Moving checkout to #{dir}")
+            FileUtils.cp_r(tmpdir + '/.', dir)
+          end
         end
       end
-      @repo_directory
+      @profile_directory || @repo_directory
     end
 
     def cache_key
-      resolved_ref
+      return resolved_ref unless @target_profile_path
+      OpenSSL::Digest::SHA256.hexdigest(resolved_ref + @target_profile_path)
     end
 
     def archive_path
-      @repo_directory
+      @profile_directory || @repo_directory
     end
 
     def resolved_source
-      { git: @remote_url, ref: resolved_ref }
+      source = { git: @remote_url, ref: resolved_ref }
+      source[:profile_path] = @target_profile_path if @target_profile_path
+      source
     end
 
     private
--- a/test/functional/gitfetcher_test.rb
+++ b/test/functional/gitfetcher_test.rb
@@ -7,42 +7,106 @@ describe 'profiles with git-based dependencies' do
   include FunctionalHelper
   before(:all) do
     @tmpdir = Dir.mktmpdir
-    @profile_dir = File.join(@tmpdir, "test-profile")
-    @git_dep_dir = File.join(@tmpdir, "git-dep")
+    @profile_dir = File.join(@tmpdir, 'test-profile')
+    @git_dep_dir = File.join(@tmpdir, 'git-dep')
+    @deep_git_dep_dir = File.join(@tmpdir, 'deep-git-dep')
 
     Dir.chdir(@tmpdir) do
-      inspec("init profile git-dep")
-      inspec("init profile test-profile")
+      inspec('init profile git-dep')
+      inspec('init profile test-profile')
+
+      # InSpec will search directories and execute successfully if there is just
+      # one profile. We create two here to ensure we execute both as intended.
+      deep_profiles = File.join(@deep_git_dep_dir, 'profiles')
+      FileUtils.mkdir_p(deep_profiles)
+      inspec('init profile ' + File.join(deep_profiles, 'deep-profile'))
+      inspec('init profile ' + File.join(deep_profiles, 'another-deep-profile'))
     end
 
-    Dir.chdir(@git_dep_dir) do
-      CMD.run_command("git init")
-      CMD.run_command("git add .")
-      CMD.run_command("git config user.name \"test\"")
-      CMD.run_command("git config user.email \"test@yahoo.com\"")
-      CMD.run_command("git commit -m \"initial commit\" --no-gpg-sign")
-      CMD.run_command("git commit -m \"another commit\" --allow-empty --no-gpg-sign")
-      CMD.run_command("git tag antag")
-    end
-
-    inspec_yml = YAML.load(File.read(File.join(@profile_dir, "inspec.yml")))
+    inspec_yml = YAML.load(File.read(File.join(@profile_dir, 'inspec.yml')))
     inspec_yml["depends"] = [
       {
         'name' => 'git-dep',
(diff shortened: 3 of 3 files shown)


jerryaldrichiii:
@trevor-vaughan (and InSpec team) let me know what you think. I'm not set on `profile_path` as the UX, I've seen `directory`, `rel`, and `path` I believe.

trevor-vaughan:
@jerryaldrichiii Honestly, out of those, `profile_path` is the most descriptive to a new user in my opinion.

`directory` is my second choice, `rel` doesn't make any sense to me, and `path` feels like it could be anywhere on the system.

Alternatively, something like `subdirectory` would also be more descriptive.

trevor-vaughan (commented):
:+1:

chris-rock:
I like that functionality 👍 

There are still a few things to think about and we should have clear answers to those. e.g As https://github.com/inspec/inspec/issues/1237 highlights, many profiles can reside in a git repo, therefore we need to think about how this profile can be executed directly too. If `inspec exec git` is pointed to a mono-repo, how should it behave? Should it say, no profile found, should it run all profiles? Do we need a cli flag to provide the relative profile path? What happens if I have a mono-repo in git and I point `inspec` to its tar representation? What happens if the profiles have relative dependencies within the same git repo? Probably you want to use a relative path in the inspec yml, to ensure it works with local paths. I think we need a consistent story.

trevor-vaughan:
As a point to note, this is the last thing that I need to be able port my profile into https://github.com/complianceascode/content

Responses to points by @chris-rock below:

> There are still a few things to think about and we should have clear answers to those. e.g As #1237 highlights, many profiles can reside in a git repo, therefore we need to think about how this profile can be executed directly too.

AFAIK, you can already do that by just pointing to the directory.

> If inspec exec git is pointed to a mono-repo, how should it behave? Should it say, no profile found, should it run all profiles? 

Provide a list of discovered profiles and allow the user to select one (or more). Have a command line switch that allows selection of profiles from a mono-repo by name.

> Do we need a cli flag to provide the relative profile path?

Probably, yes.

> What happens if I have a mono-repo in git and I point inspec to its tar representation?

Same as above, it prompts you for the ones that you want to run unless you specify them at the command line. For an easier to implement version, it could just print what you have available and then tell you what command line flag to add so that it's less interaction and debugging.

> What happens if the profiles have relative dependencies within the same git repo?

Wouldn't these be resolved at run time and, given that they're all together, work just fine?

> Probably you want to use a relative path in the inspec yml, to ensure it works with local paths.

Yes, this makes the most sense. Doing anything else binds you to the internet anyway and is almost always incorrect for secure environments.

clintoncwolfe:
FWIW, #3618 provides rich prompting capabilities
Also refs #3297 for airgap support

aaronlippold:
Are there still open changes that need to take place on this PR or are we just waiting on a review?

clintoncwolfe:
Removing DNM flag: ready for review for potential merge

bdwyertech (approved):
Looks good to me.  I'll start looking at this from the `kitchen-inspec` perspective

test/functional/gitfetcher_test.rb
```diff
@@ -7,42 +7,106 @@
   include FunctionalHelper
   before(:all) do
     @tmpdir = Dir.mktmpdir
-    @profile_dir = File.join(@tmpdir, "test-profile")
-    @git_dep_dir = File.join(@tmpdir, "git-dep")
+    @profile_dir = File.join(@tmpdir, 'test-profile')
```
clintoncwolfe, line 10: Might suggest naming this 'top-level-profile'.  All of these are test profiles.

test/functional/gitfetcher_test.rb
```diff
@@ -7,42 +7,106 @@
...
     FileUtils.rm_rf(@tmpdir)
   end
 
-  it 'executes a profile with a git based dependency' do
-    out = inspec("exec #{@profile_dir} --no-create-lockfile")
-    out.stderr.must_equal ''
-    out.exit_status.must_equal 0
+  it 'executes a profile with a git based dependencies' do
+    Dir.mktmpdir do |tmpdir|
+      out = inspec("exec #{@profile_dir} " \
```
clintoncwolfe, line 105: Try to use `run_inspec_process` when writing new functional tests. It has options like CWD, JSON output, no lockfile by default, etc.

test/functional/gitfetcher_test.rb
```diff
@@ -7,42 +7,106 @@
...
-      CMD.run_command("git config user.name \"test\"")
-      CMD.run_command("git config user.email \"test@yahoo.com\"")
-      CMD.run_command("git commit -m \"initial commit\" --no-gpg-sign")
-      CMD.run_command("git commit -m \"another commit\" --allow-empty --no-gpg-sign")
-      CMD.run_command("git tag antag")
+      # InSpec will search directories and execute successfully if there is just
+      # one profile. We create two here to ensure we execute both as intended.
+      deep_profiles = File.join(@deep_git_dep_dir, 'profiles')
+      FileUtils.mkdir_p(deep_profiles)
+      inspec('init profile ' + File.join(deep_profiles, 'deep-profile'))
```
clintoncwolfe, line 22: Try to use `run_inspec_process` when writing new functional tests. It has options like CWD, etc.

test/functional/gitfetcher_test.rb
```diff
@@ -7,42 +7,106 @@
...
 
   after(:all) do
     FileUtils.rm_rf(@tmpdir)
   end
 
-  it 'executes a profile with a git based dependency' do
-    out = inspec("exec #{@profile_dir} --no-create-lockfile")
-    out.stderr.must_equal ''
-    out.exit_status.must_equal 0
+  it 'executes a profile with a git based dependencies' do
```
clintoncwolfe, line 103: I think this is under-tested.  As @chris-rock mentioned on the PR commentary, there are several distinct use cases to examine here, and only one is attempted (though it is a complex one).

I'd like to see more use cases tested.  For example, this doesn't test the case in which there are multiple profiles in a git repo, and only one is wanted.

docs/profiles.md
```diff
@@ -224,6 +222,7 @@ depends:
   tag:     desired_version
   commit:  pinned_commit
   version: semver_via_tags
+  profile_path: optional/path/to/profile
```
clintoncwolfe, line 225: Maybe `optional/relative/path/within/repo`?

clintoncwolfe, line 225: Hrm, we actually have a bit of a naming conflict.  `path` is a valid key here, and specifies to use the local file fetcher. `profile_path` is only valid if `git` is also present, and indicates the path within the repo.

`path_within_repo`?

lib/fetchers/git.rb
```diff
@@ -38,10 +38,12 @@ def self.resolve(target, opts = {})
 
     def initialize(remote_url, opts = {})
       @branch = opts[:branch]
+      @profile_directory = nil
       @tag = opts[:tag]
       @ref = opts[:ref]
       @remote_url = remote_url
       @repo_directory = nil
+      @target_profile_path = opts[:profile_path]
```
clintoncwolfe, line 46: It might be better to avoid using the word 'target' in new code, unless you are specifically referring to the scanning target.

clintoncwolfe (changes requested):
Thanks @jerryaldrichiii! I think in a situation with many different use cases, we need to be careful to test the majority of them.  So, that's the main request - test more and different situations.

lib/fetchers/git.rb
stevendanna, line 46: The rest of this PR has a good deal of conditional logic here.  I'm wondering if that conditional logic could be removed by making the _default_ profile_path `./`.  This of course depends on what behavior we want when someone happens to point at a git repository with multiple profiles.

jerryaldrichiii:
Hey all, given my job change I won't have time to work on this PR.

Feel free to adopt and make any changes needed :smile:

docs/profiles.md
clintoncwolfe: Today 'relative_path' feels good.

miah (changes requested):
Needs adoption.

kekaichinose:

__Why do I care?__

__What do I need to do?__

kekaichinose:
Reopening because I wasn't fully up to speed on ask here, and since the PR addresses the OP's problem. Let's review this and get this out the door.

clintoncwolfe:
This PR has been adopted on #4217 - closing this one. Thanks @jerryaldrichiii!

Issue #3521, opened 2018-10-16
v3 - Plugin - Unknown error occured  - installation failed. 


burdena:
### 🎛 Description

**Choose one:** is this a 🐛 bug report 

Briefly describe the issue and the expected behavior. Also, please be aware of our [Code of Conduct](https://github.com/inspec/inspec/blob/master/CODE_OF_CONDUCT.md).

### 🌍 InSpec and Platform Version

Tell us which version of InSpec (`inspec version` or SHA of a commit) and Operating System distribution you are using

### 🤔 Replication Case

Unable to install any plugins.

$inspec plugin install inspec-iggy
Unknown error occured  - installation failed

$inspec plugin install inspec-k8s
Unknown error occured  - installation failed.

 $inspec version
3.0.0

$uname -a
Darwin Adams-MBP 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64

burdena:
MAC OSX 10.13.6 (17G65)

burdena:
also tried via the docker image 

docker run -it --rm chef/inspec plugin install inspec-iggy
Unknown error occured  - installation failed.

docker run -it --rm chef/inspec version
3.0.0

kmf:
Same here.

jquick:
Sorry about this. We still have some unsupported plugins showing up as installable. `inspec-iggy` is being updated for plugins system. Once this is released https://github.com/inspec/inspec-iggy/pull/17 you should be able to install this plugin. `inspec-k8s` and `inspec-release` are actually just placeholders and are not real. They will be invalidated on tomorrows release to remove any confusion.

If you do want to test installing a plugin please use:
`inspec-resource-lister`

burdena:
Cheers for the update, any idea when the new iggy release will happen ?

jquick:
inspec iggy `0.4.0` was released last night and should work correctly for you:
```
$ inspec plugin install inspec-iggy
Fetching: inspec-iggy-0.4.0.gem (100%)
inspec-iggy plugin, version 0.4.0, installed from rubygems.org
```

burdena:
I had to explicitly ask for 0.4.0 as only 0.2.0 was showing as availalbe.

$inspec plugin install inspec-iggy
Unknown error occured  - installation failed.

$ inspec plugin search inspec-

 Plugin Name                   Versions Available
[...]

PR #3525, opened 2018-10-17
Filter out inspec-k8s and inspec-release

miah:
ref: https://github.com/inspec/inspec/issues/3521


jquick (approved):
Tested this out. Looks good. Thanks @miah

clintoncwolfe (approved):

PR #3533, opened 2018-10-21
Add ksh (Korn Shell) resource

aklyachkin:
Trivial change - added support for ksh (Korn Shell) resource.


test/unit/resources/ksh_test.rb
```diff
@@ -0,0 +1,29 @@
+# encoding: utf-8
+# author: Christoph Hartmann
```
jerryaldrichiii: Could you remove these author lines? We track authorship via Git so no need to replace them.

jerryaldrichiii (changes requested):
Thank you for your first contribution @aklyachkin! Only a few suggestions.

In addtion to removing the author lines, could you also do DCO sign off by doing a `git commit --amend -s`? This is a little beyond a Chef's "Obvious Fix" policy I believe, see: https://docs.chef.io/community_contributions.html#the-obvious-fix-rule

For more info on DCO you can check out: https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

aklyachkin:
Hi, I changed both things. Although I still think it's quite trivial - i just took bash resource and changed two letters in it ;-)

jerryaldrichiii (approved):

jerryaldrichiii:
I can see that argument :smile: couldn't hurt to have the sign off though. Better safe than sorry.

jquick (approved):
Looks good to me! Thanks @aklyachkin

PR #3539, opened 2018-10-23
Fix rendering and other issues in the Style Guide

jerryaldrichiii:


docs/style.md
```diff
@@ -2,45 +2,53 @@
 
 This is a set of recommended InSpec rules you should use when writing controls.
 
-## Control files
+# Control Files
 
-### 1. All controls should be located in the "controls" directory and end in ".rb"
+## Place control files in "controls/" and end them with ".rb"
```
jquick: I think leaving the `in the "controls" directory` is clearer here then the forward slash.

docs/style.md
```diff
@@ -2,45 +2,53 @@
...
-Use: `controls/ssh_config.rb`
+Avoid: `controls/ssh_config`</br>
+Use: `controls/ssh_config.rb`</br>
 
-Avoid: `controls/ssh/config.rb`
+Avoid: `controls/ssh/config.rb`</br>
 Use: `controls/ssh_config.rb`
 
-### 2. Avoid using "controls" or "control" in the name of your control files
+## Avoid "controls"/"control" in your control filenames
```
jquick: same here

docs/style.md
```diff
@@ -2,45 +2,53 @@
...
 
+# Controls
 
-## Controls
+## Avoid wrapping controls in conditional statements
 
-### 4. Do not wrap controls in conditional statements
-
-Reason: This will create dynamic profiles whose controls depend on the execution. The problem here is that we cannot render the profile or provide its information before scanning a system. We want to be able to inform users of the contents of their profiles before they run them. It is valid to skip controls that are not necessary for a system, as long as you do it via `only_if` conditions. Ruby's internal conditionals will hide parts of the profile to static analysis and should thus be avoided.
+This will create dynamic profiles whose controls depend on the execution. The
```
jquick, line 50: Why are we setting line length here (and other places in the file)? I think we should allow the renderer to wrap as it pleases. If it looks bad on render we should update those settings.

jerryaldrichiii: It doesn't look bad on render either way but the file had both forced line wraps and no line wraps.

I went with forced line wraps for easier editing and making sure it was standard in the file...but I'm not set on that idea.

I prefer it because it makes reviewing in GitHub easier.

jerryaldrichiii: `s/doesn't look bad either way/doesn't affect the render either way/`

It will render the same regardless of the line wraps in the file.

docs/style.md
jerryaldrichiii: Adding the extra words causes it to wrap in an ugly way when rendered. :disappointed:

docs/style.md
jerryaldrichiii: Same issue here, making this `"controls" or "controls" cause it to wrap in a bad way when rendered.

docs/style.md
clintoncwolfe: Consider using backticks for technical literals: `controls` rather than "controls" .

docs/style.md
```diff
@@ -136,42 +154,46 @@ end
...
 resolved, installed, vendored, and protected from conflicts. We aim to avoid
-exposing this complexity to users of InSpec, to make it a great tool even if
-you are not a developer.
-
-Developers may still use external gem dependencies but should vendor it
-with their plugins or resource packs.
+exposing this complexity to users of InSpec, to make it a great tool even if you
+are not a developer.
 
+Developers may still use external gem dependencies but should vendor it with
```
clintoncwolfe: I think this should be rephrased, as plugins and resource packs treat gems differently.

Perhaps:

Plugins should declare gem dependencies in their gemspec, and then rely on the plugin installation facility to install and manage dependencies.

I'm not sure how to ship gems with resource packs.  I don't think vendoring them will work (how would that load path be set?)

Feel free to address that on a separate PR.

docs/style.md
```diff
@@ -2,45 +2,53 @@
...
-### 1. All controls should be located in the "controls" directory and end in ".rb"
+## Place control files in "controls/" and end them with ".rb"
 
-Reason: Most syntax highlighters will render InSpec files correctly across a wide list of tools.
+Most syntax highlighters will render InSpec files correctly across a wide list
+of tools.
 
-Avoid: `controls/ssh_config`
-Use: `controls/ssh_config.rb`
+Avoid: `controls/ssh_config`</br>
```
clintoncwolfe: Throughout the file, this PR adds HTML breaks in lists.  I think it would look better, and be more semantic, if the lists used a list marker:

* Avoid: `controls/ssh_config`
* Use: `controls/ssh_config.rb`

We could add levels, as well:

* Avoid: `controls/ssh_config`
  * Use: `controls/ssh_config.rb`

clintoncwolfe (commented):
Just some thoughts about HTML injection, and the gem deps issue.

docs/style.md
miah: Maybe make "controls/" a link to the directory in the repository so its a little more obvious?

docs/style.md
jerryaldrichiii: I agree it should be fixed. I'll add your suggestion to this PR. I'm also not sure about vendoring gems in resource packs.

docs/style.md
jerryaldrichiii: I'm not a fan of levels in this case...but I like it better than `</br>` and have no other solutions.

I'll change them 😄

jerryaldrichiii: Actually, how about the following since it works better with multiple avoids?:

Avoid: 
  - controls/ssh_config

Use: 
  - controls/ssh_config.rb

clintoncwolfe: I like that, too!

clintoncwolfe (approved):
Thanks, looks great!

jquick (approved):
Thanks @jerryaldrichiii

PR #3541, opened 2018-10-24
Update to safe navigation exit code search

jquick:

Currently there is an issue upstream in ChefClient where a runner is executed with no profiles added. In these cases we do not have any default profiles so we need to safe navigation the check.


jerryaldrichiii (approved):
Great catch!

PR #3543, opened 2018-10-26
Modernize omnibus config and reduce omnibus package size

tas50:
Sync over a lot of changes we've made to the omnibus config for chef. 1 change per commit.


aaronlippold:
Osx 10.14 as well?

tas50:
We don't have a test VM for 10.14

jquick (approved):
Thanks @tas50

miah (approved):
Thanks @tas50

Issue #3459, opened 2018-10-01
inspec exec command line ordering does not work as expected


trevor-vaughan:
### 🎛 Description

**Choose one:** 🐛 bug report

In most command line utilities, I am used to the following syntax:

`<COMMAND> <OPTIONS> <NON_OPTION_ITEMS>`

With InSpec, I just tried to switch from using `--format` to `--reporter` and found that what I expected to work, as shown above, did not.

I expected that `inspec exec --reporter json /path/to/stuff` would execute but instead it gave me an error that `/path/to/stuff` is not a valid reporter type.

I then tried `inspec exec --reporter json -- /path/to/stuff` since `--` is common for 'end of options' but this also did not work.

I checked the docs at https://www.inspec.io/docs/reference/cli/ but the `exec` section didn't have any indication of this strange behavior in the main documentation or option docs that I could find.

### 🌍 InSpec and Platform Version

InSpec 2.3.5 on CentOS 7 as installed via the official YUM repos.

### 💁 Possible Solutions

1. Make the CLI work like most traditional utilities (preferred)
2. Update the docs to note options that have to go after the path
3. Make the path an option and don't allow users to set it at the command line without a flag

aaronlippold:
The updated CLI structures are located here.
https://www.inspec.io/docs/reference/reporters/

However, you are correct, the Thor CLI does diverge a bit from the GNU
standards we are all used too. The updates would have to be made there but
I do know there are some issues with getting a fully matching style due to
some issues in Thor itself. ( I don't recall specifically the reasons ) but
that is where we would have to start.

I do know however that --reporter should be at the end given it currently
uses a space as its parsing token aka --reporter cli json:<path_to_my_file>
is what you are currently looking for.

Also the first reporter defaults to on-screen display so if you want to
display yaml or json to the screen the recommended way is:

inspec exec <PROFILE> --reporter json:-
inspec exec <PROFILE> --reporter yaml:-

aaronlippold:
`inspec exec <profile> <target> <attributes> (<json-config>)* <options> <reporter>` is what I have found to be the most stable ordering

* json-config will take the place of many of the other <option-blocks>

aaronlippold:
inspec exec <profile> <target> <attributes> (<json-config>)* <options>
<reporter> is what I have found to be the most stable ordering

   - json-config will take the place of many of the other

--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
[...]

PR #3547, opened 2018-10-29
Allow end of options during Thor array parsing

jquick:

Fixes https://github.com/inspec/inspec/issues/3459

This change allows you to use the typical end of options `--` after the Thor array type. This will allow you to do executions like this:
```
inspec exec --reporter json -- profile
```

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -6,6 +6,20 @@ require 'thor'
 require 'inspec/log'
 require 'inspec/profile_vendor'
 
+# Allow end of options during array type parsing
+# https://github.com/erikhuda/thor/issues/631
+class Thor::Arguments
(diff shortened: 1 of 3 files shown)


miah (approved):

lib/inspec/base_cli.rb
```diff
@@ -6,6 +6,20 @@
 require 'inspec/log'
 require 'inspec/profile_vendor'
 
+# Allow end of options during array type parsing
+# https://github.com/erikhuda/thor/issues/631
+class Thor::Arguments
```
clintoncwolfe, line 11: Code, docs and examples are all great - though I have the usual reservations about monkeypatching, I think we have no other practical choice here.  Thanks @jquick!

clintoncwolfe (approved):

Issue #3549, opened 2018-10-29
Allow each inspec check warning to be enabled or disabled


clintoncwolfe:
### 🎛 Description

🙋 feature request

Many linters and checkers have a mechanism to disable individual (or even categories) of warnings and errors.  Many users have expressed a desire for inspec check to be more lenient in a specific way for their environment (for example @james-stocks would like "No tests in control" to be ignored in some cases). In some cases, complex CI wrappers are added to ignore a particular warning.

One approach:
 * Internally name and categorize all warnings, and make sure the warning system uses the same code path for all warnings
 * Add CLI options to `inspec check` based on the warning names: `--no-warn empty_test`, for example.
 * Add a CLI option to `inspec check` to list all known warnings and their names.
 * Add support for reading the warning configuration from the inspec config JSON file.

### 🌍 InSpec and Platform Version

Currently on 3.0.x

### 🤔 Replication Case / Prior Art

RuboCop uses --except to disable rules

gcc uses a sophisticated warning suppression system - see [gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

Refs #3348

PR #3552, opened 2018-10-30
Check for non-unique control names within a profile during check and exec

### LINKED ISSUE #822: inspec does not check that every control ID is unique.
### Description

Currently, if you were to have two control ids that are identical the last one written is the impact value that will be used.  The inspec check should validate that all control ids are unique in a given profile.

RHEL 6
INSPEC 0.22.1
### Replication Case

```
inspec init profile secureprofile_01
cp secureprofile_01/controls/example.rb secureprofile_01/controls/tmp.rb
```

Modify the tmp.rb to be

``` ruby
title '/tmp profile'

control 'tmp-1.0' do
  impact 0.3
  title 'Create /tmp directory'
  desc 'A /tmp directory must exist'
  describe file('/tmp') do
    it { should be_directory }
  end
end

control 'tmp-1.1' do
  impact 0.3
  title '/tmp directory should be owned by root'
  desc 'A /tmp directory must be owned by the root user'
  describe file('/tmp') do
    it { should be_owned_by 'root' }
  end
end
```

Now when viewing the impact the impact will be 0.7 not 0.3.  The inspec check should catch that the control names matched to prevent inconsitant behavior.
### Possible Solutions

Add code to check for control id uniqueness.
### Stacktrace

None

clintoncwolfe:
When two controls have the same name in the same profile, current behavior is to silently merge them.  That is probably masking a problem - typical use case is to only redefine a control when inheriting from another profile.

This PR makes the following changes:
* Adds the profile_id field to the source_location datastructure attached to `Inspec::Rule` objects. This allows us to distinguish between legit inheritance (the profile IDs would be different) and a clobbering condition.
* Modifies `Inspec::Profile#check` to scan the rules after they are loaded, examining any that have a `merge_count > 0`, and then issuing an *error* when a clobber condition is detected. The error message from `inspec check` includes the source locations of both the clobberer and the clobberee.
* Modifies `Inspec::Runner#load` to perform a similar check, but only issue a *warning* to the log destination, during `inspec exec`.  Messaging is similar.
* Functional tests for `inspec check` and `inspec exec`.

Fixes #822 

`inspec exec` example:
```
[cwolfe@lodi inspec-cli]$ be inspec exec test/unit/mock/profiles/dupe-controls
[2018-10-29T20:33:56-04:00] WARN: Control dupe-01 at test/unit/mock/profiles/dupe-controls/controls/duplicates.rb:1 is duplicated within the same profile 'dupe-controls' - clobbered by test/unit/mock/profiles/dupe-controls/controls/duplicates.rb:12

Profile: A profile containing duplicated controls, refs github issue 822 (dupe-controls)
Version: 0.1.0
Target:  local://

  ✔  dupe-01: dupe-01-title-two
     ✔  true should eq true


```

`inspec check` example:
```
[cwolfe@lodi inspec-cli]$ be inspec check test/unit/mock/profiles/dupe-controls
Location:    test/unit/mock/profiles/dupe-controls
Profile:     dupe-controls
Controls:    1
Timestamp:   2018-10-29T20:34:46-04:00
Valid:       false

  ✖  test/unit/mock/profiles/dupe-controls/controls/duplicates.rb:1: Control dupe-01 is duplicated in profile dupe-controls - clobbered by test/unit/mock/profiles/dupe-controls/controls/duplicates.rb:12
  !  test/unit/mock/profiles/dupe-controls/controls/duplicates.rb:1: Control dupe-01 has no descriptions

Summary:     1 errors, 1 warnings
```


lib/inspec/runner.rb
```diff
@@ -102,6 +102,25 @@ def load
 
       all_controls.each do |rule|
         register_rule(rule) unless rule.nil?
+        # Check for duplicate controls and warn
```
jquick, line 105: I feel like this logic may be better here:
https://github.com/inspec/inspec/blob/master/lib/inspec/rule.rb#L249

Maybe save the profile_id in the history and warn out if a new one comes that is the same?

jquick (commented):
Looking pretty good! Just a placement question.

lib/inspec/runner.rb
clintoncwolfe: I don't think we should do UX work in the Rule - both check and exec will execute that, and if we want to tell the user about a problem, the two commands have different mechanisms for UX

In fact, the code does store the profile_id in the history - it's included in the source_location (see rule.rb line 41)

miah (approved):
Thanks @clintoncwolfe

lib/inspec/profile.rb
```diff
@@ -422,7 +421,24 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
...
         warn.call(sfile, sline, nil, id, "Control #{id} has impact < 0.0") if control[:impact].to_f < 0.0
         warn.call(sfile, sline, nil, id, "Control #{id} has no tests defined") if control[:checks].nil? or control[:checks].empty?
-      }
+
+        # Check for duplicate IDs within the same profile
+        rule = control[:rule_obj]
+        next unless Inspec::Rule.merge_count(rule) > 0
+        profile_id = Inspec::Rule.profile_id(rule)
+        merges = Inspec::Rule.merge_changes(rule)
+        merges = merges.select do |merge_location|
```
zenspider, line 430: It looks like y'all use the weirich method for blocks (`do`/`end` for side-effects, `{}` for result—something I strongly advocate for), so this select should use `{}`.

clintoncwolfe: I don't think we have any set pattern project-wide. I typically use do/end for multiline blocks, and `{}` for one-liners.

clintoncwolfe:
I'm going to withdraw this PR; if/when we implement a rule-based static check system, it would become irrelevant anyway. Refs #3549

PR #3553, opened 2018-10-30
Allow help args after Thor commands

jquick:

This will allow the common `-h, --help, help` args after the Thor command.

Ex:
```
jquick@Jareds-MBP:~/Chef/inspec
(jq/cli_help_commands)> _inspec exec help
Usage:
  inspec exec PATHS
...

jquick@Jareds-MBP:~/Chef/inspec
(jq/cli_help_commands)> _inspec plugin -h
Commands:
  inspec plugin help [COMMAND]               # Describe subcommands or one specific subcommand
  inspec plugin install [-v VERSION] PLUGIN  # Installs a plugin from rubygems.org, a gemfile, or a path 
...

jquick@Jareds-MBP:~/Chef/inspec
(jq/cli_help_commands)> _inspec habitat profile --help
Commands:
  inspec habitat profile create PATH     # Create a one-time Habitat artifact for the profile found at PATH
  inspec habitat profile help [COMMAND]  # Describe subcommands or one specific subcommand
...

jquick@Jareds-MBP:~/Chef/inspec
(jq/cli_help_commands)> _inspec habitat help profile
Commands:
  inspec habitat profile create PATH     # Create a one-time Habitat artifact for the profile found at PATH
  inspec habitat profile help [COMMAND]  # Describe subcommands or one specific subcommand
...
```

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -297,8 +297,11 @@ begin
   # This allows you to use any of the normal help commands after the normal args.
   help_commands = ['-h', '--help', 'help']
   (help_commands & ARGV).each do |cmd|
-    match = ARGV.delete(cmd)
-    ARGV.size > 1 ? ARGV.insert(-2, match) : ARGV.unshift(match)
+    # move the help argument to one place behind the end for Thor to digest
+    if ARGV.size > 1
+      match = ARGV.delete(cmd)
+      ARGV.insert(-2, match)
+    end
   end
 
   # Load v2 plugins
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -293,6 +293,14 @@ class Inspec::InspecCLI < Inspec::BaseCLI
 end
 
 begin
+  # Handle help commands
+  # This allows you to use any of the normal help commands after the normal args.
+  help_commands = ['-h', '--help', 'help']
+  (help_commands & ARGV).each do |cmd|
+    match = ARGV.delete(cmd)
+    ARGV.size > 1 ? ARGV.insert(-2, match) : ARGV.unshift(match)
+  end
+
   # Load v2 plugins
   v2_loader = Inspec::Plugin::V2::Loader.new
   v2_loader.load_all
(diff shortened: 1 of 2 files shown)


test/functional/inspec_exec_test.rb
```diff
@@ -52,6 +52,13 @@
     File.stat("#{outpath}/foo/bar/test.json").size.must_be :>, 0
   end
 
+  it 'can execute the exec help' do
+    out = inspec("exec --help")
```
clintoncwolfe, line 56: Could we have another test for `inspec exec help` specifically?  That's a different code path.

lib/inspec/cli.rb
```diff
@@ -293,6 +293,14 @@ def run_command(opts)
 end
 
 begin
+  # Handle help commands
+  # This allows you to use any of the normal help commands after the normal args.
+  help_commands = ['-h', '--help', 'help']
+  (help_commands & ARGV).each do |cmd|
+    match = ARGV.delete(cmd)
+    ARGV.size > 1 ? ARGV.insert(-2, match) : ARGV.unshift(match)
```
clintoncwolfe: The Perl programmer in me is doing a cartwheel, but I do think a comment explaining what you (think) you are doing would be good.

A bare ternary if without assignment.  Sheesh.  I can't believe RuboCop doesn't grouse about that :-)

clintoncwolfe (changes requested):
Two minor things - a request for a comment, and I think one more test would be good.

test/functional/inspec_exec_test.rb
clintoncwolfe: I'm looking for exactly 'inspec exec help'.  I know the implementation actually treats that the same as --help, but that is not obvious to an external user. If we change the implementation, I want to know that `inspec exec help` still gives a usage message, *not* trying to find a profile named `help`.

jquick: Added one for this as well

clintoncwolfe (approved):

PR #3559, opened 2018-10-31
documents inspec habitat profile setup command

nellshamrell:
I thought I needed to add a command to the habitat integration to create a habitat directory with a plan file in it in a profile....but it turned out that command already exists, it just wasn't documented anywhere.  This adds the info about that command to the docs.

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>

### DIFF
--- a/docs/habitat.md
+++ b/docs/habitat.md
@@ -128,7 +128,7 @@ $ habitat profile create ~/profiles/frontend1
 Create a Habitat directory that includes a plan file, config hooks, and more in a profile directory.
 
 This is the same process that is used by `inspec habitat profile create` - but this adds the generated Habitat
- directory and file to your system so that you can commit them to source control. If you commit these files to Github, you can connect that plan to the [https://www.habitat.sh/docs/using-builder/](Habitat Builder Service).
+ directory and file to your system so that you can commit them to source control. If you commit these files to GitHub, you can connect that plan to the [Habitat Builder Service](https://www.habitat.sh/docs/using-builder/).
 
 #### Syntax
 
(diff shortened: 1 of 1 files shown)


docs/habitat.md
```diff
@@ -151,6 +123,38 @@ $ habitat profile create ~/profiles/frontend1
 [2017-03-28T13:30:18-04:00] INFO: Copying artifact to /Users/aleff...
 ```
 
+### inspec habitat profile setup
+
+Create a Habitat directory that includes a plan file, config hooks, and more in a profile directory.
+
+This is the same process that is used by `inspec habitat profile create` - but this adds the generated Habitat
+ directory and file to your system so that you can commit them to source control. If you commit these files to Github, you can connect that plan to the [https://www.habitat.sh/docs/using-builder/](Habitat Builder Service).
```
clintoncwolfe: I think you have that hyperlink syntax backwards - it's:
```
[Link Text](url)
```
[Like This](https://docs.chef.io)

It misrenders:
screen shot 2018-10-31 at 20 33 47

jerryaldrichiii (approved):
I would change`Github` to `GitHub`, but won't hold up this excellent documentation over it!

Thanks!

clintoncwolfe (dismissed):
Thanks so much for this!  I tried out the instructions and they work great.  Just one little gotcha, a broken link.  Thanks!

docs/habitat.md
jerryaldrichiii: Great catch! A heuristic I use is "squared circle". Not sure if it helps though.

nellshamrell:
Fixed the link and the spelling of GitHub - ty for the suggestions :)

jquick (approved):
Thanks @nellshamrell

miah (approved):
Thank you @nellshamrell !

PR #3561, opened 2018-11-01
Fixes (some) ruby warnings related to functional tests

TheLonelyGhost:


test/functional/inspec_exec_test.rb
```diff
@@ -53,7 +53,7 @@
   end
 
   it 'can execute the profile with a target_id passthrough' do
-    outpath = Dir.tmpdir
+    Dir.tmpdir
```
clintoncwolfe: I still don't see this being used - we're not in a tmpdir block.  Could this be removed entirely?

clintoncwolfe (approved):
Thanks - just one note about whether a tmpdir call could be removed, but it's not a blocker.

jquick (approved):
Thanks @TheLonelyGhost

PR #3562, opened 2018-11-01
Add Git SSH and HTTP basic auth support to `inspec exec`

### LINKED ISSUE #3555: Add support for `inspec exec` and private Git repos (GitHub or GitLab)
### 🎛 Description

🙋 feature request

Currently there is no method to use `inspec exec` and a private GitHub repo. Below are some examples that I think should work but do not:

#### Via SSH (`git@github.com`)

```
inspec exec git@github.com:chef/private-testing-profile.git
```

Fails with:
```
Could not fetch inspec profile in "git@github.com:chef/private-testing-profile.git"
```
This fails since `target` is passes as a String and no fetcher knows how to handle it. I think the Git fetcher should parse this string and do a Git clone.

#### Via `username:password` and HTTPS

```
PASSWORD=hunter2 inspec exec https://jerryaldrichiii:$PASSWORD@github.com/chef/private-testing-profile.git
```

Fails with:

```
userinfo not supported.  [RFC3986]
```

> I think we should also prompt for password if only username is given, but I'll open that for discussion.

#### Via [git-credential-cache](https://git-scm.com/docs/git-credential-cache)

```
git config credential.helper cache
git ls-remote https://jerryaldrichiii@github.com/chef/private-testing-profile
# Auth happens here
inspec exec https://github.com/chef/private-testing-profile
```

Fails with:

```
[2018-10-30T11:19:40-07:00] WARN: URL target https://github.com/chef/private-testing-profile.git transformed to https://github.com/chef/private-testing-profile/archive/master.tar.gz. Consider using the git fetcher
OpenURI::HTTPError: 404 Not Found
```

### 💁 Possible Solutions

I think leveraging the Git fetcher when URLs do not end with `.tar.gz`/`.zip` or when they end with `.git` might be the solution. That will allow all but the username only method to work out of the box (since it uses `git clone`). The username only method does not prompt for a password when ran via InSpec so we will need to capture it somehow.

> Example below requires you to change the Git fetcher to use the `target` as a String and not just a Hash

Example:

```
inspec exec https://jerryaldrichiii@github.com/chef/private-testing-profile.git
```

Fails with:

```
Error running 'git ls-remote "https://jerryaldrichiii@github.com/chef/private-testing-profile.git" "master*"': fatal: could not read Password for 'https://jerryaldrichiii@github.com': No such device or address
```

jerryaldrichiii:
This adds support and documentation for the following `inspec exec` usages:
                      
Git via SSH (uses SSH keys if repo is private):     
```                                                                                                                                   
inspec exec git@github.com:dev-sec/linux-baseline.git
```                                                                                                                              
                                                      
Git via HTTPS (.git is required):                                                                                                        
```          
inspec exec https://github.com/dev-sec/linux-baseline.git
```               
                          
Private Git via HTTPS (.git is required):
```
inspec exec https://API_TOKEN@github.com/dev-sec/linux-baseline.git
```                              
                                                 
Private Git via HTTPS and cached credentials (.git is required):
```                                                     
git config credential.helper cache                       
git ls-remote https://github.com/dev-sec/linux-baseline.git
inspec exec https://github.com/dev-sec/linux-baseline.git
```                      
                                  
Web hosted fileshare (supports .zip): 
```              
inspec exec https://webserver/linux-baseline.tar.gz
```                                                                                                    
                                            
Web hosted fileshare with basic authentication (supports .zip):
```                                   
inspec exec https://username:password@webserver/linux-baseline.tar.gz
```

This closes #3555

### DIFF
--- a/lib/fetchers/git.rb
+++ b/lib/fetchers/git.rb
@@ -29,7 +29,11 @@ module Fetchers
     priority 200
 
     def self.resolve(target, opts = {})
-      new(target[:git], opts.merge(target)) if target.respond_to?(:has_key?) && target.key?(:git)
+      if target.is_a?(String)
+        new(target) if target.start_with?('git@') || target.end_with?('.git')
+      elsif target.respond_to?(:has_key?) && target.key?(:git)
+        new(target[:git], opts.merge(target))
+      end
     end
 
     def initialize(remote_url, opts = {})
--- a/lib/fetchers/git.rb
+++ b/lib/fetchers/git.rb
@@ -30,7 +30,7 @@ module Fetchers
 
     def self.resolve(target, opts = {})
       if target.is_a?(String)
-        new(target) if target.start_with?('git@') || target.end_with?('.git')
+        new(target, opts) if target.start_with?('git@') || target.end_with?('.git')
       elsif target.respond_to?(:has_key?) && target.key?(:git)
         new(target[:git], opts.merge(target))
       end
--- a/lib/fetchers/url.rb
+++ b/lib/fetchers/url.rb
@@ -95,6 +95,7 @@ module Fetchers
 
     def initialize(url, opts)
       @target = url
+      @target_uri = parse_uri(@target)
       @insecure = opts['insecure']
       @token = opts['token']
       @config = opts
@@ -119,6 +120,16 @@ module Fetchers
     end
 
     private
+
+    def parse_uri(target)
+      if target.is_a?(String)
+        URI.parse(target)
+      elsif target[:url]
+        URI.parse(target[:url])
+      end
+    rescue URI::Error
+      nil
+    end
 
     def sha256
       file = @archive_path || temp_archive_path
@@ -155,9 +166,8 @@ module Fetchers
         version: @config['profile'][2],
       }.to_json
 
-      uri = URI.parse(@target)
       opts = http_opts
-      opts[:use_ssl] = uri.scheme == 'https'
+      opts[:use_ssl] = @target_uri.scheme == 'https'
 
       if @insecure
         opts[:verify_mode] = OpenSSL::SSL::VERIFY_NONE
@@ -165,12 +175,12 @@ module Fetchers
         opts[:verify_mode] = OpenSSL::SSL::VERIFY_PEER
       end
 
-      req = Net::HTTP::Post.new(uri)
+      req = Net::HTTP::Post.new(@target_uri)
       opts.each do |key, value|
         req.add_field(key, value)
       end
       req.body = json
-      res = Net::HTTP.start(uri.host, uri.port, opts) { |http|
+      res = Net::HTTP.start(@target_uri.host, @target_uri.port, opts) { |http|
         http.request(req)
       }
 
@@ -188,7 +198,7 @@ module Fetchers
     def download_archive_to_temp
       return @temp_archive_path if !@temp_archive_path.nil?
       Inspec::Log.debug("Fetching URL: #{@target}")
-      remote = open(@target, http_opts)
+      remote = open_via_uri(@target)
       @archive_type = file_type_from_remote(remote) # side effect :(
       archive = Tempfile.new(['inspec-dl-', @archive_type])
       archive.binmode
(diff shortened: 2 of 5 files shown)


aaronlippold:
Great work

jquick:
Can we get some tests that show this fetcher activates for the specified string urls

lib/fetchers/git.rb
```diff
@@ -29,7 +29,11 @@ class Git < Inspec.fetcher(1)
     priority 200
 
     def self.resolve(target, opts = {})
-      new(target[:git], opts.merge(target)) if target.respond_to?(:has_key?) && target.key?(:git)
+      if target.is_a?(String)
+        new(target) if target.start_with?('git@') || target.end_with?('.git')
```
clintoncwolfe: In the other branch of the conditional, `opts` is passed along to `new`, but here, it isn't.  Is that intentional?

lib/fetchers/url.rb
```diff
@@ -120,6 +121,16 @@ def to_s
 
     private
 
+    def parse_uri(target)
+      if target.is_a?(String)
+        URI.parse(target)
+      elsif target[:url]
+        URI.parse(target[:url])
+      end
+    rescue URI::Error
```
clintoncwolfe: If we can't parse the URI for some reason, I'd think we should communicate that back to the user somehow.

Similar logic is used in `resolve_from_string` at line 39, but the circumstances are different: `resolve` is speculative, and is supposed to return nil if the fetcher could not handle the input.

`parse_uri` seems to only be called from `initialize`, which at that point should not be speculative - we've committed to servicing what is thought to be a valid URL. If that is not true, we need to abort.

lib/inspec/cli.rb
```diff
@@ -168,9 +168,70 @@ def archive(path)
...
+      ```
+      inspec exec /path/to/a_test.rb
+      ```
+
+    Git via SSH (uses SSH keys if repo is private):
+      ```
+      inspec exec git@github.com:dev-sec/linux-baseline.git
+      ```
+
+    Git via HTTPS (.git is required):
```
clintoncwolfe: Consider clarifying that "(the '.git' suffix is required)"

On first reading I thought you meant that the `git` executable was required, and the `.` was a typo.

lib/inspec/cli.rb
```diff
@@ -168,9 +168,70 @@ def archive(path)
...
+      ```
+
+    Private Git via HTTPS and cached credentials (.git is required):
+      ```
+      git config credential.helper cache
+      git ls-remote https://github.com/dev-sec/linux-baseline.git
+      inspec exec https://github.com/dev-sec/linux-baseline.git
+      ```
+
+    Web hosted fileshare (supports .zip):
```
clintoncwolfe: Suggest rephrasing as "(also supports .zip)"

lib/inspec/cli.rb
```diff
@@ -168,9 +168,70 @@ def archive(path)
...
+      ```
+      inspec exec /path/to/profile
+      ```
+
+    Local single test (doesn't allow attributes or custom resources)
+      ```
+      inspec exec /path/to/a_test.rb
+      ```
+
+    Git via SSH (uses SSH keys if repo is private):
```
clintoncwolfe: Assuming this clones, valid SSH keys are required whether the repo is public or not:

screen shot 2018-11-08 at 00 17 11

clintoncwolfe (changes requested):
Great work, @jerry!

One concern about exception handling in the URL fetcher and one question about how options are handled in the git fetcher are my blockers; otherwise it's just some docs clarifications.

Thanks so much, I'm sure many people will be happy to have this!

lib/fetchers/git.rb
jerryaldrichiii: Hmm...I did not see any `opts` passed during testing, but I suppose passing `opts` couldn't hurt. I'll add it.

lib/inspec/cli.rb
jerryaldrichiii: This is why we have multiple eyes looking! Will change. Thanks!

lib/inspec/cli.rb
jerryaldrichiii: Couldn't hurt. Will do.

lib/inspec/cli.rb
jerryaldrichiii: Hmm...interesting...didn't know that. Will update the the comment.

jerryaldrichiii: Hmm...interesting...didn't know that. Will update the comment.

lib/fetchers/url.rb
jerryaldrichiii: Great catch, I'll rework to attempt to a parse regardless of what is passed.

clintoncwolfe (approved):

jquick (approved):
Thanks @jerryaldrichiii

aaronlippold:
will this work with `inspec json .... ` as well.

mechastorm:
just wondering if this useful feature has been documented in the doc site?

aaronlippold:
I believe so but if not please feel free to push up a PR :-)

PR #3564, opened 2018-11-01
aws_vpc: accept 17 hexadecimal characters for vpc_id

### LINKED ISSUE #3518: aws_vpc expect handle only 8 hexadecimal characters instead both 8 and 17
Description:
InSpec doesn't handle vpc_id which contain 17 characters and expect just 8

Case:
describe aws_subnets.where(vpc_id: 'vpc-012345678987654321') do
  it { should exist }

Expected result:
✔   aws_subnets exist
Actual result:
× The vpc ID 'vpc-01234567898765432' does not exist
VPC_ID should be in the format "vpc-" followed by 8 or 17 hexadecimal characters.

inspec version:
2.3.24

kchistova:
Signed-off-by kchistova <kchistova@gmail.com>

Fixes #3518 	
EDIT: (@clintoncwolfe) - move issue ref to PR desc

### DIFF
--- a/lib/resources/aws/aws_vpc.rb
+++ b/lib/resources/aws/aws_vpc.rb
@@ -30,8 +30,8 @@ class AwsVpc < Inspec.resource(1)
       allowed_scalar_type: String,
     )
 
-    if validated_params.key?(:vpc_id) && validated_params[:vpc_id] !~ /^vpc\-[0-9a-f]{8}/
-      raise ArgumentError, 'aws_vpc VPC ID must be in the format "vpc-" followed by 8 hexadecimal characters.'
+    if validated_params.key?(:vpc_id) && validated_params[:vpc_id] !~ /^vpc\-([0-9a-f]{8})|(^vpc\-[0-9a-f]{17})$/
+      raise ArgumentError, 'aws_vpc VPC ID must be in the format "vpc-" followed by 8 or 17 hexadecimal characters.'
     end
 
     validated_params
--- a/test/unit/resources/aws_vpc_test.rb
+++ b/test/unit/resources/aws_vpc_test.rb
@@ -16,12 +16,20 @@ class AwsVpcConstructorTest < Minitest::Test
     AwsVpc.new
   end
 
-  def test_accepts_vpc_id_as_scalar
+  def test_accepts_vpc_id_as_scalar_eight_sign
     AwsVpc.new('vpc-12345678')
   end
 
-  def test_accepts_vpc_id_as_hash
+  def test_accepts_vpc_id_as_scalar
+    AwsVpc.new('vpc-12345678987654321')
+  end
+
+  def test_accepts_vpc_id_as_hash_eight_sign
     AwsVpc.new(vpc_id: 'vpc-1234abcd')
+  end
+
+  def test_accepts_vpc_id_as_hash
+    AwsVpc.new(vpc_id: 'vpc-abcd123454321dcba')
   end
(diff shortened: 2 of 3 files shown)


clintoncwolfe:
Thanks @kchistova! Could you take a moment and sign your commits using `git commit -s`?  We use commit signing to indicate that the contributor has read and understood the [Developer Certification of Origin](https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco), which basically means it's your code you're contributing, it doesn't introduce a new license, etc.  Thanks!

clintoncwolfe:
Hrm, I think the DCO detector issued a false negative here - the commit is signed-off.

lib/resources/aws/aws_vpc.rb
```diff
@@ -30,8 +30,8 @@ def validate_params(raw_params)
       allowed_scalar_type: String,
     )
 
-    if validated_params.key?(:vpc_id) && validated_params[:vpc_id] !~ /^vpc\-[0-9a-f]{8}/
-      raise ArgumentError, 'aws_vpc VPC ID must be in the format "vpc-" followed by 8 hexadecimal characters.'
+    if validated_params.key?(:vpc_id) && validated_params[:vpc_id] !~ /^vpc\-([0-9a-f]{8})|(^rtb\-[0-9a-f]{17})$/
```
clintoncwolfe: `rtb-` is the prefix for *routing tables*, I think - don't you mean `vpc-` in both places?

test/unit/resources/aws_vpc_test.rb
```diff
@@ -16,14 +16,22 @@ def test_empty_params_ok
     AwsVpc.new
   end
 
-  def test_accepts_vpc_id_as_scalar
+  def test_accepts_vpc_id_as_scalar_eight_sigh
```
clintoncwolfe: What does `eight_sigh` mean?

test/unit/resources/aws_vpc_test.rb
```diff
@@ -16,14 +16,22 @@ def test_empty_params_ok
     AwsVpc.new
```
clintoncwolfe, line 16: Could you also update the examples in docs/resources/aws_vpc.md.erb?

clintoncwolfe (changes requested):
Thanks for this - it really helps. 

A few things, all minor - only the regex change is a real blocker.  Thanks!

lib/resources/aws/aws_vpc.rb
kchistova: Yes, sorry

test/unit/resources/aws_vpc_test.rb
kchistova: 8 sign i meant

lib/resources/aws/aws_vpc.rb
kchistova: Fixed

test/unit/resources/aws_vpc_test.rb
kchistova: Changed to eight_sign

test/unit/resources/aws_vpc_test.rb
kchistova: Done

clintoncwolfe (approved):

jquick (approved):
Looks good to me! Thanks @kchistova

PR #3565, opened 2018-11-01
Detect FreeBSD in port resource

### LINKED ISSUE #3563: `Port` resource not supported on FreeBSD, but helper code exists.
https://github.com/inspec/inspec/blob/d622f469672aea8708cc6331dc638c4cd12d6cf0/lib/resources/port.rb#L69

on FreeBSD this returns 'bsd' as the OS family, so port matching does not appear supported on the platform.

However the code exists and should work, I believe it's just not matching here.

dijit:
I understand this is rather zealous, since it will match openbsd- but at least it will allow FreeBSD to work, since we have code for that already.

Fixes #3563
Edit (@clintoncwolfe) - move issue ref into PR desc

### DIFF
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -66,7 +66,7 @@ module Inspec::Resources
         LsofPorts.new(inspec)
       elsif os.windows?
         WindowsPorts.new(inspec)
-      elsif ['freebsd'].include?(os[:family])
+      elsif ['bsd'].include?(os[:family])
         FreeBsdPorts.new(inspec)
       elsif os.solaris?
         SolarisPorts.new(inspec)
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -66,7 +66,7 @@ module Inspec::Resources
         LsofPorts.new(inspec)
       elsif os.windows?
         WindowsPorts.new(inspec)
-      elsif ['bsd'].include?(os[:family])
+      elsif 'freebsd'.include?(os[:family])
         FreeBsdPorts.new(inspec)
       elsif os.solaris?
         SolarisPorts.new(inspec)


dijit:
Nvm, simply dropping the brackets solved the issue too.

clintoncwolfe:
Thanks @dijit!  Could you take a moment and sign your commits using `git commit -s`?  We use commit signing to indicate that the contributor has read and understood the [Developer Certification of Origin](https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco), which basically means it's your code you're contributing, it doesn't introduce a new license, etc.  Thanks!

lib/resources/port.rb
```diff
@@ -66,7 +66,7 @@ def port_manager_for_os
         LsofPorts.new(inspec)
       elsif os.windows?
         WindowsPorts.new(inspec)
-      elsif ['freebsd'].include?(os[:family])
+      elsif 'freebsd'.include?(os[:family])
```
clintoncwolfe, line 69: @jquick do we auto-generate predicates for each platform?  If so, we could have `os.freebsd?`.

`os[:family]` is 'bsd' here, which is correct; the platform would be `freebsd`.  I think relying on the fact that one is a substring of the other is not the intent of the code - we are asking specifically if we are on freebsd.

So I agree we need a conditional to match FreeBSD, but I think it should be something higher-level like `os.freebsd?`.

jquick (approved):
Thanks @dijit

clintoncwolfe (changes requested):
Thanks for the contribution! However, I think the test that is being used could be made safer. It's not obvious how to do that, so I've asked our platform detect expert to weigh in.

lib/resources/port.rb
jquick: Yes with the new platforms work `os.freebsd?` should work as expected.

clintoncwolfe:
As this is a one-line change, I'm going to adopt it to add the os-based predicate call.

clintoncwolfe:
Hrm, looks like we don't get platform-name-specific predicates automatically.  I spun up a freebsd 9.3 vagrantbox from our Kitchen config, and played:

```
[cwolfe@lodi inspec-review]$ be inspec version
3.0.28
[cwolfe@lodi inspec-review]$ be inspec detect -t ssh://vagrant@127.0.0.1:2222

== Platform Details

Name:      freebsd
Families:  bsd, unix, os
Release:   9.3-RELEASE-p49
Arch:      amd64
[cwolfe@lodi inspec-review]$ be inspec shell -t ssh://vagrant@127.0.0.1:2222 -c 'os.freebsd?' 2>&1 | head -2
bundler: failed to load command: inspec (/Users/cwolfe/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/bin/inspec)
NoMethodError: undefined method `freebsd?' for Operating System Detection:#<Class:0x00007ff2fed65d28>
[cwolfe@lodi inspec-review]$ be inspec shell -t ssh://vagrant@127.0.0.1:2222 -c 'os.bsd?'
true
[cwolfe@lodi inspec-review]$ be inspec shell -t ssh://vagrant@127.0.0.1:2222 -c 'os.name'
freebsd
[cwolfe@lodi inspec-review]$ be inspec shell -t ssh://vagrant@127.0.0.1:2222 -c 'os.family'
bsd
```

It does appear that sockstat (the CLI tool that the code relies on) is standard on FreeBSD and NetBSD.  Note that MacOS X is also in the bsd family, but does not include sockstat; however we test for Darwin earlier in the conditionals. 
I think changing the conditional to be `os.bsd?` would be OK.

clintoncwolfe:
Adopted on #3579; thanks again for your contribution!

PR #3567, opened 2018-11-01
Change usage of `Dir.home` to `Inspec.config_dir`

jerryaldrichiii:
This prevents errors on systems without `ENV['HOME']` (e.g. Habitat services) by allowing the specification of the location of `.inspec` via `ENV['INSPEC_CONFIG_DIR']`.


jquick (approved):
Thanks @jerryaldrichiii

miah (approved):
Thanks @jerryaldrichiii

PR #3575, opened 2018-11-03
Create a class to handle the plugins.json file

### LINKED ISSUE #3350: Make a class for the plugins.json file
### 🎛 Description

🐛 bug report, in the sense of a code smell

Both Inspec::Plugin::V2::Loader and Inspec::Plugin::V2::Installer need to interact with the plugin configuration file.  Currently, both simply create it if needed, both read it, Loader validates it, and Installer modifies it; Registry exposes primitives to answer question about it.

So, yeah, no one thing has responsibility for this, and one thing should.  Let's make that thing.

2.2.78

### 💁 Possible Solutions

Make a class that is responsible for creating, reading, modifying, validating, and querying the contents of the plugins.json file.

Should also be the authority on where it is, too.

clintoncwolfe:
Fixes #3350 	

This is an internal refactor for the plugins system.  As tech debt, we had implemented the reading, validation, and writing of the plugin state file, `~/.inspec/plugins.json` scattered accross several files, with no one class having responsibility.  Testing was minimal.

This PR creates a class to handle the statefile and gate modifications to it with validation.  Unit tests are included.

The Loader and Installer classes are modified to use the ConfigFile class, and their tests are updated.


jquick (approved):
Code and tests look good to me! Thanks @clintoncwolfe

miah (approved):
Looks great, excellent test coverage and error handling. A++ Would review again =)

PR #3576, opened 2018-11-05
aws_security_group: Query against other security group ids in allow_* matchers

j00p34:
This pull request adds the functionality to use security group references in the controls for security group

### DIFF
--- a/docs/resources/aws_security_group.md.erb
+++ b/docs/resources/aws_security_group.md.erb
@@ -91,7 +91,7 @@ The following examples show how to use this InSpec audit resource.
     end
 
     # Ensure that the canary_deployments Security Group only allows access from one specific security group id on port 443.
-    describe aws_security_group(group_name: 'careful_updates') do
+    describe aws_security_group(group_name: 'canary_deployments') do
 
       it { should allow_in_only(port: 443, security_group: "sg-33334444") }
     end
--- a/docs/resources/aws_security_group.md.erb
+++ b/docs/resources/aws_security_group.md.erb
@@ -12,7 +12,6 @@ SGs are a networking construct which contain ingress and egress rules for networ
 
 While this resource provides facilities for searching inbound and outbound rules on a variety of criteria, there is currently no support for performing matches based on:
 
- * References to other Security Groups
  * References to VPC peers or other AWS services (that is, no support for searches based on 'prefix lists').
 
 <br>
@@ -90,6 +89,26 @@ The following examples show how to use this InSpec audit resource.
       # Expect exactly three rules.
       its('outbound_rules.count') { should cmp 3 }
     end
+
+    # Ensure that one of the security groups in a list allows access from one of the groups in another list
+    # In this example we get the list of groups from Terraform output
+    # load data from Terraform output
+    content = inspec.profile.file("terraform.json")
+    params = JSON.parse(content)
+
+      control 'master to node access security group' do
+        desc 'one of the node security groups should allow all access from:
+      one of the master security groups, so masters can reach all nodes'
+        describe.one do
+          params['node_security_group_ids']['value'].each do |nodeSG|
+            params['master_security_group_ids']['value'].each do |masterSG|
+              describe aws_security_group(id: nodeSG) do
+                it { should allow_in(security_group: masterSG) }
+              end
+            end
+          end
+        end
+      end
 
 <br>
 
--- a/docs/resources/aws_security_group.md.erb
+++ b/docs/resources/aws_security_group.md.erb
@@ -84,24 +84,28 @@ The following examples show how to use this InSpec audit resource.
       end
 
       # If you have one rule with two CIDRs:
-      it { should allow_out(ipv4_range: [ '10.7.23.12/32', '10.8.23.12/32' ] }
+      it { should allow_out(ipv4_range: [ '10.7.23.12/32', '10.8.23.12/32' ]) }
 
       # Expect exactly three rules.
       its('outbound_rules.count') { should cmp 3 }
     end
 
-    # Ensure that one of the security groups in a list allows access from one of the groups in another list
-    # In this example we get the list of groups from Terraform output
-    # load data from Terraform output
-    content = inspec.profile.file("terraform.json")
-    params = JSON.parse(content)
+    # Ensure that the canary_deployments Security Group only allows access from one specific security group id on port 443.
+    describe aws_security_group(group_name: 'careful_updates') do
+
+      it { should allow_in_only(port: 443, security_group: "sg-33334444") }
+    end
+
+    # Ensure that one of the security groups in a list allows access from one of the groups in another list.
+    # If you have two lists of groups, check if one of the groups in the first list allows access for one of the
+    # groups in the second list.
 
       control 'master to node access security group' do
         desc 'one of the node security groups should allow all access from:
       one of the master security groups, so masters can reach all nodes'
         describe.one do
-          params['node_security_group_ids']['value'].each do |nodeSG|
-            params['master_security_group_ids']['value'].each do |masterSG|
+          [ 'sg-11112222', 'sg-33334444' ].each do |nodeSG|
+            [ 'sg-55556666', 'sg-77778888' ].each do |masterSG|
               describe aws_security_group(id: nodeSG) do
                 it { should allow_in(security_group: masterSG) }
               end
@@ -279,6 +283,7 @@ The matchers accept a key-value list of search criteria.  For a rule to match, i
   * position - A one-based index into the list of rules. If provided, this restricts the evaluation to the rule at that position. You may also use the special values `:first` and `:last`. `position` may also be used to enable `allow_in_only` and `allow_out_only` to work with multi-rule Security Groups.
   * protocol - Specifies the IP protocol. 'tcp', 'udp', and 'icmp' are some typical values. The string "-1" or 'any' is used to indicate any protocol.
   * to_port - Determines if a rule exists whose port range ends at the specified number. The word 'to_' does *not* relate to inbound/outbound directionality; it relates to the port range ("counting _to_"). `to_port` is an exact criterion; so if the rule allows 1000-2000 and you specify a `to_port` of 1999, it does not match.
+  * security_group - Specifies a security-group id, to be checked as permissible origin (for `allow_in`) or destination (for `allow_out`) for traffic. Each AWS Security Group rule may have multiple allowed source or destination security groups.
 
     describe aws_security_group(group_name: 'mixed-functionality-group') do
       # Allow RDP from defined range
@@ -302,6 +307,9 @@ The matchers accept a key-value list of search criteria.  For a rule to match, i
 
       # Do not allow unrestricted IPv4 access.
       it { should_not allow_in(ipv4_range: '0.0.0.0/0') }
+
+      # Allow unrestricted access from security-group.
+      it { should allow_in(security_group: 'sg-11112222') }
     end
 
     # Suppose you have a Group that should allow SSH and RDP from
--- a/lib/resources/aws/aws_security_group.rb
+++ b/lib/resources/aws/aws_security_group.rb
@@ -194,7 +194,7 @@ class AwsSecurityGroup < Inspec.resource(1)
     return true unless criteria.key?(:security_group)
     query = criteria[:security_group]
     return false unless rule[:user_id_group_pairs]
-    rule[:user_id_group_pairs].any? {|group| query == group[:group_id]}
+    rule[:user_id_group_pairs].any? { |group| query == group[:group_id] }
   end
 
   def validate_params(raw_params)
--- a/lib/resources/aws/aws_security_group.rb
+++ b/lib/resources/aws/aws_security_group.rb
@@ -44,6 +44,7 @@ class AwsSecurityGroup < Inspec.resource(1)
     # allow_{in_out}_only require either a single-rule group, or you
     # to select a rule using position.
     return false unless rules.count == 1 || criteria.key?(:position)
+    return false if criteria.key?(:security_group) && !rules[0][:user_id_group_pairs].count == 1
     criteria[:exact] = true
     allow(rules, criteria)
   end
@@ -58,6 +59,7 @@ class AwsSecurityGroup < Inspec.resource(1)
       matched &&= allow__match_protocol(rule, criteria)
       matched &&= allow__match_ipv4_range(rule, criteria)
       matched &&= allow__match_ipv6_range(rule, criteria)
+      matched &&= allow__match_security_group(rule, criteria)
       matched
     end
   end
@@ -67,6 +69,7 @@ class AwsSecurityGroup < Inspec.resource(1)
       :from_port,
       :ipv4_range,
       :ipv6_range,
+      :security_group,
       :port,
       :position,
       :protocol,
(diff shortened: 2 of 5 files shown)


lib/resources/aws/aws_security_group.rb
```diff
@@ -44,6 +44,7 @@ def allow_only(rules, criteria)
     # allow_{in_out}_only require either a single-rule group, or you
     # to select a rule using position.
     return false unless rules.count == 1 || criteria.key?(:position)
+    return false if criteria.key?(:security_group) && !rules[0][:user_id_group_pairs].count == 1
```
clintoncwolfe: I don't think we want to add this case here. If `criteria.key?(:position)` is true, and refers to the second rule, but the first rule does not have a security group reference, this will return a false negative.

test/unit/resources/aws_security_group_test.rb
```diff
@@ -320,6 +340,114 @@ def describe_security_groups(query)
...
+          group_name: 'zeta',
+          vpc_id: 'vpc-12345678',
+          ip_permissions: [
+            OpenStruct.new({
+              from_port: nil,
+              to_port: nil,
+              ip_protocol: "-1",
+              user_id_group_pairs: [
+                OpenStruct.new({
+                  description: "String",
```
clintoncwolfe: I think this is confusing - if you are not using or testing a field, it is better to omit it.  If the value 'String' is present, that might imply to someone that the value is important, or that we are setting up type checking - when in fact it isn't.

docs/resources/aws_security_group.md.erb
```diff
@@ -91,6 +90,26 @@ The following examples show how to use this InSpec audit resource.
...
+          params['node_security_group_ids']['value'].each do |nodeSG|
+            params['master_security_group_ids']['value'].each do |masterSG|
+              describe aws_security_group(id: nodeSG) do
+                it { should allow_in(security_group: masterSG) }
+              end
+            end
+          end
+        end
+      end
+
```
clintoncwolfe, line 116: Please add documentation under the Matchers section for this criteria - around line 281.

docs/resources/aws_security_group.md.erb
```diff
@@ -91,6 +90,26 @@ The following examples show how to use this InSpec audit resource.
       its('outbound_rules.count') { should cmp 3 }
     end
 
+    # Ensure that one of the security groups in a list allows access from one of the groups in another list
+    # In this example we get the list of groups from Terraform output
+    # load data from Terraform output
+    content = inspec.profile.file("terraform.json")
```
clintoncwolfe: I think this is a powerful example, but will also be intimidating to new users.  We also don't want to imply that you must use Terraform to use this feature.  I'd recommend replacing this with a simplified example.

clintoncwolfe (changes requested):
This is a fantastic contribution, @j00p34 !  Just a couple of minor adjustments, then I think we can move ahead.

aaronlippold:
Hello, just my two cents, I thought that example was great but I understand
the feedback. Could we as a middle ground perhaps add in the simplified
example and also keep the more complex example labeling the minimal example
as the minimum needed but as an example of an advanced use case see this
example. That way we make sure to show people what's needed and also what's
possible.

clintoncwolfe:
I'm not opposed to having a very simple example (which strictly focuses on the new capability) as well as a moderately complex example showing off what is possible - I think that's great.  

Keep in mind the goal of this document - to be a clear reference as to what the resource offers.

docs/resources/aws_security_group.md.erb
j00p34: I had some doubt about this already, I chose to put it in because it took me a while to  figure out how to do this properly. This is an actual use case for us and I do think it might be useful. Would it be acceptable if I remove the reference to terraform, and add a less complex example?

lib/resources/aws/aws_security_group.rb
j00p34: :-)  Missed that one.
I've added a test for that specific case and updated the code.

test/unit/resources/aws_security_group_test.rb
j00p34: That's a good example of copy-paste programming. I copied that stuff from the API docs vanilla example. Forgot to change the content after that. I've removed the strings optional according to the documentation. The remaining have sensible default content now.

docs/resources/aws_security_group.md.erb
j00p34: done

docs/resources/aws_security_group.md.erb
j00p34: @clintoncwolfe PTAL

docs/resources/aws_security_group.md.erb
```diff
@@ -84,24 +84,28 @@ The following examples show how to use this InSpec audit resource.
...
       its('outbound_rules.count') { should cmp 3 }
     end
 
-    # Ensure that one of the security groups in a list allows access from one of the groups in another list
-    # In this example we get the list of groups from Terraform output
-    # load data from Terraform output
-    content = inspec.profile.file("terraform.json")
-    params = JSON.parse(content)
+    # Ensure that the canary_deployments Security Group only allows access from one specific security group id on port 443.
+    describe aws_security_group(group_name: 'careful_updates') do
```
clintoncwolfe: +1 for a simple example; however 'canary_deployments' != 'careful_updates' :-)

docs/resources/aws_security_group.md.erb
```diff
@@ -84,24 +84,28 @@ The following examples show how to use this InSpec audit resource.
...
+    # groups in the second list.
 
       control 'master to node access security group' do
         desc 'one of the node security groups should allow all access from:
       one of the master security groups, so masters can reach all nodes'
         describe.one do
-          params['node_security_group_ids']['value'].each do |nodeSG|
-            params['master_security_group_ids']['value'].each do |masterSG|
+          [ 'sg-11112222', 'sg-33334444' ].each do |nodeSG|
+            [ 'sg-55556666', 'sg-77778888' ].each do |masterSG|
```
clintoncwolfe, line 108: This is great as a moderately complicated capability example.  Thanks!

docs/resources/aws_security_group.md.erb
```diff
@@ -260,6 +283,7 @@ The matchers accept a key-value list of search criteria.  For a rule to match, i
   * position - A one-based index into the list of rules. If provided, this restricts the evaluation to the rule at that position. You may also use the special values `:first` and `:last`. `position` may also be used to enable `allow_in_only` and `allow_out_only` to work with multi-rule Security Groups.
   * protocol - Specifies the IP protocol. 'tcp', 'udp', and 'icmp' are some typical values. The string "-1" or 'any' is used to indicate any protocol.
   * to_port - Determines if a rule exists whose port range ends at the specified number. The word 'to_' does *not* relate to inbound/outbound directionality; it relates to the port range ("counting _to_"). `to_port` is an exact criterion; so if the rule allows 1000-2000 and you specify a `to_port` of 1999, it does not match.
+  * security_group - Specifies a security-group id, to be checked as permissible origin (for `allow_in`) or destination (for `allow_out`) for traffic. Each AWS Security Group rule may have multiple allowed source or destination security groups.
```
clintoncwolfe, line 286: Hrm, I see a problem here.  Since a single rule can have multiple security groups as origins/destinations, but the criteria only allows you to specify a single sg_id, the user doesn't know if a multi-sg rule will match-on-any, or not match. And how does that behavior change for allow_{in,out}_only?

ipv4_range has the same problem, and solves it by allowing an array as the criterion value. If a single IP was provided in the query, we boost it to a single-element Array. If  we are in exact mode (the `_only` form), then the list of IPs on the rule must set-wise-match the list of IPs on the criterion: in other words, an `all?` match.  See line 176.

If we're not in exact mode, the IP matcher runs in `any?` mode (it also has to contend with CIDR subsetting, which `security_group` references do not).

So, I think we need to have `security_group` take an array, and clarify how that is handled.

lib/resources/aws/aws_security_group.rb
```diff
@@ -41,9 +41,18 @@ def allow_out_only?(criteria = {})
   private
 
   def allow_only(rules, criteria)
+    rules = allow__focus_on_position(rules, criteria)
     # allow_{in_out}_only require either a single-rule group, or you
     # to select a rule using position.
     return false unless rules.count == 1 || criteria.key?(:position)
+    if criteria.key?(:security_group)
```
clintoncwolfe, line 48: Just to repeat, I don't think security group requires any special handling at all here.

lib/resources/aws/aws_security_group.rb
```diff
@@ -41,9 +41,18 @@ def allow_out_only?(criteria = {})
   private
 
   def allow_only(rules, criteria)
+    rules = allow__focus_on_position(rules, criteria)
```
clintoncwolfe, line 44: I don't think this should be added.  `allow_only` says 'Exactly this rule must be present. If no position was given, it must be the only rule. If position was given, focus only on that rule, but allow no inexact (range- or set-based matching).'

So, this line enforces the conditional in the above statement:
```
return false unless rules.count == 1 || criteria.key?(:position)
```

Next, we turn on exact-matching mode; and we proceed into regular rule processing in `allow`. After validating args, `allow` calls `allow__focus_on_position(rules, criteria)`, after which there is exactly one rule to match (since we either started with only one rule, or a position was specified).

Each criteria gets a chance to fail the match; and 
```
 matched &&= allow__match_security_group(rule, criteria)
```
will do the job just fine. No need for special handling for any criteria in allow_only.

clintoncwolfe (changes requested):
Great work @j00p34! The example, docs, and test updates are great.

However, I do see two major issues.

1. I don't think the changes to `allow_only` are needed. Happy to help test that.
2. You mention that a single rule can reference multiple SGs.  We need to handle that more directly (IP ranges must do that, as well).  I can help if you like.  

Thanks!

docs/resources/aws_security_group.md.erb
j00p34: Well, I'm not sure. Maybe I should just remove that sentence. The thing is: You can't actually create a rule with multiple security groups as source/destination in the AWS Web interface. What happens is that if you create multiple identical rules which only differ on the security group they refer to, the API returns a struct containing one rule definition with multiple user_id_group_pairs in an array as targets/sources. I don't expect users to actually want to test for a list of security groups allowed for a single rule. This would be counterintuitive as it's not the way rules are created. Users will expect there to be separate rules while the API returns one rule with multiple group references, that's why I implemented it to work with single sg's only

lib/resources/aws/aws_security_group.rb
j00p34: Allow_only: I assumed that allow_only should check if there's a single rule allowing access to/from a single criteria. Because the similar security group rules are merged into a single rule, the rule containing multiple security group id's drops through the check for rules.count as that can be one while there are actually multiple security group rules. That's why I check the user_id_group_pairs.count. As a user I would expect a test for allow only on a security group criteria to allow only traffic to/from exact that single security group, which is what it checks for now.

position: ok so if I understand correctly, the position of the rule is not relevant for the security groups? I must admit that it's not completely clear to me what position indicates or what use this has. I assumed it specifies a position in the list of rules, (which is what it tests for obviously) but the position of the rule is only relevant if you know it upfront. So it could be used to check of a rule has changed position or something. If you would actually expect the rules to be in a particular position as created in the aws interface, that could possibly fail if you use security-groups with similar rules (because they are merged together). I just implemented it in the security group check because it was already there. 

I am perfectly willing to remove/change anything, but maybe you could make it a little more clear what exactly the reuirements for allow_only are and what position implies? If it's just that I'm not getting it that's also fine. Then I'll just remove the "superfluous" code and that's also fine with me. I will have to take some time trying to understand the code a little better then.

docs/resources/aws_security_group.md.erb
j00p34: I will have to polish my copy-paste skills

docs/resources/aws_security_group.md.erb
clintoncwolfe: Understood - that sounds like an edge case.  I'm happy with just documenting its current behavior.

lib/resources/aws/aws_security_group.rb
clintoncwolfe: Ahh, the key fact that "multiple rules that only target security groups get combined" breaks a lot of the assumptions behind the way that allow_only works.  

Indeed, so long as AWS isn't combining or rearranging rules on its own, position is a (potentially) useful way of isolating an individual rule, and then applying exact matching.  But since AWS is re-arranging things, something like what you have added is needed.  Makes me wonder if ipv4/6 range is also being collapsed / expanded.

So, I'm OK with the code, but I think some comments explaining why the `security_group` criterion requires special handling would be great.

lib/resources/aws/aws_security_group.rb
clintoncwolfe: (I'm now convinced it does)

clintoncwolfe:
DCO check fail is a false negative - signoff is present.

docs/resources/aws_security_group.md.erb
j00p34: correction: you can actually create them at once. Only they are expanded to multiple rules in the web interface and collapsed to a single rule when retrieving them via the API. But as they are aways displayed separately, I still think treating them as separate rules is the most intuitive approach. In my opinion this also accounts for ipv4 ranges. I think counting the rules+checking every single rule would be enough to ensure a security group allows only access as specified. (because AWS denies traffic by default). There could be a check for allow_only but that should check if there is only a single rule , or if there's a single rule that matches specific criteria.

 ```
 def allow_only(rules, criteria)
   rules = allow__focus_on_position(rules, criteria)
   return false unless count_sg_rules(rules) == 1
   allow(rules, criteria)
 end
```

lib/resources/aws/aws_security_group.rb
j00p34: >  Makes me wonder if ipv4/6 range is also being collapsed / expanded.

I run this through the debugger and it is indeed the case. When multiple similar rules are created they are collapsed into one ip_permission hash with multiple ip_ranges / ipv_6_ranges /   user_id_group_pairs

 >Indeed, so long as AWS isn't combining or rearranging rules on its own, position is a (potentially) useful way of isolating an individual rule, and then applying exact matching.

Could it be possible that the position check was initially based on this comment:  https://github.com/inspec/inspec/issues/2574#issuecomment-369775428 ?

Which I think must be a confusion between security groups, which don't allow 'deny' rules and don't depend on position:
see the second point here:
https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html#VPCSecurityGroups

and network ACL's that do allow 'deny' rules and also depend on position:
https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html

I think some kind of check for exact rules that searches all rules and sees if there's an exact match might be more appropriate.

j00p34: I have now tested this and the position matching actually fails when two similar rules (only different in target) are before the position.

clintoncwolfe: To clarify something about `position`, and the intent of `allow_*` in general: the matcher is not intended to emulate the evaluated behavior of AWS security groups, but only to examine the data structure that is returned by the `describe_security_groups` API call.  In that context, having a `position` element makes sense, as you might be interested in checking a rule located at a particular index.  

Having AWS expand / collapse rules breaks that badly, and that should likely be addressed but it's outside the scope of this PR.

clintoncwolfe: I've opened #3602 to address concerns about position; let's not let that block this PR.

clintoncwolfe (approved):

jquick (approved):
Looks good! Thanks for your hard work on this @j00p34

PR #3577, opened 2018-11-05
Add only_if to Inspec objects

james-stocks:
This lets us generate controls that have `only_if` clauses.

### DIFF
--- a/lib/inspec/objects/control.rb
+++ b/lib/inspec/objects/control.rb
@@ -2,7 +2,7 @@
 
 module Inspec
   class Control
-    attr_accessor :id, :title, :descriptions, :impact, :tests, :tags, :refs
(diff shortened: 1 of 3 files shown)


aaronlippold:
+1

aaronlippold:
I wonder if this begs the question that we should take a look at inspect
object and see if there any other missing elements i e things we can do in
the DSL but we can't replicate on the object side. We might have zero gaps.

Just a thought

clintoncwolfe (approved):
Thanks @james-stocks! My only feedback: consider also modifying `add_test` to accept an `only_if` clause (as an option hash, ideally, ahem), like on #3586.

jquick (approved):
Thanks @james-stocks

PR #3579, opened 2018-11-06
port: Correctly detect FreeBSD

### LINKED ISSUE #3563: `Port` resource not supported on FreeBSD, but helper code exists.

clintoncwolfe:
This PR adopts #3565 and was originally contributed by @dijit .

Fixes #3563 

This PR corrects the portability conditionals in the `port` resource for FreeBSD.  After the platform API changes, that code would never match; this change makes it match for anything in the BSD family (except Darwin/MacOS X, which matches earlier).

Manually tested; we have a FreeBSD config in our kitchen.vagrant.yml, but is broken and cannot converge (though it can get to the point where it is running, and SSH is listening)
screen shot 2018-11-05 at 19 07 34

### DIFF
--- a/lib/resources/port.rb
+++ b/lib/resources/port.rb
@@ -66,7 +66,7 @@ module Inspec::Resources
         LsofPorts.new(inspec)
       elsif os.windows?
         WindowsPorts.new(inspec)
-      elsif ['freebsd'].include?(os[:family])
+      elsif ['bsd'].include?(os[:family])
         FreeBsdPorts.new(inspec)
       elsif os.solaris?
         SolarisPorts.new(inspec)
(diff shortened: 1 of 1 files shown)


jquick (approved):
Thanks @clintoncwolfe

miah (approved):

jerryaldrichiii (approved):

PR #3583, opened 2018-11-06
Added xml resource support for ints, bools, and string responses

greenantdotcom:
Methods like…

* `count()` return `Integer` values
* `boolean()` return `TrueClass`/`FalseClass` values
* `concat()` return `String` values

…but threw exceptions because those types weren't supported.

This adds support to the `xml` resource, and adds tests to verify some of those examples.

Signed-off-by: Mark Hughes <greenantdotcom@users.noreply.github.com>


clintoncwolfe (approved):
Thanks so much - code, docs, and tests, all in one clear, useful feature - thanks for a great contribution!

jquick (approved):
Really nice, thanks @greenantdotcom

PR #3586, opened 2018-11-07
Allow add_test to accept negation

rachelrice:
This will allow for negated tests to be generated with add_tests.

Signed-off-by: Rachel Rice <rrice@chef.io>

### DIFF
--- a/lib/inspec/objects/describe.rb
+++ b/lib/inspec/objects/describe.rb
@@ -57,8 +57,8 @@ module Inspec
       @variables = []
     end
 
-    def add_test(its, matcher, expectation)
-      test = Inspec::Describe::Test.new(its, matcher, expectation, false)
+    def add_test(its, matcher, expectation, negated=false)
+      test = Inspec::Describe::Test.new(its, matcher, expectation, negated)
       tests.push(test)
       test
     end
--- a/lib/inspec/objects/describe.rb
+++ b/lib/inspec/objects/describe.rb
@@ -57,8 +57,8 @@ module Inspec
       @variables = []
     end
 
-    def add_test(its, matcher, expectation, negated = false)
-      test = Inspec::Describe::Test.new(its, matcher, expectation, negated)
+    def add_test(its, matcher, expectation, opts = {})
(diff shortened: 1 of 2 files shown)


lib/inspec/objects/describe.rb
```diff
@@ -57,8 +57,8 @@ def initialize
       @variables = []
     end
 
-    def add_test(its, matcher, expectation)
-      test = Inspec::Describe::Test.new(its, matcher, expectation, false)
+    def add_test(its, matcher, expectation, negated = false)
```
james-stocks: +1 on doing this.
I notice in the existing unit tests that it's hinted that you can use `obj.add_test(:aba, 'cmp', 'cba').negate!` for a negated test, and this change would not be needed.
I would personally not have guessed/assumed that the method returned the added test and I could chain a method call on it. 
We're also processing a large number of tests with a random mixture of negated ones. It would be nicer to have a single method call that we pass the boolean to instead of needing to write a branch that does two different calls to `add_test`

clintoncwolfe: My only concern here is that we have a method with three positional args, adding a fourth optional positional arg.  I'd prefer adding an optional options hash, to allow for future growth in a flexible way:

```
def add_test(its, matcher, expectation, opts = {})
  test = Inspec::Describe::Test.new(its, matcher, expectation, opts[:negated])
  tests.push(test)
  test
end
```

clintoncwolfe (approved):
Thanks for this - it will obviously help you in your work, and seems straightforward and has a good test.  My only complaint, and it's not enough to block acceptance, is that I would prefer hash-style parameter passing once the arity exceeds 3.

Thanks again, @rachelrice!

lib/inspec/objects/describe.rb
james-stocks: +1 on this approach; and @clintoncwolfe pointed out on https://github.com/inspec/inspec/pull/3577 that `only_if` is another property we want to manage, so we'd have at least two possible opts for a starter

james-stocks (approved):
Splendid

jquick (approved):
Nice job @rachelrice

PR #3603, opened 2018-11-09
Improvements to the functional helper run_inspec_process

clintoncwolfe:
This PR adds features I keep dragging around with me from branch to branch.  It only affects the functional test helper, so this is real "inside baseball" stuff.

It adds a `run_inspec_helper` method, similar to that available in the plugin test harness.  Its main difference is that it can take a hash of options, and then later commits build on that to allow running with a tmpdir created, generating a temporary plugin file, and automatically processing JSON ourput.

### DIFF
--- a/test/functional/helper.rb
+++ b/test/functional/helper.rb
@@ -94,22 +94,25 @@ module FunctionalHelper
     inspec(commandline, assemble_env_prefix(env))
   end
 
-  # This version allows additional options.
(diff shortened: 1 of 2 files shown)


jquick (approved):
Thanks @clintoncwolfe! Looks good.

miah (approved):
Great additions. Thanks @clintoncwolfe

PR #3606, opened 2018-11-13
filesystem: improve Windows support

mhackethal:
improve filesystem.rb to support windows.
Split into 2 classes LinuxFileSystemResource / WindowsFileSystemResource
Add filesystem to verify a FS-type ( currently not for linux because missing test server )
Size on Windows is converted to GB - discussion about this welcome

Signed-off-by: markus hackethal <mh@it31.de>

### DIFF
--- a/lib/resources/filesystem.rb
+++ b/lib/resources/filesystem.rb
@@ -5,7 +5,7 @@ module Inspec::Resources
     supports platform: 'windows'
     desc 'Use the filesystem InSpec resource to test file system'
     example "
-      'filesystem' currently not supported for Linux
+      
       describe filesystem('/') do
         its('size') { should be >= 32000 }
         its('type') { should eq false }
--- a/lib/resources/filesystem.rb
+++ b/lib/resources/filesystem.rb
@@ -56,9 +56,9 @@ module Inspec::Resources
     def name
       info = @fsman.info(@partition)
       info[:name]
-    end  
+    end
   end
-   
+
   class FsManagement
     attr_reader :inspec
     def initialize(inspec)
@@ -68,7 +68,7 @@ module Inspec::Resources
 
   class LinuxFileSystemResource < FsManagement
     def info(partition)
-      begin
+
         cmd = inspec.command("df #{partition} --output=size")
         raise Inspec::Exceptions::ResourceFailed, "Unable to get available space for partition #{partition}" if cmd.stdout.nil? || cmd.stdout.empty? || !cmd.exit_status.zero?
         value = cmd.stdout.gsub(/\dK-blocks[\r\n]/, '').strip
@@ -77,7 +77,7 @@ module Inspec::Resources
           size: value.to_i,
           filesystem: false,
         }
-      end
+
     end
   end
 
--- a/lib/resources/filesystem.rb
+++ b/lib/resources/filesystem.rb
@@ -20,7 +20,7 @@ module Inspec::Resources
     def initialize(partition)
       @partition = partition
       @cache = nil
-      #select file system manager
+      # select file system manager
       @fsman = nil
 
       os = inspec.os
@@ -33,7 +33,7 @@ module Inspec::Resources
       end
     end
 
-    def info 
+    def info
       return @cache if !@cache.nil?
       return {} if @fsman.nil?
       @fsman.info(@partition)
@@ -41,17 +41,18 @@ module Inspec::Resources
 
     def to_s
       "FileSystem #{@partition}"
-    end	
+    end
 
     def size
       info = @fsman.info(@partition)
       info[:size]
-    end		
+    end
 
     def filesystem
       info = @fsman.info(@partition)
       info[:filesystem]
     end
+
     def name
       info = @fsman.info(@partition)
       info[:name]
@@ -72,7 +73,7 @@ module Inspec::Resources
         raise Inspec::Exceptions::ResourceFailed, "Unable to get available space for partition #{partition}" if cmd.stdout.nil? || cmd.stdout.empty? || !cmd.exit_status.zero?
         value = cmd.stdout.gsub(/\dK-blocks[\r\n]/, '').strip
         {
-          name: partition,				
+          name: partition,
           size: value.to_i,
           filesystem: false,
         }
@@ -92,7 +93,7 @@ module Inspec::Resources
       begin
         fs = JSON.parse(cmd.stdout)
         rescue JSON::ParserError => e
-        raise Inspec::Exceptions::ResourceFailed,
+          raise Inspec::Exceptions::ResourceFailed,
           'Failed to parse JSON from Powershell. ' \
           "Error: #{e}"
         end
@@ -103,5 +104,5 @@ module Inspec::Resources
         filesystem: fs['FileSystem'],
       }
     end
-  end	
+  end
 end
--- a/lib/resources/filesystem.rb
+++ b/lib/resources/filesystem.rb
@@ -36,7 +36,7 @@ module Inspec::Resources
     def info
       return @cache if !@cache.nil?
       return {} if @fsman.nil?
-      @fsman.info(@partition)
+      @cache = @fsman.info(@partition)
     end
 
     def to_s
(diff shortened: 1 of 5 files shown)


jquick:
Looking good @mhackethal. You will need to reformat this Resource according to our Rubocop rules. You can test the file via `bundle exec rubocop filesystem.rb`.

mhackethal:
Thanks for the hint. I already recognized the "rubocop style". 
@jquick anyway to rerun it online? I dont have the required environment to perform it.

mhackethal:
@jquick Would you mind to assist me? I struggle currently with the unit testing. I adjusted the unit test for Filesystem but it didnt pass. https://travis-ci.org/inspec/inspec/jobs/454909571
Atm i am a little bit lost since i can not detect the root cause. A short hint from your side or a link to right "howto" would be very helpful so i can do it better next time.

Thanks 🥇

lib/resources/filesystem.rb
```diff
@@ -2,30 +2,104 @@ module Inspec::Resources
...
+        @fsman = WindowsFileSystemResource.new(inspec)
+      else
+        raise Inspec::Exceptions::ResourceSkipped, 'The `filesystem` resource is not supported on your OS yet.'
+      end
+    end
+
+    def info
+      return @cache if !@cache.nil?
+      return {} if @fsman.nil?
+      @fsman.info(@partition)
```
jquick: Do you want to set the cache here?
```
@cache = @fsman.info(@partition)
```

test/unit/resources/filesystem_test.rb
```diff
@@ -2,27 +2,23 @@
...
+  it 'verify filesystem on linux' do
+    resource = MockLoader.new(:arch).load_resource('filesystem','/') 
+    _(resource.size).must_be :>=, 1
+    _(resource.partition).must_equal '/'
+    _(resource.size).must_equal 28252316
+  end
+  
+  # arch windows
+  it 'verify filesystem on windows' do
+    resource = MockLoader.new(:arch).load_resource('filesystem','c:')
```
jquick: I think you want `:windows` here instead of `:arch`

jquick (commented):
Looking really nice @mhackethal!

jquick:
Looks like its still having issues. Ill pull it down this afternoon and play with it a bit and see what I can find. I think we may need to mock some of the commands that are used in the testing.

mhackethal:
Yes youre right. I am currently try to understand how "mock" is working. I found the folders
but i dont understand the strucuture and how to create a output file. At least i undestand that the test is not done against a real resource. Thats why unit testing fails. Since the windows cmd is "get-wmiobject" and a file mock/cmd/get-wmiobject already exists i am now totally lost 👎 

Sorry for the additional work on your side. I really gave it a try . Many thanks for your support.

jquick:
@mhackethal - I was able to get this working here https://github.com/mhackethal/inspec/pull/1

This mocks the long windows command you run and returns the cmd file accordingly. Since the command is long I used the sha of the command instead which is supported.

jquick:
You can also run your unit tests like this to check it:
```
((HEAD detached at mhackethal/patch-1))> bundle exec m test/unit/resources/filesystem_test.rb
Run options: -n "/^(test_0001_verify\\ filesystem\\ on\\ linux|test_0002_verify\\ filesystem\\ on\\ windows|test_0001_fails\\ on\\ FreeBSD\\ \\(unix\\-like\\))$/" --seed 21363

# Running:

.Command not mocked:
    $disk = Get-WmiObject Win32_LogicalDisk -Filter "DeviceID='c:'"
    $disk.Size = $disk.Size / 1GB
    $disk | select -property DeviceID,Size,FileSystem | ConvertTo-Json
    SHA: 2df93b941844efec1eee05c311cb4b4e8e3fd8b31808b785a70bd1a46298c716
E.

Finished in 0.402526s, 7.4529 runs/s, 7.4529 assertions/s.

  1) Error:
Inspec::Resources::FileSystemResource#test_0002_verify filesystem on windows:
Inspec::Exceptions::ResourceSkipped: Unable to get available space for partition c:
    /Users/jquick/Chef/inspec/lib/resources/filesystem.rb:90:in `info'
    /Users/jquick/Chef/inspec/lib/resources/filesystem.rb:47:in `size'
    /Users/jquick/Chef/inspec/test/unit/resources/filesystem_test.rb:15:in `block (2 levels) in <top (required)>'

3 runs, 3 assertions, 0 failures, 1 errors, 0 skips
```

This was how I easily got the SHA for the command.

mhackethal:
is it possible to get the SHA key when i edit over GitHub Website?
i was able to trace it down that i Need an Addition to /helper/resource,rb

mhackethal:
And again some Kind of failure 👎 

```
 1) Failure:
428
PluginInstallerInstallationTests#test_install_a_gem_with_invalid_depends_from_rubygems_org [C:/projects/inspec/test/unit/plugin/v2/installer_test.rb:221]:
429
Expected # encoding: UTF-8
430
"can't activate inspec-test-fixture-0.1.2, already activated inspec-test-fixture-0.1.0" to include # encoding: UTF-8
431
"Could not find 'fake_plugin_dependency' (>= 0)".

```
No idea what i am doing wrong.
Maybe you can support again @jquick

jquick:
Hey @mhackethal , That is not related to your changes. It is an issue we see sometimes in our testing framework and I need to nail down why it's happening. I think the last thing we need for you is to sign off your commit. You can do this by:

```
git commit --amend -s
git push -f
```

mhackethal (commented):
Signed-off-by: markus hackethal <mh@it31.de>

jquick:
Sorry @mhackethal, our policy is that it has to be part of the commit message if possible.

mhackethal:
yes i see was just test. Since i do everything over github website. I have to change a file to submit the sign-off. But i wanted to test if there is a different way. 
I will setup a Dev environment next days but atm no time

jquick:
Lookin good!

mhackethal:
some off-topic question would you mind to check #3607 too?
I get travis and appveyor error which are not related to my change. But can not solve it.

jquick (approved):
Thanks for all your work on this @mhackethal !

lib/resources/filesystem.rb
```diff
@@ -2,30 +2,104 @@ module Inspec::Resources
   class FileSystemResource < Inspec.resource(1)
     name 'filesystem'
     supports platform: 'linux'
+    supports platform: 'windows'
     desc 'Use the filesystem InSpec resource to test file system'
     example "
+      'filesystem' currently not supported for Linux
```
miah: Is this accurate? We say we support it above.

miah (changes requested):
Thank you for this contribution @mhackethal! I just have one minor question about the the example in the description.

lib/resources/filesystem.rb
mhackethal: hi @miah 
when you ask for i think its a bit missleading. The resource 'filesystem' works on Windows and Linux. But the resource itself has an Option .filesystem this is currently only available on Windows.

```
      describe filesystem('/') do 
        its('size') { should be >= 32000 } 
        its('filesystem') { should eq false }   <-- this is Always 'false' on linux 
      end 

      describe filesystem('c:') do 
        its('size') { should be >= 90 } 
        its('filesystem') { should eq 'NTFS' } <-- for Windows it return the "filesystem" for the resource
     end
```

I am fine if you want to update the documentation according to a better wording,

miah: Since we're adding this attribute could we name it something else? I'd go with `type` to make it match our other FS related resource [mount](https://www.inspec.io/docs/reference/resources/mount/).

```
describe filesystem('/') do
  its('type') { should eq 'NTFS' }
end
```

mhackethal: update to 'type' . Thanks for the good feedback

miah: Thanks! I think we can remove that 'filesystem is not supported' part of the description too.

mhackethal: removed.

miah (approved):
Perfect. Thanks for this contribution @mhackethal !!

PR #3613, opened 2018-11-15
Support cran packages for non-utf8 locales

atward:
{Initial_Punctuation} {Final_Punctuation} does not include ```'```, which causes cran resource to fail when outputting to a non-utf8 locale tty.


lamont-granquist (approved):

lamont-granquist:
@clintoncwolfe once i sat down and decoded the regexp and played in pry a bit this seems obviously correct.

clintoncwolfe (approved):
Thanks - our apologies for letting this sit for so long!

PR #3618, opened 2018-11-15
UI Object for for CLI work

clintoncwolfe:
While working on the plugin installer, and other CLI plugins, it became clear that we would benefit from having a UI object that enforced some consistency and made the code less UX+logic coupled.

This PR adds Inspec::UI, which provides low and high-level formatting.  It provides table formatting.

Two options control the object's behavior.

`--color/--no-color` has been promoted to be a global option.
`--interactive/--no-interactive` is a new global option.  

Extensive unit and functional tests are provided. A fixture plugin, `inspec-test-ui`, is used to exercise each of the major features.

Here's how some of the features render on my screen.  The table makes me super happy.

screen shot 2018-11-15 at 17 44 59

### DIFF
--- a/lib/inspec/ui.rb
+++ b/lib/inspec/ui.rb
@@ -1,5 +1,5 @@
 module Inspec
-  # Provides simple UI interaction primitives for CLI commands and plugins.
+  # Provides simple terminal UI interaction primitives for CLI commands and plugins.
   class UI
 
     ANSI_CODES = {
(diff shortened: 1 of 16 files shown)


miah (approved):
This is great. Thanks @clintoncwolfe !

lib/inspec/ui.rb
```diff
@@ -0,0 +1,216 @@
...
+        red: "\e[38;5;9m", # 256-color light red
+        green: "\e[38;5;41m",  # 256-color light green
+        yellow: "\e[33m",
+        cyan: "\e[36m",
+        white: "\e[37m",
+        grey: "\e[38;5;247m",  # 256-color medium grey
+      },
+    }.freeze
+
+    GLYPHS = {
```
jquick, line 20: I can see this being used in the future to clean up the `inspec/reporters/cli.rb`. If that is the case we will want to have a windows batch of valid symbols and colors here also. I think this is good for phase 1 though.

jquick (approved):
Really nice job! Some colors and symbols may not show up on windows as expected but it looks like most of the CLI output stuff is unchanged as of now.

PR #3560, opened 2018-11-01
Silence RSpec 'should' Warning

clintoncwolfe:
Fixes #952 

This PR adds a functional test to trigger a case in which RSpec prints a warning to STDERR, but worse, a summary of deprecations to STDOUT, which breaks JSON and other reporters.  The warning was an edge case, triggered by calling `.should` (note the dot) explicitly on any object, rather than as a DSL method. After adding the test, the PR adds configuration code to RSpec to inform it that we do intend to use `should`.

The InSpec project generally advises auditors to use the `should` syntax (as opposed to the RSpec-favored `expect` syntax) for readability.  You are of course free to use whatever syntax you please.

PR #3620, opened 2018-11-16
Revert setting RSpec expectation syntax to 'should' mode

### LINKED ISSUE #952: Rspec deprecation warnings added to test summary creating invalid json file
### Description

I'm running a Chef Inspec exec profile in a Jenkins job and it returns Rspec deprecation warnings. The problem is that some lines from these warnings are included in the json output that inspec exec command creates.  I'm left with a file that contains : 
1 deprecation warning total
{"version":"0.30.0","profiles":{  ..... }

 Inspec 0.30.0, Ruby 2.0.0p645, Rspec 3.5.2., RHEL 6.6
### Replication Case

Details and code snippets are here: 
http://stackoverflow.com/questions/39048601/chef-inspec-throws-rspec-deprecation-errors
Command executing inspec profile: 

```
  VmIp=$(vmrun getGuestIPAddress output-vmware-iso/${templateName}.vmx -wait)
    inspec exec  ./inspecRepo/ --format json-min -t ssh://vagrant@${VmIp} --password vagrant --sudo | tee Test_Results.json
```
### Stacktrace

[Test_Results.txt](https://github.com/chef/inspec/files/431156/Test_Results.txt)

### LINKED ISSUE #3616: "Expect" command missing from latest version
### 🎛 Description

This is a feature request. 🙋

In previous versions of inspec (3.0.25 and before) it was possible to run a test using the command `expect`. Eg:

`it 'this test would not fail' do
     expect(3).to eq(3)
end`

Versions newer than 3.0.25 no longer have this feature, so any test which uses `expect` now fails. Our work has a lot of dependencies on this command, and refactoring is out of the question if this can just be put back in.

This command works in version 3.0.25 and before. We use RHEL7.3.

### 🤔 Replication Case

To replicate the issue, just try to use the `expect` command for a test.

### 💁 Possible Solutions

Not sure.

### 💻 Stacktrace

`undefined method expect for #<RSpec::ExampleGroups::RhuiREGIONRhelServerRhsclRepo:0x00000003c83a40> Did you mean?  exec`

clintoncwolfe:
This PR reverts #3560, with the only change beyond the reversion being to add explanatory comments to a test.

#3560, included in inspec 3.0.46, explicitly set the RSpec expectation syntax to be `should` mode (as opposed to `expect` mode, or not setting it at all).  This fixed #952, but also broke `expect` syntax.  

While the InSpec project docs do not mention `expect` and generally prefer `should` for readability for non-programmers, RSpec has been moving to `expect` for a while, and many developers prefer it and have implemented their profiles using it. As a project, we don't want to insist on `should`; and besides, the circumstance in which #952 occurred are very rare and avoidable.

Hence the reversion.

Fixes #3616

### DIFF
--- a/lib/inspec/rspec_extensions.rb
+++ b/lib/inspec/rspec_extensions.rb
@@ -1,13 +1,5 @@
 require 'inspec/attribute_registry'
-require 'rspec/core'
 require 'rspec/core/example_group'
-
-# Setup RSpec to allow use of `should` syntax without warnings
-RSpec.configure do |config|
-  config.expect_with(:rspec) do |rspec_expectations_config|
-    rspec_expectations_config.syntax = :should
-  end
-end
 
 # This file allows you to add ExampleGroups to be used in rspec tests
 #
--- a/test/functional/filter_table_test.rb
+++ b/test/functional/filter_table_test.rb
@@ -69,7 +69,8 @@ describe '3103 default methods for filter table' do
     cmd += ' --controls ' + controls.join(' ')
     cmd = inspec(cmd)
 
-    # RSpec keeps issuing a deprecation count to stdout; I can't seem to disable it.
+    # RSpec keeps issuing a deprecation count to stdout
+    # See https://github.com/inspec/inspec/pull/3560
     output = cmd.stdout.split("\n").reject {|line| line =~ /deprecation/}.join("\n")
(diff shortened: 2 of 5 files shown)


jquick (approved):
Thanks @clintoncwolfe

PR #3623, opened 2018-11-20
www: remove jquery sticky on the sidebar

arlimus:
the sticky sidebar is really annoying when the window is too small as it moves offscreen or doesnt allow clicking entries at the  bottom


jquick (approved):
Thanks @arlimus

PR #3627, opened 2018-11-21
Remove deprecated shadow matchers

### LINKED ISSUE #3631: Remove deprecated `shadow` resource matchers
### 🎛 Description

🐛 bug report?

We have deprecation warnings for some the `shadow` resource matchers. We should remove them.

If we don't remove them before 4.0 we should change `warn` to `Inspec::Log.warn()` to allow hiding the deprecation warnings from the output (otherwise things like the `junit` reporter will be broken).

### 💁 Possible Solutions

Merging: https://github.com/inspec/inspec/pull/3627

Aha! Link: https://chef.aha.io/features/SH-2204

jerryaldrichiii:
This is a breaking change that we have had warnings about for a while. However, this will break some profiles so it should be merged with caution.

If we chose not to merge this, we should change `warn` to `Inspec::Log.warn()` so the DEPRECATION warnings can be hidden.

Fixes #3631

EDIT: (@clintoncwolfe) add link to issue

### DIFF
--- a/lib/resources/shadow.rb
+++ b/lib/resources/shadow.rb
@@ -23,11 +23,11 @@ module Inspec::Resources
          'the system and/or as users that own running processes.'
     example "
       describe shadow do
-        its('user') { should_not include 'forbidden_user' }
+        its('users') { should_not include 'forbidden_user' }
       end
 
       describe shadow.user('bin') do
-        its('password') { should cmp 'x' }
+        its('passwords') { should cmp 'x' }
         its('count') { should eq 1 }
       end
     "
@@ -53,12 +53,6 @@ module Inspec::Resources
       .register_column(:inactive_days, field: 'inactive_days')
       .register_column(:expiry_dates, field: 'expiry_date')
       .register_column(:reserved, field: 'reserved')
-    # These are deprecated, but we need to "alias" them
-    filtertable
-      .register_custom_property(:user) { |table, value| table.resource.user(value) }
-      .register_custom_property(:password) { |table, value| table.resource.password(value) }
-      .register_custom_property(:last_change) { |table, value| table.resource.last_change(value) }
-      .register_custom_property(:expiry_date) { |table, value| table.resource.expiry_date(value) }
 
     filtertable.register_custom_property(:content) { |t, _|
       t.entries.map do |e|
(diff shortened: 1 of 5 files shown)


jquick:
We should open an issue on this. We may just wait on this till 4.0 drops. If that is the case we will close and keep the branch alive.

jerryaldrichiii:
I opened an issue.  :smile:

clintoncwolfe:
Please rebase against master to resolve conflict.

lib/resources/shadow.rb
```diff
@@ -53,12 +53,6 @@ def initialize(path = '/etc/shadow', opts = {})
       .register_column(:inactive_days, field: 'inactive_days')
       .register_column(:expiry_dates, field: 'expiry_date')
       .register_column(:reserved, field: 'reserved')
-    # These are deprecated, but we need to "alias" them
-    filtertable
-      .register_custom_property(:user) { |table, value| table.resource.user(value) }
```
clintoncwolfe, line 58: See comment below

lib/resources/shadow.rb
```diff
@@ -88,38 +82,6 @@ def filter(query = {})
       Shadow.new(@path, content: content, filters: @filters + filters)
     end
 
-    # Next 4 are deprecated methods.  We define them here so we can emit a deprecation message.
-    # They are also defined on the Table, above.
-    def user(query = nil)
-      warn '[DEPRECATION] The shadow `user` property is deprecated and will be removed' \
```
clintoncwolfe, line 94: So, this was written prior to the `Inspec.deprecate` facility, so this is *currently* acting like a deprecation group set to `warn`.

If we simply remove these in v4, the user experience will be `no method 'user' for class CLASSx018374 etc...` which is pretty ugly.  Each control will fail, with a source code error.

Alternatively, we could add a deprecation group right now, set it action to warn (for v3) and then set to `exit` in v4, and physically remove only in v5.

That gives a better user experience - the user gets a message that clearly tells them what to do (because we write a nice message) and we also ensure that they actually make the change because the inspec run exits abnormally.

clintoncwolfe (changes requested):
I think we should probably use the deprecation facility for this, so users will have a better experience in v4.

lib/resources/shadow.rb
clintoncwolfe: ... and in fact that is done on https://github.com/inspec/inspec/pull/3829/files#diff-dea4fdbc73cceaa1ff8488720ce40d14 .

I think we should just close this PR.

clintoncwolfe:
Closing in favor of #3829 and #3838

PR #3628, opened 2018-11-23
add support for aws kms key aliases

rikturnbull:
Signed-off-by: Rik Turnbull <37418542+rikturnbull@users.noreply.github.com>

Adds support for kms key aliases:

  describe aws_kms_key('alias/my-kms-key') do
    it { should exist }
  end

### DIFF
--- a/docs/resources/aws_kms_key.md.erb
+++ b/docs/resources/aws_kms_key.md.erb
@@ -10,7 +10,7 @@ Use aws_kms_key to verify the properties of a single key. Use aws_kms_keys to ve
 
 AWS Key Management Service (AWS KMS) is a managed service that makes it easy for you to create and control the encryption keys used to encrypt your data. AWS KMS lets you create master keys that can never be exported from the service and which can be used to encrypt and decrypt data based on policies you define.
 
-Each AWS KMS Key is uniquely identified by its key_id or arn.
+Each AWS KMS Key is uniquely identified by its key_id, arn or alias.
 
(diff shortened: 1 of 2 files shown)


clintoncwolfe (changes requested):
Hi! Thanks so much for the contribution! Could you add some tests - at least at the unit level?  We also like to see integration tests, using Terraform to setup and then run InSpec controls using your new feature.

rikturnbull:
Now I'm wondering whether a kms key alias should be a first class resource in itself - rather than as a unique identifier for a kms key. If you're happy as it stands I'll go ahead and add some unit tests.

I can have a go at integration tests - is it possible to isolate individual tests?

clintoncwolfe:
Generally, we try to split things apart at the same resolution as the AWS API does.  So, if they have a separate data type for an Alias, it might make sense to make a separate InSpec resource.  The main question is, aside from the name and the referred-to key, will an Alias ever bear additional information (such as creator, permissions, creation date, etc?)

For isolating tests, unfortunately it's rather primitive.  Each terraform file tries to isolate a service as best it can.  You can run individual control files instead for the entire test suite, as well.

clintoncwolfe:
Hi! We're moving future development on the AWS resources into the inspec-aws resource pack. There is an open issue for this feature there, at https://github.com/inspec/inspec-aws/issues/13

I'm closing this here, as we are no longer accepting new AWS features in inspec/inspec.

PR #3630, opened 2018-11-26
Add support for aws vpc names

rikturnbull:
Allows AWS VPCs to be identified by Name label:

    # Find a VPC by Name
    describe aws_vpc(vpc_name: 'my-vpc') do
      it { should exist }
    end

### DIFF
--- a/docs/resources/aws_vpc.md.erb
+++ b/docs/resources/aws_vpc.md.erb
@@ -29,7 +29,7 @@ This resource first became available in v2.0.16 of InSpec.
 
 ## Syntax
 
-An `aws_vpc` resource block identifies a VPC by id. If no VPC ID is provided, the default VPC is used.
+An `aws_vpc` resource block identifies a VPC by id or Name tag. If no VPC ID or Name tag is provided, the default VPC is used.
 
     # Find the default VPC
     describe aws_vpc do
@@ -46,6 +46,10 @@ An `aws_vpc` resource block identifies a VPC by id. If no VPC ID is provided, th
       it { should exist }
     end
 
+    # Find a VPC by Name
+    describe aws_vpc(vpc_name: 'my-vpc') do
+      it { should exist }
+    end
 <br>
 
 ## Examples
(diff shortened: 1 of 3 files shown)


rikturnbull:
I had not considered multiple VPCs could have the same Name tag.

What's the recommended pattern for dealing with this - fail if multiple VPCs match?

aaronlippold:
If you have only one, return a string, if you have more than one, return an array of strings of the names. We will also have to add docs to include this and *if the name* doesn't produce a unique VPC then we have to at a minimum document this or find a way to warn the user they are not getting only a single VPC. Lastly we will / should provide a way ( via filter tables perhaps ) to allow them to use the 'name' and some other identifying element in a `.where` clause to ensure they only get one if that is what is needed etc.

clintoncwolfe:
As submitted, this is a *singular* resource.  Standard behavior for singular AWS resources when its resource parameters match multiple resources is to throw an error - see [aws_security_group](https://www.inspec.io/docs/reference/resources/aws_security_group/#syntax) for example.  

You could certainly add the same feature (query by name) to `aws_vpcs`, the plural resource - and in that case, if you match multiple, that's just fine.

rikturnbull:
I see. I'm trying to write tests before the resources are built which means I have to select them based on naming. I can see a potential solution if I add `:tags` to the `aws_vpc` resource:

`.register_column(:tags, field: :tags)`

and then (there's probably a better way to do this):

```
  # check 'my-vpc' exists and is unique
  describe aws_vpcs.where{ !tags.nil? and tags.find{|h| h[:key] == "Name"}[:value] == "my-vpc"} do
    its('count') { should cmp 1 }
  end

  # check 'my-vpc' is available
  aws_vpcs.where{ !tags.nil? and tags.find{|h| h[:key] == "Name"}[:value] == "my-vpc"}.vpc_ids.each do |vpc_id|
    describe aws_vpc(vpc_id) do
      its ('state') { should eq 'available' }
    end
  end

```
Is this correct?

rikturnbull:
I'll close this. I've raised a new one instead: https://github.com/inspec/inspec/pull/3672

rikturnbull:
Going round in circles - leaving this open as a singular resource, now with unit tests.
See: https://github.com/inspec/inspec/pull/3672#issuecomment-450558443

clintoncwolfe:
Closing in favor of https://github.com/inspec/inspec-aws/issues/25

PR #3632, opened 2018-11-26
Add SQLcl to Oracledb_session Doc

ibsavage:
SQLcl can be used with the oracledb_session resource. It requires a slightly different option compared to sqlplus that wasn't previously noted in the documentation.


jquick (approved):
Thanks @ibsavage !

PR #3633, opened 2018-11-26
Modify URL fetcher to accept URI

### LINKED ISSUE #3626: NoMethodError when downloading profile from automate server
### 🎛 Description

🐛 bug report

There appears to be a regression in Inspec 3.0.46 that is breaks the Audit cookbooks ability to detect when an Inspec profile is not available from the Automate server.  Which then prevents it from displaying a user friendly error about the profile is missing on the Automate server. 

For example when using Audit 7.3.0 and Inspect-core 2.2.102 you will normally see a `RuntimeError: Unable to find requested profile on path: '/compliance/organizations/support/owners/admin/compliance/badprofile/tar' on the Automate system.` when selecting an invalid profile.  

On the same system when using Inspec-core 3.0.46 a `NoMethodError: undefined method '[]'` exception is raised in Inspect instead.

Failing system: Audit 7.3.0, Inspec 3.0.46

### 🤔 Replication Case

Setup chef-client run with Audit cookbook and the following attributes

```
    "audit": {
      "profiles": {
        "badprofile": { "compliance": "admin/badprofile" }
      },
      "inspec_version": "3.0.46"
    }
```

When chef-client runs it will then throw 

### 💁 Possible Solutions

In https://github.com/inspec/inspec/blob/master/lib/fetchers/url.rb#L126 it appears that `target` is already a URI object and should just be returned instead of being parsed. 

Updating it to the following fixes the problem and lets it throw the `profile not found error` as expected.

```
    def parse_uri(target)
      return URI.parse(target) if target.is_a?(String)
      return target if target.is_a?(URI)

      URI.parse(target[:url])
    end
```

### 💻 Stacktrace

Gist with stacktraces from both versions of inspec
https://gist.github.com/teknofire/786bad65ed9c3741d6bbc9c0e2430913

jerryaldrichiii:
When https://github.com/inspec/inspec/pull/3562 was merged it added a method to parse a URI passed as a URL. However, some users (at least the Audit cookbook) were passing a URI and not a URL so there is no need to parse it.

This adds code to use the URI if a URI was passed.

Many thanks to @teknofire for discovering this and submitting a very detailed issue.

This closes #3626

### DIFF
--- a/lib/fetchers/url.rb
+++ b/lib/fetchers/url.rb
@@ -24,6 +24,8 @@ module Fetchers
         resolve_from_string(target[:url], opts, target[:username], target[:password])
       elsif target.is_a?(String)
         resolve_from_string(target, opts)
+      elsif target.is_a?(URI)
+        resolve_from_string(target.to_s, opts)
       end
     end
 
@@ -94,7 +96,7 @@ module Fetchers
     attr_reader :files, :archive_path
 
     def initialize(url, opts)
-      @target = url
+      @target = url.to_s
       @target_uri = url.is_a?(URI) ? url : parse_uri(url)
       @insecure = opts['insecure']
       @token = opts['token']
--- a/lib/fetchers/url.rb
+++ b/lib/fetchers/url.rb
@@ -95,7 +95,7 @@ module Fetchers
 
     def initialize(url, opts)
       @target = url
-      @target_uri = parse_uri(@target)
+      @target_uri = url.is_a?(URI) ? url : parse_uri(url)
       @insecure = opts['insecure']
       @token = opts['token']
       @config = opts
--- a/test/unit/fetchers/url_test.rb
+++ b/test/unit/fetchers/url_test.rb
@@ -34,7 +34,14 @@ describe Fetchers::Url do
       res.expects(:open).returns(mock_open)
       _(res).must_be_kind_of Fetchers::Url
       _(res.resolved_source).must_equal({url: 'https://chef.io/some.tar.gz', sha256: expected_shasum})
+    end
 
+    it 'handles an https URI' do
+      uri = URI.parse('https://chef.io/some.tar.gz')
+      res = Fetchers::Url.resolve(uri)
+      res.expects(:open).returns(mock_open)
+      _(res).must_be_kind_of Fetchers::Url
+      _(res.resolved_source).must_equal({url: 'https://chef.io/some.tar.gz', sha256: expected_shasum})
     end
 
     it 'doesnt handle other schemas' do


teknofire (approved):

clintoncwolfe (dismissed):
Thanks @jerryaldrichiii! 

However, we can't accept this without tests.  It looks like the URL fetcher has some unit tests [already](https://github.com/inspec/inspec/blob/master/test/unit/fetchers/url_test.rb); you should able to just add some there, passing a URI object instead of a string.

jerryaldrichiii:
Ah, that is 100% on me. Sorry.

Adding tests now...and looking at it from that perspective I think I'll change a few things as well (e.g. add support to `#resolve` just in case)

jquick (approved):
Thanks @jerryaldrichiii !

lib/fetchers/url.rb
```diff
@@ -24,6 +24,8 @@ def self.resolve(target, opts = {})
         resolve_from_string(target[:url], opts, target[:username], target[:password])
       elsif target.is_a?(String)
         resolve_from_string(target, opts)
+      elsif target.is_a?(URI)
+        resolve_from_string(target.to_s, opts)
```
clintoncwolfe, line 28: Hrm.  If I'm reading this right, the user passes a URI object, we render it to a String on line 28, call resolve_from_string, which parses the string into a URI object, on line 31?  That makes us rely on the round-tripping ability of URI, which seems risky.

Perhaps we should have a resolve_from_uri, which resolve_from_string calls?

test/unit/fetchers/url_test.rb
```diff
@@ -34,7 +34,14 @@
       res.expects(:open).returns(mock_open)
       _(res).must_be_kind_of Fetchers::Url
       _(res.resolved_source).must_equal({url: 'https://chef.io/some.tar.gz', sha256: expected_shasum})
+    end
 
+    it 'handles an https URI' do
```
clintoncwolfe, line 39: I think this is a great start for a basic test!  However, I think if someone is handing you a URI, it may be because it is something more exotic than a HTTPS address. Consider adding a test with something harder to parse - perhaps a URL with an embedded password, or an FTP URI, or a file:/// URI.

clintoncwolfe (dismissed):
Thanks @jerryaldrichiii ! I'd like to see an implementation that doesn't rely on re-parsing the URI, and maybe more than basic testing.  Thanks!

clintoncwolfe (approved):
Thanks for the effective fix, @jerryaldrichiii!

PR #3644, opened 2018-12-01
Fix GitHub alert for rack dependency

jerryaldrichiii:
After digging I see that PR #975 fixed `bundle install` on 2.1.9 about two years ago. The minimum required version of Ruby for InSpec is 2.2. This updates our Gemfile to remove any logic for Ruby `< 2.2.2`.

The Gemfile should only be used for development...but I could be wrong here. Let me know if you think there will be any issues.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,11 +2,6 @@
 source 'https://rubygems.org'
 gemspec name: 'inspec'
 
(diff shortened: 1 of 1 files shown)


jquick (approved):
Thanks @jerryaldrichiii !

miah (approved):
Thanks @jerryaldrichiii

PR #3645, opened 2018-12-02
Remove Habitat package run time dependencies from Habitat plan

jerryaldrichiii:
This removes the runtime dependencies on Hab pkgs and instead modifies the `PATH` environment variable to use the OS binaries where the InSpec Habitat package is installed.

It should be noted that this is counter to what Habitat intends in most cases. In general, it is preferable to use only Habitat packages as runtime dependencies to get all the benefits that Habitat provides.

We elected not to do this for the InSpec Habitat package since the list of binaries that would need to be installed to support all InSpec resources would be prohibitively expensive (both in disk space and network requirements). If you wish to use Habitat packaged binaries with this package you can use:

`hab pkg install origin/my-binary --binlink`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,7 @@ Currently, this method of installation only supports Linux. See the [Habitat sit
 Download the `hab` binary from the [Habitat](https://www.habitat.sh/docs/get-habitat/) site.
 
 ```bash
-hab pkg install chef/inspec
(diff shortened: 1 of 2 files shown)


jquick (approved):
Love the reduction in dependencies! Thanks @jerryaldrichiii

miah (approved):
Thanks @jerryaldrichiii

PR #3647, opened 2018-12-03
Further slim the files we ship in the gem artifact

tas50:
The artifact doesn't need the gemspec, gemfile or rakefile since we're not shipping tests already. Also there's no need for the changelog or maintainer file. For those a user should reference the github repo since these files are buried deep on the filesystem and not designed to be read through.


jerryaldrichiii (approved):

jquick (approved):
Thanks @tas50

PR #3656, opened 2018-12-07
Update formatting of heading on doc for aws_config_delivery_channel

tmonk42:
Heading is displaying incorrect on https://www.inspec.io/docs/reference/resources/aws_config_delivery_channel/


clintoncwolfe:
Thanks! Accepting without DCO as trivial.

PR #3660, opened 2018-12-10
Incorrect mysql session failure when rows are returned starting with the string 'error'

### LINKED ISSUE #3659: mysql_session: Failure when a query returns a row that begins with the string 'error'
This happens for example if you use SHOW VARIABLES which has a variable named error_count.
```
sql = mysql_session('my_user','password')

describe sql.query('SHOW VARIABLES') do
  its('stdout') { should_not match(/test/) }
end
```

The warning that will appear:
> Can't connect to MySQL instance for SQL checks.

It's because of the last part of this regex, it checks for lines that start with error

https://github.com/inspec/inspec/blob/v3.0.64/lib/resources/mysql_session.rb#L33

```
 if out =~ /Can't connect to .* MySQL server/ || out.downcase =~ /^error/
```

A real error seems to always start with the text ERROR and is followed by a space and a number
```
ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'VARIABLEsS' at line 1
```

A row/result always followed by a tab character and the value
```
error_count     0
```

Adding a simple space to the regex will fix it, i'll make a pull request for it :-)

wiebe:
See #3659 for further details.

Fixes #3659 	

EDIT: (@clintoncwolfe) move issue reference into PR body

### DIFF
--- a/lib/resources/mysql_session.rb
+++ b/lib/resources/mysql_session.rb
@@ -30,7 +30,7 @@ module Inspec::Resources
       mysql_cmd = create_mysql_cmd(q, db)
       cmd = inspec.command(mysql_cmd)
       out = cmd.stdout + "\n" + cmd.stderr
-      if out =~ /Can't connect to .* MySQL server/ || out.downcase =~ /^error/
+      if out =~ /Can't connect to .* MySQL server/ || out.downcase =~ /^error /
         # skip this test if the server can't run the query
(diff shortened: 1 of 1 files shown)


clintoncwolfe (approved):
Simple change, with big impact.  Thanks so much @wiebe!

jquick (approved):
Thanks @wiebe !

PR #3663, opened 2018-12-11
Remove extra files from inspec-core + file appbundling inspec

tas50:
InSpec could no longer be appbundled since it was missing a gemfile and
gemspec. I added those back. I also trimmed down the gems in the
inspec-core gem.


tas50:
The sooner this could get released the better. I *may* have broken the DK / Workstation builds with the previous cleanup gone too far.

jquick (approved):
Thanks @tas50 !

clintoncwolfe (approved):
Thanks!

Issue #3661, opened 2018-12-10
Formalize InSpec config file handling


clintoncwolfe:
### 🎛 Description

🙋 feature request

Currently, the `--json-config` option takes a path to a file (defaults to `~/.inspec/config.json`).  That config file currently is treated as follows:

 * We read it.  There is no validation or file format version or schema.
 * We merge it into the hash of CLI options that Thor parsed.  This makes it very powerful!

However, some things have just been glued in - like compliance server credentials.  There is a need to store other things in the config file - one example might be credential sets for Habitat (which like Compliance has a multipart, hard-to-URLify credential set).

Ideally, we'd be able to use that config file for things that don't exactly map to the CLI args, yet also doesn't feel glued in.  We'd also like to be able to validate the file, and reason about file format versions without heuristics.

Finally, the name of the option '--json-config' ties us to the format, in a rather strange way.  In the future, it would be nice to be able to read other formats, like YAML. Whether that comes to pass or not, `--config` is shorter and less unexpected.  Let's add support for `--config`, document it, and still support `--json-config` but plan to deprecate it.

### 🌍 InSpec and Platform Version

3.0.64

### 💁 Possible Solutions

Create a class to represent the file.  Have the class able to run a jsonschema-based validator, and store the schemae based on file version, which we detect from the first key, "file_version":"1.1".  If not present, assume v0.0.

For v1, consider this format:

```
{
  "file_format": "1.0", 
  "cli_args": {
    "log-level": "warn"
  },
  "credentials": {
    "habitat": {
      "my-creds-01": {
        "ssh_host": "somehost.somewhere",
        "ssh_key": "/some/path/to.key",
        "cli_path": "/opt/habitat/embedded/bin/hab",
        "api_key": "1234567890",
        "api_host": "anotherhost.somewhere"
      },
      "my-creds-02": {
        ...
      }
    },
    "compliance": {
      "default": {
         ... used for reporting, not scan targeting ....
      }
    }
  }
}
```

The point here being that we could store credentials first by connection name (URI schema / train plugin suffix), then keyed by a user-defined name, so they could have
multiple credential sets per backend.

A user might use it as:
```
$ inspec exec someprofile -t habitat://my-creds-01
$ inspec exec someprofile -t habitat://my-creds-02
```

For backends that provide their own credential storage facility, we would need to decide on the appropriate priority.  For example for AWS, 

1. use environment variables; if missing,
2. use instance role credentials; if missing,
[...]

PR #3669, opened 2018-12-12
Expose additional WinRM options for transport, basic auth, and SSPI 

### LINKED ISSUE #391: Expose additional winrm options
### Description

On FIPS enabled clients the WinRM transport can only use `:ssl` **transport** and `basic_auth`. NTLM auth uses `MD4` is is not at all feasible.

 ### Train and Platform Version

All verisons

### Replication Case

Ruby just barks when using `MD5` in FIPS mode. `MD4` is not at all available. 

The only working config on FIPS machine:

```
require 'winrm'

opts =
    {
    transport: :ssl,
    endpoint: 'https://<some server>:5986/wsman',
    user: 'Administrator',
    password: '<redacted>,
    no_ssl_peer_verification: true,
    basic_auth_only: true
    }

conn = WinRM::Connection.new(opts)

shell = conn.shell(:powershell)
puts shell.run('ls').stdout
shell.close
```

frezbo:

Fixes: https://github.com/inspec/train/issues/391

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -69,11 +69,11 @@ module Inspec
         desc: 'Use SSL for transport layer encryption (WinRM).'
       option :self_signed, type: :boolean,
         desc: 'Allow remote scans with self-signed certificates (WinRM).'
-      option :transport, type: :string,
+      option :winrm_transport, type: :string, default: 'negotiate',
         desc: 'Specify which transport to use, defaults to negotiate (WinRM).'
-      option :disable_sspi, type: :boolean,
+      option :winrm_disable_sspi, type: :boolean,
         desc: 'Whether to use disable sspi authentication, defaults to false (WinRM).'
-      option :basic_auth, type: :boolean,
+      option :winrm_basic_auth, type: :boolean,
         desc: 'Whether to use basic authentication, defaults to false (WinRM).'
       option :json_config, type: :string,
         desc: 'Read configuration from JSON file (`-` reads from stdin).'
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -69,6 +69,12 @@ module Inspec
         desc: 'Use SSL for transport layer encryption (WinRM).'
       option :self_signed, type: :boolean,
         desc: 'Allow remote scans with self-signed certificates (WinRM).'
+      option :transport, type: :string,
+        desc: 'Specify which transport to use, defaults to negotiate (WinRM).'
+      option :disable_sspi, type: :boolean,
+        desc: 'Whether to use disable sspi authentication, defaults to false (WinRM).'
+      option :basic_auth, type: :boolean,
+        desc: 'Whether to use basic authentication, defaults to fallse (WinRM).'
       option :json_config, type: :string,
         desc: 'Read configuration from JSON file (`-` reads from stdin).'
       option :proxy_command, type: :string,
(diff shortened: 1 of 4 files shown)


clintoncwolfe:
I'm all for making it easier to use transports in the real world, with the settings needed to do so.  However, I think we need to think carefully about how we go about this: we already have a proliferation of options.

#3661 proposes using the configuration file to store credential sets - allowing transports to have extensive options and arguments.  With #3661, you could have multiple credentials for different winrm targets, with each configured as it needs.  Adding new options to that would be a non-event; InSpec would simply pass all information along to Train.

However, that's not convenient for more casual use.  Command line args are really what's wanted.  We face a few challenges:

1. Organizing CLI opts effectively
1. Allowing Train plugins to dynamically configure InSpec's CLI option parser (Thor) 

The second point is a long-term problem - winrm is part of core Train for today, and we can bear some overtight coupling on that.

For the first issue, I propose that we prefix all winrm options with `--winrm-` - so this PR would add `--winrm-transport`, `--winrm-disable-sspi`, and `--winrm-basic-auth`.  We'd then also alias existing WinRM options to include that prefix, and at some (major version) point deprecate the "prefix-less" versions of the options.  We could also use the Thor `group` facility to improve help output. 

Obviously, a similar approach could be taken for other transports that expose options (most egregiously SSH).

frezbo:
@clintoncwolfe Would just prefixing the three options I added with `winrm` would be enough to get this going forward? Do you want me to update the **train** PR too, https://github.com/inspec/train/pull/392?

clintoncwolfe:
Yep, that'll do for this PR.  I'll put in a review.

lib/inspec/base_cli.rb
```diff
@@ -69,6 +69,12 @@ def self.target_options # rubocop:disable MethodLength
         desc: 'Use SSL for transport layer encryption (WinRM).'
       option :self_signed, type: :boolean,
         desc: 'Allow remote scans with self-signed certificates (WinRM).'
+      option :transport, type: :string,
```
clintoncwolfe: As discussed, please prefix these with `--winrm-`

clintoncwolfe (changes requested):
As discussed, just looking for the prefix change; we'll also need to change the version pin on Train in the gemspec.

lib/inspec/base_cli.rb
```diff
@@ -69,6 +69,12 @@ def self.target_options # rubocop:disable MethodLength
...
+        desc: 'Specify which transport to use, defaults to negotiate (WinRM).'
```
clintoncwolfe, line 73: I suggest we add a list of valid transport names.  Also, the help text says it defaults to `:negotiate`, but the Thor option doesn't have a default (perhaps omission causes the WinRM library to negotiate - but if so, a comment would be good)

clintoncwolfe (changes requested):

frezbo:
@clintoncwolfe PR updated, I will update the **gemspec** once a new version of train is released.

lib/inspec/base_cli.rb
frezbo: resolved.

lib/inspec/base_cli.rb
frezbo: resolved.

clintoncwolfe:
Train 1.6.3 has been released, with the new options included.

frezbo:
@clintoncwolfe The tests fail due to the newer ruby versions packaging `bundler` along with it. I have also seen this locally. The `functional` tests pass locally.

clintoncwolfe (approved):
Thanks so much, @frezbo !

jerryaldrichiii (approved):
Great work @frezbo!

frezbo:
👍

PR #3671, opened 2018-12-14
habitat packaging: Fix inspec wrapper linking to /bin/bash instead of the bash hab package

teknofire:
Adds missing `pkg_deps` for `core/bash`, otherwise
https://github.com/inspec/inspec/blob/master/habitat/plan.sh#L57 would
return an empty string for the pkg path and output `/bin/bash`
instead of the expected `/hab/pkgs/core/bash/.../bin/bash` path.

@jerryaldrichiii This looks like it might have been broken for some time but a recent update for #3645 caused this to fail on me

Signed-off-by: Will Fisher <wfisher@chef.io>

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -55,7 +55,7 @@ wrap_inspec_bin() {
   local real_bin="$GEM_HOME/gems/inspec-${pkg_version}/bin/inspec"
   build_line "Adding wrapper $bin to $real_bin"
   cat <<EOF > "$bin"
-#!$(pkg_path_for bash)/bin/bash
+#!$(pkg_path_for core/bash)/bin/bash
 export SSL_CERT_FILE=$(pkg_path_for cacerts)/ssl/cert.pem
 set -e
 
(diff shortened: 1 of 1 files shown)


jerryaldrichiii (changes requested):
Great catch @teknofire!

We did something slightly unconventional and removed hab dependencies and instead modified InSpec to use system packages (see https://github.com/inspec/inspec/blob/master/habitat/plan.sh#L62). That being said, ensuring we have an interpreter to read our bin wrapper is crucial. I mistakenly thought this was provided by `core/coreutils`.

Care to change the line below to `pkg_path_for core/bash` instead of just `pkg_path_for bash`?

https://github.com/inspec/inspec/blob/master/habitat/plan.sh#L57

teknofire:
@jerryaldrichiii updated with the requested change

jerryaldrichiii (approved):
Great catch @teknofire!

clintoncwolfe (approved):
Thanks!

Issue #3641, opened 2018-11-30
Attempt awspec compatibility


clintoncwolfe:
### 🎛 Description

🙋 feature request

Let's try to make a bridge to the awspec kit, so InSpec can leverage the resources.

skpaterson:
Hi @clintoncwolfe - are you still targeting this work for the InSpec 4.0 release?

clintoncwolfe:
No, we're not.  I did some spike work on it in https://github.com/inspec/inspec/tree/cw/awspec - I got to the point where it was clear we'd have to alter the inheritance tree (both awspec resources and inspec resource demand to dictate the superclass).

Not impossible, and the payoff is large!

skpaterson:
Thanks @clintoncwolfe will migrate this across in that case.

Issue #3621, opened 2018-11-16
Plural-Singular Iteration 


clintoncwolfe:
### 🎛 Description

🙋 feature request

I've been thinking about this for a while, as a fluency issue.

Currently we implement plural and singular resources separately, unaware of each other.  [We say](https://blog.chef.io/2018/06/14/understanding-singular-and-plural-inspec-resources/) plural resources are good for finding things, and singular are good for examining in detail.  That leads to filter-and-examine patterns like this:

```
control "No group should allow in port 22" do
  aws_security_groups.where(vpc_: 'vpc-123').security_group_ids.each do |security_group_id|
    describe aws_security_group(security_group_id) do
      it { should_not allow_in port: 22 }
    end
  end
end
```

(Just using AWS Security Group as an example - this would apply to any plural-singular pair).

It would be great if the plural resource knew about the singular resource, and was able to be smarter, maybe this:

```
control "No group should allow in port 22" do
  aws_security_groups.where(vpc_: 'vpc-123').each do |security_group|
    # This is an actual aws_security_group object, not just the ID
    describe security_group do
      it { should_not allow_in port: 22 }
    end
  end
end
```

### 🌍 InSpec and Platform Version

3.0.46

### 💁 Possible Solutions

* Use a ResourceDSL plugin to add linkage DSL methods to the plural resources (so that a plural resource can learn the associated singular resource, and the ID value(s) to pass) 
* Add a hidden constructor that allows a plural resource to inflate a singular resource using data it may have already fetched

Issue #3139, opened 2018-06-13
Create a common interface for AWS tag queries


clintoncwolfe:
### 🎛 Description

🙋 feature request

This issue supercedes #2592, which had a lot of other discussion.  Refs #3138 

Most AWS resources support tagging - attaching a (sometimes limited) number of key-value pairs.  We'd like to expose this to users in several ways.

Unfortunately, the API for fetching tags varies widely by services, so each resource will need to implement the actual fetch, while a shared library might be able to handle the properties / matchers /filter criteria definitions.

Thoughts on desired behavior:

#### Singular resources

* `tags` property returning a hash

Already implemented by aws_ec2_instance, PR for `aws_s3_bucket` on #3138 

* `has_tag?(tag_name)` matcher

* `has_tag_value?(tag_name, tag_value)` matcher

#### Plural resources

* `tags` merged hash property. Are values boosted to arrays?

* has_tag(tag_name) filter_criteria

* has_tag_value(tag_name, tag_value) block-only filter_criteria

#### Dynamic properties?

Azure also defines custom properties, based on the observed tags: `name_tag`, `env_tag`, etc.  Is that desirable?  How do you handle absence?  Is that possible to implement as filter criteria on a plural?

#### Case sensitivity?

AWS tags are sometimes titlecased (Name is, traditionally).

### 🌍 InSpec and Platform Version

2.2.10

clintoncwolfe:
Refs:

#3138 	
#3415 
#3416 
#3417 
#3672

PR #3672, opened 2018-12-16
Add support for tags to aws_vpcs

rikturnbull:
Adds tag property to aws_vpcs to allow tag based queries:

```
  # check 'my-vpc' exists and is unique
  describe aws_vpcs.where{ !tags.nil? and tags.find{|h| h[:key] == 'Name'}[:value] == 'my-vpc'} do
    its('count') { should cmp 1 }
  end
```

### DIFF
--- a/docs/resources/aws_vpcs.md.erb
+++ b/docs/resources/aws_vpcs.md.erb
@@ -83,6 +83,15 @@ Filters the results to include only those VPCs that have the given DHCP Option S
       it { should_not exist }
     end
 
+### tags
+
+Check the VPC with Name tag 'my-vpc' exists and is unique
+
+    # Check 'my-vpc' exists
+    describe aws_vpcs.where{ !tags.nil? and tags.find { |h| h[:key] == 'Name' }[:value] == 'my-vpc' }
+      its('count') { should cmp 1 }
+    end
+
 ## Properties
 
 ### cidr_blocks
--- a/lib/resources/aws/aws_vpcs.rb
+++ b/lib/resources/aws/aws_vpcs.rb
@@ -15,6 +15,7 @@ class AwsVpcs < Inspec.resource(1)
   filter.register_custom_matcher(:exists?) { |x| !x.entries.empty? }
   filter.register_column(:cidr_blocks, field: :cidr_block)
         .register_column(:vpc_ids, field: :vpc_id)
+        .register_column(:tags, field: :tags)
   # We need a dummy here, so FilterTable will define and populate the dhcp_options_id field
   filter.register_column(:dummy, field: :dhcp_options_id)
         .register_column(:dhcp_options_ids) { |obj| obj.entries.map(&:dhcp_options_id).uniq }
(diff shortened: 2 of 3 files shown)


clintoncwolfe:
I'd like to hold off on this until we decide how to handle tags on plural resources.  What you have is practical - and if you renamed the field to be something like 'raw_tags', I think we might accept it.  We're wanting to reserve the method 'tags' for something that does something more sophisticated, hoping to avoid the awkward search within an array of hashes (which you include in your example, and we think would be a common use case).

See #3139 and discuss there if you have ideas.

rikturnbull:
The actual problem I'm trying to solve is a lookup of a single resource using a name (which is often a tag). I'm currently writing AWS tests against a fork and building up a worrying number which will need migrating in the future. That fork supports the following:

```
describe aws_subnet(subnet_name: 'my-subnet') do
  it { should exist }
end

describe aws_route_table(route_table_name: 'my-route-table') do
  it { should exist }
end

describe aws_vpc(vpc_name: 'my-vpc') do
  it { should exist }
end
```

This PR for the plural `aws_vpcs` resource was to create a pattern (to dispense with the fork) so that I can do a find against the plural resource and then examine the richer singular resource (second attempt at https://github.com/inspec/inspec/pull/3630).

Unfortunately, as you suggest this results in the rather inelegant:

```
# must first check 'my-vpc' exists and is unique
describe aws_vpcs.where{ !tags.nil? and tags.find{|h| h[:key] == "Name"}[:value] == "my-vpc"} do
  its('count') { should cmp 1 }
end

# now can check 'my-vpc' attributes
aws_vpcs.where{ !tags.nil? and tags.find{|h| h[:key] == "Name"}[:value] == "my-vpc"}.vpc_ids.each do |vpc_id|
  describe aws_vpc(vpc_id) do
    its ('state') { should eq 'available' }
  end
end
```

I can see a number of open discussions that are relevant here (in particular as this naming lookup seems to be a pattern used by awspec):

https://github.com/inspec/inspec/issues/3139
https://github.com/inspec/inspec/issues/3641
https://github.com/inspec/inspec/issues/3621

I'm not entirely sure how to move forward. Perhaps the pattern should indeed be as I've described:

```
describe aws_RESOURCE(RESOURCE_name: 'my-label') do
  it { should exist }
end
```

and also just:

```
describe aws_RESOURCE('my-label') do
  it { should exist }
end
```

I've resubmitted https://github.com/inspec/inspec/pull/3630 with unit tests.

clintoncwolfe:
I think we should keep in mind the goal of this change - to allow searching by VPC name.  That's easy to approve and agree to.  The fact that a VPC name comes from a tag is an implementation detail.  So long as you don't expose the tag functionality, this PR stays out of the impasse of "how to query tags".

So, I would recommend #3630 be a simplified PR to add name-querying support to the singular `aws_vpc` resource (without exposing tags for now), and this PR to do the same for the plural resource.

tbugfinder:
> I think we should keep in mind the goal of this change - to allow searching by VPC name. That's easy to approve and agree to. The fact that a VPC name comes from a tag is an implementation detail. So long as you don't expose the tag functionality, this PR stays out of the impasse of "how to query tags".
> 
> So, I would recommend #3630 be a simplified PR to add name-querying support to the singular `aws_vpc` resource (without exposing tags for now), and this PR to do the same for the plural resource.

@clintoncwolfe I think tag:Name is just a single specific tag. I would prefer having access to all tags not just a specific one.

ghost:
This is being added via https://github.com/inspec/inspec-aws/issues/25 which is currently WIP.

PR #3673, opened 2018-12-18
docker.containers: Ensure .labels returns an Array

### LINKED ISSUE #3653: docker: label matching not working with multiple labels
### 🎛 Description

🐛 bug report

Label matching doesn't seem to work correctly when we have multiple labels. It seems like `include` is doing an exact match? Please see below.

OS: Linux
```

$ inspec version
3.1.3
```
### 🤔 Replication Case

Test:

```
describe docker.containers.where { names == "amazing_noether" } do
    its('labels') { should include 'version=foo' }
end
```

Run an example container:

```
$ sudo docker run --label="app=redis" --label="version=foo" -d redis
```
Run test:

```
 sudo inspec  exec test.rb

Profile: tests from test.rb (tests from test.rb)
Version: (not specified)
Target:  local://

  #<Inspec::Resources::DockerContainerFilter:0x000000000587aac8> with names == "amazing_noether"
     ×  labels should include "version=foo"
     expected ["app=redis,version=foo"] to include "version=foo"

```
The issue seems to be that `["app=redis, version=foo"]` should really be `["app=redis", "version=foo"]`. I tried my hand at fixing this up, but not yet been successful.

### 💁 Possible Solutions

Not sure.

### 💻 Stacktrace

NA

### LINKED ISSUE #3654: docker: "include" labels support for `where`
### 🎛 Description

🙋 feature request

One of the examples for the docker resource is:

```
describe docker.containers.where { names == 'flamboyant_colden' } do
  it { should be_running }
end
```

It also mentions that i can use `labels` in the where criteria.  How do I say, select containers which includes a label? Is something like this possible `..where { labels include 'my=label'}` ?

NA
### 🤔 Replication Case

NA
### 💁 Possible Solutions

I would try to implement this feature if given some guidance.

jerryaldrichiii:
This modifies `labels` in FilterTable to split values on `,`.

Without this the value for `labels` would be `'foo=bar,spam=eggs'`. Which would make `docker.containers.where { labels.include?('foo=bar') }` not behave as expected.

Many thanks to @amitsaha for helping me solve this!

This fixes #3653


jerryaldrichiii:
Could use someone with more FilterTable knowledge to let me know if this is "correct" holistically.

This allows for `docker.containers.where { labels.include?('foo=bar') }` but doesn't really do much else and changes the value of `lables` to return an Array of Arrays instead of an Array of Strings.

A user could also have done `docker.containers.where { labels.match?(/foo=bar/) }`.

miah (approved):
Thanks @jerryaldrichiii! I think this looks solid. =)

lib/resources/docker.rb
```diff
@@ -203,8 +203,13 @@ def parse_json_command(labels, subcommand)
         # does not include a slash since that is not a valid character in a container name
         j['names'] = j['names'].split(',').find { |c| !c.include?('/') } if j.key?('names')
 
+        # Split labels on ','
+        # This allows `its('labels') { should include 'foo=bar' }` to work
```
clintoncwolfe: Please do not use 'foo' in code.

docs/resources/docker.md.erb
```diff
@@ -136,7 +136,7 @@ Or execute the profile directly via URL:
       its('commands') { should_not include '/bin/sh' }
       its('images') { should_not include 'u12:latest' }
       its('ports') { should include '0.0.0.0:1234->1234/tcp' }
-      its('labels') { should include 'License=GPLv2,Vendor=CentOS' }
+      its('labels') { should include ['License=GPLv2', 'Vendor=CentOS'] }
```
clintoncwolfe: This is probably not what is intended.  This would indicate that you are expecting `labels` to be an Array containing a nested Array containing two Strings in exactly that order.

The existing code was more accurate but not very usable - the `labels` field would be a String, and you can search a string for substrings using `include`.

Calling `include` on an Array does something completely different, looking at each element in isolation - it does not look for "subarrays", as your example would imply.

If you want to look for two different labels when `labels` is an Array, you'll want to do so in two tests:

```
# labels is an Array of Strings; look for individual elements
its('labels') { should include 'License=GPLv2' }
its('labels') { should include 'Vendor=CentOS' }
```

lib/resources/docker.rb
```diff
@@ -203,8 +203,13 @@ def parse_json_command(labels, subcommand)
...
+        j['labels'] = j['labels'].split(',') if j.key?('labels')
```
clintoncwolfe: I encourage you to think of the [Boy Scout Rule](https://deviq.com/boy-scout-rule/).  Consider replacing the `j`, `k`, and `v` variables with names that are meaningful to humans and amenable to grepping.

clintoncwolfe: Not sure if I understand the code flow here, so this may not be a concern:

Is it possible for the JSON data to not contain the key 'labels', as the conditional suggests?  In that case, how will your users expect the `labels` field to behave in the InSpec resource?  As-is, I suspect that the `labels column will end up with nil values, so 

```
its('labels') { should include 'something-i-care-about' }
```

will throw a NoMethod error, looking for #include? on Nil::NilClass.

Consider setting it to be an empty array if there are no labels to parse.

test/unit/mock/cmd/docker-ps-a
```diff
@@ -1,4 +1,4 @@
-{"Command":"\"/bin/bash\"","CreatedAt":"2017-04-24 10:29:12 +0200 CEST","ID":"3def9aa450f8bd772c3d5b07e27ec934e5f58575e955367a0aca2d93e0687536","Image":"ubuntu:12.04","Labels":"","LocalVolumes":"0","Mounts":"","Names":"sleepy_khorana","Networks":"bridge","Ports":"","RunningFor":"29 minutes","Size":"0 B","Status":"Exited (127) 2 seconds ago"}
```
clintoncwolfe, line 1: Seems like it might be a good idea to *keep* a test case in which there are no labels.

clintoncwolfe (changes requested):
Thanks for the turnaround on this, @jerryaldrichiii ! My main concerns are stylistic, getting a better example, and making sure we handle the no-labels case well.  Thanks!

lib/resources/docker.rb
jerryaldrichiii: Ah, even though it is a comment it is wrong regardless. This PR does not allow that. Will update (and replace `foo` with something meaningful).

Just curious, why shouldn't I use `foo`?

lib/resources/docker.rb
jerryaldrichiii: Can do!

jerryaldrichiii: I just copied what was above to match style. Given the Boy Scout Rule you mentioned, I'll change both to be more Ruby like (e.g. `if j['names]`). I'll also set `labels` to an empty array if the field isn't present.

test/unit/mock/cmd/docker-ps-a
jerryaldrichiii: All the other containers do not have labels and there are tests for them.

Docker does not return valid JSON but instead a JSON for each result so this may explain things, see the file for more context.

docs/resources/docker.md.erb
jerryaldrichiii: I think while not ideal, the way I have it is correct. `docker.containers.labels` returns and Array of Arrays. Which is why I highlighted in the first comment on the PR my hesitancy (https://github.com/inspec/inspec/pull/3673#issuecomment-448057106)

Unless I'm misunderstanding something?:

```
   11: it 'check docker container parsing' do
    12:   require 'pry'; binding.pry
 => 13:   _(resource.containers.ids).must_equal ['3def9aa450f8bd772c3d5b07e27ec934e5f58575e955367a0aca2d93e0687536', 'd94f854370d2b02912e8fc636502bc72b74fbd567a7eba3fc6a52045bb28904e', '5a83c301f30ccd48579a74a84af6fdd0c0e0d66aacc7bb52abfa2ba2544c6c0c', '5a83c301f30ccd48579a74a84af6fdd0c0e0d66aacc7bb52abfa2ba2544c6c0c']
    14:   _(resource.containers.names).must_equal ['sleepy_khorana', 'laughing_austin', 'heuristic_almeida', 'laughing_lamport']            
    15:   _(resource.containers.labels).must_equal [['app=example', 'version=1.5.4'], [], [], []]                                           
    16: end

[1] pry(#<Inspec::Resources::Docker::docker>)> resource.containers.labels                                                                   
=> [["app=example", "version=1.5.4"], [], [], []]
[2] pry(#<Inspec::Resources::Docker::docker>)> resource.containers.labels.include?('app=example')                                           
=> false
[3] pry(#<Inspec::Resources::Docker::docker>)> 
```

lib/resources/docker.rb
clintoncwolfe: Anytime you need to provide an example, try to think of a meaningful example that a user might encounter in the real world.  To my eyes, "foo bar baz" etc reads as unhelpful - it implies that you don't care about the user (or your fellow engineers) enough to provide a useful example.

It's not a hard and fast issue - using it in a comment is no great crime - and it certainly is subjective.

jerryaldrichiii: Sure, I guess it boils down to "What harm is there to NOT using foo?". Nothing but benefits really.

docs/resources/docker.md.erb
clintoncwolfe: Ahh, you're right, yes, as implemented dockers.containers.labels will indeed return an Array of Arrays.  I was speaking out of the wrong end of my digestive tract.  Your example code, however, is not likely to match: it will find containers that have exactly those two labels in that order, which is painful.

You've asked privately about FilterTable, and yes, here is where a FilterTable feature will help - we really want that column flattened and de-duped, so it is an Array of Strings.  You will get exactly that effect if you add `style: :simple` to the `register_column` call for `labels` on DockerContainer (resources/docker.rb:16). See the FilterTable Usage guide [here](https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md#a-colors-method-that-will-return-a-flattened-and-uniqd-array-of-values) and [here](https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md#how-do-i-declare-my-interaction-with-filtertable).

With style: simple enabled, we would now have:

```
# labels (really) is an Array of Strings; look for individual elements

# At least one matching container has the License=GPLv2 label
its('labels') { should include 'License=GPLv2' }
# At least one matching container has the Vendor=CentOS label
its('labels') { should include 'Vendor=CentOS' }
```

Using style: :simple has consequences.  You can now only do 'OR' operations. You lose the ability to identify containers that have specific pairs of labels, that is, 'AND'.  However, that can be overcome using chaining.

```
docker_containers.where { labels.include 'License=GPLv2' }
                 .where { labels.include 'Vendor=CentOS' }
```

test/unit/mock/cmd/docker-ps-a
clintoncwolfe: Ahh, OK.  I just saw the line deletion and didn't poke around further.

docs/resources/docker.md.erb
jerryaldrichiii: I think OR operations only are fine here! I'll update and change the examples.

clintoncwolfe (approved):

test/unit/resources/docker_container_test.rb
```diff
@@ -24,6 +24,7 @@
       _(resource.tag).must_equal '12.04'
       _(resource.command).must_equal '/bin/bash'
       _(resource.ports).must_equal ''
+      _(resource.labels).must_equal ['app=example', 'version=1.5.4']
```
clintoncwolfe, line 27: This currently has a failing unit test, which needs to be fixed.

Also, you mentioned that there was an existing unit test which tested a container with zero labels; could you show me where that is?

clintoncwolfe (changes requested):
Looks like there is a failing unit test on the label reader.  Other issues are resolved, good work!

See https://travis-ci.org/inspec/inspec/jobs/469749939#L1103

test/unit/resources/docker_container_test.rb
jerryaldrichiii: Yup, looks like I needed to fix the `docker_container` resource since `labels[0]` is no longer correct.

Also, I didn't see where a test where a docker container was explicitly returning `[]` when no labels are used. Couldn't hurt to have an explicit test instead of relying on `docker.containers.labels` being correct. I'll make it so.

clintoncwolfe (approved):

clintoncwolfe:
Our TravisCI setup currently has an issue, breaking unit and functional tests.

I ran unit and functional tests locally, success, on ref
```
d72b65df (HEAD -> ja/fix-docker-labels, origin/ja/fix-docker-labels) Add test for `docker_container` with no labels
```

PR #3678, opened 2018-12-19
Update Ruby versions used for testing; fix Travis CI issues

tas50:
This PR updates Ruby versions used in CI testing to the latest patchlevels:
 * 2.3.6 => 2.3.8
 * 2.4.3 => 2.4.5
 * 2.5.1 => 2.5.3

Pile of bugfixes and 2 CVEs:

    CVE-2018-16396: Tainted flags are not propagated in Array#pack and String#unpack with some directives
    CVE-2018-16395: OpenSSL::X509::Name equality check does not work correctly

The upgrade has two other consequences:
 * bundler is now included with the version of rubygems included in each of these rubies.  Additionally, TravisCI's ruby build pack includes this newer rubygem; **this causes all TravisCI jobs to halt**, since we were installing bundler as a pre-install job, and it wanted overwrite confirmation.  Removing the extra install request resolved the issue.
 * rubygems has a few small internal changes that cause the plugin installer system to fail tests.  Both fixes are backward compatible.
   * It now hits both api.rubygems.org and plain rubygems.org in API requests. Our unit tests needed to be updated to allow this.
   * The search library, Gem::RemoteFetcher, is no longer auto-loaded; so we explicitly load it in the installer library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ bundler_args: "--without integration tools maintenance deploy"
 before_install:
 - gem update --system
 - gem --version
-- rvm @global do gem install bundler
 - bundle --version
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,41 +15,41 @@ before_install:
 - bundle --version
 matrix:
   include:
-  - rvm: 2.3.6
-  - rvm: 2.4.3
-  - rvm: 2.5.1
-  - rvm: 2.4.3
+  - rvm: 2.3.8
+  - rvm: 2.4.5
+  - rvm: 2.5.3
+  - rvm: 2.4.5
     script: bundle exec rake $SUITE
     env: SUITE="test:functional"
-  - rvm: 2.4.3
+  - rvm: 2.4.5
     script: bundle exec rake $SUITE
     bundler_args: "--without tools maintenance deploy"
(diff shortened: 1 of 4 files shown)


clintoncwolfe (dismissed):
This is wedging in TravisCI, with bundle trying to overwrite its own binstub.  Any thoughts?

https://travis-ci.org/inspec/inspec/jobs/469876553#L871

Update: wedging is resolved by #3687

clintoncwolfe:
I'm going to adopt this; it's intertwined with #3687 , which is due to TravisCI updating their OS ruby.  This has some cascading effects.

tas50:
Clinton you might wanna peak at the ohai travis. We have some of these same issues. Bundler in travis is :(

clintoncwolfe:
It looks like our bundler issues - timing out while RVM waited at an interactive prompt, to allow installation of bundler when an executable already existed - was simple to resolve.  Th extra install command was part of our TravisCI config!

We do have more serious issues due to internal API changes within rubygems.  The InSpec plugin facility uses rubygems internals, so we need to make some adjustments.

jerryaldrichiii (approved):
Looks good to me! Thanks @tas50!

miah (approved):
Thanks @tas50 !

Issue #3594, opened 2018-11-08
Unable to make oracledb_session work at all.


g1ps:
### 🎛 Description

Again, I need to ask for assistance. After many hours of trying I can't get this to work at all.

My env is this:

Machine A -> Machine B -> Machine C

A is running the inspec tests. It connects to B which, in turn, can access C and has Oracle tools installed and working. A is RHEL, B & C are Windows machines. I am running this, based on the example in the docs, which also fails:
````
  sql = oracledb_session(user: 'system', password: 'sys', host: '<C's name>', service: '<service>')

 describe sql.query('SELECT NAME FROM v$database;').rows do
    its('count') { should eq 1 }
  end
````
inspec always returns:
````
  []
     x  count should eq 1

     expected: 1
          got: 0

     (compared using ==)
````
Which suggests strongly to me that it isn't connecting to the DB.
I have tried everything I can find (which is almost nothing) and think of. The query works if I run it directly. The user is fine. The details are correct. I tried connecting directly to C, the DB server. No change. I tried various additional parameters: `sqlplus_bin`, `as_db_role`. I tried the host IPs instead of names. I've tried other users. I've tried several queries.

Warning and error `--log-level` output show nothing. debug output shows nothing that looks like a problem to me. The query it shows will run if I run it manually. The login string it generates works on the command line on B. I'm out of ideas. What am I missing here?

### 🌍 InSpec and Platform Version

inspec 3.0.12

**Comment by tbugfinder** [comment]
What's the output of `puts sql`?

**Comment by g1ps** [comment]
Run where?

**Comment by tbugfinder** [comment]
```
  sql = oracledb_session(user: 'system', password: 'sys', host: '<C's name>', service: '<service>')

puts sql

puts sql.query('SELECT NAME FROM v$database;')

 describe sql.query('SELECT NAME FROM v$database;').rows do
    its('count') { should eq 1 }
  end
```

g1ps:
OK, thanks.
````
Oracle Session
SQL ResultSet

Profile: tests from test_spec.rb (tests from test_spec.rb)
Version: (not specified)
Target:  winrm://********@http://**************:5985/wsman:3389

  []
     x  count should eq 1

     expected: 1
          got: 0

     (compared using ==)

Test Summary: 0 successful, 1 failure, 0 skipped
-sh-4.2$
````

clintoncwolfe:
@miah any thoughts?

[...]

PR #3680, opened 2018-12-19
Fix oracledb_session SQLCL and CSV output parsing

gsreynolds:
## Description

The `oracledb_session` resource is non-functional against the following: `Oracle Standard Edition One (11.2.0.4.v18)` on RDS using `sqlcl-18.3.0.259.2029`

* sqlcl_bin option isn't used if provided
* silent options for sqlcl and sqlplus binaries not passed, need to remove command prompt, command echoing and connection banners to avoid breaking CSV parsing
* Fix 'no implicit conversion of Regexp into String' with string `.delete`
* Leading blank line in output breaks CSV parsing

Potentially related to #3594

The CSV output is preferred but appears not to be covered with the unit test mocking (test/unit/mock/cmd/oracle-result is the HTML format)

## InSpec Version
3.0.52 (last change to `oracledb_session` resource was v2.2.35 in https://github.com/inspec/inspec/pull/3170)

## Replication Case
```ruby
control 'test' do
  impact 1.0
  title 'Test'
  desc 'Test oracledb_session'
  sql = oracledb_session(user: '<USER>', password: '<PASS>', host: '<HOST>', port: '1521', service: 'ORCL', sqlcl_bin: '<PATH TO SQLCL>')
  describe sql.query('select username, profile from dba_users;').column('profile') do
    it { should include 'DEFAULT' }
  end      

  describe sql.query('SELECT NAME AS VALUE FROM v$database;').row(0).column('value') do
    its('value') { should eq 'ORCL' }
  end
end
```

## Output
Before:
```
Profile: InSpec Profile (oracledb)
Version: 0.1.0
Target:  local://

  ×  test: TEST (2 failed)
     ×  [] should include "DEFAULT"
     expected [] to include "DEFAULT"
     ×  SQL Column value should eq "ORCL"

     expected: "ORCL"
          got: ""

     (compared using ==)
```
After:
```
Profile: InSpec Profile (oracledb)
Version: 0.1.0
Target:  local://

  ✔  test: TEST
     ✔  ["RDSADMIN", "RDSADMIN", "DEFAULT", "RDSADMIN", "DEFAULT", "DEFAULT", "DEFAULT", "RDSADMIN", "DEFAULT"] should include "DEFAULT"
     ✔  SQL Column value should eq "ORCL"

```

### DIFF
--- a/lib/resources/oracledb_session.rb
+++ b/lib/resources/oracledb_session.rb
@@ -99,7 +99,9 @@ module Inspec::Resources
     end
 
     def parse_csv_result(stdout)
-      output = stdout.delete(/\r/)
+      output = stdout.delete("\r")
+      # Remove blank first line if it exists
+      output = output.gsub(/^$\n/, '')
       table = CSV.parse(output, { headers: true })
 
       # convert to hash
--- a/lib/resources/oracledb_session.rb
+++ b/lib/resources/oracledb_session.rb
@@ -40,7 +40,7 @@ module Inspec::Resources
       @su_user = opts[:as_os_user]
       @db_role = opts[:as_db_role]
 
-      # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
+      # we prefer sqlcl although it is way slower than sqlplus, but it understands csv properly
       @sqlcl_bin = opts[:sqlcl_bin] || 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
       @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'
 
--- a/lib/resources/oracledb_session.rb
+++ b/lib/resources/oracledb_session.rb
@@ -41,7 +41,7 @@ module Inspec::Resources
       @db_role = opts[:as_db_role]
 
       # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
-      @sqlcl_bin = 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
+      @sqlcl_bin = opts[:sqlcl_bin] || 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
       @sqlplus_bin = opts[:sqlplus_bin] || 'sqlplus'
 
       return fail_resource "Can't run Oracle checks without authentication" if @su_user.nil? && (@user.nil? || @password.nil?)
--- a/lib/resources/oracledb_session.rb
+++ b/lib/resources/oracledb_session.rb
@@ -68,11 +68,11 @@ module Inspec::Resources
       query = verify_query(escaped_query)
       query += ';' unless query.end_with?(';')
       if @db_role.nil?
-        command = %{#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
+        command = %{#{bin} -S "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
       elsif @su_user.nil?
-        command = %{#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
+        command = %{#{bin} -S "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
       else
-        command = %{su - #{@su_user} -c "env ORACLE_SID=#{@service} #{bin} / as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC"}
+        command = %{su - #{@su_user} -c "env ORACLE_SID=#{@service} #{bin} -S / as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC"}
       end
       cmd = inspec.command(command)
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -473,7 +473,7 @@ class MockLoader
       "7d1a7a0f2bd1e7da9a6904e1f28981146ec01a0323623e12a8579d30a3960a79" => cmd.call('mssql-result'),
       # oracle
       "bash -c 'type \"sqlplus\"'" => cmd.call('oracle-cmd'),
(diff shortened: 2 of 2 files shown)


btm:
@gsreynolds can you look at the travis and appveyor failures on this?

gsreynolds:
Hi @btm, from what I can see the failures are unrelated to the PR. 

All the Travis failures are:
```
$ rvm @global do gem install bundler
bundler's executable "bundle" conflicts with /home/travis/.rvm/rubies/ruby-2.3.6/bin/bundle
Overwrite the executable? [yN]  
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

The two Appveyor fails are apparently also unrelated:
```
inspec json::inspec json with a inheritance profile#test_0002_populates code from child profiles:
NoMethodError: undefined method `[]' for nil:NilClass
    C:/projects/inspec/test/functional/inspec_json_profile_test.rb:145:in `block (3 levels) in <top (required)>'
```

I've rebased on master and force pushed to see if that clears the failures.

gsreynolds:
Hi @btm, I fixed an oversight on my part in `test/helper.rb` which has fixed the unit tests and Travis now passes.

However, Appveyor is still suffering from random failures, Appveyor passed once but when I tided my git history and pushed again, it has now failed again :(

lib/resources/oracledb_session.rb
```diff
@@ -41,7 +41,7 @@ def initialize(opts = {})
       @db_role = opts[:as_db_role]
 
       # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
-      @sqlcl_bin = 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
+      @sqlcl_bin = opts[:sqlcl_bin] || 'sql' unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
```
clintoncwolfe, line 44: The comment mentions something called sqlci, the option is named sqlcl, and the actual value is either 'sql' or 'sqlplus'.  Could you clarify?  Assume the maintainer of this code is not familiar with Oracle.

clintoncwolfe, line 44: This appears to introduce a new resource parameter, `sqlcl_bin` but does not document that in the documentation.

test/helper.rb
```diff
@@ -473,7 +473,7 @@ def md.directory?
       "7d1a7a0f2bd1e7da9a6904e1f28981146ec01a0323623e12a8579d30a3960a79" => cmd.call('mssql-result'),
       # oracle
       "bash -c 'type \"sqlplus\"'" => cmd.call('oracle-cmd'),
-      "1998da5bc0f09bd5258fad51f45447556572b747f631661831d6fcb49269a448" => cmd.call('oracle-result'),
+      "bfff2fe305897dbd92d32be74662278dfd898c97fae5c34e929e40ad2d92732c" => cmd.call('oracle-result'),
```
clintoncwolfe, line 476: Why does the hash change?  I don't see a matching change to a command mock file...

lib/resources/oracledb_session.rb
```diff
@@ -68,11 +68,11 @@ def query(q)
       query = verify_query(escaped_query)
       query += ';' unless query.end_with?(';')
       if @db_role.nil?
-        command = %{#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
+        command = %{#{bin} -S "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
```
clintoncwolfe, line 71: A set of comments picking these apart - or links to online Oracle docs - would help the maintainability of this code. A lot happens in different ways, and I don't see the reasoning.

clintoncwolfe (changes requested):
Thanks for the contribution! However, since this adds a new resource parameter, we need to document that.  Also, clarifying comments would really help; the InSpec team does not work with Oracle databases very much, so anything you can do to explain the code to people unfamiliar with it would be great.

Finally, and I understand this is likely out of scope, I would feel a lot more comfortable if we had real integration tests. While a unit test based on canned output is a good start, I would feel a lot better about this if we had a set of rake tasks to setup, test more thoroughly, and then teardown some RDS instances; something like #3684.

lib/resources/oracledb_session.rb
gsreynolds: `sqlcl_bin` was an existing resource parameter that is documented but was previously unused in the code.
https://github.com/inspec/inspec/blob/8cdd7cfe0309528e2c866e1c311ee854e012e94a/docs/resources/oracledb_session.md.erb#L34

gsreynolds: The comment is incorrect, i've corrected it to `sqlcl`

test/helper.rb
gsreynolds: The hash changed as the `-S` silent options were added in the commands (I'm not sure why there isn't a mock file)

lib/resources/oracledb_session.rb
gsreynolds: This logic was added in the previous PR that touched this resource: https://github.com/inspec/inspec/pull/3170. The way the logic is constructed appears to do:
1. User & password authentication
2. User & password authentication plus assuming a database role
3. User & password authentication plus assuming a database role executing the binary as root

lib/resources/oracledb_session.rb
clintoncwolfe: Ahh, thanks!

lib/resources/oracledb_session.rb
clintoncwolfe: Yes, and I found it confusing then, as well.  What I was asking for here is comments *in code* explaining what is going on, or a URL *in a code comment* where a maintainer with little knowledge of Oracle can go to understand what is happening, and why it happens three different ways.

miah (changes requested):
Thanks for the contribution. This looks solid, could you address the changes requested by @clintoncwolfe after that this has my approval.

clintoncwolfe:
(boilerplate message)

miah (approved):
I think this fixes a few issues we're currently dealing with, so lets get this pulled through. @clintoncwolfe can you give this another look over? We can add the comment about the login stuff afterwards (I'll do it).

clintoncwolfe (approved):
Thanks!

miah:
I'm leaning towards getting #4654 merged and closing this one out. The only additional that this one does is with sqlcl, but there are no tests. I have sqlcl locally,  so i can start to build that out. But I want to get sqlplus into a zone of stability first =)

PR #3687, opened 2018-12-20
Fix TravisCI hanging due to Bundler being installed twice

clintoncwolfe:
Remove bundle install command from travis config, as it is already installed with Ruby.


miah (approved):
Looks good!

clintoncwolfe:
This has some knock-on effects with #3678 	- I'm going to combine them.

PR #3690, opened 2018-12-26
Centralized Deprecation Facility

### LINKED ISSUE #465: Unify deprecation handling in resources
Get here:

``` ruby
if condition
  deprecated 'Please use xyz instead of abc', '1.0'
end
```

Which leads to

``` bash
[DEPRECATION] Please use xyz instead of abc. This method will be remove in version 1.0.
```

Global access for deprecation call
Ability to enable/disable deprecations

clintoncwolfe:
Implements a centralized deprecation handling facility as [specified here](https://github.com/inspec/inspec/issues/465#issuecomment-444598679).

This PR:
 * Adds a set of classes in `lib/utils/deprecation` to handle deprecations
 * Reads and validates a config file from `etc/deprecations.json`
 * Provides a developer doc in `docs/dev/deprecation.md` which provides an overview of how it can be used
 * Defines Inspec.deprecate, a top-level entry point to use the system
 * Unit tests for the classes
 * A test fixture profile and functional tests to exercise ignoring, warning, failing a control, and exiting inspec in response to a deprecation

This PR does *not* add any changes to the InSpec main codebase to actually use the new system.

Closes #465


frezbo:
@clintoncwolfe Awesome work.

lib/utils/deprecation/config_file.rb
```diff
@@ -0,0 +1,108 @@
...
+require 'inspec/globals'
+
+module Inspec
+  module Deprecation
+    class ConfigFile
+      GroupEntry = Struct.new(:name, :action, :prefix, :suffix, :exit_status, :occurences)
+
+      # What actions may you specify to be taken when a deprecation is encountered?
+      VALID_ACTIONS = [
+        # Hard exit inspec, no stacktrace, exit code specified or DEFAULT_HARD_EXIT_STATUS
```
clintoncwolfe: I'm not sure DEFAULT_EXIT_STATUS is used - I think it just calls Inspec::UI.exit

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,109 @@
+
+require 'utils/deprecation/config_file'
+require 'inspec/log'
+
+module Inspec
+  module Deprecation
+    class Deprecator
+      attr_reader :config, :groups
+
+      # This is used only in functional testing
```
clintoncwolfe: Is adding code to the main codebase to facilitate testing a good idea?

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      def initialize(opts = {})
+        @@test_cfg_io ||= nil # rubocop: disable Style/ClassVars
+        @config = Inspec::Deprecation::ConfigFile.new(opts[:config_io] || @@test_cfg_io)
+        @groups = @config.groups
+      end
+
+      def handle_deprecation(group_name, message, opts = {})
+        group = groups[group_name.to_sym] # Will be nil if unrecognized group
```
clintoncwolfe: Throughout, the code has awkward conditionals because it cannot rely on having a GroupEntry record.  How could that be handled better?

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,109 @@
...
+        assembled_message = assemble_message(message, group, group_name, opts)
+
+        action = (group ? group[:action] : :warn) || :warn
+        action_method = ('handle_' + action.to_s + '_action').to_sym
+        send(action_method, assembled_message, group)
+      end
+
+      private
+
+      def annotate_stack_information(opts)
```
clintoncwolfe, line 34: How do you feel about stack-probing heuristics?

clintoncwolfe (commented):
After re-reading this a while later, here are a few things that stuck out to me.

lib/utils/deprecation/config_file.rb
```diff
@@ -0,0 +1,108 @@
...
+        end
+        unless @raw_data['file_version'] == '1.0.0'
+          raise Inspec::Deprecation::InvalidConfigFileError, "Unrecognized file_version '#{@raw_data['file_version']}' - supported versions: 1.0.0"
+        end
+      end
+
+      def validate_unknown_group_action
+        seen = (@raw_data['unknown_group_action'] || @unknown_group_action).to_sym
+        unless VALID_ACTIONS.include?(seen)
+          raise Inspec::Deprecation::UnrecognizedActionError, "Unrecognized action for unknown groups '#{seen}' - supported actions: #{VALID_ACTIONS.map(&:to_s).join(', ')}"
```
Vasu1105: If I am reading it correctly I think message should be `Unrecognised action '#{seen} for unknown groups ...`

lib/utils/deprecation/deprecator.rb
lamont-granquist: brittle, but necessary.  the client does this in its error message decorator and in the deprecations it throws.

lib/utils/deprecation/config_file.rb
jerryaldrichiii: I would like that or:

```
"Unrecognized action for unknown groups: '#{seen}' - supported actions:
```

jerryaldrichiii: Or, we could match the below and do:

```
"Unrecognized action for unknown groups - saw '#{seen}', supported fields:
```

lib/utils/deprecation/deprecator.rb
jerryaldrichiii: We've done it in the past (I think by checking for a mock transport). I'd like to avoid it if possible...but I don't know how here.

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,109 @@
...
+        # Within a control block, that is actually an RSpec:ExampleGroup
+        stack.each do |frame|
+          return true if frame.path.end_with?('rspec/core/example_group.rb')
+        end
+
+        false
+      end
+
+      def handle_ignore_action(message, group)
+        # Inspec::Log.debug("Ignoring deprecation message from #{opts[:caller][:path]}:#{opts[:caller][:lineno]}" )
```
jerryaldrichiii: Can we add a debug message here? Seems like it was started, but the one commented doesn't use `message` or `group`.

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      def handle_log_action(message, level, _group)
```
jerryaldrichiii: Do we need a debug log level here for ignore?

docs/dev/deprecation.md
```diff
@@ -0,0 +1,101 @@
...
+## Overview
+
+The centralized deprecation facility is used to deprecate (retire, withdraw, etc) features both within InSpec core code, InSpec core resources, plugins, and custom resources.
+
+The most important goal of the deprecation facility was to collect decisions about what actions should be taken for a deprecation in a central place, so that it is easier to handle major version changes.
+
+### Example
+
+```
+# In some code in Inspec core
```
tas50:
```suggestion
# In some code in InSpec core
```

lib/utils/deprecation/config_file.rb
```diff
@@ -0,0 +1,108 @@
+require 'stringio'
+require 'json'
+require 'inspec/globals'
+
+module Inspec
+  module Deprecation
+    class ConfigFile
+      GroupEntry = Struct.new(:name, :action, :prefix, :suffix, :exit_status, :occurences)
```
tas50: occurrences has 2 Rs

lib/inspec/ui.rb
```diff
@@ -32,6 +32,7 @@ class UI
     EXIT_NORMAL = 0
     EXIT_USAGE_ERROR = 1
     EXIT_PLUGIN_ERROR = 2
+    EXIT_FATAL_DEPRECATION = 3
```
jerryaldrichiii, line 35: Can we add a corresponding docs change?

lib/utils/deprecation/config_file.rb
jerryaldrichiii: Don't see `:occurences` anywhere else. Can we drop it?

jerryaldrichiii: Oh, I lied, I don't see where we modify it though.
https://github.com/inspec/inspec/pull/3690/files#diff-f1f40a3f8cb29613584f1d9843ca3366R101

lib/utils/deprecation/deprecator.rb
tyler-ball: What about breaking open the class in a `test/spec_helper_deprecator.rb` class and monkey patching it during specs only? It looks like if you did that you would only need to change the `def initialize` method on the class.

It means it becomes possible for the monkey patching to get out of sync (if the real `initialize` method changes in some way) but keeps test-only code out of the class.

It is more work and a little more obfuscated logic, so with the simplicity of this test support code right now it may be better to just leave this alone. Thoughts?

lib/utils/deprecation/config_file.rb
```diff
@@ -0,0 +1,108 @@
...
+      ].freeze
+
+      VALID_GROUP_FIELDS = %w{action suffix prefix exit_status}.freeze
+
+      attr_reader :groups, :unknown_group_action
+
+      def initialize(io = nil)
+        io ||= open_default_config_io
+        begin
+          @raw_data = JSON.parse(io.read)
```
tyler-ball, line 27: One thought - hand writing JSON sucks. What about using YAML or TOML instead? I realize that Inspec developers will be the only ones modifying this file so I'll leave it up to y'all

lib/utils/deprecation/config_file.rb
```diff
@@ -0,0 +1,108 @@
...
+        @raw_data['groups'].each do |group_name, group_info|
+          validate_group_entry(group_name, group_info)
+        end
+      end
+
+      def validate_file_version
+        unless @raw_data.key?('file_version')
+          raise Inspec::Deprecation::InvalidConfigFileError, 'Missing file_version field'
+        end
+        unless @raw_data['file_version'] == '1.0.0'
```
tyler-ball, line 69: Could probably simplify this by making the `file_version` a simple integer. No need to use semver with here, as I imagine it won't be changing a ton in the future

lib/utils/deprecation/deprecator.rb
clintoncwolfe: from review :
``` attr accessor :opts  def initialize(opts = {})   @opts = opts end  def config   @config ||= Inspec::Deprecation::ConfigFile.new(opts[:config_io]) end ```  ```   hash = { config_io: StringIO.new(DEPRECATION_CONFIG) }   expect(obj).to receive(:config).and_return(hash) ``` expect(obj).to receive(:opts).and_return(hash)

File: lib/utils/deprecation/deprecator.rb (hunk 3 above)
Comment (tyler-ball) [reply lib/utils/deprecation/deprecator.rb]: One alternative is to _require_ a GroupEntry record for every type of deprecation. But that gets really hard to test because we would need 100% test coverage. That seems more brittle than your current approach

**Review by tyler-ball — APPROVED** [review]
I had a few minor comments but overall I think this looks like a well documented, designed and implemented change 👍 👍

**Review by jerryaldrichiii — DISMISSED** [review]
A few questions and a request for a docs addition.

See: for style changes https://github.com/inspec/inspec/pull/3725 (non-blocking)

File: docs/dev/deprecation.md (hunk 8 above)
Comment (jerryaldrichiii) [reply docs/dev/deprecation.md]: Check out: https://github.com/inspec/inspec/pull/3725

The suggestion feature is quite cumbersome and doesn't DCO sign-off.

File: test/unit/mock/profiles/deprecation/resource_pack/libraries/deprecation_tester.rb [hunk 13]
```diff
@@ -0,0 +1,62 @@
...
+      "an_ignored_group" : { "action": "ignore" },
+      "a_group_that_will_fail" : { "action": "fail_control" }
+    }
+  }
+  EOC
+
+  def fail_me
+    Inspec::Deprecation::Deprecator.class_test_cfg_io(StringIO.new(DEPRECATION_CFG))
+
+    #deprecate(:a_group_that_will_fail, 'This should fail')
```
jerryaldrichiii: Are these comments needed? I removed them https://github.com/inspec/inspec/pull/3725 but...want to make sure that was the right action.

lib/utils/deprecation/config_file.rb
clintoncwolfe: Thank you, @Vasu1105, for pointing out that `seen` referred to an action, not a group.

test/unit/mock/profiles/deprecation/resource_pack/libraries/deprecation_tester.rb
clintoncwolfe: The background here is that the original issue asked for a new resource DSL method, `deprecate`; and did not extend to making the facility available throughout the codebase.

Once `Inspec.deprecate` was added, having bare `deprecate` seemed a bit magical, and was a distinction without a difference, so I didn't go that route.  The commented out lines were cruft, and should indeed be removed.

lib/utils/deprecation/config_file.rb
clintoncwolfe: Very good catch Jerry, thanks.  I'm removing it.

lib/utils/deprecation/config_file.rb
clintoncwolfe: InSpec uses JSON for config files today - the main config, the plugins config, and the plugins filter config.

User-visible files are YAML, such as the profile metadata and attributes.

We could certainly open an issue to migrate all to YAML; that would also involve changes to A2.

Within the scope of this PR, "what format should InSpec use for internal config?", the answer is JSON.

clintoncwolfe: That said, how about let's add a field, `comment`, to the valid group entries.  That won't be used by the deprecation system, but would be available to the person editing the file to explain why the entry is there.

lib/utils/deprecation/deprecator.rb
clintoncwolfe: I've added code to filter out RSpec, as suggested.

lib/utils/deprecation/deprecator.rb
clintoncwolfe: That sounds like a good approach, consistent with the others.

lib/utils/deprecation/deprecator.rb
```diff
@@ -0,0 +1,108 @@
...
+        @config = Inspec::Deprecation::ConfigFile.new(opts[:config_io] || @@test_cfg_io)
+        @groups = @config.groups
+      end
+
+      def handle_deprecation(group_name, message, opts = {})
+        group = groups[group_name.to_sym] # Will be nil if unrecognized group
+        annotate_stack_information(opts)
+        assembled_message = assemble_message(message, group, group_name, opts)
+
+        action = (group ? group[:action] : :warn) || :warn
```
clintoncwolfe: This right here is a bug - if there is no group, we're supposed to use the unknown_group_action.  Instead, this always warns.

lib/utils/deprecation/deprecator.rb
clintoncwolfe: I've modified the code to create a GroupEntry when no group is found in the config file.

lib/inspec/cli.rb
```diff
@@ -175,14 +175,19 @@ def archive(path)
   end
 
   desc 'exec LOCATIONS', 'run all test files at the specified LOCATIONS.'
+  # TODO: find a way for Thor not to butcher the formatting of this
```
jerryaldrichiii, line 178: It's not looking so good rendered either:

2019-01-15-225555_713x391_scrot

This gets us closer, but Thor still doesn't like it.

```diff
    Exit codes:
      ```
      0:   Normal exit, all tests passed
      1:   Usage or general error
      2:   Error in plugin system
      3:   Fatal deprecation encountered
      100: Normal exit, at least one test failed
      101: Normal exit, at least one test skipped but none failed
      ```
```

2019-01-15-230358_714x503_scrot

Maybe we should ask @kagarmoe?

clintoncwolfe: @jerryaldrichiii thanks for checking the www version, I didn't think of that at all.  

I'm going to add the backticks to get it looking nice on the web, and open an issue to look into improving Thor formatting.

https://github.com/inspec/inspec/issues/3730

jerryaldrichiii (approved):
Thanks for responding to all my feedback! Looking good!

tyler-ball (approved):

PR #3691, opened 2018-12-26
iis_site: Add ability to check certificate hash on bindings

robwafle:
Work in progress, I believe I added the required tests.  Looking forward to feedback.  Very excited as this will be my first open source commit, ever.

I did not modify the existing has_binding method because it would have been a breaking change to include the certificate hash.   I figure if someone does want to check the certificate hash, they can go ahead and use the new method.

I also added the sign-off to the commit, and per CONTRIBUTING.md I incremented the minor version to 3.3.0 since I added a new feature.


VERSION
```diff
@@ -1 +1 @@
-3.2.7
\ No newline at end of file
+3.3.0
```
clintoncwolfe, line 1: Hey, thanks for the contribution!  I've re-read our CONTRIBUTING.md, and it's a bit misleading - in fact, our Continuous Integration system is in charge of all version increments, so this particular change should be reverted.

I've opened #3692 to update the CONTRIBUTING.md.

lib/resources/iis_site.rb
```diff
@@ -107,12 +117,17 @@ def iis_site(name)
         "#{k['protocol']} #{k['bindingInformation']}#{k['protocol'] == 'https' ? " sslFlags=#{k['sslFlags']}" : ''}"
       }
 
+      bindings_array_with_cert_hash = site['bindings']['Collection'].map { |k|
```
clintoncwolfe, line 120: Hrm.  'bindings_array' is really regrettable (and not your doing). It collapses a rich datastructure into a string, which people are now matching against. The field selection, formatting, and order now must never change lest they break other tests.  The fact that there are multiple bindings just makes things worse; the implementation of `has_binding?` is very brittle.

Which puts you, @robwafle, in an tough position - to add a field, it looks like the only thing to do might be to add a new property, which also stringifies, but with your new field added.  When someone else wants another field, they'd have to add it as a property too, but we'd also have to consider combinations....

I'm no IIS expert, but I think we could have better matchers here.  We could do this: 
 * Store the entire, un-stringified binding information in the `info` hash
 * implement a new matcher, `has_binding_with?(criteria = {})`

Then, `has_binding_with` could accept criteria, and you would return true if there was at lease one binding that met all the criteria.

Again, I'm no IIS expert, but from the examples it would be easy to see:
```
describe iis_site(...) do
  it { should have_binding_with(schema: :https, port: 443, hostname: 'www.contoso.com') }
  it { should have_binding_with(certificate_hash: '123456789sdfasdf', port: 443) }
end
```

This would allow people to test the cert hash if they wanted to.  We could add new criteria in the future without breaking existing users' code.  

What do you think?

clintoncwolfe (changes requested):
Thanks for your contribution!  A minor thing: we need to have the machines update the version, not the humans :-).  That's what makes me say "changes requested".

A bigger thing: I've proposed an alternate approach to unpacking and querying the bindings. That's not a requested change, but the opening of a discussion.  I'd be happy to help you out in trying out some ideas.

VERSION
robwafle: Thanks!  Will do!

lib/resources/iis_site.rb
robwafle: Yes.  I couldn't agree more.  My initial change is functional and was able to solve my immediate problem, but what you proposed is without a doubt leaps and bounds better.  I discussed this with another gentleman on the chef community slack #inspec channel and we were thinking the same.  I will update this PR to add the have_binding_with() method as above and remove the has_binding_with_cert() method.

aaronlippold: Glad we got this worked out :) 👍

robwafle:
Quick note for anyone who sees this PR:  I've been asked to put this on hold and I'm using a local inspec profile to override the iis_site resource within our own test kitchen.  Please comment if you are looking forward to this functionality as well.

clintoncwolfe:
As this has sat quietly for a while, and is not on the InSpec core team's roadmap, I'm going to close this.

Issue #3694, opened 2018-12-28
Remove personal repo references from tests


clintoncwolfe:
### 🎛 Description

🐛 bug report

Several tests refer to Git repos that are owned by individuals (generally test fixture profiles).  As we cannot guarantee their future availability, we should consolidate and move those to be in repos that are under the inspec/ Github org.

Seach for these (not a comprehensive list):
 * [ ] nathenharvey
 * [ ] chris-rock
 * [ ] jquick
 * [ ] clintoncwolfe
 * [ ] arlimus
 * [ ] adamleff
 
### 🌍 InSpec and Platform Version

3.2.4

### 💁 Possible Solutions

Consolidate if possible into the main inspec repo.  If not, make new, properly named repos under inspec org with a README.

Issue #3351, opened 2018-09-04
Stop using example profiles as test fixtures


clintoncwolfe:
### 🎛 Description

🐛 bug report

In places like [test/functional/helper.rb](https://github.com/inspec/inspec/blob/master/test/functional/helper.rb), we store paths to test fixtures (testing resources on the filesystem, typically).  Many of these fixtures point to profiles that are in [examples/](https://github.com/inspec/inspec/blob/master/examples), which is a set of profiles intended to be good teaching examples.

While the needs of testing and teaching may overlap somewhat, they also diverge significantly.  For example, if you add controls to an example, that may cause a test that was relying the count of controls to fail.  Such coupling is harmful.

### 🌍 InSpec and Platform Version

2.2.78

### 💁 Possible Solutions

Initially, copy the example profiles into the test fixture area.  Consider stripping them down to the essential testing need, and / or consolidating them.

clintoncwolfe:
Looks like we use things from Nathen's personal repo, too:

```
  1) Failure:
inspec exec::when using profiles on the supermarket#test_0002_can run supermarket profiles from inspec.yml [C:/projects/inspec/test/functional/inspec_exec_test.rb:359]:
Expected # encoding: UTF-8
"[2018-11-06T10:52:24+00:00] WARN: URL target https://github.com/nathenharvey/tmp_compliance_profile/ transformed to https://github.com/nathenharvey/tmp_compliance_profile/archive/master.tar.gz. Consider using the git fetcher\n" to include # encoding: UTF-8
"Profile Summary: \e[38;5;41m1 successful control\e[0m, \e[38;5;9m1 control failure\e[0m, 0 controls skipped\n".
```

clintoncwolfe:
Refs #596

clintoncwolfe:
Fixed by #3889

PR #3696, opened 2018-12-30
Stabilize profile export functional test

clintoncwolfe:
The profile export command, `inspec json`, has a functional test that verifies it works when targeting a profile with a dependency.  As written, the test pointed at a dependency maintained by the devsec project.  Whenever the dependent repo would change, the checksum would change, causing the lockfile to go stale and the test to fail.  [example](https://travis-ci.org/inspec/inspec/jobs/473136872#L1122) 

Since we don't care about fetching in this test, this PR changes the test to examine a local profile using the `path:` fetcher, so that the checksum will never go stale again.

Refs #3694, #3351

### DIFF
--- a/test/functional/inspec_json_profile_test.rb
+++ b/test/functional/inspec_json_profile_test.rb
@@ -131,17 +131,18 @@ describe 'inspec json' do
   end
 
   describe 'inspec json with a inheritance profile' do
-    let(:profile) { File.join(examples_path, 'meta-profile') }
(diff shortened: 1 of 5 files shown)


miah (approved):
I reran the travis tests and everything looks good. Thanks @clintoncwolfe !

clintoncwolfe:
As this is a testing-only change, and the tests are failing on most PRs, I'm going to go ahead and merge this so people can rebase.

PR #3700, opened 2019-01-03
processes: allow very wide arguments

caarlos0:
prevents output from getting cut if there are a lot of arguments...


caarlos0:
cc/ @clintoncwolfe

clintoncwolfe (approved):
Thanks for the contribution!

lib/resources/processes.rb
```diff
@@ -104,7 +104,7 @@ def ps_axo
           command: 12,
         }
       else
-        command = 'ps axo pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user,command'
+        command = 'ps -ww axo pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user,command'
```
lamont-granquist, line 107: if i'm reading this correctly, this line does not run on either linux or windows, which means this is a catch-all for other unix-like O/Sen and this command is now changing from strictly BSD-style of arguments to mixed SysV-BSD style which will quite probably break whichever O/Sen winds up on this path?

also makes no sense how inspec can support Solaris, HP-UX, etc when it looks like they wind up on this branch of code, though, so i may be missing something...

lamont-granquist (changes requested):

lib/resources/processes.rb
clintoncwolfe: There is a very good chance that the answer to "how is that working on SysV?" is "it doesn't".  We don't CI test on a SysV derivative, as far as I know - see https://github.com/inspec/inspec/blob/master/.kitchen.yml .

miah (approved):
@caarlos0 Thank you for this contribution!

lib/resources/processes.rb
miah: Since we have #3724 logging this issue we should punt for now and circle back in the future.

lamont-granquist: agreed.  this line should get dropped though.  this breaks MacOS:

% ps -ww xa
ps: illegal argument: xa
% ps -ww
  PID TTY           TIME CMD
[...]
% ps xa
  PID   TT  STAT      TIME COMMAND

miah: Maybe we can throw in another `elsif os.macos?` and define a functional ps for it then further refine that after we #3724 lands.

clintoncwolfe:
(boilerplate message)

lib/resources/processes.rb
zenspider: Poke

lamont-granquist: that line should probably just not be changed.  macos was only illustrative of the problem.  you have to deal with aix, solaris, bsd, mac, etc in this not-linux/not-windows codeblock.

it probably should not be changed in this diff, and then there needs to be an overhaul of the inspec functional testing and this needs to get greatly expanded.  there's no guarantee that this works side of the conditional works with mac right now either, it is just definitely broken on mac if you accept this change.

zenspider: We need to push on this PR as it is still a real problem.

`ps axwwo ...` works fine on linux and osx.

As far as other platforms go, I think they already have problems with this code so I'm not sure how it's relevant. SysV platforms aren't being branched here and should be using `ps -ef` (or `-F`) instead. I'm pretty sure they won't work with the current code (at least, not the way we think they will).

lamont-granquist: Right but `ps xaww` != `ps -ww xa`.

If you change the line here to `ps axwwo` that'd be fine, but it can't be merged like this.

caarlos0:
I don't have the fork anymore... feel free to pull the pr and continue from there :)

PR #3705, opened 2019-01-05
Pin mixlib-log to < 3.0.0

clintoncwolfe:
InSpec supports Ruby < 2.5 but `mixlib-log` no longer does.

This adds a pin to `mixlib-log`.

See: https://github.com/chef/mixlib-log/pull/47

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'rspec-its', '~> 1.2'
   spec.add_dependency 'pry', '~> 0'
   spec.add_dependency 'hashie', '~> 3.4'
-  spec.add_dependency 'mixlib-log'
+  spec.add_dependency 'mixlib-log', '< 3.0.0' # Mixlib does not support Ruby <2.5
   spec.add_dependency 'sslshake', '~> 1.2'
   spec.add_dependency 'parallel', '~> 1.9'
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@clintoncwolfe this can probably be closed now?  mixlib-log 3.0.1 supports ruby 2.4 which is likely the oldest ruby you can support.

users on ancient chef-12 less than 12.9.x or so can always use bundler and create a Gemfile which pins mixlib-log to < 3.0 -- that requires them to learn to use bundler, which i think is 100% reasonable.

clintoncwolfe:
We currently support Ruby 2.3, and I'm not sure we need to. We'll discuss.

lamont-granquist:
that's actually what i meant, mixlib-log supports ruby 2.3:

https://github.com/chef/mixlib-log/blob/master/mixlib-log.gemspec#L14

monday before coffee.

lamont-granquist:
also, to reiterate, this doesn't mean that inspec doesn't now support ruby 2.2 either -- it means everyone on ruby 2.2 needs to use a Gemfile which manually pins via `gem "mixlib-log", "< 3.0"` and it'll still work fine -- and that is only because those people are on such an old rubygems+bundler as well that dep resolution based on ruby version is broken.

miah (approved):

clintoncwolfe:
Based on the above discussion, I'm closing this.

PR #3706, opened 2019-01-07
inspec json CLI: Output control descriptions using labels

### LINKED ISSUE #3682: Control `descriptions` inconsistent json representation
### 🎛 Description

🐛 bug report 

Both profile json (from `inspec json`) and results json (from `inspec exec`) contain an array of controls. The JSON representation for a `description` of a control differs between the two.

3.0.52
macOS

### 🤔 Replication Case

Take any profile, and run `inspec json .`

The resulting descriptions for some control in the array of controls will look like:
```
"descriptions": {
	"default": "My description",
	"caveat": "My caveat"
},
```

Execute the profile using `inspec exec`, view results as json and observe any control in the array of controls:
```
"descriptions": [
	{
		"label": "default",
		"data": "My description"
	},
	{
		"label": "caveat",
		"data": "My caveat"
	}
],
```

samcornwell:
Fixes #3682

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -49,7 +49,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     o[:vendor_cache] = Inspec::Cache.new(o[:vendor_cache])
 
     profile = Inspec::Profile.for_target(target, o)
-    info = profile.info
+    info = profile.json
     # add in inspec version
     info[:generator] = {
       name: 'inspec',
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -325,6 +325,15 @@ module Inspec
       res
     end
 
+    def json
+      json = info
+      json[:controls].each do |c|
+        c[:descriptions] = Inspec::Reporters::Json.convert_descriptions(c[:descriptions])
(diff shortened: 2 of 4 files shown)


clintoncwolfe (changes requested):
After discussion internally at Chef, we have confirmed that changing the format of the output of `inspec json` will break other product like Chef Automate 2.

So, here is the proposed approach.

1. Add an additional commandline arg to `inspec json`, named `--schema-version`.  If absent, default to schema version 1.0.0, which is the current (pre #3706) output.
2. Also accept `--schema-version 1.1.0`.  In that case, produce output with the description fields broken out, as on #3706.
3. If version != 1.0.0, also add a top-level JSON key, 'schema_version', with the value of the schema version included.

I understand that may be more than you signed up for :-) when you submitted this pull request. A Chef engineer can adopt this PR if you'd like.

Thanks again!

clintoncwolfe:
(boilerplate message)

aaronlippold:
I will have The team update the comments and provide a timeline on when we think we'll be able to continue this. It may be the end of next week given we have customer deliverables we are focusing on for the beginning of this coming week. @samcornwell

PR #3708, opened 2019-01-07
postgres_session: Fix postgres directory searching on non-debian os

### LINKED ISSUE #3707: Postgres resource can't find data directory on non-debian systems
### 🎛 Description

🐛 bug report 
When running on a non-Debian installation, Postgres resource reports that it is unable to find...anything:
```
Unable to find the PostgreSQL data_dir in expected location(s), please
        execute "psql -t -A -p <port> -h <host> -c "show hba_file";" as the PostgreSQL
        DBA to find the non-standard data_dir location.
Default postgresql configuration directory:  does not exist. Postgresql may not be installed or we've misidentified the configuration directory.
Default postgresql data directory:  does not exist. Postgresql may not be installed or we've misidentified the data directory.
```

InSpec version: 3.2.6
Platform: OS X High Sierra

### 🤔 Replication Case

Run the following:
```
docker run --rm -d --name testdb postgres:alpine
inspec exec https://github.com/dev-sec/postgres-baseline -t docker://testdb
```

### 💁 Possible Solutions

`resources/postgres.rb` line 82 breaks the directory search loop in the first iteration, leaving `data_dir_loc` as `nil`.  This cascades to all subsequent location checks.

Change line 82 to conditionally break, e.g., 
```
        break if inspec.directory(dir).exist?
```
### 💻 Stacktrace

N/A

Cerebus:
Only break the postgres data directory search when the directory is found.  Fixes #3707

### DIFF
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -77,10 +77,7 @@ module Inspec::Resources
         '/var/lib/postgresql/data',
       ]
 
-      dir_list.each do |dir|
-        data_dir_loc = dir if inspec.directory(dir).exist?
-        break if inspec.directory(dir).exist?
-      end
+      data_dir_loc = dir_list.detect { |i| inspec.directory(i).exist? }
       
       if data_dir_loc.nil?
         warn 'Unable to find the PostgreSQL data_dir in expected location(s), please
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -81,7 +81,7 @@ module Inspec::Resources
         data_dir_loc = dir if inspec.directory(dir).exist?
         break if inspec.directory(dir).exist?
       end
-
+      
       if data_dir_loc.nil?
         warn 'Unable to find the PostgreSQL data_dir in expected location(s), please
         execute "psql -t -A -p <port> -h <host> -c "show hba_file";" as the PostgreSQL
--- a/lib/resources/postgres.rb
+++ b/lib/resources/postgres.rb
@@ -69,7 +69,6 @@ module Inspec::Resources
     end
 
     def locate_data_dir_location_by_version(ver = @version)
-      data_dir_loc = nil
       dir_list = [
         "/var/lib/pgsql/#{ver}/data",
         '/var/lib/pgsql/data',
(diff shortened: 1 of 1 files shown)


aaronlippold:
Hi, we need you to sign your contribution.
```
git co <my-branch>
git --amend -s
git --force push upstream master
```

clintoncwolfe:
Thank you for your contribution! However, before we can accept it, we ask that you sign your commits, to indicate that you are the author of the code and that it otherwise meets the terms of our [Developer Certification of Origin](https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) policy.

After you have read the DCO, please sign your code, as follows:
```
git commit --amend -s
git push -f
```

Thank you!

lib/resources/postgres.rb
```diff
@@ -79,7 +79,7 @@ def locate_data_dir_location_by_version(ver = @version)
 
       dir_list.each do |dir|
         data_dir_loc = dir if inspec.directory(dir).exist?
-        break
+        break if inspec.directory(dir).exist?
```
clintoncwolfe: I agree this needs to be a conditional, but I don't think we need to perform the same check as was performed on line 81 - that's another round-trip filesystem check.  What about

```
break if data_dir_loc
```

clintoncwolfe (changes requested):
I'd rather avoid the second directory check - should be straightforward.  Let me know if you need any help!

lib/resources/postgres.rb
Cerebus: I'm not a ruby guy so I had to poke around for an idiomatic version.  See next commit for a 1-line version.

clintoncwolfe: Hey, that is much prettier than what I suggested.  Thanks for your contribution!

clintoncwolfe (dismissed):

clintoncwolfe:
Ahh, linter isn't happy.

You can run the linter manually:

inspec-src $ bundle exec rubocop

clintoncwolfe (approved):
Thanks, and thanks for your patience with the linter!

PR #3709, opened 2019-01-08
[ROLLUP] Formalize config file and add credset support

### LINKED ISSUE #3661: Formalize InSpec config file handling

clintoncwolfe:
This PR:
  * Introduces `Inspec::Config`, which represents the configuration file
  * Adds a version field to the config file, and validates it when possible
  * Interacts with Train such that transport credentials are passes in more reliably without having to invent new URI schemes for each transport
  * Allows you to store as many credential sets as needed in the config file, with whatever options the transport needs
  * Unifies the URI scheme for the credset mechanism as `--target transport://credsetname`
  * Does its best to preserve functionality for Compliance and Automate, though those need to be tested manually
  * Allows you to use the more intuitive `--config` option.  `--json-config` remains supported but is now undocumented.

*NOTE 1*  This PR must be merged after https://github.com/inspec/train/pull/394 is merged and releases, and the train constraint adjusted accordingly.  A Gemfile entry for train pointing at the branch for 394 must also be removed.

*NOTE 2* This PR will likely be split into smaller pieces.

Closes #3661

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -75,8 +75,9 @@ module Inspec
         desc: 'Whether to use disable sspi authentication, defaults to false (WinRM).'
       option :winrm_basic_auth, type: :boolean,
         desc: 'Whether to use basic authentication, defaults to false (WinRM).'
-      option :json_config, type: :string,
+      option :config, type: :string,
         desc: 'Read configuration from JSON file (`-` reads from stdin).'
+      option :json_config, type: :string, hide: true
       option :proxy_command, type: :string,
         desc: 'Specifies the command to use to connect to the server'
       option :bastion_host, type: :string,
@@ -118,7 +119,7 @@ module Inspec
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
     end
 
-    def self.detect(params: {}, indent: 0, color: 39)
+    def self.format_platform_info(params: {}, indent: 0, color: 39)
       str = ''
       params.each { |item, info|
         data = info
@@ -254,8 +255,8 @@ module Inspec
       # logging singleton Inspec::Log. Eventually it would be nice to
       # move internal debug logging to use this logging singleton.
       #
-      loc = if o.log_location
-              o.log_location
+      loc = if o['log_location']
+              o['log_location']
             elsif suppress_log_output?(o)
               STDERR
             else
@@ -263,14 +264,14 @@ module Inspec
             end
 
       Inspec::Log.init(loc)
-      Inspec::Log.level = get_log_level(o.log_level)
+      Inspec::Log.level = get_log_level(o['log_level'])
 
       o[:logger] = Logger.new(loc)
       # output json if we have activated the json formatter
       if o['log-format'] == 'json'
         o[:logger].formatter = Logger::JSONFormatter.new
       end
-      o[:logger].level = get_log_level(o.log_level)
+      o[:logger].level = get_log_level(o['log_level'])
     end
   end
 end
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -118,91 +118,6 @@ module Inspec
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
     end
 
-    def self.default_options
-      {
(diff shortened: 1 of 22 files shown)


lib/inspec/base_cli.rb
```diff
@@ -409,27 +251,27 @@ def vendor_deps(path, opts)
 
     def configure_logger(o)
```
Vasu1105, line 252: may be we can also replace `o` with more descriptive argument name ?

lib/inspec/cli.rb
```diff
@@ -39,12 +40,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: 'A list of controls to include. Ignore all other tests.'
   profile_options
   def json(target)
-    o = opts.dup
+    o = config
```
Vasu1105, line 43: here to ^

lib/inspec/cli.rb
```diff
@@ -39,12 +40,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
...
     diagnose(o)
     o['log_location'] = STDERR
```
Vasu1105, line 45: I am not sure but in some we are using key as symbols and here we are using string. Can we use symbol here too ?

lib/inspec/base_cli.rb
clintoncwolfe: We're currently working with a skeleton crew for PR reviews, so we're being careful not to introduce changes that would drive up line counts.  This PR does not touch the line you mention, so it neither improves nor worsens the situation.  I'd rather change it as well.

We would certainly welcome a separate PR in which you make any readability and clarity changes you see fit.

lib/inspec/cli.rb
clintoncwolfe: I agree, it's a problem throughout InSpec and Train.  Inspec::Config uses a HashWithIndifferentAccess class provided by Thor to allow InSpec and Train to continue accessing it using either approach.

The original authors of InSpec were not Ruby programmers, and used a mix.  If you want to fix this we'd welcome it, but it would require very thorough testing.

clintoncwolfe:
I'm closing this in favor of the three split our PR's; we can just consider them on an individual basis.

PR #3710, opened 2019-01-08
Config file formalization and credential sets: part 1, Config Class

clintoncwolfe:
This is part 1 of #3709.  It is part of a series to address #3661.

This PR implements a class, Inspec::Config, which represents the config file.  It is responsible for many tasks that were formerly in BaseCLI:

 * Providing per-command defaults
 * Merging defaults, config file, and CLI options
 * Parsing and validating formatter options
 * Performing other miscellaneous option finalization

Much of this class's code was copied from lib/inspec/base_cli.rb .  It will be removed in part 2 of the PR series.

This PR has a no prerequisites and may be merged directly to master, or to a rollup branch.

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -1,0 +1,286 @@
+# Represents InSpec configuration.  Merges defaults, config file options,
+# and CLI arguments.
+
+require 'pp'
+
+module Inspec
+  class Config
+    # These are options that apply to any transport
+    GENERIC_CREDENTIALS = %w{
+      backend
+      sudo
+      sudo_password
+      sudo_command
+      sudo_options
+      shell
+      shell_options
+      shell_command
+    }.freeze
+
+    extend Forwardable
+
+    # Many parts of InSpec expect to treat the Config as a Hash
+    def_delegators :@final_options, :each, :delete, :[], :[]=, :key?
+    attr_reader :final_options
+
+    def initialize(cli_opts = {}, cfg_io = nil, command_name = nil)
+      @command_name = command_name || (ARGV.empty? ? nil : ARGV[0].to_sym)
+      @defaults = Defaults.for_command(@command_name)
+
+      @cli_opts = cli_opts.dup
+      cfg_io = resolve_cfg_io(@cli_opts, cfg_io)
+      @cfg_file_contents = read_cfg_file_io(cfg_io)
+
+      @merged_options = merge_options
+      @final_options = finalize_options
+    end
+
+    def diagnose
+      return unless self[:diagnose]
+      puts "InSpec version: #{Inspec::VERSION}"
+      puts "Train version: #{Train::VERSION}"
+      puts 'Command line configuration:'
+      pp @cli_opts
+      puts 'JSON configuration file:'
+      pp @cfg_file_contents
+      puts 'Merged configuration:'
+      pp @merged_options
+      puts
+    end
+
+    #-----------------------------------------------------------------------#
+    #                         Reading Config Files
+    #-----------------------------------------------------------------------#
+
+    # Regardless of our situation, end up with a readable IO object
+    def resolve_cfg_io(cli_opts, cfg_io)
+      unless cfg_io
+        # TODO: deprecate --json-config
+        path = cli_opts[:config] || cli_opts[:json_config]
+        if path == '-'
+          Inspec::Log.warn 'Reading JSON config from standard input' if STDIN.tty?
+          path = STDIN
+        elsif path.nil?
+          default_path = File.join(Inspec.config_dir, 'config.json')
+          path = default_path if File.exist?(default_path)
+        elsif !File.exist?(path)
+          raise ArgumentError, "Could not read configuration file at #{path}"
+        end
+        cfg_io = File.open(path) if path
+        cfg_io ||= StringIO.new('{ "version": "1.1" }')
+      end
+      cfg_io
+    end
+
+    def read_cfg_file_io(cfg_io)
+      contents = cfg_io.read
+      begin
+        @cfg_file_contents = JSON.parse(contents)
+        validate_config_file_contents!
+      rescue JSON::ParserError => e
+        raise Inspec::ConfigError::MalformedJson, "Failed to load JSON configuration: #{e}\nConfig was: #{contents}"
+      end
+      @cfg_file_contents
+    end
+
+    def file_version
+      @cfg_file_contents['version'] || :legacy
+    end
+
+    def legacy_file?
+      file_version == :legacy
+    end
+
+    def config_file_cli_options
+      if legacy_file?
+        # Assume everything in the file is a CLI option
+        @cfg_file_contents
+      else
+        @cfg_file_contents['cli_options'] || {}
+      end
+    end
+
+    def config_file_reporter_options
+      # This is assumed to be top-level in both legacy and 1.1.
+      # Technically, you could sneak it in the 1.1 cli opts area.
+      @cfg_file_contents.key?('reporter') ? { 'reporter' => @cfg_file_contents['reporter'] } : {}
+    end
+
+    #-----------------------------------------------------------------------#
+    #                            Validation
+    #-----------------------------------------------------------------------#
+    def validate_config_file_contents!
+      version = @cfg_file_contents['version']
+
+      # Assume legacy format, which is unconstrained
+      return unless version
+
+      unless version == '1.1'
+        raise Inspec::ConfigError::Invalid, "Unsupported config file version '#{version}' - currently supported versions: 1.1"
+      end
+
+      valid_fields = %w{version cli_options credentials compliance reporter}.sort
+      @cfg_file_contents.keys.each do |seen_field|
+        unless valid_fields.include?(seen_field)
+          raise Inspec::ConfigError::Invalid, "Unrecognized top-level configuration field #{seen_field}.  Recognized fields: #{valid_fields.join(', ')}"
+        end
+      end
+    end
+
+    def validate_reporters(reporters)
+      return if reporters.nil?
+      # TODO: move this into a reporter plugin type system
+      valid_types = [
+        'automate',
+        'cli',
+        'documentation',
+        'html',
+        'json',
+        'json-automate',
+        'json-min',
+        'json-rspec',
+        'junit',
+        'progress',
+        'yaml',
+      ]
+
+      reporters.each do |reporter_name, reporter_config|
+        raise NotImplementedError, "'#{reporter_name}' is not a valid reporter type." unless valid_types.include?(reporter_name)
+
+        next unless reporter_name == 'automate'
+        %w{token url}.each do |option|
+          raise Inspec::ReporterError, "You must specify a automate #{option} via the config file." if reporter_config[option].nil?
+        end
+      end
+
+      # check to make sure we are only reporting one type to stdout
+      stdout_reporters = 0
+      reporters.each_value do |reporter_config|
+        stdout_reporters += 1 if reporter_config['stdout'] == true
+      end
+
+      raise ArgumentError, 'The option --reporter can only have a single report outputting to stdout.' if stdout_reporters > 1
+    end
+
+    #-----------------------------------------------------------------------#
+    #                         Merging Options
+    #-----------------------------------------------------------------------#
+    def merge_options
+      options = Thor::CoreExt::HashWithIndifferentAccess.new({})
+
+      # Lowest precedence: default, which may vary by command
+      options.merge!(@defaults)
+
+      # Middle precedence: merge in any CLI options defined from the config file
+      options.merge!(config_file_cli_options)
+      # Reporter options may be defined top-level.
+      options.merge!(config_file_reporter_options)
+
+      # Highest precedence: merge in any options defined via the CLI
+      options.merge!(@cli_opts)
+
+      options
+    end
+
+    #-----------------------------------------------------------------------#
+    #                          Finalization
+    #-----------------------------------------------------------------------#
+    def finalize_options
+      options = @merged_options.dup
+
+      finalize_set_top_level_command(options)
+      finalize_parse_reporters(options)
+      finalize_handle_sudo(options)
+      finalize_compliance_login(options)
+
+      Thor::CoreExt::HashWithIndifferentAccess.new(options)
+    end
+
+    def finalize_set_top_level_command(options)
+      options[:type] = @command_name
+      Inspec::BaseCLI.inspec_cli_command = @command_name # TODO: move to a more relevant location
+    end
+
+    def finalize_parse_reporters(options) # rubocop:disable Metrics/AbcSize
+      # default to cli report for ad-hoc runners
+      options['reporter'] = ['cli'] if options['reporter'].nil?
+
+      # parse out cli to proper report format
+      if options['reporter'].is_a?(Array)
+        reports = {}
+        options['reporter'].each do |report|
+          reporter_name, destination = report.split(':', 2)
+          if destination.nil? || destination.strip == '-'
+            reports[reporter_name] = { 'stdout' => true }
+          else
+            reports[reporter_name] = {
+              'file' => destination,
+              'stdout' => false,
+            }
+            reports[reporter_name]['target_id'] = options['target_id'] if options['target_id']
+          end
+        end
+        options['reporter'] = reports
+      end
+
+      # add in stdout if not specified
+      if options['reporter'].is_a?(Hash)
+        options['reporter'].each do |reporter_name, config|
+          options['reporter'][reporter_name] = {} if config.nil?
+          options['reporter'][reporter_name]['stdout'] = true if options['reporter'][reporter_name].empty?
+          options['reporter'][reporter_name]['target_id'] = options['target_id'] if options['target_id']
+        end
+      end
+
+      validate_reporters(options['reporter'])
+      options
+    end
+
+    def finalize_handle_sudo(options)
+      # Due to limitations in Thor it is not possible to set an argument to be
+      # both optional and its value to be mandatory. E.g. the user supplying
+      # the --password argument is optional and not always required, but
+      # whenever it is used, it requires a value. Handle options that were
+      # defined in such a way and require a value here:
+      %w{password sudo-password}.each do |option_name|
+        snake_case_option_name = option_name.tr('-', '_').to_s
+        next unless options[snake_case_option_name] == -1 # Dubious - does thor always set -1 for missing value?
+        raise ArgumentError, "Please provide a value for --#{option_name}. For example: --#{option_name}=hello."
+      end
+
+      # Infer `--sudo` if using `--sudo-password` without `--sudo`
+      if options['sudo_password'] && !options['sudo']
+        options['sudo'] = true
+        Inspec::Log.warn '`--sudo-password` used without `--sudo`. Adding `--sudo`.'
+      end
+    end
+
+    def finalize_compliance_login(options)
+      # check for compliance settings
+      # This is always a hash, comes from config file, not CLI opts
+      if options.key?('compliance')
+        require 'plugins/inspec-compliance/lib/inspec-compliance/api'
+        InspecPlugins::Compliance::API.login(options['compliance'])
+      end
+    end
+
+    class Defaults
+      DEFAULTS = {
+        exec: {
+          'reporter' => ['cli'],
+          'show_progress' => false,
+          'color' => true,
+          'create_lockfile' => true,
+          'backend_cache' => true,
+        },
+        shell: {
+          'reporter' => ['cli'],
+        },
+      }.freeze
+
+      def self.for_command(command_name)
+        DEFAULTS[command_name] || {}
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+      pp @merged_options
+      puts
+    end
+
+    #-----------------------------------------------------------------------#
+    #                         Reading Config Files
+    #-----------------------------------------------------------------------#
+
+    # Regardless of our situation, end up with a readable IO object
+    def resolve_cfg_io(cli_opts, cfg_io)
```
jerryaldrichiii, line 56: Are there tests for the paths through this? Also, should we validate that the `cfg_io` passed is a `StringIO`?

NOTE: I'm also curious why not to use a guard clause. I suspect this is a preference thing, if it is could you help me learn, offline or here, your choice.

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+      unless cfg_io
+        # TODO: deprecate --json-config
```
jerryaldrichiii: Is there an issue to track this? This feels like there is more context here than this TODO line indicates.

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+        path = cli_opts[:config] || cli_opts[:json_config]
+        if path == '-'
+          Inspec::Log.warn 'Reading JSON config from standard input' if STDIN.tty?
```
jerryaldrichiii: Is STDIN being a TTY required? I'm wondering if this would affect users using SSH commands as a job runner (which may or not require a TTY).

```
ssh somehost 'inspec exec someprofile --config -' <<-EOF
{
  INSERT JSON HERE THAT MAY USE ENV VARS ON SSH HOST
}
EOF
```

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+    def config_file_reporter_options
+      # This is assumed to be top-level in both legacy and 1.1.
+      # Technically, you could sneak it in the 1.1 cli opts area.
+      @cfg_file_contents.key?('reporter') ? { 'reporter' => @cfg_file_contents['reporter'] } : {}
+    end
+
+    #-----------------------------------------------------------------------#
+    #                            Validation
+    #-----------------------------------------------------------------------#
+    def validate_config_file_contents!
```
jerryaldrichiii, line 118: I'm confused why this is a `!` method, it doesn't seem to modify any state.

test/unit/config_test.rb
```diff
@@ -0,0 +1,417 @@
...
+      end
+    end
+
+    describe 'when the file has malformed json' do
+      let(:fixture_name) { 'malformed_json' }
+      it 'should throw an exception' do
+        ex = proc { cfg }.must_raise(Inspec::ConfigError::MalformedJson)
+        ex.message.must_include 'Failed to load JSON config'
+        ex.message.must_include 'unexpected token'
+        ex.message.must_include 'version'
```
jerryaldrichiii: Took me a while to understand where `'version'` came from. Could we change it to:

```
ex.message.must_match /Config was:.*version.*$/
```

Not necessary, but would have made it easier for me at a glance.

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+      version = @cfg_file_contents['version']
+
+      # Assume legacy format, which is unconstrained
+      return unless version
+
+      unless version == '1.1'
+        raise Inspec::ConfigError::Invalid, "Unsupported config file version '#{version}' - currently supported versions: 1.1"
+      end
+
+      valid_fields = %w{version cli_options credentials compliance reporter}.sort
```
jerryaldrichiii, line 128: What's the reason for the `sort` here?

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+      @cfg_file_contents.keys.each do |seen_field|
+        unless valid_fields.include?(seen_field)
+          raise Inspec::ConfigError::Invalid, "Unrecognized top-level configuration field #{seen_field}.  Recognized fields: #{valid_fields.join(', ')}"
+        end
+      end
+    end
+
+    def validate_reporters(reporters)
```
jerryaldrichiii: I'm curious how the schema will be validated in future PRs, right now I don't see a way to avoid

```
reporter: "json"
```

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+
+      Thor::CoreExt::HashWithIndifferentAccess.new(options)
+    end
+
+    def finalize_set_top_level_command(options)
+      options[:type] = @command_name
+      Inspec::BaseCLI.inspec_cli_command = @command_name # TODO: move to a more relevant location
+    end
+
+    def finalize_parse_reporters(options) # rubocop:disable Metrics/AbcSize
```
jerryaldrichiii, line 210: There's a lot going on here, I think this is covered end-to-end in your unit tests, but I'd like to see this method isolated and tests created for it for the sake of regression prevention.

I also feel you could split this into a few methods based on Hash or Array which might make it a bit easier to digest (and avoid having to disable the AbcSize cop)

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+
+    def finalize_handle_sudo(options)
+      # Due to limitations in Thor it is not possible to set an argument to be
+      # both optional and its value to be mandatory. E.g. the user supplying
+      # the --password argument is optional and not always required, but
+      # whenever it is used, it requires a value. Handle options that were
+      # defined in such a way and require a value here:
+      %w{password sudo-password}.each do |option_name|
+        snake_case_option_name = option_name.tr('-', '_').to_s
+        next unless options[snake_case_option_name] == -1 # Dubious - does thor always set -1 for missing value?
```
jerryaldrichiii: This one goes a bit back, context can be found here:

https://github.com/inspec/inspec/pull/1918

https://github.com/inspec/inspec/pull/1918/files#diff-2b20044866abd29a7a781f5760238403R30

I think we should link it or summarize it here.

test/unit/config_test.rb
```diff
@@ -0,0 +1,417 @@
...
+    let(:cfg_io) { StringIO.new(ConfigTestHelper.fixture(fixture_name)) }
+    let(:seen_fields) { cfg.final_options.keys.sort }
+
+    # These two test cases have the same options but in different file versions.
+    describe 'when the CLI opts are present in a 1.1 file' do
+      let(:fixture_name) { :like_legacy }
+      it 'should read the options' do
+        expected = ['color', 'reporter', 'target_id', 'type'].sort
+        seen_fields.must_equal expected
+        final_options['color'].must_equal "true"  # Dubious
```
jerryaldrichiii: Dubious how so? Could you elaborate in the comment?

test/unit/config_test.rb
```diff
@@ -0,0 +1,417 @@
...
+  end
+
+  # ========================================================================== #
+  #                      Miscellaneous Option Finalization
+  # ========================================================================== #
+
+  describe 'option finalization' do
+    it 'raises if `--password/--sudo-password` are used without value' do
+      # When you invoke `inspec shell --password`  (with no value for password,
+      # though it is setup to expect a string) Thor will set the key with value -1
```
jerryaldrichiii, line 252: See other comment about context and 

https://github.com/inspec/inspec/pull/1918/files#diff-2b20044866abd29a7a781f5760238403R30

test/unit/config_test.rb
```diff
@@ -0,0 +1,417 @@
+
```
jerryaldrichiii, line 1: I'm holding off on the `chefstyle/rubocop -a` PR until the final product, but am still willing to do the work premerge as it takes me very little time. I'll avoid any style issues in this PR for this reason.

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+    def file_version
+      @cfg_file_contents['version'] || :legacy
+    end
+
+    def legacy_file?
+      file_version == :legacy
+    end
+
+    def config_file_cli_options
+      if legacy_file?
```
jerryaldrichiii, line 101: I don't see where `file_version` or `legacy_file` are used anywhere else. 
Any reason we shouldn't remove those methods and change this to?:

```
# Assume legacy file if no version is specified
if !@cfg_file_contents['version']
```

lib/inspec/config.rb
```diff
@@ -0,0 +1,286 @@
...
+
+      @cli_opts = cli_opts.dup
+      cfg_io = resolve_cfg_io(@cli_opts, cfg_io)
+      @cfg_file_contents = read_cfg_file_io(cfg_io)
+
+      @merged_options = merge_options
+      @final_options = finalize_options
+    end
+
+    def diagnose
```
jerryaldrichiii, line 38: `pp` is fine here, but I'm wonder if the work you did with UI would prevent the need for it.

jerryaldrichiii (commented):
Gave this a look through, but I'm hesitant to give this an Approve/Reject.

Specifically, I'm having trouble following the split context across PRs and am not sure I can hold the whole plan/direction in my head. I might need to look at the functional tests/docs changes when they are added to help understand it. With that in mind, I would like us use a feature branch and then have a final review of the end product.

Of course, my trepidation might also resolved when we do a group review.

lib/inspec/config.rb
clintoncwolfe: `!` is also often used to mark methods that are expected to generate exceptions.

test/unit/config_test.rb
clintoncwolfe: When we read a JSON file, should we expect the String "true" to be returned as a String, or as `true`, an instance of TrueClass?

lib/inspec/config.rb
clintoncwolfe: There is nothing in the Inspec::UI work that deals with pretty-printing deep structures.

lib/inspec/config.rb
clintoncwolfe: No, I just personally always felt that `--json-config` was a terrible name for 'the option that tells you where the config file is'.  

I can add it to #3661.

Note that `--json-config` is still supported, it just isn't documented. We could make that more explicit.

Also note that any comments like `TODO: deprecate...` are intended to be replaced by proper deprecations once #3690 lands.

lib/inspec/config.rb
clintoncwolfe: We don't expect cfg_io to be a StringIO - it can be anything IO-like. In actual use that's likely to be an IO::File, while in testing it is easier to use a StringIO.  If there is no file, then we use a StringIO to treat it in the same way.

clintoncwolfe: Guard clauses are used to short-circuit logic, usually a break, next, or return.  Use them you have things like `return true if condition...`.

In this case, there is a lot of logic on either side of the unless.  That's not a situation for a guard clause.

lib/inspec/config.rb
clintoncwolfe: As mentioned in the description, most of the code was literally copied from base_cli.rb . I don't vouch for the logic of checking for a TTY, but it is how inspec currently behaves:

https://github.com/inspec/inspec/blob/feefdcf9afba80f4e666e6dad5ffce6bd30019dd/lib/inspec/base_cli.rb#L359

test/unit/config_test.rb
clintoncwolfe: I greatly prefer to check for individual elements of an UX message on an element-by-element basis. Rather than altering the test to only match on one regex (that could fail for any number of reasons, and will only grow in complexity over time), checking for one thing at a time make it obvious which piece is missing.

A comment with example output would be informative though - I'll add that.

lib/inspec/config.rb
clintoncwolfe: Honestly not sure!  I think so the user sees a nice sorted list in the 'valid entries' in the error message? I could certainly sort in source.

lib/inspec/config.rb
clintoncwolfe: I'm not sure I follow your concern.  There is no intended change to the logic of parsing or validating the reporters.

lib/inspec/config.rb
clintoncwolfe: This is pretty much directly copied out of base_cli.rb. I don't claim to make any changes to the code, nor improve the testing on it.

https://github.com/inspec/inspec/blob/feefdcf9afba80f4e666e6dad5ffce6bd30019dd/lib/inspec/base_cli.rb#L136

lib/inspec/config.rb
clintoncwolfe: A big point of this PR series is to formalize the config file format, and get it into a shape where can version the format, and validate it.

Having the Config object be able to report on the detected version is thus key to the intent of the PR.

lib/inspec/config.rb
jerryaldrichiii: I don't guess I'm following, I don't see any logic surrounding unless, couldn't the first line be:

```
return cfg_io if cfg_io
```

jerryaldrichiii: I do think we need some sort of type checking on `cfg_io` as well too, otherwise anything passed into `initialize`, that doesn't respond to `.read` will fail here: 

https://github.com/inspec/inspec/blob/09e0579f5c34b642de939dbfb54aaf220249bf03/lib/inspec/config.rb#L76

lib/inspec/config.rb
jerryaldrichiii: Eh, that's a lot of load on the developer `.sort` is fine here, CPU cycles are cheap :smile:

lib/inspec/config.rb
jerryaldrichiii: Ah, I was thinking about the Boy Scout Rule you've mentioned in the past. Does it only apply to small things (I recognize adding tests for this would be significant work...at least for me)

lib/inspec/config.rb
jerryaldrichiii: I was concerned if someone had an invalid config (though valid JSON). For example if someone used just `reporter: "json"` then https://github.com/inspec/inspec/blob/09e0579f5c34b642de939dbfb54aaf220249bf03/lib/inspec/config.rb#L147 would fail with a stacktrace.

test/unit/config_test.rb
jerryaldrichiii: Ah, could we elaborate in the comment?

lib/inspec/config.rb
jerryaldrichiii: Given https://github.com/inspec/inspec/pull/3710/files/09e0579f5c34b642de939dbfb54aaf220249bf03#r248157588 shouldn't this method also be a `!` method?

test/unit/config_test.rb
clintoncwolfe: This is just copied from the base_cli_test.rb

https://github.com/inspec/inspec/blob/feefdcf9afba80f4e666e6dad5ffce6bd30019dd/test/unit/base_cli_test.rb#L11

lib/inspec/config.rb
clintoncwolfe: Good idea!  This was just copied from base_cli.rb, and didn't follow that pattern.

lib/inspec/config.rb
clintoncwolfe: I personally don't see that (passing a non-IO object) as likely - this is an internal InSpec object - but it's no big deal to add a duck check.

clintoncwolfe: And regarding the guard clause: you are totally right, I just didn't read down far enough.  This is an ideal case for a guard clause.

lib/inspec/config.rb
clintoncwolfe: So, when Inspec::Config finds a problem, it throws an exception.  Things that consume Inspec::Config (none of which are included on this PR) are then responsible for catching exceptions and handling them - which might mean reporting a graceful error to the user, with no stacktrace.

The existing code throws exceptions, as well - see https://github.com/inspec/inspec/blob/feefdcf9afba80f4e666e6dad5ffce6bd30019dd/lib/inspec/base_cli.rb#L171 

The existing code calls validate_reporters through a long chain, eventually surfacing as the `opts` call which is used as one of the first things in nearly every command method.  And each command method provides a facility for handling exceptions - for example, here is `inspec exec`'s exception handler block:

https://github.com/inspec/inspec/blob/master/lib/inspec/cli.rb#L252

lib/inspec/config.rb
clintoncwolfe: It's a rule of thumb, I guess, and relies on intuition.

The goal of this PR series is to get the config reading and validation into a proper class, and then expose a simpler way of reading credentials. So, I must copy a lot of code from where it was (scattered as class methods in Inspec::BaseCLI) to its new home. (On the 2nd PR of the series, I delete it from its previous location - which perhaps I should have done on this PR). In order to help people follow the changes, I'm just doing simple copying as much as possible. Making improvements (like adding tests, renaming vars, etc) would add noise to the PR.

We could improve testing of the reporter parser on a later PR, either on this series, or treat that as a separate issue - I think that might take some real effort to really exercise that facility.  And the problem with adding tests, you know... you may find bugs, which then you must fix! Which makes me thing adding tests for the reporter parser is probably best as a separate issue.

test/unit/config_test.rb
```diff
@@ -0,0 +1,420 @@
...
+    describe 'when the file has malformed json' do
+      let(:fixture_name) { 'malformed_json' }
+      it 'should throw an exception' do
+        ex = proc { cfg }.must_raise(Inspec::ConfigError::MalformedJson)
+        # Failed to load JSON configuration: 765: unexpected token at '{ "hot_garbage": "a", "version": "1.1",
+        # '
+        # Config was: "{ \"hot_garbage\": \"a\", \"version\": \"1.1\", \n"
+        ex.message.must_include 'Failed to load JSON config'  # The message
+        ex.message.must_include 'unexpected token'  # The specific parser error
+        ex.message.must_include 'hot_garbage' # A sample of the unacceptable contents
```
robbkidd, line 72: I appreciate your example data. 

I see you.

lib/inspec/config.rb
```diff
@@ -0,0 +1,292 @@
...
+
+      cfg_io = File.open(path) if path
+      cfg_io || StringIO.new('{ "version": "1.1" }')
+    end
+
+    def determine_cfg_path(cli_opts)
+      path = cli_opts[:config] || cli_opts[:json_config] # TODO: deprecate --json-config see #3661
+
+      if path == '-'
+        Inspec::Log.warn 'Reading JSON config from standard input' if STDIN.tty?
```
marcparadise, line 70: Is reading config from stdin a new behavior, or something that was previously supported?

lib/inspec/config.rb
```diff
@@ -0,0 +1,292 @@
...
+        next unless reporter_name == 'automate'
+        %w{token url}.each do |option|
+          raise Inspec::ReporterError, "You must specify a automate #{option} via the config file." if reporter_config[option].nil?
+        end
+      end
+
+      # check to make sure we are only reporting one type to stdout
+      stdout_reporters = 0
+      reporters.each_value do |reporter_config|
+        stdout_reporters += 1 if reporter_config['stdout'] == true
```
marcparadise, line 165: Can we move this into the loop at line 153, and avoid the second loop?

lib/inspec/config.rb
```diff
@@ -0,0 +1,292 @@
...
+    #-----------------------------------------------------------------------#
+    def merge_options
+      options = Thor::CoreExt::HashWithIndifferentAccess.new({})
+
+      # Lowest precedence: default, which may vary by command
+      options.merge!(@defaults)
+
+      # Middle precedence: merge in any CLI options defined from the config file
+      options.merge!(config_file_cli_options)
+      # Reporter options may be defined top-level.
```
marcparadise, line 182: Making sure I understand this - this was due to supporting it in previous versions in different locations/'cli', but it's now required under `reporting` at the top level?

lib/inspec/config.rb
```diff
@@ -0,0 +1,292 @@
...
+      Thor::CoreExt::HashWithIndifferentAccess.new(options)
+    end
+
+    def finalize_set_top_level_command(options)
+      options[:type] = @command_name
+      Inspec::BaseCLI.inspec_cli_command = @command_name # TODO: move to a more relevant location
+    end
+
+    def finalize_parse_reporters(options) # rubocop:disable Metrics/AbcSize
+      # default to cli report for ad-hoc runners
```
marcparadise, line 211: To simplify testing, would it be worthwhile to break this into two separate impl methods, `finalize_parse_reporters_array` and `finalize_parse_reporters_hash`? I know that it's existing code, but this might be a good opportunity to clean it a bit (unless that's planned/done in subsequent PRs?)

marcparadise (changes requested):
This looks like it will make config/cli handling quite a bit easier to reason about.   A few questions and a couple of comments inline.

lib/inspec/config.rb
clintoncwolfe: It was previously supported - most of this is copied from lib/inspec/base_cli.rb .

clintoncwolfe:
As this PR breakout is confusing everyone, I'm withdrawing this and will re-open in a different way.

PR #3713, opened 2019-01-09
Config file formalization and credential sets: part 3, credential sets

clintoncwolfe:
This is the last part of #3661, and a follow-on to #3750.

This PR adds the ability to add named sets of credentials (or more generally speaking, any options that a Train transport would accept) to the InSpec config file.

```
...
"credentials": {
  "ssh": {
    "myset1": {
      "user":"bob",
      "keepalive": true, 
      "hostname": "myhost",
    }
  }
}
...
```

The `--target` option is altered to allow a user to specify a credential set to use by `ssh://myset1` (more generally, `transport://credsetname`).

Notes on operation:
 * If a matching credential set can't be found, Train is asked to parse the `--target` value as a URI, preserving the previous behavior.
 * You can use a credential set, then override individual options on the CLI.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -42,18 +42,14 @@ Config files must contain a top-level key, `version`, which indicates the file f
 }
 ```
 
+version
+: Should have the value '1.1'.
 
-### version
+cli_options
+: Any long-form command line option, without the leading dashes.
 
-Should have the value '1.1'.
-
-### cli_options
-
-Any long-form command line option, without the leading dashes.
-
-### credentials
-
-Train-transport-specific options. Store the options keyed first by transport name, then by a name you'll use later on. The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
+credentials
+: Train-transport-specific options. Store the options keyed first by transport name, then by a name you'll use later on. The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
 
 For example, if the config file contains:
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -49,11 +49,11 @@ Should have the value '1.1'.
 
 ### cli_options
 
-In this key, you can place any long-form command line option, without the leading dashes.
+Any long-form command line option, without the leading dashes.
 
 ### credentials
 
-Under this key, you may store any Train-transport-specific options.  You store the options keyed first by transport name, then by a name you choose to refer to them later.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
+Train-transport-specific options. Store the options keyed first by transport name, then by a name you'll use later on. The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
 
 For example, if the config file contains:
 
@@ -72,7 +72,7 @@ For example, if the config file contains:
 }
 ```
 
-Then you can use `--target winrm://myconn` to connect to the host, with the given extra options.
+Then use `-t winrm://myconn` to connect to the host, with the given extra options.
 
 Each Train transport offers a variety of options. By using the credential set facility, you are able to easily set options that are not accessible via the Train URI.
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,0 +1,83 @@
+# The InSpec Configuration File
+
+This documents the InSpec configuration file format introduced in version 3.5 of InSpec.
+
+## Config File Location
+
+By default, InSpec looks for a config file in `~/.inspec/config.json`.  InSpec does not need a configuration file to run.
+
+You may also specify the location using `--config`.  For example, to run the shell using a config file in `/etc/inspec`, use `inspec shell --config /etc/inspec/config.json`.
+
+## Config File Format Versions
+
+Config files must contain a top-level key, `version`, which indicates the file format. This allows us to add new fields without breaking old installations.
+
+## Version 1.1
+
+### Complete Example
+
+```
+{
+  "version": "1.1",
+  "cli_options":{
+    "color": "true"
+  },
+  "credentials": {
+    "ssh": {
+      "my-target": {
+        "host":"somewhere.there.com",
+        "user":"bob"
+      }
+    }
+  },
+  "reporter": {
+    "automate" : {
+    "stdout" : false,
+    "url" : "https://YOUR_A2_URL/data-collector/v0/",
+    "token" : "YOUR_A2_ADMIN_TOKEN",
+    "insecure" : true,
+    "node_name" : "inspec_test_node",
+    "environment" : "prod"
+  }
+}
+```
+
+
+### version
+
+Should have the value '1.1'.
+
+### cli_options
+
+In this key, you can place any long-form command line option, without the leading dashes.
+
+### credentials
+
+Under this key, you may store any Train-transport-specific options.  You store the options keyed first by transport name, then by a name you choose to refer to them later.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
+
+For example, if the config file contains:
+
+```
+{
+  "credentials": {
+    "winrm": {
+      "myconn": {
+        "user": "Administrator",
+        "host": "prod01.east.example.com",
+        "disable_sspi": true,
+        "connection_retries": 10
+      }
+    }
+  }
+}
+```
+
+Then you can use `--target winrm://myconn` to connect to the host, with the given extra options.
+
+Each Train transport offers a variety of options. By using the credential set facility, you are able to easily set options that are not accessible via the Train URI.
+
+You may have as many credential sets in the config file as you require.
+
+### reporter
+
+You may also set output (reporter) options in the config file.  See the [Reporters Page](https://www.inspec.io/docs/reference/reporters/) for details.
--- a/docs/config.md
+++ b/docs/config.md
@@ -34,7 +34,7 @@ Config files must contain a top-level key, `version`, which indicates the file f
     "automate" : {
     "stdout" : false,
     "url" : "https://YOUR_A2_URL/data-collector/v0/",
-    "token" : "YOUR_A2_ADMIN_TOKEN",
+    "token" : "YOUR_A2_API_TOKEN",
     "insecure" : true,
     "node_name" : "inspec_test_node",
     "environment" : "prod"
--- a/docs/config.md
+++ b/docs/config.md
@@ -78,6 +78,12 @@ Each Train transport offers a variety of options. By using the credential set fa
 
 You may have as many credential sets in the config file as you require.
 
+If you use a target URI and the portion after the `://` cannot be matched to credential set name, InSpec will send the URI to Train to be parsed as a Train URI.  Thus, you can still do `ssh://someuser@myhost.com`.
+
+You can use a credential set, and then override individual options using command line options.
+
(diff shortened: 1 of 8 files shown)


lib/inspec/config.rb
```diff
@@ -81,8 +83,9 @@ def unpack_train_credentials
       credentials.merge!(_utc_generic_credentials)
 
       _utc_determine_backend(credentials)
-      credentials.merge!(Train.unpack_target_from_uri(final_options[:target] || '')) # TODO: this will be replaced with the credset work
       transport_name = credentials[:backend].to_s
+
+      _utc_merge_credset(credentials, transport_name)
```
Vasu1105, line 89: @clintoncwolfe  just want to understand why we are using _ in front of methods. Is it because we want to define it private ?

clintoncwolfe: For me, it means that the method is the result of a linter-forced refactoring.  In other words, to build the functionality of unpack_train_credentials, I needed to break it into many small pieces that would have no possible use to anything else.

If it turns out that one of the _utc methods are useful outside of the context of unpacking train credentials, I'd rename it.

clintoncwolfe:
I'm withdrawing this for now until we get the first pass of config file work in.

robbkidd:
@clintoncwolfe With the `cw/cred-set-support-02` branch merged to `master` in #3750, should this PR be updated to target `master` for the merge of `cw/cred-set-support-03`? Have I got that wrong? Or can I review the much smaller [diff between current `master` and `-03`](https://github.com/inspec/inspec/compare/4ad2c64f212280c1af932b011b9db77155854fa1...cw/cred-set-support-03)?

clintoncwolfe:
Thanks @robbkidd - `master` is indeed the correct target, making this an 86-line delta.

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+
+Should have the value '1.1'.
+
+### cli_options
+
+In this key, you can place any long-form command line option, without the leading dashes.
+
+### credentials
+
+Under this key, you may store any Train-transport-specific options.  You store the options keyed first by transport name, then by a name you choose to refer to them later.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
```
jerryaldrichiii: Thoughts on:

```suggestion
Under this key, you may store any Train-transport-specific options.  You store the options keyed first by transport name, then by a name you choose to refer to them later.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport_name://credset`.
```

docs/reporters.md
```diff
@@ -124,7 +124,7 @@ This renders html code to view your tests in a browser. It includes all the test
 
 ## Automate Reporter
 
-The automate reporter type is a special reporter used with the Automate 2 suite. To use this reporter you must pass in the correct configuration via a json config `--json-config`.
+The automate reporter type is a special reporter used with the Automate 2 suite. To use this reporter you must pass in the correct configuration via a json config `--config`.
```
jerryaldrichiii:
```suggestion
The Automate reporter type is a special reporter used with Chef Automate (Version 2). To use this reporter you must pass in the correct configuration via a json config `--config`.
```

lib/inspec/config.rb
```diff
@@ -137,6 +140,33 @@ def _utc_determine_backend(credentials)
...
+      elsif final_options.key?(:target)
+        # Not sure what target looked like at all!
+        # Let train parse it.
+        credentials.merge!(Train.unpack_target_from_uri(final_options[:target]))
+      end
+    end
+
+    def _utc_find_credset_name(_credentials, transport_name)
+      return nil unless final_options[:target]
+      match = final_options[:target].match(%r{^#{transport_name}://(?<credset_name>[a-z_\-0-9]+)$})
```
jerryaldrichiii: I don't think this RegEx will capture capital letters. Do we mandate the credset_name be lower case?

```
[1] pry(main)> "FOO".match(%r{(?<credset_name>[a-z_\-0-9]+)$})
=> nil
[2] pry(main)> "foo".match(%r{(?<credset_name>[a-z_\-0-9]+)$})
=> #<MatchData "foo" credset_name:"foo">
```

How about (which will catch any word, digit, or hypen):
```
[1] pry(main)> "fFo1_-".match(%r{(?<credset_name>[\w\d\-]+)$})
=> #<MatchData "fFo1_-" credset_name:"fFo1_-">
```
```suggestion
      match = final_options[:target].match(%r{^#{transport_name}://(?<credset_name>[\w\d\-]+)$})
```

jerryaldrichiii: I feel like this could also use some unit tests with capitals, hypens, digits, etc.

Also, curious if URI can parse it to avoid regex all together.

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+        "host":"somewhere.there.com",
+        "user":"bob"
+      }
+    }
+  },
+  "reporter": {
+    "automate" : {
+    "stdout" : false,
+    "url" : "https://YOUR_A2_URL/data-collector/v0/",
+    "token" : "YOUR_A2_ADMIN_TOKEN",
```
jerryaldrichiii:
```suggestion
    "token" : "YOUR_A2_API_TOKEN",
```

jerryaldrichiii (changes requested):
Looks good on a first pass, mostly docs suggestions and a RegEx change.

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+    "token" : "YOUR_A2_API_TOKEN",
+    "insecure" : true,
+    "node_name" : "inspec_test_node",
+    "environment" : "prod"
+  }
+}
+```
+
+
+### version
```
kagarmoe: Let's consider using a definition list for these:

```
version
: Should have the value '1.1'
```

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+
+Should have the value '1.1'.
+
+### cli_options
+
+In this key, you can place any long-form command line option, without the leading dashes.
```
kagarmoe:
```suggestion
Any long-form command line options, without the leading dashes.
```

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+
+### credentials
+
+Under this key, you may store any Train-transport-specific options.  You store the options keyed first by transport name, then by a name you choose to refer to them later.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
```
kagarmoe:
```suggestion
Train-transport-specific options.  Store the options keyed first by transport name, then by a name you'll use later on.  The combination of transport name and your chosen name can be used in the `--target` option to `inspec exec`, as `--target transport-name://connection-name`.
```
Use "...then a name for the connection" if it is always a connection-name.

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+        "host": "prod01.east.example.com",
+        "disable_sspi": true,
+        "connection_retries": 10
+      }
+    }
+  }
+}
+```
+
+Then you can use `--target winrm://myconn` to connect to the host, with the given extra options.
```
kagarmoe:
```suggestion
Then use `--target winrm://myconn` to connect to the host, with the given extra options.
```

docs/config.md
```diff
@@ -0,0 +1,89 @@
...
+
+Each Train transport offers a variety of options. By using the credential set facility, you are able to easily set options that are not accessible via the Train URI.
+
+You may have as many credential sets in the config file as you require.
+
+If you use a target URI and the portion after the `://` cannot be matched to credential set name, InSpec will send the URI to Train to be parsed as a Train URI.  Thus, you can still do `ssh://someuser@myhost.com`.
+
+You can use a credential set, and then override individual options using command line options.
+
+Credential sets are intended to work hand-in-hand with the underlying credentials storage facility of the transport. For example, if you have a `~/.ssh/config` file specifying that the sally-key.pem file should be used with the host `somehost.com`, and you have a credential set that specifies that host, then when Train tries to connect to that host, the SSH library will automatically use the SSH config file to use the indicated key.
```
kagarmoe, line 81:
```suggestion
Credential sets are intended to work hand-in-hand with the underlying credentials storage facility of the transport. 
For example, if your `~/.ssh/config` file specifies using `sally-key.pem` file with the host `somehost.com`, and you also have a credential set specifying that host, then when Train tries to connect to that host, the SSH library will automatically use the SSH config file with the indicated key.
```

docs/reporters.md
```diff
@@ -6,7 +6,7 @@ title: InSpec Reporters
 
 Introduced in InSpec 1.51.6
 
-InSpec allows you to output your test results to one or more reporters. You can configure the reporter(s) using either the `--json-config` option or the `--reporter` option. While you can configure multiple reporters to write to different files, only one reporter can output to the screen(stdout).
+InSpec allows you to output your test results to one or more reporters. You can configure the reporter(s) using either the `--config` (or `--json-config`, prior to v3.6) option or the `--reporter` option. While you can configure multiple reporters to write to different files, only one reporter can output to the screen(stdout).
```
kagarmoe:
```suggestion
InSpec allows you to output your test results to one or more reporters. Configure the reporter(s) using either the `--config` (or `--json-config`, prior to v3.6) option or the `--reporter` option. While you can configure multiple reporters to write to different files, only one reporter can output to the screen(stdout).
```

Also, it would be great to define what a reporter is at the top of this doc.

docs/reporters.md
kagarmoe:
```suggestion
The automate reporter type is a special reporter used with the [Chef Automate](https://automate.chef.io/). To use this reporter you must pass in the correct configuration via a json config `--config`.
```

kagarmoe (dismissed):
Some small changes.

docs/reporters.md
clintoncwolfe: For the first mention I've placed `automate` in backticks, because it refers to the technical name of the reporter, and case matters.  I've placed the second reference in a link.

lib/inspec/config.rb
clintoncwolfe: Okay, we'll make it take capital letters too.  I want to keep it as simple as possible, though.

We don't want URI to try to parse it, because URI accepts a much broader set of things than what we want to accept. http://somewhere.com/some/path is a quite valid URI; we could use URI to parse that, and then still be obliged to verify there is no path, the username is empty, the host contains no dots, etc.

This looks like parsing, but it isn't: it's just pattern matching with '://' as the separator, and very specific things on each side.

If we fail matching it here - because it really is a URI - we fall back to asking Train to parse it, which is looking for URIs, and uses `URI` to parse under the hood.

jerryaldrichiii (approved):

docs/config.md
clintoncwolfe: Each paragraph here describes a feature of the credset system.  Here, we're describing the transport-specific fallback feature, and the example given is intended to illustrate just that one feature.

clintoncwolfe:
I think the DCO violation is due to me accepting a github "suggestion."

lamont-granquist (approved):

docs/config.md
mjingle: @clintoncwolfe - Kimberly is just correcting the passive tense in the sentence structure. The correction still has your great content contribution.

docs/config.md
mjingle: Seconded on the definition list suggestion.

docs/config.md
clintoncwolfe: Ahh, it looks like the suggestion was to split the para.  Gotcha.

miah (approved):
Thanks @clintoncwolfe

PR #3717, opened 2019-01-11
Fix extra '&&' in windows functional tests, add diagnostics

clintoncwolfe:
This PR makes two changes to the functional test helper:

 1. Fixes a problem in which '&& ' was added as a spurious prefix to windows command invocations, which powershell did not like.
 2. Adds a method, `diagnose!`, to the run result object, which make it easy to dump the the invocation, STDOUT, and STDERR while running under CI.

### DIFF
--- a/test/functional/helper.rb
+++ b/test/functional/helper.rb
@@ -33,6 +33,15 @@ module Inspec
       @train_result = train_result
       @payload = OpenStruct.new
     end
+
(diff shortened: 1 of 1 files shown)


clintoncwolfe:
As this is a testing-harness change that blocks other PR reviews, I'm self-approving it.

PR #3719, opened 2019-01-12
Move attribute unit tests to a more expected location

clintoncwolfe:


clintoncwolfe:
Simple move of two files with no changes; unit tests are detected using a broad shell glob so this has no impact, other than helping people find them. Merging without review.

PR #3720, opened 2019-01-12
Plugins: Relocate plugin activate() method to Activator class, add simpler search method

clintoncwolfe:
This pays some technical debt from the early days of writing the plugin system. At first, Activators were bare structs, containers for data with no behavior.  It was the registry's job to find activators, and then activate them.

Over time, those two operations fell into a clear pattern: search the registry for an activator using certain criteria, take the first match, and then pass its ids to Registry to have it activated. That was silly, since you already had an Activator object as the result of the first search.

This PR:
 * Puts the activate() method in Activator where it belongs
 * Adds a convenience method to Registry which searches for Activators with the assumption you expect exactly one match.

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -40,7 +40,7 @@ module Inspec
           if hook
             # OK, load the hook if it hasn't been already.  We'll then know a module,
             # which we can then inject into the context
-            registry.activate(:control_dsl, method_name) unless hook.activated?
+            hook.activate unless hook.activated?
             # Inject the module's methods into the context.
             # implementation_class is the field name, but this is actually a module.
             self.class.include(hook.implementation_class)
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -40,7 +40,8 @@ module Inspec
           if hook
             # OK, load the hook if it hasn't been already.  We'll then know a module,
             # which we can then inject into the context
-            hook.activate unless hook.activated?
+            hook.activate
+
             # Inject the module's methods into the context.
             # implementation_class is the field name, but this is actually a module.
             self.class.include(hook.implementation_class)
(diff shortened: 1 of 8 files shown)


lib/inspec/plugin/v2/activator.rb
```diff
@@ -17,5 +17,26 @@ def activated?(new_value = nil)
...
       self[:'activated?'] = new_value
     end
+
+    # Load a plugin, but if an error is encountered, store it and continue
+    def activate
+      return if activated?
+      # rubocop: disable Lint/RescueException
+      begin
+        impl_class = self[:activation_proc].call
+        self[:'activated?'] = true
```
Vasu1105: Here I guess we can define this :'activated?' as :activated? as both means the same. If you check the output of `:'activated?'` and `:activated?` on irb both are same.

clintoncwolfe: Hey, thanks for that! In Ruby 1.9, the quotes were required if the symbol name had unusual characters, but you're right, `self[:activated?]` is safe in 2.3.8, 2.4.5, and 2.5.3, so I'll us that.

miah (approved):
Thanks for paying down our debt @clintoncwolfe!

lib/inspec/control_eval_context.rb
```diff
@@ -40,7 +40,7 @@ def method_missing(method_name, *arguments, &block)
           if hook
             # OK, load the hook if it hasn't been already.  We'll then know a module,
             # which we can then inject into the context
-            registry.activate(:control_dsl, method_name) unless hook.activated?
+            hook.activate unless hook.activated?
```
robbkidd: Any harm in simplifying all the `hook.activate unless hook.activated?` to be only `hook.activate`?

The `activate()` method has [a quick `return if activated?` already present](https://github.com/inspec/inspec/pull/3720/files#diff-80fde85ee69a12f9aad186be59ae05c3R23).

robbkidd:
All in all, huge fan of plugins/activators knowing how to register/activate themselves. 👍 

huge fan

lib/inspec/control_eval_context.rb
clintoncwolfe: A fine idea, sir.

jerryaldrichiii (approved):
Looks good to me! Thanks @clintoncwolfe!

PR #3723, opened 2019-01-14
allow bundler 2.x and bump integration gems

lamont-granquist:
the bundler bump is necessary for work with bundler 2.x installed.

have to see if the berkshelf + test-kitchen bumps can go through without turning tests red, but this gets rid of berkshelf <= 6.x and the ridley / celluloid deps.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'ffi', '>= 1.9.14'
 gem 'aws-sdk', '~> 2'
 
 group :test do
-  gem 'bundler', '~> 1.5'
+  gem 'bundler', '>= 1.5'
   gem 'minitest', '~> 5.5'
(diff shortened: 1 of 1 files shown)


clintoncwolfe:
On this, especially watch the internal build channel #inspec-notify to verify that the build now succeeds - the bundler constraint was breaking the build.

clintoncwolfe:
It looks like ruby 2.3.8 could not resolve its deps with Berkshelf 6+; to get to 6 we'd need to move back to Thor 0.19, which re-surfaces a bug; to get to berkshelf 7, we need to drop support for ruby 2.3 .

clintoncwolfe:
As this is a gem deps issue only affecting testing and breaking the build, I'm going to go ahead and merge this.

PR #3725, opened 2019-01-14
Adjust style elements on deprecation facility PR

jerryaldrichiii:
Some of this is personal preference, most is our RuboCop.

Let me know if you disagree with anything 😄.

### DIFF
--- a/test/functional/logging_test.rb
+++ b/test/functional/logging_test.rb
@@ -1,7 +1,6 @@
 # Logging and deprecation facilities functional tests
 
 require 'functional/helper'
-require 'byebug'
 
 describe 'Deprecation Facility Behavior' do
   include FunctionalHelper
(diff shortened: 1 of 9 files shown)


test/functional/logging_test.rb
```diff
@@ -1,27 +1,25 @@
...
 
   describe 'when the deprecation is in a custom resource and the deprecate DSL method is used' do
     describe 'when the action is to fail the control' do
-
       describe 'when the resource is called in a control' do
         let(:profile_name) { 'typical' }
         let(:control_flag) { '--controls deprecate_fail_mode' }
 
-        it "should result in a failed control" do
+        it 'should result in a failed control' do
```
miah, line 22: I think chefstyle will want us to change these all to `"`, so might want to leave these for now, but I agree on principal =)

miah (approved):
solid. thanks for these @jerryaldrichiii. I'm +1 on this, my only qualm is the `'` vs `"` bits. I'm generally a fan of only using `'` unless interpolation is happening, but with chefstyle coming at some point in the future we might want to leave these as is.

tas50 (approved):

tas50:
💥

test/functional/logging_test.rb
jerryaldrichiii: Yup, I am all for one OR the other, but must be the same in a single file.

miah: In that case they should just all be `"` because with `'` we'll still need to change to `"` for anything that does interpolation (unless you use a percent literal like `%Q[]`, which I don't think we do often/anywhere)

jerryaldrichiii: I agree, but this breaks our RuboCop currently (even though `test/` isn't scanned).

miah: Ok, lets just note that then and we'll churn this after chefstyle =)

test/unit/mock/profiles/deprecation/resource_pack/libraries/deprecation_tester.rb
```diff
@@ -1,62 +1,59 @@
...
+        "a_group_that_will_fail": { "action": "fail_control" }
+      }
     }
-  }
   EOC
 
   def fail_me
     Inspec::Deprecation::Deprecator.class_test_cfg_io(StringIO.new(DEPRECATION_CFG))
 
-    #deprecate(:a_group_that_will_fail, 'This should fail')
```
clintoncwolfe: This is not a style change, and should be reviewed on the main PR.

PR #3729, opened 2019-01-16
Allow 'Proprietary' as a valid license term.

james-stocks:
Currently our supported profiles output a warning for every instance of `inspec check` because our license metadata `Proprietary, All rights reserved` is not considered valid.
This commit allows for a string beginning with `Proprietary` to be considered valid, as well as any valid SPDX value.

This is part of our overall desire to be able to use `inspec check` in CI and fail on warnings.

### DIFF
--- a/test/unit/profiles/profile_test.rb
+++ b/test/unit/profiles/profile_test.rb
@@ -294,7 +294,7 @@ describe Inspec::Profile do
 
       it 'prints ok messages and counts the controls' do
         logger.expect :info, nil, ["Checking profile in #{home}/mock/profiles/#{profile_id}"]
-        logger.expect :warn, nil, ["License 'Invalid License Name' needs to be in SPDX format. See https://spdx.org/licenses/."]
+        logger.expect :warn, nil, ["License 'Invalid License Name' needs to be in SPDX format or marked as 'Proprietary'. See https://spdx.org/licenses/."]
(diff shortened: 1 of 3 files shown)


test/unit/profiles/profile_test.rb
```diff
@@ -294,7 +294,7 @@
 
       it 'prints ok messages and counts the controls' do
         logger.expect :info, nil, ["Checking profile in #{home}/mock/profiles/#{profile_id}"]
-        logger.expect :warn, nil, ["License 'Invalid License Name' needs to be in SPDX format. See https://spdx.org/licenses/."]
+        logger.expect :warn, nil, ["License 'Invalid License Name' needs to be in SPDX format or marked as 'Proprietary'. See https://spdx.org/licenses/."]
```
james-stocks, line 297: This file does not actually get included in `rake test`. I needed to use `m` to run it. 
Separate to the change in this PR, I'll look into why the file's not getting included in the test run.

clintoncwolfe (approved):
Thanks @stocksy - and good work on that regex to catch commas on boundaries.

test/unit/profiles/profile_test.rb
clintoncwolfe: I opened #3733 to track, thanks

jerryaldrichiii (approved):
Great work @james-stocks!

PR #3731, opened 2019-01-16
WIP Add and adhere to `chefstyle`

miah:
DO NOT MERGE (yet).. This is part of a spike.

This PR incorporates [chefstyle](https://github.com/chef/chefstyle) into InSpec! This change will help ensure consistency in style across all Ruby projects at Chef. Previously, we were not enforcing style on tests, and code examples. This changes all that.

I have not incorporated `chefstyle` into Rake. I only removed and disabled the original Rubocop setup, and ran `chefstyle -a` This was just a quick 'look and see' =) 

When I first pulled in `chefstyle` here were the stats:

```
20604  Style/StringLiterals
545    Layout/SpaceAroundOperators
386    Layout/SpaceInsideHashLiteralBraces
215    Layout/TrailingWhitespace
207    Style/TrailingCommaInArguments
175    Layout/SpaceInsideBlockBraces
150    Style/WordArray
133    Style/TrailingCommaInHashLiteral
124    Style/BlockDelimiters
122    Style/HashSyntax
104    Layout/SpaceAfterComma
89     Layout/AlignHash
88     Layout/TrailingBlankLines
68     Layout/ExtraSpacing
57     Layout/EmptyLines
51     Style/PercentLiteralDelimiters
46     Lint/ParenthesesAsGroupedExpression
31     Lint/AmbiguousBlockAssociation
29     Style/AndOr
26     Layout/IndentHeredoc
26     Lint/AmbiguousRegexpLiteral
26     Lint/UselessAssignment
24     Layout/IndentationConsistency
23     Layout/SpaceAfterColon
19     Layout/MultilineOperationIndentation
17     Style/MutableConstant
16     Layout/LeadingCommentSpace
16     Style/NumericPredicate
15     Layout/SpaceBeforeBlockBraces
14     Style/MultilineIfModifier
12     Layout/IndentationWidth
12     Style/MultilineIfThen
10     Lint/ImplicitStringConcatenation
9      Layout/DotPosition
9      Layout/EmptyLineBetweenDefs
9      Style/RedundantReturn
9      Style/SelfAssignment
9      Style/SignalException
8      Layout/MultilineBlockLayout
8      Layout/Tab
7      Layout/EmptyLinesAroundMethodBody
7      Layout/IndentAssignment
7      Naming/ConstantName
6      Performance/RedundantBlockCall
5      Layout/EmptyLinesAroundAccessModifier
4      Layout/BlockAlignment
4      Layout/SpaceAroundEqualsInParameterDefault
[...]

PR #3732, opened 2019-01-17
SPIKE: Investigate implementing ChefStyle

jerryaldrichiii:
This PR is to capture conversation and investigate the amount of work that would be required to implement [chefstyle](https://github.com/chef/chefstyle) in InSpec to match [chef](https://github.com/chef/chef). After investigating, I realized it would only be an hour or so to: swap out `rubocop.yml`, run `chefstyle -a`, and fix the errant tests.

However, this process uncovered a few items that may cause issues. Mainly:
  - I'm unsure about how long it would take review this change (how much do we trust `rubocop -a`?).
  - There are a few AWS resource errors that may be valid and need to be investigated
  - We need to find a way to do the `chefstyle -a` commit as another user (or accept that one user will have the metric of changing ~17,000 lines of code)
    - I've attempted this...but I want to be 100% sure I don't get "credit" for modifying so many lines. I'm fine with signing the commit or having another author do this PR however.

This PR supersedes  #3731 due to changes in `rubocop.yml` prior to `chefstyle -a`. Thank you @miah for the initial work into this.

@clintoncwolfe @miah thoughts?


jerryaldrichiii:
Looks like the functional tests aren't too hairy, more quote changes. I'll make it right.

examples/profile-aws/controls/iam_users_console_users_mfa.rb
```diff
@@ -1,8 +1,8 @@
-control 'iam_users_console_users_mfa' do
-  title 'Users with console access must have MFA enabled'
+control "iam_users_console_users_mfa" do
+  title "Users with console access must have MFA enabled"
   impact 1.0
 
-  describe aws_iam_users.where { has_console_password and not has_mfa_enabled } do
-    its('entries') { should be_empty }
+  describe aws_iam_users.where { has_console_password && (not has_mfa_enabled) } do
```
miah, line 5: I'm curious if this works as expected. I know that `and` is used in lots of profiles and wonder if this is part of the rspec/inspec dsl and how this change affects that?

`and` is a [Ruby keyword](https://docs.ruby-lang.org/en/2.2.0/keywords_rdoc.html), but its also used by [rspec to create compound exceptions](https://relishapp.com/rspec/rspec-expectations/docs/compound-expectations)

jerryaldrichiii:
Discussion from standup:
  - Need to break out failing tests from AWS
  - Need to do this change when we reduce PR count (due to merge conflicts)
  - Should we do a minor bump? (SemVer doesn't require it)

jerryaldrichiii:
Closing this, the amount of work has been scoped and potential issues identified.

Essentially, the amount of work is relatively small, but the change touches nearly every file so we should only do this only when our outstanding PR count is low. Also, we should debate whether a minor version bump is needed when we do the work.

PR #3739, opened 2019-01-21
Fix AWS tests found during ChefStyle spike

jerryaldrichiii:
While doing https://github.com/inspec/inspec/pull/3732 I found a few style issues with the AWS tests.

This fixes those style issues and also fixes a duplicated test method name.

### DIFF
--- a/test/unit/resources/aws_elb_test.rb
+++ b/test/unit/resources/aws_elb_test.rb
@@ -13,15 +13,15 @@ class AwsElbConstructorTest < Minitest::Test
   end
 
   def test_empty_params_rejected
-    assert_raises(ArgumentError) { AwsElb.new }    
+    assert_raises(ArgumentError) { AwsElb.new }
   end
 
   def test_string_accepted
-    AwsElb.new 'my-elb'    
+    AwsElb.new 'my-elb'
   end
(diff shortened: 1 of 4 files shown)


clintoncwolfe (changes requested):
Please split this into two PRs, one for style-in-tests issues, and one for the rename.  Thanks!

miah (approved):
Looks good. Thanks @jerryaldrichiii !

miah:
I don't think splitting this PR would be useful. I do think thats a useful behavior on large PR's, but this is 4 files with a 23 line diff.

clintoncwolfe (approved):
Alrighty then.

jerryaldrichiii:
I also don't think it is a valuable use of time to split this PR.

I see the value of splitting large PRs into chunks for review. In this case though I don't think the size/content of this PR hinders the ability to review.

If we decide that PRs like this should be split, then we should document it somewhere. I'm not opposed to splitting, but I would like the qualifier to not be subjective.

EDIT: I see @clintoncwolfe approved this just as I sent the above message. Didn't mean for it to look like I was adding on after the fact. I think we should chat about it, but not here.

PR #3743, opened 2019-01-21
Clean up unit test output

jerryaldrichiii:
This reduces the amount of extraneous unit test output. I was not able to get all of the extra output removed easily and have created https://github.com/inspec/inspec/issues/3741 and https://github.com/inspec/inspec/issues/3742 to track the remaining work. This gets us close though.

See:

[current_output.txt](https://github.com/inspec/inspec/files/2780720/current_output.txt) vs [cleaner_output.txt](https://github.com/inspec/inspec/files/2780721/cleaner_output.txt)

Many thanks to @miah helping me with this a while back!

PR #3740, opened 2019-01-21
iis_app_pool:  Fixes error with 'should not exist'

jerryaldrichiii:
This adopts #3419 and fixes the mocked command (/me shakes fists at newline chars)

Thank you so much @TheLonelyGhost for doing 99.9% of the work here!

PR #3744, opened 2019-01-21
Fix SSL tests by allowing multiple cipher counts

jerryaldrichiii:
The value of `ciphers` is 681 on my localhost, but 993 on Travis.

This modifies the test to allow both values.

When/If this is merged I will need to rebase:
 - https://github.com/inspec/inspec/pull/3739
 - https://github.com/inspec/inspec/pull/3740
 - https://github.com/inspec/inspec/pull/3743


clintoncwolfe (approved):
It looks brittle , but really is no more brittle than the previously existing code.  Thanks!

miah (approved):
Agree that this is kinda brittle as the list of ciphers may vary based on Linux/UNIX openssl configuration.

PR #3712, opened 2019-01-08
Config file formalization and credential sets: part 2, using the Config system

clintoncwolfe:
This is part 2 of #3709.

This PR alters the CLI and some related code to rely on the Inspec::Config class introduced on #3710 .

Prerequisites: this PR must be merged after #3710 and after https://github.com/inspec/train/pull/394

NOTE: This PR contains a modification to the Gemfile, which must be removed prior to merge; after https://github.com/inspec/train/pull/394 the modification may be removed.

Issue #3667, opened 2018-12-11
Convert Reporters to a v2-API Plugin Type


clintoncwolfe:
### 🎛 Description

Refs #3638

🙋 feature request

Reporters (output formats) are semi-modular (a "v0" plugin).  Most are based off of the RSpec JSON format; but there are kludges involved.  Adding a new reporter is not a clear or easy task, but is often requested.

* [x] Make a Reporter v2-API plugin type
  * [x] Should support one-shot (wait-for-everything) mode
  * ~Should support incremental (progessive) mode~ (deferred to #4144) 
* Clean up existing reporters as needed
* Convert existing reporters into plugins
  * [ ] automate
  * [ ] cli #5069
  * [ ] json
  * [ ] json-automate
  * [ ] json-min #5062
  * [ ] junit #5084
  * [ ] yaml
* [x] Document how to make a new reporter

### 🌍 InSpec and Platform Version

3.0.61+

No work has started as of Dec 11; we are very interested in what people would like in terms of the API.

juju4:
would be interesting to report
- as log (classical syslog, json...)
- as metrics (graphite, prometheus...)

PR #3750, opened 2019-01-24
Formalize Config File

clintoncwolfe:
This PR implements part of #3661.

 * Creates a dedicated Inspec::Config class to represent the runtime configuration
 * Moves config file reading and merging from lib/inspec/base_cli.rb to the config class. **No attempt is made to refactor or otherwise improve the code moved from base_cli** - that is out of scope for this PR.
 * Adds validation to the config file
 * Modifies the CLI command handlers to use the config object.

This PR obsoletes #3710 and #3712.

A later PR, similar to #3713, will be required to complete #3661. 

~~BEFORE MERGING, train/pull/394 must be merged, and the Gemfile reset in this PR.~~

### DIFF
--- a/lib/bundles/inspec-supermarket/cli.rb
+++ b/lib/bundles/inspec-supermarket/cli.rb
@@ -30,7 +30,7 @@ module Supermarket
     desc 'exec PROFILE', 'execute a Supermarket profile'
     exec_options
     def exec(*tests)
-      o = opts(:exec).dup
+      o = config
       diagnose(o)
       configure_logger(o)
 
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -41,16 +41,16 @@ module Inspec
     # @param [Hash] config for the transport backend
     # @return [TransportBackend] enriched transport instance
     def self.create(config) # rubocop:disable Metrics/AbcSize
-      conf = Train.target_config(config)
-      name = Train.validate_backend(conf)
-      transport = Train.create(name, conf)
+      train_credentials = config.unpack_train_credentials
+      transport_name = Train.validate_backend(train_credentials)
+      transport = Train.create(transport_name, train_credentials)
       if transport.nil?
-        raise "Can't find transport backend '#{name}'."
+        raise "Can't find transport backend '#{transport_name}'."
       end
 
       connection = transport.connection
       if connection.nil?
-        raise "Can't connect to transport backend '#{name}'."
+        raise "Can't connect to transport backend '#{transport_name}'."
       end
 
       # Set caching settings. We always want to enable caching for
@@ -85,9 +85,9 @@ module Inspec
 
       cls.new
     rescue Train::ClientError => e
-      raise "Client error, can't connect to '#{name}' backend: #{e.message}"
+      raise "Client error, can't connect to '#{transport_name}' backend: #{e.message}"
     rescue Train::TransportError => e
-      raise "Transport error, can't connect to '#{name}' backend: #{e.message}"
+      raise "Transport error, can't connect to '#{transport_name}' backend: #{e.message}"
     end
   end
 end
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -75,8 +75,9 @@ module Inspec
         desc: 'Whether to use disable sspi authentication, defaults to false (WinRM).'
       option :winrm_basic_auth, type: :boolean,
         desc: 'Whether to use basic authentication, defaults to false (WinRM).'
-      option :json_config, type: :string,
+      option :config, type: :string,
         desc: 'Read configuration from JSON file (`-` reads from stdin).'
+      option :json_config, type: :string, hide: true
       option :proxy_command, type: :string,
         desc: 'Specifies the command to use to connect to the server'
       option :bastion_host, type: :string,
@@ -118,92 +119,7 @@ module Inspec
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
     end
 
-    def self.default_options
-      {
-        exec: {
-          'reporter' => ['cli'],
-          'show_progress' => false,
-          'color' => true,
-          'create_lockfile' => true,
-          'backend_cache' => true,
-        },
-        shell: {
-          'reporter' => ['cli'],
-        },
-      }
-    end
-
-    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize
-      # default to cli report for ad-hoc runners
-      opts['reporter'] = ['cli'] if opts['reporter'].nil?
-
-      # parse out cli to proper report format
-      if opts['reporter'].is_a?(Array)
-        reports = {}
-        opts['reporter'].each do |report|
-          reporter_name, target = report.split(':', 2)
-          if target.nil? || target.strip == '-'
-            reports[reporter_name] = { 'stdout' => true }
-          else
-            reports[reporter_name] = {
-              'file' => target,
-              'stdout' => false,
-            }
-            reports[reporter_name]['target_id'] = opts['target_id'] if opts['target_id']
-          end
-        end
-        opts['reporter'] = reports
-      end
-
-      # add in stdout if not specified
-      if opts['reporter'].is_a?(Hash)
-        opts['reporter'].each do |reporter_name, config|
-          opts['reporter'][reporter_name] = {} if config.nil?
-          opts['reporter'][reporter_name]['stdout'] = true if opts['reporter'][reporter_name].empty?
-          opts['reporter'][reporter_name]['target_id'] = opts['target_id'] if opts['target_id']
-        end
-      end
-
-      validate_reporters(opts['reporter'])
-      opts
-    end
-
-    def self.validate_reporters(reporters)
-      return if reporters.nil?
-
-      valid_types = [
-        'automate',
-        'cli',
-        'documentation',
-        'html',
-        'json',
-        'json-automate',
-        'json-min',
-        'json-rspec',
-        'junit',
-        'progress',
-        'yaml',
-      ]
-
-      reporters.each do |k, v|
-        raise NotImplementedError, "'#{k}' is not a valid reporter type." unless valid_types.include?(k)
-
-        next unless k == 'automate'
-        %w{token url}.each do |option|
-          raise Inspec::ReporterError, "You must specify a automate #{option} via the json-config." if v[option].nil?
-        end
-      end
-
-      # check to make sure we are only reporting one type to stdout
-      stdout = 0
-      reporters.each_value do |v|
-        stdout += 1 if v['stdout'] == true
-      end
-
-      raise ArgumentError, 'The option --reporter can only have a single report outputting to stdout.' if stdout > 1
-    end
-
-    def self.detect(params: {}, indent: 0, color: 39)
+    def self.format_platform_info(params: {}, indent: 0, color: 39)
       str = ''
       params.each { |item, info|
         data = info
@@ -286,86 +202,12 @@ module Inspec
       false
     end
 
-    def diagnose(opts)
-      return unless opts['diagnose']
-      puts "InSpec version: #{Inspec::VERSION}"
-      puts "Train version: #{Train::VERSION}"
-      puts 'Command line configuration:'
-      pp options
-      puts 'JSON configuration file:'
-      pp options_json
-      puts 'Merged configuration:'
-      pp opts
-      puts
+    def diagnose(_ = nil)
+      config.diagnose
     end
 
-    def opts(type = nil)
-      o = merged_opts(type)
-
-      # Due to limitations in Thor it is not possible to set an argument to be
-      # both optional and its value to be mandatory. E.g. the user supplying
-      # the --password argument is optional and not always required, but
-      # whenever it is used, it requires a value. Handle options that were
-      # defined above and require a value here:
-      %w{password sudo-password}.each do |v|
-        id = v.tr('-', '_').to_sym
-        next unless o[id] == -1
-        raise ArgumentError, "Please provide a value for --#{v}. For example: --#{v}=hello."
-      end
-
-      # Infer `--sudo` if using `--sudo-password` without `--sudo`
-      if o[:sudo_password] && !o[:sudo]
-        o[:sudo] = true
-        warn 'WARN: `--sudo-password` used without `--sudo`. Adding `--sudo`.'
-      end
-
-      # check for compliance settings
-      if o['compliance']
-        require 'plugins/inspec-compliance/lib/inspec-compliance/api'
-        InspecPlugins::Compliance::API.login(o['compliance'])
-      end
-
-      o
-    end
-
-    def merged_opts(type = nil)
-      opts = {}
-
-      # start with default options if we have any
-      opts = BaseCLI.default_options[type] unless type.nil? || BaseCLI.default_options[type].nil?
-      opts['type'] = type unless type.nil?
-      Inspec::BaseCLI.inspec_cli_command = type
-
-      # merge in any options from json-config
-      json_config = options_json
-      opts.merge!(json_config)
-
-      # merge in any options defined via thor
-      opts.merge!(options)
-
-      # parse reporter options
-      opts = BaseCLI.parse_reporters(opts) if %i(exec shell).include?(type)
-
-      Thor::CoreExt::HashWithIndifferentAccess.new(opts)
-    end
-
-    def options_json
-      conffile = options['json_config']
-      @json ||= conffile ? read_config(conffile) : {}
-    end
-
-    def read_config(file)
-      if file == '-'
-        puts 'WARN: reading JSON config from standard input' if STDIN.tty?
-        config = STDIN.read
-      else
-        config = File.read(file)
-      end
-
-      JSON.parse(config)
-    rescue JSON::ParserError => e
-      puts "Failed to load JSON configuration: #{e}\nConfig was: #{config.inspect}"
-      exit 1
+    def config
+      @config ||= Inspec::Config.new(options) # 'options' here is CLI opts from Thor
     end
 
     # get the log level
@@ -409,12 +251,12 @@ module Inspec
 
     def configure_logger(o)
       #
-      # TODO(ssd): This is a big gross, but this configures the
+      # TODO(ssd): This is a bit gross, but this configures the
       # logging singleton Inspec::Log. Eventually it would be nice to
       # move internal debug logging to use this logging singleton.
       #
-      loc = if o.log_location
-              o.log_location
+      loc = if o['log_location']
+              o['log_location']
             elsif suppress_log_output?(o)
               STDERR
             else
@@ -422,14 +264,14 @@ module Inspec
             end
 
       Inspec::Log.init(loc)
-      Inspec::Log.level = get_log_level(o.log_level)
+      Inspec::Log.level = get_log_level(o['log_level'])
 
       o[:logger] = Logger.new(loc)
       # output json if we have activated the json formatter
       if o['log-format'] == 'json'
         o[:logger].formatter = Logger::JSONFormatter.new
       end
-      o[:logger].level = get_log_level(o.log_level)
+      o[:logger].level = get_log_level(o['log_level'])
     end
   end
 end
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -15,6 +15,7 @@ require 'inspec/plugin/v2'
 require 'inspec/runner_mock'
 require 'inspec/env_printer'
 require 'inspec/schema'
+require 'inspec/config'
 
 class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :log_level, aliases: :l, type: :string,
@@ -45,12 +46,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: 'A list of controls to include. Ignore all other tests.'
   profile_options
   def json(target)
-    o = opts.dup
+    o = config
     diagnose(o)
     o['log_location'] = STDERR
     configure_logger(o)
 
-    o[:backend] = Inspec::Backend.create(target: 'mock://')
+    o[:backend] = Inspec::Backend.create(Inspec::Config.mock)
     o[:check_mode] = true
     o[:vendor_cache] = Inspec::Cache.new(o[:vendor_cache])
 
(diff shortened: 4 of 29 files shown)


lib/bundles/inspec-supermarket/cli.rb
```diff
@@ -30,7 +30,7 @@ def profiles
     desc 'exec PROFILE', 'execute a Supermarket profile'
     exec_options
     def exec(*tests)
-      o = opts(:exec).dup
+      o = config
```
clintoncwolfe, line 33: Nearly all CLI command implementation method begin with a call to the `opts` method, which would read the config, merge, and finagle the options. `config` does all that too, but does so in a more self-contained way.

You'll see a similar change with every CLI command method, sometimes with logging or other adjustments included.

lib/inspec/backend.rb
```diff
@@ -41,16 +41,16 @@ def inspect
     # @param [Hash] config for the transport backend
     # @return [TransportBackend] enriched transport instance
     def self.create(config) # rubocop:disable Metrics/AbcSize
-      conf = Train.target_config(config)
-      name = Train.validate_backend(conf)
-      transport = Train.create(name, conf)
+      train_credentials = config.unpack_train_credentials
```
clintoncwolfe, line 44: These two stanzas demonstrate the strange coupling between InSpec and Train. InSpec is in charge of obtaining connection options / credentials from the user, but Train is responsible for loading the correct Transport, and validating the options.  Finally, InSpec asks Train to create a connection, implicitly using the options the transport validated.

This PR is mostly about putting config in a clean class, so no attempt is made to decouple that mess; here we really just rename some vars and methods to get some clarity.

lib/inspec/base_cli.rb
```diff
@@ -118,92 +119,7 @@ def self.exec_options
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
     end
 
-    def self.default_options
```
clintoncwolfe, line 121: A major piece of this PR is simply moving code from `lib/inspec/base_cli.rb` to `lib/inspec/config.rb`. Most of the methods were class methods on the CLI base class, which wasn't really needed (though they did not access state); so when moved to Config, they became instance methods. Also, they were given clearer (IMHO) names.

Many test files and other classes sometimes called these methods; so you'll see changes in other files to update them fetching it from the new place.

lib/inspec/base_cli.rb
```diff
@@ -118,92 +119,7 @@ def self.exec_options
...
-          'create_lockfile' => true,
-          'backend_cache' => true,
-        },
-        shell: {
-          'reporter' => ['cli'],
-        },
-      }
-    end
-
-    def self.parse_reporters(opts) # rubocop:disable Metrics/AbcSize
```
clintoncwolfe, line 136: Reporter parsing and validation will eventually be reworked on #3667. For now, we just move the code; it could certainly be clarified.

lib/inspec/base_cli.rb
```diff
@@ -286,86 +202,12 @@ def suppress_log_output?(opts)
...
-      puts 'Command line configuration:'
-      pp options
-      puts 'JSON configuration file:'
-      pp options_json
-      puts 'Merged configuration:'
-      pp opts
-      puts
-    end
-
-    def opts(type = nil)
```
clintoncwolfe, line 302: This is the main entry point that most CLI commands have to the options system.  This was replaced with `config`, which instantiates a Config object.

lib/inspec/base_cli.rb
```diff
@@ -286,86 +202,12 @@ def suppress_log_output?(opts)
...
-      o = merged_opts(type)
-
-      # Due to limitations in Thor it is not possible to set an argument to be
-      # both optional and its value to be mandatory. E.g. the user supplying
-      # the --password argument is optional and not always required, but
-      # whenever it is used, it requires a value. Handle options that were
-      # defined above and require a value here:
-      %w{password sudo-password}.each do |v|
```
clintoncwolfe, line 310: I call all such 'tweaking' by the name 'finalization'. Muddy waters here; this PR only isolates it, doesn't improve.

lib/inspec/cli.rb
```diff
@@ -39,12 +40,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
...
   profile_options
   def json(target)
-    o = opts.dup
+    o = config
     diagnose(o)
     o['log_location'] = STDERR
     configure_logger(o)
 
-    o[:backend] = Inspec::Backend.create(target: 'mock://')
+    o[:backend] = Inspec::Backend.create(Inspec::Config.mock)
```
clintoncwolfe, line 54: Anytime InSpec needs to "pretend" to do a run - like to export the contents of a profile, or perform analysis of a profile - we setup a "mock" backend, and actually run the profile. That shakes out to also requiring a dummy config, so I added a constructor that makes a dummy.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      sudo_options
+      shell
+      shell_options
+      shell_command
+    }.freeze
+
+    extend Forwardable
+
+    # Many parts of InSpec expect to treat the Config as a Hash
+    def_delegators :@final_options, :each, :delete, :[], :[]=, :key?
```
clintoncwolfe, line 24: Really, the opts Hash was a mess; things just write and delete from it, and duplicate it, willy-nilly.  This class is not much better, as it just wraps up the Hash and delegates the common Hash operations, but it now has a chance of intercepting things, should we wish too.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+    def self.mock(opts = {})
+      Inspec::Config.new({ backend: :mock }.merge(opts), StringIO.new('{}'))
+    end
+
+    def initialize(cli_opts = {}, cfg_io = nil, command_name = nil)
+      @command_name = command_name || (ARGV.empty? ? nil : ARGV[0].to_sym)
+      @defaults = Defaults.for_command(@command_name)
+
+      @cli_opts = cli_opts.dup
+      cfg_io = resolve_cfg_io(@cli_opts, cfg_io)
```
clintoncwolfe, line 37: This is kind of ugly, and could be simplified, I think.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      @cfg_file_contents = read_cfg_file_io(cfg_io)
+
+      @merged_options = merge_options
+      @final_options = finalize_options
+    end
+
+    def diagnose
```
clintoncwolfe, line 44: Moved from base_cli.rb without changes.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+    #      so this method doesn't load defaults.
+    #  * individual InSpec CLI options (which in many cases may have the
+    #      transport name prefixed, which is stripped before being added
+    #      to the creds hash)
+    #  * the --target CLI option, which is interpreted:
+    #     - as an arbitrary URI, which is parsed by Train.unpack_target_from_uri
+
+    def unpack_train_credentials
+      # Internally, use indifferent access while we build the creds
+      credentials = Thor::CoreExt::HashWithIndifferentAccess.new({})
```
clintoncwolfe, line 77: /whimper

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+
+      # Helper methods prefixed with _utc_ (Unpack Train Credentials)
+
+      credentials.merge!(_utc_generic_credentials)
+
+      _utc_determine_backend(credentials)
+      credentials.merge!(Train.unpack_target_from_uri(final_options[:target] || '')) # TODO: this will be replaced with the credset work
```
clintoncwolfe, line 84: This line relies on https://github.com/inspec/train/pull/394

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      transport_name = credentials[:backend].to_s
+      _utc_merge_transport_options(credentials, transport_name)
+
+      # Convert to all-Symbol keys
+      credentials.each_with_object({}) do |(option, value), creds|
```
clintoncwolfe, line 89: Wait, what?

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+        creds[option.to_sym] = value
+        creds
+      end
+    end
+
+    private
+
+    def _utc_merge_transport_options(credentials, transport_name)
```
clintoncwolfe, line 97: The idea here is that we currently document options like '--host', '--keyfile', '--port'; as things get more complicated and we add more transports with more options, we'd like to be able to namespace them (so the `ssh` transport would get '--ssh-host', '--ssh-keyfile', etc.

We'd continue to support the unprefixed versions of anything we have documented; but we want to make things like #3669 more straightforward and predictable.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      end
+      credentials[:backend] = transport_name.to_s # these are indeed stored in Train as Strings.
+    end
+
+    #-----------------------------------------------------------------------#
+    #                         Reading Config Files
+    #-----------------------------------------------------------------------#
+
+    # Regardless of our situation, end up with a readable IO object
+    def resolve_cfg_io(cli_opts, cfg_io)
```
clintoncwolfe, line 144: I gave myself a headache re-reading this one

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+    def config_file_reporter_options
+      # This is assumed to be top-level in both legacy and 1.1.
+      # Technically, you could sneak it in the 1.1 cli opts area.
+      @cfg_file_contents.key?('reporter') ? { 'reporter' => @cfg_file_contents['reporter'] } : {}
+    end
+
+    #-----------------------------------------------------------------------#
+    #                            Validation
+    #-----------------------------------------------------------------------#
+    def validate_config_file_contents!
```
clintoncwolfe, line 217: This PR does not yet make any attempt to make it easy to validate or consume multiple versions, besides the legacy and 1.1 versions.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+
+      valid_fields = %w{version cli_options credentials compliance reporter}.sort
+      @cfg_file_contents.keys.each do |seen_field|
+        unless valid_fields.include?(seen_field)
+          raise Inspec::ConfigError::Invalid, "Unrecognized top-level configuration field #{seen_field}.  Recognized fields: #{valid_fields.join(', ')}"
+        end
+      end
+    end
+
+    def validate_reporters!(reporters)
```
clintoncwolfe, line 235: Moved (nearly) verbatim from base_cli.rb

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      # Highest precedence: merge in any options defined via the CLI
+      options.merge!(@cli_opts)
+
+      options
+    end
+
+    #-----------------------------------------------------------------------#
+    #                          Finalization
+    #-----------------------------------------------------------------------#
+    def finalize_options
```
clintoncwolfe, line 293: Every step of this feels like overly mixed concerns; but this PR just aims to isolate and gather this sort of thing, and hopefully make it obvious that this sort of thing should be happening somewhere else.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+
+      Thor::CoreExt::HashWithIndifferentAccess.new(options)
+    end
+
+    def finalize_set_top_level_command(options)
+      options[:type] = @command_name
+      Inspec::BaseCLI.inspec_cli_command = @command_name # TODO: move to a more relevant location
+    end
+
+    def finalize_parse_reporters(options) # rubocop:disable Metrics/AbcSize
```
clintoncwolfe, line 309: Moved from base_cli.rb

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+          options['reporter'][reporter_name]['stdout'] = true if options['reporter'][reporter_name].empty?
+          options['reporter'][reporter_name]['target_id'] = options['target_id'] if options['target_id']
+        end
+      end
+
+      validate_reporters!(options['reporter'])
+      options
+    end
+
+    def finalize_handle_sudo(options)
```
clintoncwolfe, line 344: Moved (with some re-arrangement) from base_cli.rb

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      end
+
+      # Infer `--sudo` if using `--sudo-password` without `--sudo`
+      if options['sudo_password'] && !options['sudo']
+        options['sudo'] = true
+        Inspec::Log.warn '`--sudo-password` used without `--sudo`. Adding `--sudo`.'
+      end
+    end
+
+    def finalize_compliance_login(options)
```
clintoncwolfe, line 363: Moved (with some re-arrangement) from base_cli.rb

test/functional/inspec_archive_test.rb
```diff
@@ -49,23 +49,11 @@
     end
   end
 
-  it 'archive wont overwrite existing files' do
+  it 'archive will overwrite existing files even without --overwrite' do
```
clintoncwolfe, line 52: So, ah, maybe we should talk about this, but archive always clobbers.  Discovered during testing, unrelated to this PR.

test/functional/inspec_exec_test.rb
```diff
@@ -555,4 +557,142 @@
...
+
+    describe 'when using the --config option to read from a custom file' do
+      let(:cli_args) { '--config ' + File.join(config_dir_path, 'json-config', 'good.json') }
+      it 'should see the custom target ID value' do
+        stderr.must_be_empty
+        seen_target_id.must_equal 'from-config-file'
+      end
+    end
+
+    # TODO - Use bash redirection to populate STDIN
```
clintoncwolfe: Open to suggestions on this one.

test/functional/inspec_exec_test.rb
```diff
@@ -555,4 +557,142 @@
...
+    #       stderr.must_be_empty
+    #       seen_target_id.must_equal 'from-config-file'
+    #       # TODO
+    #     end
+    #   end
+    # end
+
+    describe 'when reading from the default location' do
+      # Should read from File.join(config_dir_path, 'fakehome-2', '.inspec', 'config.json')
+      let(:env) { { 'HOME' => File.join(config_dir_path, 'fakehome-2') } }
```
clintoncwolfe, line 604: Here's how we currently handle ENV vars with parallelized tests: since these are *functional* tests, they're all running `inspec` subprocesses - so we only set env vars in the subprocess invocation, and don't touch the test harness's ENV.

test/unit/base_cli_test.rb
```diff
@@ -7,105 +7,19 @@
 describe 'BaseCLI' do
   let(:cli) { Inspec::BaseCLI.new }
 
-  describe 'opts' do
```
clintoncwolfe, line 10: These tests were moved to test/unit/config_test.rb

test/unit/profiles/profile_test.rb
```diff
@@ -90,6 +90,7 @@
     let(:profile_id) { 'windows-only' }
 
     it 'loads our profile but skips loading controls' do
+      skip 'Mock loader always supports all platforms - bad test, ref #3750 '
```
clintoncwolfe, line 93: I found a bug - I'm not sure this test was being run, as I don't think this test would have passed prior to this PR (or if it did, I have no idea how).

clintoncwolfe (commented):
I've gone through the PR and left some notes, hopefully that should reduce the load!

clintoncwolfe:
Integration tests under kitchen are failing due to needing the updated Train branch https://github.com/inspec/train/pull/394 - see https://travis-ci.org/inspec/inspec/jobs/485232092#L1959

lib/inspec/backend.rb
```diff
@@ -41,16 +41,16 @@ def inspect
...
+      transport_name = Train.validate_backend(train_credentials)
```
tas50, line 45: thank you 1,000 times for using less generic variable names throughout this PR

miah (approved):
This is great. Thanks @clintoncwolfe

lib/inspec/dsl.rb
```diff
@@ -3,6 +3,7 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 require 'inspec/log'
+require 'stringio'
```
jerryaldrichiii: Is this require needed? I thought it is over in Inspec::Config.

test/functional/inspec_archive_test.rb
jerryaldrichiii: Should/Could we create an issue?

EDIT: Hmmm...this was supposed to be a response to your comment regarding it always clobbering.

lib/inspec/config.rb
```diff
@@ -0,0 +1,380 @@
...
+      cfg_io = File.open(path) if path
+      cfg_io || StringIO.new('{ "version": "1.1" }')
+    end
+
+    def determine_cfg_path(cli_opts)
+      path = cli_opts[:config] || cli_opts[:json_config] # TODO: deprecate --json-config see #3661
+
+      if path == '-'
+        Inspec::Log.warn 'Reading JSON config from standard input' if STDIN.tty?
+        path = STDIN
```
jerryaldrichiii: I know this was copied over, but in this case STDIN is a IO object and not a string. So the path returned doesn't behave as expected in `resolve_cfg_io`. See final comment for a patch that gives an example of how I made reading from STDIN possible and added a test which also "fixes" this.

EDIT: s/see final comment/see above/

jerryaldrichiii (changes requested):
This looks good, save for the fact that I couldn't get running from a config passed via STDIN to work.  Attached is a patch that made it work for me.

This was the command that I was using to test:

```
cat test/unit/mock/config_dirs/json-config/good.json | inspec exec test/unit/mock/profiles/simple-metadata --config -
```

[allow_stdin.txt](https://github.com/inspec/inspec/files/2816042/allow_stdin.txt)

test/functional/inspec_archive_test.rb
clintoncwolfe: Thanks for the nudge, I've opened https://github.com/inspec/inspec/issues/3774

lib/inspec/config.rb
clintoncwolfe: *Very* good catch, @jerryaldrichiii - I temporarily added your proposed test (but not the fix yet, in the spirit of TDD) and I got:
```
Expected "/Users/cwolfe/sandbox/inspec/inspec-train/lib/inspec/config.rb:150:in `initialize': no implicit conversion of IO into String (TypeError)
```

Right on the money.

lib/inspec/config.rb
```diff
@@ -0,0 +1,391 @@
...
+      cfg_io || StringIO.new('{ "version": "1.1" }')
+    end
+
+    def check_for_piped_config(cli_opts)
+      cli_opt = cli_opts[:config] || cli_opts[:json_config]
+      Inspec.deprecate(:cli_option_json_config, '') if cli_opts.key?(:json_config)
+
+      return nil unless cli_opt
+      return nil unless cli_opt == '-'
+      # This warning is here so that if a user invokes inspec with --config=-,
```
jerryaldrichiii, line 161: Ah! That's why.

lib/inspec/config.rb
```diff
@@ -0,0 +1,391 @@
...
+
+      cfg_io = File.open(path) if path
+      cfg_io || StringIO.new('{ "version": "1.1" }')
+    end
+
+    def check_for_piped_config(cli_opts)
+      cli_opt = cli_opts[:config] || cli_opts[:json_config]
+      Inspec.deprecate(:cli_option_json_config, '') if cli_opts.key?(:json_config)
+
+      return nil unless cli_opt
```
jerryaldrichiii, line 159: Curious why both of these are needed. Wouldn't the second guard be sufficient because if it is nil it will evaluate to `false`?

lib/inspec/config.rb
```diff
@@ -0,0 +1,391 @@
...
+
+      path = determine_cfg_path(cli_opts)
+
+      cfg_io = File.open(path) if path
+      cfg_io || StringIO.new('{ "version": "1.1" }')
+    end
+
+    def check_for_piped_config(cli_opts)
+      cli_opt = cli_opts[:config] || cli_opts[:json_config]
+      Inspec.deprecate(:cli_option_json_config, '') if cli_opts.key?(:json_config)
```
jerryaldrichiii, line 157: I feel like this should be moved outside of this method. It feels out of place since it has nothing to do with a piped config.

Maybe `resolve_cfg_io` or higher up?

lib/inspec/config.rb
```diff
@@ -0,0 +1,391 @@
...
+      return nil unless cli_opt == '-'
+      # This warning is here so that if a user invokes inspec with --config=-,
+      # they will have an explanation for why it appears to hang.
+      Inspec::Log.warn 'Reading JSON config from standard input' if STDIN.tty?
+      STDIN
+    end
+
+    def determine_cfg_path(cli_opts)
+      path = cli_opts[:config] || cli_opts[:json_config]
+      Inspec.deprecate(:cli_option_json_config, '') if cli_opts.key?(:json_config)
```
jerryaldrichiii, line 169: Same as other feedback, is there a place to put this to so we don't have to repeat it?

test/unit/profiles/profile_test.rb
jerryaldrichiii: Missed this on the other review. Could you open up an issue?

jerryaldrichiii (approved):
Looks good to me. Nothing that would block a merge, just a few questions.

lib/inspec/config.rb
clintoncwolfe: I'm calling Inspec.deprecate in both places where the code would have to be altered when the deprecation is completed, and the fucntionality removed.

lib/inspec/config.rb
clintoncwolfe: You're right, `nil == '-'` does evaluate to false, so we could rely on the second guard only. I didn't consider that, and we could make that change, in the spirit of brevity.

I think of guards like this: one by one, think of each reason you can get out of doing the work you were asked to do. So, here, I have two reasons: They didn't pass in a --config option at all (guard 1) or they did use a --config option but it was not '-'.  Two different reasons, so two different guards, for clarity.

I usually favor clarity over brevity, though sometimes the more concise version is clearer.

lib/inspec/config.rb
clintoncwolfe: My intent is indeed to repeat it. `Inspec.deprecate` takes over the job of `# TODO: deprecate ...` comments, which we would have had no concerns with repeating each place the feature exists. When the time comes to remove that feature, we don't want to do a bunch of tracing to determine all changes that must be made; instead, rely on the easy grep-ability of the deprecation group name, `cli_option_json_config`.

test/unit/profiles/profile_test.rb
clintoncwolfe: Yes, https://github.com/inspec/train/issues/405

lib/inspec/config.rb
jerryaldrichiii: Hmm...never heard that argument before. I really don't have an opinion one way or another given this new info :smile:. Not like this will affect performance.

PR #3751, opened 2019-01-25
Add CLI options to disable loading plugins

clintoncwolfe:
From time to time, it is useful to disable loading user or core plugins, especially during testing. This PR does that by adding two new options, `--disable-user-plugins` and `--disable-core-plugins`.  

Disabling user plugins is useful when your plugin setup is corrupted. This would allow (for example) `inspec plugin uninstall inspec-bad-plugin --disable-user-plugins`, without resorting to locating and removing the user plugin directory.

Disabling core plugins is needed when testing plugins that may talk to each other - in this case, to isolate user attribute provider plugins during testing.

Simple tests are provided to ensure the two options work, and that they are included in usage help.

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -32,6 +32,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :interactive, type: :boolean,
     desc: 'Allow or disable user interaction'
 
+  class_option :disable_core_plugins, type: :string, banner: '', # Actually a boolean, but this suppresses the creation of a --no-disable...
+    desc: 'Disable loading all plugins that are shipped in the lib/plugins directory of InSpec. Useful in development.'
+
+  class_option :disable_user_plugins, type: :string, banner: '',
+    desc: 'Disable loading all plugins that the user installed.'
+
   desc 'json PATH', 'read all tests in PATH and generate a JSON summary'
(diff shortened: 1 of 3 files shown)


lib/inspec/plugin/v2/plugin_types/cli.rb
```diff
@@ -1,5 +1,13 @@
 require 'inspec/base_cli'
 
+# Under some conditions (unit testing in particular) we may end
```
jerryaldrichiii: I think we may have a merge conflict with: https://github.com/inspec/inspec/pull/3758/files#diff-0fdac3e925822c85c27f47879d14c973R3

I'm fine with either message but I prefer the one line approach over there. Up to you though. I have no strong opinion.

jerryaldrichiii (approved):
Looks good to me. Usually I'm against more CLI arguments, but I see these being useful.

lib/inspec/plugin/v2/plugin_types/cli.rb
clintoncwolfe: AHA! Well, we found the PR in which I hit that issue, at least. We'll work out any conflict.

miah (approved):

clintoncwolfe:
Bumping minor for new CLI features

PR #3757, opened 2019-01-26
Add `--help` to CLI plugin activation criteria

### LINKED ISSUE #3753: 'inspec --help' does not load plugins, so no usage message
### 🎛 Description

🐛 bug report

When I run ```inspec --help``` the existence of init command is not even mentioned

If I run ```inspec help```, then it is visible.

3.2.6

jerryaldrichiii:
This ensures that all the following result in the same CLI output:
  - `inspec`
  - `inspec help`
  - `inspec --help`

This closes #3753.

### DIFF
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -89,13 +89,9 @@ module Inspec::Plugin::V2
         # all following ||= ops.
         activate_me = false
 
-        # If the user invoked `inspec help`, activate all CLI plugins, so they can
-        # display their usage message.
-        activate_me ||= cli_args.first == 'help'
-
-        # Likewise, if they invoked simply `inspec`, they are confused, and need
-        # usage info.
-        activate_me ||= cli_args.empty?
+        # If the user invoked `inspec help`, `inspec --help`, or only `inspec`
+        # then activate all CLI plugins so they can display their usage message.
+        activate_me ||= ['help', '--help', nil].include?(cli_args.first)
 
         # If there is anything in the CLI args with the same name, activate it.
         # This is the expected usual activation for individual plugins.
--- a/test/functional/inspec_test.rb
+++ b/test/functional/inspec_test.rb
@@ -28,4 +28,16 @@ describe 'command tests' do
       out.exit_status.must_equal 0
     end
   end
+
+  describe 'help' do
+    it 'outputs the same message when using `help`, `--help`, or no args' do
+      outputs = [
+        inspec('help').stdout,
+        inspec('--help').stdout,
+        inspec('').stdout,
+      ]
+
+      outputs.uniq.length.must_equal 1
+    end
+  end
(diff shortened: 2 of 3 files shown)


lib/inspec/plugin/v2/loader.rb
```diff
@@ -89,13 +89,9 @@ def activate_mentioned_cli_plugins(cli_args = ARGV)
...
-        # If the user invoked `inspec help`, activate all CLI plugins, so they can
-        # display their usage message.
-        activate_me ||= cli_args.first == 'help'
-
-        # Likewise, if they invoked simply `inspec`, they are confused, and need
-        # usage info.
-        activate_me ||= cli_args.empty?
+        # If the user invoked `inspec help`, `inspec --help`, or only `inspec`
+        # then activate all CLI plugins so they can display their usage message.
+        activate_me ||= ['help', '--help', nil].include?(cli_args.first)
```
clintoncwolfe, line 94: This seems a bit less clear, but I think we can live with it.

test/functional/inspec_test.rb
```diff
@@ -28,4 +28,16 @@
       out.exit_status.must_equal 0
     end
   end
+
+  describe 'help' do
+    it 'outputs the same message when using `help`, `--help`, or no args' do
```
clintoncwolfe: This test (though missing --help) already exists at https://github.com/inspec/inspec/blob/fd0ef925c4e32c68e42cbaf449f49b09b198a2f9/test/functional/plugins_test.rb#L71

Arguably it should be in `inspec_test.rb`, as you have it , but it shouldn't be in both places.  Could you consolidate them?

test/functional/inspec_test.rb
```diff
@@ -28,4 +28,16 @@
...
+      outputs = [
+        inspec('help').stdout,
+        inspec('--help').stdout,
+        inspec('').stdout,
+      ]
+
+      outputs.uniq.length.must_equal 1
```
clintoncwolfe, line 42: Ahh, this test says "all three of these should say the same message, but we don't care what the contents of the message are".  

In fact, we very much do care - the OP on #3753 reported that `init` was missing from the output.  You can do something like https://github.com/inspec/inspec/blob/fd0ef925c4e32c68e42cbaf449f49b09b198a2f9/test/functional/plugins_test.rb#L76 to verify that plugins are indeed being loaded.

clintoncwolfe (changes requested):
Thanks for the quick response on this, @jerryaldrichiii ; just need to consolidate a duplicated test, and make sure we are testing the right thing.  Thanks!

test/functional/inspec_test.rb
jerryaldrichiii: Sure can! Great catch.

clintoncwolfe (approved):
Thanks!

miah (approved):

PR #3758, opened 2019-01-26
tests: Fix profile_context_test NameError

### LINKED ISSUE #3721: Investigate failing test when using `m` and `test/unit/profiles/profile_context_test.rb`
Running the below

```
bundle exec m test/unit/profiles/profile_context_test.rb --line 114 
```

Errors with:

```
Inspec::ProfileContext::its default DSL#test_0007_does not provide the expect keyword in the global DLS [/home/jerry/Projects/github.com/inspec/inspec/test/unit/profiles/profile_context_test.rb:114]:
[NoMethodError] exception expected, not
Class: <NameError>
Message: <"uninitialized constant Inspec::Plugin\nDid you mean?  Inspec::Plugins">
---Backtrace---
/home/jerry/Projects/github.com/inspec/inspec/lib/inspec/profile_context.rb:158:in `instance_eval'
(eval):1:in `load_with_context'
/home/jerry/Projects/github.com/inspec/inspec/lib/inspec/profile_context.rb:158:in `instance_eval'
/home/jerry/Projects/github.com/inspec/inspec/lib/inspec/profile_context.rb:158:in `load_with_context'
/home/jerry/Projects/github.com/inspec/inspec/lib/inspec/profile_context.rb:144:in `load_control_file'
/home/jerry/Projects/github.com/inspec/inspec/test/unit/profiles/profile_context_test.rb:69:in `block in load'
---------------
```

This works when doing `bundle exec rake` so I suspect there is a path through the code that fails to load something but I couldn't find it after searching for 20 minutes.

jerryaldrichiii:
This closes #3721

### DIFF
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -31,6 +31,7 @@ module Inspec::DSL
   # This is called when an unknown method is encountered
   # "bare" in a control file - outside of a control or describe block.
   def method_missing(method_name, *arguments, &block)
+    require 'inspec/plugin/v2'
     # Check to see if there is a outer_profile_dsl plugin activator hook with the method name
     registry = Inspec::Plugin::V2::Registry.instance
     hook = registry.find_activators(plugin_type: :outer_profile_dsl, activator_name: method_name).first
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -3,6 +3,7 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 require 'inspec/log'
+require 'inspec/plugin/v2'
 
 module Inspec::DSL
   def require_controls(id, &block)
@@ -31,7 +32,6 @@ module Inspec::DSL
   # This is called when an unknown method is encountered
   # "bare" in a control file - outside of a control or describe block.
   def method_missing(method_name, *arguments, &block)
-    require 'inspec/plugin/v2'
     # Check to see if there is a outer_profile_dsl plugin activator hook with the method name
     registry = Inspec::Plugin::V2::Registry.instance
     hook = registry.find_activators(plugin_type: :outer_profile_dsl, activator_name: method_name).first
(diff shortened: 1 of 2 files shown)


lib/inspec/dsl.rb
```diff
@@ -31,6 +31,7 @@ def require_resource(options = {})
   # This is called when an unknown method is encountered
   # "bare" in a control file - outside of a control or describe block.
   def method_missing(method_name, *arguments, &block)
+    require 'inspec/plugin/v2'
```
clintoncwolfe: @jerryaldrichiii, can you explain why this location was chosen for the `require`? It would seem more typical to require in the top of `lib/inspec/dsl.rb`, or in `lib/inspec/profile_context.rb` (which is where the error appeared).

clintoncwolfe (changes requested):
Surprising location for a require; I'd want to know why (and explain with a comment).

lib/inspec/dsl.rb
jerryaldrichiii: Great call out. I could use some words on describing the behavior succinctly. I believe the issue is because we are doing an `instance_eval`. The only places I could get it to work were to place the require here: https://github.com/inspec/inspec/blob/master/lib/inspec/profile_context.rb#L152 or the place you see it. I did the latter since it is closer to https://github.com/inspec/inspec/blob/852085969fe85d7a69157eb4304e075625ca8e68/lib/inspec/dsl.rb#L36

Placing at the top of the files like you would expect leads me down a rabbit hole that I couldn't get to the bottom of (issues like not being able to find `Inspec::BaseCLI`).

clintoncwolfe: If you'd like to pair or group work on this, I'd be happy to do so; I'm familiar with the missing require for BaseCLI from the plugin/v2/plugin_types/cli type.  Simple fix there.

jerryaldrichiii: Oh? Great! I'll bring it up in standup and we can get some time scheduled if the usual parking lot time doesn't cover it.

clintoncwolfe (approved):
Thanks @jerryaldrichiii

miah (approved):

PR #3759, opened 2019-01-28
Allow :value in addition to :default for setting attribute values

### LINKED ISSUE #3756: Attributes: Rename 'default' option
### 🎛 Description

Currently, to set the value of an attribute in control code, you say:

```
attribute('user', default: 'bob')
```

Likewise, in the metadata file, if you set a value, you use:
```
attributes:
  user:
    default: bob
```

In addition, in control code, we have the `value=` method...

attribute('user').value = 'susan'

This sets up a scenario where it might imply that there is some kind of precedence going on, like the Chef Client's default/normal/override/automatic. That has been a major source of confusion in the chef community, which we'd like to avoid.

Moving forward, we want to be clear that while there may be multiple sources for an attribute's value (--attrs, metadata, inline in code) the precedence is entirely determined by the source, not by some additional precedence level dimension.

v3.2.x

### 💁 Possible Solutions

Introduce 'value' as the only accepted option. Deprecate 'default' over the next two major versions.

Refs #3748

clintoncwolfe:
Closes #3756

This PR implements #3756, by using the word 'value' instead of 'default' to set attribute values. 

 * A deprecation group is added, action 'ignore'.
 * 'default' is still accepted, but internally is immediately converted to 'value'.
 * unit tests are added to check that the 'default' word still works
 * All test fixtures and examples are updated to use 'value'
 * The reference docs are updated.

### DIFF
--- a/test/unit/attributes/attribute_test.rb
+++ b/test/unit/attributes/attribute_test.rb
@@ -52,6 +52,16 @@ describe Inspec::Attribute do
       attribute.value = 'new_value'
       attribute.value.must_equal 'new_value'
     end
+
+    it 'accepts the legacy ":default" option' do
+      attribute = Inspec::Attribute.new('test_attribute', default: 'a_default')
+      attribute.value.must_equal 'a_default'
+    end
+
+    it 'accepts the legacy ":default" and ":value" options' do
+      attribute = Inspec::Attribute.new('test_attribute', default: 'a_default', value: 'a_value')
+      attribute.value.must_equal 'a_value'
+    end
   end
 
   describe 'validate required method' do
--- a/test/unit/attributes/attribute_test.rb
+++ b/test/unit/attributes/attribute_test.rb
@@ -6,65 +6,70 @@ require 'inspec/objects/attribute'
 describe Inspec::Attribute do
   let(:attribute) { Inspec::Attribute.new('test_attribute') }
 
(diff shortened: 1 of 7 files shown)


tas50 (approved):
Seems pretty straight forward and killing the name "default" which has a very specific meaning in Chef is great

test/unit/attributes/attribute_test.rb
```diff
@@ -6,65 +6,80 @@
...
 
-    it 'does not error if a value is specified' do
+    it 'does not error if a value is specified by value=' do
       attribute = Inspec::Attribute.new('test_attribute', required: true)
       attribute.value = 'test_value'
       attribute.value.must_equal 'test_value'
     end
 
     it 'returns an error if no value is set' do
+      # Assigning the cli_command is needed because in check mode, we don't error
```
jerryaldrichiii, line 80: Odd...I feel like we should error...

jerryaldrichiii (approved):

test/unit/attributes/attribute_test.rb
clintoncwolfe: In check mode (`inspec check`) it is assumed that we won't have a --attrs file passed in (we might not even support the option).

So, for better or worse, don't treat a missing-but-required attribute as an error in check mode.  We could change that policy, but remember things like A2 run check.

PR #3767, opened 2019-01-29
filesystem: Add 'type' and 'free' attributes

jmassardo:
This change updates the `filesystem` resource to add the `free` and `type` attributes.

* `free` is the amount of available space on the partition
* `type` is the file system type, i.e. NTFS, ext4, swap, etc.


jerryaldrichiii (approved):
Thanks for this @jmassardo! This looks good to me.

clintoncwolfe (changes requested):
Code looks solid and test coverage is good.  I think my only concern is about `free`; whenever you have a quantified property, there is the potential for confusion and off-by-a-factor-of-1024 errors.  What would you think about naming that property `kb_free`, `free_kb`, `kilobytes_free`, or similar?

jmassardo:
@clintoncwolfe I was following the precedent set on the `size` attribute. I documented the value of the `free` attribute in the documentation page.

docs/resources/filesystem.md.erb
```diff
@@ -25,12 +25,15 @@ A `filesystem` resource block declares tests for disk space in a partition:
...
+      its('free') { should be >= 5000000 }
+      its('type') { should cmp 'ext4' }
     end
 
 where
 
 * `filesystem('/')` states that the resource will look at the root (/) partition.
-* `size` is measured in kilobytes (KB).
+* `size` is the total partition size and is measured in kilobytes (KB).
+* `free` is the available space on the partition and is measured in kilobytes (KB).
```
clintoncwolfe, line 36: Based on further discussion, I believe that we cannot in fact know the units here for either `size` or `free`: the units may vary based on OS, disk size, fs type, and block size.

Based on that, I think releasing `free` at this time is not appropriate.  It would not be safe to do a comparison test (Imagine asking someone if you had $20 in your wallet, and you responded, "Well, I have 3 bills.").

We could however add `free_percent`, since while we don't know the units, we do know they will be the same for both `free` and `size`.

clintoncwolfe (changes requested):
As I said before, the code here is genreally solid, and we really appreciate the inclusion of docs and unit tests.

The implementation of `type` is great.

However, as implemented, `free` (and `size`, not introduced on this PR) have unpredictable units, and cannot be used in comparisons. That must be addressed, as people rely on InSpec to be correct.

clintoncwolfe:
Closing by adoption, #3778

PR #3773, opened 2019-01-30
Fix Inspec::Attribute.to_ruby and add unit test

james-stocks:
#3756 introduced a regression, in which calling `to_ruby` on an Attribute object would throw an exception because it calls a method, `default`, which was renamed.

This PR fixes that.

### DIFF
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -93,7 +93,7 @@ module Inspec
     def to_ruby
       res = ["#{ruby_var_identifier} = attribute('#{@name}',{"]
       res.push "  title: '#{title}'," unless title.to_s.empty?
-      res.push "  default: #{default.inspect}," unless default.to_s.empty?
+      res.push "  value: #{value.inspect}," unless value.to_s.empty?
       res.push "  description: '#{description}'," unless description.to_s.empty?
       res.push '})'
       res.join("\n")
--- a/lib/inspec/objects/attribute.rb
+++ b/lib/inspec/objects/attribute.rb
@@ -94,6 +94,10 @@ module Inspec
       res = ["#{ruby_var_identifier} = attribute('#{@name}',{"]
       res.push "  title: '#{title}'," unless title.to_s.empty?
       res.push "  value: #{value.inspect}," unless value.to_s.empty?
+      # to_ruby may generate code that is to be used by older versions of inspec.
+      # Anything older than 3.4 will not recognize the value: option, so
+      # send the default: option as well. See #3759
+      res.push "  default: #{value.inspect}," unless value.to_s.empty?
       res.push "  description: '#{description}'," unless description.to_s.empty?
       res.push '})'
       res.join("\n")
--- a/test/unit/attributes/attribute_test.rb
+++ b/test/unit/attributes/attribute_test.rb
@@ -225,4 +225,16 @@ describe Inspec::Attribute do
       attribute.send(:valid_numeric?, '1/2').must_equal false
     end
   end
+
+  describe 'to_ruby method' do
+    it 'generates the code for the attribute' do
+      attribute = Inspec::Attribute.new('application_port', description: 'The port my application uses', value: 80)
+      attribute.to_ruby.must_equal <<-RUBY.chomp
+attr_application_port = attribute('application_port',{
+  value: 80,
+  description: 'The port my application uses',
+})
+RUBY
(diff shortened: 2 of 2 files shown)


lib/inspec/objects/attribute.rb
```diff
@@ -93,7 +93,7 @@ def to_hash
     def to_ruby
       res = ["#{ruby_var_identifier} = attribute('#{@name}',{"]
       res.push "  title: '#{title}'," unless title.to_s.empty?
-      res.push "  default: #{default.inspect}," unless default.to_s.empty?
+      res.push "  value: #{value.inspect}," unless value.to_s.empty?
```
clintoncwolfe, line 96: Do you expect profiles generated by this to work on older InSpecs? If so, you should add an additional line:

```
res.push "  default: #{value.inspect}," unless value.to_s.empty?
```

This way any generated code will contain both a `value:` and a `default:` option; newer InSpecs will use `value:`, and older inspecs will use `default:`

test/unit/attributes/attribute_test.rb
```diff
@@ -225,4 +225,16 @@
       attribute.send(:valid_numeric?, '1/2').must_equal false
     end
   end
+
+  describe 'to_ruby method' do
+    it 'generates the code for the attribute' do
+      attribute = Inspec::Attribute.new('application_port', description: 'The port my application uses', value: 80)
+      attribute.to_ruby.must_equal <<-RUBY.chomp
```
clintoncwolfe: This is just a preference, but I find it is easier to maintain (as well as handle failures) when the individual things you are looking for are split out.  It also makes it less susceptible to whitespace changes. Like this:

ruby_code = attribute.to_ruby
ruby_code.must_include "attribute('application_port'"   # Should have the declaration
ruby_code.must_include 'value: 80'   # We care that we use the 'value' attribute
ruby_code.must_include 'default: 80'   # We care that we use the 'default' attribute

test/unit/attributes/attribute_test.rb
```diff
@@ -225,4 +225,16 @@
...
+attr_application_port = attribute('application_port',{
+  value: 80,
+  description: 'The port my application uses',
+})
+RUBY
+    end
```
clintoncwolfe, line 248: I think we might also want to try to eval the code.  We care that we not only generate lines with certain things in it, but that code be valid code.

clintoncwolfe (dismissed):
Great work!  I think 2 things are needed: 
1. We need to add `default:` to the gerneated code, **along** with the `value:`
2. I think we should probably attempt to eval (or just compile) the code in the unit test.

Thanks!

lib/inspec/objects/attribute.rb
james-stocks: Yes it will have to work on older InSpec versions to not be a breaking change. Both older and current InSpec should be good with the extra key being passed, right?

test/unit/attributes/attribute_test.rb
james-stocks: I think it's generally better to not get tripped up by whitespace changes - but in this case isn't correct whitespacing part of the responsibility of the method?

test/unit/attributes/attribute_test.rb
james-stocks: That sounds good. It looks like `eval` always returns `nil` and all I can test is that nothing raises.
It also looks like MiniTest has no assertion for "must_not_raise" and all I can do is call eval() bare and let its unhandled exception fail the test?

Let me know if you had a better approach in mind.

lib/inspec/objects/attribute.rb
clintoncwolfe: Yes.

test/unit/attributes/attribute_test.rb
clintoncwolfe: Yes, that's how it works - nothing is raised; and then I usually put a test afterwards, to ensure it worked.

jerryaldrichiii (approved):
Looks good to me! Thanks for your feedback @clintoncwolfe.

test/unit/attributes/attribute_test.rb
james-stocks: Awesome! Thank you!

james-stocks (commented):
+1 to Clinton's improvements on top of my initial attempt

miah (approved):

PR #3778, opened 2019-02-04
filesystem: Add free_kb, size_kb, type, and percent_free properties

clintoncwolfe:
This PR is an adoption of #3767.

In this PR:
 * A serious bug is found in the `size` property; on linux it reports in KB, on Windows it reports in GB; KB was the documented unit. As existing tests may rely on that behavior, the `size` property is scheduled for future deprecation. A warning is added to the docs.
 * `size_kb` is introduced, with guaranteed units.
 * The `free` property from #3767 is renamed `free_kb` and its units corrected.
 * `percent_free` is added, range 0..100
 * The `type` property introduced on #3767 is retained without changes
 * The unit tests are tightened to test the actual value returned, rather than simply `>= 1` (which allowed the GB/KB issue on windows to slip through).

Many thanks to @jmassardo for the original work on this to add `free` and `type`, and to @miah for finding the pre-existing issues with units.


docs/resources/filesystem.md.erb
```diff
@@ -41,11 +52,33 @@ The following examples show how to use this InSpec audit resource.
...
+
+    describe filesystem('/') do
+      its('free_kb') { should be >= 5000000 }
     end
 
 <br>
 
+### Test if the C:\ partition is NTFS
+
+    describe filesystem('c:\') do
```
jerryaldrichiii, line 70: Probably doesn't matter, but I'd capitalize `c` here.

lib/resources/filesystem.rb
```diff
@@ -6,12 +6,16 @@ class FileSystemResource < Inspec.resource(1)
...
     example "
       describe filesystem('/') do
-        its('size') { should be >= 32000 }
-        its('type') { should eq false }
+        its('size_kb') { should be >= 32000 }
+        its('free_kb') { should be >= 3200 }
+        its('type') { should cmp 'ext4' }
+        its('percent_free') { should be >= 20 }
       end
       describe filesystem('c:') do
```
jerryaldrichiii, line 14: Ah, well this is lowercase and no slash. I think we should standardize but don't care enough to hold a PR over it.

jerryaldrichiii (approved):
Looks good to me. Just a few comments.

jmassardo:
@clintoncwolfe this looks awesome! Thanks for taking across the finish line for us!

docs/resources/filesystem.md.erb
miah: Windows drive letters are case insensitive, capital might make it stand out a little in docs but its not a requirement. <3

lib/resources/filesystem.rb
miah: Agree that we should at least be consistent.

miah (approved):
Thanks @clintoncwolfe

PR #3784, opened 2019-02-06
Consolidate attribute-related test-fixture profiles

clintoncwolfe:
This test-only change moves the test fixture profiles that are used in attribute testing into one place.


clintoncwolfe:
I'm going to go ahead and merge this, as it is a testing-only change and just a file move.

PR #3785, opened 2019-02-07
inspec-habitat: create mock backend properly 

jerryaldrichiii:
With the changes in PR #3750 `Inspec::Backend.create` needs to support both being passed a Hash and being passed an Inspec::Config. This adds a line to convert a passed Hash to an Inspec::Config.

This also adds unit tests for Inspec::Backend because they were missing.

### DIFF
--- a/test/unit/backend_test.rb
+++ b/test/unit/backend_test.rb
@@ -1,5 +1,3 @@
-# encoding: utf-8
-
 require 'helper'
 
 describe 'Backend' do # rubocop:disable Metrics/BlockLength
--- a/test/unit/backend_test.rb
+++ b/test/unit/backend_test.rb
@@ -1,0 +1,102 @@
+# encoding: utf-8
+
+require 'helper'
+
+describe 'Backend' do # rubocop:disable Metrics/BlockLength
+  let(:backend) { Inspec::Backend.create(Inspec::Config.mock) }
+
+  describe 'create' do # rubocop:disable Metrics/BlockLength
+    it 'accepts a Hash' do
+      backend_from_hash = Inspec::Backend.create(backend: 'mock')
+      backend_from_hash.is_a?(Inspec::Backend::Base).must_equal true
+    end
+
+    it 'accepts an Inspec::Config' do
+      backend.is_a?(Inspec::Backend::Base).must_equal true
+    end
+
+    it 'raises an error if no transport backend can be found' do
+      Train.stub :create, nil do
+        err = proc { backend }.must_raise RuntimeError
+        err.message.must_equal "Can't find transport backend 'mock'."
+      end
+    end
+
+    it 'raises an error if no connection can be created' do
+      mock_transport = Minitest::Mock.new
+      mock_bad_connection = Minitest::Mock.new
+      mock_transport.expect :nil?, false
+      mock_transport.expect :connection, mock_bad_connection
+      mock_bad_connection.expect :nil?, true
+
+      Train.stub :create, mock_transport do
+        err = proc { backend }.must_raise RuntimeError
+        err.message.must_equal "Can't connect to transport backend 'mock'."
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


clintoncwolfe:
#3750 did a lot of work to *eliminate* places where `Inspec::Backend#create` was being passed a Hash. As this is all internal code and we own inspec-habitat, I'd expect inspec-habitat to be modified to pass a Config.

That lets us move forward in a Config-based world, without having to support two calling modes.

clintoncwolfe:
The Inspec::Backend unit tests are a great addition!

jerryaldrichiii:
Yup, I agree @clintoncwolfe. I'm taking a stab at reworking that plugin now actually. That change didn't seem like it belonged in this PR.

This method needs to support a Hash though. Otherwise, #3750 is a breaking change.

clintoncwolfe:
It's only a breaking change if we consider the interface to `Inspec::Backend` to be a published interface.  I don't; if people want to run InSpec from ruby, they are supposed to use `Inspec::Runner`; if people want to create Train transports, they should use `Train#create`.  None of that is clear from InSpec's docs, but it does reflect our integration with other tooling.

To me, the change that is needed here is https://github.com/inspec/inspec/blob/9d7ca85e7d409e7822e6036fbd4242a3b7647eaa/lib/plugins/inspec-habitat/lib/inspec-habitat/profile.rb#L106 should be changed to call `Config.mock`, like [here](https://github.com/inspec/inspec/pull/3750/files#diff-9d107f8e84fe350fa821fed1838dbda8R54).  This is indeed a bug - I missed a spot where we create a mock backend, because it is not tested (see #3786 for that).

jerryaldrichiii:
Fair enough. I'd like to see what we interfaces we consider public and what we don't documented at some point though. This will break any one who passes a Hash and I don't see an issue with converting it.

I don't think we pass a Hash with any of our products though.

jerryaldrichiii:
Been thinking a bit (thanks @clintoncwolfe for helping me at our standup). People should be using `Inspec::Runner` and that is the interface that should allow a Hash (it already does). As far as documenting public interfaces that seems futile. InSpec is fundamentally a CLI application, and if people are using it as a library (incorrectly) then likely they will encounter way more issues than just this.

test/unit/backend_test.rb
```diff
@@ -0,0 +1,100 @@
+# encoding: utf-8
```
miah: FYI, We don't need to add these top line utf-8 comments anymore. This changed with 2.0, all 'scripts' are considered to be utf-8 by default.

https://ruby-doc.org/core-2.1.4/Encoding.html#class-Encoding-label-Script+encoding

jerryaldrichiii: Ah! Awesome. Will fix.

test/unit/backend_test.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      cache_disabled_config = Inspec::Config.new(backend_cache: false)
+      cache_disabled_backend = Inspec::Backend.create(cache_disabled_config)
+      cache_disabled_backend.backend.cache_enabled?(:file).must_equal false
+      cache_disabled_backend.backend.cache_enabled?(:command).must_equal false
+    end
+
+    it 'enables caching when using a mock backend' do
+      backend.backend.cache_enabled?(:file).must_equal true
+      backend.backend.cache_enabled?(:command).must_equal true
```
miah, line 45: Can we change the name here so we don't have 'backend.backend' ?

clintoncwolfe (approved):
Thanks for the discussion and taking the other approach, @jerryaldrichiii .  It's good to have more test coverage, too.

miah (approved):

test/unit/backend_test.rb
jerryaldrichiii: To what though? I also struggled because I'd have to use a custom one or change the `let(:backend)` block.

Issue #3779, opened 2019-02-05
Attributes in inspec.yml of an overlay profile does not override attributes values in dependent profile


rx294:
### 🎛 Description
Attributes defined in inspec.yml of an overlay profile **does not** override attribute default values in the dependent profile.

The only way seems to be to use the old style `--attrs attributes.yml`

This is this a 🐛 bug report

The expected behavior is attributes defined in the inspec.yml of the overlay profile will override attrs values without having to use `--attrs`

### 🌍 InSpec and Platform Version
3.x

aaronlippold:
Please label as a bug. This will bring up two issues as I can see:

What is the load order and precedence. 

We will either have to set that the order of the listed depends will be last in first our or add a prop to the attribute to be 'overwrite' or 'priority' I think. The second option would be needed if we can't guarantee in what order the 'depends' are loaded or stacked.

I think there is an existing issue where @jerryaldrichiii or @jquick talked about this a bit in terms of precedence I will see if I can track it down.

clintoncwolfe:
I think we need to pin down the version in which this appeared - it's likely this was a regression from #3756 , released in v3.4.1

clintoncwolfe:
My understanding of the repro:

inspec version is 3.4.1 or greater
Profile W depends on profile B
Profile B contains statements in its control files of the form `attribute('an-attribute', default: 'inline-value')`
Profile W's metadata file, inspec.yml, contains

```attributes:
  - name: an-attribute
    default: value-from-wrapper-metadata
```

Upon execution, the expected value of `attribute('an-attribute')` is `value-from-wrapper-metadata`
Actual value is `inline-value`

rx294:
Hey @clintoncwolfe 
In the scenario I tried out both profile W and B and attributes defined inspec.yml like so
profile B
```
  - name: db_name
    description: 'name of the specific DB being evaluated within the MSSQL server'
    type: string
    default: 'value for B'
```

profile W 
```
  - name: db_name
    description: 'name of the specific DB being evaluated within the MSSQL server'
    type: string
    default: 'value for W'
```

and the control files grab the attribs using attribute('an-attribute')....we do not have a default value defined in the controls.

During the exec of Profile W ... attribute('db_name') is expected to return 'value for W'
but, currently its still returning 'value for B'

clintoncwolfe:
What happens if you use the YAML key `value` instead of `default`, in both places?

profile B
```
  - name: db_name
[...]

PR #3788, opened 2019-02-08
Add functional tests to test combinations of attribute precedence

clintoncwolfe:
Refs #3779 

I've written a set of functional tests to check all combinations of attribute precedence I could think of, along with what I think is the expected behavior.

Currently, it fails 11 of the 23 tests.  We should discuss.


clintoncwolfe:
This is generally covered by expanded tests on #3875 - I'm closing this

PR #3791, opened 2019-02-08
Don't read the configuration twice

stevendanna:
We recently found that reading configuration from stdin is broken as
of 3.6.  This appears to be the result of attempting to read the
configuration twice. Since STDIN may be a pipe, reading from it a
second time yields nothing, leading to a json parse error:

```
Failed to load JSON configuration: 765: unexpected token at ''
Config was:
```

This is a likely incomplete fix. The gist here is that we are calling
Inspec::Config.new twice, leading to multiple reads of STDIN.  Here is
output from a instrumented inspec showing the issue:

```
RETURNING STDIN: #<IO:0x00007f7ef00a1920>
READING STDIN from /Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/lib/inspec/config.rb:38:in `initialize'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/lib/inspec/base_cli.rb:210:in `new'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/lib/inspec/base_cli.rb:210:in `config'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/lib/inspec/cli.rb:277:in `detect'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
/Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/bin/inspec:12:in `<top (required)>'
/opt/chefdk/embedded/bin/inspec:23:in `load'
/opt/chefdk/embedded/bin/inspec:23:in `<main>'
RETURNING STDIN: #<IO:0x00007f7ef00a1920>
READING STDIN from /Users/ssd/.chefdk/gem/ruby/2.5.0/gems/inspec-3.6.2/lib/inspec/config.rb:38:in `initialize'
[...]

PR #3793, opened 2019-02-08
inspec detect: Fix for --config trying to read from STDIN twice

### LINKED ISSUE #3792: Piping config broken on certain CLI commands in v3.6.2
### 🎛 Description

🐛 bug report

@stevendanna reports on #3791 that when reading config from STDIN, certain code paths read STDIN twice, resulting in a failed JSON parse on the second attempt.  The parse exception is caught and the run is gracefully aborted, but expected behavior is to read the config and operate normally.

Config is intended to generally be safe to instantiate multiple times; this is relied upon in several code paths, including anything calling `run_command` (this includes `detect` and `shell`, possibly `check`, but apparently not `exec`; this means our functional test (which calls exec) did not catch it).

3.6.2, regressed via #3750 

### 🤔 Replication Case

```
echo '{"backend":"docker","host":"something","reporter":{"json":{}},"backend_cache":true}' | inspec detect --json-config=-
```

### 💁 Possible Solutions

Some alternatives:
 * Add a class var to Inspec::Config; read STDIN into it, and check there next time.
 * After reading STDIN, try to rewind it.
 * Trace each time we call Config.new() multiple times and adjust
 * Make Config a singleton

### 💻 Stacktrace

See #3791 

clintoncwolfe:
Fixes #3792 
Alternative to #3791 

This PR:
 * Expands functional testing to include piping config to `shell` and `detect` in addition to `exec`
 * Alters Inspec::Config to cache any reads from STDIN

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -177,8 +177,15 @@ module Inspec
       path
     end
 
+    # When reading STDIN, read it once into a class variable and cache it.
+    # Don't cache other IO objects though, as we use several different StringIOs
+    # during unit testing. Refs #3792
+    def self.stdin_contents # rubocop: disable Lint/IneffectiveAccessModifier
+      @stdin_content ||= STDIN.read
+    end
+
     def read_cfg_file_io(cfg_io)
-      contents = cfg_io.read
+      contents = cfg_io == STDIN ? self.class.stdin_contents : cfg_io.read
       begin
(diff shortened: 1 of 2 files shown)


jerryaldrichiii (approved):
Looks good to me once RuboCop is happy.

clintoncwolfe:
As this is needed for a hotfix release, and I've discussed the fix with several people, I'm moving forward with only one approval.

Issue #3746, opened 2019-01-22
Don't use hardcoded path for sysctl in kernel_parameter resource


mogthesprog:
### 🎛 Description

 🐛 bug report

I've recently started using the [devsec.io Linux-Baseline](https://github.com/dev-sec/linux-baseline) tests to test our linux infrastructure. The tests make heavy use of the `kernel_parameter` resource to test the linux kernel parameters, which under the hood uses `sysctl` to check kernel configuration.

It looks like the `kernel_parameter` resource [has a hardcoded path for sysctl](https://github.com/inspec/inspec/blob/7deed13425c299ded99ff6e4095b1e0db969a826/lib/resources/kernel_parameter.rb#L22), which isn't always the correct path for sysctl.

### 🌍 InSpec and Platform Version

```
$ inspec version
3.2.6
```

### 🤔 Replication Case

1. Clone the `linux-baseline` inspec tests
```
git clone https://github.com/dev-sec/linux-baseline.git /tmp/linux-baseline
```
1. Run the inspec tests on a linux system, in my case this is a remote system
```
inspec exec \
    -t ssh://${ssh_username}@${instance_ip} \
    --key-files=${ssh_private_key_file} \
    --sudo \
        /tmp/linux-baseline
```

### 💁 Possible Solutions

We could use the `PATH` to resolve a valid instance of `sysctl`, and maybe fall back to the current hardcoded path provided in the code already(?). It's important to note that some operating systems don't make the `sysctl` binary available to unprivileged users, so we'd also need to make sure that we're checking the privileged PATH when `--sudo` is passed.

### 💻 Stacktrace

No stacktrace, but you will see `nil` values where the kernel will always return an integer:

```
   ↺  sysctl-31b: Secure Core Dumps - dump path
      ↺  Skipped control due to only_if condition.
   ×  sysctl-32: kernel.randomize_va_space
      ×  Kernel Parameter kernel.randomize_va_space value should eq 2

   expected: 2
         got: nil
```

`nil` here should be an integer.

clintoncwolfe:
Would it make sense to broaden this issue to cover all `command` invocations when running under sudo?

clintoncwolfe:
attn @miah 	any thoughts?

mogthesprog:
Any news on this? it's a bit of a blocker for security testing if you can't use sysctl.

I may have time in a week or so to get this building locally and fix it myself if nobody gets to it. I'd rather get the thumbs up before spending time implementing it.

[...]

Issue #450, opened 2016-02-11
resource 'command' not including PATH items '/sbin' and '/usr/sbin'


jeremymv2:
When I use the 'command' resource I need to specify the entire path to the command when the command exists in /sbin or /usr/sbin or /usr/local/sbin.

With serverspec I can optionally define a PATH this way:
PATH environment variable
You can add PATH environment variable in spec_helper.rb or other places like this.
`set :path, '/sbin:/usr/local/sbin:$PATH'`

Is this something that can be added to inspec? This could make migrating existing serverspec tests to inspec easier for folks.

chris-rock:
@jeremymv2 Thank you very much for bringing this topic up. I have some questions, I'd like to understand:

Is the PATH set, when you log into the machine directly? Could you please elaborate a little bit more, why you need the PATH? Is the reason multi os support? May it happen, that the command is available in various locations on the same systems?

I get the use case and I can see value in adding a feature that helps to address the topic. We need to find some answer before we are implementing a solution:
- Is the path only valid for one, multiple or a group of rules?
- setting `path` vs `mybin='/sbin/command'; describe command(mybin) ...`
- how is this portable across multiple operating systems? The path may be dependent on the OS, but in some cases may not?

jeremymv2:
@chris-rock 
- Testing primarily with the centos-6.6 and centos-7.1 images I see that the PATH for the 'vagrant' user does include the sbin dirs: `/usr/local/sbin:/usr/sbin:/sbin`.
- The command resource for things in sbin dirs only seems to work if I specify the full path like: `describe command('/usr/sbin/sestatus -v') do`
- The `sestatus` command only exists as `/usr/sbin/sestatus`
- With serverspec I can utilize a spec_helper.rb that has `set :path, '/sbin:/usr/local/sbin:/usr/sbin:$PATH' unless os[:family] == 'windows'`
- When I set the PATH just in one place in the spec_helper.rb for serverspec, I can then utilize it across all of my spec files and rules.
- For organizations that have many spec files, this is helpful to them to only have to define the PATH in one place when converting a lot of serverspec rules to inspec

jerryaldrichiii:
Quick notes that may help. All of the below is executed on a CentOS 7.3 node.

It seems like executing with `--sudo` will "fix" the issue, but the real issue is related to `PATH`

## SSH remote command (without sudo)
```
[...]

PR #3800, opened 2019-02-11
Search the PATH for executables on many resources

mogthesprog:
As the title suggests.

More details in #3746 

On doing this more formally across the board, would it be a case of following the same pattern here? or would there be a better way to do this?

We can either leave this as a small PR, or fix all of the executables now?

A quick stab with grep gives me the below?

```
morganj@localhost> grep -rn "/bin\|/usr\|/sbin" lib/resources
lib/resources/docker_container.rb:23:        its('command') { should eq 'nc -ll -p 1234 -e /bin/cat' }
lib/resources/ksh.rb:13:        its('stdout') { should match /bin/ }
lib/resources/ksh.rb:19:      ksh('...', path: '/usr/bin/ksh93')
lib/resources/kernel_module.rb:45:        lsmod_cmd = '/sbin/lsmod'
lib/resources/kernel_module.rb:85:        '/sbin/modinfo'
lib/resources/kernel_module.rb:93:        '/sbin/modprobe'
lib/resources/pip.rb:20:      describe pip('django', '/path/to/virtualenv/bin/pip') do
lib/resources/iptables.rb:73:      %w{/usr/sbin/iptables /sbin/iptables iptables}.each do |cmd|
lib/resources/zfs_dataset.rb:27:      inspec.command("/sbin/zfs get -Hp all #{@zfs_dataset}").exit_status == 0
lib/resources/zfs_dataset.rb:40:      cmd = inspec.command("/sbin/zfs get -Hp all #{@zfs_dataset}")
lib/resources/gem.rb:27:                        '/opt/chef/embedded/bin/gem'
lib/resources/gem.rb:30:                      '/opt/opscode/embedded/bin/gem'
lib/resources/aide_conf.rb:14:        its('selection_lines') { should include '/sbin' }
lib/resources/aide_conf.rb:17:      describe aide_conf.where { selection_line == '/bin' } do
lib/resources/parse_config.rb:6:#  audit = command('/sbin/auditctl -l').stdout
lib/resources/zfs_pool.rb:26:      inspec.command("/sbin/zpool get -Hp all #{@zfs_pool}").exit_status == 0
lib/resources/zfs_pool.rb:34:      cmd = inspec.command("/sbin/zpool get -Hp all #{@zfs_pool}")
lib/resources/nginx.rb:15:      describe nginx('/etc/sbin/') do
lib/resources/nginx.rb:24:    def initialize(nginx_path = '/usr/sbin/nginx')
lib/resources/file.rb:194:        perm_cmd = "su -s /bin/sh -c \"test -#{flag} #{path}\" #{user}"
lib/resources/auditd.rb:33:      unless inspec.command('/sbin/auditctl').exist?
lib/resources/auditd.rb:35:              'Command `/sbin/auditctl` does not exist'
lib/resources/auditd.rb:38:      auditctl_cmd = '/sbin/auditctl -l'
lib/resources/auditd.rb:73:      @status_content ||= inspec.command('/sbin/auditctl -s').stdout.chomp
lib/resources/bash.rb:13:        its('stdout') { should match /bin/ }
lib/resources/bash.rb:19:      bash('...', path: '/bin/bash')
lib/resources/command.rb:12:        its('stdout') { should match /bin/ }
lib/resources/users.rb:115:  #   its('shell') { should eq '/bin/bash' }
lib/resources/users.rb:127:  #   it { should have_login_shell '/bin/bash' }
lib/resources/users.rb:408:      # returns: root:x:0:0:root:/root:/bin/bash
lib/resources/users.rb:541:      # returns: root:*:0:0:Charlie &:/root:/bin/csh
lib/resources/package.rb:196:      brew_path = inspec.command('brew').exist? ? 'brew' : '/usr/local/bin/brew'
lib/resources/service.rb:111:      #   Systemd runs with PID 1 as /sbin/init.
lib/resources/service.rb:112:      #   Upstart runs with PID 1 as /sbin/upstart.
lib/resources/service.rb:114:      # Upstart runs with PID 1 as /sbin/init.
lib/resources/service.rb:135:          SysV.new(inspec, service_ctl || '/usr/sbin/service')
lib/resources/service.rb:142:          SysV.new(inspec, service_ctl || '/sbin/service')
lib/resources/service.rb:160:          SysV.new(inspec, service_ctl || '/sbin/service')
```

Let me know.

### DIFF
--- a/lib/utils/which.rb
+++ b/lib/utils/which.rb
@@ -1,0 +1,16 @@
+# encoding: utf-8
+
+# Cross-platform way of finding an executable in the $PATH.
+#
+#   which('ruby') #=> /usr/bin/ruby
+def which(cmd)
+  exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+  ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
+    exts.each { |ext|
+      exe = File.join(path, "#{cmd}#{ext}")
+      return exe if File.executable?(exe) && !File.directory?(exe)
+    }
+  end
+
+  raise Inspec::Exceptions::ResourceFailed, "Command not found: `#{cmd}`"
+end
--- a/lib/resources/auditd.rb
+++ b/lib/resources/auditd.rb
@@ -4,6 +4,7 @@ require 'forwardable'
 require 'utils/filter_array'
 require 'utils/filter'
 require 'utils/parser'
+require 'utils/which'
 
 module Inspec::Resources
   class AuditDaemon < Inspec.resource(1)
@@ -30,12 +31,12 @@ module Inspec::Resources
     "
 
     def initialize
-      unless inspec.command('/sbin/auditctl').exist?
+      unless inspec.command(auditctl_os_cmd).exist?
         raise Inspec::Exceptions::ResourceFailed,
-              'Command `/sbin/auditctl` does not exist'
+              "Command `#{auditctl_os_cmd}` does not exist"
       end
 
-      auditctl_cmd = '/sbin/auditctl -l'
+      auditctl_cmd = "#{auditctl_os_cmd} -l"
       result = inspec.command(auditctl_cmd)
 
       if result.exit_status != 0
@@ -54,6 +55,10 @@ module Inspec::Resources
       parse_content
     end
 
+    def auditctl_os_cmd
+      which('auditctl')
+    end
+
     filter = FilterTable.create
     filter.register_column(:file,         field: 'file')
           .register_column(:list,         field: 'list')
@@ -70,7 +75,7 @@ module Inspec::Resources
     filter.install_filter_methods_on_resource(self, :params)
 
     def status(name = nil)
-      @status_content ||= inspec.command('/sbin/auditctl -s').stdout.chomp
+      @status_content ||= inspec.command(auditctl_os_cmd).stdout.chomp
 
       # See: https://github.com/inspec/inspec/issues/3113
       if @status_content =~ /^AUDIT_STATUS/
--- a/lib/resources/iptables.rb
+++ b/lib/resources/iptables.rb
@@ -1,5 +1,7 @@
 # encoding: utf-8
 
+require 'utils/which'
+
 # Usage:
 # describe iptables do
 #   it { should have_rule('-P INPUT ACCEPT') }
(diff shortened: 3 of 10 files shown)


mogthesprog:
Any chance I can get someone to take a look at this?

mogthesprog:
I'm unsure why the tests above are failing actually.

Is it because my `which()` function isn't working well with the `MockLoader`?

looks that way from the below in `helper.rb`

```
      # zfs output for pool tank
      '/sbin/zpool get -Hp all tank' => cmd.call('zpool-get-all-tank'),
```

zpool may not be resolving correctly now, since i'm using which to discover the zpool path... 

Any guidance on how i can update MockLoader to support this?

mogthesprog:
just incase you hadn't seen this... @clintoncwolfe

lib/utils/which.rb
```diff
@@ -0,0 +1,16 @@
+# encoding: utf-8
+
+# Cross-platform way of finding an executable in the $PATH.
+#
+#   which('ruby') #=> /usr/bin/ruby
+def which(cmd)
```
clintoncwolfe, line 6: I don't think we want this implemented as a global method.  If this were in a module, say `Inspec::Util::Which` which provided a method `which` (as defined below) it would be cleaner.

Then, in each resource (or even in the resource superclass) we could do `include Inspec::Util::Which` without defining it everywhere on all objects. Surely some object somewhere implements an unrelated `which`, and it would be bad to potentially clobber it.

lib/utils/which.rb
```diff
@@ -0,0 +1,16 @@
...
+  exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+  ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
+    exts.each { |ext|
+      exe = File.join(path, "#{cmd}#{ext}")
+      return exe if File.executable?(exe) && !File.directory?(exe)
+    }
+  end
+
+  raise Inspec::Exceptions::ResourceFailed, "Command not found: `#{cmd}`"
```
clintoncwolfe, line 15: Refs #3174

lib/utils/which.rb
clintoncwolfe, line 6: Also, this seems like a great candidate for Memoize. Its return value should be the same for any given input, and running it again is potentially expensive.

clintoncwolfe, line 6: It might be good to take an optional second argument, providing a search path and defaulting to ENV['PATH'].  Refs #450, attn @miah

clintoncwolfe (changes requested):
Thanks so much for this, it's a great idea.  

So yeah, there are a many test failures that need to be addressed.
Also, we'd prefer `which` to be made available as a mixin.
Finally, a couple of thoughts around enhancing `which`, but those ideas are optional.

Thanks again, and apologies for the delay in review.

lib/utils/which.rb
miah: Taking a option second argument would definitely be helpful.

clintoncwolfe:
As this PR has extensive test failures, and would also require a fair bit of rework to math our desired design, I'm going to close this for now. Please do re-open if you have time in the future.

Issue #3802, opened 2019-02-11
Rename attributes to inputs


clintoncwolfe:
### 🎛 Description

🙋 feature request

Per #3748, InSpec `attributes` will be renamed to be InSpec `inputs` to avoid confusion with Chef attributes.

Thusly:

 * [x] Rename test files - #3796 
 * [x] Rename code files - #3803
    * [x] Check for external inclusion from other projects
 * [x] Add deprecation group - #3805
 * [x] Rename classes and non-DSL methods #3811
 * [x] Decide what to do about DEFAULT_ATTRIBUTE #3832 
 * [ ] Create aliases for user-visible features, with deprecation hooks
   * [x] DSL methods #3853
   * [x] --attrs #3855
   * [ ] audit cookbook `default['audit']['attributes'] https://github.com/chef-cookbooks/audit/issues/360
   * [x] kitchen-inspec `verifier`/`attributes` option https://github.com/inspec/kitchen-inspec/issues/224
   * [x] Metadata attributes #3854
 * [x] Update docs to mention 'inputs' in addition to attributes #3880
 * [ ] Update examples and code generator templates #3878
 * [ ] Update Learn Chef materials #3881
 * [ ] Alter `attributes` field in output of `inspec schema` (major version + 1)
 * [ ] Update json reporter to use inputs
 * [ ] Enable deprecation warnings (major version + 1)
 * [ ] Update docs to remove mention of attributes (major version + 2)
 * [ ] Remove aliases (attributes stop working) (major version + 3)

luckylittle:
I use `inspec 4.16.0` and i've got this: `input('cis_level', value: 2)` in my controls, but it is complaining that: `WARN: Input 'cis_level' does not have a value. Use --input-file to provide a value for 'cis_level' or specify a value with attribute('cis_level', value: 'somevalue', ...)`. Which one should be used then? `input` or `attribute`?

clintoncwolfe:
We'll address your issue on #4445 - thanks @luckylittle

PR #3796, opened 2019-02-08
Attributes-Inputs Rename: Move test files

clintoncwolfe:
As part if the renaming of 'attributes' to 'inputs', this PR:

 * Moves the test fixture profiles
 * Renames the unit test files
 * Renames the unit tests
 * updates the test files to use the new file locations

Part of #3802

PR #3803, opened 2019-02-11
Attribute-Input Rename: Rename codefiles

clintoncwolfe:
As part of #3802, this PR renames the files that have attribute in their name.  Their contents are not altered, other than updating `require` statements.

This PR is dependent on #3796.


miah (approved):
Thanks @clintoncwolfe !

jerryaldrichiii (approved):

PR #3805, opened 2019-02-11
Attribute-Input Rename: Add deprecation group

clintoncwolfe:
As part of #3802, this simple PR adds a deprecation group to the list of deprecations, so that future code changes can reference it.

This PR does not depend on anything, but is not needed until after #3803.


miah (approved):
Thanks @clintoncwolfe !

jerryaldrichiii (approved):

Issue #3761, opened 2019-01-28
Rename DEFAULT_ATTRIBUTE to ATTRIBUTE_WITH_NO_VALUE


clintoncwolfe:
### 🎛 Description

🙋 feature request

The InSpec attribute system allows you to have an attribute with no value set. If you stringify the attribute, you'll get output indicating including the name of the attribute and a message that it has no value.

If you perform type checks agains the value of a valueless attribute, you'll find it is of the class DEFAULT_ATTRIBUTE. This is unfortunate, as the word `default` was previously used to assign values to attributes (though see #3759). Some profiles explicitly check for this.

The meaning of the class is more like ATTRIBUTE_WITHOUT_VALUE - a placeholder value that lets you make method calls, etc without hitting a no-method-on-NilClass error.

It also allows the user to explicitly set nil, if desired.

### 🌍 InSpec and Platform Version

3.3.14

### 🤔 Replication Case

```
describe attribute('myattr') do
  its('to_s') { should include "Attribute 'myattr' does not have a value. " }
  it { should be_kind_of Inspec::Attribute::DEFAULT_ATTRIBUTE }
end
```

### 💁 Possible Solutions

Consider renaming the class to something that is more clearly about the lack of value.

Consider adding a `.has_value?` predicate.

clintoncwolfe:
Fixed on #3832

PR #3811, opened 2019-02-13
Attribute->Input Rename:  Rename Classes and Methods

clintoncwolfe:
This PR performs a sweeping internal rename of attribute-related functionality.  Both Class names and method names are changed, replacing `s/attribute/input/`.  

The replacement was performed manually and with great care to ensue that no public interfaces were changed - especially not:
 * The DSL (`attribute()` exists, `input()` does not yet)
 * Metadata still uses `attribute:` in YAML
 * The CLI option is still `--attrs`
 * Options to Runner are still `:attribute`

Replacement was performed under `lib/` and `test/`. No docs changes are needed at this time, nor examples.

Part of #3802.

### DIFF
--- a/lib/inspec/objects/input.rb
+++ b/lib/inspec/objects/input.rb
@@ -3,19 +3,8 @@
 require 'utils/deprecation'
 
 module Inspec
-  class Input
-    attr_accessor :name
-
-    VALID_TYPES = %w{
-      String
-      Numeric
-      Regexp
-      Array
-      Hash
-      Boolean
-      Any
-    }.freeze
-
+  # TODO: move this
+  class Attribute
     DEFAULT_ATTRIBUTE = Class.new do
       def initialize(name)
         @name = name
@@ -40,6 +29,21 @@ module Inspec
         "Input '#{@name}' does not have a value. Skipping test."
       end
     end
+  end
+
+  class Input
+    attr_accessor :name
+
+    VALID_TYPES = %w{
+      String
+      Numeric
+      Regexp
+      Array
+      Hash
+      Boolean
+      Any
+    }.freeze
+
 
     def initialize(name, options = {})
       @name = name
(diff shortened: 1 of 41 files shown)


lib/inspec/objects/input.rb
```diff
@@ -3,19 +3,8 @@
 require 'utils/deprecation'
 
 module Inspec
+  # TODO: move this
```
jerryaldrichiii, line 6: Could you link an issue or add some more context?

test/unit/resources/postgres_hba_conf_test.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: utf-8
 # copyright: 2017
 # author: Aaron Lippold, lippold@gmail.com
-# author: Rony Xavier, rx294@nyu.edu 
+# author: Rony Xavier, rx294@nyu.edu
```
jerryaldrichiii, line 4: Hmm...not sure how I feel about this. I'd like to remove the line, but it is an external contributor.

jerryaldrichiii (approved):
Looks good to me, just 2 tiny comments.

test/unit/resources/postgres_hba_conf_test.rb
lamont-granquist: All this is doing is just fixing the whitespace.

I'd actually like to nuke all the author lines across the board from the whole chef-client codebase.  I never add them anymore for myself.

lib/inspec/objects/input.rb
clintoncwolfe: It will be moved on the next PR, when #3761 is implemented.

test/unit/resources/postgres_hba_conf_test.rb
jerryaldrichiii: Yeah, I was referring to the fact that usually we remove these lines...but not sure in this case due to it being an external contributor.

lamont-granquist (approved):

PR #3814, opened 2019-02-14
Update the integration tests to use dokken-images and newer Chef technologies

tas50:
Modernize how the integration tests run to utilize the changes in Chef over the last few years.

- Drop dependencies on yum/apt/windows/openssl cookbooks and instead use built in resources
- Update from Chef 14.3 -> 14.10
- Update from berkshelf 5 to 7
- Run the integration tests on Ruby 2.5 not 2.4
- Remove a lot of unnecessary chef code by using new helpers
- Remove 3 extra times we tries to setup apt in the various recipes
- Consume cookbooks from Supermarket vs. pulling from Github on each run
- Use bento slugs in the vagrant config so we always get the latest releases of each major distro release
- Update platform versions in the kitchen configs to the latest we support
- Use dokken-images which come with all the necessary files for VM like testing. This eliminates some of the setup that had to be done in the cookbook
- Update platforms we test on in Travis
  - Add Debian 9
  - Add Ubuntu 18.04
  - Add Oracle Linux 6
  - CentOS 6.8 -> 6.10
  - Debian 8.5 -> 8.11
  - Fedora 24 -> 29
  - Ubuntu 12.04 -> 14.04


.travis.yml
```diff
@@ -1,53 +1,67 @@
-sudo: required
```
tas50, line 1: Travis deprecated this config so there's no point in including it now

Gemfile
```diff
@@ -22,7 +22,9 @@ group :test do
 end
 
 group :integration do
-  gem 'berkshelf', '~> 5.2'
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.4')
+    gem 'berkshelf', '~> 7'
+  end
```
lamont-granquist, line 28: berks 7 brings in chef as a gem which breaks the spiffy ruby version stuff, so you probably do need to do this.

miah (approved):
Amazing. Thank you @tas50 !

Gemfile
```diff
@@ -22,7 +22,9 @@ group :test do
 end
 
 group :integration do
-  gem 'berkshelf', '~> 5.2'
+  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.4')
```
jerryaldrichiii, line 26: Could we add a comment and a date for removal (e.g. when 2.4 EOLs)

test/cookbooks/os_prepare/metadata.rb
```diff
@@ -5,11 +5,7 @@
 description 'This cookbook prepares the test operating systems'
 version '1.0.0'
 
-depends 'apt'
-depends 'yum'
 depends 'runit'
-depends 'postgresql'
-depends 'windows'
+depends 'postgresql', '= 7.1.3'
```
jerryaldrichiii, line 8: Could we add a comment for this specific pin?

jerryaldrichiii (approved):
Looks good to me once comments are added.

miah (approved):

PR #3820, opened 2019-02-20
Moved the AWS init profile to the correct location.

skpaterson:
The current location of the AWS profile seems distinct from OS and GCP.  Should we move it for consistency?
`lib/plugins/inspec-init/lib/inspec-init/templates/profiles/aws` -> `lib/plugins/inspec-init/templates/profiles/aws`


skpaterson:
Apologies, I'd edited the above comment and removed the original finding whilst investigating. Seemingly there is a regression on master currently for generating the AWS init profile e.g.

```
$ inspec init profile --platform aws aws-test
ERROR: Unable to generate profile: No template available for platform 'aws' (expected one of: gcp, os)
```

The AWS functional profile test also seems missing so I've re-added that.

clintoncwolfe (approved):
Great catch - thanks @skpaterson!

miah (approved):

Issue #3822, opened 2019-02-20
Externalize AWS resources, use AWS SDK 3


clintoncwolfe:
### 🎛 Description

This is an implementation issue to track work related to pushing the AWS resources out to a resource pack, and using AWS SDK v3.

Obsoletes #2571 
Obsoletes https://github.com/inspec/train/issues/350

 * [x] Revive https://github.com/inspec/inspec-aws - thanks to @skpaterson!
 * [x] Create inspec 4 branch (See 4-stable)
 * [x] Release train 2.0 with no AWS deps
 * [x] Release train-aws 0.1.0 with aws-sdk-3 deps
 * [x] On 4-stable branch, rely on train ~> 2.0 and train-aws ~> 0.1
 * [x] On 4-stable branch to retain `aws_` resources, but now against v3 #3823
 * [x] On 3-stable/master branch, add deprecation hooks in `ignore` mode #3824
 * [x] Migrate open PRs and issues to https://github.com/inspec/inspec-aws
 * [ ] On 4-stable branch, set deprecation group to warn
 * [x] On 4-stable branch, update `inspec init --platform aws` to generate profiles with a `depends` on the resource pack - #3831
 * [ ] On 4-stable branch, update docs/platforms

skpaterson:
Thanks for this @clintoncwolfe ! Couple of updates:
* Migrate open PRs and issues to https://github.com/inspec/inspec-aws  - we're currently working on open PRs and targeting early next week for transferring/closing out AWS related issues
* On 4-stable branch, update `inspec init --platform aws` to generate profiles with a `depends` on the resource pack - see #3831

btm:
@clintoncwolfe can this be closed now?

clintoncwolfe:
Yes!

PR #3823, opened 2019-02-20
v4: Support AWS SDK v3 in Core AWS Resources

clintoncwolfe:
NOTE: this PR targets version 4 of InSpec.

While AWS resources are being removed from core and placed in a resource pack (See #3822), we must retain them in core so we don't break profiles that have not yet been updated to use the resource pack.

This PR does two things:
 * Updates the `require` that previously pulled in `aws-sdk` v2, to pull in v3 and a handful of service gems
 * Suppresses the resource overwrite warning if the resource named is `aws_*`.  This quiets noise when a user DOES use the new resources.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,8 @@
 # encoding: utf-8
 source 'https://rubygems.org'
 
+gem 'train-aws', path: '../train-aws'
+
 gemspec name: 'inspec'
 
 gem 'ffi', '>= 1.9.14'
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,6 @@
 # encoding: utf-8
 source 'https://rubygems.org'
 
-gem 'train-aws', path: '../train-aws'
-
 gemspec name: 'inspec'
 
 gem 'ffi', '>= 1.9.14'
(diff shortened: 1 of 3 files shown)


miah (approved):

skpaterson (commented):
I'm assuming here that the policy is to retain the resources in the next major version rather than just removing them in v4 after the v3 deprecation warning?

clintoncwolfe:
Yes.  We are currently on version 3.x; so if we followed "strict" deprecation policy, we would warn on 4, fail on 5, and remove on 6.  

We could be more aggressive, and warn in v3, fail on 4, and remove in 5.

By "fail" I mean that we'd change the deprecation action from "ignore" to "warn" then "fail".  That would cause anyone who uses the old resource to have the control / test fail (or a warning issued, if outside a describe block).  In order for that to work, the resources need to remain in the core codebase for a few more major versions - if we literally removed them, anyone using inspec without the resource pack would see a "No such method aws_iam_groups" error, which would be a bad experience.

skpaterson (approved):
Great, thanks @clintoncwolfe 👍

PR #3824, opened 2019-02-20
Add a deprecation group for core AWS resources

clintoncwolfe:
As part of #3822, this PR adds an inactive deprecation group, then adjusts the AWS resource mixin to call the deprecation hook when a core AWS resource is accessed.

While this PR has the deprecation set to 'ignore' (so it behaves silently), here is sample output when set to 'warn' mode:

screen shot 2019-02-20 at 18 06 01
screen shot 2019-02-20 at 18 06 16


miah (approved):

skpaterson (approved):
👍

PR #3827, opened 2019-02-21
add json example from http request

jtimberman:
I was looking for an example of how to do this, but I didn't find
anything in the documentation. This is a simple examle in the `json`
resource that returns a JSON blob back from an `http` request to a
"`/health`" endpoint.

Signed-off-by: Joshua Timberman <joshua@chef.io>

### DIFF
--- a/docs/resources/json.md.erb
+++ b/docs/resources/json.md.erb
@@ -82,6 +82,24 @@ The `name` matcher tests the value of the filename as read from a JSON file vers
 
 <br>
 
+### Test JSON output from an HTTP API
+
+Our example API has a `/health` endpoint, which looks like this:
+
+    {
+      "service": {
+        "port": 3000,
+        "status": "ok"
+      }
+    }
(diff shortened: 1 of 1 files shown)


docs/resources/json.md.erb
```diff
@@ -82,6 +82,24 @@ The `name` matcher tests the value of the filename as read from a JSON file vers
...
+Our example API has a `/health` endpoint, which looks like this:
+
+    {
+      "service": {
+        "port": 3000,
+        "status": "ok"
+      }
+    }
+
+Example test:
```
clintoncwolfe, line 96: Just a thought, but it might be nice to link to the `rspec-its` docs, which provide detail on what to pass to `its`.

clintoncwolfe (approved):
Great example, thanks! One minor suggestion but nothing to stop a merge.

miah (approved):

Issue #3715, opened 2019-01-10
Use Inspec::UI consistently


clintoncwolfe:
### 🎛 Description

#3618 introduced a facility for standardizing UI interactions, but did not apply those changes to the InSpec codebase.

So... let's use Inspec::UI throughout the codebase.

Consider carefully any changes that would be breaking, and make those on a separate PR.

### 🌍 InSpec and Platform Version

3.2.x

### 🤔 Replication Case

List key points here?

### Related issues

#1700, #3648, #3580, #3556, #3513, #2384, #3818

wyardley:
@clintoncwolfe thanks for this.
Question though, it doesn't seem like `lib/inspec/ui.rb` currently has any kind of support for color terminals that don't support 256 colors, so that even if we switched to this, #1700 would still be an issue, right?

miah:
I've fixed most of these in #4000, but there are 3 remaining:

```
  2) Failure:
cli command plugins#test_0001_is able to respond to a plugin-based cli subcommand [/home/miah/projects/github/inspec_cli/test/functional/plugins_test.rb:63]:
--- expected
+++ actual
@@ -1 +1,2 @@
-""
+"[2019-04-30T11:15:39-07:00] WARN: DEPRECATION:  Please call `cli.ui` directly
+"


  3) Failure:
inputs::run profile with yaml inputs#test_0004_errors with required input not defined [/home/miah/projects/github/inspec_cli/test/functional/inputs_test.rb:103]:
--- expected
+++ actual
@@ -1 +1,2 @@
-""
+"[2019-04-30T11:19:49-07:00] WARN: DEPRECATION:  Please call `cli.ui` directly
+"


  4) Failure:
inputs::run profile with yaml inputs#test_0003_errors with invalid input types [/home/miah/projects/github/inspec_cli/test/functional/inputs_test.rb:93]:
--- expected
+++ actual
@@ -1 +1,2 @@
-""
+"[2019-04-30T11:19:49-07:00] WARN: DEPRECATION:  Please call `cli.ui` directly
+"

```

wyardley:
I don't think I can reopen it myself; @clintoncwolfe is this still something that might happen someday/

PR #3829, opened 2019-02-22
Use deprecation facility throughout code

### LINKED ISSUE #3745: Use deprecation facility throughout code
### 🎛 Description

🙋 feature request?

Throughout the code, we have comments and warning messages mentioning to deprecate a feature in the future.  With #3690, we now have a centralized deprecation facility.

 * Search throughout the codebase and find deprecation comments and warning messages.
 * Create groups in etc/deprecations.json
 * Replace the deprecation comments with calls to Inspec.deprecate

3.2.x

jerryaldrichiii:
This converts all current deprecation warnings/TODOs to use the `Inspec.deprecate()` deprecation facility.

This also modifies `Inspec.deprecate()` to only require 1 argument.

This closes #3745

### DIFF
--- a/etc/deprecations.json
+++ b/etc/deprecations.json
@@ -1,29 +1,108 @@
 {
   "file_version": "1.0.0",
-  "unknown_group_action": "ignore",
+  "unknown_group_action": "exit",
   "groups": {
     "attrs_value_replaces_default": {
       "action": "ignore",
       "prefix": "The 'default' option for attributes is being replaced by 'value' - please use it instead."
+    },
+    "aws_resources_in_resource_pack": {
+      "comment": "See #3822",
+      "action": "ignore",
+      "prefix": "AWS resources shipped with core InSpec are being to moved to a resource pack for faster iteration. Please update your profiles to depend on git@github.com:inspec/inspec-aws.git ."
     },
     "cli_option_json_config": {
       "action": "ignore",
       "prefix": "The --json-config option is being replaced by the --config option.",
       "comment": "See #3661"
     },
-    "filesystem_property_size": {
+    "file_resource_be_mounted_matchers": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "host_resource_proto_usage": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "inspec_ui_methods": {
+      "action": "ignore",
+      "suffix": "Please call `cli.ui` directly"
+    },
+    "mssql_session_pass_option": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "oracledb_session_pass_option": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "property_filesystem_size": {
       "action": "ignore",
       "comment": "See #3778"
+    },
+    "property_processes_list": {
+      "action": "warn",
+      "suffix": "This property will be removed in InSpec 4.0."
+    },
+    "properties_aws_iam_user": {
+      "action": "warn",
+      "suffix": "This property will be removed in InSpec 4.0."
+    },
+    "properties_shadow": {
+      "action": "warn",
+      "suffix": "This property will be removed in InSpec 3.0."
     },
     "rename_attributes_to_inputs": {
       "action": "ignore",
       "prefix": "InSpec Attributes are being renamed to InSpec Inputs to avoid confusion with Chef Attributes.",
       "comment": "See #3802"
     },
-    "aws_resources_in_resource_pack": {
-      "comment": "See #3822",
-      "action": "ignore",
-      "prefix": "AWS resources shipped with core InSpec are being to moved to a resource pack for faster iteration. Please update your profiles to depend on git@github.com:inspec/inspec-aws.git ."
+    "resource_apache": {
+      "action": "warn",
+      "suffix": "This resource will be removed in InSpec 4.0."
+    },
+    "resource_azure_generic_resource": {
+      "action": "warn",
+      "prefix": "The azure_generic_resource is deprecated. Please use a specific resource. See: 'https://github.com/inspec/inspec/issues/3131'"
+    },
+    "resource_iis_website": {
+      "action": "warn",
+      "suffix": "This resource will be removed in InSpec 4.0.",
+      "comment": "Needed for ServerSpec compatibility"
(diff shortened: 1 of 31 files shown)


etc/deprecations.json
```diff
@@ -1,29 +1,108 @@
 {
   "file_version": "1.0.0",
-  "unknown_group_action": "ignore",
+  "unknown_group_action": "exit",
```
clintoncwolfe, line 3: Did you intend to commit this?  I often set this during development, but I think in a release it would be safer to have this be `ignore`.

etc/deprecations.json
```diff
@@ -1,29 +1,108 @@
...
+    "file_resource_be_mounted_matchers": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "host_resource_proto_usage": {
+      "action": "warn",
+      "suffix": "This will not be supported in InSpec 4.0."
+    },
+    "inspec_ui_methods": {
+      "action": "ignore",
```
clintoncwolfe, line 28: might add comment refs #3715

lib/inspec/config.rb
```diff
@@ -155,7 +155,7 @@ def resolve_cfg_io(cli_opts, cfg_io)
 
     def check_for_piped_config(cli_opts)
       cli_opt = cli_opts[:config] || cli_opts[:json_config]
-      Inspec.deprecate(:cli_option_json_config, '') if cli_opts.key?(:json_config)
+      Inspec.deprecate(:cli_option_json_config) if cli_opts.key?(:json_config)
```
clintoncwolfe, line 158: 💯

miah (approved):
Legit. Thanks @jerryaldrichiii

lib/matchers/matchers.rb
```diff
@@ -107,6 +107,7 @@
 # Deprecated: You should not use this matcher anymore
 RSpec::Matchers.define :be_running do
   match do |service|
+    Inspec.deprecate(:serverspec_compatibility, 'The service `be_running?` matcher is deprecated.')
```
clintoncwolfe, line 110: I think this group name should probably be more specific. While we are reprioritizing serverspec compatibility, I don't think we're ripping it out all at once, either.

lib/resources/aws/aws_iam_user.rb
```diff
@@ -22,7 +22,7 @@ class AwsIamUser < Inspec.resource(1)
   alias has_console_password? has_console_password
 
   def name
-    warn "[DEPRECATION] - Property ':name' is deprecated on the aws_iam_user resource.  Use ':username' instead."
+    Inspec.deprecate(:properties_aws_iam_user, 'The aws_iam_users `name` property is deprecated. Please use `username` instead')
```
clintoncwolfe, line 25: Typo in the message here - this is in the `aws_iam_user` resource, but teh message refers to the plural.

lib/resources/filesystem.rb
```diff
@@ -52,7 +52,7 @@ def size_kb
     end
 
     def size
-      Inspec.deprecate(:filesystem_property_size, 'The `size` property did not reliably use the correct units.  Please use `size_kb` instead.')
```
clintoncwolfe, line 55: I don't know that it's worth renaming them all, but RESOURCENAME_WHAT_NAME reads better to me than WHAT_RESOURCENAME_NAME.

lib/resources/host.rb
```diff
@@ -49,7 +49,7 @@ def initialize(hostname, params = {})
       @port = params[:port]
 
       if params[:proto]
-        warn '[DEPRECATION] The `proto` parameter is deprecated. Use `protocol` instead.'
+        Inspec.deprecate(:host_resource_proto_usage, 'The host resource `proto` parameter is deprecated. Please use `protocol`.')
```
clintoncwolfe, line 52: proto_param - these are called resource parameters.

lib/resources/users.rb
```diff
@@ -213,42 +213,38 @@ def warndays
 
     # implement 'mindays' method to be compatible with serverspec
     def minimum_days_between_password_change
-      deprecated('minimum_days_between_password_change', "Please use: its('mindays')")
+      Inspec.deprecate(:resource_user, 'The user resource `minimum_days_between_password_change` property is deprecated. Please use `mindays`.')
```
clintoncwolfe, line 216: I'd change all of these to something like :user_resource_serverspec_compat .  Makes the intent and scope more explicit.

lib/utils/parser.rb
```diff
@@ -88,7 +88,7 @@ def parse_mount_options(mount_line, compatibility = false)
       # parse options as array
       mount_options[:options] = mount[5].gsub(/\(|\)/, '').split(',')
     else
-      # parse options as serverspec uses it, tbis is deprecated
+      Inspec.deprecate(:serverspec_compatibility, 'Parsing mount options in this fashion is deprecated')
```
clintoncwolfe, line 91: This should def. be more specific.

test/helper.rb
```diff
@@ -649,3 +649,13 @@ def load_resource(*args)
   m = MockLoader.new(:ubuntu1404)
   m.send('load_resource', *args)
 end
+
+# Used to capture `Inspec.deprecate()` with warn action
+def expect_deprecation_warning
+  @mock_logger = Minitest::Mock.new
```
clintoncwolfe, line 655: Does this actually need to be an instance variable?

test/helper.rb
```diff
@@ -649,3 +649,13 @@ def load_resource(*args)
   m = MockLoader.new(:ubuntu1404)
   m.send('load_resource', *args)
 end
+
+# Used to capture `Inspec.deprecate()` with warn action
+def expect_deprecation_warning
```
clintoncwolfe, line 654: We might want to make this looser - like `expect_deprecation_action(opts)`.  With no args, it expects that `Inspec.deprecate` will be called at least once.  With args, we might insist on a particular action, or restrict which group names we expect to see.

The goal here would be that a deprecation will naturally progress through ignore, warn, fail/exit over its life, letting us change the test code less frequently.  If we use a more general check like this, we would not be insisting on a warning, just a check that we know it is scheduled for deprecation.

@miah, thoughts?

clintoncwolfe (changes requested):
Thanks for all this @jerryaldrichiii - we'll be adopting it.  

Most of my changes are around specificity of the deprecation group names; and an idea around reducing the amount of test changes we'd need to do.

clintoncwolfe:
Superceded by #3952

PR #3831, opened 2019-02-22
Update AWS profile generation for InSpec 4 to depend on the new resou…

skpaterson:
…rce pack.

NOTE: this PR targets version 4 of InSpec.

### DIFF
--- a/lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
+++ b/lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
@@ -7,7 +7,7 @@ aws_vpc_id = attribute('aws_vpc_id', default: '', description: 'Optional AWS VPC
 
 # You add controls here
 control 'aws-single-vpc-exists-check' do                                    # A unique ID for this control.
-  only_if { aws_vpc_id != ''}                                               # Only run this control if the `aws_vpc_id` atrtibute is provided.
+  only_if { aws_vpc_id != ''}                                               # Only run this control if the `aws_vpc_id` attribute is provided.
   impact 1.0                                                                # The criticality, if this control fails.
   title 'Check to see if custom VPC exists.'                                # A human-readable title.
   describe aws_vpc(aws_vpc_id) do                                           # The test itself.
--- a/lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
+++ b/lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
@@ -3,24 +3,28 @@
 
 title 'Sample Section'
 
-aws_vpc_id = attribute('aws_vpc_id')
+aws_vpc_id = attribute('aws_vpc_id', default: '', description: 'Optional AWS VPC identifier.')
 
-# you add controls here
-control 'aws-vpc-check' do                                                  # A unique ID for this control.
+# You add controls here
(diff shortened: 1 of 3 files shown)


lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
```diff
@@ -3,24 +3,28 @@
...
     it { should exist }
   end
 end
 
-# Plural resources can be inspected to check for specific resource details.
-control 'aws-vpcs-check' do
-  impact 1.0
-  title 'Check in all the VPCs for default sg not allowing 22 inwards'
-  aws_vpcs.vpc_ids.each do |vpc_id|
-    describe aws_security_group(vpc_id: vpc_id, group_name: 'default') do
```
clintoncwolfe: Aww, it saddens me that this test would be removed.  It's very practical, and shows the power of inspec as a security auditing tool. Is there a strong reason to remove this?

lib/plugins/inspec-init/templates/profiles/aws/inspec.yml
```diff
@@ -6,11 +6,15 @@ copyright_email: you@example.com
...
 summary: An InSpec Compliance Profile For AWS
 version: 0.1.0
-inspec_version: '>= 2.3.5'
+inspec_version: '~> 4'
 attributes:
 - name: aws_vpc_id
-  required: true
-  description: 'The Custom AWS VPC Id'
+  required: false
+  default: ''
```
clintoncwolfe, line 15: If you mean "there should be no value unless the user provides one", you should either remove this or comment it out. As-is, this will set an empty-string vpc-id, which is probably not what you want.

clintoncwolfe (commented):
Just a couple of comments.  Also, I need to re-base this properly.

lib/plugins/inspec-init/templates/profiles/aws/controls/example.rb
skpaterson: Happy to replace this.

lib/plugins/inspec-init/templates/profiles/aws/inspec.yml
skpaterson: This was so users didn't have to specify a VPC identifier e.g. 
```
   only_if { aws_vpc_id != ''}                                               # Only run this control if the `aws_vpc_id` atrtibute is provided.
```
There's a note and link to related documentation in the README also.

clintoncwolfe (approved):
Thanks!

lib/plugins/inspec-init/templates/profiles/aws/inspec.yml
miah: Can we put a comment with a link to that documentation above this. My worry here is with people copypasta'ing the example for eternity.

miah (changes requested):
Looks good, just want a minor change around documentation on inspec.yml in the profile template.

lib/plugins/inspec-init/templates/profiles/aws/inspec.yml
skpaterson: Thanks @miah I've added a comment to the `inspec.yml`.

miah (approved):
Thanks @skpaterson

PR #3836, opened 2019-02-26
Definition list formatting

mjingle:
Formatting config.md with definition list formatting.


clintoncwolfe (approved):
Thanks, looks great!

PR #3844, opened 2019-02-28
json resource: Add handling for `command:` error

jerryaldrichiii:
This provides the user with messages if `json(command: 'command_that_errors')` exits non-zero and/or has STDERR.

Without this the resource will report "No output from command" when really the output is contained in STDERR.

### DIFF
--- a/test/unit/resources/json_test.rb
+++ b/test/unit/resources/json_test.rb
@@ -86,7 +86,7 @@ describe "Inspec::Resources::JSON" do
     end
   end
 
-  describe "#load_raw_from_file" do
+  describe "#load_raw_from_command" do
     let(:cmd_str)  { "curl localhost" }
     let(:resource) { Inspec::Resources::JsonConfig.allocate }
     let(:inspec)   { mock }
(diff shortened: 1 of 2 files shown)


lib/resources/json.rb
```diff
@@ -96,10 +96,25 @@ def load_raw_from_file(path)
     end
 
     def load_raw_from_command(command)
-      command_output = inspec.command(command).stdout
-      raise Inspec::Exceptions::ResourceSkipped, "No output from command: #{command}" if command_output.nil? || command_output.empty?
+      command_result = inspec.command(command)
 
-      command_output
+      case command_result.exit_status
```
clintoncwolfe: I don't think this should look at exit code at all. There are tools that indicate status by exit code, and non-zero does not mean "failure" (`grep` is an example).

lib/resources/json.rb
```diff
@@ -96,10 +96,25 @@ def load_raw_from_file(path)
...
+      when 0
+
+        # TODO: Handle case where command exits 0 but STDERR is found.
+
+        if command_result.stdout.empty?
+          raise Inspec::Exceptions::ResourceSkipped, "No output from command: #{command}"
```
clintoncwolfe: Might be good to leave a TODO here in each place you raise an exception - something like `TODO: Handle resource unable to run, see #3174"

test/unit/resources/json_test.rb
```diff
@@ -100,18 +100,27 @@
       inspec.expects(:command).with(cmd_str).returns(command)
     end
 
-    it 'raises an exception if command stdout is nil' do
-      command.expects(:stdout).returns(nil)
+    it 'skips the resource if command exits zero and stdout is empty' do
```
clintoncwolfe: As above, I think this is over-specified - just check for STDERR, and report the error code in the message.

clintoncwolfe (changes requested):
Great catch! I think we shouldn't introduce checking for exit code, and just rely on the output.  Also small request to add some comments.

jerryaldrichiii:
Hmmm...good point. I'll just include the exit code in any output/warnings.

EDIT: Thinking about this and there may be cases where STDERR is not empty but command exited 0. Examples like warn/debug messages (I think our json reporter does that even)

clintoncwolfe:
The issue here is only around what to do when STDOUT is empty, not STDERR.

```
 if STDOUT is not empty
   parse it as JSON
 else
   Build an error message, using anything we can scrounge from stderr; include exit code as well
 end
```

jerryaldrichiii:
Hey all, given a job change I won't be able to commit to finishing this PR in a timely manner.

Feel free to adopt :smile:

test/unit/resources/json_test.rb
```diff
@@ -86,30 +86,61 @@
...
+      resource = run_json_cmd "echo 'not-valid-json'"
+
+      assert_resource_failed resource, "unexpected token at 'not-valid-json"
     end
 
-    it "returns the command output" do
-      command.expects(:stdout).returns("json goes here")
-      resource.send(:load_raw_from_command, cmd_str).must_equal "json goes here"
+    it "empty stdout, message in stderr" do
+      resource = run_json_cmd "echo bad args 1>&2 && false"
```
clintoncwolfe: Looks like the `&&` breaks the windows tests...

clintoncwolfe (changes requested):
Code looks great, just have a broken test to fix and then should be all set.

test/unit/resources/json_test.rb
zenspider: uuuugh. THAT thing again. I'll figure it out.

zenspider: This is blocked by #4465, which I just filed.

miah (approved):

clintoncwolfe (approved):
Thanks!

PR #3858, opened 2019-03-01
Force omnibus to link to train 1.7.4

clintoncwolfe:
Without this, the omnibus build defaults to Train 2.0.0, which is not compatible with inspec 3.x .


clintoncwolfe:
Packaging merge, self-approving.

PR #3860, opened 2019-03-02
Move all gem installation to Gemfile/bundle install

lamont-granquist:
Do not use omnibus-software gem definitions for installing gems,
and use the Gemfile in the inspec project as the source of truth.


lamont-granquist:
ChefDK does this correctly:

https://github.com/chef/chef-dk/blob/7279ee80d1eaa5044b6af853f4014b3606bed42d/omnibus/config/software/chef-dk.rb#L33-L45

(do not look at chef/chef it does not do this correctly)

lamont-granquist:
Most of the rest of our software projects like this check in their Gemfile.lock so that git manages their pins that are ultimately used when bundle installing in omnibus builds to make things deterministic and checked-in to git.  In practice of course we tend to merge Gemfile.lock bumps with zero testing against any sort of actual omnibus build in jenkins-CI so builds break all the time anyway, so that's more an FYI...

If you wanted to get really advanced you could commit a Gemfile.lock here though and then have a travis job that kicked off an omnibus build against the travis instance, then run the test suite against that built omnibus image.  Wouldn't catch if some gem bump only broke Solaris or something, but would catch breakage on Ubuntu.

Gemfile
```diff
@@ -1,11 +1,17 @@
 # encoding: utf-8
 source 'https://rubygems.org'
 
-gemspec name: 'inspec'
```
tas50, line 4: Isn't gemspec the proper way to source the local gem or do we have to do this here?

lamont-granquist: this was necessary, although i can't quite recall what blew up now.  copied this from the chef/chef Gemfile.

clintoncwolfe: Note that we have two gemspecs, because we build two gems - inspec and inspec-core.  I think this line was there mainly to pass along the gemspec name explicitly.

lamont-granquist: `gem "inspec", path: "."` only picks up the inspec gem out of the inspec.gemspec though.  the name is there explicitly.

clintoncwolfe: Perhaps I'm misunderstanding, but we also need to be able to build `inspec-core`, which has a separate gemspec (and different (lighter) dependencies). Would that still work?

lamont-granquist: You don't build that for including into the Gemfile though right?  Because you weren't including `inspec-core` here before either...

clintoncwolfe (approved):
Thanks!

Gemfile
miah: The Gemfile shouldn't interfere with inspec-core builds as we're building it with `gem build`.

miah (approved):
Thanks, the reason we had split these out was due to AIX incompatibilities where we required patched versions of `unf_ext`. We had a similar issue with `train` where we needed to follow a specific branch. I appreciate the cleanup =)

clintoncwolfe:
Looks like unf_ext has bit us again - this won't build now :-(

PR #3873, opened 2019-03-06
Add caching to Inspec::Config

clintoncwolfe:
Miah and I discovered that Config will usually want to act like a singleton; this adds a class-level cache which may be used to get a handle on the first instance.

Needed for both the Attributes and Telemetry work.


miah (approved):
Thanks!

PR #3875, opened 2019-03-07
Internal overhaul of Inputs system

### LINKED ISSUE #1059: Dependencies: All attributes are scoped
https://github.com/chef/inspec/issues/888

### LINKED ISSUE #3852: Inputs: Consolidate input access to Registry, and use value voting
### 🎛 Description

🙋 feature request

Currently, inputs (formerly known as `attributes`) are created in several locations, and precedence is determined by exection order.

To make a plugin system, we need to consolidate this, and enforce a new precedence system.

Tell us which version of InSpec (`inspec version` or SHA of a commit) and Operating System distribution you are using

### 🤔 Replication Case

### 💁 Possible Solutions

### 💻 Stacktrace

clintoncwolfe:
Closes #3852
Closes #1059 1059

This (admittedly large) PR overhauls the internals of the Input system, while greatly improving test coverage and isolation.

First, the manner in which Inputs are registered is altered.  Previously, Inputs would be registered on an ad-hoc basis, with the timing varied depending on which avenue the input used to be set. This impacted precedence greatly, and made it difficult to reason about or add new ways of setting Inputs. Additionally, some Inpuuts were not bound to any particular profile, which breaks our model. So, Inputs that have been provided by profile metadata, `--attrs` files, kitchen-inspec, the audit cookbook, or inline in DSL, are now registered during profile construction.

This solved a lot of issues, but made precedence harder. So, Inputs now each have an Event stream associated with them. Each time the Input has its value changed, an Event is recorded, along with a priority (0-100). To determine the current value, the event with the highest priority wins. I then set priorities as follows:
 * 20 - inline in control code
 * 30 - profile metadata
 * 40 - from CLI files like `--attrs`
 * 40 - from Runner API (kitchen-inspec or audit cookbook)

Inline calls and profile metadata files can explicitly set a priority on a per-input basis.

Events may also be used for diagnosis.

Tests are now more clearly isolated, so that scoping, validation, profile inheritance, and other features can be verified.


miah (approved):

aaronlippold:
What priorities do input set via the CLI have in the priority chain?

PR #3885, opened 2019-03-11
json reporter: Add network identifiers

kevin-j-smith:
inspec/train enhanced to add hostname, domainname, fqdn and network to the platform section of it's os export.  Added logic to properly handle and report these findings.

#3869

### DIFF
--- a/lib/inspec/formatters/base.rb
+++ b/lib/inspec/formatters/base.rb
@@ -68,6 +68,11 @@ module Inspec::Formatters
       run_data[:platform] = {
         name: platform(:name),
         release: platform(:release),
+        network: platform(:network),
+        hostname: platform(:hostname),
+        domain_name: platform(:domain_name),
+        fqdn: platform(:fqdn),
+        guid: platform(:guid),
         target: backend_target,
       }
     end
--- a/lib/resources/platform.rb
+++ b/lib/resources/platform.rb
@@ -19,7 +19,7 @@ module Inspec::Resources
     end
 
     # add helper methods for easy access of properties
-    %w{family release arch}.each do |property|
+    %w{family release arch network hostname fqdn domain_name guid}.each do |property|
       define_method(property.to_sym) do
         @platform[property]
       end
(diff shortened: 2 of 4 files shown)


kevin-j-smith:
This commit will be more impactful once inspec/train has accepted PR: inspec/train#418 https://github.com/inspec/train/pull/418

lib/inspec/formatters/base.rb
```diff
@@ -68,6 +68,11 @@ def stop(notification)
       run_data[:platform] = {
         name: platform(:name),
         release: platform(:release),
+        network: platform(:network),
```
clintoncwolfe, line 71: As commented on #3869, I really think this data should come from something like an Ohai resource (see #3886). You can access resources even in places like this by using the Resource Registry.

lib/resources/platform.rb
```diff
@@ -19,7 +19,7 @@ def initialize
     end
 
     # add helper methods for easy access of properties
-    %w{family release arch}.each do |property|
+    %w{family release arch network hostname fqdn domain_name guid}.each do |property|
```
clintoncwolfe, line 22: As mentioned elsewhere, I don't think this should be added to platform.

clintoncwolfe (changes requested):
Thanks for the contribution!

However, as discussed on #3869, we'd rather see the network identifiers collected by something an Ohai resource, or something else - not from platform.

Additionally, in its final form, we generally expect tests and documentation to be included with PRs.  Thanks!

clintoncwolfe:
(boilerplate message)
Hi! This PR has been untouched for more than two weeks, in the "Changes Requested" state. Please either make the requested changes, or continue the discussion in comments.

kekaichinose:

__Why do I care?__

__What do I need to do?__

kekaichinose:
Sorry, prematurely pulled the trigger on this. Reopening.

clintoncwolfe:
We're going to go ahead and close this, as it now has merge conflicts, hasn't been updated, and we've updated the originating issue with next steps. Thanks!

Issue #3845, opened 2019-02-28
Magically associate inputs with controls to force outcome


clintoncwolfe:
### 🎛 Description

Refs #3017 

One idea that comes up occasionally (see also the secondary topic on #3840) is a desire to alter the output of a control based on an input (formerly known as attributes). Today, you would need to have something that has a naming standard for inputs, explicitly examines the loaded controls, then call something like `skip_control` on each that was marked to skip or fail (or, from #3017, have a finer-grained outcome, like 'waivered' or 'exception' or 'POAM').

If we adopted a naming standard for the inputs, InSpec itself could handle the latter two parts.

### 🌍 InSpec and Platform Version

3.6.x

### 💁 Possible Solutions

Post ideas below for naming conventions.

clintoncwolfe:
Attn @TheLunaticScripter
Attn @mschmeitz

aaronlippold:
I like this idea: the one thing to keep in mind is that the above classes waiver, plan of action and milestones (POAM) and exception are all a reporting decisions that are separate from the test results themselves and for the most part we accreditors still want and should see the results of the tests by default regardless of if we decide in our assessment that the results of those tests do or do not have impact on the final evaluation.

This mainly says that we should not have the tests skipped by default - run the tests and show the result. In our reporter however we collect them in the 'Non-Applicable' group ( the general term for the above classes ) for reporting and the reporting impact should be none or 0. Lastly, I think in this class we could add an opt to say skip the tests but that must be a user set field and - I stress - not the default.
The example: We did an assessment on a system and for the test event and the Approving Official selected a set of tests that would be Non-Applicable. We provided a dependent profile that updated the desc of those controls to include agree additional verbiage of why they were Non-Applicable for reporting. The approving official agreed that these controls did not need to be 'assessed' - count against the final evaluation - however that they still wanted to know the state of those controls (aka the test results either good or bad) so that they could have a clear understanding of the relative risk of the system.

zenspider:
This ties to #4408.

clintoncwolfe:
Implemented on #4427

PR #3840, opened 2019-02-27
sort control filenames

mschmeitz:
Often controls filenames are numbered, sorting them makes the output more logical instead of randomized.

This also enables our use case: Running inspec with attributes file 'exceptions.yml' that contains exceptions. The profile controls folder contains files like file-01.rb, file-02.rb etc. and a zzSkip_Control.rb This last file contains no tests only logic that iterates over all controls, if a control is in the exceptions file it is skipped. This enables a multi layered model; controls defined in exceptions.yml are skipped during the profile handling, instead of the skip_control examples that require a 'master' profile to skip exceptions in 'child' profile.

Also, the 'master'-'child' model only can handle a single layer, the multi layer model can have multiple layers; layer 3 profile (the one that is executed) depends on layer 2 profiles, depends on layer 1 profiles. The skip_control is available on each layer when using the technique as described above.

Fixes #3797 (sort control files)

EDIT: link to open issue

PR #3887, opened 2019-03-13
inspec exec: new CLI option '--sort-files' to sort files before executing controls

mschmeitz:
Add parameter '--sort_files' for 'inspec exec' to sort files before executing controls.

Often controls filenames are numbered, sorting them makes the output
logical instead of randomized.

This also enables our use case: Running inspec with attributes file
'exceptions.yml' that contains exceptions. The profile controls folder
contains files like file-01.rb, file-02.rb etc. and a zzSkip_Control.rb
This last file contains no tests, only logic that iterates over all
controls; When a control is defined in an (--attrs) exceptions file it
will be skipped.

This enables a multi layered model; controls defined in exceptions.yml
are skipped during the profile handling, instead of the skip_control
examples that require a 'master' profile to skip exceptions in 'child'
profile.

Also, the 'master'-'child' model only can handle a single layer, the
multi layer model can have multiple layers; layer 3 profile (the one
that is executed) depends on layer 2 profiles, depends on layer 1
profiles. The skip_control is available on each layer when using the
technique as described above.

Signed-off-by: Mike Schmeitz <mike.schmeitz@ictelligent.nl>

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -117,6 +117,8 @@ module Inspec
         desc: 'Show progress while executing tests.'
       option :distinct_exit, type: :boolean, default: true,
         desc: 'Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures.'
+      option :sort_files, type: :boolean, default: false,
+        desc: 'Sort the files in the controls folder before execution. (default: false)'
     end
 
     def self.format_platform_info(params: {}, indent: 0, color: 39)
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -301,6 +301,9 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: 'A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell'
   option :distinct_exit, type: :boolean, default: true,
     desc: 'Exit with code 100 if any tests fail, and 101 if any are skipped but none failed (default).  If disabled, exit 0 on skips and 1 for failures.'
+  option :sort_files, type: :boolean, default: false,
+    desc: 'Sort the files in the controls folder before execution. (default: false)'
+
   def shell_func
     o = config
     diagnose(o)
--- a/lib/inspec/file_provider.rb
+++ b/lib/inspec/file_provider.rb
@@ -256,6 +256,10 @@ module Inspec
       parent.binread(abs_path(file))
     end
 
+    def sort_files
+      @files.sort!
+    end
+
     private
 
     def get_prefix(fs)
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -55,6 +55,7 @@ module Inspec
     def self.for_path(path, opts)
       file_provider = FileProvider.for_path(path)
       rp = file_provider.relative_provider
+      rp.sort_files if opts[:sort_files]
 
       # copy embedded dependencies into global cache
       copy_deps_into_cache(rp, opts) unless opts[:vendor_cache].nil?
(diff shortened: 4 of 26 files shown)


mschmeitz:
replaces #3840

clintoncwolfe (changes requested):
This looks pretty good, and is a straightforward implementation.  However, we need to have some functional testing on this.  Have a look at the functional tests in `test/functional/inspec_exec_test.rb` (and I would encourage you to use the `run_inspec_process` helper method, rather than the older `inspec` method).

Some cases I'd like to see covered:
 - a flat profile with ~10 control files executes them in no particular order when the --sort-files option is absent
 - a flat profile with ~10 control files executes them in collation order when the --sort-files option is present
 - a profile that wraps another profile loads files in ... some defined way when the --sort-files option is present.

Thanks! The third test case is especially important - we need to have well-defined behavior when adding new features.

clintoncwolfe:
(boilerplate message)

mschmeitz:
Added tests using dependent profiles: they failed! 
Haven't had (enough) time to fix this.  My time constraint and the fact that this sort option is not mandatory at this moment makes that I don't take action on this PR. From my side, this PR can be closed.

aaronlippold:
I don't see the tests pushed to the PR. If you push the tests - even if broken - I can take a look at fixing them.

mschmeitz:
@aaronlippold I have updated the PR and included the tests. There is a readme file that lists the structure, you should run the /test/unit/mock/profiles/sort-controls/profile_layer_top

My requirement is that files within a single profile (even when it is a dependency  profile), executes the tests in a sorted manner, where file sorting is sufficient.

Current setup only sorts the first profile but not its dependencies. I was not (yet?) able to parse the sort parameter to the profiles that are handled as dependency.

I do appreciate your help and am willing to anticipate if I can!

aaronlippold:
My guess it is when we have dependent profiles we have to have some sort of
recursion going on that you're calling sort on each dependent profile down
the chain and then finally sorting the primary profile.

aaronlippold:
Ping, I plan on putting folks on this in the next week.

mschmeitz:
It has been a while and I think I wanted to take a short cut with the 'sort-files' option. What our organization really needs is handling exceptions when using inherited profiles. 

The idea we have is to use a 3 layered model. 
- The 3rd and bottom layer contains the controls, for example the complete CIS baseline, additional SSH tests, tests for Jenkins, tests created by DevOps teams, tests downloaded from (and maintained on) the internet etc. 
- The 2nd layer 'cherry picks' these controls, filters unwanted tests, and bundles them together using skips/requires. 
- The 1st layer combines the profiles of the 2nd layer. So we fire one profile, that profile belongs to a specific type of infra component (call it building block, role, like a web server, proxy server etc). The web server profile contains 'CIS profile', 'webserver security profile', 'generic company wide security profile', 'webserver buildingblock profile' 

So for the infra component 'web server' a set of 2nd layer profiles is defined, one of these profiles contains 'CIS profile' with a, by the organization, defined set of controls. And, one control that must be applicable is 'no HTTP deamon'. This is fine on most servers but is not relevant (and conflicts) on web servers. So we either need to remove the 2nd layer (which is very cumbersome; when we want to add 1 control we need to update all the profiles). 

As an alternative we can define the controls as exceptions in an exception.yaml file. At execution we parse that file using the 'attributes' option and then it 'iterates' over *ALL* the controls (*in all profiles*), and assigns the exceptions as 'skip_control' to have no fails. (See the start of this thread, here's the need for sorting).  

But what we really need is that we can add exceptions, to filter controls that reside on the 3rd layer (i.e. on the child of the child of the parent profile).

Since that feature (gather all controls including inherited, and then filter based on exception.yaml) is far more complicated as 'sort files', and the fact that I can deliver my use case with 'sort files', my choice was to go for the 'sort files' option.

clintoncwolfe:
I can understand wanting `--sort-control-files` or whatever; but using this as a very roundabout way to get waivers / exceptions is pretty painful. 

See #3845

kekaichinose:

__Why do I care?__

__What do I need to do?__

PR #3889, opened 2019-03-14
Decoupling test profiles from example profiles

srb3:
Signed-off-by: devoptimist <sbrown@chef.io>

ref: https://github.com/inspec/inspec/issues/3351


test/functional/helper.rb
```diff
@@ -55,16 +55,15 @@ module FunctionalHelper
   let(:exec_inspec) { File.join(repo_path, 'bin', 'inspec') }
   let(:mock_path) { File.join(repo_path, 'test', 'unit', 'mock') }
   let(:profile_path) { File.join(mock_path, 'profiles') }
-  let(:examples_path) { File.join(repo_path, 'examples') }
   let(:integration_test_path) { File.join(repo_path, 'test', 'integration', 'default') }
 
-  let(:example_profile) { File.join(examples_path, 'profile') }
-  let(:meta_profile) { File.join(examples_path, 'meta-profile') }
+  let(:example_profile) { File.join(profile_path, 'profile') }
```
clintoncwolfe: I might suggest placing the copy of the examples under one directory together.  Having test profiles in folders named `profile/`, `meta-profile/`, etc might be confusing, while having a group of test fixtures under `old-examples/` would be a bit more self-explanatory. 

It might not be obvious, but a few profiles are already grouped one-level-down - see `inputs/` for example.

The `test/unit/mock/profiles/` area is pretty disorganized, so any additions (ideally) will increase its organization :-)

test/functional/helper.rb
```diff
@@ -177,8 +176,8 @@ def run_inspec_process(command_line, opts = {})
   # @param &block [Type] actual test block
   def prepare_examples(dir = nil, &block)
     Dir.mktmpdir do |tmpdir|
-      FileUtils.cp_r(examples_path, tmpdir)
-      bn = File.basename(examples_path)
+      FileUtils.cp_r(profile_path, tmpdir)
```
clintoncwolfe: yeah.... def. want to group them :-).  By placing the old examples at the top, you're now obliged to copy the *entire* set of test-fixtures (often for each test) so this method now does a lot more work.

If you place them under `old-examples/`, you can just copy `old-examples/`.

test/functional/helper.rb
```diff
@@ -55,16 +55,15 @@ module FunctionalHelper
   let(:exec_inspec) { File.join(repo_path, 'bin', 'inspec') }
   let(:mock_path) { File.join(repo_path, 'test', 'unit', 'mock') }
   let(:profile_path) { File.join(mock_path, 'profiles') }
-  let(:examples_path) { File.join(repo_path, 'examples') }
```
clintoncwolfe, line 58: You probably don't want to delete this.  See below.

test/functional/inheritance_test.rb
```diff
@@ -5,11 +5,11 @@
 
 describe 'example inheritance profile' do
   include FunctionalHelper
-  let(:path) { File.join(examples_path, 'inheritance') }
-  let(:input_file) { File.join(examples_path, 'profile-attribute.yml') } # TODO rename attributes in examples
+  let(:path) { File.join(profile_path, 'inheritance') }
```
clintoncwolfe: Ahh, ouch.  I may not have communicated this clearly, but I proposed copying the example profiles (check), then altering [examples_path](https://github.com/inspec/inspec/pull/3889/files#diff-652d43958084a48aa2c78f72b152c98aL58) to point to the new location - which gets you out of having to do all the careful search-and-replace work here (and the many files to follow).  They would still call `examples_path`, but they would point to a different location.

Now we could just accept this (changing `examples_path` -> `profile_path`) but in doing so we lose a lot of information that is valuable for later work.  Down the line, we want to make all the tests that used to rely on examples now rely on dedicated test-fixtures of their own. If we make this change, we can't tell which tests are looks at the old examples location, because many things look at the `profile_path`.

Check in with me when you read this, we can explore this further.

clintoncwolfe (changes requested):
Two changes requested:
 * Place all the old examples in a subdirectory
 * Don't replace calls to `examples_path` (yet)

Great turnaround on this - happy to Zoom with you to go over things!

test/functional/helper.rb
srb3: totally agree on improving the organisation of the mock profiles dir, I moved all the old example profiles to an old-examples sub dir. This should help preserve some context on their origin.

test/functional/inheritance_test.rb
srb3: after talking with @clintoncwolfe we decided to revert back to using the examples_path variable so we could keep visibility of the tests that use the example profiles; for use when further refactoring is needed

clintoncwolfe (approved):
Thanks, this makes it possible for us to update the examples freely!

miah (approved):

PR #3892, opened 2019-03-15
Add cloudtrail logging? method to validate that a cloudtrail is enabled.

adriankierman:
Adds new check for logging enabled. This is useful because a key security and compliance issue is having the audit log turned off. This provides a check for that in inspec.

Usage:
```
describe aws_cloudtrail_trail('my_trail_name') do
    it { should be_logging }
  end
```

Testing:
An additional set of unit tests and a fixture have been added.

To test this run
```
bundle exec m ./test/unit/resources/aws_cloudtrail_trail_test.rb
```

inspec_tests_run

Signed-off-by: Adrian Kierman <adriankierman@hotmail.com>

### DIFF
--- a/docs/resources/aws_cloudtrail_trail.md.erb
+++ b/docs/resources/aws_cloudtrail_trail.md.erb
@@ -55,6 +55,12 @@ The following examples show how to use this InSpec audit resource.
       it { should be_encrypted }
     end
 
+### Test that the specified trail has logging turned on (ie is not disabled)
+
+    describe aws_cloudtrail_trail('trail-name') do
+      it { should be_logging }
+    end
+
 ### Test that the specified trail is a multi-region trail
 
     describe aws_cloudtrail_trail('trail-name') do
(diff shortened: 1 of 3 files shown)


clintoncwolfe:
Hi @adriankierman! Thanks for this great change! However, at this point all new development on AWS resources is being done on the https://github.com/inspec/inspec-aws repo, where AWS resources are being packaged as a resource pack. Please fork that repo, and submit your PR there.  Thanks!

attn @skpaterson

skpaterson:
Thanks @adriankierman, just to confirm this method is already present in the new resource pack so you can do:
```
describe aws_cloudtrail_trail('single_trail_arn') do
    it { should be_logging }
end

# or across all trails via

aws_cloudtrail_trails.trail_arns.each do |trail_arn|
    describe aws_cloudtrail_trail(trail_arn) do
        it { should be_logging }
    end
end
```

adriankierman:
Thanks - I'll use the resource pack.

PR #3895, opened 2019-03-18
file: add `more_permissive_than?(mode)` matcher

aaronlippold:
This adds a matcher to the `file` resource called `more_permissive_than?(mode)` to enable you to validate that a file is no more permissive or no less permissive than the mode specified. Supports non-windows file permission systems.

Fixes #3894


clintoncwolfe:
I reviewed this on https://github.com/inspec/inspec/pull/3894

aaronlippold:
All review comments addressed.

docs/resources/file.md.erb
```diff
@@ -119,6 +119,8 @@ not
 
 Without the zero prefix for the octal value, InSpec will interpret it as the _decimal_ value 644, which is octal 1024 or `-----w-r-T`, and any test for a file that is `-rw-r--r--` will fail.
 
+#### Note: If you want to validate that a file has an upper or lower bound in `mode` see the [`more_permissive_than?`(mode)](#more_permissive_than?(mode)) matcher below.
```
clintoncwolfe: Hash marks should only be used for headers. It also reads odd, like an advertisement for a feature in the documentation. I'd prefer to see this line removed.

docs/resources/file.md.erb
```diff
@@ -541,3 +543,14 @@ The `exist` matcher tests if the named file exists:
 The `have_mode` matcher tests if a file has a mode assigned to it:
 
     it { should have_mode }
+
+### `more_permissive_than?(mode)`
```
clintoncwolfe: The matcher name in the header should be as seen by the user - `be_more_permissive_than`

docs/resources/file.md.erb
```diff
@@ -541,3 +543,14 @@ The `exist` matcher tests if the named file exists:
...
+
+`more_permissive_than?(mode)` takes the maximum desired mode - in `octal format`
+('0644' or '0777') - of your file as a `String` and returns a `Boolean`. It will
+return `true` if your file has a mode with greater permissions than specified.
+
+    describe file('/etc/passwd') do
+      it { should_not be_more_permissive_than?('0644') }
```
clintoncwolfe: RSpec drops the `?` - please correct the examples.

lib/resources/file.rb
```diff
@@ -133,6 +133,35 @@ def sticky
 
     alias sticky? sticky
 
+    def more_permissive_than?(max_mode = nil)
+      raise Inspec::Exceptions::ResourceFailed, 'The file' + file.path + 'doesn\'t seem to exist' unless exist?
+      raise ArgumentError, 'You must provide a value for the `maximum permission target` for the file.' if max_mode.nil?
+      raise ArgumentError, 'You must proivde the `maximum permission target` as a `String`, you provided: ' + max_mode.class.to_s unless max_mode.is_a?(String)
```
clintoncwolfe, line 139: typo, proivde

lib/resources/file.rb
```diff
@@ -133,6 +133,35 @@ def sticky
 
     alias sticky? sticky
 
+    def more_permissive_than?(max_mode = nil)
+      raise Inspec::Exceptions::ResourceFailed, 'The file' + file.path + 'doesn\'t seem to exist' unless exist?
+      raise ArgumentError, 'You must provide a value for the `maximum permission target` for the file.' if max_mode.nil?
```
clintoncwolfe: Consider `maximum allowable permission`. `target` has a special meaning to InSpec users.

lib/resources/file.rb
```diff
@@ -133,6 +133,35 @@ def sticky
...
+      # Conversely, we dont care about the bits that are currently 1 because they
+      # cannot be any more permissive and we can safely ignore them.
+      #
+      # 2. Calculate the above result of ANDing the actual mode and the inverse
+      # mode. This will determine if any of the bits that would indicate a more
+      # permissive mode are set in the actual mode.
+      #
+      # 3. If the result is 0000. If the result is 0000, the files mode is equal
+      # to or less permissive than the desired mode (PASS). Otherwise, the files
+      # mode is more permissive than the desired mode (FAIL).
```
clintoncwolfe, line 157: Great comment!

clintoncwolfe (changes requested):
Just some docs updates, then I think it should be all set.  Thanks!

docs/resources/file.md.erb
aaronlippold: The thought I had was to make it easy for them to find this if when they were looking at mode it would point them to the right matcher so they were less likely to try to use `its('mode') { should cmp <= 0644 }`

clintoncwolfe (approved):

miah (approved):
Thanks @aaronlippold

PR #3897, opened 2019-03-18
Support address matchers on interface resource

### LINKED ISSUE #1830: "address" is mentioned in the doc, but not implemented
The interface resource documentation: 

https://www.inspec.io/docs/reference/resources/interface/

mentions "address", but in reality it isn't implemented yet!

Can you please implement it, with other basic checks, like: netmask? 

Thanks.

mattlqx:
Adds missing functionality to `interface`. Fixes #1830

```
describe interface("eth0") do
  its('ipv4_addresses') { should include '1.2.3.4' }
end
```

And so on... see diff/docs for additional matchers.


docs/resources/interface.md.erb
```diff
@@ -54,6 +54,30 @@ The `speed` matcher tests the speed of the network interface, in MB/sec:
 
     its('speed') { should eq 1000 }
 
+### ipv4_addresses
+
+The `ipv4_addresses` matcher tests if the specified address exists on the named network interface:
```
clintoncwolfe: Please s/matcher/property/ for all of these.  See [property](https://www.inspec.io/docs/reference/glossary/#property) vs [matcher](https://www.inspec.io/docs/reference/glossary/#matcher).

docs/resources/interface.md.erb
```diff
@@ -66,3 +90,14 @@ The `be_up` matcher tests if the network interface is available:
 
     it { should be_up }
 
+### be_ipv4_addressable
```
clintoncwolfe: I think this reads a bit awkward. What about `have_an_ipv4_address`?

lib/resources/interface.rb
```diff
@@ -87,28 +113,53 @@ def interface_info(iface)
...
         speed: speed,
+        ipv4_addresses: family_addresses['inet'],
+        ipv6_addresses: family_addresses['inet6'],
       }
     end
+
+    private
+
+    def addresses(iface)
+      cmd = inspec.command("echo -n 'inet6 ' && /sbin/ip -br -6 address show dev #{iface} && echo '' && " \
```
clintoncwolfe: I think I get what you're doing - avoiding doing two shellouts by prefixing with the family - but this took a bit of staring-at to get to that point. I think this should probably just loop over a two-element hash (family => args) and shellout once per iteration, for clarity and maintainability.

clintoncwolfe: Looks like this needs to either be need be ported to MacOS (no /sbin/ip), or add a guard clause indicating it is not supported (along with a `## Limitations` section in the docs).

@miah do you have any portability concerns?

lib/resources/interface.rb
```diff
@@ -41,6 +43,30 @@ def speed
...
+
+    def ipv4_addresses
+      ipv4_cidrs.map { |i| i.split('/')[0] }
+    end
+
+    def ipv6_addresses
+      ipv6_cidrs.map { |i| i.split('/')[0] }
+    end
+
+    def ipv4_cidrs
```
clintoncwolfe, line 74: While I think this is useful, I think some people might want to see it as an netmask, too.  That could be a separate feature request if you feel that isn't worth it.

clintoncwolfe (changes requested):
This is great! A minor docs change, a portability concern, and a clarity request, and then I think we'll be closer to a merge.  Thanks!

docs/resources/interface.md.erb
mattlqx: Can do.

docs/resources/interface.md.erb
mattlqx: Sure.

lib/resources/interface.rb
mattlqx: I can give it a go.

lib/resources/interface.rb
mattlqx: Yeah, I was trying to save the extra round-trip. But I can break it out.

mattlqx: I wasn't concerned with macOS as this class is already Linux specific. e.g. `/sys/class/net` doesn't exist on macOS either and it's called LinuxInterface.

clintoncwolfe: Yes, good call.

clintoncwolfe (approved):

miah (approved):
Thank you @mattlqx

PR #3898, opened 2019-03-19
Fixes resource example indentation

burtlo:
The opening and closing mechanic varied between all the various resources. This changes them all to use a HEREDOC with a tilde to remove leading whitespace. This removes the need for the special method to trim the `#print_example` method from shell.


miah (approved):
Thank you @burtlo! Appreciate this cleanup =)

clintoncwolfe (approved):
Thanks so much!

PR #3899, opened 2019-03-21
crontab: Add validation for path param

skpaterson:


skpaterson:
I see the build failed but can't easily re-trigger it.

clintoncwolfe:
Looks like the appveyor build failure is due to mixlib-cli 2.x not working on Ruby 2.3 . I'll look into that separately.

clintoncwolfe (approved):
Thanks! This looks like a clearer way to catch a problem and clarify behavior.

miah (approved):
Thanks @skpaterson!

Issue #3846, opened 2019-02-28
Telemetry


miah:
The InSpec team would like to gather, and report on details regarding usage to help make better decisions about how InSpec is being used.

Data we might want to collect:
 - Runtime environment (OS, Ruby Version)
 - Target(s) (unix, windows, network device, http, database, etc)
 - Runtime environment specs (memory, architecture, ipv(4,6)
 - InSpec features used (attributes, profiles, resources, plugins, custom resources, deprecations)

This is an Epic, so here is the task breakdown:

 - [ ] Write report locally
 - - [ ] Collect local environment details
 - - [ ] Collect remote environment details
 - - [ ] Collect deprecation details
 - [x] Opt-in / Opt-out
 - [ ] Send report to telemetry endpoint
 - - [ ] Fast-fail without network access / airgap
 - - [ ] Ability to set different endpoints
 - - [ ] Internal Telemetry Dashboard
 - [ ] Document Telemetry in docs/dev

clintoncwolfe:
Suggested data point - Ruby exceptions https://github.com/inspec/inspec/pull/3903#issuecomment-475866638

aaronlippold:
So does this encapsulate also the open issue we have with pass through metadata from the user as well as the discussion around the ahi resource and the expansion of the reporters to collect and report on this type of information?

aaronlippold:
I don't want us to diverge too much with those three different open items if we're going to hopefully centralize them here in a consolidated conversation.

clintoncwolfe:
Telemetry is about optionally collecting runtime data and sending it to Chef, for use in future product decisions.

If you want to capture network identifying data (as on #3869) and include it in the results of the run, with the intended recipient being the user, then you want a Reporter.

PR #3903, opened 2019-03-22
[WIP] - Telemetry data series from Reporters

miah:
Adds a core telemetry interface to InSpec.

Collects local details about Ruby Environment
Collects details about the number of profiles, controls, and tests.
Collects details about the unique types of tests and their average run time.
Can generate a simple report (that is not written anywhere just yet).

ref: https://github.com/inspec/inspec/issues/3846

### DIFF
--- a/lib/inspec/runner_rspec.rb
+++ b/lib/inspec/runner_rspec.rb
@@ -75,6 +75,7 @@ module Inspec
     def run(with = nil)
       with ||= RSpec::Core::Runner.new(nil)
       @rspec_exit_code = with.run_specs(tests)
+      Inspec.telemetry_report(@formatter.results)
       @formatter.results
     end
 
--- a/lib/inspec/utils/telemetry.rb
+++ b/lib/inspec/utils/telemetry.rb
@@ -1,0 +1,94 @@
+require 'singleton'
+require 'inspec/utils/telemetry/config'
+require 'inspec/utils/telemetry/data_series'
+require 'inspec/utils/telemetry/global_method'
+require 'inspec/utils/telemetry/report'
+
+# Inspec Telemetry!
+# This is a Singleton and acts as a collector for incoming data.
+module Inspec::Telemetry
+  class Collector
+    include Singleton
+
+    def initialize
+      @data_series_collection = []
+    end
+
+    # A collection of Inspec::Telemetry::DataSeries objects
+    # @return [Array]
+    attr_reader :data_series_collection
+
+    #
+    # @return [TrueClass]
+    def report_add(report)
+      parse_report(report)
+    end
+
+    # TBD
+    def deprecation_add(deprecation)
+    end
+
+    private
+
+    # Parse a Hash 'formatted report' from rspec and generate a Telemetry Report.
+    # @return [Inspec::Telemetry::Report]
+    def parse_report(report)
+      parse_profiles(report[:profiles])
+      generate_report
+    end
+
+    # Parses profiles, and if they include controls parse them too.
+    # Adds all profiles to our data collection.
+    # @return [TrueClass]
+    def parse_profiles(profiles)
+      profiles.each do |profile|
+        data = data_series
+        data.type = "profile"
+        data.controls = profile[:controls].count
+        profile[:controls].nil? ? next : parse_controls(profile[:controls])
+        add_data_series(data)
+      end
+    end
+
+    # Parses controls, and if they include tests parse them too.
+    # Adds all controls to our data collection.
+    # @return [TrueClass]
+    def parse_controls(controls)
+      controls.each do |control|
+        data = data_series
+        data.type = "control"
+        data.tests = control[:results].count
+        control[:results].nil? ? next : parse_tests(control[:results])
+        add_data_series(data)
+      end
+    end
+
+    # Parses tests and adds them to our data collection.
+    # @return [TrueClass]
+    def parse_tests(tests)
+      tests.each do |test|
+        data = data_series
+        data.type = "test"
+        data.resource = test[:resource_title].to_s.split.first
+        data.run_time = test[:run_time]
+        add_data_series(data)
+      end
+    end
+
+    # @return [TrueClass]
+    def generate_report
+      Inspec::Telemetry::Report.new(data_series_collection)
+    end
+
+    # Interface to DataSeries
+    # @return [Inspec::Telemetry::Dataseries]
+    def data_series
+      Inspec::Telemetry::DataSeries.new
+    end
+
+    # @return [TrueClass]
+    def add_data_series(data)
+      @data_series_collection << data
+    end
+  end
+end
--- a/lib/inspec/utils/telemetry/config.rb
+++ b/lib/inspec/utils/telemetry/config.rb
@@ -1,0 +1,14 @@
+require 'inspec/config'
+
+# - configuration
+# -- opt in/out
+# -- endpoint
+# -- filter list for on/off
+
+module Inspec::Telemetry
+  class Config
+    def initialize
+      @config = Inspec::Config.cached
+    end
+  end
+end
--- a/lib/inspec/utils/telemetry/data_series.rb
+++ b/lib/inspec/utils/telemetry/data_series.rb
@@ -1,0 +1,25 @@
+#  - data series / attributes = [ run timing, profiles, tests, os details, deprecation ]
+#  -- name
+#  -- data type
+#  -- on/off switch
+module Inspec::Telemetry
+  class DataSeries
+    def initialize
+      @enable = true
+    end
+
+    attr_accessor :resource, :type, :run_time, :controls, :tests
+
+    def enable
+      @enable = true
+    end
+
+    def disable
+      @enable = false
+    end
+
+    def enabled?
+      @enable
+    end
+  end
+end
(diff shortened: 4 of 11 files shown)


test/unit/utils/telemetry/data_series_test.rb
```diff
@@ -0,0 +1,41 @@
+require 'inspec/utils/telemetry/data_series'
+require_relative '../../../helper.rb'
+
+class TestTelemetryDataSeries < Minitest::Test
+  def setup
+    @data_series = Inspec::Telemetry::DataSeries.new
+  end
+
+  def test_initialization
+    assert_equal Inspec::Telemetry::DataSeries, @data_series.class
```
clintoncwolfe, line 10: Consider using assert_kind_of for a clearer message on failure

lib/inspec/utils/telemetry/data_series.rb
```diff
@@ -0,0 +1,25 @@
+#  - data series / attributes = [ run timing, profiles, tests, os details, deprecation ]
+#  -- name
+#  -- data type
+#  -- on/off switch
+module Inspec::Telemetry
+  class DataSeries
+    def initialize
+      @enable = true
```
clintoncwolfe, line 8: Consider naming this `@enabled`

lib/inspec/utils/telemetry/data_series.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    attr_accessor :resource, :type, :run_time, :controls, :tests
```
clintoncwolfe, line 11: Design thoughts, here.

As implemented, this DataSeries can only be associated with elements from an InSpec Report.  So, adding new types of data to track will be difficult or nonsensical; we can't add elements to the DataSeries that don't appear in a Report.

I would expect DataSeries to be a container that might have attributes:
 * name
 * data (an Array or similar, of plain data, like strings or integers)
Behavior would include:
 * push or `:<<`, some way of appending to its internal data structure
 * to_json, to_h or other marshalling

During the run, you might collect data series from all sorts of places.
 * there be a data series that tracks a list of resource names used (name: resources/names)
 * a data series that counts deprecations touched (name: deprecations/count, Integer)
 
Point being, don't tie the data representation to *what is in the report*.  At this stage the Telemetry system should be very, very generic, and need not be able to process a report at all.

lib/inspec/utils/telemetry/config.rb
```diff
@@ -0,0 +1,14 @@
+require 'inspec/config'
+
+# - configuration
+# -- opt in/out
+# -- endpoint
+# -- filter list for on/off
+
+module Inspec::Telemetry
+  class Config
+    def initialize
```
clintoncwolfe, line 10: Looks like this class doesn't do anything yet, and presumably isn't used by anything either. Consider removing this file from the PR, then having a tightly-focused PR that includes the config system along with whatever needs it (presumably opt-in).

lib/inspec/runner_rspec.rb
```diff
@@ -75,6 +75,7 @@ def tests
     def run(with = nil)
       with ||= RSpec::Core::Runner.new(nil)
       @rspec_exit_code = with.run_specs(tests)
+      Inspec.telemetry_report(@formatter.results)
```
clintoncwolfe, line 78: Consider renaming this, perhaps to 
`InSpec::Telemetry.process_reporter_results`.  It's a more awkward name, but it should only be called exactly once - here in the runner.

lib/inspec/utils/telemetry/global_method.rb
```diff
@@ -0,0 +1,8 @@
+require 'inspec/utils/telemetry'
+
+module Inspec
+  def self.telemetry_report(report)
```
clintoncwolfe, line 4: I suggest not making this a global method (instead a singleton class method, or similar). It will only be called once, so no need to make it easy to get to (and it's maybe not worth "polluting" the `Inspec.*` space).

From my perspective, there is a greater need for a very general global method, something like:

`Inspec.telemetry(:some_data_series_name, value_to_record)`

Its implementation might:
 * Locate the Telemetry collector instance 
 * look up or create a data series
 * append the data point to the data series

It has no opinion on the name of data series, or on the values passed.

lib/inspec/utils/telemetry/endpoint/file.rb
```diff
@@ -0,0 +1,4 @@
+module Inspec::Telemetry::Endpoint
```
clintoncwolfe, line 1: As this is nearly empty, and other endpoint files are empty, I might suggest removing them from this PR, and having a subsequent PR which implements adding a file output, then a later PR which adds an HTTP output, etc.

lib/inspec/utils/telemetry.rb
```diff
@@ -0,0 +1,94 @@
...
+    attr_reader :data_series_collection
+
+    #
+    # @return [TrueClass]
+    def report_add(report)
+      parse_report(report)
+    end
+
+    # TBD
+    def deprecation_add(deprecation)
```
clintoncwolfe, line 28: I actually think Telemetry should not know about deprecations (or possibly even Reporter Results!) at all.

If Telemetry offers a generic way to record data points from anywhere in the codebase, then if the Deprecation system cares about participating in telemetry, it can call the generic data collector from the Deprecation code.

This allows us to add telemetry collection points *anywhere*.

lib/inspec/utils/telemetry.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    # @return [TrueClass]
+    def generate_report
+      Inspec::Telemetry::Report.new(data_series_collection)
+    end
+
+    # Interface to DataSeries
+    # @return [Inspec::Telemetry::Dataseries]
+    def data_series
```
clintoncwolfe, line 85: Unless you're doing something special, I think this method should probably go away. Just call the constructor where you need it.

lib/inspec/utils/telemetry.rb
```diff
@@ -0,0 +1,94 @@
...
+      Inspec::Telemetry::DataSeries.new
+    end
+
+    # @return [TrueClass]
```
clintoncwolfe, line 89: Does this return TrueClass?  I thought it would return the array.

lib/inspec/utils/telemetry/report.rb
```diff
@@ -0,0 +1,130 @@
+require 'json'
+
+# Generate a Telemetry Report from a collection of DataSeries objects.
+module Inspec::Telemetry
+  class Report
```
clintoncwolfe, line 5: I think we should find a new name for this. Telemetry has to deliver its results, and the Runner has results, both are called Report or Reporter or `report` methods.

lib/inspec/utils/telemetry/report.rb
```diff
@@ -0,0 +1,130 @@
...
+    # Initialize a new report with a data series collection
+    # @param [Array] of [Inspec::Telemetry::DataSeries]
+    def initialize(data_series_collection)
+      @data_series_collection = data_series_collection.clone
+    end
+
+    # Generate a report.
+    # @return [Hash]
+    def generate
```
clintoncwolfe, line 14: I think this should be very generic.  Loop over the list of data series, ask them to marshall themselves, assemble in a structure, then (in future PR) send that structure to an endpoint(s).

In other words, this should not know about things like profiles, controls, etc - if it does, we have to change this every time we add or remove a data series.  Plugins might load that want to define data series, or might not be loaded - this would have to have conditionals for lots of things.

lib/inspec/utils/telemetry/report.rb
```diff
@@ -0,0 +1,130 @@
...
+    def generate_json
+      generate.to_json
+    end
+
+    private
+
+    # Determine how many of a type of item exists in our collection.
+    # @param [String] collection_type, the item you'd like counted, eg profile, control, test
+    # @return [Integer]
+    def count_collection_type(collection_type)
```
clintoncwolfe, line 35: I agree we'll have to have some sort of summarization facility... counting is on obvious case. Might want to be thinking about what else.

aaronlippold:
I think one thing we should add in here is if there are any exceptions raised during the run of any of the profiles and a reference to their data. Right now I'm parsing this out from the results itself but I think this is not more correct home for that data.

aaronlippold:
Also, if you look here in the `summery` tool you will see other data that we have found useful to collect that may have a good home in telemetry. The other small tool  for parsing the total number of high failed medium passes low skips etc might also be useful as a standard data item but you may or may not agree that it belongs in telemetry. 

https://github.com/mitre/inspec_tools.

aaronlippold:
I think this is great, along with the Target node asset information like IP + MAC address etc and this data we're really starting to cook with gas :-)

clintoncwolfe (changes requested):
So, I think this is a great start, but I think it might want to be split up in to several PRs:

 * a base, extremely generic PR that does not know about Reporter Results at all, and only knows how to do generic data series
 * something that knows how to pick apart a ReporterResult and create data series using the generic calls
 * The beginnings of the Config / opt-in system
 * the beginnings of the endpoint system

clintoncwolfe:
@aaronlippold, a good place to collect ideas for Telemetry would be on #3846 . Anything mentioned here will live and die with this specific PR :-)

PR #3906, opened 2019-03-28
automate reporter: Add top-level 'passthrough' field

### LINKED ISSUE #3874: add "projects" to pass through metadata
### 🎛 Description

 🙋 feature request

in order to complete the rbac work being done in automate, we're going to need to pass a field called `projects` through on reports, with an array of strings as the value (`projects: ['foo', 'bar']`).
https://github.com/inspec/inspec/blob/master/lib/inspec/reporters/automate.rb#L31

note: while we're at it, maybe correcting the spelling on `recipes` would be good? I don't actually make use of the roles or recipes fields there, just noticed the spelling there.

3.7.1

clintoncwolfe:
* Adds a new option to the `json-automate` and `automate` reporters, `passthrough`. Whatever data is placed there in the config file will be exposed under the top-level field `passthrough`.
 * Removes the unused `recipies` (sic) field
 * Internally moves the implementation of passthrough fields from `automate` to the `json-automate` reporter. Since `automate` inherits from `json-automate`, this lets both reporters get the fields. Since `automate` is difficult to test, this allows the fields to be checked.
 * Adds functional tests to cover the changes 

Closes #3874

### DIFF
--- a/lib/inspec/reporters/automate.rb
+++ b/lib/inspec/reporters/automate.rb
@@ -28,10 +28,6 @@ module Inspec::Reporters
 
       final_report[:report_uuid] = @config['report_uuid'] || uuid_from_string(final_report[:end_time] + final_report[:node_uuid])
 
-      # optional json-config passthrough options
-      %w{node_name environment roles recipies job_uuid}.each do |option|
-        final_report[option.to_sym] = @config[option] unless @config[option].nil?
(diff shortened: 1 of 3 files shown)


vjeffrey (approved):
thank you!!

miah (approved):

PR #3911, opened 2019-03-31
Don't include plugin test files in the gem file

tas50:
One of the test files is causing chef omnibus failures on Solaris for some reason, but really there's no reason to ship all these test files in the install artifact. This slims down the gem and install size a good amount. I'm open to a better regex if someone has opinions.


jaymalasinha (approved):
👍

miah (approved):

clintoncwolfe (approved):
Thanks!

PR #3917, opened 2019-04-04
Create omnibus buildkite pipelines

jeremiahsnapp:
This creates buildkite pipelines that will be used for omnibus builds/tests. It does not actually replace the current build/test infrastructure. That will be done in a separate PR once we finish testing these new pipelines.


schisamo (approved):
tenor-140565714

miah (approved):

PR #3918, opened 2019-04-05
Fix PATH env var in new omnibus test scripts

jeremiahsnapp:
Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>

### DIFF
--- a/omnibus/omnibus-test.ps1
+++ b/omnibus/omnibus-test.ps1
@@ -29,8 +29,7 @@ Start-Process "$package_file" /quiet -Wait
 
 Write-Output "--- Testing $channel $product $version"
 
-$INSTALL_DIR = "C:\opscode\inspec"
-$Env:PATH = "${INSTALL_DIR};${Env:PATH}"
+$Env:PATH = "C:\opscode\inspec\bin;${Env:PATH}"
 
 Write-Output "Running verification for $product"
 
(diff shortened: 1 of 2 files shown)


tduffield (approved):

omnibus/omnibus-test.ps1
```diff
@@ -39,5 +38,4 @@ Write-Output "Running verification for $product"
 $Env:GEM_HOME = "C:\SHOULD_NOT_EXIST"
 $Env:GEM_PATH = "C:\SHOULD_NOT_EXIST"
 
-Set-Location -Path C:\opscode\inspec\bin
```
clintoncwolfe, line 42: What other impact would removing the CWD setting have?

clintoncwolfe (commented):
PATH changes seem sensible; I don't know what impact the Set-Location change would have.

miah (approved):
Seems reasonable. I suspect the intention here is to make sure we are able to execute the installed omnibus binaries while in a different directory to ensure they work from PATH and ensure there are no 'relativity' issues?

omnibus/omnibus-test.ps1
jeremiahsnapp: I don't see any impact at all in my testing.

I originally copied the following lines from [ci/verify-inspec.bat](https://github.com/inspec/inspec/blob/master/ci/verify-inspec.bat#L8-L9) and converted the `cd` to powershell's native `Set-Location -Path`.

```
cd C:\opscode\inspec\bin
inspec version
```

I just did some more thorough testing though and realized that the line was never really necessary since `C:\opscode\inspec\bin` is already getting prepended to the `PATH` so I removed it to avoid confusion about its presence.

clintoncwolfe (approved):

PR #3924, opened 2019-04-08
Add -y to all apt-get update commands in integration tests

clintoncwolfe:
Intended to fix failing Debian 8 Travis CI tests, for example https://travis-ci.org/inspec/inspec/jobs/514248344

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -47,14 +47,14 @@ platforms:
     image: dokken/debian-8
     pid_one_command: /bin/systemd
     intermediate_instructions:
-      - RUN /usr/bin/apt-get update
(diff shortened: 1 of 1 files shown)


kitchen.yml
```diff
@@ -104,7 +104,7 @@ suites:
   - recipe[audit]
   verifier:
     inspec_tests:
-      - test/integration/default
+      - test/integration/default -y
```
tduffield: I don't think this one is necessary?

clintoncwolfe: Thanks, good catch

tduffield (approved):

miah (approved):

PR #3928, opened 2019-04-09
Telemetry Object Model

miah:
Based on the review from #3903, and discussions during pairing sessions I simplified the objects for data series and the telemetry collector. 

ref: https://github.com/inspec/inspec/pull/3903

### DIFF
--- a/lib/inspec/utils/telemetry/collector.rb
+++ b/lib/inspec/utils/telemetry/collector.rb
@@ -1,0 +1,66 @@
+require 'inspec/utils/telemetry/data_series'
+require 'singleton'
+
+module Inspec::Telemetry
+  # A Singleton collection of data series objects.
+  class Collector
+    include Singleton
+
+    def initialize
+      @data_series = []
+      @enabled = true
+    end
+
+    # Add a data series to the collection.
+    # @return [True]
+    def add_data_series(data_series)
+      @data_series << data_series
+    end
+
+    # Is the Telemetry system enabled or disabled?
+    # Always true until we add configuration parsing.
+    # @return [True, False]
+    def telemetry_enabled?
+      @enabled
+    end
+
+    # A way to disable the telemetry system.
+    # @return [True]
+    def disable_telemetry
+      @enabled = false
+    end
+
+    # The entire data series collection.
+    # @return [Array]
+    def list_data_series
+      @data_series
+    end
+
+    # Finds the data series object with the specified name and returns it.
+    # If it does not exist then creates a new data series with that name
+    # and returns it.
+    # @return [Inspec::Telemetry::DataSeries]
+    def find_or_create_data_series(name)
+      ds = @data_series.select { |data_series| data_series.name.eql?(name) }
+      if ds.empty?
+        new_data_series = Inspec::Telemetry::DataSeries.new(name)
+        @data_series << new_data_series
+        new_data_series
+      else
+        ds.first
+      end
+    end
+
+    # Lists all the data series objects that match the specified name.
+    # @return [Array]
+    def list_data_series_by_name(name)
+      @data_series.select { |ds| ds.name.eql?(name) }
+    end
+
+    # Blanks the contents of the data series collection.
+    # @return [True]
+    def reset
+      @data_series = []
+    end
+  end
+end
--- a/lib/inspec/utils/telemetry/collector.rb
+++ b/lib/inspec/utils/telemetry/collector.rb
@@ -51,12 +51,6 @@ module Inspec::Telemetry
       end
     end
 
-    # Lists all the data series objects that match the specified name.
-    # @return [Array]
-    def list_data_series_by_name(name)
-      @data_series.select { |ds| ds.name.eql?(name) }
-    end
-
     # Blanks the contents of the data series collection.
     # @return [True]
     def reset
--- a/lib/inspec/utils/telemetry/data_series.rb
+++ b/lib/inspec/utils/telemetry/data_series.rb
@@ -8,13 +8,10 @@ module Inspec::Telemetry
     def initialize(name)
       @name = name
       @enabled = true
-    end
-
-    attr_reader :name
-
-    def data
       @data ||= []
     end
+
+    attr_reader :data, :name
 
     # This needs to also be set by configuration.
     def enabled?
--- a/lib/inspec/utils/telemetry/data_series.rb
+++ b/lib/inspec/utils/telemetry/data_series.rb
@@ -1,0 +1,45 @@
+require 'json'
+
+# A minimal Dataseries Object
+# Stores the name of the data series and an array of data.
+# Stored data should be a object that supports #to_s
+module Inspec::Telemetry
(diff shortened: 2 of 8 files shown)


lib/inspec/utils/telemetry/global_methods.rb
```diff
@@ -0,0 +1,22 @@
+require 'inspec/utils/telemetry'
+
+module Inspec
+  # A Global method to add a data series object to the Telemetry Collection.
+  # `data_series_name`s are unique, so `:dependency_group` will always return
+  # the same object.
+  # `data_point` is optional, you may also supply a block with several data points.
+  # All data points should allow #to_s
+  # @return [True]
```
clintoncwolfe: I'm not sure I understand why this would be specified to return true.

test/unit/utils/telemetry/data_series_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    ds = Inspec::Telemetry::DataSeries.new('fizz')
+    assert_empty ds.data
+    assert ds.push 'bar'
+    assert_equal ['bar'], ds.data
+  end
+
+  def test_to_h
+    ds = Inspec::Telemetry::DataSeries.new('fizz')
+    ds << 'foo'
+    assert_kind_of Hash, ds.to_h
```
clintoncwolfe, line 36: I think this should test the contents of the Hash.  Just from reading this, I'm not sure what the contents would be - what are the important keys?  What types are their values? Symbols or Strings for keys?

test/unit/utils/telemetry/data_series_test.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def test_to_json
+    ds = Inspec::Telemetry::DataSeries.new('fizz')
+    ds << 'foo'
+    assert_kind_of String, ds.to_json
+    assert JSON.parse(ds.to_json)
```
clintoncwolfe, line 46: As above, I think this should do some checking as to the contents of the parsed JSON - what is expected to be in it?

lib/inspec/utils/telemetry/data_series.rb
```diff
@@ -0,0 +1,45 @@
+require 'json'
+
+# A minimal Dataseries Object
+# Stores the name of the data series and an array of data.
+# Stored data should be a object that supports #to_s
+module Inspec::Telemetry
+  class DataSeries
+    def initialize(name)
+      @name = name
+      @enabled = true
```
clintoncwolfe, line 10: If `@data` is intrinsic to the data series, it should be initialized in the constructor

lib/inspec/utils/telemetry/data_series.rb
```diff
@@ -0,0 +1,45 @@
...
+      @enabled = false
+    end
+
+    def <<(appending_data)
+      data << appending_data
+    end
+
+    alias push <<
+
+    def to_h
```
clintoncwolfe, line 31: We'll need to check with the real-world endpoints to verify this is the structure they want.  { name1 => [data1], name2 => [data2] } might also be plausible.

test/unit/utils/telemetry/collector_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    @collector.reset
+  end
+
+  def test_collector_singleton
+    assert_equal Inspec::Telemetry::Collector.instance, @collector
+  end
+
+  def test_add_data_series
+    data_series = Inspec::Telemetry::DataSeries.new('/resource/File')
+    assert @collector.add_data_series(data_series)
```
clintoncwolfe: What is being asserted here? `assert` checks to see if a value is truthy; can you say anything stronger about the return value of `add_data_series`?

test/unit/utils/telemetry/collector_test.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def test_list_data_series
+    assert @collector.list_data_series
```
clintoncwolfe: I would expect tests on a listing method to:
 * verify it is empty at first
 * add a few data series
 * make sure you get those data series back
 * check the type of the list
 * check the type of the list elements

test/unit/utils/telemetry/collector_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    assert_equal :deprecation_group, dg.name
+    assert_equal @collector.find_or_create_data_series(:deprecation_group), dg
+  end
+
+  def test_list_data_series_names
+    data_series = Inspec::Telemetry::DataSeries.new('/resource/File')
+    data_series2 = Inspec::Telemetry::DataSeries.new('/resource/Dir')
+    @collector.add_data_series(data_series)
+    @collector.add_data_series(data_series2)
+    list = @collector.list_data_series_by_name('/resource/File')
```
clintoncwolfe: I don't understand why this would take a parameter, or if it does take aparameter, why it would be an Array return. As data series names are unique, if you ask for something by name, you would only ever get back one dataseries.

I could see how having a `list_data_series_names` method (takes no args) could be useful, but I don't see how this is useful by itself.

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -0,0 +1,66 @@
...
+        @data_series << new_data_series
+        new_data_series
+      else
+        ds.first
+      end
+    end
+
+    # Lists all the data series objects that match the specified name.
+    # @return [Array]
+    def list_data_series_by_name(name)
```
clintoncwolfe: I think this should be rethought to account for uniqueness, or simply removed.

clintoncwolfe (commented):
Really great stuff here, I think I have only a few minor concerns:

 * I'd like to see more specific testing, especially of to_h
 * Most methods here need not return anything; it seems odd to test for return values but not other behavior
 * I think list_data_series_by_name should probably be removed since data series names are unique.

None of these are enough to block merging, but I wanted to give you a chance to respond and/or address them.  So, a Comment-only review from me at this time.

test/unit/utils/telemetry/data_series_test.rb
miah: Ok I can improve this.

test/unit/utils/telemetry/data_series_test.rb
miah: An this =)

test/unit/utils/telemetry/collector_test.rb
miah: I'll add another test to check that the data series was added correctly.

test/unit/utils/telemetry/collector_test.rb
miah: Ok

test/unit/utils/telemetry/collector_test.rb
miah: Ok

clintoncwolfe (approved):
Thanks, looks great!

PR #3939, opened 2019-04-11
Test with webmock 3+

tas50:
Webmock 2.x does not support Ruby 2.6.


clintoncwolfe (approved):
Thanks!

miah (approved):

PR #3940, opened 2019-04-11
Remove the omnibus override to use train 1.X

tas50:
We don't want this anymore


clintoncwolfe (approved):
Yep yep

miah (approved):

PR #3948, opened 2019-04-15
Improve omnibus test scripts

jeremiahsnapp:
* Reload Env:PATH to get changes from inspec installation
* Do not add inspec's bin path to bash's PATH environment variable

By avoiding explicitly making changes to PATH or changing directory into the project's bin directory we force the tests to completely rely on the results of the projects' install scripts.

### DIFF
--- a/omnibus/omnibus-test.sh
+++ b/omnibus/omnibus-test.sh
@@ -13,11 +13,9 @@ check-omnibus-package-signed "$package_file"
 
 echo "--- Testing $channel $product $version"
 
-export INSTALL_DIR=/opt/inspec
-export PATH="/opt/inspec/bin:$PATH"
-
 echo "Verifying ownership of package files"
 
+export INSTALL_DIR=/opt/inspec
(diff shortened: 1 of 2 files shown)


miah (approved):

tduffield (approved):

omnibus/omnibus-test.sh
```diff
@@ -34,5 +32,4 @@ echo "Running verification for $product"
 export GEM_HOME=/SHOULD_NOT_EXIST
 export GEM_PATH=/SHOULD_NOT_EXIST
 
-export PATH="$PATH:/usr/local/bin"
 inspec version
```
jaymalasinha, line 35: Do we need to pass the last exit code here as well ?

jaymalasinha (approved):

omnibus/omnibus-test.sh
jeremiahsnapp: no bash tests will handle failed exit codes properly. it's only in powershell that we have to explicitly test for failed exit codes.

PR #3953, opened 2019-04-17
Require Ruby 2.4 or later

tas50:
Ruby 2.3 is no longer supported. Drop this release from InSpec and
instead support the 3 latest Ruby releases.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -55,6 +55,8 @@ Metrics/PerceivedComplexity:
   Max: 11
 NumericLiterals:
   MinDigits: 10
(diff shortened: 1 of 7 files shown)


clintoncwolfe:
The ruby version change for RuboCop has us now failing linting.  I'll push to this branch with fixes for that.

tas50:
That regex match cop is the worst. It autocorrects things that break all over chef. We disabled it in chefstyle.

clintoncwolfe (approved):
Thanks!

miah (approved):

PR #3954, opened 2019-04-17
First step to get inspec names changeable

artem-sidorenko:
for community distributions

Please see this as a first PR. The intention is to get feedback on the direction. All other PRs will follow once the direction is clear and this PR is accepted.

### DIFF
--- a/lib/inspec/dist.rb
+++ b/lib/inspec/dist.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Inspec
+  module Dist
+    # When referencing a product directly, like InSpec
+    PRODUCT_NAME = 'Chef InSpec'
+
+    # The inspec executable
+    EXEC_NAME = 'inspec'
+  end
+end
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -16,6 +16,7 @@ require 'inspec/runner_mock'
 require 'inspec/env_printer'
 require 'inspec/schema'
 require 'inspec/config'
+require 'inspec/dist'
 
 class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :log_level, aliases: :l, type: :string,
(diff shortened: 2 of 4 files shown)


clintoncwolfe:
Attn @nellshamrell

nellshamrell:
@clintoncwolfe - looks fine to me, do you have time to review this post the next InSpec release? When can @artem-sidorenko expect an approval or request for changes?

clintoncwolfe:
We will be able to look at this soon after April 30th, 2019.

lib/inspec/dist.rb
```diff
@@ -0,0 +1,9 @@
+module Inspec
+  module Dist
+    # When referencing a product directly, like InSpec
+    PRODUCT = 'InSpec'.freeze
+
+    # The inspec executable
+    EXEC = 'inspec'.freeze
```
zenspider: Since our minimum version of ruby is 2.4, I'd suggest using frozen string literals with the magic comment:

```ruby
# frozen_string_literal: true
```

and removing the manual freeze calls.

That said, I don't see this as a blocker at all.

zenspider (changes requested):
minor feedback. happy to be overridden on this.

lib/inspec/dist.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Inspec
+  module Dist
+    # When referencing a product directly, like InSpec
+    PRODUCT = 'InSpec'
```
clintoncwolfe: This value should (As of v4) be 'Chef InSpec'

lib/inspec/dist.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    # The inspec executable
+    EXEC = 'inspec'
```
clintoncwolfe: For clarity, I might suggest calling this `Inspec::Dist::EXEC_NAME`

lib/inspec/dist.rb
clintoncwolfe: For clarity, I might suggest calling this `Inspec::Dist::PRODUCT_NAME`

clintoncwolfe (commented):
This all seems good to me, with a small clarity quibble. 

Also, are you tracking that we have pushed the executable out to a separate gem?

artem-sidorenko:
@clintoncwolfe thanks for the feedback, I addressed all points

> Also, are you tracking that we have pushed the executable out to a separate gem?

Thank you for the hint, I was aware about the plans to do so, but I wasn't aware it already happened. I'll have a look.

clintoncwolfe (approved):
Thanks for this - looking forward to moving ahead!

miah (approved):
Thanks @artem-sidorenko

zenspider (approved):
Looks great to me. Thanks!

PR #3958, opened 2019-04-18
Provide logger to license-acceptance for help debugging in the future

tyler-ball:
Signed-off-by: tyler-ball <tball@chef.io>

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'train-aws', '~> 0.1'
 
   # Implementation dependencies
-  spec.add_dependency 'license-acceptance', '~> 0.2'
+  spec.add_dependency 'license-acceptance', '>= 0.2.13', '< 2.0'
   spec.add_dependency 'thor', '~> 0.20'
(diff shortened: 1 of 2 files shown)


miah (approved):

lamont-granquist (approved):

tyler-ball:
I need to double check that the latest released version of the `license-acceptance` gem has support for this logger, but I think it does. If it does not I will release a newer version.

After I check this, it should be ready for merging.

btm (approved):
we did discuss we'll need to fix the version pin for license-acceptance the gemspec.

clintoncwolfe:
Refs #3979 - I wonder if we will have the logger setup at the right time

tyler-ball:
@clintoncwolfe I don't think the logger has been setup successfully at the point the Acceptor runs. In local testing I provided `--log-level=debug` but the logger still had its initial setting (warn I think). I think this PR should still be merged so that if logger configuration is moved earlier license acceptance continues to Just Work.

Also checked - the latest released version of this gem will support this code.

tyler-ball:
And updated the version constraint to support 1.0 when it is released next week

PR #3969, opened 2019-04-22
[DO NOT MERGE] Remove `inspec` binstub from v4 rubygem

clintoncwolfe:
This is for the relicensing effort. The license-bound executable is being removed from the gem, so that the gem may be used independently. The executables will still be available by Omnibus install (which is what you see at downloads.chef.io) as well as a Habitat package.

 * [x] binstub moved
 * [x] binstub excluded in gemspecs
 * [x] functional tests updated
 * [x] Habitat plan updated
 * [x] omnibus config updated
 * [ ] Dockerfile updated
 * [ ] homebrew updated (optional on this PR)


inspec-core.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
   spec.files = %w{README.md LICENSE} + Dir.glob('{bin,lib,etc}/**/*', File::FNM_DOTMATCH)
                                           .reject { |f| File.directory?(f) || f =~ /aws|azure|gcp/ || f =~ %r{lib/plugins/.*/test/} }
 
-  spec.executables   = %w{inspec}
+  spec.executables   = []
```
miah: I think we can remove this line entirely from both specs unless we want to be explicit here.

miah (commented):
Looks good so far.

inspec-core.gemspec
clintoncwolfe: Done, thanks

habitat/plan.sh
```diff
@@ -46,6 +46,14 @@ do_install() {
     gem install inspec-*.gem --no-document
   popd
 
+  # Binstub removed from gem distro, so it will not have been installed
+  # Do so manually
+  pushd "$pkg_prefix"
+    mkdir -p "$GEM_HOME/gems/inspec-${pkg_version}/bin/"
+    # Note that we generate a wrapper; this is the real executable
+    cp "$HAB_CACHE_SRC_PATH/$pkg_dirname/test/inspec-test-binstub" "$GEM_HOME/gems/inspec-${pkg_version}/bin/inspec"
```
tyler-ball, line 54: Something feels weird about our "real" binstubs living in the test folder, but I know this is the pattern we established so ᕕ( ᐛ )ᕗ

tyler-ball (approved):

clintoncwolfe:
DNM - missing dockerfile update

clintoncwolfe:
Closing in favor #3982

PR #3978, opened 2019-04-24
Update plugin test fixtures for Ruby 2.6 and remove for 2.3

clintoncwolfe:
This PR updates the set of in-codebase test fixture gems that are used to exercise the plugin system during the unit and functional tests.

Because gems are stored by a directory named after the minor version of the ruby binary (for example, `2.6.0/gems` for gems installed using ruby 2.6.2), the test fixture area is currently laid out with such directories in place.  When we change minor ruby versions, we are thus obliged to regenerate the plugin gems.

An unmaintained support script to do that was located and updated, but is not likely to be viable long-term.

No attempt is made here to improve the test fixture situation, or to defend this as a technique - simply, this PR maintains the existing practice without comment.


miah (approved):
Thanks @clintoncwolfe!

PR #3982, opened 2019-04-25
Package inspec executable as a separate gem

### LINKED ISSUE #3981: Package inspec executable as a separate gem
### 🎛 Description

Currently, with #3969, we have removed the inspec executable from the gem, but during omnibus and habitat builds we are simply copying it back in. That means that RubyGems won't be able to lookup the executable based on the gem name, which may cause problems; it also makes updating the Dockerfile hard.

4.1.10+

### 💁 Possible Solutions

Create a inspec-bin gem or similar.

clintoncwolfe:
Obsoletes #3969 
Closes #3981 

See above for what this is, and https://github.com/chef/chef/pull/8413/ for the approach taken.


clintoncwolfe:
Worked by a team of Chef engineers; they've had eyeballs on it all day - hence no github approvals

PR #3983, opened 2019-04-25
Appveyor Config Changes

miah:
Enable 64bit Ruby for 24, 25, and 26.
Disable MSBuild mode entirely.


clintoncwolfe (approved):
Thanks!

zenspider (approved):
I think I would have preferred to change the PATH line, but this is totally fine. Gives us a bit more granularity.

PR #3986, opened 2019-04-26
Tb/bin as gem

tyler-ball:


omnibus/Gemfile
```diff
@@ -12,7 +12,7 @@ source 'https://rubygems.org'
 
 # Use entries from chef's Gemfile
 gem 'omnibus', git: 'https://github.com/chef/omnibus', branch: 'master'
-gem 'omnibus-software', git: 'https://github.com/chef/omnibus-software', branch: 'master'
+gem 'omnibus-software', git: 'https://github.com/tyler-ball/omnibus-software', branch: 'master'
```
tduffield: are we sure we want to commit this?

tyler-ball: Whoops, no, that was for testing purposes - thanks for the catch

Issue #709, opened 2016-05-04
add directories resource


arlimus:
### Use-Case
- Test all dot-files (`\..*`) in user's home directories.
- Test all directories in a path (`/**/`)
- Test all files owned by user `adam` in a path

Stop running manual `find` commands.
Unify this across operating systems

arlimus:
Interface ideas:

``` ruby
file.search('/path/to/start', user: 'adam', type: 'directory', name: 'myfile')
```

Then there's the power-component to this task: regex/glob:

``` ruby
# regex:
file.search('/path/to/start', name: /\..netrc/)

# glob:
file.search('/path/to/start/**/.netrc')
```

arlimus:
Glob seems to be well-established by now, not just in Unix-world but also on Windows:

http://stackoverflow.com/questions/8677628/recursive-file-search-using-powershell

It feels like the right interface to use. Although regex is more powerful, an idea might be to glob by default and offer regex as an option on top. When translating this into the different systems, it's also easier to glob than to regex. Ultimately it's what most users are more familiar with...

chris-rock:
It would also be interesting to cover a use-case reported by @aaronlippold:

```
control 'filepermission-mask-05.01' do
  impact 0.5
  tag severity: 'medium'

  title 'System must Verify that Shared Library Files Have Root Ownership'
  desc 'Files from shared library directories are loaded into the address space of processes
        (including privileged ones) or of the kernel itself at runtime. Proper ownership is
        necessary to protect the integrity of the system. '

  tag cce: 'CCE-27424-1'
  tag disa: 'DISA FSO RHEL-06-000046'
  tag cci: 'CCI-001499'
  tag nist: 'AC-6'

  ref 'RHEL-SG'

 Dir.foreach("/lib") do |fname|
 describe file("/lib/#{fname}") do
  its('owner') { should cmp 'root' }
 end
 end

  Dir.foreach("/usr/lib") do |fname|
 describe file("/usr/lib/#{fname}") do
  its('owner') { should cmp 'root' }
 end
 end

 Dir.foreach("/lib64") do |fname|
 describe file("/lib64/#{fname}") do
  its('owner') { should cmp 'root' }
 end
 end

 Dir.foreach("/usr/lib64") do |fname|
 describe file("/usr/lib64/#{fname}") do
  its('owner') { should cmp 'root' }
 end
 end

end
~``
```

aaronlippold:
ontrol 'filepermission-mask-05.01' do
[...]

PR #3993, opened 2019-04-29
new resource: pam

### LINKED ISSUE #3992: Add a PAM Resource
### 🎛 Description

🙋 feature request

A PAM resource should be available that handles the complexity of full-stack PAM matches.

trevor-vaughan:
* Collates the entire PAM stack into a single resource for full
  processing.
* Adds matchers for 'match_pam_rule' with modifiers for:
  * any_with_args
  * all_with_args
  * all_without_args
  * any_with_integer_arg
* Adds matchers for 'match_pam_rules' which handles multi-line matches
  with fuzzy and exact matching options.

Fixes #3992

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -123,9 +123,14 @@ namespace :test do
   Rake::TestTask.new(:unit) do |t|
     t.libs << "test"
     t.test_files = Dir.glob([
+      "test/unit/**/pam_test.rb",
+    ])
+=begin
+    t.test_files = Dir.glob([
       "test/unit/**/*_test.rb",
       "lib/plugins/inspec-*/test/unit/**/*_test.rb",
     ])
+=end
     t.warning = !!ENV["W"]
     t.verbose = !!ENV["V"] # default to off. the test commands are _huge_.
     t.ruby_opts = ["--dev"] if defined?(JRUBY_VERSION)
--- a/lib/inspec/resources/pam.rb
+++ b/lib/inspec/resources/pam.rb
@@ -1,0 +1,544 @@
+# Pluggalbe Authentication Modules (PAM) resources and matchers for InSpec
+#
+# copyright: 2019, Onyx Point, Inc.
+# @author Trevor Vaughan <tvaughan@onyxpoint.com>
+module Inspec::Resources
+  class Pam < Inspec.resource(1)
+    # These are aliases for one another
+    attr_reader :rules, :lines
+
+    # These are here for useful interfaces into the module stack based on
+    # common searches
+    attr_reader :services, :types, :modules
+
+    name 'pam'
+
+    supports platform: 'unix'
+
+    desc 'Use the InSpec Pluggable Authentication Modules (PAM) resource to test the given system PAM configuration'
+    example "
+      # Query for a match:
+
+      describe pam('/etc/pam.d/system-auth') do
+        its('rules') { should match_pam_rule('password sufficient pam_unix.so sha512') }
+      end
+
+      # Query everything for a match without specific arguments
+      # You can use a Ruby regexp match for everything except arguments
+
+      describe pam('/etc/pam.d') do
+        its('rules') { should match_pam_rule('.* .* pam_unix.so').all_without_args('nullok') }
+      end
+
+      # Query for multiple lines
+
+      describe pam('/etc/pam.d/password-auth') do
+        required_rules = [
+          'auth required pam_faillock.so',
+          'auth sufficient pam_unix.so try_first_pass'
+        ]
+        its('rules') { should match_pam_rules(required_rules) }
+      end
+
+      # Query for multiple rules without any rules in between them
+
+      describe pam('/etc/pam.d/password-auth') do
+        required_rules = [
+          'auth required pam_faillock.so',
+          'auth sufficient pam_unix.so try_first_pass'
+        ]
+        its('rules') { should match_pam_rules(required_rules).exactly }
+      end
+    "
+
+    # Ensure that users know that errors are coming from the PAM resource
+    class PamError < StandardError; end
+
+    # Create a new PAM resource
+    #
+    # @param [String] path
+    #   Path to the target PAM configuration file or directory
+    def initialize(path = '/etc/pam.d')
+      # To know what we were actually derived from
+      @path           = path
+
+      # Easy access helpers
+      @services       = {}
+      @types          = {}
+      @modules        = {}
+
+      pam_config_file = get_file(path)
+
+      @rules          = Pam::Rules.new(pam_config_file)
+      @lines          = @rules
+
+      @top_config     = false
+
+      if path.strip == '/etc/pam.conf'
+        @top_config = true
+      end
+
+      parse_content(pam_config_file)
+    end
+
+    # Process a PAM configuration file
+    #
+    # @param [String] path
+    #   Path to the file or directory to process
+    # @param [String] service_name
+    #   The PAM Service under which the content falls
+    #   Mainly used for recursive processing
+    def parse_content(path, service_name = nil)
+      config_files = Array(path)
+
+      if path.directory?
+        config_files = list_files("#{path}/*")
+      end
+
+      config_files.each do |config_file|
+        next unless config_file.content
+
+        # Support multi-line continuance and skip all comments and blank lines
+        rules = config_file.content.gsub("\\\n",' ').lines.map(&:strip).delete_if do |line|
+          line  =~ /^(\s*#.*|\s*)$/
+        end
+
+        service = service_name
+        unless service || @top_config
+          service = config_file.basename
+        end
+
+        rules.each do |rule|
+          new_rule = Pam::Rule.new(rule, {:service_name => service})
+
+          # If we hit an 'include' or 'substack' statement, we need to derail and
+          # delve down that tail until we hit the end
+          #
+          # There's no recursion checking here but, if you have a recursive PAM
+          # stack, you're probably not logging into your system anyway
+          if ['include','substack'].include?(new_rule.control)
+            # Support full path specification includes
+            if new_rule.module_path[0].chr == '/'
+              pam_sub_config_file = get_file(new_rule.module_path)
+            else
+              if File.directory?(path.path)
+                pam_sub_config_file = get_file(File.join(path.path, new_rule.module_path))
+              else
+                pam_sub_config_file = get_file(File.join(File.dirname(path.path), new_rule.module_path))
+              end
+            end
+
+            if pam_sub_config_file.exist?
+              parse_content(pam_sub_config_file, service)
+            end
+          else
+
+            unless new_rule.type && new_rule.control && new_rule.module_path
+              raise PamError, "Invalid PAM config found at #{config_file}"
+            end
+
+            @services[new_rule.service] ||= []
+            @services[new_rule.service] << new_rule
+
+            @types[new_rule.type] ||= []
+            @types[new_rule.type] << new_rule
+
+            @modules[new_rule.module_path] ||= []
+            @modules[new_rule.module_path] << new_rule
+
+            @rules.push(new_rule)
+          end
+        end
+      end
+    end
+
+    def to_s
+      "PAM Config[#{@path}]"
+    end
+
+    def service(service_name)
+      @services[service_name]
+    end
+
+    def type(type_name)
+      @types[type_name]
+    end
+
+    def module(module_name)
+      @modules[module_name]
+    end
+
+    private
+
+    # These are primarily down here to assist with unit test stubbing since I
+    # could not figure out how to stub the `inspec` object.
+
+    # Get a list of File resources on the remote system
+    def list_files(path)
+      inspec.bash("ls #{path}").stdout.lines.map{|f| inspec.file(f.strip) }
+    end
+
+    # Get a specific File resource from the remote system
+    def get_file(path)
+      inspec.file(path)
+    end
+
+    # The list of rules with a bunch of helpers for matching in the future
+    #
+    # We do fuzzy matching across the board when checking for internal rule
+    # matches
+    class Rules < Array
+      def initialize(pam_config_file)
+        @pam_config_file = pam_config_file
+      end
+
+      def services
+        self.collect{|l| l.service}.sort.uniq
+      end
+
+      def service
+        svcs = self.collect{|l| l.service}.sort.uniq
+        if svcs.length > 1
+          raise PamError, %(More than one service found: '[#{svcs.join("', '")}]')
+        end
+
+        svcs.first
+      end
+
+      def first?(rule, opts={:service_name => nil})
+        raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
+
+        service_name = get_service_name(opts[:service_name])
+
+        _rule = Pam::Rule.new(rule, {:service_name => service_name})
+
+        rules_of_type(_rule.type, opts).first == _rule
+      end
+
+      def last?(rule, opts={:service_name => nil})
+        raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
+
+        service_name = get_service_name(opts[:service_name])
+
+        _rule = Pam::Rule.new(rule, {:service_name => service_name})
+
+        rules_of_type(_rule.type, opts).last == _rule
+      end
+
+      def rules_of_type(rule_type, opts={:service_name => nil})
+        raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
+
+        service_name = get_service_name(opts[:service_name])
+
+        if @services[service_name]
+          @services[service_name].find_all { |l| l.type == rule_type }
+        else
+          []
+        end
+      end
+
+      # Determines if one or more rules are contained in the rule set
+      #
+      # @param [Array[String] rules The Rules to find
+      # @param [Hash] opts Options for the include processor
+      # @option opts [Boolean] :exact
+      #   If set, no rules may be present between the rules provided in `rules`
+      #   If unset, the rules simply need to be in the correct order, other rules
+      #   may appear between them
+      # @option opts [String] :service_name The PAM Service under which the rules
+      #   should be searched
+      # @return [Boolean] true if found, false otherwise
+      def include?(rules, opts={:exact => false, :service_name => nil})
+        raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
+
+        service_name = get_service_name(opts[:service_name])
+
+        rules = Array(rules).map{|l| Pam::Rule.new(l, {:service_name => service_name})}
+
+        retval = false
+
+        if opts[:exact]
+          # This requires everything between the first and last rule to match
+          # exactly
+
+          first_entry = index(rules.first)
+          last_entry = index(rules.last)
+
+          if first_entry && last_entry
+            retval = (self[first_entry..last_entry] == rules)
+          end
+        else
+          # This match allows other rules between the two in question
+          retval = (rules.select{|l| super(l)} == rules)
+        end
+
+        return retval
+      end
+      alias_method :match, :include?
+
+      # An alias for setting `:exact => true` in the `include` method
+      def include_exactly?(rules, opts={})
+        include?(rules, opts.merge({:exact => true}))
+      end
+      alias_method :match_exactly, :include_exactly?
+
+      # Convert the data structure to an Array suitable for an RSpec diff
+      #
+      # @return [Array[String]]
+      def to_a
+        self.sort_by{|l| l.type}.map{|l| l.to_s}
+      end
+
+      # Convert the data structure to a String
+      #
+      # @return [String]
+      def to_s
+        to_a.join("\n")
+      end
+
+      private
+
+      # Get the service name out of the configuration target
+      #
+      # @param [String] svc_name Optional name of the service that should be
+      #    returned
+      #
+      # @return String
+      def get_service_name(svc_name = nil)
+        return svc_name if svc_name
+
+        if !svc_name && @pam_config_file.directory?
+          raise PamError, 'You must pass ":service_name" as an option!'
+        else
+          return @pam_config_file.basename
+        end
+      end
+    end
+
+    # A single Rule object that has been processed
+    #
+    # Rule equality is a fuzzy match that can accept regular expression matches
+    # within the string to compare
+    class Rule
+      attr_reader :to_s
+      attr_reader :service, :silent, :type, :control, :module_path, :module_arguments
+
+      def initialize(rule, opts = {})
+        @to_s = rule.strip.gsub(/\s+/,' ')
+
+        rule_regex = %r%
+          ^                                                        # Start of Rule
+          \s*                                                      # Ignore initial Whitespace
+          (?<silent>-)?                                            # Capture Silent Flag
+          #{'(?<service_name>.+?)\s+' unless opts[:service_name] } # Capture Service
+          (?<type>.+?)\s+                                          # Capture Type
+          (?<control>(\[.+\]|.+?))\s+                              # Capture Control
+          (?<module_path>.+?(\.so)?)                               # Capture Module Path
+          (\s+(?<module_args>.+?))?                                # Capture Module Args
+          $                                                        # End of Rule
+        %x
+
+        match_data = rule.match(Regexp.new(rule_regex, Regexp::EXTENDED))
+
+        unless match_data
+          raise PamError, "Invalid PAM configuration rule: '#{rule}'"
+        end
+
+        @service          = opts[:service_name] ? opts[:service_name] : match_data[:service_name]
+        @silent           = match_data[:silent] == '-'
+        @type             = match_data[:type]
+        @control          = match_data[:control]
+        @module_path      = match_data[:module_path]
+        @module_arguments = match_data[:module_args] ? match_data[:module_args].strip.split(/\s+/) : []
+      end
+
+      # Determine if a rule matches
+      #
+      # @param [String, Pam::Rule] to_cmd
+      #   The PAM rule to compare against
+      def match?(to_cmp)
+        to_cmp = Pam::Rule.new(to_cmp, {:service_name => @service}) if to_cmp.is_a?(String)
+
+        # The simple match first
+        self.class == to_cmp.class &&
+          @service.match(Regexp.new("^#{to_cmp.service}$")) &&
+          @type.match(Regexp.new("^#{to_cmp.type}$")) &&
+
+          # Add escape characters to the square brackets and pipe symbol in each
+          # control line
+          @control.match(Regexp.new("^#{to_cmp.control.gsub(/(\[|\])/, '\\\\\\1')}$")) &&
+          @module_path.match(Regexp.new("^#{to_cmp.module_path}$")) &&
+          (
+            # Quick test to pass if to_cmp module_arguments are a subset
+            (to_cmp.module_arguments - @module_arguments).empty? ||
+              # All module_arguments in to_cmp should Regex match something
+              to_cmp.module_arguments.all? do |arg|
+                !@module_arguments.grep(Regexp.new("^#{arg}$")).empty?
+              end
+          )
+      end
+      alias_method :==, :match?
+      alias_method :eql?, :==
+    end
+
+    # Custom Matchers
+    #
+    # TODO explain somewhere that :all_with_args, :all_without_args, :all_with_integer_arg
+    # will cause match_pam_rule to return true when there are no potential matches
+    RSpec::Matchers.define :match_pam_rule do |expected|
+      def matching_integer_arg? (line)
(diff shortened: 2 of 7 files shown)


lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
   end
 end
+
+# The following are required for the PAM resource
+#
+# Copyright 2019 - Onyx Point, Inc.
+#
+# TODO explain somewhere that :all_with_args, :all_without_args, :all_with_integer_arg
+# will cause match_pam_rule to return true when there are no potential matches
+RSpec::Matchers.define :match_pam_rule do |expected|
```
clintoncwolfe: Matchers that are specific to a resource (like this one) should be defined in the resource file .  You don't need to call `RSpec::Matchers.define` - just define a method ending in `?`.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
+# encoding: utf-8
+# copyright: 2019, Onyx Point, Inc.
+
+class Pam < Inspec.resource(1)
+  # These are aliases for one another
+  attr_reader :rules, :lines
```
clintoncwolfe: I think we should probably call them by the logical name (`rules`) only

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+
+  # These are here for useful interfaces into the module stack based on
+  # common searches
+  attr_reader :services, :types, :modules
+
+  name 'pam'
+
+  supports platform: 'unix'
+
+  desc 'Use the InSpec pam resource to test the given system pam configuration'
```
clintoncwolfe: Please expand the `PAM` acronym here, so that users unfamiliar with it may understand.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+
+    describe pam('/etc/pam.d/system-auth') do
+      its('rules') { should match_pam_rule('password sufficient pam_unix.so sha512' }
+    end
+
+    # Query everything for a match without specific arguments
+    # You can use a Ruby regexp match for everything except arguments
+
+    describe pam('/etc/pam.d') do
+      its('rules') { should match_pam_rule('.* .* pam_unix.so').all_without_args('nullok' }
```
clintoncwolfe: Missing trailing paren

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+  def initialize(path = '/etc/pam.d')
+    # To know what we were actually derived from
+    @path          = path
+
+    # Easy access helpers
+    @services      = {}
+    @types         = {}
+    @modules       = {}
+
+    config_target = inspec.file(path)
```
clintoncwolfe: Please do not use the variable name 'target' within new InSpec code, unless it is refering to the location data that allows us to connect to the remote machine. It is overused and confusing already :-) !

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+  # Process a PAM configuration file
+  #
+  # @param [String] path The path to the file or directory to process
+  # @param [String] service_name The PAM Service under which the content falls.
+  #   Mainly used for recursive processing
+  def parse_content(path, service_name = nil)
+    config_files = Array(path)
+
+    if path.directory?
+      config_files = inspec.bash("ls #{path}/*").stdout.lines.map{|f| inspec.file(f.strip) }
```
clintoncwolfe: Dir globbing would come along on #709....

Does this require bash?  Could we use command here, in case the system does not have `bash`?

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+  def module(module_name)
+    @modules[module_name]
+  end
+
+  # The list of rules with a bunch of helpers for matching in the future
+  #
+  # We do fuzzy matching across the board when checking for internal rule
+  # matches
+  class Rules < Array
+    def initialize(config_target)
```
clintoncwolfe: As mentioned, try to avoid `target`

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+      rules_of_type(_rule.type, opts).last == _rule
+    end
+
+    def rules_of_type(rule_type, opts={:service_name => nil})
+      raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
+
+      service_name = get_service_name(opts[:service_name])
+
+      if @services[service_name]
+        @services[service_name].find_all do |l|
```
clintoncwolfe: Might be more readable if you expand `l` to be a real word.  I think you mean line? Is that the same as `rule`?  This has to be maintained by people who are probably not familiar with PAM.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+      else
+        []
+      end
+    end
+
+    # Determines if one or more rules are contained in the rule set
+    #
+    # @param [Array[String] rules The Rules to find
+    # @param [Hash] opts Options for the include processor
+    # @option opts [Boolean] :exact
```
clintoncwolfe: I think the term `exact` is misleading here - how about `contiguous`?

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+    #   If set, no rules may be present between the rules provided in `rules`
+    #   If unset, the rules simply need to be in the correct order, other rules
+    #   may appear between them
+    # @option opts [String] :service_name The PAM Service under which the rules
+    #   should be searched
+    # @return [Boolean] true if found, false otherwise
+    def include?(rules, opts={:exact => false, :service_name => nil})
+      raise PamError, 'opts must be a hash' unless opts.is_a?(Hash)
```
clintoncwolfe: Maybe ArgumentError here?

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+    #
+    # @return [Array[String]]
+    def to_a
+      self.sort_by{|l| l.type}.map{|l| l.to_s}
+    end
+
+    # Convert the data structure to a String
+    #
+    # @return [String]
+    def to_s
```
clintoncwolfe: InSpec resources typically prefix the data with something like the display name of the resource, e.g. `PAM configuration ...`

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+    #
+    # @param [String] svc_name Optional name of the service that should be
+    #    returned
+    #
+    # @return String
+    def get_service_name(svc_name = nil)
+      return svc_name if svc_name
+
+      if !svc_name && @config_target.directory?
+        raise PamError, 'You must pass ":service_name" as an option!'
```
clintoncwolfe: This seems like an ArgumentError

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+      else
+        return @config_target.basename
+      end
+    end
+  end
+
+  # A single Rule object that has been processed
+  #
+  # Rule equality is a fuzzy match that can accept regular expression matches
```
clintoncwolfe: I think it would be good to specify exactly what is meant by "fuzzy"

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+  # within the string to compare
+  class Rule
+    attr_reader :to_s
+    attr_reader :service, :silent, :type, :control, :module_path, :module_arguments
+
+    def initialize(rule, opts = {})
+      @to_s = rule.strip.gsub(/\s+/,' ')
+
+      rule_regex = <<-'EOM'
+        # Start of Rule
```
clintoncwolfe: :100: for the well-commented regexes!

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+    end
+
+    def match?(to_cmp)
+      to_cmp = Pam::Rule.new(to_cmp, {:service_name => @service}) if to_cmp.is_a?(String)
+
+      # The simple match first
+      self.class == to_cmp.class &&
+        @service.match(Regexp.new("^#{to_cmp.service}$")) &&
+        @type.match(Regexp.new("^#{to_cmp.type}$")) &&
+        @control.match(Regexp.new("^#{to_cmp.control.gsub(/(\[|\])/, '\\\\\\1')}$")) &&
```
clintoncwolfe: OK, I think this one needs to be clarified.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+
+      @service          = opts[:service_name] ? opts[:service_name] : match_data[:service_name]
+      @silent           = match_data[:silent] == '-'
+      @type             = match_data[:type]
+      @control          = match_data[:control]
+      @module_path      = match_data[:module_path]
+      @module_arguments = match_data[:module_args] ? match_data[:module_args].strip.split(/\s+/) : []
+    end
+
+    def match?(to_cmp)
```
clintoncwolfe: Throughout this method, I'm struggling to understand what to_cmp is - a rule to compare against?

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    case @args_type
+    when :all_with_args, :all_without_args, :all_with_integer_arg
+      retval = true
+    when :any_with_args, :any_with_integer_arg
+      retval = false
+    end
+
+    if [:all_with_integer_arg, :any_with_integer_arg].include? @args_type
+      unless Numeric.method_defined?(@args[:operator])
+        fail("Error: Operator '#{@args[:operator]}' is an invalid numeric comparison operator.")
```
clintoncwolfe: Please raise an exception, and mention what it is - a PAM operator, otherwise this may be confusing.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+      end.flatten.join("\n")
+    end
+
+    retval
+  end
+
+  diffable
+
+  # TODO make these an array of args so that we can actually chain them together
+  chain :any_with_args do |args|
```
clintoncwolfe: Overall, I understand the desire to make a matcher that is very RSPec-like.  However, we generally don't expect the user to know RSpec at all, and we have found that chaining matchers is harder to write compared to having matchers that take arguments.

This isn't a blocker for me, but it would have to be well-documented.

clintoncwolfe (changes requested):
A great implementation! A few things I would be looking for:
 * move the custom matcher into the resource, so it is specific to the PAM resource. We don't want people calling the matcher on anything else.
 * A few variable naming gripes, not a blocker
 * Consider not using RSpec chained matchers.  We don't generally do that in other InSpec resources.
 * Needs tests (unit at least) and a dedicated doc file.

Thanks!

lib/matchers/matchers.rb
trevor-vaughan: @clintoncwolfe Ok, will try to do this. For some reason, I couldn't get it to work previously, do you happen to have an example handy that I can look at?

lib/resources/pam.rb
trevor-vaughan: @clintoncwolfe I did this to match the prose in the STIGs and other policies where they say "make sure the following lines are in <whatever>". Do you still want this changed? I figure it's just syntactic sugar to appease reviewers.

lib/resources/pam.rb
trevor-vaughan: No idea how that ever worked :-|

lib/resources/pam.rb
trevor-vaughan: Ok, will fix.

lib/resources/pam.rb
trevor-vaughan: I think I had issues pulling just `stdout` from `inspec.command`. I can look again.

lib/resources/pam.rb
trevor-vaughan: Heh, I think `contiguous` may also confuse people.

Would `exact_order` be better?

lib/resources/pam.rb
trevor-vaughan: Isn't this more understandable? You know that `PamError` comes from the PAM resource not some random place.

lib/resources/pam.rb
trevor-vaughan: This was for comparison purposes, not sure if we want to go throwing in random stuff.

lib/resources/pam.rb
trevor-vaughan: I can give it a shot.

lib/resources/pam.rb
trevor-vaughan: Yep

lib/matchers/matchers.rb
trevor-vaughan: Honestly, this just read better and required less code to figure out in the tests.

I get that users shouldn't have to know rspec but, at a certain point, they're going to have to if they want to write tests that aren't insane levels of weird BASH-fu.

trevor-vaughan:
@clintoncwolfe Thanks for the thorough review! This is exactly what I was looking for before I started hitting the tests heavily.

Not sure when I'll be able to get it all done but will attempt to chip away over time and more than happy for any of the maintainers to update the PR.

lib/matchers/matchers.rb
miah: @trevor-vaughan Check the AWS resources, there are a few examples:

https://github.com/inspec/inspec/blob/master/lib/resources/aws/aws_ec2_instance.rb#L86-L94

https://www.inspec.io/docs/reference/resources/aws_ec2_instance/

miah (changes requested):
Thanks for the contribution @trevor-vaughan. Looks like there are a few changes to address and I'll give this another review. \

lib/matchers/matchers.rb
zenspider: `fail` == `raise`. I prefer `raise` because not everyone knows that... is there a specific exception class that should be raised tho? (I'm new—no clue)

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    end
+  end
+
+  match do |actual|
+    case @args_type
+    when :all_with_args, :all_without_args, :all_with_integer_arg
+      retval = true
+    when :any_with_args, :any_with_integer_arg
+      retval = false
+    end
```
zenspider: This one feels weird to me, but I'm new to this code still. Generally when you have  some sort of condition where one side evaluates to true and the other false you just push the assignment up to be the condition itself.

In this case? I dunno. You should push the assignment up, for sure... but do you drop it down to something simple like `retval = [:all_with_args, :all_without_args, :all_with_integer_arg].include? @args_type` ? you won't catch that the case where `@args_type` isn't one of the other two... but you're already not doing that.

I think... think... in this case, push the return up and add an else that raises an exception saying `@args_type` wasn't handled. This covers all bases that I can think of.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+
+          if potentials && !potentials.empty?
+            actual_munge[service] ||= []
+            actual_munge[service] += potentials.map(&:to_s)
+
+            potentials.each do |potential|
+              case @args_type
+              when :all_without_args
+                retval = !potential.module_arguments.join(' ').match?(@args)
+                throw :stop_searching unless retval
```
zenspider: UGH... I think I'm getting this. This is to pop out of two layers of enumerator? This is getting hairy. Feels like this should be extracted to its own method so a plain `return` will suffice. Also seems like there's some enumerable method(s) that could work better but I'm still just scratching the surface of this.

This is starting to get complex enough that it will be neigh impossible to properly test.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    end
+
+    if actual_munge.empty?
+      @actual = actual.to_s
+    elsif actual_munge.keys.length == 1
+      @actual = actual_munge.values.flatten.join("\n")
+    else
+      @actual = actual_munge.map do |service, lines|
+        lines.map do |line|
+          service + ' ' + line
```
zenspider: Please avoid `String#+` wherever possible. Use interpolation or `String#%` or even `Array.join`.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    when :all_with_args
+      res += ", all with args #{@args}"
+    when :all_without_args
+      res += ", all without args #{@args}"
+    when :all_with_integer_arg
+      res += ", all with arg #{@args[:key]} #{@args[:operator]} #{@args[:value]}"
+    when :any_with_integer_arg
+      res += ", any with arg #{@args[:key]} #{@args[:operator]} #{@args[:value]}"
+    when :any_with_args
+      res += ", any with args #{@args}"
```
zenspider: pull `res +=` up

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    end
+    res
```
zenspider: and/or get rid of this. The above could be `res + case ... end`

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+  end
+end
+
+RSpec::Matchers.define :match_pam_rules do |expected|
+  match do |actual|
+    @expected = expected.to_s
+    @actual = actual.to_s
```
zenspider: are these actually needed if they're not used below?

lib/resources/pam.rb
zenspider: Probably also always refer to it as `PAM` in comments.

lib/resources/pam.rb
zenspider: It's just a big string, right? Is it ever actually evaluated?

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+  # These are here for useful interfaces into the module stack based on
+  # common searches
+  attr_reader :services, :types, :modules
+
+  name 'pam'
+
+  supports platform: 'unix'
+
+  desc 'Use the InSpec pam resource to test the given system pam configuration'
+  example "
```
zenspider: Please use a heredoc for something this big / stringy.

`example <<-EXAMPLE` or something similar.

lib/resources/pam.rb
zenspider: Thank you for lining these up. Totally helps. (please fix the alignment on `config_target` (or all the others to it)).

lib/resources/pam.rb
zenspider: What will `ls` do on a windows box? Is this moot because PAM isn't available there? I honestly have no idea. 😛

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+
+      rules.each do |rule|
+        new_rule = Pam::Rule.new(rule, {:service_name => service})
+
+        # If we hit an 'include' or 'substack' statement, we need to derail and
+        # delve down that tail until we hit the end
+        #
+        # There's no recursion checking here but, if you have a recursive PAM
+        # stack, you're probably not logging into your system anyway
+        if ['include','substack'].include?(new_rule.control)
```
zenspider: %w[...] is great here.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+
+    def match?(to_cmp)
+      to_cmp = Pam::Rule.new(to_cmp, {:service_name => @service}) if to_cmp.is_a?(String)
+
+      # The simple match first
+      self.class == to_cmp.class &&
+        @service.match(Regexp.new("^#{to_cmp.service}$")) &&
+        @type.match(Regexp.new("^#{to_cmp.type}$")) &&
+        @control.match(Regexp.new("^#{to_cmp.control.gsub(/(\[|\])/, '\\\\\\1')}$")) &&
+        @module_path.match(Regexp.new("^#{to_cmp.module_path}$")) &&
```
zenspider: This is heavy handed. Try `@module_path =~ /^#{to_cmp.module_path}$/ &&`

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+        (
+          # Quick test to pass if to_cmp module_arguments are a subset
+          (to_cmp.module_arguments - @module_arguments).empty? ||
+            # All module_arguments in to_cmp should Regex match something
+            to_cmp.module_arguments.all? do |arg|
+              !@module_arguments.grep(Regexp.new("^#{arg}$")).empty?
```
zenspider: same here... Are `to_cmp.module_arguments` regexps fragments? Or just values? If the latter, then it looks like you're trying to say something like:

```ruby
(to_cmp.module_arguments & @module_arguments) == to_cmp.module_arguments
```

zenspider: and I would pull several of these calls out into their own local vars to avoid repetition and make things cleaner reading.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    actual_munge = {}
+
+    @expected = expected.to_s
+
+    if @args_type
+      catch :stop_searching do
+        actual.services.each do |service|
+          expected_line = Pam::Rule.new(expected, {:service_name => service})
+
+          potentials = actual.find_all do |line|
```
zenspider: Please use the Weirich method (braces for value, do/end for side-effect) on find_all and friends.

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+    end
+
+    actual_munge = {}
+
+    @expected = expected.to_s
+
+    if @args_type
+      catch :stop_searching do
+        actual.services.each do |service|
+          expected_line = Pam::Rule.new(expected, {:service_name => service})
```
zenspider: no braces needed

lib/matchers/matchers.rb
```diff
@@ -338,3 +338,164 @@ def format_expectation(negate)
...
+                retval = potential.module_arguments.join(' ').match?(@args)
+                throw :stop_searching unless retval
+              when :all_with_integer_arg
+                retval = matching_integer_arg? potential
+                throw :stop_searching unless retval
+              when :any_with_integer_arg
+                retval = matching_integer_arg? potential
+                throw :stop_searching if retval
+              when :any_with_args
+                retval = potential.module_arguments.join(' ').match?(@args)
```
zenspider: Help me out here. The `join` implies a string and that this is `String#match?`. But `@args` is plural implying a collection and above it is used as a hash with things like `@args[:value]`. This doesn't match `#match?`. Clue me?

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+          (?<type>.+?)\s+
+        # Capture Control
+          (?<control>(\[.+\]|.+?))\s+
+        # Capture Module Path
+          (?<module_path>.+?(\.so)?)
+        # Capture Module Args
+          (\s+(?<module_args>.+?))?
+        # End of Rule
+          $
+      EOM
```
zenspider:
```ruby
# original string-based regexp
rule_regex = <<-'EOM'
           # Start of Rule
             ^
           # Ignore initial Whitespace
             \s*
           # Capture Silent Flag
             (?<silent>-)?
         EOM

         unless opts[:service_name]
           rule_regex += <<-'EOM'
             # Capture Service
               (?<service_name>.+?)\s+
           EOM
         end

         rule_regex += <<-'EOM'
           # Capture Type
             (?<type>.+?)\s+
           # Capture Control
             (?<control>(\[.+\]|.+?))\s+
           # Capture Module Path
             (?<module_path>.+?(\.so)?)
           # Capture Module Args
             (\s+(?<module_args>.+?))?
           # End of Rule
             $
         EOM
```
versus:
```ruby
# regexp with /x flag & interpolation:
%r%
  # Start of Rule
    ^
  # Ignore initial Whitespace
    \s*
  # Capture Silent Flag
    (?<silent>-)?
  # Capture Service
    #{'(?<service_name>.+?)\s+' unless opts[:service_name]}
  # Capture Type
    (?<type>.+?)\s+
  # Capture Control
    (?<control>(\[.+\]|.+?))\s+
  # Capture Module Path
    (?<module_path>.+?(\.so)?)
  # Capture Module Args
    (\s+(?<module_args>.+?))?
  # End of Rule
    $
%x
```
versus:
```ruby
# cleaned up:
sn = opts[:service_name]
%r%
    ^                                      # Start of Rule
    \s*                                    # Ignore initial Whitespace
    (?<silent>-)?                          # Capture Silent Flag
    #{'(?<service_name>.+?)\s+' unless sn} # Capture Service
    (?<type>.+?)\s+                        # Capture Type
    (?<control>(\[.+\]|.+?))\s+            # Capture Control
    (?<module_path>.+?(\.so)?)             # Capture Module Path
    (\s+(?<module_args>.+?))?              # Capture Module Args
    $                                      # End of Rule
%x
```

Now that the comments aren't being noise, I can start to pattern match potential problems. Should some of those `.+?` fragments be `\S+` or are they allowed to have whitespace in them? Do you want to capture that last `\s+` in the `module_args` bit? It could be `\s*` on the tail of the previous to be constent but still compatible w/ the last `?`.

lib/resources/pam.rb
```diff
@@ -0,0 +1,370 @@
...
+      @module_path      = match_data[:module_path]
+      @module_arguments = match_data[:module_args] ? match_data[:module_args].strip.split(/\s+/) : []
+    end
+
+    def match?(to_cmp)
+      to_cmp = Pam::Rule.new(to_cmp, {:service_name => @service}) if to_cmp.is_a?(String)
+
+      # The simple match first
+      self.class == to_cmp.class &&
+        @service.match(Regexp.new("^#{to_cmp.service}$")) &&
```
zenspider: `obj.match(Regexp.new("^#{str_expr}$")) => obj =~ /^#{str_expr}$/`. Am I missing something?

zenspider (changes requested):

kekaichinose:
@trevor-vaughan Can you verify where you're at on this PR?

trevor-vaughan:
@kekaichinose Heh, I'm currently somewhere between no time and extra no time :-(.

If someone else wants to take a whack at it, I've left it open to edits from maintainers.

jerryaldrichiii:
Talked a bit with @trevor-vaughan in Slack. Flagging this for adoption.

miah:
Hi @trevor-vaughan. Have you been able to write any tests for this? Anything would be better than nothing.

lib/matchers/matchers.rb
trevor-vaughan: Honestly, I was just building this from other examples. I'm also not 100% sure how it works.

Added the exception case in case someone adds something that isn't supported.

lib/resources/pam.rb
trevor-vaughan: Honestly, I'll probably have to test this one some more to be able to answer all of your questions. I'll start with it as-is and then it can be refined over time.

Thanks for compressing the regex though, that was helpful and I'll update it accordingly.

lib/resources/pam.rb
trevor-vaughan: It was easier to read in the larger logic chaining.

lib/matchers/matchers.rb
trevor-vaughan: Pretty sure I inherited the notation from the general matchers to keep it consistent. Honestly, some of the matcher bits were pulled together from examples because not much of it was super clear and relies on internal RSpec items that I didn't dig very far into.

trevor-vaughan:
@miah Making progress but can't figure out how to properly stub this https://github.com/inspec/inspec/pull/3993/files#diff-f454d67d1eac3dc9154b7e7fab1a18a8R63

Help would be appreciated.

lib/resources/pam.rb
trevor-vaughan: @zenspider No idea honestly. Is there another way in Inspec to get a listing of a directory on a remote host?

trevor-vaughan:
Ok, I think I'll just make a couple of methods to stub them out and that should work. Still making progress.

trevor-vaughan:
Hi All. Need some help figuring out why these expectations aren't being matched if possible: https://github.com/inspec/inspec/pull/3993/files#diff-d013311217b3b201a25622574fa60531R27-R28

lib/resources/pam.rb
zenspider: I disagree. Longer is not easier to read and the extra calls means extra reasoning about what it is doing:

```ruby
obj.match(Regexp.new("^#{str_expr}$"))
obj =~ /^#{str_expr}$/
```

and logical chaining just means even more reasoning, so please be as clear and succinct as possible.

clintoncwolfe:
Hi, this is is stale enough that I think we're going to close this for now. If you decide to revisit this, we welcome it.

PR #3998, opened 2019-04-30
Fix promotion of the inspec-bin gem in Expeditor

tas50:
The previous config failed to promote inspec-bin to stable.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -6,8 +6,7 @@ product_key: inspec
 rubygems:
  - inspec
  - inspec-core
- - inspec-bin:
-     gemspec_path: ./inspec-bin/
+ - inspec-bin
 
 docker_images:
   - inspec:


tyler-ball (approved):

jaymalasinha (changes requested):
Referring the docs https://expeditor.chef.io/docs/integrations/rubygems/#configuration  current config looks correct as the default gemspec_path is root ( so we need `./inspec-bin/`.  Not sure why the publish for inspec-bin fails though @tduffield ?

tduffield (changes requested):
Yeah, this will actually break things.

tyler-ball:
@tduffield @tas50 What is the correct fix to get gems promoting?

tduffield:
@tyler-ball 1. make sure that `rubygems@chef.io` is an owner of the gem (if the gem does not already exist). 2. If the gem already exists, `gem push` returns a non-zero exit code which would cause the failure I believe.

PR #4000, opened 2019-04-30
Apply new UI code to CLI and plugin-manager-cli

miah:
This PR replaces existing 'red', 'bold' and 'puts '-' * 50' type code with the new UI code. My goal was simply updating the code to use the new UI, I didn't attempt to do things better. =)

I think when we discuss overhauling the UI/UX of InSpec we should create some templates for UI output so that we can further simplify this code.

ref: https://github.com/inspec/inspec/issues/3715

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -123,15 +123,15 @@ class Inspec::InspecCLI < Inspec::BaseCLI
       puts JSON.generate(result)
     else
       %w{location profile controls timestamp valid}.each do |item|
-        prepared_string = format('%-12s %s',
+        prepared_string = format("%-12s %s",
                                  "#{item.to_s.capitalize} :",
                                  result[:summary][item.to_sym])
         ui.plain_line(prepared_string)
       end
       puts
 
-      if result[:errors].empty? and result[:warnings].empty?
-        ui.plain_line('No errors or warnings')
+      if result[:errors].empty? && result[:warnings].empty?
+        ui.plain_line("No errors or warnings")
       else
         item_msg = lambda { |item|
           pos = [item[:file], item[:line], item[:column]].compact.join(":")
@@ -198,7 +198,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     result = profile.check
 
     if result && !o[:ignore_errors] == false
-      o[:logger].info 'Profile check failed. Please fix the profile before generating an archive.'
+      o[:logger].info "Profile check failed. Please fix the profile before generating an archive."
       return ui.exit Inspec::UI::EXIT_USAGE_ERROR
     end
 
@@ -308,7 +308,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     if o["format"] == "json"
       puts res.to_json
     else
-      ui.headline('Platform Details')
+      ui.headline("Platform Details")
       ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0, color: 36)
     end
   rescue ArgumentError, RuntimeError, Train::UserError => e
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -148,7 +148,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
         ui.plain_line("Summary:     #{errors}, #{warnings}")
       end
     end
-    ui.exit 1 unless result[:summary][:valid]
+    ui.exit Inspec::UI::EXIT_USAGE_ERROR unless result[:summary][:valid]
   rescue StandardError => e
     pretty_handle_exception(e)
   end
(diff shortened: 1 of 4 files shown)


lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -57,23 +56,22 @@ def search(search_term) # rubocop: disable Metrics/AbcSize
...
+        ui.line
+        ui.plain(" #{search_results.count} plugin(s) found")
         puts
 
-        exit 2 if search_results.empty?
+        ui.exit 2 if search_results.empty?
       rescue Inspec::Plugin::V2::SearchError => ex
         Inspec::Log.error ex.message
-        exit 1
+        ui.exit 1
```
clintoncwolfe: Consider using `Inspec::UI::EXIT_USAGE_ERROR` instead of `1`

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -117,13 +115,14 @@ def update(plugin_name)
         begin
           installer.update(plugin_name)
         rescue Inspec::Plugin::V2::UpdateError => ex
-          puts(red { 'Update error: ' } + ex.message + ' - update failed')
```
clintoncwolfe: The formatting here would be a bit different - the original has only 'Update error:` in red, whereas the new code would have teh whole line in red.

I think it looks better with just the first part in red, myself.

Similar feedback for other locations - of course, there may be something about the `ui.red` method that makes that hard.

clintoncwolfe (commented):
Looks great, and very thorough! I have just two things, and neither is a blocker:
 * maybe use the exit code constants
 * consider breaking out the color calls to only highlight part of the line in the plugin manager

lib/inspec/cli.rb
```diff
@@ -99,36 +99,32 @@ def check(path) # rubocop:disable Metrics/AbcSize
       puts JSON.generate(result)
     else
       %w{location profile controls timestamp valid}.each do |item|
-        puts format('%-12s %s', item.to_s.capitalize + ':',
-                    mark_text(result[:summary][item.to_sym]))
+        prepared_string = format('%-12s %s',
+                                 (item.to_s.capitalize + ':'),
```
zenspider: This isn't your change, but we should avoid string concatenation whenever possible. That `+ ':'` could just be part of the format string for free.

lib/inspec/cli.rb
```diff
@@ -99,36 +99,32 @@ def check(path) # rubocop:disable Metrics/AbcSize
...
+        result[:errors].each { |item| ui.red " #{Inspec::UI::GLYPHS[:script_x]}  #{item_msg.call(item)}\n" }
+        result[:warnings].each { |item| ui.yellow " !  #{item_msg.call(item)}\n" }
 
         puts
-        puts format('Summary:     %s%d errors%s, %s%d warnings%s',
-                    red, result[:errors].length, rst,
-                    yellow, result[:warnings].length, rst)
+
+        errors = ui.red(result[:errors].length.to_s + ' errors', print: false)
+        warnings = ui.yellow(result[:warnings].length.to_s + ' warnings', print: false)
```
zenspider: In the same sense, whenever you see `x.to_s + y`, that can be much better expressed with interpolation or `String#%`:

```ruby
# using interpolation
errors = ui.red("#{result[:errors].length} errors", print: false)
warnings = ui.yellow("#{result[:warnings].length} warnings", print: false)

# using String#% (aka format)
errors = ui.red("%d errors" % [result[:errors].length], print: false)
warnings = ui.yellow("%d warnings" % [result[:warnings].length], print: false)
```

Totally minor in this context tho.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -346,37 +362,49 @@ def install_from_remote_gem_verson_preflight_check(plugin_name, requested_versio
...
-        puts 'If you disagree with this determination, please accept our apologies for the misunderstanding, and open an issue at https://github.com/inspec/inspec/issues/new'
-        exit 2
+        ui.red("Plugin on Exclusion List - #{plugin_name} is listed as an " \
+               'incompatible gem - refusing to install.')
+        ui.plain("Rationale: #{ex.details.rationale}")
+        ui.plain('Exclusion list location: ' +
+                 File.join(Inspec.src_root, 'etc', 'plugin_filters.json'))
+        ui.plain('If you disagree with this determination, please accept ' \
+                 'our apologies for the misunderstanding, and open an issue ' \
+                 'at https://github.com/inspec/inspec/issues/new')
```
zenspider: THANK YOU

zenspider (approved):
Some feedback to consider but no requested changes. Looks good.

miah:
I've got two more tests to update on this one, update incoming.

miah:
I just rebased against master so this doesn't get too stale. =)

clintoncwolfe (approved):

miah:
I rebased this against master, after the tests complete if they are still green I'm going to merge.

PR #4007, opened 2019-05-01
Improve testing on inputs that do not have values

### LINKED ISSUE #3919: Using InSpec Inputs (Attributes) requires `description:`
### 🎛 Description

When using attributes, `description:` is required, but I don't see it in the documentation nor does the error hint that that is the case.


This fails:

```
[~/Demo/InSpec]$ cat example.rb                                                                                                                                                                                                                                                           
example = attribute('example')

describe example do
  it { should_not be_empty }
end
[~/Demo/InSpec]$ cat inputs.yml                                                                                                                                                                                                                                                           
example: 'This is an example'
[~/Demo/InSpec]$ inspec exec example.rb --attrs inputs.yml                                                                                                                                                                                                                                
Profile 'tests from example.rb' does not have an input with name 'my_example'
```

This succeeds:

```
[~/Demo/InSpec]$ cat example.rb                                                                                                             
example = attribute('example', description: 'This is required')

describe example do
  it { should_not be_empty }
end
[~/Demo/InSpec]$ cat inputs.yml                                                                                                             
example: 'This is an example'
[~/Demo/InSpec]$ inspec exec example.rb --attrs inputs.yml                                                                                  

Profile: tests from example.rb (tests from example.rb)
Version: (not specified)
Target:  local://

  This is an example
     ✔  should not be empty

```

3.6.15

### 💁 Possible Solutions

Change the error output to indicate `description:` is required, or make `description:` not mandatory. I personally prefer the latter.

### LINKED ISSUE #3955: Attribute has to be defined prior to being called.
### 🎛 Description

🐛 I have a boolean attribute in a profile if the attribute is not scoped out in either the `inspec.yml` or the profile itself. When I use the attribute in an `only_if` there is no warning or error that the attribute is being ignored. If I use a `puts` with the attribute I get a message that the profile doesn't have an attribute with that name. 

3.9.0 on windows and linux

### 🤔 Replication Case
test file
```ruby
puts ''
puts ''
puts attribute('bar')

control 'foo' do 
  title 'foo'
  desc 'foo'
  only_if { attribute('bar') != false }
  describe file('foo') do
    it{ should exist }
  end
end
```

insepce.yml
```
name: foo
title: foo
summary: foo
version: 0.1.0
maintainer: foo
```

Output
```
Profile 'foo' does not have an input with the name 'bar'
```

if the profile is run without the puts and using the `--attrs` with the `bar` attribute set to false there is no error and the control is not skipped.

### 💁 Possible Solutions

When an attribute is specified that is not declared even in an only_if produce an error, or when an attribute is used in a profile that has not been defined infer it's type and warn in the output or debug log that the attribute has been used but had a value of `nil`.

clintoncwolfe:
Closes #3955 
Closes #3919 

3955 pointed out that InSpec v3 has undefined and unpleasant behavior around using inputs (attributes) that do not have values. In the input core overhaul, merged into inspec v4.2 and released on 4.3.2, this issues is resolved, but had no specific tests.

3919 pointed out that adding a description to an attribute was apparently enough to get inspec v3 to no longer complain about the valueless input. As v4.2+ does not exhibit this behavior, we just need tests to show that valueless attributes are accepted, without description. 

This PR adds such tests to cover both issues.

### DIFF
--- a/test/functional/inputs_test.rb
+++ b/test/functional/inputs_test.rb
@@ -106,4 +106,14 @@ describe 'inputs' do
 
   #   # TODO - add test for backwards compatibility using 'attribute' in DSL
   end
+
+  describe 'when using a profile with undeclared (valueless) inputs' do
+    it 'should warn about them and not abort the run' do
+      cmd = "exec #{inputs_profiles_path}/undeclared"
+      result = run_inspec_process(cmd, json: true)
+      result.stderr.must_include "WARN: Input 'undeclared_01'"
+      result.stderr.must_include 'does not have a value'
+      result.must_have_all_controls_passing
+    end
+  end
 end
--- a/test/unit/mock/profiles/inputs/undeclared/controls/undeclared.rb
+++ b/test/unit/mock/profiles/inputs/undeclared/controls/undeclared.rb
@@ -1,0 +1,28 @@
+control 'start_marker' do
+  describe('dummy_test_01') do
+    it { should cmp 'dummy_test_01'}
+  end
+end
+
+control 'undeclared_in_control_body' do
+  attribute('undeclared_01')
+  assignment_outcome = attribute('undeclared_02')
+  describe('dummy_test_02') do
+    it { should cmp 'dummy_test_02'}
+  end
+end
+
+control 'undeclared_in_only_if' do
+  only_if { attribute('undeclared_03') }
+  describe('dummy_test_03') do
+    it { should cmp 'dummy_test_03'}
+  end
+end
+
+attribute('undeclared_04')
+
+control 'end_marker' do
+  describe('dummy_test_04') do
+    it { should cmp 'dummy_test_04'}
+  end
+end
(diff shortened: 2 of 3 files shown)


miah (approved):

test/functional/inputs_test.rb
```diff
@@ -106,4 +106,14 @@
 
   #   # TODO - add test for backwards compatibility using 'attribute' in DSL
   end
+
+  describe 'when using a profile with undeclared (valueless) inputs' do
+    it 'should warn about them and not abort the run' do
+      cmd = "exec #{inputs_profiles_path}/undeclared"
+      result = run_inspec_process(cmd, json: true)
+      result.stderr.must_include "WARN: Input 'undeclared_01'"
+      result.stderr.must_include 'does not have a value'
```
zenspider, line 115: Is this 2 lines for readability? You can use `must_match` to skip the middle bits but ensure they're one warning.

clintoncwolfe: I like to check for separate elements of an error message under separate expectations. Really I should have done it one three lines: A warning, it mentions the input by name, and the complaint is that it does not have a value.

zenspider: Separate include checks could pass via completely unrelated lines, or in any order, which is why I would prefer a regexp that checks everything in one fell swoop.

Accepting despite this, so we can get it through.

zenspider (approved):

Issue #3880, opened 2019-03-08
Attribute->Input rename - Documentation Update


clintoncwolfe:
### 🎛 Description

As part of #3802, update existing documentation to refer to attributes as inputs.

Retain mentions of the former name.

Check for broken links going to anchor URLs whose section names are changing.

clintoncwolfe:
This is being worked on branch [cw/input-docs](https://github.com/inspec/inspec/tree/cw/input-docs)

Issue #3853, opened 2019-03-01
Attribute -> Input rename: Rename DSL access points


clintoncwolfe:
### 🎛 Description

🙋 feature request

Renamed the control DSL method `attribute` to `input`.

Continue to support `attribute`.

Like `attribute`, `input` should automatically evaluate to the value of the Input object.

Introduce a new DSL method, `input_object`, which returns the unevaluated Input.  For advanced usage, document the diagnostic features of the object.

Update docs.

Update tests (this has a large footprint).

clintoncwolfe:
This was merged on #4008

PR #4008, opened 2019-05-01
Rename 'attribute' DSL method to 'input'

clintoncwolfe:
As part of the attribute => input rename (#3802), this PR adds a new control DSL method, `input`, and makes `attribute` a wrapper for it, with an ignore-mode deprecation call.

The code could use some DRYing up - this code doesn't introduce any new duplication, but the existing method is defined in three places, with nearly the same implementation.

A 'new way' and 'old way' test fixture is provided.

Docs are intentionally omitted; we plan a big, nice new page detailing Inputs on #3880.

Partial fix on #3853 ; we also need a massive PR to go through and update all of the test fixtures and examples to use the new DSL method.

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -26,8 +26,7 @@ module Inspec
         with_resource_dsl resources_dsl
 
         # allow attributes to be accessed within control blocks
-        # TODO: deprecate name, use input()
-        define_method :attribute do |input_name, options = {}|
+        define_method :input do |input_name, options = {}|
           if options.empty?
             # Simply an access, no event here
             Inspec::InputRegistry.find_or_register_input(input_name, profile_id).value
@@ -43,6 +42,11 @@ module Inspec
         # Will return nil on a miss.
         define_method :input_object do |input_name|
           Inspec::InputRegistry.find_or_register_input(input_name, profile_id)
+        end
+
+        define_method :attribute do |name, options = {}|
+          Inspec.deprecate(:attrs_dsl, "Input name: #{name}, Profile: #{profile_id}")
+          input(name, options)
         end
 
         # Support for Control DSL plugins.
@@ -182,9 +186,7 @@ module Inspec
           profile_context_owner.register_rule(control, &block) unless control.nil?
         end
 
-        # method for inputs; import input handling
-        # TODO: deprecate name, use input()
-        define_method :attribute do |input_name, options = {}|
+        define_method :input do |input_name, options = {}|
           if options.empty?
             # Simply an access, no event here
             Inspec::InputRegistry.find_or_register_input(input_name, profile_id).value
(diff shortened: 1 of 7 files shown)


miah (approved):

lib/inspec/rspec_extensions.rb
```diff
@@ -65,14 +65,30 @@ def method_missing(method_name, *arguments, &block)
 
 class RSpec::Core::ExampleGroup
   # This DSL method allows us to access the values of inputs within InSpec tests
-  def attribute(name)
-    Inspec::InputRegistry.find_or_register_input(name, self.class.metadata[:profile_id]).value
+  def input(input_name, options = {})
```
zenspider, line 68: Maybe add `options[:value] ||= input_name` to DRY up your tests below and keep it API compatible to the old name?

lib/inspec/control_eval_context.rb
```diff
@@ -35,7 +35,7 @@ def self.rule_context(resources_dsl, profile_id)
             options[:priority] = 20
             options[:provider] = :inline_control_code
             evt = Inspec::Input.infer_event(options)
-            Inspec::InputRegistry.find_or_register_input(input_name, profile_name, event: evt).value
+            Inspec::InputRegistry.find_or_register_input(input_name, profile_id, event: evt).value
```
zenspider, line 37: This doesn't track, so I assume it is the subject of the question in the commit message.

Is it possible / worthwhile to write a test to reproduce this failure (and thus prevent regressions)?

lib/inspec/rspec_extensions.rb
zenspider: (where "below" means in the whole diff, not in this commit... that confused myself :P)

lib/inspec/control_eval_context.rb
clintoncwolfe: This codepath is well exercised by tests - anytime we make an input / attribute with options (for example, a value) this should get touched. For example, the 'scoping' test profile tests this.

I really don't know how this was working at all when we merged in the inputs overhaul. Not reassuring, but it is well-tested at this point.

lib/inspec/rspec_extensions.rb
clintoncwolfe: Do you mean `options[:value] ||= value` ?

zenspider: Ah. Sorry for being unclear. If you default `options[:value] ||= input_name` (my original comment had `name`, confusingly), then that can clean up your diffs below and dry it up a bit:

```diff
-  subject { attribute('test-03') } 
+  subject { input('test-03') }
```

clintoncwolfe: It is valid for an input (which is a user-facing DSL feature) to have an specified value, as well as it to have an explicit nil or an explicit false. Defaulting to the name of the input would be very surprising behavior; I don't think it's appropriate.

I'm sure the tests could probably be DRY'd up.

zenspider (approved):

PR #4025, opened 2019-05-03
Enable user telemetry opt-in / opt-out on cli

### LINKED ISSUE #3849: Telemetry - Opt In/Out
As a user I want to opt in or out of Telemetry Collection. I should be able to do this with command line arguments, as well as through a configuration file. I may want to configure this through the ENVIRONMENT too.

miah:
This adds `--enable-telemetry` and `--no-enable-telemetry` to the cli. Currently telemetry is enabled by default.

I wanted to also add the ability to opt-out and specify additional configuration via our config file, but thats going to require another issue as its not as straight forward (our config file is versioned and making changes is currently non-trivial).

closes https://github.com/inspec/inspec/issues/3849

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -39,6 +39,9 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :disable_user_plugins, type: :string, banner: '',
     desc: 'Disable loading all plugins that the user installed.'
 
+  class_option :telemetry, type: :boolean,
+    desc: 'Allow or disable telemetry', default: true
+
   require 'license_acceptance/cli_flags/thor'
   include LicenseAcceptance::CLIFlags::Thor
 
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -39,8 +39,8 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :disable_user_plugins, type: :string, banner: '',
     desc: 'Disable loading all plugins that the user installed.'
 
-  class_option :telemetry, type: :boolean,
-    desc: 'Allow or disable telemetry', default: true
+  class_option :enable_telemetry, type: :boolean,
+    desc: 'Allow or disable telemetry', default: false
 
   require 'license_acceptance/cli_flags/thor'
   include LicenseAcceptance::CLIFlags::Thor
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -67,6 +67,10 @@ module Inspec
       puts
     end
 
+    def telemetry_enabled?
+      final_options['telemetry']
+    end
+
     #-----------------------------------------------------------------------#
     #                      Train Credential Handling
     #-----------------------------------------------------------------------#
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -70,7 +70,7 @@ module Inspec
     # return all telemetry options from config
     # @return [Hash]
     def telemetry_options
-      final_options.select { |key,_| key.include?('telemetry') }
+      final_options.select { |key, _| key.include?('telemetry') }
     end
 
     #-----------------------------------------------------------------------#
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -67,8 +67,10 @@ module Inspec
       puts
     end
 
-    def telemetry_enabled?
-      final_options['telemetry']
+    # return all telemetry options from config
+    # @return [Hash]
+    def telemetry_options
+      final_options.select { |key,_| key.include?('telemetry') }
     end
 
     #-----------------------------------------------------------------------#
--- a/lib/inspec/utils/telemetry/collector.rb
+++ b/lib/inspec/utils/telemetry/collector.rb
@@ -12,6 +12,7 @@ module Inspec::Telemetry
       load_config
     end
 
+    # Allow loading a configuration, useful when testing.
     def load_config(config = Inspec::Config.cached)
       @config = config
     end
--- a/lib/inspec/utils/telemetry/collector.rb
+++ b/lib/inspec/utils/telemetry/collector.rb
@@ -11,6 +11,7 @@ module Inspec::Telemetry
 
     def initialize
       @data_series = []
+      @telemetry_toggled_off = false
       load_config
     end
 
@@ -29,13 +30,16 @@ module Inspec::Telemetry
     # telemetry, if not default to false.
     # @return [True, False]
     def telemetry_enabled?
-      config_telemetry_options.fetch('enable_telemetry', false)
+      if @telemetry_toggled_off
+        false
+      else
+        config_telemetry_options.fetch('enable_telemetry', false)
+      end
     end
 
     # A way to disable the telemetry system.
-    # @return [True]
     def disable_telemetry
-      @enabled = false
+      @telemetry_toggled_off = true
     end
 
     # The entire data series collection.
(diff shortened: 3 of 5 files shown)


test/unit/utils/telemetry/collector_test.rb
```diff
@@ -40,4 +40,14 @@ def test_reset_singleton
...
   end
+
+  def test_telemetry_disabled
+    @collector.load_config(Inspec::Config.mock(telemetry: false))
+    refute @collector.telemetry_enabled?
+  end
+
+  def test_telemetry_enabled
+    @collector.load_config(Inspec::Config.mock(telemetry: true))
+    assert @collector.telemetry_enabled?
```
miah, line 56: I feel like these tests are pretty silly since I'm just testing that the configuration got loaded and a True/False was set, is there a better way to do this?

lib/inspec/config.rb
```diff
@@ -67,6 +67,10 @@ def diagnose
       puts
     end
 
+    def telemetry_enabled?
+      final_options['telemetry']
+    end
+
```
miah, line 75: I felt it was better to add an accessor here rather than our collector needing to know the format of our options hash and reaching inside.

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -21,7 +27,11 @@ def add_data_series(data_series)
     # Always true until we add configuration parsing.
     # @return [True, False]
     def telemetry_enabled?
-      @enabled
+      if @config.telemetry_enabled?.nil?
+        @enabled = true
+      else
+        @enabled = @config.telemetry_enabled?
+      end
     end
```
miah, line 38: I know this is a bit weird if `@config` is nil and thus the `telemetry_enabled?` method won't exist. This might be more problematic if InSpec is used as a 'library', but if the cli/config is loaded properly we should be good. Suggestions?

zenspider: I'm not sure I understand your comment here. If `@config` is nil, that first line will fail calling `telemetry_enabled?` (unless we're doing hinky things on `NilClass`) and will raise a `NoMethodError`.

test/unit/utils/telemetry/collector_test.rb
zenspider: I would `assert`/`refute` before the `load_config` to show state change. Otherwise, I think these tests are fine.

zenspider (changes requested):
Minor change requested to 2 tests.

lib/inspec/config.rb
clintoncwolfe: I'm unsure about this. Should the config object have *general* knowledge of the  major subsystems of InSpec ? I think yes (it already know about train credentials, for example). 

However, asking specifically if the telemetry system is enabled means that when you add another config option - like endpoint URL - you'll need to add another method to the Config (or else scatter such methods between the Collector and Config).

So, I think if you asked the Config, "I don't care how you got it, but can you give me all the Telemetry options as a hash" would be fine - Collector doesn't have to care about Config's innards, and vice-versa.

clintoncwolfe: I'm unsure about this. Should the config object have *general* knowledge of the  major subsystems of InSpec ? I think yes (it already know about train credentials, for example). 

However, asking specifically if the telemetry system is enabled means that when you add another config option - like endpoint URL - you'll need to add another method to the Config (or else scatter such methods between the Collector and Config).

So, I think if you asked the Config, "I don't care how you got it, but can you give me all the Telemetry options as a hash" would be fine - Collector doesn't have to care about Config's innards, and vice-versa.

lib/inspec/cli.rb
```diff
@@ -39,6 +39,9 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   class_option :disable_user_plugins, type: :string, banner: '',
     desc: 'Disable loading all plugins that the user installed.'
 
+  class_option :telemetry, type: :boolean,
+    desc: 'Allow or disable telemetry', default: true
```
clintoncwolfe: I'm pretty sure this should default to false - we're doing opt-in on other Chef products, IIRC...

clintoncwolfe: Very weak suggestion, Consider naming this `--enable-telemetry` - I always like verb phrases

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -8,7 +9,12 @@ class Collector
```
clintoncwolfe, line 9: I suggest exposing @config via attr_reader, so we avoid the ivar accesses (which are vulnerable to typos)

lib/inspec/utils/telemetry/collector.rb
clintoncwolfe: Inspec::Config.cached will still work if `inspec` is used as a library. Since that is the default, @config would only be nil if someone were actively trying to break it :-)

clintoncwolfe (changes requested):
Thanks! I have a couple of changes I'd like:
 * Pretty sure this should be opt-in - let's verify with other teams, maybe Mike
 * I think the Telemetry-Config interface should know less about each other - see comment. A Hash would probably be fine for an interface.

lib/inspec/utils/telemetry/collector.rb
miah: My only issue is that `Inspec::Config.cached` doesn't work 'right' if you are testing the collector class. This is why I'm passing in a mock config object.

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -17,17 +26,20 @@ def add_data_series(data_series)
...
-    # Always true until we add configuration parsing.
+    # The loaded configuration should have a option to configure
+    # telemetry, if not default to false.
     # @return [True, False]
     def telemetry_enabled?
-      @enabled
+      if @telemetry_toggled_off
+        false
+      else
+        config_telemetry_options.fetch('enable_telemetry', false)
```
zenspider, line 36: I usually like my booleans to be positively oriented, because logical statements like this usually throw me for a bit of a De Morgan-y loop mentally. 

Also, if statements that have a return value of true/false in them is usually a sign that the logic can be cleaned up into an or/and expression... but see point #1. 😛

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -52,9 +64,18 @@ def find_or_create_data_series(name)
     end
 
     # Blanks the contents of the data series collection.
+    # Reset telemetry toggle
     # @return [True]
-    def reset
+    def reset!
```
zenspider, line 69: good

lib/inspec/utils/telemetry/collector.rb
```diff
@@ -52,9 +64,18 @@ def find_or_create_data_series(name)
...
       @data_series = []
+      @telemetry_toggled_off = false
+    end
+
+    private
+
+    # Minimize exposure of Inspec::Config interface
+    def config_telemetry_options
```
zenspider, line 77: good!

test/unit/utils/telemetry/collector_test.rb
```diff
@@ -37,7 +37,24 @@ def test_find_or_create_data_series
   def test_reset_singleton
     data_series = Inspec::Telemetry::DataSeries.new('/resource/File')
     @collector.add_data_series(data_series)
-    @collector.reset
+    @collector.reset!
     assert_equal 0, @collector.list_data_series.count
```
zenspider, line 41: Not part of your PR but `assert_empty` is more expressive here.

zenspider (approved):
Some minor feedback but looks good!

lib/inspec/utils/telemetry/collector.rb
miah: Ok. I'll give this one some more thought and see how I can make it better.

clintoncwolfe (approved):
I'm happy with it - thanks!

PR #4023, opened 2019-05-03
Modernize use of Minitest.

zenspider:
+ Turn off verbosity in Rakefile by default. Use `rake V=1` to turn back on.
+ MiniTest -> Minitest everywhere.
+ MiniTest::Unit::TestCase -> Minitest::Test everywhere.
+ Updated minitest doco urls to official and up-to-date site.
+ Normalize requires. Only needs "minitest/autorun" and "minitest/pride".

Signed-off-by: Ryan Davis <zenspider@chef.io>

PR #4031, opened 2019-05-06
Fix up sporadically failing tests and ensure tests can run independently

zenspider:
Most of the info is in the commit messages... I should whip up some way to collate them into something I can paste here. 😕

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ before_install:
 - gem update --system
 - gem --version
 - bundle --version
+env:
+  - SLOW=1
 matrix:
   include:
   - rvm: 2.4.5
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -113,9 +113,8 @@ module Inspec::Plugin::V2
     end
 
     def self.plugin_gem_path
-      # I can't believe there isn't a simpler way of getting this
-      # 2.4.2.p123 => 2.4.0
-      ruby_abi_version = (Gem.ruby_version.segments[0, 2] << 0).join('.')
+      require 'rbconfig'
+      ruby_abi_version = RbConfig::CONFIG['ruby_version']
       File.join(Inspec.config_dir, 'gems', ruby_abi_version)
     end
 
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -115,6 +115,7 @@ module Inspec::Plugin::V2
     def self.plugin_gem_path
       require 'rbconfig'
       ruby_abi_version = RbConfig::CONFIG['ruby_version']
+      # TODO: why are we installing under the api directory for plugins?
       File.join(Inspec.config_dir, 'gems', ruby_abi_version)
     end
 
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -246,26 +246,26 @@ RSpec::Matchers.define :cmp do |first_expected| # rubocop:disable Metrics/BlockL
     # if actual and expected are strings
     if expected.is_a?(String) && actual.is_a?(String)
       return actual.casecmp(expected) == 0 if op == :==
-      return Gem::Version.new(actual).method(op).call(Gem::Version.new(expected)) if
+      return Gem::Version.new(actual).send(op, Gem::Version.new(expected)) if
         version?(expected) && version?(actual)
     elsif expected.is_a?(Regexp) && (actual.is_a?(String) || actual.is_a?(Integer))
       return !actual.to_s.match(expected).nil?
     elsif expected.is_a?(String) && integer?(expected) && actual.is_a?(Integer)
-      return actual.method(op).call(expected.to_i)
+      return actual.send(op, expected.to_i)
     elsif expected.is_a?(String) && boolean?(expected) && [true, false].include?(actual)
-      return actual.method(op).call(to_boolean(expected))
+      return actual.send(op, to_boolean(expected))
     elsif expected.is_a?(Integer) && integer?(actual)
-      return actual.to_i.method(op).call(expected)
+      return actual.to_i.send(op, expected)
     elsif expected.is_a?(Float) && float?(actual)
-      return actual.to_f.method(op).call(expected)
+      return actual.to_f.send(op, expected)
     elsif actual.is_a?(Symbol) && expected.is_a?(String)
-      return actual.to_s.method(op).call(expected)
+      return actual.to_s.send(op, expected)
     elsif octal?(expected) && actual.is_a?(Integer)
-      return actual.method(op).call(expected.to_i(8))
+      return actual.send(op, expected.to_i(8))
     end
 
     # fallback to simple operation
-    actual.method(op).call(expected)
+    actual.send(op, expected)
   rescue NameError => _
     false
   rescue ArgumentError
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -217,13 +217,13 @@ RSpec::Matchers.define :cmp do |first_expected| # rubocop:disable Metrics/BlockL
   def float?(value)
     Float(value)
     true
-  rescue ArgumentError => _ex
+  rescue ArgumentError, TypeError
     false
   end
 
   def octal?(value)
     return false unless value.is_a?(String)
-    !(value =~ /\A0+\d+\Z/).nil?
+    !(value =~ /\A0+[0-7]+\Z/).nil?
   end
 
   def boolean?(value)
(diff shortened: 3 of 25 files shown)


.travis.yml
```diff
@@ -12,6 +12,8 @@ before_install:
 - gem update --system
 - gem --version
 - bundle --version
+env:
+  - SLOW=1
```
miah, line 16: I like this, and that the slow tests get marked as skipped so we don't forget about them =)

miah (approved):

.travis.yml
zenspider: Exactly. That was the intent.

If you know of a better way to globally set an env on appveyor, I'd appreciate it.

lib/inspec/plugin/v2/loader.rb
```diff
@@ -113,9 +113,8 @@ def plugin_gem_path
     end
 
     def self.plugin_gem_path
-      # I can't believe there isn't a simpler way of getting this
-      # 2.4.2.p123 => 2.4.0
-      ruby_abi_version = (Gem.ruby_version.segments[0, 2] << 0).join('.')
```
clintoncwolfe, line 118: great!

lib/matchers/matchers.rb
```diff
@@ -247,26 +247,26 @@ def try_match(actual, op, expected) # rubocop:disable Metrics/CyclomaticComplexi
     # if actual and expected are strings
     if expected.is_a?(String) && actual.is_a?(String)
       return actual.casecmp(expected) == 0 if op == :==
-      return Gem::Version.new(actual).method(op).call(Gem::Version.new(expected)) if
```
clintoncwolfe, line 249: Could you comment as to why these changes are needed to get the tests running more cleanly?

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -569,6 +569,8 @@ def test_refuse_install_when_plugin_on_exclusion_list
   end
 
   def test_error_install_with_debug_enabled
+    skip "this test requires bundler to pass" unless defined? ::Bundler
```
clintoncwolfe, line 572: Are there test conditions under which Bundler is not available? We require it to run the test suite...

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -581,11 +583,13 @@ def test_error_install_with_debug_enabled
 #-----------------------------------------------------------------------------------------#
 #                               inspec plugin update
 #-----------------------------------------------------------------------------------------#
-class PluginManagerCliUpdate < MiniTest::Test
+class PluginManagerCliUpdate < Minitest::Test
   include CorePluginFunctionalHelper
   include PluginManagerHelpers
 
   def test_when_a_plugin_can_be_updated
+    skip "this test requires bundler to pass" unless defined? ::Bundler
```
clintoncwolfe, line 591: Likewise here, we should have bundler...

test/integration/default/controls/audit_spec.rb
```diff
@@ -14,7 +14,8 @@
 
 # Grab bundled inspec version. This should be the same as the one
 # passed for audit cookbook. If its not, you should do a `bundle install`
-inspec_version = Gem.loaded_specs['inspec'].version.to_s
+inspec_version = Inspec::VERSION
+# or: Gem.loaded_specs['inspec'].version.to_s rescue Inspec::VERSION
```
clintoncwolfe, line 17: Could you explain this change?

test/unit/matchers/matchers_test.rb
```diff
@@ -0,0 +1,118 @@
...
+require "rspec/expectations"
+require "matchers/matchers"
+
+require "pry"
+
+describe 'inspec matchers' do
+  describe 'cmp matcher' do
+    include RSpec::Matchers
+
+    def rspec_expect value, &block
```
clintoncwolfe, line 17: This probably needs in-code comments to explain what is going on. We shouldn't assume maintainers have any particular knowledge of RSpec.

test/unit/matchers/matchers_test.rb
```diff
@@ -0,0 +1,118 @@
...
+      ::RSpec::Expectations::ExpectationTarget.for(value, block)
+    end
+
+    def assert_cmp exp, act
```
clintoncwolfe: I think `exp` means `expectation`, but I don't know what `act` refers to. Could you use more expressive variable names?

test/unit/plugin/v2/installer_test.rb
```diff
@@ -51,18 +58,19 @@ def teardown
       Dir.glob(File.join(@config_dir_path, 'empty', '*')).each do |path|
         next if path.end_with? '.gitkeep'
         FileUtils.rm_rf(path)
-      end
+      end if @config_dir_path
     end
 
     # Clean up any activated gems
     Gem.loaded_specs.delete('inspec-test-fixture')
+    Gem.loaded_specs.delete('ordinal_array')
```
clintoncwolfe, line 69: I think I know what this fixes, but just for completeness, could you comment explaining what this is for?

test/unit/plugin/v2/installer_test.rb
```diff
@@ -195,13 +209,18 @@ def test_install_a_pinned_gem_from_rubygems_org
   def test_install_a_gem_with_conflicting_depends_from_rubygems_org
     ENV['INSPEC_CONFIG_DIR'] = File.join(@config_dir_path, 'empty')
 
+    spec = Gem::Specification._all.find { |s| s.name == "rake" }
```
clintoncwolfe, line 216: Could you explain this?

clintoncwolfe (commented):
This looks generally good; however on the sporadic failure fixes, there are some that I am not clear on - adding some comments on the PR indicating what fixes what would be very helpful. I also had a clarity concern or two.

Those aren't really blockers; if you add the comments, I can flip to Approve.

test/unit/plugin/v2/installer_test.rb
zenspider: What are you looking for here beyond the comment above it? That some tests activate different dependencies and will wind up causing test ordering issues?

test/unit/matchers/matchers_test.rb
zenspider: "actual". I use this all over (esp in the doco for minitest) so I make assumptions about it's "obviousness". What I really want is a better spoken language that has same-length antonyms (true vs false, local vs remote, etc).

zenspider:
This diff is bigger than it needs to be because it was rebased over master and then my modernize_minitest branch. If we merge the latter, then this diff should clean up a fair amount.

miah:
Since we merged #4023 you might have to rebase this branch against master to clean that diff up.

lib/matchers/matchers.rb
zenspider: They're not. They're part of the general cleanup of stuff that I saw as I went through. The main reason this should go in is this:

```
2896 % ./cmp.rb 
Warming up --------------------------------------
         method+call   154.352k i/100ms
                send   248.925k i/100ms
Calculating -------------------------------------
         method+call      2.603M (± 1.4%) i/s -     13.120M in   5.041969s
                send      6.425M (± 3.1%) i/s -     32.111M in   5.003057s

Comparison:
                send:  6424944.9 i/s
         method+call:  2602626.5 i/s - 2.47x  slower
```

Especially considering that this is on the matcher side of things, and thus, hit by all users.

test/unit/matchers/matchers_test.rb
zenspider: fixed

test/unit/matchers/matchers_test.rb
zenspider: fixed

miah:
Lets rebase this now that #4066 has landed.

miah (approved):

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
zenspider: We don't require bundler to run the test suite. It runs fine without it with the exception of some of these test assumptions. 

Requiring bundler of our users and especially contributors makes total sense, but the developers on this project shouldn't be beholden to an unnecessary requirement like bundler, especially when it just slows us down.

```
4402 % time ntimes 10 bundle exec ~/.rbenv/versions/2.6.2/bin/ruby -Ilib:test -e 0

real	0m5.506s
user	0m4.504s
sys	0m0.883s
4403 % time ntimes 10 ~/.rbenv/versions/2.6.2/bin/ruby -Ilib:test -e 0

real	0m0.876s
user	0m0.694s
sys	0m0.146s
```

Some of these commits are effectively decoupling our tests from bundler.

test/integration/default/controls/audit_spec.rb
zenspider: The active code. That is using our VERSION constant instead of finding the same value indirectly all the way through rubygems and back. If the spec isn't loaded, then that will fail out. There's no reason that the spec needs to be loaded for this test to run or pass.

test/unit/plugin/v2/installer_test.rb
zenspider: Decouples from bundler.

miah:
This makes us go GREEN on Travis. I'm going to merge this to get us into a better place. @clintoncwolfe please review this still.

PR #4036, opened 2019-05-08
Add missing tests for DataSeries#enabled? #disable

miah:


test/unit/utils/telemetry/data_series_test.rb
```diff
@@ -45,4 +45,16 @@ def test_to_json
     assert_equal '{"name":"fizz","data":["foo"]}', ds.to_json
     assert JSON.parse(ds.to_json)
   end
+
+  def test_enabled
+    ds = Inspec::Telemetry::DataSeries.new('fizz')
+    assert ds.enabled?
```
zenspider, line 51: Eventually I want to move us off of bare `assert` as the messages aren't that good (and I'm not the biggest fan of always having our own string message as they're often very redundant). In this case:

```ruby
assert_operator ds, :enabled?
```

would fail with a message like "Expected [object] to be enabled?"

zenspider (approved):
Looks good. Thanks for the extra tests!

test/unit/utils/telemetry/data_series_test.rb
miah: Ignore the confused misclick emoji. Thanks for the tip =)

clintoncwolfe (approved):
Thanks, looks good!

PR #4037, opened 2019-05-08
Make the postgres resource more defensive

alexpop:
I've got the following exceptions from the postgres core resourse:

When scanning an ubuntu 18.04 server without postgres installed:
```
NoMethodError: undefined method `chomp' for nil:NilClass
  /Users/apop/git/inspec/lib/resources/postgres.rb:139:in `cluster_from_dir'
  /Users/apop/git/inspec/lib/resources/postgres.rb:43:in `determine_dirs'
  /Users/apop/git/inspec/lib/resources/postgres.rb:13:in `initialize'
  /Users/apop/git/inspec/lib/inspec/plugin/v1/plugin_types/resource.rb:96:in `initialize'
  /Users/apop/git/inspec/lib/inspec/resource.rb:51:in `new'
```

When scanning my osx laptop without postgres installed:
```
NoMethodError: undefined method `empty?' for nil:NilClass
  /Users/apop/git/inspec/lib/resources/postgres.rb:19:in `initialize'
  /Users/apop/git/inspec/lib/inspec/plugin/v1/plugin_types/resource.rb:96:in `initialize'
  /Users/apop/git/inspec/lib/inspec/resource.rb:51:in `new'
  /Users/apop/git/inspec/lib/inspec/resource.rb:51:in `block (3 levels) in create_dsl'
  ../postgres-baseline/controls/postgres_spec.rb:40:in `load_with_context'
```

The changes in this PR are increasing the defensiveness of the resource against unexpected setups.

### DIFF
--- a/lib/inspec/resources/postgres.rb
+++ b/lib/inspec/resources/postgres.rb
@@ -47,8 +47,8 @@ module Inspec::Resources
         @version = version_from_psql
         if @version.to_s.empty?
           if inspec.directory('/var/lib/pgsql/data').exist?
-            warn 'Unable to determine PostgreSQL version: psql did not return
-              a version number and unversioned data directories were found.'
+            warn 'Unable to determine PostgreSQL version: psql did not return' \
+                 'a version number and unversioned data directories were found.'
           else
             @version = version_from_dir('/var/lib/pgsql')
           end
@@ -131,7 +131,7 @@ module Inspec::Resources
         'main'
       else
         dirs = inspec.command("ls -d #{dir}/*/").stdout.lines
-        if dirs.count == 0
+        if dirs.empty?
           warn "No postgresql clusters configured or incorrect base dir #{dir}"
           return nil
         end
--- a/lib/inspec/resources/postgres.rb
+++ b/lib/inspec/resources/postgres.rb
@@ -15,7 +15,7 @@ module Inspec::Resources
       # print warnings if the dirs do not exist
       verify_dirs
 
-      if !@version.nil? && !@conf_dir.empty?
+      if !@version.to_s.empty? && !@conf_dir.to_s.empty?
         @conf_path = File.join @conf_dir, 'postgresql.conf'
       else
         @conf_path = nil
@@ -38,16 +38,17 @@ module Inspec::Resources
         # installed as well as multiple "clusters" to be configured.
         #
         @version = version_from_psql || version_from_dir('/etc/postgresql')
-        @cluster = cluster_from_dir("/etc/postgresql/#{@version}")
-        @conf_dir = "/etc/postgresql/#{@version}/#{@cluster}"
-        @data_dir = "/var/lib/postgresql/#{@version}/#{@cluster}"
+        if !@version.to_s.empty?
+          @cluster = cluster_from_dir("/etc/postgresql/#{@version}")
+          @conf_dir = "/etc/postgresql/#{@version}/#{@cluster}"
+          @data_dir = "/var/lib/postgresql/#{@version}/#{@cluster}"
+        end
       else
         @version = version_from_psql
-        if @version.nil?
+        if @version.to_s.empty?
           if inspec.directory('/var/lib/pgsql/data').exist?
             warn 'Unable to determine PostgreSQL version: psql did not return
-             a version number and unversioned data directories were found.'
-            nil
+              a version number and unversioned data directories were found.'
           else
             @version = version_from_dir('/var/lib/pgsql')
           end
@@ -130,6 +131,10 @@ module Inspec::Resources
         'main'
       else
         dirs = inspec.command("ls -d #{dir}/*/").stdout.lines
+        if dirs.count == 0
+          warn "No postgresql clusters configured or incorrect base dir #{dir}"
+          return nil
+        end
         first = dirs.first.chomp.split('/').last
         if dirs.count > 1
           warn "Multiple postgresql clusters configured or incorrect base dir #{dir}"


lib/resources/postgres.rb
```diff
@@ -16,7 +16,7 @@ def initialize
       # print warnings if the dirs do not exist
       verify_dirs
 
-      if !@version.nil? && !@conf_dir.empty?
+      if !@version.to_s.empty? && !@conf_dir.to_s.empty?
```
alexpop:
```ruby
[6] pry(main)> "test".empty?
=> false

[7] pry(main)> "".empty?
=> true

[8] pry(main)> nil.empty?
NoMethodError: undefined method `empty?' for nil:NilClass
from (pry):8:in `__pry__'

[9] pry(main)> nil.to_s.empty?
=> true

[10] pry(main)> "test".to_s.empty?
=> false

[11] pry(main)> "".to_s.empty?
=> true
```

lib/resources/postgres.rb
```diff
@@ -39,16 +39,17 @@ def determine_dirs
...
+        end
       else
         @version = version_from_psql
-        if @version.nil?
+        if @version.to_s.empty?
           if inspec.directory('/var/lib/pgsql/data').exist?
             warn 'Unable to determine PostgreSQL version: psql did not return
-             a version number and unversioned data directories were found.'
-            nil
+              a version number and unversioned data directories were found.'
```
miah: Could you wrap the string like this:

```
warn 'Unable to determine PostgreSQL version: psql did not return' \
     'a version number and unversioned data directories were found.'
```

miah (changes requested):
I think this is a big step in the right direction. Thanks @alexpop

lib/resources/postgres.rb
zenspider: Why is `@version.nil?` converted as well?

As far as the second one goes, I think it is better to use a safe-call to preserve meaning (tho, I'd rather we walk upstream of that nil and figure out how to defend against it at the source):

```
>> nil&.empty?
=> nil
>> ""&.empty?
=> true
>> "test"&.empty?
=> false
```

lib/resources/postgres.rb
```diff
@@ -39,16 +39,17 @@ def determine_dirs
...
-        @data_dir = "/var/lib/postgresql/#{@version}/#{@cluster}"
+        if !@version.to_s.empty?
+          @cluster = cluster_from_dir("/etc/postgresql/#{@version}")
+          @conf_dir = "/etc/postgresql/#{@version}/#{@cluster}"
+          @data_dir = "/var/lib/postgresql/#{@version}/#{@cluster}"
+        end
       else
         @version = version_from_psql
-        if @version.nil?
+        if @version.to_s.empty?
```
zenspider: Is the concern that it _could_ be an empty string instead of nil?

lib/resources/postgres.rb
```diff
@@ -131,6 +132,10 @@ def cluster_from_dir(dir)
         'main'
       else
         dirs = inspec.command("ls -d #{dir}/*/").stdout.lines
+        if dirs.count == 0
```
zenspider: `dirs.empty?` has clearer intent.

zenspider (changes requested):

lib/resources/postgres.rb
alexpop: 👍 changed

lib/resources/postgres.rb
alexpop: thank you, changed

lib/resources/postgres.rb
alexpop: Yes, current and future changes of version parsing could provide a nil or empty string. Wanted to guard against both while at it.

lib/resources/postgres.rb
alexpop: Gave the save operator a try but doesn't look to handle the nil as we need it for these conditions.
```ruby
$ pry
[1] pry(main)> @version
=> nil
[2] pry(main)> puts "UNDEFINED" if @version&.empty?
=> nil
[3] pry(main)> puts "UNDEFINED" if @version.to_s.empty?
UNDEFINED
=> nil
[4] pry(main)> @version=''
=> ""
[5] pry(main)> puts "UNDEFINED" if @version&.empty?
UNDEFINED
=> nil
[6] pry(main)> puts "UNDEFINED" if @version.to_s.empty?
UNDEFINED
=> nil
[7] pry(main)> @version='1.2.3'
=> "1.2.3"
[8] pry(main)> puts "UNDEFINED" if @version&.empty?
=> nil
[9] pry(main)> puts "UNDEFINED" if @version.to_s.empty?
=> nil
```

alexpop: Nils or empty strings can come from three methods
```
version_from_psql
version_from_dir
locate_data_dir_location_by_version
```
and possibly more in the future.

I wanted to guard parts of the code vulnerable to empty values to avoid the stacktraces I've got without trying hard.

lib/resources/postgres.rb
zenspider: Having the version **parsing** come back nil is not a sign of an _error_?

lib/resources/postgres.rb
zenspider: It's the covering it up and moving on part that I'm a bit worried about, but I'm not familiar with the tools being used yet.

Just a coding smell: silently hiding nil in succession is usually a bandaid on a bandaid.

zenspider:
I wish our testing setup was a bit more reasonable so we could require tests on this type of thing... I think I'm gonna merge this for now and we'll try to backfill later.

PR #4039, opened 2019-05-08
[WIP] Fix Appveyor Failures

clintoncwolfe:
PR to investigate and repair the current AppVeyor failures on master


miah:
Since we landed #4066 I don't think this PR is needed. We still want some vagrant/test-kitchen magic for our Windows tests but I don't think this is it.

PR #4041, opened 2019-05-08
Remove `# encoding: utf8` magic comments

miah:
This PR removes every occurrence of a `# encoding` magic comment. The only other change is a small change to `test/unit/profiles/profile_test.rb` due to changes in the profile being tested.


zenspider (approved):
Looks great

clintoncwolfe (approved):
Many thanks!

tas50:
I really wanted to do this when I first read through inspec. Looks good

PR #4042, opened 2019-05-08
Relocate core resources from lib/resources to lib/inspec/resources

miah:
This moves all of the core resources from `lib/resources` to `lib/inspec/resources`.

I've left the aws and azure resources in place as I think the next step for them is to remove them entirely and depend on their resource packs instead.


zenspider (approved):
Hah! Damn. That's quite the list. Assuming this passes on CI I think this is great.

clintoncwolfe (approved):
Yaaaaaassss!

miah:
Expect a few more like this over the next few weeks. This issue has been bothering me for a while =)

miah:
I fixed the broken tests here: https://github.com/inspec/inspec/pull/4046

PR #3988, opened 2019-04-26
Add Coveralls.io support to InSpec

miah:
The goal with this PR is to add 'coveralls' to our pipeline. This means all PR's would get checked for test coverage and coveralls would give us a inline report. This would be useful for contributors getting immediate feedback and help make sure we all do better with test coverage.

I also have a similar PR to add codeclimate which can also do test coverage (but I haven't enabled that feature as it requires changing travis/appveyor settings). codeclimate is mostly focused on code complexity and maintainability.

https://coveralls.io/github/inspec/inspec

https://github.com/inspec/inspec/pull/4044

Signed-off-by: Miah Johnson <miah@chia-pet.org>

PR #4044, opened 2019-05-08
Add codeclimate to our pipeline

miah:
The goal with this PR is to add 'codeclimate' to our pipeline. This means all PR's would get checked for complexity and maintainability. We can tune its reports through the included yaml file and I suspect that at first it might be brutal. =)

https://codeclimate.com/github/inspec/inspec

https://github.com/inspec/inspec/pull/3988

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,19 @@
+version: "2"
+checks:
+  file-lines:
+    enabled: false
+plugins:
+  fixme:
+    enabled: true
+    config:
+      strings:
+        - TODO
+  flog:
+    enabled: true
+  markdownlint:
+    enabled: true
+  rubocop:
+    enabled: true
+    channel: rubocop-0-49
+    config:
+      file: .rubocop.yml


.codeclimate.yml
```diff
@@ -0,0 +1,19 @@
+version: "2"
+checks:
+  file-lines:
+    enabled: false
+plugins:
+  fixme:
+    enabled: true
+    config:
+      strings:
+        - TODO
```
zenspider, line 10: I also use HACK, FIX, REFACTOR... ummm... 

But I also don't think I've added any either (yet)

.codeclimate.yml
```diff
@@ -0,0 +1,19 @@
...
+  flog:
+    enabled: true
```
zenspider, line 12: Flay? (I think it is called `duplication`)

That one is gonna be brutal as well. We might want to bump thresholds to start and slowly ramp them down over time.

.codeclimate.yml
```diff
@@ -0,0 +1,19 @@
+version: "2"
+checks:
+  file-lines:
+    enabled: false
```
zenspider, line 4: I honestly wish this wasn't disabled... I like to code on a fairly small laptop and it's been a bit of a pain.

.codeclimate.yml
zenspider: It might also be this:

```
  similar-code:
    enabled: true
  identical-code:
    enabled: true
```

zenspider (approved):
I think this is a good start. It's prolly gonna hurt. (yay!)

.codeclimate.yml
zenspider: Confirmed that it has been renamed: https://docs.codeclimate.com/docs/duplication-concept

.codeclimate.yml
miah: I did some quick greps and it seems like TODO was the only one present in our repo. I haven't looked at train, kitchen-inspec, etc so this list might grow/change soon. (I'll likely end up copying this config to all our repos in the future).

.codeclimate.yml
miah: Correct we should get those automatically as they are enabled by default. I'm only disabling the defaults and adding optional plugins in this configuration.

miah: Do you have any suggestions for a good starting threshold for flog?

.codeclimate.yml
miah: I'm in agreement. I'll gladly remove this entry, I know our code severely fails on this one though.

@clintoncwolfe whats your thought?

clintoncwolfe: Well, philosophically, I'm the "I want to build features and fix bugs on a product, not deal with linting issues constantly", so I don't care if the files are big, etc. We have editors, we have scrolling mice, we have search.

On new work, I'm all for short files, why not; but I don't see much gain from reducing file length in isolation from other refactorings (I'm all for splitting lib/inspec/cli.rb, for example, but it's not file-length that motivates me).

That is entirely an opinion, though.

miah: Lets leave it disabled for now =)

zenspider: Oh. Derp. I assumed it was about line width, not file length.

aaronlippold:
Can I also suggest a bit of static analysis as well.

On going static analysis That's part of the regular PR process would go a
long way in demonstrating a focus on security as well as function.

.codeclimate.yml
zenspider: One MILLION!

PR #4046, opened 2019-05-08
Fix broken tests after removing magic comments.

miah:


zenspider (approved):
Not on you—at all—but these tests are brittle and should be reworked to not be susceptible to this sort of change.

miah:
I agree, I've been looking at coverage/index.html and coveralls report a bunch now trying to think about how we can improve some of these. But right now I'm mostly concerned with resources.

PR #4066, opened 2019-05-15
Emergency test fixes!

zenspider:
This fixes our test framework contention and normalizes the number of tests we run regardless of platform/version.

This also cleans up a number of issues that made it harder to debug than it should have been. I'd like to push us towards more consistency, esp on the test side of things. We need to have them deterministic and absolutely dependable.

See commentary in individual commits.

### DIFF
--- a/examples/plugins/inspec-resource-lister/test/helper.rb
+++ b/examples/plugins/inspec-resource-lister/test/helper.rb
@@ -23,3 +23,4 @@ require 'minitest/spec'
 require 'minitest/autorun'
 
 # You might want to put some debugging tools here.  We run tests to find bugs, after all.
+# require 'byebug'
--- a/examples/plugins/inspec-resource-lister/test/helper.rb
+++ b/examples/plugins/inspec-resource-lister/test/helper.rb
@@ -23,4 +23,3 @@ require 'minitest/spec'
 require 'minitest/autorun'
 
 # You might want to put some debugging tools here.  We run tests to find bugs, after all.
-require 'byebug'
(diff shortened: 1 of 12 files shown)


examples/plugins/inspec-resource-lister/test/helper.rb
```diff
@@ -23,4 +23,3 @@
 require 'minitest/autorun'
 
 # You might want to put some debugging tools here.  We run tests to find bugs, after all.
-require 'byebug'
```
jerryaldrichiii, line 26: Maybe comment out this line?

I'm all for removing it, but if we do maybe the comment above too.

test/helper.rb
```diff
@@ -16,7 +23,58 @@
   add_group 'Backends', 'lib/inspec/backend'
 end
 
-require 'minitest/autorun'
+##
```
jerryaldrichiii, line 26: Wizardry (in my eyes) with supporting documentation. So so so good.

jerryaldrichiii (approved):
Looks good to me. Just some comments.

Thank you for this. Those 'byebug' lines have bothered me for quite some time.

examples/plugins/inspec-resource-lister/test/helper.rb
miah: I'm +1 on removing all byebug references, and the comment.

test/helper.rb
miah: Wow.

miah (approved):
Its 00:40. I'm reviewing code. This has been a journey. Thank you.

PR #4076, opened 2019-05-15
Add kitchen rake testing

lamont-granquist:
I'd suggest starting to structure kitchen testing like this with a
directory and subdirs to keep the kitchen testing gems out of the
root Gemfile entirely.

This still mounts the root dir in /inspec so the root Gemfile
is still what you're running the rake tests against.

By having an extra layer of subdirs, then you can split up different
use cases like the rake-testing vs. the audit cookbook testing.
Otherwise it'll be a mess of having to manage different kitchen.yml
files that require way too many different drivers/provisioners/verifiers

### DIFF
--- a/kitchen/rake-tests/kitchen.yml
+++ b/kitchen/rake-tests/kitchen.yml
@@ -1,0 +1,24 @@
+driver:
+  name: docker
+
+platforms:
+- name: ubuntu-18.04
+  driver_config:
+    image: ubuntu:18.04
+    platform: ubuntu
+    provision_command:
+      - curl -L https://omnitruck.chef.io/install.sh | sudo bash -s -- -P inspec
+      - apt-get -y install build-essential
+    require_chef_omnibus: false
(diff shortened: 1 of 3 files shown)


zenspider:
> `cd kitchen/rspec-tests && bundle install && bundle exec kitchen test`

zenspider (approved):
Thank you so much!

zenspider:
Merging so we can get our hands on it.

kitchen/rake-tests/kitchen.yml
```diff
@@ -0,0 +1,24 @@
...
+    require_chef_omnibus: false
+    volume: <%= File.expand_path("../..", File.dirname(__FILE__)) %>:/inspec
+
+provisioner:
+  name: shell
+
+verifier:
+  name: shell
+  remote_exec: true
+  command: cd /inspec && PATH=/opt/inspec/embedded/bin:$PATH bundle exec rake test
```
miah, line 21: Wow easier than I thought it would be!

PR #4077, opened 2019-05-15
 Build the InSpec gem directly in Kitchen so we can run locally 

tas50:
Before you had to kick off kitchen via Rake as the Rake task build the local gem we injected into the cookbook. Now we do it within Test Kitchen using a feature that didn't exist when this was all written. Also --output is your friend and greatly reduces the complexity of all this.

This also has some cleanup for the private chef kitchen file to nuke some old platforms.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -138,19 +138,7 @@ namespace :test do
     concurrency = ENV['CONCURRENCY'] || 1
     os = args[:os] || ENV['OS'] || ''
     ENV['DOCKER'] = 'true' if ENV['docker'].nil?
-    puts "Building current InSpec gem for audit cookbook testing..."
-    output = %x[gem build inspec-core.gemspec]
-    puts output
-    gem_name = output.split("\n")[-1].split(':')[1].strip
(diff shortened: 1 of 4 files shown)


zenspider (approved):
Thank you!

Rakefile
```diff
@@ -138,19 +138,7 @@ namespace :test do
...
-    path = File.dirname(__FILE__)
-    File.rename(File.join(path, gem_name), File.join(path, 'inspec-core-local.gem'))
-    destination = File.join(path, 'test', 'cookbooks', 'os_prepare', 'files', 'inspec-core-local.gem')
-    begin
-      FileUtils.cp(File.join(path, 'inspec-core-local.gem'), destination)
-      sh("bundle exec kitchen test -c #{concurrency} #{os}")
-    ensure
-      FileUtils.rm(destination)
-    end
+    sh("bundle exec kitchen test -c #{concurrency} #{os}")
```
miah, line 141: much nicer

miah (approved):

PR #4079, opened 2019-05-16
Fix @unique_controls uninitialized warning

miah:
## Description

memoize @unique_controls to a Set if nil and
check that its empty? to continue and add controls

/home/miah/projects/github/inspec-master/lib/inspec/reporters/cli.rb:169: warning: instance variable @unique_controls not initialized

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/reporters/cli.rb
+++ b/lib/inspec/reporters/cli.rb
@@ -166,9 +166,9 @@ module Inspec::Reporters
     end
 
     def all_unique_controls
-      return @unique_controls unless @unique_controls.nil?
+      @unique_controls ||= Set.new
+      return @unique_controls unless @unique_controls.empty?
 
-      @unique_controls = Set.new
       run_data[:profiles].each do |profile|
         profile[:controls].map { |control| @unique_controls.add(control) }
       end
--- a/lib/inspec/reporters/cli.rb
+++ b/lib/inspec/reporters/cli.rb
@@ -166,14 +166,11 @@ module Inspec::Reporters
     end
 
     def all_unique_controls
-      @unique_controls ||= Set.new
-      return @unique_controls unless @unique_controls.empty?
-
-      run_data[:profiles].each do |profile|
-        profile[:controls].map { |control| @unique_controls.add(control) }
-      end
-
-      @unique_controls
+      @unique_controls ||= begin
+                             run_data[:profiles].map { |profile|
+                               profile[:controls]
+                             }.uniq
+                           end
     end
 
     def profile_summary
--- a/test/unit/reporters/cli_test.rb
+++ b/test/unit/reporters/cli_test.rb
@@ -288,13 +288,6 @@ describe Inspec::Reporters::CLI do
     it 'return unique controls' do
       report.send(:all_unique_controls).count.must_equal 4
     end
-
-    it 'return unique controls cached' do
-      instance_variable_get(:@unique_controls).must_be_nil
-      report.send(:all_unique_controls).count.must_equal 4
-      assert = report.instance_variable_get(:@unique_controls)
-      assert.count.must_equal 4
-    end
   end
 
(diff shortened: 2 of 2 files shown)


test/unit/reporters/cli_test.rb
```diff
@@ -288,13 +288,6 @@ def windowize(string)
     it 'return unique controls' do
       report.send(:all_unique_controls).count.must_equal 4
     end
-
-    it 'return unique controls cached' do
-      instance_variable_get(:@unique_controls).must_be_nil
-      report.send(:all_unique_controls).count.must_equal 4
-      assert = report.instance_variable_get(:@unique_controls)
-      assert.count.must_equal 4
-    end
```
zenspider, line 297: great!

lib/inspec/reporters/cli.rb
```diff
@@ -166,9 +166,9 @@ def format_with_color(color_name, text)
...
 
     def all_unique_controls
-      return @unique_controls unless @unique_controls.nil?
+      @unique_controls ||= Set.new
+      return @unique_controls unless @unique_controls.empty?
 
-      @unique_controls = Set.new
       run_data[:profiles].each do |profile|
         profile[:controls].map { |control| @unique_controls.add(control) }
       end
```
zenspider: This one I'm gonna turn into an opportunity to evolve this. Getting rid of the warning is absolutely desired, but this method is a bit of a mess (not on you at all). But! Now we have the opportunity to clean this up!

1) original
```ruby
    def all_unique_controls
      return @unique_controls unless @unique_controls.nil?

      @unique_controls = Set.new
      run_data[:profiles].each do |profile|
        profile[:controls].map { |control| @unique_controls.add(control) }
      end

      @unique_controls
    end
```
2) don't access uninitialized ivar (yay!) This is definitely one way to do it. Another is to simply set it in `initialize`, but the point of this method is to be a "lazy accessor" so it doesn't compute the set until it is used.
```ruby
    def all_unique_controls
      @unique_controls ||= Set.new
      return @unique_controls unless @unique_controls.empty?

      run_data[:profiles].each do |profile|
        profile[:controls].map { |control| @unique_controls.add(control) }
      end

      @unique_controls
    end
```
3) New let's focus on the content. That map stands out as very odd since it isn't going anywhere. So don't map if you're not mapping (tho, we are, as we will see below)
```ruby
    def all_unique_controls
      @unique_controls ||= Set.new
      return @unique_controls unless @unique_controls.empty?

      run_data[:profiles].each do |profile|
        profile[:controls].each do |control|
          @unique_controls.add(control)
        end
      end

      @unique_controls
    end
```
4) Use idiomatic caching method structure: `@x ||= begin ... end`
```ruby
    def all_unique_controls
      @unique_controls ||= begin
                             uc = Set.new

                             run_data[:profiles].each do |profile|
                               profile[:controls].each do |control|
                                 uc.add(control)
                               end
                             end

                             uc
                           end
    end
```
5) Finally, now that we can see what this is: a lazy accessor that is trying to return a unique collection of profile controls, let's strip it back to the essentials -- all this is really doing is getting the profile's controls (read: a map) and returning a unique collection (read: .uniq) of them, so just do that:
```ruby
    def all_unique_controls
      @unique_controls ||= begin
                             run_data[:profiles].map { |profile|
                               profile[:controls]
                             }.uniq
                           end
    end
```

zenspider (changes requested):
Your code is totally acceptable as is, but I want to push this further.

Thank you!

lib/inspec/reporters/cli.rb
miah: Perfect feedback. Thank you!

zenspider (approved):
Fantastic. I'm gonna push this through now so that CI is a bit more readable on your rake overhaul branch (please rebase)

PR #4080, opened 2019-05-16
Do not lint test kitchen support files

clintoncwolfe:
Avoid CI failures in code installed by bundler, as seen [here](https://ci.appveyor.com/project/chef/inspec/builds/24577117/job/yib5uplg00fph44h).


miah (commented):
Can we just `rubocop -a` and fix the files rather than ignore them?

clintoncwolfe:
From what I could tell, the files in question were being installed by bundler, and were not part of our codebase - hence the exclusion, rather than a correction. If #4082 works, I'd prefer it, as well.

PR #4081, opened 2019-05-16
Fix @profile_name not initialized

miah:
I believe that if `@conf['profile']` doesn't exist @profile_name may not be initialized.

```
inspec-master/lib/inspec/profile_context.rb:35: warning: instance variable @profile_name not initialized
```

### DIFF
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -25,7 +25,7 @@ module Inspec
       @profile_id = profile_id
       @backend = backend
       @conf = conf.dup
-      @profile_name = @conf['profile'].profile_name || @profile_id if @conf['profile']
+      @profile_name = @conf.has_key?('profile') ? @conf['profile'].profile_name : @profile_id
       @skip_only_if_eval = @conf['check_mode']
       @rules = {}
       @control_subcontexts = []


lib/inspec/profile_context.rb
```diff
@@ -25,7 +25,7 @@ def initialize(profile_id, backend, conf)
       @profile_id = profile_id
       @backend = backend
       @conf = conf.dup
-      @profile_name = @conf['profile'].profile_name || @profile_id if @conf['profile']
+      @profile_name = @conf.has_key?('profile') ? @conf['profile'].profile_name : @profile_id
```
zenspider, line 28: I think I'm a bit old school in saying this, but I think I prefer:
```ruby
@profile_name = @conf['profile'] ? @conf['profile'].profile_name : @profile_id
```
over your version... literally just pulling that `if` condition down to a ternary. But yours is actually a bit more polite and correct than mine.

zenspider (approved):
good. thank you!

lib/inspec/profile_context.rb
zenspider: This took me a while because of lack of coffee, but this is basically `(a = b) if c` but really should be `a = (b if c)`.

miah: Right, that || if was throwing me too. I'll update. Thanks!

PR #4086, opened 2019-05-16
Accept the Chef License during automated testing

tyler-ball:
## Description
@tas50 This should fix the errors you saw in https://travis-ci.org/inspec/inspec/jobs/533031727 while testing with the latest versions of our libraries

## Related Issue
N/A

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ before_install:
 - gem update --system
 - gem --version
 - bundle --version
+env:
+  global:
+    - CHEF_LICENSE=accept-no-persist
 matrix:
   include:
   - rvm: 2.4.5
--- a/Rakefile
+++ b/Rakefile
@@ -76,26 +76,6 @@ namespace :test do
     end or fail 'Failures'
   end
 
-  task :accept_license do
-    FileUtils.mkdir_p(File.join(Dir.home, '.chef', 'accepted_licenses'))
-    # If the user has not accepted the license, touch the acceptance
-    # file, but also touch a marker that it is only for testing.
-    unless File.exist?(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec'))
-      puts "\n\nTemporarily accepting Chef user license for the duration of testing...\n"
-      FileUtils.touch(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec'))
-      FileUtils.touch(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec.for_testing'))
-    end
-
-    # Regardless of what happens, when this process exits, check for cleanup.
-    at_exit do
-      if File.exist?(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec.for_testing'))
-        puts "\n\nRemoving temporary Chef user license acceptance file that was placed for test duration.\n"
-        FileUtils.rm_f(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec'))
-        FileUtils.rm_f(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec.for_testing'))
-      end
-    end
-  end
-
   Rake::TestTask.new(:functional) do |t|
     t.libs << 'test'
     t.test_files = Dir.glob([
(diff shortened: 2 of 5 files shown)


miah (approved):
Thanks @tyler-ball

Rakefile
```diff
@@ -76,26 +76,6 @@ namespace :test do
...
-
-    # Regardless of what happens, when this process exits, check for cleanup.
-    at_exit do
-      if File.exist?(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec.for_testing'))
-        puts "\n\nRemoving temporary Chef user license acceptance file that was placed for test duration.\n"
-        FileUtils.rm_f(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec'))
-        FileUtils.rm_f(File.join(Dir.home, '.chef', 'accepted_licenses', 'inspec.for_testing'))
-      end
-    end
-  end
```
zenspider, line 97: Are you intending for devs to have just accepted the license and use our home folders? The ENV above is only for travis, no?

.travis.yml
```diff
@@ -12,6 +12,9 @@ before_install:
 - gem update --system
 - gem --version
 - bundle --version
+env:
+  global:
+    - CHEF_LICENSE=accept-no-persist
```
zenspider, line 17: Probably needs to go into appveyor as well.

zenspider (changes requested):
Not sure about regular rake / ruby runs.

.travis.yml
tyler-ball: Thanks, incoming

Rakefile
tyler-ball: Correct. Thats the pattern I have used so far. It does require accepting the license in all the CI systems (Travis, Appveyor, Buildkite, etc...)

I think this is somewhat of a philosophical question. We could certainly set `CHEF_LICENSE` in the `spec_helper.rb` (or somewhere else top level that will apply to all the tests). That would ensure the tests pass in all the CI systems but would also be accepting the license if someone ran the InSpec tests locally (which is approved in the EULA).

As I type all that out... maybe it is better to do it in the test setup/teardown rather than the CI system. Especially if we set `accept-no-persist` it would only accept it for that test run. Thoughts @zenspider @tas50 ?

zenspider: I think we're gonna have to do something more resilient because a number of our tests (and it should be more) set the HOME to a fake/mock directory.

I don't have an opinion about going the env route vs setup/teardown. The former is probably easier tho.

tyler-ball:
It _looks_ like this fixed the functional tests. Perhaps the unit tests are failing because of updated dependencies from TK 1.x -> 2.x?

tas50:
I'm concerned that we're working around a bug in TK and kitchen-dokken by specifying the env vars. They shouldn't be needed. We were installing Chef 14 yet we were getting prompted for the license. Users are going to run into that same issue and we need to figure out what's causing it vs. setting the env var to work around it.

Rakefile
zenspider: Also, while I'm not fond of the code you're removing above, I'm a big proponent that anyone should be able to check out the code, install deps, run rake, and see green. This violates that idea... but I'm not sure what the right balance is.

zenspider (changes requested):

tyler-ball:
Related: https://github.com/someara/kitchen-dokken/pull/177

kekaichinose:
This needs a meeting. @clintoncwolfe will gather InSpec team, @tas50, and @tyler-ball to come to a decision on how to test this during tomorrow's Engineering Foundation CR meeting.

clintoncwolfe:
After discussing in the eng-foundation code review, we decided this PR is really about two things:
 1. Moving the license acceptance to Travis instead of Rake
 2. Using ENV instead of filesystem touches

For 1, we agreed that we really don't want that - we want to keep it in Rake, so that contributors can run the tests using Rake without any surprises.

For 2, it's a refactoring request but there are some wrinkles around a test needing to turn off the licensing (more global state...).  It's legit, and this code was written in a hurry; I've moved the refactor request into issue #4200 to capture it.

Based on these conclusions, and the merge conflicts, I'm closing this PR.  Thanks for all the thought-provoking conversation!

PR #4089, opened 2019-05-17
Fix github url regex. Add period for repo

mirskiy:
For url transforming, the github url regex didn't allow periods in the repo name. Github allows periods in the repo name.

## Description

I allowed periods in the repo in the github url regex.

## Types of changes

## Checklist:

Thanks!

### DIFF
--- a/lib/fetchers/url.rb
+++ b/lib/fetchers/url.rb
@@ -65,8 +65,8 @@ module Fetchers
     # https://bitbucket.org/username/repo/commits/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc is transformed to
     # https://bitbucket.org/username/repo/get/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc.tar.gz
 
-    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w-]+)(\.git)?(/)?$}
-    GITHUB_URL_WITH_TREE_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w-]+)/tree/(?<commit>[\w\.]+)(/)?$}
+    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w.-]+?)(\.git)?(/)?$}
+    GITHUB_URL_WITH_TREE_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w.-]+)/tree/(?<commit>[\w\.]+)(/)?$}
     BITBUCKET_URL_REGEX = %r{^https?://(www\.)?bitbucket\.org/(?<user>[\w-]+)/(?<repo>[\w-]+)(\.git)?(/)?$}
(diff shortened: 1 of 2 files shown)


lib/fetchers/url.rb
```diff
@@ -65,7 +65,7 @@ def self.resolve_from_string(target, opts, username = nil, password = nil)
     # https://bitbucket.org/username/repo/commits/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc is transformed to
     # https://bitbucket.org/username/repo/get/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc.tar.gz
 
-    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w-]+)(\.git)?(/)?$}
+    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w.-]+)(\.git)?(/)?$}
     GITHUB_URL_WITH_TREE_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w-]+)/tree/(?<commit>[\w\.]+)(/)?$}
```
miah: I think we'd need it here too.

miah (changes requested):
Thanks @mirskiy. Could you also add it to the tree regex?

lib/fetchers/url.rb
zenspider: agreed. good catch.

lib/fetchers/url.rb
```diff
@@ -65,7 +65,7 @@ def self.resolve_from_string(target, opts, username = nil, password = nil)
     # https://bitbucket.org/username/repo/commits/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc is transformed to
     # https://bitbucket.org/username/repo/get/95ce1f83d5bbe9eec34c5973f6894617e8d6d8cc.tar.gz
 
-    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w-]+)(\.git)?(/)?$}
+    GITHUB_URL_REGEX = %r{^https?://(www\.)?github\.com/(?<user>[\w-]+)/(?<repo>[\w.-]+)(\.git)?(/)?$}
```
zenspider: good catch! I totally have repos with periods in their names!

zenspider (changes requested):

lib/fetchers/url.rb
mirskiy: Added to GITHUB_URL_WITH_TREE_REGEX.

mirskiy:
I updated the tree regex. I also changed the github url regex to be non-greedy for the repo - because we allow periods, it was capturing the ".git" in some of the test cases as part of the repo name (ex. github.com/inspec/inspec.git had repo as inspec.git). Those test cases are now passing.

I rebased against master and it looks like all the test cases are passing when I run locally, but I don't understand AppVeyor and Coveralls so let me know what I need to do to pass those.

miah (approved):
Thanks @mirskiy, dont worry about appveyor or coveralls at the moment. We're still straightening some things out with them =)

zenspider (approved):
Great. Thank you!

mirskiy:
@miah Is there anything else needed to merge this? Thanks!

PR #4107, opened 2019-05-18
Minor require cleanup for telemetry.

zenspider:
Changes `require_relative` to plain require and adds a missing require needed if global_methods_test was run by itself.


lib/inspec/utils/telemetry.rb
```diff
@@ -1,2 +1,3 @@
 require 'inspec/utils/telemetry/collector'
 require 'inspec/utils/telemetry/data_series'
+require 'inspec/utils/telemetry/global_methods'
```
miah, line 3: Will we have a circular require?

https://github.com/inspec/inspec/pull/4083

PR #4111, opened 2019-05-18
Fixes a test interaction bug between installer and loader.

zenspider:
Not sure how it was supposed to work in the first place. The plugin in
question depends on ordinal_array and that wasn't in that fixture dir.
Switching to 2 is all it apparently needed.

This does cause yet another double-load warning (VERSION already
defined). I don't know how we're supposed to get around that with the
way we've set up these tests.

## Types of changes

## Checklist:


miah (approved):

zenspider:
Merged early because it is destabilizing the other PRs

PR #4113, opened 2019-05-18
Just use math.

zenspider:
I don't know what the reasoning is behind the rjust & the format
call... String converting between octal to binary in order to get an
integer? Instead, this converts from an octal string and uses
straight-forward binary manipulation and logic.

Please avoid `cond ? false : true` like constructs. Use `!cond` instead.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/file.rb
+++ b/lib/inspec/resources/file.rb
@@ -155,10 +155,10 @@ module Inspec::Resources
       # to or less permissive than the desired mode (PASS). Otherwise, the files
       # mode is more permissive than the desired mode (FAIL).
 
-      max_mode = max_mode.rjust(4, '0')
-      binary_desired_mode = format('%04b', max_mode).to_i(2)
-      desired_mode_inverse = (binary_desired_mode ^ 0b111111111)
-      (desired_mode_inverse & file.mode).zero? ? false : true
+      max_mode = max_mode.to_i(8)
+      inv_mode = 0777 ^ max_mode
+
+      inv_mode & file.mode != 0
     end
 
     def to_s


miah (approved):

aaronlippold:
The rjust call was so that someone could pass in 0644 or 644 and either would work. This avoided the problem that uses were having with file mode in and above property when they would pass in 644 and get a failure 06444 and get a pass.

aaronlippold:
The conversion to straight binary strings allows us to ensure that we test everything at and below the upper bound and conversing for a lower bound.

aaronlippold (commented):
The formatting was there to ensure if I passed in three digit or four digit variance for mode or even five digit, 00644, I always got the same binary string calculations back. If you test these cases with the updated code and the math still comes out the same from a binary perspective I have no problem with simplifying the code. I just wanted to ensure that no matter what style of integer string for mode that the user might use, that the function still worked correctly. noting that most compliance documentation says things like ensure the files are mode 644 or less permissive, but a lot of our current code requires the user to use 0644 I wanted to avoid that confusion for the user.

The original implementation was derived from tested c code used in scap and each step is outlined in the above comments with the math behind them. Again, if you found a simpler solution I'm all for it :-)

zenspider:
Hey @aaronlippold. Thank you for the explanation! I'm still not sure I'm tracking (like the `"%04b"` part for a 9 bit binary?)... but I think the maths (specifically the string->int conversion) should be fine. Here are your cases from above:

```ruby
>> '0644'.to_i 8
=> 420
>> '644'.to_i 8
=> 420
>> '00644'.to_i 8
=> 420
```

There's a ton of `ArgumentError` checks above that prevent many other cases from occuring, so unless I'm still missing something, I think the conversion should be equivalent.

aaronlippold:
Cool, great no worries. I think I played around a lot with different implementations and I thought the way I did it was the safest no matter what the heck that user put in. And maybe my brain just wasn't working well that day and I was overthinking things. Just double check that when you put in the strings 00644, 0644 and 644 etc that we protect the user from both themselves and the silly compliance document writers. I always make the assumption that the user would have no idea of the difference of those three. The biggest problem this addresses is that when the user would write things like " mode less than or equal to 0644" with a straight file mode property that they would get a pass if it was 644 think the less than case was being covered. This is technically a false pass. I actually wanted to put a note from the documentation by the mode property so that user understood that the mode property itself was only an equals check.

aaronlippold:
And yes I put all those argument catches in there to make sure they couldn't shoot themselves in the foot

aaronlippold:
I think I attempted to do it your way the first time actually and I must have fat fingered something, so I'm glad you were able to do it better :-)

aaronlippold:
I think one of the other cases I was trying to cover I just recalled was
what if they passed in the number 0644 or 644 versus the string 0644 or
644. That might have been why I was doing the formatting stuff. So we want
to double-check that as well. AKA they didn't quote it.

aaronlippold:
This is where I wish Ruby had type casting

miah (approved):

PR #4116, opened 2019-05-19
Clean more requires

zenspider:

### DIFF
--- a/lib/inspec/expect.rb
+++ b/lib/inspec/expect.rb
@@ -24,7 +24,7 @@ module Inspec
     def example_group
       that = self
 
-      opts = { 'caller' => calls[0][3] }
+      opts = { 'caller' => calls[0][3] } # TODO: this needs overhaul. no magic #s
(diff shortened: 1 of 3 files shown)


miah (approved):

lib/inspec/expect.rb
```diff
@@ -24,7 +24,7 @@ def not_to(*args, &block)
     def example_group
       that = self
 
-      opts = { 'caller' => calls[0][3] }
+      opts = { 'caller' => calls[0][3] } # FIX: this needs overhaul. no magic #s
```
miah: We should use `# TODO` so codeclimate will complain =)

test/helper.rb
```diff
@@ -101,10 +101,9 @@ module Kernel
 require 'inspec/config'
 require 'fetchers/mock'
 require 'inspec/dependencies/cache'
```
miah, line 103: I doubt the majority of these requires are required in this file.

PR #4132, opened 2019-05-23
Changed version check network timeout from 1 to 3 seconds.

zenspider:
This fixes one problem in our tests where a slow/bad network connection causes our version checker to come back nil, causing rubygems to output a warning, causing our functional to fail on non-empty stderr checks.


miah (approved):

.gitignore
```diff
@@ -1,42 +1,40 @@
-Gemfile.lock
-Gemfile.local
-coverage
 *.gem
+.attribute.yml
```
clintoncwolfe, line 2: These changes seem unrelated to the purpose of the PR. Could you clean it up a bit?

lib/utils/latest_version.rb
```diff
@@ -1,19 +1,13 @@
...
   def latest
     uri = URI('https://rubygems.org/api/v1/gems/inspec.json')
-    res = Net::HTTP.start(
-      uri.host, uri.port, use_ssl: uri.scheme == 'https',
-      open_timeout: 0.5, read_timeout: 0.5
-    ) { |http|
-      http.get(uri.path)
-    }
-    inspec_info = JSON.parse(res.body)
+    inspec_info = JSON.parse(uri.read)
```
clintoncwolfe: Because InSpec is often executed in airgapped environments, we need to fail fast when there is no hope of a network connection - hence the specification of timeouts. I think we need to retain that.

clintoncwolfe (changes requested):
Looks like you caught a lot on your gitignore, but the main thing is we really do need some kind of fast timeouts on the version lookup. We don't handle airgapping consistently, but we don't want to regress.

lib/utils/latest_version.rb
miah: I wonder if we should just rip this functionality out entirely. I'd rather we have a command that checks if we're outdated and version just reporting the version; quickly.

Considering this feature is based around the version of the Gem, and we really want people to use chefdk / omnibus versions. If its a feature to be left in, we should be querying the latest omnibus version.

`inspec outdated` (probably pick a better name) - reports whether the user has a outdated version of inspec, what the latest version of inspec is, and the url to download it.

clintoncwolfe (approved):
I think the code is fine here, but please update the PR description (which will go in the Changelog) to reflect the fact that the version check timeout is changing from 1 to 3 seconds - that may impact some customer somewhere.

lib/utils/latest_version.rb
clintoncwolfe: I could see `inspec outdated`, as you suggest, but that would mean breaking change in `inspec version`'s behavior. What about adding `inspec version --no-check`?

PR #4136, opened 2019-05-24
Dist branding for some of plugins

artem-sidorenko:
- plugins/inspec-plugin-manager-cli
- plugins/inspec-artifact
- plugins/inspec-habitat
- plugins/inspec-compliance

## Description
Branding for community distributions

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-compliance/lib/inspec-compliance/api.rb
+++ b/lib/plugins/inspec-compliance/lib/inspec-compliance/api.rb
@@ -2,6 +2,7 @@
 require 'net/http'
 require 'uri'
 require 'json'
+require 'inspec/dist'
 
 require_relative 'api/login'
 require_relative 'configuration'
@@ -16,6 +17,7 @@ module InspecPlugins
     # API Implementation does not hold any state by itself,
     # everything will be stored in local Configuration store
     class API
+      include Inspec::Dist
       extend InspecPlugins::Compliance::API::Login
 
       # return all compliance profiles available for the user
@@ -293,7 +295,7 @@ module InspecPlugins
         if response.code == '400'
           Inspec::Log.debug(
             "Received 400 from #{url}#{automate_endpoint} - " \
-            'assuming target is a Chef Automate2 instance',
+            "assuming target is a #{AUTOMATE_PRODUCT_NAME}2 instance",
           )
           true
         else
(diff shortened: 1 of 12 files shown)


miah (approved):
Seems reasonable. Thanks!

lib/plugins/inspec-compliance/lib/inspec-compliance/api.rb
```diff
@@ -293,7 +294,7 @@ def self.target_is_automate2_server?(url, insecure)
         if response.code == '400'
           Inspec::Log.debug(
             "Received 400 from #{url}#{automate_endpoint} - " \
-            'assuming target is a Chef Automate2 instance',
+            "assuming target is a #{Inspec::Dist::AUTOMATE_PRODUCT_NAME}2 instance",
```
zenspider: Should this be its own constant just in case we decide to rename something yet again? I don't know how far out that is...

lib/plugins/inspec-compliance/lib/inspec-compliance/cli.rb
```diff
@@ -1,15 +1,17 @@
 
+require 'inspec/dist'
+
 require_relative 'api'
 
 module InspecPlugins
   module Compliance
     class CLI < Inspec.plugin(2, :cli_command)
-      subcommand_desc 'compliance SUBCOMMAND', 'Chef Compliance commands'
+      subcommand_desc 'compliance SUBCOMMAND', "#{Inspec::Dist::COMPLIANCE_PRODUCT_NAME} commands"
```
zenspider: These start to feel heavy. You can `include Inspec::Dist` in this class (and others) and then drop the namespacing to make it a bit more readable.

zenspider (approved):
One Q, One minor suggestion, but not blockers. Approved.

lib/plugins/inspec-compliance/lib/inspec-compliance/api.rb
miah: We just renamed everything, but the community distribution will have to choose something other than automate. At the very least we can remove the '2' here since the product is named 'Chef Automate'.

lib/plugins/inspec-compliance/lib/inspec-compliance/cli.rb
miah: @artem-sidorenko Are you comfortable making these changes? I'd love to get this all merged. Thanks =)

clintoncwolfe (approved):

lib/plugins/inspec-compliance/lib/inspec-compliance/api.rb
artem-sidorenko: I saw different methods for API endpoint type discovery within this context. Without to know all details, I saw it reasonable to leave the `2` as otherwise this debug message can't be differentiated from the message of `target_is_automate_server` below

artem-sidorenko:
@zenspider @clintoncwolfe @miah thanks for the review! I addressed the point with dropping of namespace for all classes

Issue #4133, opened 2019-05-23
Revamp resource (and others) registry 


zenspider:
Right now we load ALL resources including AWS on every single invocation of anything (including just loading test/helper.rb). This requires every file and their dependencies and costs a ton.

I'm part way through revamping this, but I'm starting to get it over my head. Unit tests are fine (and 33% faster!), but the functionals do a lot of installs and those things aren't found. I don't know the functional test (or maybe it's all `inspec` invocations) bootstrap sequencing enough to know how the plugins should be discovered (or when) and as such, many of them are all failing out.

zenspider:
Right now, all files are loaded. They hit our DSL for resources and all call `name`. Name adds the resource to a registry with the `name` string as the key and an anonymous class as the value.

THEN, (sometime after) all the keys are enumerated and methods are added (again, to an anonymous class) every time something is done. This is incredibly costly in many ways and can be greatly simplified by loading what we need when we need it. I just don't know how to do that last part for plugins/installs.

zenspider:
+ master:
  + 218 test files in 159s
  + require "inspec/base_cli" = true in 1.7849
+ clean_unnecessary_requires:
  + 218 test files in 98s
  + require "inspec/utils/telemetry" = true in 0.4844

zenspider:
#4365 addresses the rest of this for resources.
I just filed #4460 to for plugins.

Issue #4127, opened 2019-05-22
Things should have names


zenspider:
Inspec uses `Class.new` and `Module.new` with wild abandon. This must stop. It's entirely too difficult to debug any of the internals with all the clever code going on and I don't see a single use-case where these dynamic modules or classes are actually necessary. Use accessors instead of closures.

Aha! Link: https://chef.aha.io/features/SH-2322

clintoncwolfe:
Adding search keyword "Anonymous class"

clintoncwolfe:
I'm going to close this as a duplicate of #4211

PR #4142, opened 2019-05-25
Clean unnecessary requires

zenspider:
Partially dealing with #4133. This removes a bunch of very heavy requires from test/helper.rb and changes the way that resources are included at the top level. Now resources and tests can require what they directly depend on and not bring in everything all in one very expensive go.

+ master:
  + 218 tests in 159s
  + require "inspec/base_cli" = true in 1.7849
+ clean_unnecessary_requires:
  + 218 tests in 98s
  + require "inspec/utils/telemetry" = true in 0.4844


zenspider:
I still want to change the way the registry works entirely, essentially getting rid of `lib/inspec/resources.rb` as a whole and making it so that everything works on demand... but it is way too hard to debug right now and I'd rather deal with all the anonymous classes / modules first (ie #4127)

lib/inspec/resources/filesystem.rb
```diff
@@ -1,3 +1,5 @@
+require 'inspec/resources/command'
+
 module Inspec::Resources
   class FileSystemResource < Inspec.resource(1)
```
miah, line 4: Can we rename this one to `FileSystem` ?

miah (approved):

lib/inspec/resources/bash.rb
```diff
@@ -3,7 +3,7 @@
 require 'inspec/resources/command'
 
 module Inspec::Resources
-  class Bash < Cmd
+  class Bash < Command
```
clintoncwolfe: I agree that this is nicer, but I think things like this should be done on separate PRs in the future - this is massive PR, and it's important to keep focus. This change isn't actually needed to help improve require-isolation.

lib/inspec/resources/command.rb
```diff
@@ -84,4 +84,6 @@ def to_s
       output
     end
   end
+
+  Cmd = Command
```
clintoncwolfe: This sort of aliasing *definitely* makes me think a resource class rename sweep should happen on a separate PR.

lib/inspec/shell.rb
```diff
@@ -119,7 +119,7 @@ def help(topic = nil)
         end
       elsif topic == 'matchers'
         print_matchers_help
-      elsif !Inspec::Resource.registry[topic].nil?
+      elsif !Inspec::Resource.registry[topic].nil? # TODO: fix
```
clintoncwolfe: fix what? If you are leaving a note for a future developer (even you), please be kind and leave enough context to make it obvious what is problematic enough to make you leave a TODO. Thanks!

test/functional/helper.rb
```diff
@@ -8,6 +8,8 @@
 require 'yaml'
 require 'tmpdir'
 
+require 'inspec/resources' # TODO: work this back out
```
clintoncwolfe: I'm actually surprised this is needed here at all - test/functional/helper is mostly worried about spawning subprocesses, not the actual internals of the InSpec resources that the subprocesses may be exercising. What needs the resources here?

test/unit/resources/aws_billing_report_test.rb
```diff
@@ -1,5 +1,11 @@
 require 'helper'
-require_relative 'aws_billing_backend'
+require 'inspec/resource'
+require 'resources/aws/aws_billing_report'
+
+require_relative 'aws_billing_backend' # TODO: move this to mocks
```
clintoncwolfe, line 5: Note - as the AWS resources are being moved to the inspec-resource pack, development has stopped on these resources - and the resource pack has a different test harness, IIRC. So, we don't need to invest too much here.

clintoncwolfe (changes requested):
I think this is s great change, and will really open up some new features like autoloading. My major concern is that some resource classes are renamed on the PR (which is an unrelated change), while others are marked as a TODO to be renamed, and another is commented.  Likewise, xinetd file was renamed to match its class. I'd prefer that this PR just be about the `require` changes, and make the renames on a separate PR.  If that seems onerous, let's at least be all-or-nothing, and do all renames.

test/functional/helper.rb
zenspider: `test/functional/inspec_report_test.rb` for sure. Had a few other failures but it is starting to look like they're unstable tests instead.

lib/inspec/resources/filesystem.rb
zenspider: Totally... but I agree with @clintoncwolfe that we should do renames in a separate PR. I have a stash somewhere that'll report all the renames needed on first load so it should be quick and easy.

lib/inspec/resources/bash.rb
zenspider: I did this one because I was tired and getting confused by cmd the windows shell vs Command, esp once it started being subclassed by wholly different shells.

lib/inspec/shell.rb
zenspider: I will always know what this means. Unnecessary logic is unnecessary. I just didn't want to fix it in this PR. I have a slew of these I'd like to do soonish.

test/functional/helper.rb
zenspider:
```
Run options: --seed 8843

# Running:

SE

Finished in 0.146366s, 13.6644 runs/s, 0.0000 assertions/s.

  1) Error:
inspec report tests::report#test_0001_loads a json report:
NoMethodError: undefined method `file' for #<#<Class:0x00007fb64ccbf5e8>:0x00007fb64ccbead0>
Did you mean?  fail
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/dsl.rb:49:in `method_missing'
    /Users/ryan/Work/git/inspec/inspec/test/unit/mock/profiles/old-examples/profile/controls/example.rb:21:in `load_with_context'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile_context.rb:162:in `instance_eval'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile_context.rb:162:in `load_with_context'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile_context.rb:146:in `load_control_file'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile.rb:202:in `block in collect_tests'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile.rb:199:in `each'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/profile.rb:199:in `collect_tests'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/runner.rb:111:in `block in load'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/runner.rb:94:in `each'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/runner.rb:94:in `load'
    /Users/ryan/Work/git/inspec/inspec/lib/inspec/runner.rb:122:in `run'
    test/functional/inspec_report_test.rb:12:in `block (3 levels) in <main>'

2 runs, 0 assertions, 0 failures, 1 errors, 1 skips

You have skipped tests. Run with --verbose for details.
        1.65 real         0.94 user         0.49 sys
```

zenspider: I can push the require down since it seems to just be this one

zenspider:
Renames coming soon...

PR #4151, opened 2019-05-28
Nothing uses source_readers, remove it.

miah:
I believe this is all old functionality replaced by `lib/inspec/plugin/v1/plugin_types/source_reader.rb`


zenspider (approved):
Fantastic! Be sure to join #salmon in slack!

clintoncwolfe (approved):

miah:
After further investigation, this is still used. I'll file a issue with more details.

PR #4154, opened 2019-05-29
Fix missing Constant via module inclusion in CLI class

miah:
This is related to a change from earlier: https://github.com/inspec/inspec/pull/4136


zenspider (approved):
The include above isn't needed after this addition. Also, not sure how you want to do this but I had to do this exact fix to rebase my requires PR. I'll make sure it's its own commit, but they'll be different.

PR #4156, opened 2019-05-29
Allow to switch off the license check for community distributions

artem-sidorenko:
lets have a similar approach as in chef/chef: https://github.com/chef/chef/pull/8354/files#diff-847fc1c7c0af04a65c570c6f4c3381e9R25

## Description

In order to be able to pass the `enforce_license` flag from the binstub, the EULA checking code had to be moved to some Thor-related method within early initialisation phase of application.

## Types of changes

## Checklist:


lib/inspec/base_cli.rb
```diff
@@ -24,6 +25,32 @@ class << self
...
+      super(given_args, config)
+    end
+
+    # EULA acceptance
+    def self.check_license!
+      allowed_commands = ['-h', '--help', 'help', '-v', '--version', 'version']
+
+      require 'license_acceptance/acceptor'
+      begin
+        if (allowed_commands & ARGV.map(&:downcase)).empty? &&  # Did they use a non-exempt command?
```
zenspider: Might just be my brain (and it often is!), but the inverse of this makes more sense to me. Something like:

```ruby
!(ARGV.map(&:downcase) - allowed_commands).empty? # man I wish English had a proper antonym for empty.
```

zenspider (commented):
Just commentary

lib/inspec/base_cli.rb
artem-sidorenko: I did not to change the existing code, I just moved it without major changes. However I'm open for such changes/optimizations if needed

clintoncwolfe (approved):
Thanks - and this begins a refactor of cli.rb that I would love to continue :-)

lib/inspec/base_cli.rb
zenspider: Not needed. Just a suggestion. I'll get to it sooner or later.

zenspider (approved):

miah:
I resolved the merge errors, lets get this merged today.

miah:
We'll also need to resolve the chefstyle issues after merge; not a huge deal breaker.

miah (approved):

PR #4160, opened 2019-05-29
Make travis green again

zenspider:
How do you pronounce MTGA?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ branches:
 language: ruby
 cache:
   - bundler
-  directories:
+  - directories:
     - /home/travis/.rvm
 dist: xenial
 services:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ branches:
 language: ruby
 cache:
   - bundler
-  directories:
+  - directories:
     - /home/travis/.rvm
```
miah, line 9: Might need to be:

```
bundler: true
directories:
  - /home/travis/.rvm
```

miah (changes requested):

.travis.yml
zenspider: we'll worry about that later. Want green more than I want cache atm

miah (approved):

Issue #4002, opened 2019-04-30
Switch from jsonschema to json-schema?


zenspider:
jsonschema was last released in 2012 and emits warnings on load.

json-schema was last released in Oct 2018 and has ~16m downloads. It is probably a much better bet for us in the long term.

clintoncwolfe:
It is entirely possible that nothing is consuming the schema we store, and we may not need this functionality at all - need to check with the A2 team.

aaronlippold:
We use it

clintoncwolfe:
@aaronlippold Could you expand a bit - I'm curious as to what exactly your use case is?

zenspider:
A cursory check suggests we might be able to switch over with something as simple as:

```ruby
module JSON
  class Schema
    def self.validate data, schema=nil, _=nil
      JSON::Validator.validate schema, data
    end
  end
end
```

aaronlippold:
https://vuejsexamples.com/schema-based-form-generator-vue-js-2-0-component-based-on-vuetify/

Would be an example of how we plan to use it as well as possibly connections to graph QL and other things. Fundamentally it's very nice to be able to view what the current standard scheme is are for the date of structures of inspec. Is there a particular challenge in maintaining them, we have to maintain them either way don't we?

miah:
We aren't talking about removing the JSON Schemas, just changing the library we use. InSpec needs to define schemas for json files in our configuration, our interaction with automate, and telemetry endpoints.

miah:
I'm going to close this as I did the work. The next steps will be to utilize json-schema for defining JSON schemas inside InSpec =)

PR #4172, opened 2019-06-02
Replace jsonschema with json-schema

### LINKED ISSUE #4129: jsonschema dependency in Gemfile used in tests
Any test dependency should probably be specified as a development dependency in the gemspec.

```
Gemfile:  gem 'jsonschema', '~> 2.0.2'
Gemfile.lock:    jsonschema (2.0.2)
Gemfile.lock:  jsonschema (~> 2.0.2)
test/functional/inspec_check_test.rb:require 'jsonschema'
test/functional/inspec_exec_json_test.rb:require 'jsonschema'
test/functional/inspec_exec_jsonmin_test.rb:require 'jsonschema'
```

miah:
This replaces jsonschema with json-schema. I've also dropped the external  json requirement since Ruby ships with it. The tests were also updated to test the result. jsonschema was also a Gemfile dependency when it should have been in the gemspec.

fixes https://github.com/inspec/inspec/issues/4129  https://github.com/inspec/inspec/issues/4002

### DIFF
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'train-core', '~> 2.0'
   spec.add_dependency 'license-acceptance', '>= 0.2.13', '< 2.0'
   spec.add_dependency 'thor', '~> 0.20'
-  spec.add_dependency 'json-schema'
+  spec.add_dependency 'json-schema', '~> 2.8'
   spec.add_dependency 'method_source', '~> 0.8'
   spec.add_dependency 'rubyzip', '~> 1.1'
   spec.add_dependency 'rspec', '~> 3'
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'train-core', '~> 2.0'
   spec.add_dependency 'license-acceptance', '>= 0.2.13', '< 2.0'
   spec.add_dependency 'thor', '~> 0.20'
-  spec.add_dependency 'json', '>= 1.8', '< 3.0'
+  spec.add_dependency 'json-schema'
   spec.add_dependency 'method_source', '~> 0.8'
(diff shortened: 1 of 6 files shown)


clintoncwolfe (approved):
Thanks

inspec-core.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'train-core', '~> 2.0'
   spec.add_dependency 'license-acceptance', '>= 0.2.13', '< 2.0'
   spec.add_dependency 'thor', '~> 0.20'
-  spec.add_dependency 'json', '>= 1.8', '< 3.0'
+  spec.add_dependency 'json-schema'
```
zenspider: Unbounded dependencies will bite us. Please bound the top end with with `~> x.y`

test/functional/inspec_exec_json_test.rb
```diff
@@ -11,7 +11,7 @@
     data = JSON.parse(out.stdout)
     sout = inspec('schema exec-json')
     schema = JSON.parse(sout.stdout)
-    JSON::Schema.validate(data, schema)
+    JSON::Validator.validate(schema, data).wont_equal false
```
zenspider, line 14: Yes! Thank you for checking the result.

test/functional/inspec_exec_json_test.rb
```diff
@@ -21,7 +21,7 @@
...
```
zenspider, line 24: samsies.

zenspider (changes requested):

inspec-core.gemspec
miah: Ok, I figured I'd get called out on this one but I wasn't sure the right minimum here. Thanks!

test/functional/inspec_exec_json_test.rb
miah: Right, if we aren't going to check the result it almost doesn't make sense to validate it at all; unless we're expecting the test to blow up?

zenspider (approved):
Thank you! That was concerning!

PR #4177, opened 2019-06-03
Turn off parallel testing (for now?)

zenspider:
This will slow our tests down but hopefully completely eliminate the remaining test instability we have.

This also changes the way that rspec & minitest interact again, as I found that the previous way wasn't sufficient for mocha to survive rspec's invocation.


miah (approved):
Ooof.

zenspider:
Merging to push to stable while everyone is asleep :P

PR #4188, opened 2019-06-05
Fixed up interactions between minitest & simplecov.

zenspider:
Sometime during the 2.x's, `at_exit` changed its ordering. As a result,
a lot of things that were stacking `at_exit`'s broke. This is one of
those since both simplecov and minitest do their thing via `at_exit`.
This switches to simplecov w/ no defaults on, then replicates their
simplecov/defaults.rb with our own.

I'm going to try to get that entire file back upstream but it can live
here for now.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -5,19 +5,22 @@
 # Do not add any other code to this code block. Simplecov and
 # coveralls only until the next code block:
 
-require 'simplecov'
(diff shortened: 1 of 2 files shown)


miah (approved):
Its all coming together!

clintoncwolfe (approved):

PR #4189, opened 2019-06-05
Close at 400 days, with a message.

miah:
This will automatically close issues after 400 days of inactivity. 

I'm not particularly fond of the stale or close messages, would appreciate feedback there.


zenspider (approved):

clintoncwolfe (approved):
It seems clear and to the point. There is no message that is going to make people happy; we might consider asking @nellshamrell if she has some ideas about adding some friendly words.

I'm OK with the change, though :-)

zenspider:
@nellshamrell ?

PR #4192, opened 2019-06-10
Update test code to rename attribute() to input() in DSL

### LINKED ISSUE #4149: Test Fixtures - Update DSL input references
After the DSL rename for Inputs, rename all occurrences of 'attribute' in test fixtures to 'input'.

clintoncwolfe:
## Description

After #4008 renamed `attribute()` to `input()` in the DSL, all of our test fixtures that mention `attribute()` would one day emit deprecation warnings. This PR mechanically performs a bulk replace on the first commit. A code example is also included.

In addition, it was found that we were testing kitchen-inspec in this codebase, which didn't make sense - it is explicitly tested in its own codebase, where it should be.

Finally, a DSL-related bit of testing code was updated regarding how inputs are marshalled under `to_ruby`.

## Related Issue

Part of #3802
Followup to #4008
Closes #4149

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/inputs/input_test.rb
+++ b/test/unit/inputs/input_test.rb
@@ -37,14 +37,14 @@ describe Inspec::Input do
       # Should have the DSL call. This should be attribute(), not input(), for the
       # foreseeable future, to maintain backwards compatibility.
       ruby_code.must_include "attribute('application_port'"
-      ruby_code.must_include 'value: 80'
-      ruby_code.must_include 'default: 80'
+      ruby_code.must_include "value: 80"
+      ruby_code.must_include "default: 80"
       ruby_code.must_include "description: 'The port my application uses'"
 
       # Try to eval the code to verify that the generated code was valid ruby.
       # Note that the input() method is part of the DSL, so we need to
       # alter the call into something that can respond - the constructor will do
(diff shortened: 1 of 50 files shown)


miah (approved):

miah:
Overall looks fine, but some tests are now failing.

test/unit/inputs/input_test.rb
```diff
@@ -34,15 +34,17 @@
...
       ruby_code.must_include 'value: 80'
       ruby_code.must_include 'default: 80'
       ruby_code.must_include "description: 'The port my application uses'"
 
       # Try to eval the code to verify that the generated code was valid ruby.
-      # Note that the attribute() method is part of the DSL, so we need to
+      # Note that the input() method is part of the DSL, so we need to
       # alter the call into something that can respond - the constructor will do
-      ruby_code_for_eval = ruby_code.sub(/attribute\(/,'Inspec::Input.new(')
+      ruby_code_for_eval = ruby_code.sub(/input\(/,'Inspec::Input.new(')
```
zenspider: `ruby_code_for_eval` ? should I be scared?

I think there are safer ways to syntax check that don't require pulling in full parsers and the like. Not sure if it matters here since this is in our tests and presumably we control all of the input.

zenspider (approved):
One concern, but otherwise approved.

PR #4193, opened 2019-06-10
Speed up inspec

zenspider:
This attempts to shave off time from our tests by speeding up the implementation, mostly be rearranging when and how we load things. There's more to be done (I really want to remove the resource registration system!) but this should help us in the meantime.

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -1,19 +1,30 @@
 # Copyright 2015 Dominik Richter
 
-require "logger"
-require "thor"
-require "json"
-require "pp"
-require "inspec/utils/json_log"
-require "inspec/utils/latest_version"
-require "inspec/base_cli"
-require "inspec/plugin/v1"
-require "inspec/plugin/v2"
-require "inspec/runner_mock"
-require "inspec/env_printer"
-require "inspec/schema"
-require "inspec/config"
-require "inspec/dist"
+require "inspec/utils/deprecation/deprecator"
+
+module Inspec # TODO: move this somewhere "better"?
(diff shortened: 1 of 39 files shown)


lib/inspec/cli.rb
```diff
@@ -1,19 +1,33 @@
...
-require 'inspec/runner_mock'
-require 'inspec/env_printer'
-require 'inspec/schema'
-require 'inspec/config'
+require 'inspec/utils/deprecation/deprecator'
 require 'inspec/dist'
+require 'inspec/backend'
+require 'inspec/dependencies/cache'
+
+module Inspec # TODO: move this somewhere "better"?
```
miah, line 8: I fear overloading the word 'core'; but maybe something like `lib/inspec/core.rb`?

test/functional/inspec_exec_test.rb
```diff
@@ -8,6 +8,31 @@
     skip_windows!
   }
 
+#   it 'runs_directly' do
+#     require "inspec/cli"
+#     require "misc/exit_wtf"
```
miah: Is that a debugging crumb?

miah (approved):

test/functional/inspec_exec_test.rb
zenspider: ah yes. I'll remove that.

lib/inspec/cli.rb
zenspider: I do too... doesn't have to happen in this commit. that's why I leave TODO's behind as I travel through the code.

zenspider:
This is part of my optimization work, being tracked by #4212.

PR #4194, opened 2019-06-11
Apply chefstyle to codebase

zenspider:
Switched to chefstyle so we match the rest of chef.

Miah did 90%+ of the work. I mostly advised and then helped with a gnarly rebase.


miah (approved):

miah:
This had to be a quick merge because keeping this branch up dated with incoming changes was difficult.

PR #4195, opened 2019-06-12
Disable rubocop on Code Climate.

miah:
This is now managed by chefstyle which CodeClimate doesn't support. The code is linted before tests are started so style is already strictly enforced.


zenspider (approved):
Yes please!

PR #4205, opened 2019-06-17
Fix Expeditor Version Bump Script

clintoncwolfe:
## Description

The change to ChefStyle changed the string literal quotes from single quotes to double quotes. This broke the `update_version.sh` script used by Expeditor, so that `VERSION` was getting updated (to 4.5.1, 4.5.2, 4.5.3, etc) while lib/inspec/version.rb stayed at 4.5.1 . This kept the rubygem jammed at 4.5.1, so Artifactory kept overwriting it, and if we had released, it would have been a mislabelled gem artifact, containing 4.6.1 content but labeled as 4.5.1.

The update script was updated, using quote escaping as is used on chef/chef, and tested locally to verify it works.

As this is a release-blocking issue, and we need a release today for a critical issue, I will be expediting this PR.

## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/update_version.sh
+++ b/.expeditor/update_version.sh
@@ -6,5 +6,5 @@
 
 set -evx
 
-sed -i -r "s/VERSION = '.*'/VERSION = '$(cat VERSION)'/" lib/inspec/version.rb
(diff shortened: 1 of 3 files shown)


clintoncwolfe:
Expediting due to release-blocking issue.

PR #4206, opened 2019-06-17
Use the updated ruby-cleanup omnibus definition

tas50:
The contents of clean-static-libs is now part of ruby-cleanup which handles all the cleanup of ruby junk.


clintoncwolfe:
Dupe #4207

PR #4223, opened 2019-06-19
Set the umask on travis in before_install so that file resource permission matcher can be tested

zenspider:
This removes one of our timebomb skips, fixes it by manually setting the file's mode before testing, and provides better feedback in case of an actual failure.

META: this was beyond frustrating. Uninformative failure messages, changing umasks, and I suspect filesystem caching (such that changing the umask doesn't actually have an effect). Multiply that with a 10 minute feedback loop and this was not a productive cycle for what is a fairly trivial problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
 - "gem update --system '3.0'"
 - gem --version
 - bundle --version
+- umask 0022
 env:
   - SLOW=1 NO_AWS=1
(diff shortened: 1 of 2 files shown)


miah (approved):
Good find.

clintoncwolfe (approved):
Good catch, and I agree this is a regrettable thing to have to do.

PR #4225, opened 2019-06-20
[WIP] Move improved habitat scaffolding from core-plans to InSpec repo

TheLunaticScripter:

As part of the effortless project users depend on a reliable way to get updates to how InSpec is run using habitat. The easiest way to do that is through the use of scaffolding. Currently, scaffolding has been housed in core-plans but it has been recommended that the effortless scaffoldings be moved to the repositories where their parent binary lives. 

## Types of changes

## Checklist:

### DIFF
--- a/scaffolding-inspec/lib/scaffolding.sh
+++ b/scaffolding-inspec/lib/scaffolding.sh
@@ -1,0 +1,209 @@
+#
+# A scaffolding for an InSpec Profile
+#
+
+scaffolding_load() {
+  pkg_deps=(
+    "${pkg_deps[@]}"
+    "chef/inspec"
+  )
+  pkg_build_deps=(
+    "${pkg_build_deps[@]}"
+    "chef/inspec"
+    "core/jq-static"
+  )
+  pkg_svc_user="root"
+  pkg_svc_run="set_just_so_you_will_render"
+}
+
+do_default_before() {
+  if [ ! -f "$PLAN_CONTEXT/../inspec.yml" ]; then
+    message="ERROR: Cannot find inspec.yml."
+    message="$message Please build from the profile root"
+    build_line "$message"
+
+    return 1
+  fi
+
+  # Execute an 'inspec compliance login' if a profile needs to be fetched from
+  # the Automate server
+  if [ "$(grep "compliance: " "$PLAN_CONTEXT/../inspec.yml")" ]; then
+    _do_compliance_login;
+  fi
+}
+
+do_default_setup_environment() {
+  set_buildtime_env PROFILE_CACHE_DIR "$HAB_CACHE_SRC_PATH/$pkg_dirname"
+  set_buildtime_env ARCHIVE_NAME "$pkg_name-$pkg_version.tar.gz"
+
+  # InSpec loads `pry` which tries to expand `~`. This fails if HOME isn't set.
+  set_runtime_env HOME "$pkg_svc_var_path"
+
+  # InSpec will create a `.inspec` directory in the user's home directory.
+  # This overrides that to write to a place within the running service's path.
+  # NOTE: Setting HOME does the same currently. This is here to be explicit.
+  set_runtime_env INSPEC_CONFIG_DIR  "$pkg_svc_var_path"
+}
+
+do_default_unpack() {
+  # Change directory to where the profile files are
+  pushd "$PLAN_CONTEXT/../" > /dev/null
+
+  # Get a list of all files in the profile except those that are Habitat related
+  profile_files=($(ls -I habitat -I results -I "*.hart"))
+
+  mkdir -p "$HAB_CACHE_SRC_PATH/$pkg_dirname" > /dev/null
+
+  # Copy just the profile files to the profile cache directory
+  cp -r ${profile_files[@]} "$HAB_CACHE_SRC_PATH/$pkg_dirname"
+}
+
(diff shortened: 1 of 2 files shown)


scaffolding-inspec/lib/scaffolding.sh
```diff
@@ -0,0 +1,209 @@
...
+  if [ "$(grep "compliance: " "$PLAN_CONTEXT/../inspec.yml")" ]; then
+    _do_compliance_login;
+  fi
+}
+
+do_default_setup_environment() {
+  set_buildtime_env PROFILE_CACHE_DIR "$HAB_CACHE_SRC_PATH/$pkg_dirname"
+  set_buildtime_env ARCHIVE_NAME "$pkg_name-$pkg_version.tar.gz"
+
+  # InSpec loads `pry` which tries to expand `~`. This fails if HOME isn't set.
```
clintoncwolfe, line 39: `pry` is only used for `inspec shell`, which is used by humans to interactively try things out - not likely to ever be used in this scenario. I would expect that `pry` would only be loading if `shell` were being executed, but maybe it is always loading. If it is a pain point we could look into it and try to eliminate the need for setting HOME.

We don't know about all these little rough edges, it's great to learn about them.

scaffolding-inspec/lib/scaffolding.sh
```diff
@@ -0,0 +1,209 @@
...
+  set_runtime_env HOME "$pkg_svc_var_path"
+
+  # InSpec will create a `.inspec` directory in the user's home directory.
```
clintoncwolfe, line 42: Right, .inspec is where we keep a number of things, including the profile cache, config file, and user plugins. Each of those have their own individual ENV var, as well, but overriding INSPEC_CONFIG_DIR is the simplest thing to do.

**Making** .inspec/ automatically is a bit rude; do you think we should not do that, unless we absolutely must write something?

scaffolding-inspec/lib/scaffolding.sh
```diff
@@ -0,0 +1,209 @@
...
+CFG_LOG_LEVEL={{cfg.log_level}}
+CFG_LOG_LEVEL="\${CFG_LOG_LEVEL:-warn}"
+CFG_CHEF_LICENSE={{cfg.chef_license.acceptance}}
+CFG_CHEF_LICENSE="\${CFG_CHEF_LICENSE:-undefined}"
+CONFIG="{{pkg.svc_config_path}}/inspec_exec_config.json"
+PROFILE_PATH="{{pkg.path}}/{{pkg.name}}-{{pkg.version}}.tar.gz"
+
+inspec_cmd()
+{
+  inspec exec \${PROFILE_PATH} --json-config \${CONFIG} --chef-license \$CFG_CHEF_LICENSE --log-level \$CFG_LOG_LEVEL
```
clintoncwolfe, line 88: --config , not --json-config . You can still use the old name but it didn't make a lot of sense  to emphasize the format and make the option 5 chars longer.

scaffolding-inspec/lib/scaffolding.sh
```diff
@@ -0,0 +1,209 @@
...
+                 -o "$HAB_CACHE_SRC_PATH/$pkg_dirname/$pkg_name-$pkg_version.tar.gz" \
+                 --chef-license "$CHEF_LICENSE"
+}
+
+do_default_install() {
+  cp "$HAB_CACHE_SRC_PATH/$pkg_dirname/$pkg_name-$pkg_version.tar.gz" "$pkg_prefix"
+
+  mkdir -p "$pkg_prefix/config"
+  chmod 0750 "$pkg_prefix/config"
+  cat << EOF >> "$pkg_prefix/config/cli_only.json"
```
clintoncwolfe, line 122: No rush - and certainly no blocker on this PR - but I'd really like to collaborate on how we can safely introduce changes to the JSON interfaces - the config InSpec accepts, and the reports it emits. Here, I see you're using the legacy format  (which is fine). 
We actually accept a more recent format of the config, which is mainly extended for handling complex scanning credentials like those needed for *scanning* a habitat installation). We added a version number to the format, so we can identify what version people are asking us to deal with, so we can validate it. We'll support the legacy format for a long time (maybe forever!), but eventually there may be features that you need in a future config format - and I think versioned config formats are one safe way to introduce change.

clintoncwolfe (commented):
Just a few comments at this point - thanks for showing this to us!

echohack:
Hey there @TheLunaticScripter @clintoncwolfe 

Recently we've discovered that the way Expeditor works limits us from including scaffolding-chef and scaffolding-inspec in their respective parent projects. In other words, Expeditor expects each project to have a 1:1 mapping with the release lifecycle of ALL included packages.

Since both scaffolding-inspec and scaffolding-chef are implementations of best practices *around*  the core products, they may need to live on a different release lifecycle.

So we are moving these projects to separate repos so that they can be released independently.

These repos are:

github.com/chef/scaffolding-chef
github.com/chef/scaffolding-inspec

I'm currently in the process of seeding these repos with some of the initial requirements for them, at which point we'll PR this scaffolding into the scaffolding-inspec repo.

Thank you very much!! Please let me know if you have questions and we can wax about it <3

TheLunaticScripter:
After discussion with Release Engineering I've moved this PR to https://github.com/chef/scaffolding-inspec/pull/4.

PR #4232, opened 2019-06-21
Update date on our timebomb.

miah:
Bumping this till we're all back from vacation. 

This covers most of our windows tests while we're working on fixing those functionals and migrating from appveyor -> buildkite

ref: #4204


clintoncwolfe (commented):
Thank you miah!

PR #4233, opened 2019-06-22
Debian service detection for Buster

jaredledvina:
## Description
Currently, /etc/debian_version for the Debian Buster/Sid release contains "buster/sid". This causes version to be set to 0 resulting in SysV being configured as the service provider incorrectly. This commit updates the debian detection logic to work for "buster/sid" as version 10 and makes sense that we have a non-0 version before setting the init system.

## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks for the contribution and the thorough testing!

miah (approved):

PR #4236, opened 2019-06-24
Add ip6tables resource

### LINKED ISSUE #1011: ip6tables
### Description

A ip6tables resource to verify ip6 configuration as required by MITRE as requested in #563

ramereth:
## Description

This should resolve #1011 which provides an ip6tables resource to test IPv6
iptables rules. This is essentially a copy of the iptables resource with a few
renames.

In addition, I've pulled in the integration tests for iptables into ip6tables
and enabled it on docker so that it properly gets tested regularly. The test
cookbook recipe has been updated to support all of the current platforms that
are being tested.

## Related Issue
- #1011 

## Types of changes

## Checklist:


ramereth:
So it looks like the hosts that run the kitchen suites don't have the IPv6 iptables  kernel module loaded which is a shame. I wonder if running modprobe before kitchen will resolve that, but I'm not sure if that's what you'd like to do.

Also, if its preferred, I can refactor this so that the code is reused between iptables and ip6tables. I really think having a separate resource for ip6tables is important however. Let me know!

zenspider (approved):
Thank you for the PR!

miah (approved):
Thanks @ramereth

PR #4238, opened 2019-06-25
Add CLI target examples for Windows domain users

jerryaldrichiii:
This adds an example of using `-t winrm://` with a domain account. This is different because using two `@` (e.g. `winrm://user@domain@somehost`) is confusing and probably breaks the parser.

I'm not a fan of how it wraps...but also not sure how to solve it.

## Description
Current:
Screen Shot 2019-06-25 at 15 36 47

New:
Screen Shot 2019-06-25 at 15 28 55

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,17 @@ Chef InSpec makes it easy to run your tests wherever you need. More options are 
 # run test locally
 inspec exec test.rb
 
-# run test on remote host on SSH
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
Thanks for such a clear example!

miah (approved):

PR #4242, opened 2019-06-28
windows_task: Fix handling of multiple triggers

### LINKED ISSUE #1430: windows_task fails if multiple triggers
### Description

`windows_task` fails if there are multiple triggers in a task.

### Replication Case

Just create a task with 2 dummy triggers.
Any test will fail on it because of code wait for a `{ Type: ""... }` and get `[ { Type: "" ....}, { Type: "" ...} ]`

### Possible Solutions

* Use `win32ole` lib instead of a `schtasks.exe` command.
* Test if response is an array and get first (the objects in array are strictly the same)

### LINKED ISSUE #3395: Resource windows_task fails when multiple triggers defined
### 🎛 Description

🐛 bug report

When running an inspec test to verify the settings of a windows_task, it fails with "∅  no implicit conversion of String into Integer" if the task has multiple triggers defined.

I've tested this with inspec versions 1.45.13, which I subsequently upgraded to 2.2.70, both have the same behavior.

Tests run on Mac OS 10.13 against both Windows 10 Pro and Windows 10 Enterprise clients.

### 🤔 Replication Case

1. Create two tasks on a Windows 10 client.  The action does not matter, but set one to have a single trigger (does not matter what the trigger is) and set the other two have two triggers (again, the actual triggers are irrelevant)
2. Create an inspec test to verify the tasks.  What you check does not matter, so a simple check for existing is sufficient)
3. Run the test.
4. Observe the following sad output:
fail 1

5. Add a second trigger to the task that previously had only one.
6. Remove one trigger from the task that previously had multiple.
7. Re-run the inspect test, and notice that the failures switch like so:
fail 2

I've tested multiple tasks, triggers, actions, etc.  The only thing that seems to matter is that it will always fail if there is more than one trigger defined on the task.

### 💁 Possible Solutions

I have no brilliant suggestion at this time, sadly!

jerryaldrichiii:
## Description 
When using `schtasks` a list is returned if the task has multiple triggers. This merges that list with the last item taking precedence. This is how `Get-ScheduledTask` behaves.

Initially, I was going to rewrite this resource to use `Get-ScheduledTask` but the original author purposely did not do this so that PowerShell v3 would be supported. We only support PowerShell v5, but I don't want to break any current users and this change didn't seem to controversial to me.

If it gives us trouble, I recommend rewriting it to use `Get-ScheduledTask`.

## Related Issue
This fixes #3395, which was opened when #1430 was closed due to staleness. (@clintoncwolfe edit for clarity)

## Types of changes

## Checklist:


jerryaldrichiii:
Until this is merged (or if you're using an older InSpec version) here are a few workarounds you can use:

## Simple workaround

This will check if the scheduled task exists:

```ruby
task_name = 'Multiple Triggers Example'
describe command("schtasks /query /tn '#{task_name}'") do
end
```

Output:
```
Command: `schtasks /query /tn 'Single Trigger'`
   ✔  exit_status should eq 0
```
## More Advanced workaround
This will allow you to check any of the following:
  - Status
  - Logon Mode
  - Last Result
  - Task To Run
  - Run As User
  - Scheduled Task State

```ruby
task_name = 'Multiple Triggers Example'
task_query = "schtasks /query /v /fo csv /tn '#{task_name}' | ConvertFrom-Csv | Select @{N='URI';E={$_.TaskName}},@{N='State';E={$_.Status.ToString()}},'Logon Mode','Last Result','Task To Run','Run As User','Scheduled Task State' | ConvertTo-Json -Compress"

result = JSON.parse(inspec.command(task_query).stdout).reduce(:merge)
describe "Scheduled Task: #{task_name}" do
  it 'should have a state of either "Ready" or "Running"' do
    expect(result['State']).to be_in(['Ready', 'Running'])
  end
  it 'should have a Logon Mode of "Interactive only"' do
    expect(result['Logon Mode']).to eq('Interactive only')
  end
  it 'should have a Last Result of "1"' do
    expect(result['Last Result']).to_not eq('1')
  end
  it 'should execute a command matching "Taskmgr"' do
    expect(result['Task To Run']).to match(/Taskmgr/)
  end
  it 'should be enabled' do
    expect(result['Scheduled Task State']).to eq('Enabled')
  end
end
```

Output:

```
Scheduled Task: Multiple Triggers
   ✔  should have a state of either "Ready" or "Running"
   ✔  should have a Logon Mode of "Interactive only"
   ✔  should have a Last Result of "1"
   ✔  should execute a command matching "Taskmgr"
   ✔  should be enabled
```

jerryaldrichiii:
Until this is merged (or if you're using an older InSpec version) here are a few workarounds you can use:

## Simple workaround

This will check if the scheduled task exists:

```ruby
task_name = 'Multiple Triggers Example'
describe command("schtasks /query /tn '#{task_name}'") do
end
```

Output:
```
Command: `schtasks /query /tn 'Multiple Triggers Example'`
   ✔  exit_status should eq 0
```
## More Advanced workaround
This will allow you to check any of the following:
  - Status
  - Logon Mode
  - Last Result
  - Task To Run
  - Run As User
  - Scheduled Task State

```ruby
task_name = 'Multiple Triggers Example'
task_query = "schtasks /query /v /fo csv /tn '#{task_name}' | ConvertFrom-Csv | Select @{N='URI';E={$_.TaskName}},@{N='State';E={$_.Status.ToString()}},'Logon Mode','Last Result','Task To Run','Run As User','Scheduled Task State' | ConvertTo-Json -Compress"

result = JSON.parse(inspec.command(task_query).stdout).reduce(:merge)
describe "Scheduled Task: #{task_name}" do
  it 'should have a state of either "Ready" or "Running"' do
    expect(result['State']).to be_in(['Ready', 'Running'])
  end
  it 'should have a Logon Mode of "Interactive only"' do
    expect(result['Logon Mode']).to eq('Interactive only')
  end
  it 'should have a Last Result of "1"' do
    expect(result['Last Result']).to_not eq('1')
  end
  it 'should execute a command matching "Taskmgr"' do
    expect(result['Task To Run']).to match(/Taskmgr/)
  end
  it 'should be enabled' do
    expect(result['Scheduled Task State']).to eq('Enabled')
  end
end
```

Output:

```
Scheduled Task: Multiple Triggers Example
   ✔  should have a state of either "Ready" or "Running"
   ✔  should have a Logon Mode of "Interactive only"
   ✔  should have a Last Result of "1"
   ✔  should execute a command matching "Taskmgr"
   ✔  should be enabled
```

jerryaldrichiii:
I...I don't know how to make it better 😞 

Pushing up what I could though. Suggestions/fixes welcome.

jerryaldrichiii:
I...I don't know how to make it better 😞

Pushing up what I could though. Suggestions/fixes welcome.

miah (approved):

clintoncwolfe (approved):

PR #4244, opened 2019-06-28
Config final options to include additional items specific to plugins

### LINKED ISSUE #4243: Config final options to include additional items specific to plugins
## Detailed Description
I am developing an Input plugin that needs configuration information.   Without coupling it with a cli plugin or a hard-coded path, in the plugin, it would be helpful to obtain the configuration information from the Inspec configuration file.  I would prefer this over allowing additional command line flags (which is in the Config's final_options hash).  

## Context
```
inspec-bin/bin/inspec exec --target ssh://remote -- ~/git/inspec/inspec_truth_plugin/examples/controls/example.rb 

Unrecognized top-level configuration field truth_services.  Recognized fields: cli_options, compliance, credentials, reporter, version

cat ~/.inspec/config.json
{
  "version" : "1.1",
  "cli_options" : {
    "color" : true
  },
  "credentials" : {
    "ssh" : {
      "remote" : {
        "host" : "my_remote_host",
        "user" : "user1",
        "password" : "password",
        "sudo" : true,
        "sudo_password" : "password"
      }
    }
  },
  "reporter" : {
    "cli" : {
      "stdout" : true
    },
    "json" : {
      "file" : "output.json",
      "stdout" : false
    }
  },
  "truth_services" : {
    "local_file" : "example_truths.yml"
  }
}
```
## Possible Implementation
Inspec::Config lines 287
Change valid_fields to private global array with the current default values.  Through class methods allow for the list to expand for plugins to 'register' configuration fields that they expect to use.
Inspec::Config lines 346
options.merge!(@cfg_file_contents)

(@clintoncwolfe - edit for code formatting)

kevin-j-smith:
closes #4243 

## Description
Took the method described in the issue.   1) changed valid_fields to a class variable and provided a accessor method to append new values to it.   2) used this variable for config validation.  3) added cfg_file_contents to the final_options.  

### Before change with key:value in configuration
Unrecognized top-level configuration field truth_services.  Recognized fields: cli_options, compliance, credentials, reporter, version

### After change with key:value in configuration, but Input plugin not registering valid field
  ↺  example: Example
     ↺  Input 'example_input' does not have a value. Skipping test.

### After change with key:value in configuration and Input plugin registering valid field 'truth_services'
  ✔  example: Example
     ✔  WORLD should eq "WORLD"

### Example Input plugin with registration
```ruby
# frozen_string_literal: true

require 'inspec/config'
  
module InspecPlugins
  module InspecTruthPlugin
    class Plugin < Inspec.plugin(2)
      
      # Metadata
      # Must match entry in plugins.json
      plugin_name :'inspec-truth_plugin'

      # Activation hooks
      input :local_file do
        require_relative 'local_file_input.rb'
        InspecPlugins::InspecTruthPlugin::LocalFileInput
      end
    end
  end
end

Inspec::Config.regsiter_valid_field('truth_services')
```

## Related Issue
#4243 

## Types of changes
- [ X ] New content (non-breaking change)

## Checklist:
- [ X ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -23,6 +23,8 @@ module Inspec
       shell_command
     }.freeze
 
+    @@valid_fields = %w{version cli_options credentials compliance reporter}
+
     extend Forwardable
 
     # Many parts of InSpec expect to treat the Config as a Hash
@@ -75,6 +77,11 @@ module Inspec
     # @return [Hash]
     def telemetry_options
       final_options.select { |key, _| key.include?("telemetry") }
+    end
+
+    # Use this to register valid configuration fields
+    def self.regsiter_valid_field(field)
+      @@valid_fields.push(field) unless field && (@@valid_fields.include? field)
     end
 
     #-----------------------------------------------------------------------#
@@ -284,10 +291,9 @@ module Inspec
         raise Inspec::ConfigError::Invalid, "Unsupported config file version '#{version}' - currently supported versions: 1.1"
       end
 
-      valid_fields = %w{version cli_options credentials compliance reporter}.sort
       @cfg_file_contents.keys.each do |seen_field|
-        unless valid_fields.include?(seen_field)
-          raise Inspec::ConfigError::Invalid, "Unrecognized top-level configuration field #{seen_field}.  Recognized fields: #{valid_fields.join(', ')}"
+        unless @@valid_fields.include?(seen_field)
+          raise Inspec::ConfigError::Invalid, "Unrecognized top-level configuration field #{seen_field}.  Recognized fields: #{@@valid_fields.join(', ')}"
         end
       end
     end
@@ -343,6 +349,8 @@ module Inspec
 
       # Highest precedence: merge in any options defined via the CLI
       options.merge!(@cli_opts)
+
+      options.merge!(@cfg_file_contents)
 
       options
     end
--- a/test/unit/config_test.rb
+++ b/test/unit/config_test.rb
@@ -72,7 +72,7 @@ describe "Inspec::Config" do
     describe "when the file is a valid v1.1 file" do
       let(:fixture_name) { "basic" }
       it "should read the file successfully" do
-        expected = %w{create_lockfile reporter type}.sort
+        expected = %w{cli_options create_lockfile credentials reporter type version}.sort
         seen_fields.must_equal expected
       end
     end
@@ -80,7 +80,7 @@ describe "Inspec::Config" do
     describe "when the file is minimal" do
       let(:fixture_name) { "minimal" }
       it "should read the file successfully" do
-        expected = %w{reporter type}.sort
+        expected = %w{reporter type version}.sort
         seen_fields.must_equal expected
       end
     end
@@ -130,7 +130,7 @@ describe "Inspec::Config" do
     describe "when the exec command is used" do
       let(:command) { :exec }
       it "should have the correct defaults" do
-        expected = %w{color create_lockfile backend_cache reporter show_progress type}.sort
+        expected = %w{color create_lockfile backend_cache reporter show_progress type version}.sort
         seen_fields.must_equal expected
         final_options["reporter"].must_be_kind_of Hash
         final_options["reporter"].count.must_equal 1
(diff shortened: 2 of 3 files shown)


lib/inspec/config.rb
```diff
@@ -23,6 +23,8 @@ class Config
       shell_command
     }.freeze
 
+    @@valid_fields = %w{version cli_options credentials compliance reporter}
```
zenspider, line 26: Is there a reason why you went with class variable instead of constant or class instance variable?

lib/inspec/config.rb
```diff
@@ -344,6 +350,8 @@ def merge_options
       # Highest precedence: merge in any options defined via the CLI
       options.merge!(@cli_opts)
 
+      options.merge!(@cfg_file_contents)
```
zenspider, line 353: This seems off topic for this commit, no? Is it actually related?

test/unit/config_test.rb
```diff
@@ -72,15 +72,15 @@
     describe "when the file is a valid v1.1 file" do
       let(:fixture_name) { "basic" }
       it "should read the file successfully" do
-        expected = %w{create_lockfile reporter type}.sort
+        expected = %w{cli_options create_lockfile credentials reporter type version}.sort
```
zenspider, line 75: totally not on you since this already existed... but why are we sorting a static already-sorted array?

zenspider (changes requested):
Some minor questions about the commits.

lib/inspec/config.rb
kevin-j-smith: The config was not instantiated before input plugins were being registered, thus not class instance variable.   The value changes during runtime, thus not constant.  At least that was my thought processes.  I made a specific setter for this value, named for its intention.  I am amicable.

lib/inspec/config.rb
kevin-j-smith: This all comes down to what is available to input plugins.  When I was debugging, and trying to figure out exactly how this all works :), The final_options was all that was available.  I would take some KT on this if there is another way.   The user scenario was to have one configuration file for Inspec to maintain.

test/unit/config_test.rb
kevin-j-smith: Ha, yeah, OCD a little.   When reviewing the test and failures it was easier to order them to see what was missing and to make sure it was logical that they should now exist with the changes being made.  I could not find anything that was trying to double validate that the order mattered.

clintoncwolfe (changes requested):
Thanks for the contribution! However, I think we should discuss the overall approach taken. As a general project goal, we're aiming to have versioned schema configuration files. Version 1.1 is currently fixed; I think any new changes should be added by creating a new version, 1.2, rather than dynamically adding new fields at runtime. We already have a mechanism for determining what fields are valid (reading the version number).

I agree that having config options for plugins is useful - so I suggest we add a namespace for that, and increment the plugin version.  One approach might be:

```
{
  "version": "1.2",
  # Existing fields...
  "plugins": {
    "inspec-plugin-name1": {
      "private-setting1":"private-value1",
      "private-setting2":"private-value2"
    },
    "inspec-plugin-name2": {}
  }
}
```

Under each plugin's setting area, the Config system would be agnostic as to schema - no opinion as to what fields are or are not present; if a plugin wants to validate its settings, would need to do so on its own.

There could then be a corresponding API extension, something like `config.plugin_settings('inspec-plugin-name')` that would return the fetched settings, if any.

Thoughts?

lib/inspec/config.rb
clintoncwolfe: As you are the first public contributor known to be writing an input plugin, I'd love to hear more about your experiences and how we can make it a convenient API. It is extremely minimal at the moment.

kevin-j-smith:
@clintoncwolfe, yep that does make sense.  I will take another shot at the changes as per your insights, thanks

lib/inspec/config.rb
kevin-j-smith: As of yet, I have not done much with it.   As I am taking a MVP iterative approach to some use cases, I will be building out a better story.  Basically, our silo'ed ecosystem does not align well with a centralized auditor.  Too many heads.  What I am wishing to build out is a 'Oracle' service that Inspec can chat with to get quorum agreed upon values.  As of this time, the only down side was that I had no method to configure the plugin without statically defining a file.

miah (changes requested):
Hey @kevin-j-smith have you had any time to address the change requests from the rest of the team?

clintoncwolfe:
I need this feature for something I am working on; I'm adopting this PR

clintoncwolfe:
Superseded by #4406 - thanks for the contribution!

PR #4246, opened 2019-07-01
use stdlib SecureRandom to generate a password

robbkidd:
## Description

Replaces the now-yanked passgen gem with something available in the Ruby standard library. Why `.base64(23)`? Base64 with an input of 23 was reliably generating 32-character strings with uppercase, lowercase, numerics, and symbols.

From the docs on the Azure terraform provider:

> NOTE: admin_password must be between 6-72 characters long and must satisfy at least 3 of password complexity requirements from the following:
> 1. Contains an uppercase character
> 2. Contains a lowercase character
> 3. Contains a numeric digit
> 4. Contains a special character

So, this should suffice.

## Related Issue

No issue open at the moment. This should resolve the error:

```
Could not find gem 'passgen' in any of the gem sources listed in your Gemfile.
```

When `bundle install` is run on the project.

## Types of changes

- ~New content (non-breaking change)~
- ~Breaking change (a content change which would break existing functionality or processes)~

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,6 @@ group :test do
   gem "mocha", "~> 1.1"
   gem "ruby-progressbar", "~> 1.8"
   gem "webmock", "~> 3.0"
-  gem "passgen"
   gem "m"
   gem "pry", "~> 0.10"
   gem "pry-byebug"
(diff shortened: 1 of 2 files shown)


robbkidd:
Updated to a 72-character string. Move the require to within the task.

zenspider (approved):

tas50 (approved):
Seems super straight forward to me

PR #4251, opened 2019-07-02
Permit plugin to be in path with git ref

### LINKED ISSUE #3904: Cannot load git-referenced bundled plugin
Using 3.7.11 from a gem install

```
➜  my-cool-project git:(my_branch) ✗ bundle exec inspec plugin install $(bundle show inspec-something)
Unrecognizable plugin structure - inspec-something-812f55372c8e - When installing from a path, please provide the path of the entry point file - installation failed.
```

The plugin source is at `/Users/jamesstocks/code/my-cool-project/vendor/bundle/ruby/2.3.0/bundler/gems/inspec-something-812f55372c8e/lib/inspec-something.rb`

Looking at the supported plugin loading cases [here](https://github.com/inspec/inspec/blob/master/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb#L225) it looks like this setup does not match a supported case.

Workaround:
`ln -s $(bundle show inspec-something) ~/.inspec/plugins/inspec-something` works, but symlinking in a plugin like this is not recommended/supported.

james-stocks:
Fixes #3904

## Description

Fixes https://github.com/inspec/inspec/issues/3904 by allowing a plugin to be found in a folder that refers to a git SHA e.g. `/Users/jamesstocks/code/bundles/cool-project/ruby/2.4.0/bundler/gems/inspec-special-62e42ce9eca1`

## Related Issue

#3904

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
+++ b/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
@@ -229,6 +229,7 @@ module InspecPlugins
         given = given.expand_path # Resolve any relative paths
         name_regex = /^(inspec|train)-/
         versioned_regex = /^(inspec|train)-[a-z0-9\-\_]+-\d+\.\d+\.\d+$/
+        sha_ref_regex = /^(inspec|train)-[a-z0-9\-\_]+-[0-9a-f]{5,40}$/
 
         # What are the last four things like?
         parts = [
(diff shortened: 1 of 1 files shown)


lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -229,6 +229,7 @@ def install_from_path__apply_entry_point_heuristics(path) # rubocop: disable Met
         given = given.expand_path # Resolve any relative paths
         name_regex = /^(inspec|train)-/
         versioned_regex = /^(inspec|train)-[a-z0-9\-\_]+-\d+\.\d+\.\d+$/
+        sha_ref_regex = /^(inspec|train)-[a-z0-9\-\_]+-[0-9a-f]{5,40}$/
```
james-stocks, line 232: I could make a single regex that matches either of these 2 regexes.
That would keep the lines changed below simpler; but might make the regex(es) unreasonable to follow.

james-stocks:
I couldn't find existing unit tests that cover the `CliCommand` class that's changed by this PR.
I can add some unit tests if I get a pointer on the right place in the current test setup to add them.

clintoncwolfe (approved):
Thanks for this - after a zoom discussion, we determined that this is needed for CI on the Compliance Profiles team, not just for developer convenience. This PR doesn't currently have any docs or testing on it, but as it is likely to only be used by the Content team and they are willing to accept those risks, I feel OK accepting this without tests. If you want to add tests, @james-stocks, they would go here: https://github.com/inspec/inspec/blob/cec41211526845f5850929e863a02a01f0625e1e/lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb#L265

miah (approved):

PR #4269, opened 2019-07-08
Add allowed users example to Users resource

shoekstra:
This example shows how to check for expected users and to flag any other
users that should not be present.

Signed-off-by: Stephen Hoekstra <stephenhoekstra@gmail.com>

### DIFF
--- a/docs/resources/users.md.erb
+++ b/docs/resources/users.md.erb
@@ -58,6 +58,20 @@ The following examples show how to use this Chef InSpec audit resource.
       it { should exist }
     end
 
+### Test only allowed users exist
+
+    allowed_users = %w(user1 user2 user3)
+
+    users.where { uid > 1000 && uid < 65534 }.usernames.sort.each do |u|
+      describe user(u) do
+        if allowed_users.include?(u)
+          it { should exist }
+        else
+          it { should_not exist }
+        end
+      end
+    end
+
 <br>
 
 ## Matchers


jerryaldrichiii:
Brining the conversation out of Slack into here 😄 

I'm not sure I like the `if/else` bits...but it's not wrong per se.

Here's how I would have done it:

```ruby
### Ensure only users in allowed_users exist (with UIDs > 500)

    allowed_users = %w{jerry boiarde}
    current_users = users.where { uid >= 1000 }.usernames
    invalid_users = users.where { uid >= 1000 && !allowed_users.include?(username) }.usernames

    describe 'Users with uid >= 500' do
      it 'should only contain allowed users' do
        expect(invalid_users).to be_empty
      end
      it 'should contain all allowed users' do
        expect(current_users).to include(allowed_users)
      end
    end
```

Thoughts?

EDIT:

Output from command (I'm still not happy with it):
```
  Users with uid >= 500
     ×  should only contain allowed users
     expected `["nobody"].empty?` to return true, got false
     ×  should contain all allowed users
     expected ["nobody"] to include ["jerry", "boiarde"]
```

miah (approved):
Thanks @shoekstra

docs/resources/users.md.erb
```diff
@@ -58,6 +58,20 @@ The following examples show how to use this Chef InSpec audit resource.
       it { should exist }
     end
 
+### Test only allowed users exist
+
+    allowed_users = %w(user1 user2 user3)
+
+    users.where { uid > 1000 && uid < 65534 }.usernames.sort.each do |u|
```
jerryaldrichiii, line 65: No need for the upper bound I suspect.

jerryaldrichiii (approved):

jerryaldrichiii:
I changed my mind, your example is simpler.

aaronlippold:
Actually, wouldn't it be better to use the be_in matcher here? Provide much better reporting output.

docs/resources/users.md.erb
shoekstra: Maybe not, in our environment it's the `nfsnobody` user but happy to omit it for the example.

shoekstra:
@aaronlippold when I tested locally I got this:

```
Profile: inspec-shell
Version: (not specified)

  User test1
     ✔  should exist
  User test2
     ✔  should exist
...
  User test3
     ×  should not exist
     expected User test3 not to exist

Test Summary: 18 successful, 2 failures, 0 skipped
```

I'm happy with this output.

@jerryaldrichiii I like what you did and it may have been something I would do, but for the example wanted to keep it short and to the point like other examples I've come across.

Issue #4291, opened 2019-07-12
Service resource `be_running?` is deprecated but `be_active?` throws an error


russellseymour:
When a service is being tests the documentation states that to see if the service is running the matcher `be_running` should be used.

However when run a deprecation notice is output stating that this is only for backward compatibility:

```
DEPRECATION: The service `be_running?` matcher is deprecated. This is only allowed for compatibility with ServerSpec (used at C:/Users/russells/workspaces/azure/camsa-setup/test/integration/camsa/monitoring.rb:82)
```

If this is changed to use `be_active` the error on the test is as follows:

```
  Service statsd
     [PASS]  should be enabled
     [FAIL]  should be active
     expected Service statsd to respond to `active?`
```

Looking at the code the command that runs the `active?` check runs the command

systemctl is-active statsd --quiet

The service is installed on the machine as a systemd service. If I run this command on the machine I get the the output of `active` if running or `inactive` of not running.

Indeed if running the exit code is 0 and if not running it is 3 which is what is required. However the output of the test states that `active?` is not responded to.

jeremymv2:
Some additional context:
https://github.com/inspec/inspec/issues/4059#issuecomment-499397982

clintoncwolfe:
It looks like `active?` is an internal predicate used by some of the OS-specific implementations, not intended to be exposed as a resource matcher - so I don't think it is expected to work. "active" is a concept specific to certain service systems, and I think the `service` resource is trying to be more generic - I say that simply because I don't see something like `be_active` documented anywhere.

Instead, calling `be_running` really is the right thing to do here - and I've opened a PR #4295 to remove that deprecation warning.

kekaichinose:
@russellseymour per @clintoncwolfe's explanation, it sounds like your issue with `be_running?` will be resolved with the removal of the deprecation warning in PR #4295. Can you confirm?

If so, we'll make sure to link them so this issue will be closed when that PR is merged. Thanks!

clintoncwolfe:
Closing, #4295 is merged!

PR #4295, opened 2019-07-12
Un-deprecate the be_running matcher on service resource

### LINKED ISSUE #4059: `be_running` is deprecated, what should I use?
```
WARN: DEPRECATION: The service `be_running?` matcher is deprecated. This is only allowed for compatibility with ServerSpec
```

What should I use instead?

Also, this is in all the docs so those should be changed as well.

clintoncwolfe:

## Description

A close reading of what is going on here made it clear to me that there is a lot of unnecessary grief happening. The matcher is already restricted to the `service` resource (I had initially misunderstood and was under the impression it was a universal matcher); and I cannot speak to any long-term goal regarding maintaining or discarding serverspec compatibility.

Anyway, this message confuses a lot of people, and it should only have been in place when a clear alternative was offered. Let's withdraw the deprecation message and live a more peaceful life.

Thanks to @jerryaldrichiii 	to pointing that conclusion out several weeks ago.

## Related Issue

Closes #4059

See also #4291

## Types of changes

## Checklist:

### DIFF
--- a/lib/matchers/matchers.rb
+++ b/lib/matchers/matchers.rb
@@ -104,7 +104,6 @@ end
 # Deprecated: You should not use this matcher anymore
 RSpec::Matchers.define :be_running do
   match do |service|
-    Inspec.deprecate(:serverspec_compatibility, "The service `be_running?` matcher is deprecated.")
     service.running? == true
   end
 


zenspider (approved):

stevendanna (approved):
Thank you!

PR #4307, opened 2019-07-16
Removed integration tests

zenspider:
They've been red for weeks and nobody has addressed it so I'm removing
them.


miah:
+1. We haven't managed these well, and I think there is _some_ benefit to them, but I think its double coverage once we start doing integration testing against our compliance profiles. 

We should re-investigate test-kitchen integration testing with a more well defined plan in the future.

miah (approved):

PR #4314, opened 2019-07-18
Add SHA512 password example to `shadow` resource

jerryaldrichiii:
## Description

This adds an example that checks for SHA512 password hashes in `/etc/shadow`.

Many thanks to @kclinden, @sam1el, and @zenspider for helping make this possible.

Screenshot:

Screen Shot 2019-07-17 at 20 04 36

Failing output:
```
Password hashes in /etc/shadow
     ×  should only contain SHA512 hashes
     Users without SHA512 hashes: bad_user, really_bad_user
```

Passing output:
```
Password hashes in /etc/shadow
     ✔  should only contain SHA512 hashes
```

## Related Issue

N/A

## Types of changes

## Checklist:

### DIFF
--- a/docs/resources/shadow.md.erb
+++ b/docs/resources/shadow.md.erb
@@ -89,7 +89,7 @@ Use `where` to match any of the supported [filter criteria](#filter_criteria). `
 
 Use `where` with [expect syntax](https://www.inspec.io/docs/reference/profiles/#should-vs-expect-syntax) to show all users (that aren't disabled or locked) without SHA512 hashed passwords.
 
-    # Find users with password entries `*`, `!`, or don't begin with `$6$`
+    # Users with password fields that are not *, !, or don't begin with $6$
     bad_users = inspec.shadow.where { password !~ /^[*!]$|^\$6\$.*/ }.users
 
     describe 'Password hashes in /etc/shadow' do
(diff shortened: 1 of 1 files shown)


jerryaldrichiii:
Good call @kclinden, changing to:

```
# Users with password fields that are not *, !, or don't begin with $6$
```

It is hard getting it to fit on one line 😅

kclinden:
In your snippet the message variable should be failure_message.

jerryaldrichiii:
I changed it to `message` so it would not word wrap when rendered. 

EDIT: Aaaaaaaand forgot to update in both places 🤕

kekaichinose (approved):
Looks good to me

miah (approved):
Makes me wonder if we should add something more generic to the resource for checking how the password is stored.

aaronlippold:
I agree. If you look at the channel discussion on the topic this is what I
was suggesting myself. I actually started to look for a ruby gem that could
return to me the algorithm used given a hash.

aaronlippold:
This is really what we were looking for because the control asks that all
passwords have a minimum of a Sha256 but if the end user decided to use
let's say 512 or 1024 when it comes out they would still meet the spirit of
the requirement. Are resource should be able to say that all passwords
strings are 256 or higher or at least be able to check am array of
acceptable shot types. We have an equivalent check in the SSHD config
resource where we check acceptable ciphers and then use the be_in matcher
to test for the requirement.

PR #4326, opened 2019-07-23
Sort control files before loading to allow results to be predictable

### LINKED ISSUE #3797: Make reporting consistent/predictable
(edited by @zenspider to strike out prescriptive aspects of this issue)

### 🎛 Description

🙋 feature request

~~Sort control files before loading them to allow results to be predictable (This is similar to pull requests [#3887](https://github.com/inspec/inspec/pull/3887))~~

Customer wants reports to be in a consistent order.

inspec version: **4.6.9**
Operating System: **RHEL 7**, **AIX 7/6**, **WIN2008/2012/2016**, **SOLARIS 11**

### 🤔 Replication Case

~~The loading order of control files is based on OS **find** command output, it is hard to organize large number of controls into control files, especially when initialing and closing actions are required.~~

### 💁 Possible Solutions

No:

    Add sorting statement as below, and testing had been done.

    ``` git diff
    diff --git a/lib/inspec/profile.rb b/lib/inspec/profile.rb
    index 6f6f3048..be66581a 100644
    --- a/lib/inspec/profile.rb
    +++ b/lib/inspec/profile.rb
    @@ -55,6 +55,7 @@ module Inspec
         def self.for_path(path, opts)
           file_provider = FileProvider.for_path(path)
           rp = file_provider.relative_provider
    +      rp.files.sort!

           # copy embedded dependencies into global cache
           copy_deps_into_cache(rp, opts) unless opts[:vendor_cache].nil?
    ```
    Note: Currently, we patch inspec for all OS mentioned above to perform various actions collectly.

mattray:
Fixes https://github.com/inspec/inspec/issues/3797

## Description
Sorts the control files for predictable output. This is a frequent request, just submitting a PR for the simplest solution.

## Related Issue
https://github.com/inspec/inspec/pull/3887
https://github.com/inspec/inspec/pull/3840

## Types of changes
Simply sorts the control files.

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -48,6 +48,7 @@ module Inspec
     def self.for_path(path, opts)
       file_provider = FileProvider.for_path(path)
       rp = file_provider.relative_provider
+      rp.files.sort!
 
       # copy embedded dependencies into global cache
       copy_deps_into_cache(rp, opts) unless opts[:vendor_cache].nil?


zenspider:
I'm against this at a philosophical level. Why is this needed?

clintoncwolfe (commented):
It seems straightforward, but my only concern is the lack of testing - not so much because we want to verify that it works, but because we want to discover how control file sorting works under other conditions - such as multiple profiles, on the same command line, inheritance, etc. Also, should this be documented?

clintoncwolfe:
My understanding is that people want to have setup and teardown files - aaa.rb to initialize, zzz.rb to cleanup, etc.

mattray:
I submitted the patch because the customer said they wanted consistent ordering of test results and this worked for them. If we document it as a feature, then we'll need to put a lot more effort into testing the scenarios @clintoncwolfe mentioned. Would it be acceptable to merge the patch and see if this is sufficient for the other issues that have been raised in the past and revisit it if it is not?

zenspider:
Consistent ordering of test results is (or can be) different from consistent ordering of test execution.

I think this issue will require some extra discussion. I might need recalibration.

miah:
It feels like this should be a reporter modification.

aaronlippold:
In the end I think we would sort in the reporter even if we sorted on the
input we would want to guarantee that the order was as expected coming out
of the reporter. That's the interface point right.

clintoncwolfe:
@mattray can you clarify - what is wanted here? predictable execution order, or predictable results ordering?

aaronlippold:
Hi, just my two cents. we've had this discussion before where people wanted
a consistent execution order but that was in regards to them trying to use
the output of a prior test in the input of a next test. This will and can
lead to very unexpected results and not something we should encourage. In
my opinion. Each test should be independent.

The original ask if I remember correctly was that they wanted consistent
ordered results output.

Walking down the other path is dangerous in my opinion and will lead to
many bad habbits.

zenspider:
After discussion and further consideration, I'm going to have to push back on this. It's dangerous (read: leads to false positives caused by test order dependencies) and not something we want to support.

> I submitted the patch because the customer said they wanted consistent ordering of test results

This is a perfect **_description_** of the problem at hand. This should have become an issue asking for exactly that. What this PR is is a **_prescription_** of what you want, but it is the wrong solution for the actual problem at hand (a report order issue, not a run order issue). Sorry to have to go this way, but I have Very Strong Feelings™ about testing (and put out the very first test framework that actually randomized order to maintain exactly this).

zenspider:
Also: Please do file that issue so we can get it on the board and you can track it.

mattray:
@zenspider was #3797 not sufficient?

zenspider:
@mattray I overlooked that. I would say that it is NOT sufficient for the reasons I explained above. Being descriptive as to what your problem is helps us find a solution that fits both the codebase, best practices, and the customer... this ticket is not that. It is prescriptive ("sort the controls to allow X"). I will poke at it.

Issue #4282, opened 2019-07-10
Update JSON Schema to more recent draft


Mitriol:
Current JSON schema uses draft v2, as evidenced by use of the "optional": true|false keyword which was deprecated by v3/v4. If we are to move forward with creating new, versioned reporters, I think it would be sensible to:

- Clarify which fields specifically are guaranteed to exist, and which are simply optional.
  - Currently, though many fields are marked "optional" => true/false, some required fields aren't specifically marked as such, and some fields marked optional are guaranteed to be there (so we might as well mark them as such).
  - Improving this would help to catch any breaking changes to reporters.
- Modernize the current reporter schema to a more recent draft of the json schema.
  - Later versions, such as v7 allow for more specific schema to accommodate specific output conditions (e.g. we need one of URI or URL on a ref).
  - Previous statements indicate that a migration to v4 was at one point planned: see https://github.com/inspec/inspec/issues/884#issuecomment-238202187
  - Unclear which version would be the best - might as well do the latest?
- Use an update schema as a baseline for future iterations of reporters as planned in #3905 

As far as what impact I see this as having:
- The reporters themselves will be unchanged. For now we would just be updating the schema to match what the reporters actually produce.
- Better marking of required data should improve our json-schema tests.
- Doesn't really cause any harm; existing code should by and large work exactly the same.

I've already done some work to this end, as modernizing the schema was necessary for me to use quicktype.io automatic type generation (it does not support pre-v4 schema). Would appreciate any thoughts on the matter!

zenspider:
As discussed in zoom, a PR bumping us to schema v4 and bumping our schemas to remain compatible would be welcome. Efforts to unify the schemas and make them more consistent should come after that since they will most likely require major versioning.

Schwad:
Sorry stale bot, I believe this is something we're still keeping an eye on, unless someone else knows something I don't.

kekaichinose:
@mitriol, going back to @zenspider's comment from earlier, it's been more than a year since this issue was opened, and we apologize that we have not been able to address this directly. The best path forward here would be to submit a PR to get the schema updated, we'll review it to make sure it doesn't break the user experience for other folks, and merge it after that.

PR #4330, opened 2019-07-24
(WIP) Revision of schema to JSON Schema draft 7

Mitriol:
The schema output by "inspec schema <name>" has been revised to be in the JSON Schema draft 7 format, providing better support for required attributes and self-documenting schema structures. As of right now this provides no functional change to what the schema _is_ (the reporters are completely untouched), simply how it is generated and output.

However, this work should hopefully be helpful to developers hoping to understand/automatically parse the Inspec output schema, and also to be a good foundation for any future revisions/versioning of the schema.

## Related Issue
See #4282 , #3706 

## Types of changes

## Checklist:

## Future work:
- [X] Update existing schema structure to draft 7
- [X] Update tests to validate new schema version (was already done)
- [X] Make tests validate the schema itself
- [ ] Add descriptions to all fields (where appropriate)
- [X] Verify that all required data is marked required (and, conversely, that no non-required data is marked required).
- [ ] Document the current schema generation process

And, optionally, the following:
- [ ] Add versioning component/parameter to the schema generation cli, to support any future breaking changes
- [ ] Unify schema types to avoid inconsistent data (e.g. the description labels in #3682).
- [ ] Decide how to version schema generation to accommodate future changes. Should json schema <version> simply fetch a cached output of schema <version> if it is not the newest version?

Signed-off-by: Jacob Henry <jchenry@mitre.org>


lib/inspec/schema/profile_json.rb
```diff
@@ -0,0 +1,58 @@
+require_relative "./primitives"
```
miah: `require "inspec/schema/primitives"`

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
+require "json"
+require_relative "./primitives"
+require_relative "./exec_json"
+require_relative "./exec_json_min"
+require_relative "./profile_json"
```
miah: Also here, I don't think we need require_relative as require will work fine for these.

lib/inspec/schema/profile_json.rb
```diff
@@ -0,0 +1,58 @@
...
+
+# We disable this for the sake of legibility
+# rubocop:disable Layout/AlignHash
```
miah: We should avoid adding new rubocop:disables to the code.

lib/inspec/schema/exec_json.rb
```diff
@@ -0,0 +1,73 @@
...
+        "code"              => STRING,
+        "source_location"   => SOURCE_LOCATION.ref,
+        "results"           => Schema.array(CONTROL_RESULT.ref),
+      },
+    }, [CONTROL_DESCRIPTION, REFERENCE, SOURCE_LOCATION, CONTROL_RESULT])
+
+    # Based loosely on https://www.inspec.io/docs/reference/profiles/ as of July 3, 2019
+    # However, concessions were made to the reality of current reporters, specifically
+    # with how description is omitted and version/inspec_version aren't as advertised online
+    PROFILE = SchemaType.new("Exec JSON Profile", {
```
miah: even if version/inspec_version aren't as advertised we should still include them.

miah (changes requested):
Thanks @Mitriol!

lib/inspec/schema/profile_json.rb
Mitriol: Done. Having them aligned was nice for working, but unnecessary for dev.

lib/inspec/schema/exec_json.rb
Mitriol: Done

lib/inspec/schema/schema.rb
Mitriol: Changed all requires to `require "lib/inspec/schema/__file__"`

lib/inspec/schema/exec_json.rb
```diff
@@ -1,4 +1,4 @@
-require_relative "./primitives"
+require "lib/inspec/schema/primitives"
```
zenspider: no, not "lib". lib is already in the load path (via rake/gemfile/etc)... this would only work if "." were in the load path as well, which it isn't by default.

lib/inspec/schema/exec_json.rb
```diff
@@ -1,66 +1,63 @@
...
-        "resource"      => STRING,
-        "message"       => STRING,
-        "skip_message"  => STRING,
-        "exception"     => STRING,
-        "backtrace"     => OBJECT,
+        "resource" => STRING,
+        "message" => STRING,
+        "skip_message" => STRING,
+        "exception" => STRING,
+        "backtrace" => OBJECT,
```
zenspider: Do we really have a whitespace linter rule turned on? @miah?

I will fight chefstyle on this one. Visual pattern matching is a good thing to the point that we're literally evolved for it. We shouldn't make it harder.

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
+require "json"
```
zenspider: this file should be named `lib/inspec/schema/output_schema.rb`, no?

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
...
+require "lib/inspec/schema/primitives"
+require "lib/inspec/schema/exec_json"
+require "lib/inspec/schema/exec_json_min"
+require "lib/inspec/schema/profile_json"
+
+module Inspec
+  module OutputSchema
```
zenspider: and should be namespaced to match.

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Add them all to the definition hash
+      defined_types.each do |type|
+        definitions[type.name] = type.body
+      end
+
+      # Nest under "definitions" label
+      {
+        "definitions" => definitions,
+      }
```
zenspider: All of the comments are redundant distractions. Each "paragraph" of code is separated but actually all related. And all boil down to something that's both succinct and easily comprehensible:

```ruby
{
  "definitions" => schema_type.all_depends.map { |t| [t.name, t.body] }.to_h
}
```

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
...
+      require "train"
+      Train.create("mock").connection
+      Train::Platforms.export
+    end
+
+    LIST = {
+      "profile-json" => OutputSchema.finalize(Schema::ProfileJson::PROFILE),
+      "exec-json" => OutputSchema.finalize(Schema::ExecJson::OUTPUT),
+      "exec-jsonmin" => OutputSchema.finalize(Schema::ExecJsonMin::OUTPUT),
+      "platforms" => PLATFORMS,
```
zenspider: why isn't the proc here?

lib/inspec/schema/schema.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    # Helper function to automatically bundle a type with its dependencies
+    def self.finalize(schema_type)
+      schema_type.body.merge(OutputSchema.build_definitions(schema_type))
+    end
+
+    # using a proc here so we can lazy load it when we need
+    # TODO: Modernize
+    PLATFORMS = lambda do
```
zenspider: Why the TODO? Modernize in what sense? Why isn't it just done the way you want?

lib/inspec/schema/profile_json.rb
```diff
@@ -0,0 +1,58 @@
+require "lib/inspec/schema/primitives"
+
+# These type occur only when running "inspec json <file>".
+
+module Schema
+  module ProfileJson
```
zenspider: namespacing should match pathing.

lib/inspec/schema/exec_json.rb
```diff
@@ -1,66 +1,63 @@
 require "lib/inspec/schema/primitives"
 
-# We disable this for the sake of legibility
-# rubocop:disable Layout/AlignHash
```
zenspider: Honestly, I could do without this commit as a whole. I think it looks/works better before.

THAT MIGHT JUST BE ME THO. I haven't had this fight with the ppl owning chefstyle (yet).

lib/inspec/schema/schema.rb
zenspider: Also, do we benefit from the lazy aspect at all? I can't tell when/where/how "pratforms" gets hit.

zenspider (changes requested):
Thank you for the PR. Some changes requested.

Also, I thought based on our discussion that you wanted to go to v4, not v7?

lib/inspec/schema/schema.rb
aaronlippold: I think a few of these things are historical, as we were trying to be as non-invasive as possible will still achieving our initial goal. We'll go through and remove any such comments of actions to be taken and replace them with any notes of things that still need to be done.

lib/inspec/schema/schema.rb
Mitriol: If I'm being fully honest I'd never written ruby until ~2 weeks ago, so still not entirely clear on some best-practice stuff. 

To be clear, are you saying it should be
`
namespace OutputSchema
    module OutputSchema
`

Apologies if I'm being obtuse here.

lib/inspec/schema/schema.rb
Mitriol: Forgive me but what is the proc?

RE: Lazy aspect: Platforms is unchanged from the old code - I figured there must have been a reason for it at the time, but I can remove it now if necessary.

Mitriol:
> Thank you for the PR. Some changes requested.
> 
> Also, I thought based on our discussion that you wanted to go to v4, not v7?

I stated that we should go to "at least" v4. It really comes down to if we have any specific tools in mind that only accept older schemas.

I could roll back to v4 - the only change that would be required would be to remove every  `"required" => []`, since in v4 a required array had to have at least one element (if nothing was required, it was simply omitted). All other changes added functionality that the schema, as currently written, does not utilize. So even though I wrote it as "v7" it is technically also v6, v5, and (with some small alterations) v4 compatible.

lib/inspec/schema/schema.rb
Mitriol: ^ I don't know what modernize means here either haha.

lib/inspec/schema/schema.rb
miah:
```
module Inspec
  module Schema
    module OutputSchema
    end
  end
end
```

lib/inspec/schema/exec_json.rb
miah: I think we should have the chefstyle discussion, but for now I'm not a fan of adding more `# rubocop:disable`. I'd like to remove all of them from the code and just let the style system manage it while we manage the style itself.

Mitriol: On the docs page for this styling rule: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/AlignHash

It seems that the configuration `EnforcedHashRocketStyle: table` would result in styling as I had it, that would be manage automatically by rubocop. I'll leave this commit as is for now though.

lib/inspec/schema/profile_json.rb
Mitriol: All rubocop disables are gone as of commit https://github.com/inspec/inspec/pull/4330/commits/e7cb80d927b114bcbb0a3627dba84805f2be3300.

However, to get this visual alignment back we could change chefstyle rule Layout/AlignHash to use EnforcedHashRocketStyle: table.

lib/inspec/schema/exec_json.rb
Mitriol: Resolved as of https://github.com/inspec/inspec/pull/4330/commits/6dd9b4d21f8db0c4d8180795c6059f6fedb08d31.

Seemed to be working either way but yes, this is nicer.

lib/inspec/schema/schema.rb
Mitriol: Resolved as of https://github.com/inspec/inspec/pull/4330/commits/a25eabb6d2b7289edc3fccd6771b17767e2fc306

lib/inspec/schema/schema.rb
Mitriol: Resolved as of https://github.com/inspec/inspec/pull/4330/commits/e7cb80d927b114bcbb0a3627dba84805f2be3300

lib/inspec/schema/profile_json.rb
Mitriol: Resolved as of https://github.com/inspec/inspec/pull/4330/commits/e7cb80d927b114bcbb0a3627dba84805f2be3300

lib/inspec/schema/schema.rb
Mitriol: Used your change in https://github.com/inspec/inspec/pull/4330/commits/7fe53fc1fc53ba04e5579e5542011d7d5f61bba1

Mitriol:
RE: Codeclimate, some of the found "Similar blocks of code found" are in tests I didn't even write. Unsure what to really do about this? Feels outside the scope of the PR.

Also, some of the TODO's are not mine (such as the "Modernize" one). I can just remove them if that is preferred?

miah:
Hi, We're going to merge this into a intermediary branch and handle any cleanup that needs to be done before merging to master.

Thank you @Mitriol for putting this together. Sorry its taken so long =)

PR #4334, opened 2019-07-27
Found more tests that we've probably never actually run.

zenspider:
*sigh*


clintoncwolfe (approved):
Thanks!

miah (approved):

PR #4337, opened 2019-07-29
Removed lib/inspec/utils/latest_version.rb and associated code.

zenspider:
There is no reason why our tests should be failing because we deploy a
new version. That test shouldn't even be hitting the network for real.

Besides, the functionality has no place in our codebase. It's
functionality provided via `gem outdated` so why did we re-implement
it? And we expect most people to be running via omnibus, so they're
pinned to a specific version in the first place. Even if they updated,
they still couldn't run it and that would be more confusing.


miah (approved):
:1st_place_medal:

clintoncwolfe (approved):

PR #4343, opened 2019-07-30
group resource: Modified DarwinGroup to collect all users properly on macos

zenspider:
I noticed that my profile targeting group "staff" on OSX wasn't
returning ME. That seemed wrong. And it turns out it is wrong. We were
not collecting any users whose primary group was that group. (almost
all regular users are in group staff and they're all missing).

I ALSO found out that our MockLoader-based testing is piss-poor and,
through much pain and suffering, figured out how to test the group
resource directly.

### DIFF
--- a/lib/inspec/resources/groups.rb
+++ b/lib/inspec/resources/groups.rb
@@ -164,22 +164,40 @@ module Inspec::Resources
   # OSX uses opendirectory for groups, so `/etc/group` may not be fully accurate
   # This uses `dscacheutil` to get the group info instead of `etc_group`
   class DarwinGroup < GroupInfo
-    def groups
-      group_info = inspec.command("dscacheutil -q group").stdout.split("\n\n")
+    def runmap(cmd, &blk)
+      hashmap(inspec.command(cmd).stdout.lines, &blk)
+    end
 
-      groups = []
+    def hashmap(enum, &blk)
+      enum.map(&blk).to_h
+    end
+
+    def groups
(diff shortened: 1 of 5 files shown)


miah (changes requested):
You'll need to `chefstyle -a` =)

zenspider:
And probably to convert everything over to command resources? This was sloppy and only tested locally. Not sure HOW to test this tbh.. BK doesn't do mac yet.

lib/inspec/resources/groups.rb
```diff
@@ -165,21 +165,38 @@ def groups
   # This uses `dscacheutil` to get the group info instead of `etc_group`
   class DarwinGroup < GroupInfo
     def groups
-      group_info = inspec.command("dscacheutil -q group").stdout.split("\n\n")
+      group_by_id = `dscl . -list /Groups PrimaryGroupID`.lines.map { |l| name, id = l.split; [id.to_i, name] }.to_h
```
clintoncwolfe: We can't use backticks - that will run locally. InSpec is supposed to run its commands on the scanning target, that's why it is running them through inspec.command.

Unless we somehow know that we are always local here, which seems a bad assumption?

clintoncwolfe (changes requested):
While it may seem unlikely someone would use a Mac machine as a remote scan target, we have to be careful to keep local execution separate from remote execution - so we need to check the backticks.

lib/inspec/resources/groups.rb
zenspider: done!

miah (approved):

clintoncwolfe (approved):

PR #4346, opened 2019-07-30
Add integration pipelines, and a schedule (it does nothing)

miah:


kekaichinose:
@miah so what's the schedule for?

miah:
The schedule will be used for nightly integration testing, but nothing is currently hooked up. I just want to get the configs in and make sure expeditor is happy here =)

kekaichinose:
Roger that.

kekaichinose (approved):
Looks good to me

PR #4348, opened 2019-07-30
Only bump versions when we modify InSpec.

miah:
In our current configuration, you must pro-actively label a Pull Request to prevent expeditor from bumping the version. 

This PR intends to change that behavior. Now version bumping will happen only when a file matching the specified list is modified.

The list I've used should match all files that end up affecting our artifacts. It doesn't include files like, dotfiles, expeditor/build configuration, or docs.

I mostly don't want to accidentally burn versions when I am quickly iterating on buildkite stuff. I think the other exclusions just make sense. Do we want to bump versions on InSpec when we make inconsequential changes outside the core code?

(docs will have its own pipeline to fire off the website updates)

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -38,7 +38,7 @@ pipelines:
      - LANG: "C.UTF-8"
      - SLOW: 1
  - www/build:
-    description: Gather documentation from related resource packs and build website.
+    description: Build website.
     definition: .expeditor/wwwbuild.yml
  - www/deploy:
     description: Deploy website to inspec.io
@@ -91,6 +91,20 @@ merge_actions:
     ignore_labels:
      - "Version: Skip Bump"
      - "Expeditor: Skip All"
+    only_if_modified:
+     - docs/*
+     - etc/*
+     - habitat/*
+     - inspec-bin/*
+     - lib/*
+     - omnibus/*
+     - support/*
+     - tasks/*
+     - test/*
+     - Gemfile*
+     - LICENSE
+     - "*.gemspec"
+     - "*.md"
  - bash:.expeditor/update_version.sh:
     only_if: built_in:bump_version
  - built_in:update_changelog:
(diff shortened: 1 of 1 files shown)


zenspider (approved):

miah:
I had to yamllint this, so there are some minor unrelated changes included.

.expeditor/config.yml
```diff
@@ -91,6 +91,18 @@ merge_actions:
...
+     - habitat/*
+     - inspec-bin/*
+     - lib/*
+     - omnibus/*
+     - support/*
+     - tasks/*
+     - test/*
+     - Gemfile*
+     - "*.gemspec"
+     - LICENSE
```
clintoncwolfe, line 105: There are a lot of individual files that arguably could go here - CODE_OF_CONDUCT.md, CONTRIBUTING.md. I definitely think README.md should be versioned.

clintoncwolfe (changes requested):
I love this overall, and am completely on-board for CI/CD changes.

I think docs changes should, however, result in a versioned artifact. Docs are part of our public contract with the world, and so should be versionable and subject to semver.

AFAIK this does not prevent docs from having an automated website publish process as part of the promotion pipeline - I'm totally all for that, and that should be a separate issue from whether a doc change is versioned.

.expeditor/config.yml
kekaichinose: +1 on the main meta files being versioned (e.g. CODE_OF_CONDUCT.md, CONTRIBUTING.md, README.md, etc.)

kekaichinose (changes requested):
Please add main project meta files to versioning (e.g. CODE_OF_CONDUCT.md, CONTRIBUTING.md, README.md)

miah:
Thanks for the feedback, I've made those changes.

clintoncwolfe (approved):
Thanks!

kekaichinose (approved):
Awesome sauce. Thanks!

tduffield (approved):

PR #4365, opened 2019-08-06
Make inspec much faster for most commands.

zenspider:
Removes pre-loading and registration of resources by adding
method_missing to a couple key areas.

This still drops the ball in some areas, but it is a start.


zenspider:
Still has one failure because the method_missing hack doesn't bother with aws stuffs... ignore for now. I'm pushing to see the effect on BK times.

lib/inspec/cli.rb
```diff
@@ -330,6 +323,7 @@ def detect
   option :distinct_exit, type: :boolean, default: true,
     desc: "Exit with code 100 if any tests fail, and 101 if any are skipped but none failed (default).  If disabled, exit 0 on skips and 1 for failures."
   def shell_func
+    require "inspec/resources" # TODO: remove me if possible!
```
clintoncwolfe: Since the shell is the REPL, if we were multi-threaded, we could be sneaky and load fast to get to a prompt, then do our heavy loading while waiting on the user to type... if threading was safe on RSpec...

lib/inspec/backend.rb
```diff
@@ -77,6 +77,14 @@ def self.create(config) # rubocop:disable Metrics/AbcSize
           connection
         end
 
+        def method_missing(id, *args, &blk)
```
clintoncwolfe, line 81: I feel that at this point we should be careful to heavily explain what is happening when we add new elements of metaprogramming. What context is this being called in? Explain it to most novice new developer on the team, but also to the potential contributor, in comments. Right now it is fresh in your mind and obvious; it will never be easier.

I can follow this one, but I have the tribal knowledge that requiring a resource has the side effect registering it in the registry.

lib/inspec/cli.rb
clintoncwolfe: I'm not sure we can - we definitely need it for tab completion and help.  If we had multithreading, we could get to a prompt early and then load the heavy stuff while the user types...

lib/inspec/control_eval_context.rb
```diff
@@ -66,9 +66,15 @@ def method_missing(method_name, *arguments, &block)
...
-            # If we couldn't find a plugin to match, maybe something up above has it,
-            # or maybe it is just a unknown method error.
-            super
+            begin
+              id = method_name
+              require "inspec/resources/#{id}"
+              klass = Inspec::Resource.registry[id.to_s]
+              klass.new(inspec, id, *arguments)
+            rescue LoadError
+              # TODO: aws
```
clintoncwolfe: If you're referring to the aws resources in lib/resources, those are deprecated and are scheduled for removal in 5.0 - so I'm not sure it's worth adding code to handle.

clintoncwolfe (commented):
I think we need to talk about this, as it is actually a breaking change - we currently promise people that Registry is fully populated at the start of a run, and you can cross-reference both core and custom resources that way. So dynamically loading resources (in this implementation) might break that. 

We could fix that by having Registry be more intelligent and perform 2-stage loading - we can glob lib/inspec/resources, which is far cheaper than the actual require.

I also request that we comment the addition - without it, it's just another sudden chunk of metaprogramming in the swamp.  Right now it has a clear purpose, easy to understand.

EDIT - After discussion, so few people use that undocumented feature that it is not worth preserving a full registry. My concerns are addressed.

lib/inspec/cli.rb
miah: I'm curious how many users use shell, autocomplete, or help =)

I think so long as we tell people we are going to break this functionality for a limited time while are making things faster this would be fine.

miah:
> We could fix that by having Registry be more intelligent and perform 2-stage loading - we can glob lib/inspec/resources, which is far cheaper than the actual require.

@clintoncwolfe I think we'd want to avoid globbing the resources. Right now we load and register _every_ resource when InSpec starts, whether the user uses it or not. I think globbing would end up with a very similar performance hit. I think phase two of this would be moving our `lib/inspec/resources.rb` to use 'autoload' instead of 'require'.

clintoncwolfe (changes requested):
I'm happy with the implementation but would request that we add comments explaining that this is auto-loading the resources, and perhaps a comment about requiring a resource resulting in it being registered.

lib/inspec/control_eval_context.rb
miah: If I understand the change correctly, we'll need code to handle AWS until we completely remove it otherwise those resources won't be loaded and users would be forced to use inspec-aws.

miah (changes requested):
I agree that we need a comment or documentation update.

It feels like this PR should also delete `inspec/resources.rb` since we're removing all references to it.

zenspider:
@miah I would agree, but for `inspec shell`'s `help commands`.

We could still delete that file and try to populate the registry in some other way but right now I'm kinda leaning towards keeping the file as a "manifest" or some-such. I should definitely drop a comment on the top of the file so that it makes that plainly clear at this point. Thanks.

zenspider:
@clintoncwolfe totally. I'll add some comments

I'm still not sure about having 3 different `method_missing` implementations but I don't see a clear refactoring right now. But... maybe a helper method they can all use for just the resource side?

zenspider:
IDGI... this was _supposed_ to be failing on missing aws resources. I thought we had a test that triggered that... I'm not getting that failure now.

lib/inspec/backend.rb
clintoncwolfe: Thanks!

clintoncwolfe (approved):
Looks great - thanks!

lib/inspec/dsl.rb
```diff
@@ -44,7 +63,11 @@ def method_missing(method_name, *arguments, &block)
       # We still haven't called it, so do so now.
       send(method_name, *arguments, &block)
     else
-      super
+      begin
+        Inspec::DSL.method_missing_resource(@backend, method_name, *arguments)
```
miah: Is `@backend` right to use here? Can we use the accessor added on line 6?

lib/inspec/resources/platform.rb
```diff
@@ -67,19 +69,22 @@ def supported?(supports)
...
-            status = name == v
-          elsif k == :release
-            status = check_release(v)
-          else
-            status = false
-          end
+      supports.each do |support|
+        support.each do |k, v|
+          status =
+            case k
```
miah: nice!

miah (changes requested):
Can you answer this question? https://github.com/inspec/inspec/pull/4365/files#r318319942

Approve after that =)

lib/inspec/dsl.rb
zenspider: I took a whack at using `backend` and it broke some tests... :/

lib/inspec/control_eval_context.rb
zenspider: I needed to handle it to make tests pass, so... done.

lib/inspec/cli.rb
zenspider: I worked this out too.

zenspider:
I believe all comments/questions have been addressed @miah... but your link doesn't work right (prolly due to my force pushes) so I'm not sure if I'm missing anything.

miah (approved):

PR #4366, opened 2019-08-06
Ensure that resources are always initialized in all cases

zenspider:
It does NOT do what you think it does. It does NOT prevent an object
from coming back. All it does is have an uninitialized (or partially
initialized) object come back, which isn't helpful.

### DIFF
--- a/lib/inspec/plugin/v1/plugin_types/resource.rb
+++ b/lib/inspec/plugin/v1/plugin_types/resource.rb
@@ -81,6 +81,7 @@ module Inspec
           @resource_skipped = false
           @resource_failed = false
           @supports = Inspec::Resource.supports[name]
+          @resource_exception_message = nil
 
           # attach the backend to this instance
           @__backend_runner__ = backend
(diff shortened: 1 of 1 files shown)


miah (approved):

miah (approved):
I wonder about side effects on this one though

clintoncwolfe (changes requested):
I think it's actually intentional behavior with a bad comment - the intent really is not to affect the return value, but to just avoid further initializiation.

If that entire initialization block that follows were placed under an `unless`, that might be clearer, and not involve a suspicious return-from-initialize...

```
# Stop initializing if we are on an unsupported or test backend
unless supported || test_backend
  begin
    super(*args)
  rescue...
    ...
  end
end
```

clintoncwolfe (changes requested):
(double commented, whoops)

zenspider:
It might be confusingly named, but Initialization _needs_ to complete for our resources to work correctly. Simple examples are `def to_s` using ivars to generate a string.

clintoncwolfe (approved):
If that's the intent - to always initialize resources - I'll accept, and I've updated the PR title for you to reflect the intent.

PR #4372, opened 2019-08-08
DO NOT MERGE YET: `let` blocks: Changes to speed up generated tests.

zenspider:
CIS generated tests can have a lot of variables that are actually very
resource heavy. This converts all variables into their equivalent let
form, deferring all computation to actual use. This means that
"skipped" controls will have roughly zero impact.

I think there is still more to be done here and on that "skipped"
control concept. Its name is misleading at best yet still potentially
very costly.

### DIFF
--- a/test/unit/dsl/objects_test.rb
+++ b/test/unit/dsl/objects_test.rb
@@ -10,7 +10,7 @@ describe "Objects" do
       obj.expectation = "2.4.2"
       _(obj.to_ruby).must_equal '
 subject { resource }
-describe "resource" do
+describe %q(resource) do
   it { subject.version.should cmp >= "2.4.2" }
 end
 '.strip
@@ -24,7 +24,7 @@ end
       obj.expectation = "2.4.2"
       _(obj.to_ruby).must_equal '
 subject { resource.version }
-describe "resource.version" do
+describe %q(resource.version) do
   it { subject.should cmp >= "2.4.2" }
 end
 '.strip
@@ -36,7 +36,7 @@ end
       obj.expectation = Regexp.new("^Desc.+$")
       _(obj.to_ruby).must_equal '
 subject { resource.to_s }
-describe "resource.to_s" do
+describe %q(resource.to_s) do
   it { subject.should cmp(/^Desc.+$/) }
 end
 '.strip
@@ -48,7 +48,7 @@ end
       obj.expectation = 3
       _(obj.to_ruby).must_equal '
 subject { resource.to_i }
-describe "resource.to_i" do
+describe %q(resource.to_i) do
   it { subject.should cmp > 3 }
 end
 '.strip
@@ -61,7 +61,7 @@ end
       obj.expectation = "mytest"
       _(obj.to_ruby).must_equal '
 subject { resource.name[2] }
-describe "resource.name[2]" do
+describe %q(resource.name[2]) do
   it { subject.should eq "mytest" }
 end
 '.strip
@@ -73,7 +73,7 @@ end
       obj.expectation = "mytest"
       _(obj.to_ruby).must_equal '
 subject { resource.hello("world") }
-describe "resource.hello(\"world\")" do
+describe %q(resource.hello("world")) do
   it { subject.should eq "mytest" }
 end
 '.strip
@@ -85,7 +85,7 @@ end
       obj.expectation = %w{mytest mytest2 mytest3}
       _(obj.to_ruby).must_equal '
 subject { resource.hello("world") }
-describe "resource.hello(\"world\")" do
+describe %q(resource.hello("world")) do
   it { subject.should be_in ["mytest", "mytest2", "mytest3"] }
 end
 '.strip
@@ -98,7 +98,7 @@ end
       obj.expectation = %w{mytest2 mytest3 mytest4}
       _(obj.to_ruby).must_equal '
 subject { resource.hello("world") }
-describe "resource.hello(\"world\")" do
+describe %q(resource.hello("world")) do
   it { subject.should_not be_in ["mytest2", "mytest3", "mytest4"] }
 end
 '.strip
@@ -110,7 +110,7 @@ end
       obj.expectation = %w{item1 item2 item3 item4 item5}
       _(obj.to_ruby).must_equal '
 subject { ["item1","item2","item3"] }
-describe "[\"item1\",\"item2\",\"item3\"]" do
+describe %q(["item1","item2","item3"]) do
   it { subject.should be_in ["item1", "item2", "item3", "item4", "item5"] }
 end
 '.strip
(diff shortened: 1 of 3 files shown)


james-stocks:
To be clear - where raw ruby variables are defined inside a control, those variables are still evaluated even if the control is skipped?
I did not know that.

One concern with this change is that we depend upon the `inspec/object` classes for test generation, and InSpec 1 is still supported. Therefore it breaks things for us if these classes start to generate Ruby that is not compatible in InSpec 1.51.
I think InSpec has had `subject` since forever though?

Side note: we have a private ticket to make our own InSpec code generation classes and stop using the `inspec/object` classes. We should maybe get that done so that we stop relying on classes that are internal to InSpec.

test/unit/dsl/objects_test.rb
```diff
@@ -396,17 +423,20 @@
...
+      # TODO: doubled up on subject... probably skippable w/ some logic
       control.to_ruby.must_equal '
 control "variable.control.id" do
   title "Variable Control Important Title"
   desc  "The most variable control the world has ever seen"
   impact 1.0
-  a = command("which grep")
-  describe a do
-    its("exit_status") { should eq 0 }
+  let(:a) { command("which grep") }
```
james-stocks, line 432: I like the use of `let`.

The way `subject` is used makes controls look quite different from the ideal examples you see on the front page of inspec.io.
I think customers might question why we have this `subject { a }` instead of just `describe a do`

clintoncwolfe (dismissed):
We need to have a Product discussion on this. This raises the bar on technical knowledge needed to understand a profile; we've aimed to have profiles readable where possible by non-technical auditors. 

Does this deepen dependency on RSpec? 

It may be worth it; but the conversation should be had.

clintoncwolfe:
@james-stocks here's what happens with skip_control - assuming this is a control that is being included, but skipped:
```
control 'control-01' do # This will not appear in the report
  puts 'This will execute'
  describe command('touch this-executed-too') do  # This line is a big surprise 
    puts 'This DID NOT execute'
    its('stdout') { should cmp 'will not execute' } 
  end
end
```

zenspider:
@james-stocks AFAIK, inspec has always had rspec as its core, so let will always be safe. We can rename it to `var` or `delay` or something to insulate as well. `let` isn't a big concept nor is it tied to rspec, so I don't think the name matters too much. It's super easy to implement (literally map it to `define_method` and you can be done).

@clintoncwolfe this is for generated code _only_, so I don't think it's really raising a bar. That and it maps 1:1 (+1 for the subject line, but I have ideas about that for simple subjects). 

@kekaichinose I need your eyeballs on this too.

test/unit/dsl/objects_test.rb
zenspider: Again, this is only for generated code.

zenspider:
> Does this deepen dependency on RSpec?

nah. Let can be as simple as `alias let define_method` if you really want.

james-stocks:
We should maybe zoom about this, I don't understand "only for generated code". For the context we use these classes for; the generated code should ideally end up, as @clintoncwolfe says, readable where possible by non-technical auditors.

I think `let` is actually a good word for self-describing what the code is doing at that point.

zenspider:
@james-stocks "only for generated code" is in reference to the fact that this PR is only in the part of inspec that is used by our CIS generator (whose name I _always_ forget). This doesn't address hand written profiles or anything else. Just our own internal tooling to do CIS generation. Happy to zoom tho. I sent you a DM to try to schedule something soonish.

kekaichinose:
@zenspider per our team discussion, what's say we update the title to the `lazy` moniker and get with @james-stocks to get some of the fuzziness cleared up?

clintoncwolfe:
In our team discussion, I'm happy with the change overall - it's a big win in terms of profile evaluation performance. We have to decide how deeply to document / advocate for using the the feature.

I also experimented a bit on a stock inspec 4.10.4, and it seemed that `let` is not implemented by RSpec (or is not otherwise exposed in the DSL). we need to verify we have round-trip backwards compatibatility - anything we generate must be compatible with 1.x in the short term, and 3.x in the long-term (or else we back-port and work with Superteam to switch to a more recent 3.x).

zenspider:
Specifically:

```
9992 % bundle exec inspec shell -c 'describe "blah" do let(:woot) { 42 }; it "should go" do expect(woot).to cmp 42; end end'

Profile: inspec-shell
Version: (not specified)
Target:  local://

  blah
     ✔  should go

9993 % bundle exec inspec shell -c 'control "blah" do let(:woot) { 42 }; it "should go" do expect(woot).to cmp 42; end end'

Profile: inspec-shell
Version: (not specified)
Target:  local://

  ×  blah: (eval):1
     ×  Control Source Code Error (eval):1 
     undefined method `let' for #<#<Class:0x00007f986f971d28>:0x00007f986f9bbb80>

```

kekaichinose (approved):
Looks good to me. This will be helpful to folks that actually want to skip controls :)

clintoncwolfe:
So, I'm holding off reviewing this because I think to understand better what happens when older inspecs try to evaluate code that has been generated by this. If we could have a zoom session where we cover some of that, it would be helpful.

james-stocks:
Took another look at this with @zenspider , my current thoughts are:

I like the use of `lets` but am less sure about the usage of `subject`. I think whatever we do here should be in consensus with what we teach people to do on inspec.io and learn.chef.io etc

We do need to be careful with our supported profiles, we should test these changes in the CI for the profiles where the generated code gets checked and executed with all of InSpec 1,2,3 and 4

clintoncwolfe (commented):
As we discussed, I think we need to go ahead and add a `let` DSL method so we can round trip generated profiles. Ideally we'd have a functional test to generate and then execute a profile, but that might be a lot of harness.

miah (approved):

james-stocks (changes requested):
Need to be very careful merging this.
I think the InSpec code this PR will generate does not work. See @zenspider's comment [here](https://github.com/inspec/inspec/pull/4372#issuecomment-521036653)

Currently `let` will only work inside a describe. Declaring it outside the describe (like this PR will do) does not work

```
inspec shell -c 'let(:woot) { 42 }; describe "blah" do it "should go" do expect(woot).to cmp 42; end end'
Traceback (most recent call last):
...
/Users/jamesstocks/code/bundles/cool-project/ruby/2.6.0/gems/inspec-4.12.0/lib/inspec/dsl.rb:47:in `method_missing': undefined method `let' for #<#<Class:0x00007fda7f6622c0>:0x00007fda7f6615c8> (NoMethodError)
```

I think Ryan said this may be as trivial as including rspec in the right DSL class.

zenspider:
Yeah... it's not that trivial... but I woke up this morning thinking that we could change the code generator itself to generate the fix, ensuring compatibility back to 1.0. Tee hee.

kekaichinose:
@zenspider are there some other changes we need to make to get this merged?

zenspider:
@kekaichinose yeah. this is blocked by `let` (not) in contexts. I have a separate issue on that.

zenspider:
This is mostly moot at this point since we're moving the code generations out.

PR #4324, opened 2019-07-23
Initial pass to clean up functional testing wrt exit codes and output.

zenspider:
This should get windows tests running again? Hopefully? They're
running SO slowly on my vagrant box that I'm just pushing for now.
They're clean on the mac side.

Signed-off-by: Ryan Davis <zenspider@chef.io>

PR #4387, opened 2019-08-13
inspec plugins: List system plugins and other UX improvements

### LINKED ISSUE #4248: InSpec plugin list returning nil class error when train-aws installed.
# Summary:
After installing the latest MacOSX 10.4 installation .dmg, I found that a profile doing a very generic scan of AWS is returning an error with train-aws installed. It looks like there's a version coming back as nil and not being handled, but it's hard to tell.

# Context and data:

```
Scotts-MacBook-Pro-2:aws_inspec_test svidmar$ inspec plugin list
 Plugin Name                   Version   Via     ApiVer
-------------------------------------------------------
Traceback (most recent call last):
    15: from /usr/local/bin/inspec:155:in `<main>'
    14: from /usr/local/bin/inspec:155:in `load'
    13: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-bin-4.6.4/bin/inspec:11:in `<top (required)>'
    12: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
    11: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
    10: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
     9: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
     8: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:238:in `block in subcommand'
     7: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:115:in `invoke'
     6: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
     5: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
     4: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
     3: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.6.4/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb:29:in `list'
     2: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.6.4/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb:29:in `each'
     1: from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.6.4/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb:30:in `block in list'
/opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.6.4/lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb:438:in `make_pretty_version': undefined method `version' for nil:NilClass (NoMethodError)
```

Normally, I'd expect the list of plugins to show train-aws as installed, with its version in a simple table output to the command line. 

Version, profile and control information:

```
$ inspec version; cat ../aws_inspec_test/aws_inspec_test/inspec.yml; cat ../aws_inspec_test/aws_inspec_test/controls/example.rb 
4.6.4
name: aws_inspec_test
title: InSpec Profile
maintainer: The Authors
copyright: The Authors
copyright_email: you@example.com
license: Apache-2.0
summary: An InSpec Compliance Profile
version: 0.1.1
inspec_version: '>= 2.2.7'
depends:
  - name: inspec-aws
    url: https://github.com/inspec/inspec-aws/archive/master.tar.gz
supports:
  platform: aws


describe aws_cloudtrail_trails do
  its('entries.count') { should cmp 0 }
end
```

### LINKED ISSUE #4277: `inspec plugin list` should include system plugins
## Detailed Description

InSpec has plugins that it relies on directly using its gemspec - such as train-azure, train-aws, etc) - let's call these **system plugins**. Currently they do not get listed when the user executes `inspec plugin list`, which is confusing and potentially disastrous when the user attempts to then install a believed-to-be-missing plugin.

## Context

Give clarity to users as to what is actually available.

## Possible Implementation

1. Look in InSpec's source root for the plugins/ directory (to get "bundled" plugins)
2. (and) Scan the dependencies of inspec's own gemspec (to get "system" plugins)

Filter options could be added to `inspec plugin list` to include or exclude types.

clintoncwolfe:
## Description

This PR improves the `inspec plugin list` command by adding a column indicating where the plugin came from - if it was a user plugin ("user"), a gem that InSpec depends on ("system"), or a gem that is in InSpec's own source code ("core").

Additionally, system plugins are now listed (Fixes #4277). This reduces customer confusion; it is almost never the right thing to try to re-install a system plugin.

Version number display was also found to be either broken (fixes #4248) or always displaying "(unknown)" for user installs because the installed version was simply not being recorded in the statefile. Now it is.

The listing UI has been updated to use the new updated InSpec::UI system, so the table output is prettier. Many error messages are reformatted slightly.

## Related Issue

#4277, #4248

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/filter.rb
+++ b/lib/inspec/plugin/v2/filter.rb
@@ -60,4 +60,14 @@ module Inspec::Plugin::V2
       end
     end
   end
+
+  module FilterPredicates
+    def train_plugin_name?(name)
+      name.to_s.start_with?("train-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
+    end
+
+    def inspec_plugin_name?(name)
+      name.to_s.start_with?("inspec-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
+    end
+  end
 end
--- a/lib/inspec/plugin/v2/filter.rb
+++ b/lib/inspec/plugin/v2/filter.rb
@@ -63,18 +63,26 @@ module Inspec::Plugin::V2
 
   module FilterPredicates
     def train_plugin_name?(name)
-      name.to_s.start_with?("train-") && name_allowed?(name)
+      valid_plugin_name?(name, :train)
     end
 
     def inspec_plugin_name?(name)
-      name.to_s.start_with?("inspec-") && name_allowed?(name)
+      valid_plugin_name?(name, :inspec)
     end
 
-    def valid_plugin_name?(name)
-      name.to_s.match(/^(inspec|train)-/) && name_allowed?(name)
-    end
+    def valid_plugin_name?(name, kind = :either)
+      # Must have a permitted prefix.
+      return false unless case kind
+      when :inspec
+        name.to_s.start_with?("inspec-")
+      when :train
+        name.to_s.start_with?("train-")
+      when :either
+        name.to_s.match(/^(inspec|train)-/)
+      else false
+      end # rubocop: disable Layout/EndAlignment
 
-    def name_allowed?(name)
+      # And must not be on the exclusion list.
       ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
     end
   end
--- a/lib/inspec/plugin/v2/filter.rb
+++ b/lib/inspec/plugin/v2/filter.rb
@@ -63,11 +63,19 @@ module Inspec::Plugin::V2
 
   module FilterPredicates
     def train_plugin_name?(name)
-      name.to_s.start_with?("train-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
+      name.to_s.start_with?("train-") && name_allowed?(name)
     end
 
     def inspec_plugin_name?(name)
-      name.to_s.start_with?("inspec-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
+      name.to_s.start_with?("inspec-") && name_allowed?(name)
+    end
+
+    def valid_plugin_name?(name)
+      name.to_s.match(/^(inspec|train)-/) && name_allowed?(name)
+    end
+
+    def name_allowed?(name)
+      ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
     end
   end
 end
--- a/lib/inspec/plugin/v2/filter.rb
+++ b/lib/inspec/plugin/v2/filter.rb
@@ -61,6 +61,11 @@ module Inspec::Plugin::V2
     end
   end
 
+  # To be a valid plugin name, the plugin must beign with either
+  # inspec- or train-, AND ALSO not be on the exclusion list.
+  # We maintain this exclusion list to avoid confusing users.
+  # For example, we want to have a real gem named inspec-test-fixture,
+  # but we don't want the users to see that.
   module FilterPredicates
     def train_plugin_name?(name)
       valid_plugin_name?(name, :train)
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -60,14 +60,15 @@ module Inspec::Plugin::V2
       # TODO: - check plugins.json for validity before trying anything that needs to modify it.
       validate_installation_opts(plugin_name, opts)
 
-      # TODO: change all of these to return installed spec/gem/thingy
       # TODO: return installed thingy
       if opts[:path]
         install_from_path(plugin_name, opts)
       elsif opts[:gem_file]
-        install_from_gem_file(plugin_name, opts)
+        gem_version = install_from_gem_file(plugin_name, opts)
+        opts[:version] = gem_version.to_s
       else
-        install_from_remote_gems(plugin_name, opts)
+        gem_version = install_from_remote_gems(plugin_name, opts)
+        opts[:version] = gem_version.to_s
       end
 
       update_plugin_config_file(plugin_name, opts.merge({ action: :install }))
@@ -88,9 +89,9 @@ module Inspec::Plugin::V2
 
       # TODO: Handle installing from a local file
       # TODO: Perform dependency checks to make sure the new solution is valid
-      install_from_remote_gems(plugin_name, opts)
+      gem_version = install_from_remote_gems(plugin_name, opts)
 
-      update_plugin_config_file(plugin_name, opts.merge({ action: :update }))
+      update_plugin_config_file(plugin_name, opts.merge({ action: :update, version: gem_version.to_s }))
     end
 
     # Uninstalls (removes) a plugin. Refers to plugin.json to determine if it
@@ -335,13 +336,15 @@ module Inspec::Plugin::V2
       # not obliged to during packaging.)
       # So, after each install, run a scan for all gem(specs) we manage, and copy in their gemspec file
       # into the exploded gem source area if absent.
-
       loader.list_managed_gems.each do |spec|
         path_inside_source = File.join(spec.gem_dir, "#{spec.name}.gemspec")
         unless File.exist?(path_inside_source)
           File.write(path_inside_source, spec.to_ruby)
         end
       end
+
+      # Locate the GemVersion for the new dependency and return it
+      solution.detect { |g| g.name == new_plugin_dependency.name }.version
     end
 
     #===================================================================#
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -365,7 +365,7 @@ module Inspec::Plugin::V2
       # excluding any that are path-or-core-based, excluding the gem to be removed
       plugin_deps_we_still_must_satisfy = registry.plugin_statuses
       plugin_deps_we_still_must_satisfy = plugin_deps_we_still_must_satisfy.select do |status|
-        status.installation_type == :gem && status.name != plugin_name_to_be_removed.to_sym
+        status.installation_type == :user_gem && status.name != plugin_name_to_be_removed.to_sym
       end
       plugin_deps_we_still_must_satisfy = plugin_deps_we_still_must_satisfy.map do |status|
         constraint = status.version || "> 0"
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -54,7 +54,7 @@ module Inspec::Plugin::V2
         begin
           # We could use require, but under testing, we need to repeatedly reload the same
           # plugin.  However, gems only work with require (rubygems dooes not overload `load`)
-          if plugin_details.installation_type == :gem
+          if plugin_details.installation_type == :user_gem
             activate_managed_gems_for_plugin(plugin_name)
             require plugin_details.entry_point
           else
@@ -247,9 +247,9 @@ module Inspec::Plugin::V2
         status = Inspec::Plugin::V2::Status.new
         status.name = plugin_entry[:name]
         status.loaded = false
-        status.installation_type = (plugin_entry[:installation_type] || :gem)
+        status.installation_type = (plugin_entry[:installation_type] || :user_gem)
         case status.installation_type
-        when :gem
+        when :user_gem
           status.entry_point = status.name.to_s
           status.version = plugin_entry[:version]
         when :path
@@ -266,7 +266,7 @@ module Inspec::Plugin::V2
 
     def fixup_train_plugin_status(status)
       status.api_generation = :'train-1'
-      if status.installation_type == :gem
+      if status.installation_type == :user_gem
         # Activate the gem. This allows train to 'require' the gem later.
         activate_managed_gems_for_plugin(status.entry_point)
       end
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -37,7 +37,6 @@ module Inspec::Plugin::V2
 
       # Identify plugins that inspec is co-installed with
       detect_system_plugins unless options[:omit_sys_plugins]
-
     end
 
     def load_all
@@ -274,7 +273,7 @@ module Inspec::Plugin::V2
 
     def detect_system_plugins
       # Find the gemspec for inspec
-      inspec_gemspec = Gem::Specification.find_by_name('inspec', '=' + Inspec::VERSION)
+      inspec_gemspec = Gem::Specification.find_by_name("inspec", "=" + Inspec::VERSION)
 
       # Make a RequestSet that represents the dependencies of inspec
       inspec_deps_request_set = Gem::RequestSet.new(*inspec_gemspec.dependencies)
@@ -286,7 +285,7 @@ module Inspec::Plugin::V2
 
       inspec_gemspec.dependencies.each do |inspec_dep|
         next unless inspec_plugin_name?(inspec_dep.name) || train_plugin_name?(inspec_dep.name)
-        plugin_spec = runtime_solution.detect {|s| s.name == inspec_dep.name }.spec
+        plugin_spec = runtime_solution.detect { |s| s.name == inspec_dep.name }.spec
 
         status = Inspec::Plugin::V2::Status.new
         status.name = inspec_dep.name
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -274,7 +274,7 @@ module Inspec::Plugin::V2
 
     def detect_system_plugins
       # Find the gemspec for inspec
-      inspec_gemspec = Gem::Specification.find_by_name("inspec", "=" + Inspec::VERSION)
+      inspec_gemspec = Gem::Specification.find_by_name("inspec", "=#{Inspec::VERSION}")
 
       # Make a RequestSet that represents the dependencies of inspec
       inspec_deps_request_set = Gem::RequestSet.new(*inspec_gemspec.dependencies)
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -35,9 +35,9 @@ module Inspec::Plugin::V2
       # and may be safely loaded
       detect_core_plugins unless options[:omit_core_plugins]
 
-      # TODO: system plugins
-      # Load train plugin gem deps from inspec gem dependencies?
-      # Scan gem universe for all inspec- or train- gems?
+      # Identify plugins that inspec is co-installed with
+      detect_system_plugins unless options[:omit_sys_plugins]
+
     end
 
     def load_all
@@ -269,6 +269,41 @@ module Inspec::Plugin::V2
       if status.installation_type == :user_gem
         # Activate the gem. This allows train to 'require' the gem later.
         activate_managed_gems_for_plugin(status.entry_point)
+      end
+    end
+
+    def detect_system_plugins
+      # Find the gemspec for inspec
+      inspec_gemspec = Gem::Specification.find_by_name('inspec', '=' + Inspec::VERSION)
+
+      # Make a RequestSet that represents the dependencies of inspec
+      inspec_deps_request_set = Gem::RequestSet.new(*inspec_gemspec.dependencies)
+      inspec_deps_request_set.remote = false
+
+      # Resolve the request against the installed gem universe
+      gem_resolver = Gem::Resolver::CurrentSet.new
+      runtime_solution = inspec_deps_request_set.resolve(gem_resolver)
+
+      inspec_gemspec.dependencies.each do |inspec_dep|
+        next unless inspec_plugin_name?(inspec_dep.name) || train_plugin_name?(inspec_dep.name)
+        plugin_spec = runtime_solution.detect {|s| s.name == inspec_dep.name }.spec
+
+        status = Inspec::Plugin::V2::Status.new
+        status.name = inspec_dep.name
+        status.entry_point = inspec_dep.name # gem-based, just 'require' the name
+        status.version = plugin_spec.version.to_s
+        status.loaded = false
+        status.installation_type = :system_gem
+
+        if train_plugin_name?(status[:name])
+          # Train plugins are not true InSpec plugins; we need to decorate them a
+          # bit more to integrate them.
+          fixup_train_plugin_status(status)
+        else
+          status.api_generation = 2
+        end
+
+        registry[status.name.to_sym] = status
       end
     end
   end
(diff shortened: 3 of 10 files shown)


lib/inspec/plugin/v2/filter.rb
```diff
@@ -60,4 +60,14 @@ def validate_plugin_filter_file(_file_version)
...
     end
   end
+
+  module FilterPredicates
+    def train_plugin_name?(name)
+      name.to_s.start_with?("train-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
+    end
+
+    def inspec_plugin_name?(name)
+      name.to_s.start_with?("inspec-") && ! Inspec::Plugin::V2::PluginFilter.exclude?(name)
```
zenspider: These should be refactored together w/ a prefix arg. They're the same except for data.

lib/inspec/plugin/v2/loader.rb
```diff
@@ -260,8 +256,20 @@ def read_conf_file_into_registry
...
+        # Train plugins are not true InSpec plugins; we need to decorate them a
+        # bit more to integrate them.
+        fixup_train_plugin_status(status) if train_plugin_name?(plugin_entry[:name])
+
         registry[status.name] = status
       end
     end
+
+    def fixup_train_plugin_status(status)
+      status.api_generation = :'train-1'
```
zenspider, line 268: Is this value arbitrary? Why the awkward symbol? `:train_1` or "train-1" would be fine too, no?

lib/inspec/plugin/v2/loader.rb
```diff
@@ -133,7 +141,7 @@ def list_managed_gems
     # This is simply all gems that begin with train- or inspec-.
     # @return [Array[Gem::Specification]] Specs of all gems found.
     def self.list_installed_plugin_gems
-      list_managed_gems.select { |spec| spec.name.match(/^(inspec|train)-/) }
+      list_managed_gems.select { |spec| inspec_plugin_name?(spec.name) || train_plugin_name?(spec.name) }
```
zenspider: This doesn't seem like an equivalent refactoring (there's now an extra exclude check) and it went from 1 relatively cheap method call to potentially 6. I don't see these being used anywhere else so I'm not sure I see the point?

lib/inspec/plugin/v2/loader.rb
```diff
@@ -234,18 +242,6 @@ def detect_core_plugins
...
-        status = registry[train_plugin_name.to_sym]
-        status.api_generation = :'train-1'
-
-        if status.installation_type == :gem
-          # Activate the gem. This allows train to 'require' the gem later.
-          activate_managed_gems_for_plugin(train_plugin_name)
-        end
-      end
-    end
-
```
zenspider, line 248: YAY

lib/inspec/plugin/v2/loader.rb
```diff
@@ -11,9 +12,16 @@ module Inspec::Plugin::V2
   class Loader
     attr_reader :conf_file, :registry, :options
 
+    # For {inspec|train}_plugin_name?
+    include Inspec::Plugin::V2::FilterPredicates
+    extend Inspec::Plugin::V2::FilterPredicates
```
zenspider, line 17: Why? I don't see this being used at the class level (in this commit).

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -77,29 +77,29 @@ class PluginManagerCliList < Minitest::Test
   include PluginManagerHelpers
 
   def test_list_when_no_user_plugins_installed
+    skip_windows!
     result = run_inspec_process_with_this_plugin("plugin list")
 
-    skip_windows!
```
zenspider, line 82: this was here for a reason

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -77,29 +77,29 @@ class PluginManagerCliList < Minitest::Test
...
-
+    result = run_inspec_process_with_this_plugin("plugin list --all")
+    plugin_lines = result.stdout.lines.filter { |line| line.match(/^\s*(inspec|train)-/) }
+
+    # Look for a specific plugin of each type - core, bundle, and system
+    [
+      { name: "inspec-plugin-manager-cli", type: "core" },
+      { name: "inspec-supermarket", type: "bundle" },
+      { name: "train-aws", type: "gem (system)" },
+    ].each do |test_case|
```
zenspider: This seems unnecessarily complicated:

`[["blahblah", "blah"], ...].each do |name, type| ... end`

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -77,29 +77,29 @@ class PluginManagerCliList < Minitest::Test
...
+      plugin_line = plugin_lines.detect { |line| line.include? test_case[:name] }
+      refute_nil plugin_line, "#{test_case[:name]} should be detected in plugin list --all output"
+      assert_includes plugin_line, test_case[:type], "#{test_case[:name]} should be detected as a '#{test_case[:type]}' type in list --all "
```
zenspider: and thus has a bunch of redundancies here

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -478,6 +478,16 @@ def make_pretty_version(status)
...
       end
+
+      def make_pretty_install_type(status)
+        {
+          bundle: "bundle", # We could call this core, too - not much of a distinction
+          core: "core",
+          path: "path",
+          gem: "gem (user)",
+          system: "gem (system)",
+        }[status.installation_type]
```
zenspider: Should extract the static hash out to a frozen constant (unless you anticipate it going dynamic later (tho that might be a merge of the constant + dynamic))

lib/inspec/plugin/v2/loader.rb
```diff
@@ -271,5 +271,40 @@ def fixup_train_plugin_status(status)
         activate_managed_gems_for_plugin(status.entry_point)
       end
     end
+
+    def detect_system_plugins
+      # Find the gemspec for inspec
+      inspec_gemspec = Gem::Specification.find_by_name('inspec', '=' + Inspec::VERSION)
```
zenspider: `string + string` => `"string#{string}"`

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -467,13 +467,20 @@ def make_pretty_version(status)
...
+          if status.version.nil?
+            '(unknown)'
+          elsif status.version =~ /^\d+\.\d+\.\d+$/
+            status.version
+          else
+            # Assume it is a version constraint string and try to resolve
+            # TODO: this is naive, and assumes the latest version is the one that will be used. Logged on #3317
+            # In fact, the logic to determine "what version would be used" belongs in the Loader.
+            Inspec::Plugin::V2::Loader.list_installed_plugin_gems
+              .select { |spec| spec.name == status.name.to_s }
```
zenspider: do we know that status.name is a symbol (or other non-string)? It should be extracted to a lvar and stringified outside of an enumerator.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -20,17 +20,18 @@ def list
...
-        ui.bold(format(" %-30s%-10s%-8s%-6s", "Plugin Name", "Version", "Via", "ApiVer"))
-        ui.line
-        plugin_statuses.sort_by(&:name).each do |status|
-          ui.plain(format(" %-30s%-10s%-8s%-6s", status.name,
-                          make_pretty_version(status),
-                          make_pretty_install_type(status),
-                          status.api_generation.to_s))
+        ui.table do |t|
+          t.header = ["Plugin Name", "Version", "Via", "ApiVer"]
+          plugin_statuses.map { |s| [s.name.to_s, s] }.sort { |sa, sb| sa[0] <=> sb[0] }.map { |sf| sf[1] }.each do |status|
```
zenspider: One `sort_by` vs `map`+`sort`+`map`? Not sure I understand this change:
```
plugin_statuses.map { |s| [s.name.to_s, s] }.sort { |sa, sb| sa[0] <=> sb[0] }.map { |sf| sf[1] }.each do |status|
```

vs

```
plugin_statuses.sort_by { |s| s.name.to_s }.each do |status|
```

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -96,13 +99,13 @@ class PluginManagerCliList < Minitest::Test
...
-    assert_includes result.stdout, "0 plugin(s) total", "Empty list should include zero count"
-
-    assert_exit_code 0, result
+  # Listing all is now default behavior
+  def list_cases
+    [
+      { arg: "-c", name: "inspec-plugin-manager-cli", type: "core" },
+      { arg: "-c", name: "inspec-supermarket", type: "core" },
+      { arg: "-s", name: "train-aws", type: "gem (system)" },
+    ]
```
zenspider: extract to const

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -113,64 +116,80 @@ def test_list_all_when_no_user_plugins_installed
...
+    skip_windows!
+
     pre_block = Proc.new do |plugin_statefile_data, tmp_dir|
       plugin_statefile_data.clear # Signal not to write a file, we'll provide one.
       copy_in_core_config_dir("train-test-fixture", tmp_dir)
     end
 
-    result = run_inspec_process_with_this_plugin("plugin list", pre_run: pre_block)
-
-    skip_windows!
```
zenspider, line 139: ditto

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -498,22 +512,22 @@ def test_fail_install_from_nonexistant_gemfile
   end
 
   def test_install_from_rubygems_latest
+    skip_windows!
+
     install_result = run_inspec_process_with_this_plugin("plugin install inspec-test-fixture", post_run: list_after_run)
 
     success_message = install_result.stdout.split("\n").grep(/installed/).last
-    skip_windows!
```
zenspider, line 482: ...

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -74,15 +74,15 @@ def search(search_term) # rubocop: disable Metrics/AbcSize
         end
 
         puts
-        ui.bold(format(" %-30s%-50s", "Plugin Name", "Versions Available"))
+        ui.bold(format(" %-30s%-50s\n", "Plugin Name", "Versions Available"))
```
zenspider, line 85: Isn't that what `ui.line` does?

zenspider (changes requested):

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
clintoncwolfe: I prefer my data structures to have human meaningful labels.

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
clintoncwolfe: Yes, that's right. It's aiming for clarity, not brevity. If one assertion fails, you know exactly what to go fix. 

I believe in tests as useful diagnostic tools; I understand you disagree, but this is still valid.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
clintoncwolfe: No, `ui.line` makes a horizontal line. `ui.plain_line` does a `puts`, basically.

lib/inspec/plugin/v2/loader.rb
clintoncwolfe: I'm not particularly concerned about cost (even a heavy user would only have perhaps tens of managed gems), but I am very concerned about accuracy.  The previous implementation was flawed in that it ignored the exclusion list.

Here we are checking the exclusion list twice - once for inspec and once for train - so a potential refactoring would be to have a `any_plugin_name?` method, and call that once.

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
clintoncwolfe: Are you saying you wanted it after the test execution but before the assertions for a reason, which you are choosing not to share?

To me, it seemed to be a misplacement (note that I moved it above the subprocess spawn) - `run_inspec_process_with_this_plugin` is very expensive (it is spawning a subprocess). If you're not going to run the assertions, why run the subprocess? You may have some good reason, I'd love to learn it if you'd share it.

lib/inspec/plugin/v2/loader.rb
clintoncwolfe: ~~You're right - I was using it at one point, but not in current committed code. ~~

It is used at lib/inspec/plugin/v2/loader.rb:144

lib/inspec/plugin/v2/filter.rb
clintoncwolfe: (edited)

So, I had dug in my heels not to refactor this, feeling it would only make the code longer with no benefit other than deduplication of one-liners. I then slept on it, and went ahead and refactored it, but the refactor (at least the one I wrote) is indeed much bigger than the original. It is deduplicated though.

lib/inspec/plugin/v2/loader.rb
clintoncwolfe: I've added a `valid_plugin_name?` method that handles the exclusion list check just once.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
clintoncwolfe: Ahh, of course you can send a block to sort_by. Thank you. Needing to call a chain of methods had me doing the Schwarzian transform on it. Thanks for the suggestion.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
clintoncwolfe: Here is my interpretation of what you said: "I'm afraid that converting the `status.name` to a string repeatedly might be very expensive, so it might be best to do it once and store it in a variable. That way the comparision conditional in the select's block doesn't have to keep converting it."

I agree, that's wasteful, and my eye didn't notice it. Thanks!

lib/inspec/plugin/v2/loader.rb
clintoncwolfe: Well - and this is a matter of taste - as a lookup value, Symbols make more sense to me than Strings; and visually, dashes are more apealling than underscores. It's harder to type a symbol with a dash, but once it's in, it gets happily displayed many times, but typed only once.  

I looked in the inspec and train plugin docs; we are not committed to any one value of this - but not a numeric value, as that would collide with the InSpec API versioning.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
zenspider: `sort_by` _is_ the Schwartzian transform pattern.

that and `&:method_name` is `:method_name.to_proc` is `{ |obj| obj.method_name }`

miah (approved):
With bad Internet I was able to review this a few times =) I think this looks good. Thanks @clintoncwolfe!

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
clintoncwolfe: In conversation you pointed me to https://github.com/inspec/inspec/pull/4324/commits/917a531a921fbaa9dd83b154621de09cdc5097ad, which indicated you wanted the tests to always run, though not yet the assertions. I've pushed up a commit moving the skips back to their original locations.

lib/inspec/plugin/v2/loader.rb
zenspider: > visually, dashes are more apealling than underscores

to you... but underscores and non-quoted symbols are more idiomatic ruby.

lib/inspec/plugin/v2/filter.rb
```diff
@@ -63,18 +63,26 @@ def validate_plugin_filter_file(_file_version)
...
+      when :inspec
+        name.to_s.start_with?("inspec-")
+      when :train
+        name.to_s.start_with?("train-")
+      when :either
+        name.to_s.match(/^(inspec|train)-/)
+      else false
+      end # rubocop: disable Layout/EndAlignment
+
+      # And must not be on the exclusion list.
```
zenspider, line 90: This surprises me. The original code was `list_managed_gems.select { |spec| spec.name.match(/^(inspec|train)-/) }` and flogged at 4.7 and this refactoring has bloomed into a _lot_ more code now flogging at 24.0. The original intent, as I understand it, was to add 1 extra check (the exclusion list check) and make it so you can check if the name starts with `train-` in one extra spot. Am I not grokking something?

lib/inspec/plugin/v2/installer.rb
```diff
@@ -335,13 +336,15 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
...
-
       loader.list_managed_gems.each do |spec|
         path_inside_source = File.join(spec.gem_dir, "#{spec.name}.gemspec")
         unless File.exist?(path_inside_source)
           File.write(path_inside_source, spec.to_ruby)
         end
       end
+
+      # Locate the GemVersion for the new dependency and return it
+      solution.detect { |g| g.name == new_plugin_dependency.name }.version
```
zenspider, line 347: I would prefer if we move away from the `*-ect`s. They're not actually descriptive (except reject).

lib/inspec/plugin/v2/filter.rb
zenspider: tests? I don't see any for this.

lib/plugins/inspec-plugin-manager-cli/test/functional/inspec-plugin_test.rb
```diff
@@ -76,31 +99,41 @@ class PluginManagerCliList < Minitest::Test
...
-    assert_exit_code 0, result
+    plugins_seen = parse_plugin_list_lines(result.stdout)
 
-    # TODO: split
-    result = run_inspec_process_with_this_plugin("plugin list -a")
+    # Look for a specific plugin of each type - core, bundle, and system
+    LIST_CASES.each do |test_case|
+      plugin_line = plugins_seen.detect { |plugin| plugin[:name] == test_case[:name] }
+      refute_nil plugin_line, "#{test_case[:name]} should be detected in plugin list --all output"
+      assert_equal plugin_line[:type], test_case[:type], "#{test_case[:name]} should be detected as a '#{test_case[:type]}' type in list --all "
```
zenspider: assert_equal expected, actual

lib/inspec/plugin/v2/filter.rb
zenspider: This is what I think is more in line with the original intent / refactoring:

```ruby
  module FilterPredicates
    def valid_plugin_name?(name)
      name =~ /^(inspec|train)-/
    end

    def train_plugin_name?(name)  # even this seems unnecessary to me, but it gives a nice and consistent name.
      name =~ /^train-/
    end

    def excluded_plugin?(name)
      Inspec::Plugin::V2::PluginFilter.exclude?(name)
    end
  end
```

zenspider (dismissed):
I still have some strong reservations about filter.rb as a whole, the refactoring(s) involved, and the complexity this adds for what sounds like a simple PR/improvement (add source of gem to plugin list output)... That might be out of scope for this PR tho.

At 23 commits now and 13 files changed, I would say that this PR has outgrown the original description.

I'll let you be the judge of that.

lib/inspec/plugin/v2/installer.rb
clintoncwolfe: I don't see a corrective suggestion here, and I don't personally see anything wrong with detect, so I'm leaving this as-is.

lib/inspec/plugin/v2/filter.rb
clintoncwolfe: I've made a tech debt ticket to pursue further simplification, #4426 . However, the required implementation is different than that posted above - each case must check both the text name and the exclusion file.

PR #4388, opened 2019-08-13
Expand sys_info resource functionality

Vancelot11:
Signed-off-by: Kerry Vance <vancelot@osuosl.org>

## Description
Expand functionality of sys_info resource to allow different arguments to hostname function. Default behavior remains the same, so nothing should break. I was provisioning a server that used the full hostname in some files. The default behavior of 'hostname' didn't match the value used in the recipe. The value was dynamic not static, so could not be tested literally.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/sys_info.rb
+++ b/lib/inspec/resources/sys_info.rb
@@ -15,12 +15,12 @@ module Inspec::Resources
       end
 
       describe sys_info do
-        its('hostname') { should eq 'example.com' }
+        its('fqdn') { should eq 'user.example.com' }
       end
 
     EXAMPLE
 
-    %w{ alias domain fqdn ip-address short }.each do |opt|
+    %w{ domain fqdn ip-address short }.each do |opt|
       define_method(opt.to_sym) do
         hostname(opt)
       end
@@ -33,12 +33,10 @@ module Inspec::Resources
         opt = case opt
               when "f", "long", "fqdn", "full"
                 " -f"
-              when "a", "alias"
-                " -a"
               when "d", "domain"
                 " -d"
               when "i", "ip-address"
-                " -i"
+                " -I"
               when "s", "short"
                 " -s"
               else
--- a/lib/inspec/resources/sys_info.rb
+++ b/lib/inspec/resources/sys_info.rb
@@ -13,13 +13,38 @@ module Inspec::Resources
       describe sys_info do
         its('hostname') { should eq 'example.com' }
       end
+
+      describe sys_info do
+        its('hostname') { should eq 'example.com' }
+      end
+
     EXAMPLE
 
+    %w{ alias domain fqdn ip-address short }.each do |opt|
+      define_method(opt.to_sym) do
+        hostname(opt)
+      end
+    end
+
     # returns the hostname of the local system
-    def hostname
+    def hostname(opt = nil)
       os = inspec.os
       if os.linux? || os.darwin?
-        inspec.command("hostname").stdout.chomp
+        opt = case opt
+              when "f", "long", "fqdn", "full"
+                " -f"
+              when "a", "alias"
+                " -a"
+              when "d", "domain"
+                " -d"
+              when "i", "ip-address"
+                " -i"
+              when "s", "short"
+                " -s"
+              else
+                nil
+              end
+        inspec.command("hostname#{opt}").stdout.chomp
       elsif os.windows?
         inspec.powershell("$env:computername").stdout.chomp
       else
--- a/lib/inspec/resources/sys_info.rb
+++ b/lib/inspec/resources/sys_info.rb
@@ -20,7 +20,7 @@ module Inspec::Resources
 
     EXAMPLE
 
-    %w{ domain fqdn ip-address short }.each do |opt|
+    %w{ domain fqdn ip_address short }.each do |opt|
       define_method(opt.to_sym) do
         hostname(opt)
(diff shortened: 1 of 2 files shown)


miah (approved):
Seems reasonable to me, thanks @Vancelot11!

lib/inspec/resources/sys_info.rb
```diff
@@ -13,13 +13,38 @@ class System < Inspec.resource(1)
       describe sys_info do
         its('hostname') { should eq 'example.com' }
       end
+
+      describe sys_info do
+        its('hostname') { should eq 'example.com' }
```
clintoncwolfe: It looks like this example block is identical to the previous one - did you mean to edit it to include an option to `hostname`?

clintoncwolfe: Please add entries to `docs/resources/sys_info.md.erb` explaining the new options and listing the choices that are recognized.

clintoncwolfe (changes requested):
Great contribution! I'm not sure anyone would ever find this without listing it in the docs, though - could you please update the documentation? Thanks!

lib/inspec/resources/sys_info.rb
Vancelot11: Oops yeah meant to update that.

Vancelot11:
Upon closer inspection of the man pages, `alias` is deprecated in hostname so I removed it.

zenspider (approved):
This looks good. Thank you!

lib/inspec/resources/sys_info.rb
```diff
@@ -13,13 +13,36 @@ class System < Inspec.resource(1)
...
+      define_method(opt.to_sym) do
+        hostname(opt)
+      end
+    end
+
     # returns the hostname of the local system
-    def hostname
+    def hostname(opt = nil)
       os = inspec.os
       if os.linux? || os.darwin?
```
clintoncwolfe: Hrm, this is not portable to Darwin (MacOS):
```
hostname: illegal option -- I
usage: hostname [-fs] [name-of-host]
```

Neither -d nor -I works there - I think we either need a no-support exception, or an alternate implementation.

lib/inspec/resources/sys_info.rb
```diff
@@ -13,13 +13,36 @@ class System < Inspec.resource(1)
...
-        inspec.command("hostname").stdout.chomp
+        opt = case opt
+              when "f", "long", "fqdn", "full"
+                " -f"
+              when "d", "domain"
+                " -d"
+              when "i", "ip-address"
```
clintoncwolfe: One thought - it might be better to have this as ip_address. If you're using this to lookup values (as opposed to validate) you need to be able to outright call the method, like this:
```
my_host = sys_info.fqdn
my_ip = sys_info.ip-address # Would need to send a symbol)
```

clintoncwolfe (changes requested):
The docs are great! But I tried it on my mac and it has some portability issues that silently fail - it needs to either loudly fail, or just work.  Thanks!

If you don't have access to MacOS, let us know and we'll adopt the PR.

lib/inspec/resources/sys_info.rb
Vancelot11: I have a loop above that defines them as methods so they can be accessed that way. I switched it over, I didn't realize ruby doesn't like hyphens.

lib/inspec/resources/sys_info.rb
Vancelot11: I don't think I have access to a mac right now at least not that I can get setup with inspec. From the man pages though it looks like it can only do full or trim the domain. I'm not sure an alternate implementation is possible but I'm not a mac user.

Vancelot11: I refactored it so it should be portable on MacOS now. I couldn't test it but I based it off the mac hostname man page so I was able to still implement a couple of the options. I also added failures for unsupported options for all platforms.

clintoncwolfe (approved):
Thanks - changes look good! Looks like we still have a linting error, but after that is resolved, should be able to merge.

Vancelot11:
@clintoncwolfe Were you able to test in on mac? I tried to write it so that options mac has would be portable and any attempt at unsupported options would fail loudly but I wasn't able to test it on mac.

clintoncwolfe:
Yes, I tested on MacOS - thank you!

mhackethal:
Any chance you harmonize the functionality with WIndows Operating systems? Current sys_info for windows alway says:
"The `sys_info.hostname` resource is not supported with that option on your OS."
if you try to check for 'fqdn', ip_address , domain.

inspec> describe sys_info do
inspec>   its('fqdn') { should eq "foo" }
inspec>   its('ip_address') { should eq "foo" }
inspec>   its('domain') { should eq "bar" }
inspec> end

```
Profile: inspec-shell
Version: (not specified)

  System Information
     [FAIL]  fqdn should eq "foo"

     expected: "foo"
          got: "The `sys_info.hostname` resource is not

     (compared using ==)

     [FAIL]  ip_address should eq "foo"

     expected: "foo"
          got: "The `sys_info.hostname` resource is not

     (compared using ==)

     [FAIL]  domain should eq "bar"

     expected: "bar"
          got: "The `sys_info.hostname` resource is not

     (compared using ==)


Test Summary: 0 successful, 3 failures, 0 skipped
```

PR #4391, opened 2019-08-14
Deprecate platforms for 3-stable branch

jeremiahsnapp:
As of March 31st 2019, SLES 11 is no longer generally supported. Per our
support process we will no longer officially support SLES 11.

As of April 30th 2019, Ubuntu 14.04 is no longer generally supported. Per our
support process we will no longer officially support Ubuntu 14.04.

See https://docs.chef.io/platforms.html#platform-end-of-life-policy


miah (approved):
Thanks!

clintoncwolfe (approved):
Thanks!

Issue #4392, opened 2019-08-15
MssqlSession.query is not escaping double quote correctly


dalee-bis:
# Describe the problem

When `MssqlSession.query` is escaping the `"` character, it is doing so by prefixing it with a backslash (e.g. `\"`). This does not escape the quote character. It should be escaped by adding an additional double quote character (e.g. `""`).

Unfortunately, due to the current implementation, it is impossible to work around the issue (any use of `"` will add a backslash before so it cannot be doubled up.

Example SQL Query: `PRINT '"Test!"';`

Inspec Usage: `mssql_session.query("PRINT '\"Test!\"';")`

Generated command: `sqlcmd -Q "set nocount on; PRINT '\"Test\"!';" -W -w 1024 -s ','`

Output:
```
Msg 105, Level 15, State 1, Server EC2AMAZ-D3S9TNL, Line 1
Unclosed quotation mark after the character string ''.
```

Correct command: `sqlcmd -Q "set nocount on; PRINT '""Test""!';" -W -w 1024 -s ','`

Output:
```
"Test"!
```

## Possible Solution

Change https://github.com/inspec/inspec/blob/a35d37043e65ce5fd528d58cd48ecb369c74a432/lib/inspec/resources/mssql_session.rb#L56 to:

`escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '""').gsub(/\$/, '\\$')`

zenspider:
This is also true of oracledb_session.rb and probably others. We should do a sweep of all the db oriented resources.

aaronlippold:
https://github.com/inspec/inspec/pull/1939

kekaichinose:
Closed in #4393

PR #4393, opened 2019-08-15
Fixed MssqlSession.query not escaping double quote correctly

dalee-bis:
## Description

When `MssqlSession.query` is escaping the `"` character, it is doing so by prefixing it with a backslash (e.g. `\"`). This does not escape the quote character. It should be escaped by adding an additional double quote character (e.g. `""`).

## Related Issue

https://github.com/inspec/inspec/issues/4392

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/mssql_session.rb
+++ b/lib/inspec/resources/mssql_session.rb
@@ -53,7 +53,7 @@ module Inspec::Resources
     end
 
     def query(q) # rubocop:disable Metrics/PerceivedComplexity
-      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"').gsub(/\$/, '\\$')
+      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '""').gsub(/\$/, '\\$')
       # surpress 'x rows affected' in SQLCMD with 'set nocount on;'
       cmd_string = "sqlcmd -Q \"set nocount on; #{escaped_query}\" -W -w 1024 -s ','"
       cmd_string += " -U '#{@user}' -P '#{@password}'" unless @user.nil? || @password.nil?


miah (changes requested):
Thanks for the find and contribution @dalee-bis! Could you add a test for this? The test file is here: https://github.com/inspec/inspec/blob/master/test/unit/resources/mssql_session_test.rb#L65

Your commit will also need at DCO signature: https://github.com/inspec/inspec/blob/master/CONTRIBUTING.md#dco-sign-off-methods

dalee-bis:
Sorry but I'm not very familiar with Ruby. I only managed this change because it was mostly regex and easily testable at https://www.tutorialspoint.com/execute_ruby_online.php. I don't have an environment set up to ensure the test is running successfully.

Also editing the commit message is causing issues due to the privacy settings on my account. Is the DCO required for a 3 character change?

miah (approved):
You're right, this likely falls under our Obvious Fix policy so we can skip DCO.

lib/inspec/resources/mssql_session.rb
```diff
@@ -53,7 +53,7 @@ def initialize(opts = {})
     end
 
     def query(q) # rubocop:disable Metrics/PerceivedComplexity
-      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"').gsub(/\$/, '\\$')
+      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '""').gsub(/\$/, '\\$')
```
zenspider, line 56: I believe this is right for SQL... but this type of change _should_ have a test behind it if at all possible. I just saw your comment about your ruby experience so I'm gonna approve and try to follow up on this with the test (or at least file an issue for it).

zenspider (approved):

PR #4400, opened 2019-08-20
Fixed multiline error described in issue #3921

### LINKED ISSUE #3921: Filesystem resource doesn't work on RHEL 5/6 systems with long filesystem names
### 🎛 Description

**Choose one:** 🐛 bug report

The filesystem resource tends to break on RHEL 5/6 for devices with a long filesystem name, due to the filesystem wrapping to the next line. This behavior is easily avoidable by adding the `-P` (portable) option to the `df` command that is executed in `LinuxFileSystemResource`

3.9.0 on RHEL 5/6/7

### 🤔 Replication Case

On RHEL 7, if you have a long device name, it renders on one line, the way inspec expects, like so:
```
Filesystem                  Type 1K-blocks   Used Available Use% Mounted on
/dev/mapper/root_vg-root_lv xfs    2086912 807600   1279312  39% /
```

But on RHEL 5/6, it renders on two lines, like so:
```
Filesystem    Type   1K-blocks      Used Available Use% Mounted on
/dev/mapper/root_vg-root_lv
              ext3     1904472    972532    833636  54% /
```
Which means that the values that the existing code in `LinuxFileSystemResource` is pulling results in:

`["/dev/mapper/root_vg-root_lv"]`

Instead of (generated by using the existing logic, but changing the command to `df #{partition} -TP`):

`["/dev/mapper/root_vg-root_lv", "ext3", "1904472", "972372", "833796", "54%", "/"]`

### 💁 Possible Solutions

Change [lib/resources/filesystem.rb Line 95](https://github.com/inspec/inspec/blob/master/lib/resources/filesystem.rb#L95) to `cmd = inspec.command("df #{partition} -TP")` (add the P to the end)

### 💻 Stacktrace
```
     ×  FileSystem / percent_free
     divided by 0
```

Axuba:
Signed-off-by: Albert Jubany Juàrez <ajubanyjuarez@gmail.com>

## Description

Addresses the bug that happens on RedHat 5 and 6 that if a filesystem has a long name, it puts a newline right after the filesystem name, and InSpec's describe filesystem doesn't manage it.

## Related Issue

Fixes #3921

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/filesystem.rb
+++ b/lib/inspec/resources/filesystem.rb
@@ -100,7 +100,7 @@ module Inspec::Resources
         raise Inspec::Exceptions::ResourceFailed,
           "Unable to get available space for partition #{partition}"
       end
-      value = cmd.stdout.split(/\n/)[1].strip.split(" ")
+      value = cmd.stdout.gsub(/\n\s+/, " ").split(/\n/)[1].strip.split(" ")
       {
         name: partition,
         size_kb: value[2].to_i,


clintoncwolfe (commented):
So on #3921, there is a suggestion to add -P at https://github.com/inspec/inspec/blob/29bec90b1b487ab33b10988a9513c94b4b2b0a3e/lib/resources/filesystem.rb#L95 - which we noticed you did on an early commit, then rolled back. What obstacle did you encounter, and why does string manipulation of the output seem better? Thanks!

zenspider (changes requested):
Thank you for your PR!

I have the same question about `-P`, but would also like to see this PR rewritten to be a single commit. I usually do that with `git rebase -i <sha of start>` (via emacs' magit, which makes it a breeze), but you can also roll out all the commits (eg `git reset HEAD^` * N commits) and add a new commit w/ a final message (plus DCO please)... or start a new PR and just patch the current diff in. If anything, there are too many ways to do this.

Axuba:
> So on #3921, there is a suggestion to add -P at
> 
> https://github.com/inspec/inspec/blob/29bec90b1b487ab33b10988a9513c94b4b2b0a3e/lib/resources/filesystem.rb#L95
> - which we noticed you did on an early commit, then rolled back. What obstacle did you encounter, and why does string manipulation of the output seem better? Thanks!

@clintoncwolfe Hello, before committing the changes I tested it locally in my RHEL6 & RHEL7 servers, and it worked perfectly and it fixed the issue on RHEL6, while still working on RHEL7. But when I committed I noticed that the inspec-master-verify failed as you can see [here](https://buildkite.com/chef-oss/inspec-inspec-master-verify/builds/343). I wasn't sure why would it work on RedHat machines but I neither knew the environment those tests are run to.

So I basically tried to fix the issue doing a work-around, since I don't know the enviromnent those tests use, that manages to fix the issue on RedHat 6&7 servers while passing the verifier. And I came with the substitution String.

@zenspider thanks for the tips, it's the first time I do this, I do it as soon as I can :)

Axuba:
Sorry I had some problems with the branch. @zenspider Everything is in 1 commit now.

miah:
Thanks for the work on this @Axuba, decided on a slightly different fix in #4637

PR #4401, opened 2019-08-20
inputs: Accept bare input from the command line

### LINKED ISSUE #3300: Ability to set an attribute via the cli
🙋 feature request?

As part of our updates to attributes in inspec 3.x, we are also seeing more and more use of inspec in the ci/cd pipeline. The ability to pass these things only via the `attributes.yml` or via some `pipe foo` work yes, but a more stabilized approach would be great.

The ability to set an attribute from the CLI would be very useful as run / job specific data is easiest to push right form the task level. Examples would be the 'container_id'/container_name the pipeline just randomly created or a EC2 id that was just built. Further, things stored in Vault which I can request real time via a cli or perhaps a dynamic user_id or other data etc. 

`inspec exec . -t ... --attr='container_id':"${ID}" --attr=user:$USER --attrs=attributes.yml ` 

#719

clintoncwolfe:
## Description

Implements a simple CLI option, `--input` , which takes name-value pairs on the CLI and accepts them as inputs. This is much less cumbersome than `--input-file`, which requires you to first write out a YAML file. For quick demos and such, this is a pain reliever and UX delight.

Example:

```
inspec exec my-profile --input some_input=some_value
```

This is not an "input plugin" but it is a "way of reading inputs". It could not be a plugin because we don't expose the CLI option facility in any pluggable fashion.

Things that are not perfect:
 * Repeated --input flags overwrite the previous flag, rather than adding a new input. Currently you can add multiple inputs using just one --input flag, but that is less intuitive.
 * This is now the 4th think in InputRegistry that is reading off the runner config to bind inputs - we're overdue for a refactor there (#4402)

## Related Issue

This closes #3300 . 

## Types of changes

## Checklist:

### DIFF
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -63,7 +63,7 @@ That said, any profile that uses the DSL keyword `input()` (or the deprecated `a
 
 ### How can I set Inputs?
 
-As installed (without specialized plugins), Chef InSpec supports six ways of setting inputs:
+As installed (without specialized plugins), Chef InSpec supports several ways of setting inputs:
 
  * Inline in control code, using `input('input_name', value: 42)`.
  * In profile `inspec.yml` metadata files
@@ -80,7 +80,7 @@ In addition, Chef InSpec supports Input Plugins, which can provide optional inte
 
 Briefly:
 
-inline DSL < metadata < ( cli-inupt-file or kitchen-inspec or audit-cookbook ) < cli --input
+inline DSL < metadata < ( cli-input-file or kitchen-inspec or audit-cookbook ) < cli --input
 
 In addition, for inherited profiles:
 
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -46,11 +46,7 @@ Test Summary: 0 successful, 1 failure, 0 skipped
 
 That result clearly won't do. Let's override the input's default value. Create a file, `custom_amps.yml`:
 
-```yaml
-amplifier_max_volume: 11
-```
-
-We can now run that profile with `inspec exec rock_critic --input-file custom_amps.yaml`:
+We can now run that profile with `inspec exec rock_critic --input amplifier_max_volume=11`:
 
 ```
   11
@@ -71,7 +67,8 @@ As installed (without specialized plugins), Chef InSpec supports five ways of se
 
  * Inline in control code, using `input('input_name', value: 42)`.
  * In profile `inspec.yml` metadata files
- * Using the CLI option `--input-file somefile.yaml`
+ * Using the CLI option `--input name1=value1 name2=value2...` to read directly from the command line
+ * Using the CLI option `--input-file somefile.yaml` to read inputs from files
  * In kitchen-inspec, using the `verifier/inputs` settings
  * In the Audit Cookbook, using the `node[:audit][:inputs]`
 
@@ -83,7 +80,7 @@ In addition, Chef InSpec supports Input Plugins, which can provide optional inte
 
 Briefly:
 
-inline DSL < metadata < ( cli-input-file or kitchen-inspec or audit-cookbook )
+inline DSL < metadata < ( cli-inupt-file or kitchen-inspec or audit-cookbook ) < cli --input
 
 In addition, for inherited profiles:
 
@@ -122,7 +119,7 @@ As packaged, Chef InSpec uses the following priority values:
 | CLI `--input-file` option              | 40       |   No                |
 | inspec-kitchen `inputs:` section       | 40       |   No                |
 | audit cookbook `node[:audit][:inputs]` | 40       |   No                |
-
+| CLI `--input` option                   | 50       |   No                |
 
 ### What happened to "Attributes"?
 
@@ -246,9 +243,31 @@ an_input: a_value
 another_input: another_value
 ```
 
-CLI-set inputs have a priority of 40.
+CLI-input-file-set inputs have a priority of 40.
 
 As of Chef InSpec 4.3.2, this mechanism has the following limitations:
+
+  1. No [input options](#input-options-reference) may be set - only the name and value.
+  2. Because the CLI is outside the scope of any individual profile and the inputs don't take options, the inputs are clumsily copied into every profile, effectively making the CLI mechanism global.
+
+## Setting Input values using `--input`
+
+You may also provide inputs and values directly on the command line:
+
+```yaml
+inspec exec my_profile --input input_name=input_value
+```
+
+To set multiple inputs, say:
+```yaml
+inspec exec my_profile --input input_name1=input_value1 name2=value2
+```
+
+Don't repeat the `--input` flag; that will override the previous setting.
+
+CLI-set inputs have a priority of 50.
+
+As of Chef InSpec 4.12, this mechanism has the following limitations:
 
   1. No [input options](#input-options-reference) may be set - only the name and value.
   2. Because the CLI is outside the scope of any individual profile and the inputs don't take options, the inputs are clumsily copied into every profile, effectively making the CLI mechanism global.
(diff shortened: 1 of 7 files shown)


docs/inputs.md
```diff
@@ -46,11 +46,7 @@ Test Summary: 0 successful, 1 failure, 0 skipped
 
 That result clearly won't do. Let's override the input's default value. Create a file, `custom_amps.yml`:
 
-```yaml
-amplifier_max_volume: 11
-```
-
-We can now run that profile with `inspec exec rock_critic --input-file custom_amps.yaml`:
+We can now run that profile with `inspec exec rock_critic --input amplifier_max_volume=11`:
```
mjingle, line 49: Would the last sentence in line 47 be removed in the case of the new line 49?

docs/inputs.md
```diff
@@ -71,7 +67,8 @@ As installed (without specialized plugins), Chef InSpec supports five ways of se
 
  * Inline in control code, using `input('input_name', value: 42)`.
  * In profile `inspec.yml` metadata files
- * Using the CLI option `--input-file somefile.yaml`
+ * Using the CLI option `--input name1=value1 name2=value2...` to read directly from the command line
```
mjingle, line 70: Line 66 needs to be updated from "five" to "six" with this new addition :)

docs/inputs.md
```diff
@@ -246,13 +243,35 @@ an_input: a_value
...
+```yaml
+inspec exec my_profile --input input_name=input_value
+```
+
+To set multiple inputs, say:
+```yaml
+inspec exec my_profile --input input_name1=input_value1 name2=value2
+```
+
+Don't repeat the `--input` flag; that will override the previous setting.
```
mjingle:
```suggestion
Do not repeat the `--input` flag as that will override the previous setting.
```

docs/inputs.md
```diff
@@ -67,11 +63,12 @@ That said, any profile that uses the DSL keyword `input()` (or the deprecated `a
 
 ### How can I set Inputs?
 
-As installed (without specialized plugins), Chef InSpec supports five ways of setting inputs:
+As installed (without specialized plugins), Chef InSpec supports six ways of setting inputs:
```
zenspider: "several" ? @mjingle ?

docs/inputs.md
```diff
@@ -83,7 +80,7 @@ In addition, Chef InSpec supports Input Plugins, which can provide optional inte
 
 Briefly:
 
-inline DSL < metadata < ( cli-input-file or kitchen-inspec or audit-cookbook )
+inline DSL < metadata < ( cli-inupt-file or kitchen-inspec or audit-cookbook ) < cli --input
```
zenspider: `s/inupt/input/`

lib/inspec/input_registry.rb
```diff
@@ -132,10 +132,37 @@ def bind_profile_inputs(profile_name, sources = {})
       bind_inputs_from_metadata(profile_name, sources[:profile_metadata])
       bind_inputs_from_input_files(profile_name, sources[:cli_input_files])
       bind_inputs_from_runner_api(profile_name, sources[:runner_api])
+      bind_inputs_from_cli_args(profile_name, sources[:cli_input_arg])
```
zenspider, line 138: Is this reverse priority order aka last one wins?

lib/inspec/input_registry.rb
```diff
@@ -132,10 +132,37 @@ def bind_profile_inputs(profile_name, sources = {})
...
+        unless pair.include?("=")
+          if pair.end_with?(".yaml")
+            raise ArgumentError, "ERROR: --input is used for individual input values, as --input name=value. Use --input-file to load a YAML file."
+          else
+            raise ArgumentError, "ERROR: An '=' is required when using --input. Usage: --input input_name1=input_value1 input2=value2"
+          end
+        end
+        input_name, input_value = pair.split("=")
+        evt = Inspec::Input::Event.new(
+          value: input_value.sub(/,$/, ""), # Trim trailing comma if any
```
zenspider: `input_value.chomp(",")` or `input_value.delete_suffix(",")` (but I don't know what version of ruby that was introduced in)

lib/inspec/profile.rb
```diff
@@ -128,7 +128,8 @@ def initialize(source_reader, options = {})
         cli_input_files: options[:runner_conf][:input_file], # From CLI --input-file
         profile_metadata: metadata,
         # TODO: deprecation checks here
-        runner_api: options[:runner_conf][:attributes] # This is the route the audit_cookbook and kitchen-inspec take
+        runner_api: options[:runner_conf][:attributes], # This is the route the audit_cookbook and kitchen-inspec take
+        cli_input_arg: options[:runner_conf][:input]
```
zenspider: this diff is why I like to leave trailing commas... but I've not gotten used to doing it in arg lists.

test/functional/inputs_test.rb
```diff
@@ -42,6 +42,13 @@
       line = lines.detect { |l| l.include? "--attrs" }
       line.wont_be_nil
     end
+
+    it "includes the --input option" do
+      result = run_inspec_process("exec help", lock: true) # --no-create-lockfile option breaks usage help
+      lines = result.stdout.lines
+      line = lines.detect { |l| l.include? "--input" }
```
zenspider: `assert_include lines.join, "--input"`

but both versions have the problem that `--input-file` will match this... so probably `assert_match` w/ `$` or something.

test/functional/inputs_test.rb
```diff
@@ -67,6 +74,56 @@
     end
   end
 
+  describe "when using the --input inline raw input flag CLI option" do
+    let(:result) { run_inspec_process("exec #{inputs_profiles_path}/cli #{input_opt} #{control_opt}", json: true) }
+    let(:control_opt) { "" }
+
+    describe "when the --input is used once with one value" do
+      let(:input_opt) { "--input test_input_01=value_from_cli_01" }
+      let(:control_opt) { "--controls test_control_01" }
```
zenspider, line 156: These don't really need to be functional tests. Processing the command line args directly and then making sure that the options are processed correctly would suffice and cost nothing in comparison to this very heavyweight alternative.

zenspider (changes requested):
A bit heftier of an ask on testing... Feel free to punt on it, but I'd like an issue opened to do it later if so.

docs/inputs.md
mjingle: The change to "several" would be recommended. There's less maintenance this way.

mjingle:
```suggestion
As installed (without specialized plugins), Chef InSpec supports several ways of setting inputs:
```

lib/inspec/profile.rb
clintoncwolfe: I like trailing commas too, but chefstyle does not permit them.

lib/inspec/input_registry.rb
clintoncwolfe: Last one wins, if there is a tie in priority. But cli_args have higher priority anyway.

test/functional/inputs_test.rb
clintoncwolfe: OK, I cut an issue for that: https://github.com/inspec/inspec/issues/4412

zenspider (changes requested):
All my original feedback still stands?

Did you forget to push?

clintoncwolfe:
D'oh! Yes, you're right, I made most of the requested changes, forgot to push, and then requested a new review. Sorry about the wasted cycle, and thanks for your patience.

zenspider (approved):

zenspider (approved):

miah (approved):
Seems reasonable. Thanks @clintoncwolfe!

scalp42:
Amazing, thanks a lot 🙇

Issue #4417, opened 2019-08-26
Waivers Step 3: Extend JSON output to include waiver data


zenspider:
This is part of epic #4408.

* Changes to output to add extra fields to indicate wavered controls
  * Just adds new fields

Step 3: reporting: we need to add the waiver data to the control, which implies that even if a control is skipped it still has some payload in the report output, the data added is whether this control was skipped, for how long (optionally) and why.

zenspider:
* Changes to output to add extra fields to indicate wavered controls
  * Just adds new fields

Step 3: reporting: we need to add the waiver data to the control, which implies that even if a control is skipped it still has some payload in the report output, the data added is whether this control was skipped, for how long (optionally) and why.

zenspider:
From #4415:

### User Story
As an Operator, I want to report the status of waived controls so my stakeholders so that I can provide an accurate assessment of my compliance posture.

### Acceptance Criteria
* Control status reporting is updated
  * waiver state is included
  * and if a control is waived, its body of data is included
    * __Control ID_Name__ - the ID and Title of the control to be waived
    * __Expiration Date__ - the date when the waiver is set to expire
    * __Waiver Justification__ *(optional)* - a way to allow users to justify the waiver (i.e. why the given control must be waived)
    * __Skip run switch__ - allow users to skip the evaluating the waived control
* Updated reporting doesn't impact existing reporter performance

miah:
Blocked until some design issues are cleared up in #4427

Issue #4414, opened 2019-08-23
Waivers Step 2: Handle waivers during runtime


kekaichinose:
This is part of epic #4408.

### User Story
As an Operator, I need a waiver to override the existing impact of a control and optionally skip execution altogether so that I can mitigate new compliance standards

### Acceptance Criteria
* ingested waiver information modifies control list
* modified control list runs
  * and waived controls carry a 0 impact score
  * and waived controls that are configured as *skipped* are not evaluated

zenspider:
* Some kind of internal overlay to enforce skips
  * We have our choice of existing functionality to (possibly) leverage
    * skip_control
    * only_if
    * skip (rspec outcome)

* skip_control suppresses controls from from output - but waivered controls must appear in output.
* We'd like to avoid creating a new way of altering behavior of profiles - we already have Inputs. So, for the front-end side, we'd like to use Inputs for control, or something similar

Step 2: If that waiver is set to true or has an expiration date that is still valid, then we skip the run of that control (this is a new check to the runtime of a control).

zenspider:
See also: #3845

clintoncwolfe:
This was addressed on #4427

Issue #4409, opened 2019-08-22
Waivers Step 1: Input File & CLI


kekaichinose:
This is part of epic #4408.

### User Story
As an Operator, I need to have a way to define and read in waivers so I can apply them to my profiles when doing an inspec run.

### Acceptance Criteria
* define a file format (e.g. toml, yaml, etc) to store waiver configurations
  * a waiver is applied at the control level
  * a waiver entry will contain the following data
    * __Control ID_Name__ - the ID and Title of the control to be waived
    * __Expiration Date__ - the date when the waiver is set to expire
    * __Waiver Justification__ *(optional)* - a way to allow users to justify the waiver (i.e. why the given control must be waived)
    * __Skip run switch__ - allow users to skip evaluating the waived control
* be able to read in waiver file from command line

clintoncwolfe:
```
Control ID_Name - the ID and Title of the control to be waived
```
Controls have both an ID and a title - which value will be available in this field? I recommend ID.

galenemery:
I'm ➕ for ID

zenspider:
+ Input doc
  + a cli option
  + need to pin down a format
    + Control id (name?)
    + Optional expiration date
    + Waiver reason blurb
    + Skip boolean (should this run and a failure does NOT affect the overall result vs should this just not run)
  + straightforward

Draft file format:

```yaml
xccdf_org.cisecurity.benchmarks_rule_1.21_something_else:
  expiry_date: 2019-09-15
  justification: We don't like this one
  skip_run: true

xccdf_org.cisecurity.benchmarks_rule_1.1.1.4_Ensure_mounting_of_hfs_filesystems_is_disabled:
  justification: We don't like this one
```

PR #4427, opened 2019-08-28
waivers: functional test fixture and eval-time skipping

clintoncwolfe:
## Description

This is an implementation of #3845/#4414, which uses inputs to act as if `only_if` had been defined for a control. The waiver rules are applied, and the control is conditionally skipped using the only_if facility.

This PR includes a test fixture profile, which has a control and corresponding Input rotating through 11 cases of the various ways waivers may be used.

There is also a passing functional test.  It really does test that the controls are skipped when waivered; however there are several no-op assertions in the functional test file that are just placeholders for  Part 3.

There is an input file included. I made up the field names - they can certainly be adjusted as can the values.

There are placeholders in the functional test file for assertion methods for detecting3845 waiver annotations in the JSON report.

I went with a functional test after staring at the unit tests for the profile and control_eval_context unit tests. That's where the internal unit tests belong, but that is really opaque code. I was able to get a functional test up and running, and start tracing things that way much more easily. Happy to circle back later on that.

## Review Notes

Several things to discuss:

* ~~The waiver rules are currently applied BEFORE evaulating the contents of the control. I believe that means that any real `only_if` block in the profile source will have higher precedence.~~ UPDATE: the waiver rules now apply after the only_if.

* the ability to "slug" the control IDs (replacing spaces, etc) is not implemented here but needs to be implemented somewhere - not sure if here is the right place. @zenspider?

* ~~As implemented, some of the waiver logic (such as expiration date comparision) might need to be re-created in step 3 #4417. We may want to look into how to avoid that.~~ UPDATE: all logic happens here in step 2 now, and it records information in the run_data for step 3 to display, hopefully with only presentation (not business) logic.

## Related Issue

#3845 , #4414 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -43,21 +43,21 @@ module Inspec
       @__rule_id = id
       @__profile_id = profile_id
       @__checks = []
-      @__skip_rule = {} # { result: true, message: "Why" }
+      @__skip_rule = {} # { result: true, message: "Why", type: [:only_if, :waiver] }
       @__merge_count = 0
       @__merge_changes = []
       @__skip_only_if_eval = opts[:skip_only_if_eval]
-
-      # TODO: Note - if waivers are applied before the DSL instance
-      # eval, then the contents of the control block (including only_if)
-      # will override the waiver. Is that wanted? Or should waivers always win?
-      __apply_waivers
 
       # evaluate the given definition
       return unless block_given?
 
       begin
         instance_eval(&block)
+
+        # By applying waivers *after* the instance eval, we assure that
+        # waivers have higher precedence than only_if.
+        __apply_waivers
+
       rescue StandardError => e
         # We've encountered an exception while trying to eval the code inside the
         # control block. We need to prevent the exception from bubbling up, and
@@ -147,6 +147,7 @@ module Inspec
       return if @__skip_only_if_eval == true
 
       @__skip_rule[:result] ||= !yield
+      @__skip_rule[:type] = :only_if
       @__skip_rule[:message] = message
     end
 
@@ -199,9 +200,9 @@ module Inspec
       rule.instance_variable_get(:@__skip_rule)
     end
 
-    def self.set_skip_rule(rule, value, message = nil)
+    def self.set_skip_rule(rule, value, message = nil, type = :only_if)
       rule.instance_variable_set(:@__skip_rule,
-        { result: value, message: message })
+        { result: value, message: message, type: type })
     end
 
     def self.merge_count(rule)
@@ -219,9 +220,9 @@ module Inspec
       return checks(rule) unless skip_check[:result].eql?(true)
 
       if skip_check[:message]
-        msg = "Skipped control due to only_if condition: #{skip_check[:message]}"
+        msg = "Skipped control due to #{skip_check[:type]} condition: #{skip_check[:message]}"
       else
-        msg = "Skipped control due to only_if condition."
+        msg = "Skipped control due to #{skip_check[:type]} condition."
       end
 
       # TODO: we use os as the carrier here, but should consider
@@ -259,7 +260,8 @@ module Inspec
       skip_check = skip_status(src)
       sr = skip_check[:result]
       msg = skip_check[:message]
-      set_skip_rule(dst, sr, msg) unless sr.nil?
+      skip_type = skip_check[:type]
+      set_skip_rule(dst, sr, msg, skip_type) unless sr.nil?
 
       # Save merge history
       dst.instance_variable_set(:@__merge_count, merge_count(dst) + 1)
@@ -315,7 +317,8 @@ module Inspec
 
       # OK, apply a skip.
       @__skip_rule[:result] = true
-      @__skip_rule[:message] = "Skipped due to waiver: #{waiver_info["justification"] || "(no reason given)"}"
+      @__skip_rule[:type] = :waiver
+      @__skip_rule[:message] = waiver_info["justification"]
     end
 
     #
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -283,21 +283,25 @@ module Inspec
     def __apply_waivers
       input_name = @__rule_id # TODO: control ID slugging
       registry = Inspec::InputRegistry.instance
-      input = registry.inputs_by_profile[@__profile_id].dig(input_name)
+      input = registry.inputs_by_profile.dig(@__profile_id, input_name)
       return unless input
 
-      input = input.value
+      # An InSpec Input is a datastructure that tracks a profile parameter
+      # over time. Its value can be set by many sources, and it keeps a
+      # log of each "set" event so that when it is collapsed to a value,
+      # it can determine the correct (highest priority) value.
+      waiver_info = input.value
 
-      # Waivers should have a hash value with keys including skip and
+      # Waivers should have a hash value with keys possibly including skip and
       # expiration_date. We only care here if it has a skip key and it
       # is yes-like, since all non-skipped waiver operations are handled
       # during reporting phase.
-      return unless input.key?("skip")
-      return unless input["skip"].to_s.match(/y|yes|true/i)
+      return unless waiver_info.key?("skip")
+      return unless waiver_info["skip"].to_s.match(/y|yes|true/i)
 
       # OK, the intent is to skip. Does it have an expiration date, and
       # if so, is it in the future?
-      expiry = input["expiration_date"]
+      expiry = waiver_info["expiration_date"]
       if expiry
         if expiry.is_a?(Date)
           # It appears that yaml.rb automagically parses dates for us
@@ -311,7 +315,7 @@ module Inspec
 
       # OK, apply a skip.
       @__skip_rule[:result] = true
-      @__skip_rule[:message] = "Skipped due to waiver: #{input["justification"] || "(no reason given)"}"
+      @__skip_rule[:message] = "Skipped due to waiver: #{waiver_info["justification"] || "(no reason given)"}"
     end
 
     #
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -42,7 +42,7 @@ module Inspec
       @__rule_id = id
       @__profile_id = profile_id
       @__checks = []
-      @__skip_rule = {}
+      @__skip_rule = {} # { result: true, message: "Why" }
       @__merge_count = 0
       @__merge_changes = []
       @__skip_only_if_eval = opts[:skip_only_if_eval]
@@ -206,6 +206,8 @@ module Inspec
       rule.instance_variable_get(:@__merge_changes)
     end
 
+    # If a rule is marked to be skipped, this
+    # creates a dummay array of "checks" with a skip outcome
     def self.prepare_checks(rule)
       skip_check = skip_status(rule)
       return checks(rule) unless skip_check[:result].eql?(true)
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -281,7 +281,7 @@ module Inspec
     # only_if mechanism)
     # Double underscore: not intended to be called as part of the DSL
     def __apply_waivers
-      input_name = "waiver_#{@__rule_id}" # TODO: control ID slugging
+      input_name = @__rule_id # TODO: control ID slugging
       registry = Inspec::InputRegistry.instance
       input = registry.inputs_by_profile[@__profile_id].dig(input_name)
       return unless input
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -312,8 +312,8 @@ module Inspec
           # It appears that yaml.rb automagically parses dates for us
           if expiry < Date.today # If the waiver expired, return - no skip applied
             __waiver_data["message"] = "Waiver expired on #{expiry}, evaluating control normally"
+            return
           end
-          return
         else
           ui = Inspec::UI.new
           ui.error("Unable to parse waiver expiration date '#{expiry}' for control #{@__rule_id}")
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -29,6 +29,7 @@ module Inspec
       @resource_dsl
     end
 
+    attr_reader :__waiver_data
     def initialize(id, profile_id, opts, &block)
       @impact = nil
       @title = nil
@@ -292,22 +293,27 @@ module Inspec
       # over time. Its value can be set by many sources, and it keeps a
       # log of each "set" event so that when it is collapsed to a value,
       # it can determine the correct (highest priority) value.
-      waiver_info = input.value
+      # Store in an instance variable for.. later reading???
+      @__waiver_data = input.value
+      __waiver_data["skipped_due_to_waiver"] = false
+      __waiver_data["message"] = ""
 
       # Waivers should have a hash value with keys possibly including skip and
       # expiration_date. We only care here if it has a skip key and it
       # is yes-like, since all non-skipped waiver operations are handled
       # during reporting phase.
-      return unless waiver_info.key?("skip")
-      return unless waiver_info["skip"].to_s.match(/y|yes|true/i)
+      return unless __waiver_data.key?("skip") && __waiver_data["skip"]
 
       # OK, the intent is to skip. Does it have an expiration date, and
       # if so, is it in the future?
-      expiry = waiver_info["expiration_date"]
+      expiry = __waiver_data["expiration_date"]
       if expiry
         if expiry.is_a?(Date)
           # It appears that yaml.rb automagically parses dates for us
-          return if expiry < Date.today # If the waiver expired, return - no skip applied
+          if expiry < Date.today # If the waiver expired, return - no skip applied
+            __waiver_data["message"] = "Waiver expired on #{expiry}, evaluating control normally"
+          end
+          return
         else
           ui = Inspec::UI.new
           ui.error("Unable to parse waiver expiration date '#{expiry}' for control #{@__rule_id}")
@@ -318,7 +324,8 @@ module Inspec
       # OK, apply a skip.
       @__skip_rule[:result] = true
       @__skip_rule[:type] = :waiver
-      @__skip_rule[:message] = waiver_info["justification"]
+      @__skip_rule[:message] = __waiver_data["justification"]
+      __waiver_data["skipped_due_to_waiver"] = true
     end
 
     #
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -1,6 +1,7 @@
 # copyright: 2015, Dominik Richter
 
 require "method_source"
+require "date"
 require "inspec/describe"
 require "inspec/expect"
 require "inspec/resource"
@@ -46,6 +47,11 @@ module Inspec
       @__merge_count = 0
       @__merge_changes = []
       @__skip_only_if_eval = opts[:skip_only_if_eval]
+
+      # TODO: Note - if waivers are applied before the DSL instance
+      # eval, then the contents of the control block (including only_if)
+      # will override the waiver. Is that wanted? Or should waivers always win?
+      __apply_waivers
 
       # evaluate the given definition
       return unless block_given?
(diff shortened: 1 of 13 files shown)


lib/inspec/rule.rb
```diff
@@ -42,11 +43,16 @@ def initialize(id, profile_id, opts, &block)
       @__rule_id = id
       @__profile_id = profile_id
       @__checks = []
-      @__skip_rule = {}
+      @__skip_rule = {} # { result: true, message: "Why" }
```
zenspider: By piggybacking here, does that mean that "last one wins" and thus, we might lose the information of the waiver? I don't know how `only_if` works, but I assume it'll re-set some of these fields and we'll lose information. We might want to prefix fields in this PR with `waiver_` or something similar.

IOW, how does automate differentiate between a normal only_if triggering and a waiver triggering?

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
       @__checks.push([describe_or_expect, values, block])
     end
 
+    # Look for an input with a matching ID, and if found, apply waiver
+    # skipping logic. Basically, if we have a current waiver, and it says
+    # to skip, we'll replace all the checks with a dummy check (same as
+    # only_if mechanism)
+    # Double underscore: not intended to be called as part of the DSL
```
zenspider, line 285: `private`?

lib/inspec/rule.rb
```diff
@@ -42,11 +43,16 @@ def initialize(id, profile_id, opts, &block)
...
       @__checks = []
-      @__skip_rule = {}
+      @__skip_rule = {} # { result: true, message: "Why" }
       @__merge_count = 0
       @__merge_changes = []
       @__skip_only_if_eval = opts[:skip_only_if_eval]
 
+      # TODO: Note - if waivers are applied before the DSL instance
+      # eval, then the contents of the control block (including only_if)
+      # will override the waiver. Is that wanted? Or should waivers always win?
```
zenspider: I think this is what I was trying to say above. We need waivers to win OR we need to distinguish between the two via the same data structure.

test/functional/waivers_test.rb
```diff
@@ -0,0 +1,73 @@
...
+describe "waivers" do
+  include FunctionalHelper
+  let(:waivers_profiles_path) { "#{profile_path}/waivers" }
+
+  def assert_test_outcome(expected, result_json)
+    assert_equal expected, result_json["status"]
+  end
+
+  def assert_waiver_annotation(result_json)
+    # TODO - test JSON for waiver annotation
```
zenspider, line 15: these should call `flunk` to make it obvious that they're not done.

test/functional/waivers_test.rb
```diff
@@ -0,0 +1,73 @@
...
+  def refute_waiver_annotation(result_json)
+    # TODO - test JSON for waiver annotation
+    # Don't suppose we get this for free by defining assert_waiver_annotation ...
+  end
+
+  describe "a fully pre-slugged control file" do
+    it "has all of the expected outcomes" do
+      cmd = "exec #{waivers_profiles_path}/basic --input-file #{waivers_profiles_path}/basic/files/waivers.yaml"
+      run_result = run_inspec_process(cmd, json: true)
+      controls_by_id = run_result.payload.json["profiles"][0]["controls"].map { |c| [c["id"], c] }.to_h
```
zenspider: I'm curious why you use `dig` for a single level above but don't use it here for 3 levels here?

test/functional/waivers_test.rb
```diff
@@ -0,0 +1,73 @@
...
+        assert_test_outcome "passed", result
+        refute_waiver_annotation result
+      end
+
+      [
+        "02_not_waivered_fails",
+        "07_waivered_expiry_in_past_not_skipped_fails", # Should this give a special waiver expired message?
+      ].each do |control_id|
+        result = controls_by_id[control_id]["results"][0]
+        assert_test_outcome "failed", result
```
zenspider: The repetition here with `result =` and the `assert_*` between all of these `each` blocks hints that `control_id` should be the thing passed in:

```ruby
let(:controls_by_id) { run_result.payload.json.dig("profiles", 0, "controls").map { |c| [c["id"], c] }.to_h }
# ...
assert_test_outcome "failed", control_id
```

test/functional/waivers_test.rb
```diff
@@ -0,0 +1,73 @@
...
+        assert_waiver_annotation result
+      end
+
+    end
+  end
+
+  # describe "an inherited profile"
+  # describe "a profile whose control ids require transformation"
```
zenspider, line 69: These could be real describes with an `it` that has a `skip` so we have something in our output.

test/unit/mock/profiles/waivers/basic/controls/basic.rb
```diff
@@ -0,0 +1,43 @@
+control "01_not_waivered_passes" do
+  describe(true) { it { should eq true } }
```
zenspider, line 2: I find this extra verbage confusing tbh... The true/false values are not important and adding noise. The nominal cases could be something simple like:

```ruby
describe(:passes) { it { :passes } }
```
and the waived case could be
```ruby
describe(:fails) { it { fail "don't run me" } }
```

making it so you don't have to think about the positive case at all because it both tells you that it passes but also because it essentially does nothing. I don't have to `eval` it in my head. Likewise, the negative case literally tells me that it fails and then does exactly that. Again, no `eval` in my head.

test/unit/mock/profiles/waivers/basic/controls/basic.rb
```diff
@@ -0,0 +1,43 @@
...
+control "09_waivered_expiry_in_future_not_skipped_passes" do
+  describe(true) { it { should eq true } }
+end
+
+control "10_waivered_expiry_in_future_not_skipped_fails" do
+  describe(true) { it { should eq false } }
+end
+
+control "11_waivered_expiry_in_future_skipped" do
+  describe(true) { it { should eq true } }
```
zenspider, line 42: This can lead to a false positive if we don't implement anything because this dosn't fail.

test/unit/mock/profiles/waivers/basic/files/waivers.yaml
```diff
@@ -0,0 +1,44 @@
+waiver_03_waivered_no_expiry_not_skipped_passes:
+  expiration_date: never
+  justification: Sound reasoning
+  skip: no
```
zenspider, line 3: I think this file is both covering too much and missing some steps. The edge cases as I see them:

* expiration_date: nil (we said it was optional), past, future
* skip: yes, no
* test result: passing or failing

and so we _could_ have 12 waiver combinations... but in reality we don't need to. For example, past versus future doesn't need to test every combination of skip or test result, just that it does invoke properly when in the future and that past acts the same as nil. See waiver `09` and `10` for what I think of as redundant tests, and expiration as `nil` as a missing test.

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
       @__checks.push([describe_or_expect, values, block])
     end
 
+    # Look for an input with a matching ID, and if found, apply waiver
+    # skipping logic. Basically, if we have a current waiver, and it says
+    # to skip, we'll replace all the checks with a dummy check (same as
+    # only_if mechanism)
+    # Double underscore: not intended to be called as part of the DSL
+    def __apply_waivers
+      input_name = "waiver_#{@__rule_id}" # TODO: control ID slugging
```
zenspider: The design spec we came up with only used the ID and didn't prefix with `waiver_`. I thought we were keying in on the example document from the design meeting and would just be looking for "skip":

https://github.com/inspec/inspec/issues/4409#issuecomment-525049831

Is there a reason for this?

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+      registry = Inspec::InputRegistry.instance
+      input = registry.inputs_by_profile[@__profile_id].dig(input_name)
```
zenspider: I don't understand using `dig` this way. Wouldn't using it for both levels make more sense?

```ruby
input = registry.inputs_by_profile.dig(@__profile_id, input_name)
```

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+      return unless input
+
+      input = input.value
```
zenspider: This probably needs a comment explaining the datastructure involved. You did a `dig` above to `input_name`, so don't you have the value?

And I'm very confused digging into this and trying to parse:
https://github.com/inspec/inspec/blob/435747a2d2b4a52b7044eda65133bdbb19121acf/lib/inspec/input_registry.rb#L24

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+
+      # Waivers should have a hash value with keys including skip and
+      # expiration_date. We only care here if it has a skip key and it
```
zenspider, line 302: The design stated that expiration date was optional.

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+      # is yes-like, since all non-skipped waiver operations are handled
+      # during reporting phase.
+      return unless input.key?("skip")
+      return unless input["skip"].to_s.match(/y|yes|true/i)
```
zenspider: We specified this to be a boolean and should treat it as such... This could just be:
```ruby
return unless input["skip"]
```

BUT! a true `skip` meant "don't run this" and I would think we'd need to attach the waiver data somewhere in the case of a non-skipped-but-waived control, no?

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+
+      # OK, the intent is to skip. Does it have an expiration date, and
+      # if so, is it in the future?
+      expiry = input["expiration_date"]
+      if expiry
+        if expiry.is_a?(Date)
+          # It appears that yaml.rb automagically parses dates for us
+          return if expiry < Date.today # If the waiver expired, return - no skip applied
+        elsif expiry.match(/never/i) # any other values?
```
zenspider: This was not in the design.

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+      return unless input.key?("skip")
+      return unless input["skip"].to_s.match(/y|yes|true/i)
+
+      # OK, the intent is to skip. Does it have an expiration date, and
+      # if so, is it in the future?
+      expiry = input["expiration_date"]
+      if expiry
+        if expiry.is_a?(Date)
+          # It appears that yaml.rb automagically parses dates for us
+          return if expiry < Date.today # If the waiver expired, return - no skip applied
```
zenspider: I guess we never specified whether `expiration_date` meant "through" or "until"... so this might need to be a `<=`?

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+        elsif expiry.match(/never/i) # any other values?
+          # Do nothing, fall through
+        else
+          raise Inspec::Exceptions::ResourceFailed, "Unable to parse waiver expiration date '#{expiry}' for control #{@__rule_id}"
```
zenspider: We parsed it (valid yaml)... it just has invalid data.

lib/inspec/rule.rb
```diff
@@ -267,6 +275,45 @@ def __add_check(describe_or_expect, values, block)
...
+        end
+      end
+
+      # OK, apply a skip.
+      @__skip_rule[:result] = true
+      @__skip_rule[:message] = "Skipped due to waiver: #{input["justification"] || "(no reason given)"}"
```
zenspider: I might be misunderstanding the point of `@__skip_rule`, but to me this should be passing through the expiration date and whether it was skipped?

...or does the backend need to re-process all of the same input/waiver data?

The answer to this question determines the overall implementation of step 3.

zenspider: Also, since the tests are laid out for step 3... should @miah be working _on top_ of your branch?

zenspider (changes requested):
Applied a critical eye to this one since it determines a lot for step 3 (and step 1 to a lesser degree).

lib/inspec/rule.rb
clintoncwolfe: The body of each control is instance eval'd against an instance of this class - so `private` won't stop someone from writing profile code that tinkers with this stuff. (though there is a `private` declaration later in the file.)

test/functional/waivers_test.rb
clintoncwolfe: I think of it more for cases when I don't know if an intermediate is nil, vs avoiding a long chain of references. It's prettier as a dig here, certainly.

test/unit/mock/profiles/waivers/basic/controls/basic.rb
clintoncwolfe: Well, I think this comes from different perspectives. I'm not someone with years of Ruby testing experience (so my brain doesn't see the same "labor pathways" as yours), but I am someone with some InSpec exposure; so my goal is to make the test fixture look like a minimal inspec profile. I would not want someone - like a new contributor - to open a simple test file and get hung up on unfamiliar code; it should just look like InSpec, unless we have a strong reason for it not to look like InSpec. And InSpec usually has `should op value` stanzas. So, that's why I provide them.

test/unit/mock/profiles/waivers/basic/controls/basic.rb
clintoncwolfe: This is the test fixture, not the test. The test is to verify that this should skip, not pass. If it passes, that's a defect, and our test should detect that.

We can add an additional case that would fail if not skipped, if you like.

test/unit/mock/profiles/waivers/basic/files/waivers.yaml
clintoncwolfe: I think we should probably draw up a formal IO table and show it to Product to verify behavior.  In the meantime I've accepted the note about nil expiration and am adding test cases to cover.

I don't see any harm in redundancy in test cases - I'm much more concerned about thoroughness, and seeing every case covered.

lib/inspec/rule.rb
clintoncwolfe: I'll alter behavior as follows:
* if absent, no expiration date (never expires)
* if present, must be a valid parseable date
* no other values valid (no special handling of "never", etc)

lib/inspec/rule.rb
clintoncwolfe: No, it would have been parsed twice - the YAML parser, then again as a Date. I'm saying the second parsing failed.

lib/inspec/rule.rb
clintoncwolfe: No, just headcanon. Removed.

lib/inspec/rule.rb
clintoncwolfe: Yes. I was concerned because the registry delegates part of that - but yes, here, that's clearly fine.

lib/inspec/rule.rb
clintoncwolfe: I've renamed the variable and added a clarifying comment.

lib/inspec/rule.rb
clintoncwolfe: I've been assuming that the job of phase 2 is to handle skipping - alterations to the runtime execution plan - and phase three is to handle reporting.  Both phase 2 and 3 can look to the InputRegistry as the sole source of truth for waiver data. 

I think if we were to stash that data somewhere now, I'd feel like we were just making another copy of it for unclear reasons. Unless we want to avoid scanning the entire list of controls during reporting?

lib/inspec/rule.rb
clintoncwolfe: Your call :-)

lib/inspec/rule.rb
clintoncwolfe: As I understand it, `@__skip_rule` is a data structure that determines if the control has been marked to be skipped. If `@__skip_rule[:result]` is true, then the runtime will "skip" the tests (by replacing all of them with one dummy `os` check, see `prepare_checks`).  `only_if` uses the same mechanism.

I'm all for altering the fields in that structure to make it more flexible - right now it adds bad messaging, assuming all skips are "Only if." We could also invent a new mechanism.

All of this here, though, only affects the textual messaging associated with the control. For step three, we're wanting new separate fields in the JSON output. And I think - with some code deduplication - it should be the same sort of logic we see in step 2 - read from the Input, then consider if it was expired, was it to be skipped, etc.  I'm not sure how much of the Rule datastructure we have access to during the reporting phase (if any) - clear that's something to look into ASAP.

clintoncwolfe: OK, I've located where we'd stash such data in the RSpec example metadata - we have an opportunity here - https://github.com/inspec/inspec/blob/d1ffce727828853222b80ae018d969b2f6c2bc74/lib/inspec/runner_rspec.rb#L166 - however data structure there does not propagate to the run_data hash passed to the reporter structure. I'm still looking at that.

clintoncwolfe: I found where the RSpec metadata may be propagated to the reporter Hash data structure - here - https://github.com/inspec/inspec/blob/master/lib/inspec/formatters/base.rb#L152 .

At this point we could concentrate all waiver logic in step 2, and only do display in step 3.

lib/inspec/rule.rb
clintoncwolfe: I've updated the PR to now distinguish between the two types of skip; moved the apply_waivers call to ensure it takes precedence; and added functional tests to verify.

lib/inspec/rule.rb
clintoncwolfe: Having slept on it and done some more code reading, I think doing the logic once here in step 2, then attaching the data somewhere (the control metadata) seems the most sensible. I've done so on d118ed62.

lib/inspec/rule.rb
clintoncwolfe: I've cleaned up the way that the messaging is handled, and placed all of the business logic around waivers in step 2. AFAIK in step 3 now we just need to display fields - I've updated the design doc with notes on what should be available in the `run_data` structure.

lib/inspec/rule.rb
zenspider: The above is still trying to do the job that YAML already does:

```
9982 % ruby -ryaml -e 'p YAML.load("---\nkey: yes")'
{"key"=>true}
```

lib/inspec/rule.rb
zenspider: I don't really care as long as it is documented. I would probably assume that "expiration_date" means "on", thus "<".

lib/inspec/rule.rb
zenspider: yaml handles dates:

```
9985 % ruby -ryaml -e 'p YAML.load("---\nkey: 2019-09-01")'
{"key"=>#<Date: 2019-09-01 ((2458728j,0s,0n),+0s,2299161j)>}
```

I'm saying, either the yaml parses or it doesn't... the rest is about invalid data, not parsing.

test/unit/mock/profiles/waivers/basic/files/waivers.yaml
zenspider: **M**VP

zenspider (approved):
I think there's still some stuff going on that doesn't need to (eg stringifying and re-parsing already parsed yaml data)... but this should be complete enough to work on phase 3

lib/inspec/rule.rb
```diff
@@ -42,7 +44,7 @@ def initialize(id, profile_id, opts, &block)
       @__rule_id = id
       @__profile_id = profile_id
       @__checks = []
-      @__skip_rule = {}
+      @__skip_rule = {} # { result: true, message: "Why", type: [:only_if, :waiver] }
```
miah, line 47: Should this be an real object? (not a blocker, just a concern)

lib/inspec/rule.rb
```diff
@@ -206,14 +214,16 @@ def self.merge_changes(rule)
       rule.instance_variable_get(:@__merge_changes)
     end
 
+    # If a rule is marked to be skipped, this
+    # creates a dummay array of "checks" with a skip outcome
```
miah, line 218: s/dummay/dummy

miah (approved):

PR #4433, opened 2019-08-29
Use new logo branding

btm:
Updates the website logos to use our new branding. 

Screenshot from 2019-08-29 16-51-22


miah (approved):

kekaichinose (approved):
We probably should change the general theme back to match Chef's orange-ish swatch, but I suppose that's a longer story :)

btm:
@kekaichinose yeah, there's still an open question as to what we do with the mini-sites, but marketing had asked we update the logos.

kekaichinose:
Roger that. Let's get this in!

PR #4434, opened 2019-08-29
Resolve issue where the Inspec::Tag to_ruby method outputs invalid Ruby

irvingpop:
Signed-off-by: Irving Popovetsky <irving@chef.io>

## Description

The `Inspec::Tag` `to_ruby` method inspects the key, but this produces invalid Ruby.  This PR removes the `.inspect` from the key so it outputs the bare string.

What you get:
```
"cci": "CCI-123456"
```
OR:
```
:cci: "CCI-123456"
```

What you always want:
```
cci: "CCI-123456"
```

## Related Issue

This is affecting some customer-critical work that @TheLunaticScripter and I are working on in `inspec-scap`

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/dsl/objects_test.rb
+++ b/test/unit/dsl/objects_test.rb
@@ -472,10 +472,10 @@ end
       control.id = "tag.control.id"
       control.to_ruby.must_equal '
 control "tag.control.id" do
-  tag "key": "value"
-  tag "key2\'": "value\'"
-  tag "key3\"": "value\""
-  tag "key4": ["a", "b"]
+  tag key: "value"
+  tag key2\': "value\'"
+  tag key3": "value\""
+  tag key4: ["a", "b"]
 end
 '.strip
 
--- a/test/unit/dsl/objects_test.rb
+++ b/test/unit/dsl/objects_test.rb
@@ -454,28 +454,16 @@ end
       tag1.to_hash.must_equal res1
       control.add_tag(tag1)
 
-      res2 = { name: "key2'", value: "value'" }
+      res2 = { name: "key2", value: %w{a b} }
       tag2 = Inspec::Tag.new(res2[:name], res2[:value])
       tag2.to_hash.must_equal res2
       control.add_tag(tag2)
-
-      res3 = { name: "key3\"", value: "value\"" }
-      tag3 = Inspec::Tag.new(res3[:name], res3[:value])
-      tag3.to_hash.must_equal res3
-      control.add_tag(tag3)
-
-      res4 = { name: "key4", value: %w{a b} }
-      tag4 = Inspec::Tag.new(res4[:name], res4[:value])
-      tag4.to_hash.must_equal res4
-      control.add_tag(tag4)
 
       control.id = "tag.control.id"
       control.to_ruby.must_equal '
 control "tag.control.id" do
   tag key: "value"
-  tag key2\': "value\'"
-  tag key3": "value\""
-  tag key4: ["a", "b"]
+  tag key2: ["a", "b"]
 end
 '.strip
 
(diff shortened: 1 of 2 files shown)


irvingpop:
hi @zenspider I updated the tests so they pass now.   However I'm wondering if we should remove these two keys, because they definitely don't output valid Ruby: https://github.com/inspec/inspec/blob/master/test/unit/dsl/objects_test.rb#L457-L465

james-stocks:
Confusingly, the two keys with quote characters in them are valid symbols in the previous code where they had quotes. Ruby will quote the key name:

```
irb(main):016:0> a = { "hel'lo".to_sym => 'world' }
=> {:"hel'lo"=>"world"}
```

So the quotes that are incorrect for basic tags were actually needed for these two unusual tags.

Maybe `Inspec::Tag.new` should replace any non-standard key characters with underscores like

```
    def initialize(key, value)
      @key = key.gsub(/[^A-Za-z0-9_]/, '_')
      @value = value
    end
```

irvingpop:
I agree that quotes in symbols are valid Ruby, maybe my title was a bit extreme there.   But for some reason @TheLunaticScripter and I were noticing that Inspec was bombing when there were quotes even surrounding normal-looking symbols. 

John could you please post the error you were seeing?

TheLunaticScripter:
So I re-ran through and I"m not getting an error with building when the key is a double-quoted string. I think is it more the readability of  `"cci": "foo"`.

james-stocks:
To be clear, I think this PR means that lines [476](https://github.com/inspec/inspec/pull/4434/files#diff-dbdcbc27e3eee8feec09652b04df169dR476) and [477](https://github.com/inspec/inspec/pull/4434/files#diff-dbdcbc27e3eee8feec09652b04df169dR477) are no longer valid Ruby/InSpec.

My previous comment suggested altering the tag string given to Inspec::Tag to ensure it's always valid as a conventional non-quoted tag. 
Or we could add a clever check to see if the quotes are needed or not.
Or we could merge this PR as-is and assume that anyone generating InSpec code with the InSpec::Object classes is not doing anything exotic - they would need to open a subsequent PR if they are (I think that's reasonable, imo).

clintoncwolfe (approved):
My take is that we don't want to slug the tag automatically - I think we have tests that look at round-tripping the code, and that would break those tests, at least.

As-is, by not inspecting, I think we'll usually be right, and if a user has a weird tag name, they can manually slug it to something acceptable - if we need to be more aggressive on that we can be.

miah (approved):
Seems fine, tangentially can we move `inspec/objects/*` over to `inspec-scap` itself? I'm pretty sure nothing in here requires that code.

clintoncwolfe:
inspec/objects/input.rb is definitely needed by InSpec - let's do due diligence if we consider that move.

test/unit/dsl/objects_test.rb
```diff
@@ -472,10 +472,10 @@
...
       control.to_ruby.must_equal '
 control "tag.control.id" do
-  tag "key": "value"
-  tag "key2\'": "value\'"
-  tag "key3\"": "value\""
-  tag "key4": ["a", "b"]
+  tag key: "value"
+  tag key2\': "value\'"
+  tag key3": "value\""
+  tag key4: ["a", "b"]
```
zenspider: This is NOT valid ruby for `key2` and `key3`. So either this fix isn't comprehensive enough, or the test is overkill because we never intend to support such keys/values.

zenspider (changes requested):
I think I'm in agreement with @clintoncwolfe but that the test needs to reflect that.

test/unit/dsl/objects_test.rb
zenspider: @irvingpop I can't make the above call. I assume these test tags are simply invalid and should be removed?

zenspider (approved):
As long as the linux tests pass, this is good and I'll merge it. Ignore any failures on windows in this PR.

Thank you!

PR #4436, opened 2019-08-29
Fixed the formatting of `inspec exec -h` by not using long_desc.

zenspider:
long_desc does some really bad word wrapping stuff that totally breaks
our formatting of the output (like code blocks). So, don't use it. If
we like this, I will apply it to all of them.


miah (approved):
Basically, long_desc is bad :/

clintoncwolfe (approved):
Thanks!

PR #4437, opened 2019-08-29
Rough draft on waiver input

zenspider:
long_desc does some really bad word wrapping stuff that totally breaks
our formatting of the output (like code blocks). So, don't use it. If
we like this, I will apply it to all of them.

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -221,8 +221,8 @@ module Inspec
 
       path = determine_cfg_path(cli_opts)
 
-      cfg_io = File.open(path) if path
-      cfg_io || StringIO.new('{ "version": "1.1" }')
+      ver = KNOWN_VERSIONS.max
+      path ? File.open(path) : StringIO.new({ "version" => ver }.to_json)
     end
 
     def check_for_piped_config(cli_opts)
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -57,6 +57,12 @@ module Inspec
         RunnerRspec.new(@conf)
       end
 
+      if @conf[:waiver_file]
+        waivers = @conf.delete(:waiver_file)
+        @conf[:input_file] ||= []
+        @conf[:input_file].concat waivers
+      end
+
       # About reading inputs:
       #   @conf gets passed around a lot, eventually to
       # Inspec::InputRegistry.register_external_inputs.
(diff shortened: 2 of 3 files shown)


lib/inspec/cli.rb
```diff
@@ -208,9 +208,9 @@ def archive(path)
     pretty_handle_exception(e)
   end
 
-  desc "exec LOCATIONS", "run all test files at the specified LOCATIONS."
-  # TODO: find a way for Thor not to butcher the formatting of this
-  long_desc <<~EOT
```
kekaichinose: I think this change was accidentally linked from #4436 . Can we remove this?

lib/inspec/runner.rb
```diff
@@ -57,6 +57,12 @@ def initialize(conf = {})
         RunnerRspec.new(@conf)
       end
 
+      if @conf[:waiver_file]
+        waivers = cleanse_waivers @conf.delete(:waiver_file)
+        @conf[:input_file] ||= []
+        @conf[:input_file].concat waivers
```
clintoncwolfe, line 63: So, we have a choice here - if waivers are going to be YAML files, very similar to input files,  this will work fine, and we need not pre-process them at all - `cleanse_waivers` won't be needed, AFAIK. The InputRegistry expects `:input_file` to be an array of YAML pathnames, and it will try to read them as such.

So if we choose any other format, we need to either read it ourselves in the InputRegistry, or read it here (which would be surprising, I think).

lib/inspec/runner.rb
```diff
@@ -71,6 +77,12 @@ def initialize(conf = {})
       configure_transport
     end
 
+    def cleanse_waivers(waivers)
+      waivers.reject do |h|
+        # h[:expiry] && h[:reason] && h[:id] && etc ...
```
clintoncwolfe: I don't think there is any filtering to be done here - it's a list of YAML filenames; unless we exclude a whole file.

lib/inspec/config.rb
```diff
@@ -221,8 +221,8 @@ def resolve_cfg_io(cli_opts, cfg_io)
 
       path = determine_cfg_path(cli_opts)
 
-      cfg_io = File.open(path) if path
-      cfg_io || StringIO.new('{ "version": "1.1" }')
+      ver = KNOWN_VERSIONS.max
+      path ? File.open(path) : StringIO.new({ "version" => ver }.to_json)
```
clintoncwolfe, line 225: Unrelated but needed, thanks.

clintoncwolfe (changes requested):
My only concern is with `cleanse_waivers` - I don't think there is any work to be done at this stage. We'd want to look at individual (per-control) waivers elsewhere.

lib/inspec/cli.rb
zenspider: good catch

miah (approved):

clintoncwolfe (approved):
Thanks!

Issue #4438, opened 2019-09-03
Custom Reporting and Resource message handling


kevin-j-smith:
## Detailed Description
The output results coming from Inspec are, at times, not helpful.   For a pass you might see:
should cmp == 3.   But there is not much information around this.  Particularly, what command was ran for the resource and what output/command-result was returned from the target.  Therefore, we have leveraged describe/context/it(s) structure when more context is needed.   The output is not generally readable unless critically massaged.

On top of that, there are times where we run resources to further make conditional decisions on what describe to configure and run.  I can get into the code and figure out what the results are telling me, but I cannot clearly tell from the results what happened in a control to dictate that this particular describe was formed and executed the way it way.   Yes, more context.  Therefore, we have taken our nice clean control logic and added in a bunch of context to placate auditors and executive stakeholders.  Admittedly the results are much easier to read and causes us to have to consult code less often during audits.

Lastly, some of our methods have been to add in new reporting logic.  Something that would be more easily handled if Inspec supported custom reporters.  Or in this case, custom rspec formatters.  How I solved this was to add a custom_base formatter to inspec/formatters.  This class only tracks all formatters that sub-class it.  Then in the formatters I register a method 'name' that maps back to the configuration/cli-flag reporters.   When runner_rspec is being configured it will register all formatters that have been configured and have been registered as custom_base sub class.  Given the ability to load these during plugin registration I have added my custom formatters to some different plugins.  :)

## Context
As a profile writer and software developer, I can fully comprehend the information returned in the results.  And if I cannot, I can track the failure to the code and reproduce the information on the target host.  But, in the end, I am not the auditor or executive stakeholder.  Also, we have corporate vulnerability software/service that tracks other assessments.  We have been charged to supply results into that service (only one of such requests or use stories where reporting to multiple custom services is being requested on a per team basis).

## Possible Implementation
Following this up with a PR.

aaronlippold:
In this case I would use subject and be very prescriptive about what I
wanted to result to report out irrespective to what was tested. if you'd
like some examples or for me to walk through how we're handling this
challenge please feel free to email me at lippold@gmail.com.

Or you can review some of my course material on GitHub.
https://mitre-inspec-developer.netlify.com/

https://mitre-inspec-advanced-developer.netlify.com/

Yours

Aaron

zenspider:
I believe this is a duplicate of #3667. As such, closing.

PR #4439, opened 2019-09-03
Custom rspec formatters and enhanced rspec messaging

kevin-j-smith:
## Description
Needing the ability to enhance reporters to supply more information from resources without breaking the current reporters or pass/fail/skip messages.

## Related Issue
Issue #4438 

## Types of changes

## Checklist:

I do have rake failures, but I did have them from the current master branch.   Also, the default formatter accepts messages and prints them to the console so you will now see all commands and their results.  If you know a nifty way around this in the rake configuration that would be best. :)


kevin-j-smith:
The one failure is due to rpsec default formatter being used vs the inspec formatters that are used during exec.  Not sure what method you wish to use for tests.   Should I specifically ignore these lines in the test or not?

zenspider:
This PR closes #4438, but the problem is that #3667 already exists and specifically talks about making reporters extendable via the plugin system. We have #3667 tentatively scheduled for Q4. 

There are some systemic problems with this PR. First, it invents its own mechanism outside of our existing plugin system. As such, it feels like an additional ownership burden that we don't want to take on and maintain. Second, it does all of this without any testing whatsoever which is, again, an ownership/maintenance burden we don't consent to.

As such, I think I'm going to have to decline.

PR #4443, opened 2019-09-03
Fix for postfix_conf when using a non-standard config location

ramereth:
Pass the first argument of the opts array instead of the whole array which
confuses the SimpleConfig parser. Also added a test which verifies this fixes
the issue.

In addition, change the base_name to something that matches other resource base
names.

## Types of changes

## Checklist:


clintoncwolfe (approved):
Thank you!

zenspider (approved):

Issue #4174, opened 2019-06-03
Improve Windows account support


sam1el:
Inspec should support common OS features for windows as it does in Linux

## Detailed Description
Currently, to check things such as the password age of a specific user, a custom powershell block is needed and you have to parse stdout. In linux we can check against /etc/shadow directly

## Context
as a proserve engineer, having to write custom profiles is quite common. Having a shorthand version of the windows side of things would not only make this process smoother but, would make it easier for the customer to learn if they are already familiar with the linux aspect but need to write for windows as well. 

## Possible Implementation
maybe as part of OS detection in inspec, have a case stystem so if it sees something like 
` its "passowrd' should eq 'secretstuff' ' 
it switches to a windows native command like 
`net user {username}  | find /I "Password last set"` or
 `Get-LocalUser -Name {username} | select PasswordLastSet`

clintoncwolfe:
I'd imagine we'd improve portability on the `user` resource, or similar, to handle password settings.

Are there other specific use cases you are looking to simplify?

sam1el:
i'm building a small list. I am seeing issues with current profiles where, though the expected result exists on the local system in policy and registry, the profile sees it as nil. this could just be there profile. That being said, the user resource is complex enough to likely cover a wide array of resources as is. it's a good start. the sub context of local user grabs group etc. i could see a use case to maybe improve the way it reads the legal text as well for gpo'

mhackethal:
I think that fits to my topic. #4432  . Lets link it.

PR #4452, opened 2019-09-06
User windows fix

### LINKED ISSUE #4432: Resource user partly broken for windows
The user resource executed on a windows target seems to be partly broken
because of wrong usage of cache inside resource user[s].

# Describe the problem

```
You are currently running on:

    Name:      windows_server_2016_standard
    Families:  windows, os
    Release:   10.0.14393
    Arch:      x86_64
inspec> describe user("Win-E01-FA-01") do
inspec>   it { should exist }
inspec>   its('badpasswordattempts') { should be < 2 }
inspec> end
 User Win-E01-FA-01
    [PASS]  should exist
    [FAIL]  badpasswordattempts
    undefined method `badpasswordattempts' for User Win-E01-FA-01:#<Class:0x0000000009b0df98>

```
If i check the created json for local users the json provides the attribute (badpasswordattemps) as well as )mindays,maxdays,warndays and so on.
A inspec check of all these attributes is failing with error "undefined method".
Cross checking the users.rb resource shows that for linux some of attributes are moved into a "credential" cache. I think this breaks the execution on Windows.

Expectation:
its('badpasswordattempts') { should be < 5 } 
its('mindays') { should be < x }
its('maxdays') { should eq x }
and so on

Should not throw an "undefined method".
If i understand the behaviour of the resource wrong any advice is highly appreciated.

Thanks!

Trace:
```
    def mindays
      puts "i was here"
	  puts credentials.nil?
	  puts identity[:mindays]
	  identity[:mindays] unless identity[:mindays].nil?
      credentials[:mindays] unless credentials.nil?	  
    end
```
Output:

> PS D:\temp> inspec exec .\users.rb -l debug
> i was here
> true
> 0
> 

## Possible Solution

Reworks resource users.rb to reflect the "credentials" cache. And supporting offered features also on windows. 
Improve unit testing for user/users resource on windows.

I did some more Research on this Topic and created an updated users.rb:
[https://github.com/mhackethal/inspec/tree/user-windows-fix](https://github.com/mhackethal/inspec/tree/user-windows-fix)

Next to this i harmonized some Attributes 'passwordage' and 'badpasswordattempts' between Windows and Linux.

By Default Linux provides Password age as [Date] but Windows provides it as (days) since last Change. 
First try was to use the Linux date (Aug, 30 2019) but due to localization of the target System it could be Always a different style. Next to this the ruby date class is not a part of the inspec bundle so it is not convertable by ruby.
My solution was to calculate the days by executing the caluclation as inspec.command on the target. This avoids any wrong calculation due to localization of the target.

In Addition the LinuxUser class provides now badpasswordattempts by executing 'lastb' with filter to the account.  On Windows the counter is resetted after every System boot. On Linux it depends on the log Rotation of /var/log/wtmp.

Discussion About this behaviour is very welcome :)
Since its just an idea how i would implement it.

In case this update is wanted and approved i can create a PR but Need some support to pass the unit testings. ( Mock commands and so on ). I don´t have any Linux Environment where i can execute this bundle rake stuff. And on Windows this is not working.

mhackethal:
Provide expected attributes in user resource. Missing usage of cache for windows accounts
reflect windows attributes to be also available for linux.

## Description
It solves: #4432 and parts of #4174 
Expected behaviour on linux and windows should be equal.

## Related Issue

Fixes #4432
#4174 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -118,7 +118,7 @@ module Inspec::Resources
   #   its('warndays') { should eq 5 }
   #   its('passwordage') { should be >= 0 }
   #   its('maxbadpasswords') { should eq nil } // not yet supported on linux
-  #   its('badpasswordattempts') { should eq 0 } 
+  #   its('badpasswordattempts') { should eq 0 }
   # end
   # describe user('Administrator') do
   #   it { should exist }
@@ -132,8 +132,8 @@ module Inspec::Resources
   #   its('maxdays') { should eq 42 }
   #   its('warndays') { should eq nil }
   #   its('passwordage') { should eq 355 }
-  #   its('maxbadpasswords') { should eq 0 } 
-  #   its('badpasswordattempts') { should eq 0 } 
+  #   its('maxbadpasswords') { should eq 0 }
+  #   its('badpasswordattempts') { should eq 0 }
   # end
   #
   # The following  Serverspec  matchers are deprecated in favor for direct value access
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -466,7 +466,7 @@ module Inspec::Resources
       cmd = inspec.command("echo $((($(date +%s) - $(date -d '#{params["Last password change"]}' '+%s'))/86400))")
       dayslastset = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
 
-      cmd = inspec.command("lastb -w -a | awk '{print $1}' | grep #{username} | wc -l")
+      cmd = inspec.command("lastb -w -a | grep #{username} | wc -l")
       badpasswordattempts = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
 
       {
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -464,18 +464,10 @@ module Inspec::Resources
       ).params
 
       cmd = inspec.command("echo $((($(date +%s) - $(date -d '#{params["Last password change"]}' '+%s'))/86400))")
-      if cmd.exit_status != 0
-        dayslastset = nil
-      else
-        dayslastset = convert_to_i(cmd.stdout.chomp)
-      end
+      dayslastset = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
 
       cmd = inspec.command("lastb -w -a | awk '{print $1}' | grep #{username} | wc -l")
-      if cmd.exit_status != 0
-        badpasswordattempts = nil
-      else
-        badpasswordattempts = convert_to_i(cmd.stdout.chomp)
-      end
+      badpasswordattempts = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
 
       {
         mindays: convert_to_i(params["Minimum number of days between password change"]),
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -573,6 +573,31 @@ module Inspec::Resources
       res[0] unless res.empty?
     end
 
+    def meta_info(username)
+      res = identity(username)
+      return if res.nil?
+      {
+        home: res[:home],
+        shell: res[:shell],
+        domain: res[:domain],
+        userflags: res[:userflags],
+      }
+    end
+
+    def credentials(username)
+      res = identity(username)
+      return if res.nil?
+      {
+        mindays: res[:mindays],
+        maxdays: res[:maxdays],
+        warndays: res[:warndays],
+        badpasswordattempts: res[:badpasswordattempts],
+        maxbadpasswords: res[:maxbadpasswords],
+        minpasswordlength: res[:minpasswordlength],
+        passwordage: res[:passwordage],
+      }
+    end
+
     def list_users
       collect_user_details.map { |user| user[:username] }
     end
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -463,7 +463,7 @@ module Inspec::Resources
         group_re: nil,
         multiple_values: false
       ).params
-      
+
       dparse = Date.parse "#{params['Last password change']}"
       dayslastset = (Date.today - dparse).to_i
       cmd = inspec.command("lastb -w -a | grep #{username} | wc -l")
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -116,6 +116,24 @@ module Inspec::Resources
   #   its('mindays') { should eq 0 }
   #   its('maxdays') { should eq 99 }
   #   its('warndays') { should eq 5 }
+  #   its('passwordage') { should be >= 0 }
+  #   its('maxbadpasswords') { should eq nil } // not yet supported on linux
+  #   its('badpasswordattempts') { should eq 0 } 
+  # end
+  # describe user('Administrator') do
+  #   it { should exist }
+  #   its('uid') { should eq "S-1-5-21-1759981009-4135989804-1844563890-500" }
+  #   its('gid') { should eq nil } // not supported on Windows
+  #   its('group') { should eq nil } // not supported on Windows
+  #   its('groups') { should eq ['Administrators', 'Users']}
+  #   its('home') { should eq '' }
+  #   its('shell') { should eq nil } // not supported on Windows
+  #   its('mindays') { should eq 0 }
+  #   its('maxdays') { should eq 42 }
+  #   its('warndays') { should eq nil }
+  #   its('passwordage') { should eq 355 }
+  #   its('maxbadpasswords') { should eq 0 } 
+  #   its('badpasswordattempts') { should eq 0 } 
   # end
   #
   # The following  Serverspec  matchers are deprecated in favor for direct value access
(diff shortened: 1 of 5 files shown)


lib/inspec/resources/users.rb
```diff
@@ -425,10 +463,26 @@ def credentials(username)
         multiple_values: false
       ).params
 
+      cmd = inspec.command("echo $((($(date +%s) - $(date -d '#{params["Last password change"]}' '+%s'))/86400))")
```
miah: Do we need to do math in a `inspec.command`?

lib/inspec/resources/users.rb
```diff
@@ -425,10 +463,26 @@ def credentials(username)
...
+      if cmd.exit_status != 0
+        dayslastset = nil
+      else
+        dayslastset = convert_to_i(cmd.stdout.chomp)
+      end
+
+      cmd = inspec.command("lastb -w -a | awk '{print $1}' | grep #{username} | wc -l")
```
miah: I think we could just do `lastb -w -a #{username}|wc -l`

But also, we'll need to subtract 2 from that number. As we'll have a empty line and informational line about btmp.

```
$ sudo lastb -w -a miah
miah     ssh:notty    Fri Sep  6 11:12 - 11:12  (00:00)     ::1
miah     ssh:notty    Fri Sep  6 11:12 - 11:12  (00:00)     ::1
miah     ssh:notty    Fri Sep  6 11:12 - 11:12  (00:00)     ::1

btmp begins Fri Sep  6 11:12:07 2019
```

We'll also need to note that this is not a cumulative number of all bad login attempts from that user. Its only since the creation of that btmp file; because this stores all bad login attempts systems on the public internet might refresh this more often. Some distro's might even recreate this file on reboot. I don't have all the details on the lifecycle of that file, but it might be good to consider.

miah (changes requested):
Thanks for the contribution @mhackethal =)

I would love to add this functionality, thanks for starting the work on it!

We need tests and documentation before we can merge this work in. If you have any questions please don't hesitate to ask.

lib/inspec/resources/users.rb
mhackethal: hi miah, i just pulled this request because no one answered to my "discussion" Topic :)
#4432.
I opened a bug and described why i did math calculation on the target. The reason for doing it on the target is the localization. Each target could have a different localization and therefor ist close to impossible to calculate it by ruby. Next to this the ruby "date" class is not available from inside inspec. Same issue as for some AWS ressources.

I am totally open minded at this Topic. Any discussion and ideas are welcome and should be documented issue #4432.

lib/inspec/resources/users.rb
mhackethal: Same as your previous comment -> discussion should be done on issue #4432. I noted that specifics down but there is now way to determinate an all time bad Login Count. Because of the log Rotation and Retention Policy. This Count is Always depending on the local System Setup. (Same like on Windows). But every Administrator should be Aware of this. New ideas or any comment is welcome.

mhackethal:
> Thanks for the contribution @mhackethal =)
> I would love to add this functionality, thanks for starting the work on it!
> We need tests and documentation before we can merge this work in. If you have any questions please don't hesitate to ask.

Ill provide the requested documentation as soon as we have an allignment of the provided features. As i tried to explain i didnt expected to get this pull request approved. I did it to start a discussion and receive feedback. At first sight i just wanted to fix the "caching" issue for Windows but after taking a Closer look into the resource i felt like it requires some more improvements.

lib/inspec/resources/users.rb
mhackethal: Just some additional remarks. I think its better to keep go with "awk&grep" on your System i have to calculate "-2" but Maybe on another distro there are 3 empty lines are an additional info line. This could break the calculation. By using awk and grep i can ensure that it only Counts the lines of badloginattempts. The Performance should have no effect in that case.

zenspider: I don't see any need for the `awk` part. I'd go for grep + wc or even `grep -c` (but I'm not sure how universal the count flag is).

zenspider: And discussion about code should go next to the code so we know what we're talking about.

lib/inspec/resources/users.rb
```diff
@@ -425,10 +463,26 @@ def credentials(username)
...
+      if cmd.exit_status != 0
+        badpasswordattempts = nil
+      else
+        badpasswordattempts = convert_to_i(cmd.stdout.chomp)
+      end
```
zenspider:
```ruby
# lift assignment
badpasswordattempts = if cmd.exit_status != 0
                        nil
                      else
                        convert_to_i(cmd.stdout.chomp)
                      end

# inline conditional
badpasswordattempts = cmd.exit_status != 0 ? nil :convert_to_i(cmd.stdout.chomp)

# simplify
badpasswordattempts = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
```

lib/inspec/resources/users.rb
zenspider: "localization" is the reason why you subtract and divide by 86400 on the client?

lib/inspec/resources/users.rb
```diff
@@ -573,6 +629,31 @@ def identity(username)
...
+      {
+        mindays: res[:mindays],
+        maxdays: res[:maxdays],
+        warndays: res[:warndays],
+        badpasswordattempts: res[:badpasswordattempts],
+        maxbadpasswords: res[:maxbadpasswords],
+        minpasswordlength: res[:minpasswordlength],
+        passwordage: res[:passwordage],
+      }
+    end
```
zenspider, line 647: These new methods need tests.

zenspider (changes requested):

lib/inspec/resources/users.rb
mhackethal: There are 3 reason for calculation. 
1) Localization of the client 
2) harmonize output to windows. 
3) Ruby date class not a part of "inspec" package.

lib/inspec/resources/users.rb
mhackethal: Thanks for this ruby expert tips :) 
i will do the requested updates.

lib/inspec/resources/users.rb
mhackethal: Ill check if can find where to put that. Again thanks!

lib/inspec/resources/users.rb
clintoncwolfe: Hi @mhackethal ,
  [`date`](https://ruby-doc.org/stdlib-2.6.4/libdoc/date/rdoc/index.html), is the standard Ruby library that includes the `Date` class, and as such, it is included with every Ruby installation. You can (and should) do any conversion in Ruby.

zenspider: I've got to push back on this one.

1. Too complicated.
2. `date -d` doesn't do what you think it does (on all platforms).
3.  Ruby can do all the conversions for you:

```ruby
>> require "date"
=> true
>> d0 = Date.parse "2019-01-01"
=> #<Date: 2019-01-01 ((2458485j,0s,0n),+0s,2299161j)>
>> Date.today - d0
=> (266/1)
```

zenspider (changes requested):

lib/inspec/resources/users.rb
mhackethal: i am sorry. Youre totally right. Looks like i did something wrong on my end. (big sorry)
On my first tests i received an error when i used require "date" so i thought its not usable.
But looks like i did something wrong - its working now.
```

      #cmd = inspec.command("echo $((($(date +%s) - $(date -d '#{params["Last password change"]}' '+%s'))/86400))")
      #dayslastset = convert_to_i(cmd.stdout.chomp) if cmd.exit_status == 0
      dparse = Date.parse "#{params['Last password change']}"
      dayslastset = (Date.today - dparse).to_i
```
I changed the the code and tested on:
RHEL 6 , 7
Ubuntu 12.x,14,x,16,x
Debian 9.x
Debian 10.x

```
You are currently running on:
    Name:      redhat
    Release:   6.10
    Arch:      x86_64
inspec> describe user("root") do
inspec>   its('passwordage') { should eq 0 }
inspec> end

Profile: inspec-shell
Version: (not specified)
  User root
     [FAIL]  passwordage should eq 0
     expected: 0
          got: 9
     (compared using ==)
```
```
You are currently running on:
    Name:      ubuntu
    Release:   16.04
    Arch:      x86_64
inspec> describe user("root") do
inspec>   its('passwordage') { should eq 0 }
inspec> end
Profile: inspec-shell
Version: (not specified)
  User root
     [FAIL]  passwordage should eq 0
     expected: 0
          got: 76
     (compared using ==)
```

I tested both ways onetime with the "cmd" parser and one time with the ruby parser. On my systems the result was equal on both.
I personally would prefer the "cmd" way because i have the following situation in my mind:
- InSpec uses : dd/mm/yyyy
- Target uses : mm/dd/yyyy
Same for yyyy/mm/dd vs. yyyy/dd/mm

Any hint on which distro "date -d" will not do what i expect it to do? (just for personal intrest on this topic).

I will push an update to the code later. Thank you for your time and supporting me on this topic!

mhackethal: All done and updated.
Does anyone can give me some suggestion about CodeClimat and the failing BK Test for Windows?
Is this something i should solve?

kekaichinose: @miah any feedback for @mhackethal on this comment?

miah: I've reviewed the codeclimate complaints here and they are valid, but don't need to be addressed by this PR.

miah (approved):
Thanks @mhackethal!

lib/inspec/resources/users.rb
mhackethal: Can i consider this as no More Tasks for me?

miah: Depends on the other reviewers =)

zenspider (approved):

clintoncwolfe (approved):

Issue #137, opened 2015-10-22
verifiy input parameter of `user` resources


chris-rock:
```
describe user(nil) do
...
end
```

should throw an error and fail the test

arlimus:
Agreed! This is also a bit broad, changing the title to focus on the `user` resource.

adamleff:
I think this is another good case of where resources should be able to throw exceptions that descend from one of two classes (`Inspec::Exceptions::ResourceSkipped` or `Inspec::Exceptions::ResourceErrored`) with a message and properly show that in the appropriate output formatter. Yet another reference to #1205 :)

pwelch:
Closing this issue because there has been no traction on it for a while or it is no longer relevant. If this issue is still relevant, please reopen. Thanks!

Looks like this was addressed on #1205

PR #4462, opened 2019-09-10
Fix Markdown markup for the aws_security_group resource

multani:
## Description
The formatting of the [aws_security_group](https://www.inspec.io/docs/reference/resources/aws_security_group/) resource has several bugs. The "Matchers" section is particularly affected.

## Types of changes

## Checklist:

### DIFF
--- a/docs/resources/aws_security_group.md.erb
+++ b/docs/resources/aws_security_group.md.erb
@@ -28,7 +28,7 @@ This resource first became available in v2.0.16 of InSpec.
 
 ## Syntax
 
-Resource parameters: group_id, group_name, id, vpc_id
+Resource parameters: `group_id`, `group_name`, `id`, `vpc_id`
 
 An `aws_security_group` resource block uses resource parameters to search for and then test a Security Group. If no SGs match, no error is raised, but the `exists` matcher returns `false`, and all scalar properties are `nil`. List properties returned under these conditions are empty lists. If more than one SG matches (due to vague search parameters), an error is raised.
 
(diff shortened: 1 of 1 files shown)


multani:
Also, it's not clear whether this documentation should be fix in the https://github.com/inspec/inspec or the https://github.com/inspec/inspec-aws repository?

clintoncwolfe:
It should definitely be submitted to the inspec/inspec-aws repo. Thanks!

multani:
@clintoncwolfe I'd be glad to port the patch to inspec-aws but AFAIK, the current documentation at https://inspec.io still uses the Markdown from inspec/inspec, right?
For example, the content of the same doc in the inspec-aws repository [contains a properties table](https://github.com/inspec/inspec-aws/blob/master/docs/resources/aws_security_group.md#properties) which is not in https://www.inspec.io/docs/reference/resources/aws_security_group/

This is a bit confusing, it looks like if I move this patch to inspec-aws, the official documentation will still be partly broken... :/

clintoncwolfe:
So, the way this is *supposed* to work is that the inspec-aws version is the most accurate; when we update the inspec.io website we pull their markdown files and publish them; and we only keep the old aws resource files in core for people who did not update their dependencies in inspec 4.

Issues #4471 and #4470 and your #4463 all indicate that we need to work on the publishing system. All that said, the right place for any info you have is inspec/inspec-aws's markdown files.

kekaichinose:
@multani as we work to address the publishing system issues that @clintoncwolfe mentioned above, are you okay with us moving this issue over to inspec/inspec-aws project to be addressed?

Unless there's any objects, we'll just go ahead and migrate it over there for you.

clintoncwolfe:
Issue moved to [inspec/inspec-aws #137](https://github.com/inspec/inspec-aws/issues/137) via [**ZenHub**](https://www.zenhub.com/)

PR #4475, opened 2019-09-11
Update rubygems and use gem env for system details.

zenspider:
Travis used to update rubygems to the latest version before doing
anything else. We don't on BK and I just found a rubygems specific bug
that would have been exposed by this.


miah (approved):

miah (approved):

clintoncwolfe (approved):
Thanks!

PR #4483, opened 2019-09-12
Fix config clearing

### LINKED ISSUE #4481: Inspec::Config not cleared when using the Inspec::Runner API
`Inspec::Config.cache` is re-used at multiple places, this causes an issue with running profiles over a loop/simultaneously using the `Inspec::Runner` API, as the `cached_config` instance variable from `Inspec::Config` is never cleared or there is a method to clear it like `__reset` present in `Inspec::InputRegistry`.

# Describe the problem
I have a profile named **normal** and another profile named **dependent** which depends on the **normal profile**. There is an `input` named *region* in the **normal** profile with a default value of `us-gov-west-1`. 

The respective control file in question:

`~/work/inspec/normal/controls/region.rb`
```ruby
region = attribute('aws_region', default: 'us-gov-west-1', description: 'Default AWS region')

control 'aws_region' do
  title 'The default AWS region must be set using an environment variable.'

  describe command('echo ${AWS_DEFAULT_REGION}') do
    its('stdout') { should match(/^[\s]*#{region}[\s]*$/) }
  end
end
```

Now the **dependent** profile depends on the **normal** profile and just have this in controls, basically loading the meta profile.

`~/work/inspec/dependent/controls/dependent.rb`
```ruby
include_controls 'normal-foo' do
end
```

This is the code used to execute the profiles:

```ruby
# frozen_string_literal: true

require 'inspec'

opts = {
  'color' => true
}

def inspec_runner(profile, opts)
  runner.add_target(File.expand_path(profile))
  runner.run
  runner.reset
end
```

 I have `AWS_DEFAULT_REGION` set as `us-east-1` in my environment.

Calling `inspec_runner('~/work/inspec/dependent', opts)` fails as expected

Calling `inspec_runner('~/work/inspec/dependent', opts.merge(attributes: { 'region' => 'us-east-1' }))` passes as expected

but calling

```ruby
inspec_runner('~/work/inspec/dependent', opts)
inspec_runner('~/work/inspec/dependent', opts.merge(attributes: { 'region' => 'us-east-1' }))
```

The first and second test fails which was not the expected case. The second test should have passed (Second test passes in Inspec v3.x). Digging into the code, it seems the `Inspec::Config.cache` is used here: https://github.com/inspec/inspec/blob/master/lib/inspec/profile.rb#L117 and here: https://github.com/inspec/inspec/blob/master/lib/inspec/dependencies/requirement.rb#L121.

What happens in that the cached `:runner_conf` is being passed to the second invocation (which doesn't have the `:inputs` options passed in) of the `Runner` API. I have added `Inspec::Config.instance_variable_set(:@cached_config, nil)` at the end of `inspec_runner` method to work around this.

## Possible Solution
Have a reset method to clear `Inspec::Config` and call that after `Inspec::Runner.run` is finished.

If my proposed solution is good I can raise a PR with the changes.

frezbo:
Fixes #4481

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -135,6 +135,11 @@ module Inspec
       Thor::CoreExt::HashWithIndifferentAccess.new(@plugin_cfg[plugin_name] || {})
     end
 
+    # clear the cached config
+    def self.__reset
+      @cached_config = nil
+    end
+
     private
 
     def _utc_merge_transport_options(credentials, transport_name)


miah (changes requested):
Could you take a look at the tests?

lib/inspec/runner.rb
```diff
@@ -126,6 +126,8 @@ def run(with = nil)
       Inspec::Log.debug "Starting run with targets: #{@target_profiles.map(&:to_s)}"
       load
       run_tests(with)
+      # clear config after executing profile
+      Inspec::Config.__reset
```
clintoncwolfe: I disagree with adding this line here.  I think if you are a consumer of `Inspec::Runner`, you can call `Inspec::Config.__reset` yourself after calling `run`. I'm not sure it makes sense to always reset the config.

clintoncwolfe (changes requested):
Thanks for this - I think it's a great idea. I'm OK with adding a reset functionality to Inspec::Config, but I think using it should be at-will - just because you call the Runner.run doesn't mean the config should get reset.

lib/inspec/runner.rb
frezbo: yes I agree with you. I don't think kitchen-inspec needs these changes as it will always run one profile at a time I guess.

clintoncwolfe (approved):
Ok, with just an addition of a reset to Config, and no obligatory call to it, I think this is a fine addition - many thanks!

lib/inspec/runner.rb
zenspider: Agreed. But looks resolved.

zenspider (approved):

miah (approved):

PR #4485, opened 2019-09-13
Split out Inspec::Input functional code from the code generation code.

### LINKED ISSUE #4456: Split lib/inspec/objects/inputs.rb in two
That file is 14kb where all the other files in there are < 3kb.

I think the non-ruby-code-generation side of this file should be split out. For all I care, it can stay in the same namespace entirely, but the working bits should physically move.

I'm also filing #4455 at the same time, which is to move this code out of the project. That can't get done until this is done.

### LINKED ISSUE #4473: Pull input / input registry errors out of errors.rb
errors.rb should be generics only

zenspider:
Includes moving errors into Inspec::Input and Inspec::InputRegistry.

Fixes #4456.
Fixes #4473.


miah (changes requested):
`File does not exist: inspec/input`

zenspider:
I R DUM

clintoncwolfe (approved):
Thanks, a clean move!

miah (approved):

PR #4490, opened 2019-09-16
Split and parallelize some tests to make them faster

zenspider:
This takes the run time from ~3 minutes to ~1 minute (in parallel).


zenspider:
I want to do a few more functionals... Do not merge yet please.

miah (approved):
yes. 100% want faster functionals.

clintoncwolfe (approved):
The file-splitting is definitely an improvement :-) and there are a lot of places where the parallelism should be straightforwardly beneficial. I have some reservations about making the plugin tests parallel (though they need it the worst) but that just may be FUD. I'm OK with moving forward, and if we see instability down the road, we'll deal with it then.

zenspider:
@clintoncwolfe the instability that we saw was mostly from units and globals. Since these are functionals and are all shelling out, they _shouldn't_ be able to to infect each other. We might still have problems where we're not using `tmpdir` instead of the actual mock dirs.

PR #4501, opened 2019-09-18
Habitat plan for Windows

miah:
A Windows habitat plan.ps1 for InSpec!

Heavily cribbed from @robbkidd :1st_place_medal: work on [Infra](https://github.com/chef/chef/pull/8900).

### DIFF
--- a/inspec-bin/Rakefile
+++ b/inspec-bin/Rakefile
@@ -1,0 +1,10 @@
+Bundler::GemHelper.install_tasks(name: "inspec-bin")
+
+desc "force install the inspec-bin gem"
+task "install:force" do
+  sh "gem build -V inspec-bin.gemspec"
+  built_gem_path = Dir["inspec-bin-*.gem"].sort_by { |f| File.mtime(f) }.last
+  FileUtils.mkdir_p("pkg") unless Dir.exist?("pkg")
+  FileUtils.mv(built_gem_path, "pkg")
+  sh "gem install -f pkg/#{built_gem_path}"
+end
--- a/test/artifact/Rakefile
+++ b/test/artifact/Rakefile
@@ -1,0 +1,5 @@
+require "rake/testtask"
+
+Rake::TestTask.new(:default) do |t|
+  t.test_files = Dir["test/artifact/*_test.rb"].sort
+end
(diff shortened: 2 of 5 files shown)


miah:
I'm flipping the bit on this, lets get it merged. 

There is still work to do with the testing pipeline, and there are probably lots of bugs to be discovered. Definitely don't consider this supported at this point =)

test/artifact/inspec_detect_test.rb
```diff
@@ -0,0 +1,12 @@
+require "minitest/autorun"
+
+class TestArtifactDetect < Minitest::Test
+  def test_detect
+    out, err = capture_subprocess_io {
+      assert system("inspec detect --no-color")
+    }
+
+    assert_match %r%Platform Details%, out
```
zenspider: won't chefstyle get pissy about using `%r` when it doesn't have `/` in the pattern?

Huh. apparently not. But it does  blow up on the use of curlies on the assertion above. I agree with the curlies. so you might want to disable that check. I think that's the only blocker to getting the tests running.

zenspider (commented):
minor feedback about BK being red on chefstyle

inspec-bin/Rakefile
```diff
@@ -0,0 +1,10 @@
+Bundler::GemHelper.install_tasks(name: "inspec-bin")
+
+desc "force install the inspec-bin gem"
+task "install:force" do
+  sh "gem build -V inspec-bin.gemspec"
+  built_gem_path = Dir["inspec-bin-*.gem"].sort_by { |f| File.mtime(f) }.last
```
zenspider, line 6:
```ruby
[...].sort_by { |f| File.mtime(f) }.last
```
vs
```ruby
[...].max_by { |f| File.mtime(f) }
```

test/artifact/Rakefile
```diff
@@ -0,0 +1,7 @@
+require "rake/testtask"
+
+Rake::TestTask.new(:test) do |t|
+  t.test_files = FileList["*_test.rb"].sort
```
zenspider, line 4: thank you for sorting this... file systems are strange beasts.

clintoncwolfe (commented):
I'm OK with approving everything except the plan PowerShell itself - I just don't know enough PowerShell, let alone Habitat plan idioms, to be a competent reviewer here. Could we get someone who works on windows plans frequently to review this and give feedback?

PR #4502, opened 2019-09-18
Fix input hash processing on inheritance

### LINKED ISSUE #4504: `to_hash` error when running inherited profile with an input
The following error is output when running with an inherited profile:

```
/opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.17.3/lib/inspec/profile.rb:319:in map': undefined method to_hash' for #<Inspec::Input:0x0000000004d67f28> (NoMethodError)
```

# Describe the problem

Using, `inspec` from the Chef YUM repos, this evidenced itself in version `4.16.15`

The code used for inheritance was trivial:

```
require_controls 'disa_stig-el7' do
  @conf['profile'].info[:controls].each do |ctrl|
    tags = ctrl[:tags]
    if tags && tags[:subsystems]
      unless tags[:subsystems].empty?
        control ctrl[:id]
      end
    end
  end
end
```

## Possible Solution

#4502 (extremely naive but does mitigate the issue)

trevor-vaughan:
When attempting to use an inherited profile, I was getting an error
indicating that `to_hash` could not be performed on an Inspec::Input
object.

This patch only calls the to_hash method if it is available on the
corresponding object.

No differences have been found in the output or processing.

EDIT (@clintoncwolfe): Fixes #4504

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -316,7 +316,9 @@ module Inspec
         # convert to array for backwards compatability
         res[:inputs] = []
       else
-        res[:inputs] = res[:inputs].values.map(&:to_hash)
+        res[:inputs] = res[:inputs].values.map do |input|
+          input.respond_to?(:to_hash) ? input.to_hash : input
+        end
       end
       res[:sha256] = sha256
       res[:parent_profile] = parent_profile unless parent_profile.nil?


zenspider:
I'm not sure this is the right response... I suspect that I might have caused the bug in e375ea30c8 but that isn't in the release you said this bug appears in...

trevor-vaughan:
@zenspider Honestly, it's probably not the right response but it definitely functions properly based on usage. I didn't dig deeper than "make it work again".

trevor-vaughan:
For completeness, the failing release in the Chef YUM repos is `4.16.15`

clintoncwolfe (commented):
Just a quick comment to mention this is on my radar; I want to get to the bottom of why some inputs comeback with no to_hash method, rather than just accept that they will sometimes be missing. Thanks!

trevor-vaughan:
@clintoncwolfe If there's anything I can test, throw it at me!

clintoncwolfe:
OK, I think we have a fix - we shouldn't have removed the `to_hash` method yet. It's replaced on #4545 - I'm going to close this. Thanks for contributing!

PR #4513, opened 2019-09-24
Fixed Inspec::Plugin::V2::Registry#detect_system_plugins

zenspider:
+ Falls back to checking for inspec-core if inspec not found.
+ Fails gracefully if neither are found.

### DIFF
--- a/test/unit/plugin/v2/loader_test.rb
+++ b/test/unit/plugin/v2/loader_test.rb
@@ -1,9 +1,13 @@
 # Unit tests for Inspec::PluginLoader and Registry
 
-require "minitest/autorun"
-require_relative "../../../../lib/inspec/plugin/v2"
+require "helper"
+require "inspec/plugin/v2"
 
 require "train" # Needed for Train plugin testing
+
+class Inspec::Plugin::V2::Loader
+  public :detect_system_plugins
+end
 
 class PluginLoaderTests < Minitest::Test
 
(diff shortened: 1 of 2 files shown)


miah (approved):

zenspider:
Paired with @miah for review.

test/unit/plugin/v2/loader_test.rb
```diff
@@ -244,4 +248,82 @@ def test_when_a_train_plugin_is_installed_via_gem_and_required
...
+    assert_match exp_err, log_out
+  end
+
+  def test_detect_system_plugins
+    with_empty_registry do
+      exp = []
+      exp_err = "inspec gem not found, skipping detecting of system plugins\n"
+
+      assert_detect_system_plugins exp, exp_err do |loader|
+        def loader.find_inspec_gemspec(*)
```
clintoncwolfe, line 308: This feels like re-inventing mocking. Is there a reason the implementation of Inspec::Plugin::V2::Loader.find_inspec_gemspec can't be mocked?

test/unit/plugin/v2/loader_test.rb
```diff
@@ -244,4 +248,82 @@ def test_when_a_train_plugin_is_installed_via_gem_and_required
...
+      end
+    end
+
+    assert_equal exp_keys, REG_INST.registry.keys.sort
+    assert_match exp_err, log_out
+  end
+
+  def test_detect_system_plugins
+    with_empty_registry do
+      exp = []
```
clintoncwolfe, line 304: It would have helped my understanding if this was named `expected_plugins`, which it later turned out to mean.

clintoncwolfe (commented):
I think the implementation code is straightforward and solid. The testing code seems overcomplicated, and while I am sure there is a good reason for that, it may need to be explained to me. Comments and longer/more clear variable names might help explain why a more typical mocking approach could not be used.

I'm OK with this going in as-is if we need to ship but if there is time some commentary would be great.

test/unit/plugin/v2/loader_test.rb
zenspider: This is stubbing, the ruby way. Not re-invented, since this has existed in ruby since before mocking frameworks existed (literally).

Issue #4514, opened 2019-09-25
website header content overlays the main navigation


SalmonKing72:
# Describe the problem
* When scrolling on the desktop website, the header content overlays the main navigation
* When clicking the hamburger menu to open the mobile navigation, the header content overlays the navigation preventing it from being used.

## Possible Solution
update the sass to add a lower or negative z-index value for the `header`

james-stocks:
Sounds like a duplicate of https://github.com/inspec/inspec/issues/4331

What browser are you using @salmonking72 ? #4331 was for Chrome on Android.

SalmonKing72:
@james-stocks The mobile portion of this is probably duplicative of #4331. I saw the issue still occurring on Safari for iOS, also Firefox, Safari, and Chrome Canary on macOS. The z-index update in my PR related to this fixed both the desktop issue I described and the mobile issue described in #4331. 

**Chrome**
![image](https://user-images.githubusercontent.com/6287664/65612212-e65adb00-df81-11e9-8689-ae4b7ea4cbee.png)

**Firefox**
<img width="1584" alt="image" src="https://user-images.githubusercontent.com/6287664/65612408-32a61b00-df82-11e9-8246-3bd898baec50.png">

**Safari**
![image](https://user-images.githubusercontent.com/6287664/65612477-52d5da00-df82-11e9-95a5-198f9fe0c453.png)

james-stocks:
I think this was fixed by https://github.com/inspec/inspec/pull/4515 (but I messed up linking the two :) )

kekaichinose:
Closed by #4515

PR #4515, opened 2019-09-25
Fix issue 4514 - update header z-index preventing header overlaying main navigation

SalmonKing72:
## Description
Currently the header content scrolls in front of the main navigation of the website. It's especially impacting the mobile layout navigation (when clicking the hamburger menu button to display the navigation only to find it drop down behind the header). I've updated the z-index value on the `header` style so that it displays behind the main navigation. I noticed there's a couple of general classes used for setting specific z-index values on any element, so I'm happy to create another one for setting `z-index: -1;` if that's what you all see fit. But since it's specific to the header, I put the change on the `header` style.

## Types of changes

## Checklist:


james-stocks:
Fixes #4514

zenspider:
merging as obvious fix style fix

PR #4363, opened 2019-08-06
Add startuser windows service

mbaitelman:
Signed-off-by: Mendy Baitelman <mendy@baitelman.com>

Add startuser to Windows Resource

## Description
Addresses #421  
When creating windows services you can specify the user it should run as. 
This feature allows checking that the correct user is used.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:
- [X] I have read the **CONTRIBUTING** document.

PR #4367, opened 2019-08-07
Add startuser systemd service

mbaitelman:
Add a check to see which user a systemd resource is running under

Depends on #4363

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:
- [X] I have read the **CONTRIBUTING** document.

PR #4516, opened 2019-09-25
Add startuser systemd service v2

mbaitelman:
Add a check to see which user a systemd resource is running under

Depends on #4363

## Types of changes

## Checklist:


mbaitelman:
@clintoncwolfe I had to move the pull from #4367 as I had deleted my fork and had to recreate it to take ownership.

lib/inspec/resources/service.rb
```diff
@@ -304,10 +304,12 @@ def info(service_name)
 
       # LoadState values eg. loaded, not-found
       installed = params["LoadState"] == "loaded"
+      startname = params["User"] != "" ? params["User"] : nil
```
miah: Could we do this instead:

`startname = params["User"]`

I think it would be nicer to have a empty string since we're expecting a string if its set.

miah (changes requested):

clintoncwolfe (commented):
I agree with @miah re: the default for startname

lib/inspec/resources/service.rb
zenspider: Agreed. This complexity isn't necessary.

zenspider (changes requested):
I'm in agreement with miah.

lib/inspec/resources/service.rb
mbaitelman: Updated

miah (approved):
Looks great. Thanks!

mbaitelman:
Buildkite failed but Im not able to see the details.

zenspider:
I think if you rebase against master then you'll be green.

clintoncwolfe (approved):

miah:
@mbaitelman Thanks. Your code looks good, could you just rebase your branch against master? That should clear up the `buildkite/inspec-inspec-master-verify` build, when thats green we'll merge.

Ignore `buildkite/inspec-inspec-master-verify-resources` for now though =)

miah:
Looks great thanks!

PR #4518, opened 2019-09-25
Added first draft of waiver doco.

zenspider:

### DIFF
--- a/docs/waivers.md
+++ b/docs/waivers.md
@@ -1,0 +1,54 @@
+# Waivers (Beta)
+
+Waivers is a mechanism to mark controls as "waived" for various reasons
+and to control the running and/or reporting of those controls. It
+takes a YAML input file that identifies which controls are waived,
+(optionally) whether they should be skipped from running, a description of why it
+is waived, and (optionally) an expiration date for the waiver.
+
+NOTE: This mechanism is currently in beta and is ultimately meant to be
+ingested and maintained by Chef Automate UI.
+
+## Usage
+
+To use waivers, you must have a correctly formatted input file and
+invoke `inspec exec` with `--waiver-file [path]`. Currentlty, waiver results are
+only reported with the JSON reporter type, so you must also have
+`--reporter json`. This may change in the future.
+
+```
+% inspec exec --waiver-file waivers.yaml --reporter json -- path/to/profile
+```
+
+## File Format
+
+Waiver files are [input files](inputs.html "inputs files") with a
+specific format:
+
+```yaml
+control_id:
+  expiration_date: YYYY-MM-DD
+  skip: true
+  justification: "reason for waiving this control"
+```
+
++ `expiration_date` is optional. Absence means the waiver is permanent.
++ `skip` is optional. Absence means the control will run and be
+  reported, but failures in it won't make the overall run fail.
++ `justification` can be any text you want and might include a reason
+  as well as who signed off on the waiver.
+
+### Example:
+
+```yaml
+waiver_control_1_2_3:
+  expiry_date: 2019-09-15
+  justification: Not needed until Q3. @secteam
+
+xccdf_org.cisecurity.benchmarks_rule_1.1.1.4_Ensure_mounting_of_hfs_filesystems_is_disabled:
+  expiry_date: 2020-03-01
+  justification: "This might be a bug in the test. @qateam"
+  skip_run: true
+
+# ... and so on ...
+```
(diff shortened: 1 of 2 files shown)


miah (approved):

docs/waivers.md
```diff
@@ -0,0 +1,54 @@
...
+is waived, and (optionally) an expiration date for the waiver.
+
+NOTE: This mechanism is currently in beta and is ultimately meant to be
+ingested and maintained by Chef Automate UI.
+
+## Usage
+
+To use waivers, you must have a correctly formatted input file and
+invoke `inspec exec` with `--waiver-file [path]`. Currentlty, waiver results are
+only reported with the JSON reporter type, so you must also have
```
clintoncwolfe: Waiver information is at least partially visible in the default CLI output as well.

docs/waivers.md
```diff
@@ -0,0 +1,54 @@
...
+`--reporter json`. This may change in the future.
+
+```
+% inspec exec --waiver-file waivers.yaml --reporter json -- path/to/profile
```
clintoncwolfe: While this is more robust, InSpec docs more frequently give invocations in the order:
`inspec exec PROFILE OPTIONS` and omit the end-of-options marker `--`. It can be visually confusing (is there a space between `-- path/to/profile`)?

docs/waivers.md
```diff
@@ -0,0 +1,54 @@
...
+
+```yaml
+control_id:
+  expiration_date: YYYY-MM-DD
+  skip: true
+  justification: "reason for waiving this control"
+```
+
++ `expiration_date` is optional. Absence means the waiver is permanent.
++ `skip` is optional. Absence means the control will run and be
```
clintoncwolfe, line 34: This appears to be a true/false value, but the docs address it as a present / not present value.

docs/waivers.md
```diff
@@ -0,0 +1,54 @@
...
+Waiver files are [input files](inputs.html "inputs files") with a
+specific format:
+
+```yaml
+control_id:
+  expiration_date: YYYY-MM-DD
+  skip: true
+  justification: "reason for waiving this control"
+```
+
```
clintoncwolfe, line 32: I think there should probably be an entry here for control_id.

docs/waivers.md
```diff
@@ -0,0 +1,54 @@
+# Waivers (Beta)
+
+Waivers is a mechanism to mark controls as "waived" for various reasons
+and to control the running and/or reporting of those controls. It
+takes a YAML input file that identifies which controls are waived,
+(optionally) whether they should be skipped from running, a description of why it
+is waived, and (optionally) an expiration date for the waiver.
+
+NOTE: This mechanism is currently in beta and is ultimately meant to be
+ingested and maintained by Chef Automate UI.
```
clintoncwolfe: I think Product can speak more to this, but the feature is certainly exposed in such a way that is not bound to Automate. I'm not sure the docs should limit it to that scope.

clintoncwolfe (changes requested):
Great overall; main change I would suggest is to run `exec` without the `--reporter` option but with a `--waiver-file` and observer the messaging - it is broader than just for Automate consumption, I think.

docs/waivers.md
zenspider: the end of options marker was there because of the way that thor consumes args when a flag is labelled as an array. I hate it. But I'm removing all mention of json and reporters so that can go away as well.

docs/waivers.md
```diff
@@ -1,32 +1,30 @@
...
 ```
 
 ## File Format
 
-Waiver files are [input files](inputs.html "inputs files") with a
-specific format:
+Waiver files are [input files](https://www.inspec.io/docs/reference/inputs/) with a specific format:
 
 ```yaml
-control_id:
```
zenspider: no, this line needs to stay. (fixing on my side)

docs/waivers.md
```diff
@@ -38,7 +36,7 @@ control_id:
 + `justification` can be any text you want and might include a reason
   as well as who signed off on the waiver.
 
-### Example:
+### Examples:
```
zenspider, line 39: I went singular because this is one file. (I don't care either way, just explaining)

docs/waivers.md
zenspider:
```
      return unless __waiver_data.key?("skip") && __waiver_data["skip"]
```

seems like it should work either way? Tho that `key?` check doesn't seem necessary. Hashes default to nil for missing keys.

docs/waivers.md
zenspider: In standup the hangup seemed to be around "ultimately" being interpreted as "the only way" when I meant it as a future looking thing. We have many features that were intended for specific uses/users but can be used by many.

I'm changing this to "intended" after talking to Mary.

PR #4545, opened 2019-10-01
Replace input to_hash method, and add unit test for it

### LINKED ISSUE #4504: `to_hash` error when running inherited profile with an input

clintoncwolfe:
## Description

This re-adds the `to_hash` method to the Input class, which allows Inputs to be marshalled for certain types of output, especially when the inputs are in inherited profiles.

Also included is a unit test to help exercise the method.

attn: @trevor-vaughan 

## Related Issue

Fixes #4504 
Supercedes #4502
Probably regressed on #4485

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/input.rb
+++ b/lib/inspec/input.rb
@@ -319,6 +319,21 @@ module Inspec
     end
 
     #--------------------------------------------------------------------------#
+    #                               Marshalling
+    #--------------------------------------------------------------------------#
+
+    def to_hash
+      as_hash = { name: name, options: {} }
+      %i{description title identifier type required value}.each do |field|
+        val = send(field)
+        next if val.nil?
+
+        as_hash[:options][field] = val
+      end
+      as_hash
+    end
+
+    #--------------------------------------------------------------------------#
     #                           Value Type Coercion
     #--------------------------------------------------------------------------#
 
--- a/lib/inspec/input.rb
+++ b/lib/inspec/input.rb
@@ -318,6 +318,17 @@ module Inspec
       !current_value.is_a? NO_VALUE_SET
     end
 
+    def to_hash
+      as_hash = { name: name, options: {} }
+      %i{description title identifier type required value}.each do |field|
+        val = send(field)
+        next if val.nil?
+
+        as_hash[:options][field] = val
+      end
+      as_hash
+    end
+
     #--------------------------------------------------------------------------#
     #                           Value Type Coercion
     #--------------------------------------------------------------------------#
--- a/lib/inspec/input.rb
+++ b/lib/inspec/input.rb
@@ -319,21 +319,6 @@ module Inspec
     end
 
     #--------------------------------------------------------------------------#
-    #                               Marshalling
-    #--------------------------------------------------------------------------#
-
-    def to_hash
-      as_hash = { name: name, options: {} }
-      %i{description title identifier type required value}.each do |field|
-        val = send(field)
-        next if val.nil?
-
-        as_hash[:options][field] = val
-      end
-      as_hash
-    end
-
-    #--------------------------------------------------------------------------#
     #                           Value Type Coercion
     #--------------------------------------------------------------------------#
 
--- a/test/unit/inputs/input_test.rb
+++ b/test/unit/inputs/input_test.rb
@@ -25,6 +25,28 @@ describe Inspec::Input do
     end
   end
 
+  describe "marshalling" do
+    it "should be able to represent an Input as a Hash" do
+      input = Inspec::Input.new("test_input",
+        value: 12,
+        title: "Best input ever",
+        description: "important",
+        type: "Numeric",
+        required: true
+      )
+      _(input.to_hash).must_equal({
+        name: "test_input",
+        options: {
+        value: 12,
+        title: "Best input ever",
+        description: "important",
+        type: "Numeric",
+        required: true
+        }
+      })
+    end
+  end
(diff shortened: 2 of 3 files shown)


lib/inspec/input.rb
```diff
@@ -318,6 +318,21 @@ def has_value?
       !current_value.is_a? NO_VALUE_SET
     end
 
+    #--------------------------------------------------------------------------#
+    #                               Marshalling
+    #--------------------------------------------------------------------------#
```
miah: I'm all for code comments, but these seem a bit off. Could we make these compatible with rdoc? https://ruby.github.io/rdoc/

lib/inspec/input.rb
```diff
@@ -318,6 +318,21 @@ def has_value?
...
+    def to_hash
+      as_hash = { name: name, options: {} }
+      %i{description title identifier type required value}.each do |field|
+        val = send(field)
+        next if val.nil?
+
+        as_hash[:options][field] = val
+      end
+      as_hash
+    end
```
zenspider, line 330: This code is duplicated from https://github.com/inspec/inspec/blob/master/lib/inspec/objects/input.rb#L17-L26 . It should only live in one spot, esp since those two files open and define in the same class (which is another separate problem).

Why it isn't showing up in codeclimate is probably a threshold thing and this method is juuuust small enough.

test/unit/inputs/input_test.rb
```diff
@@ -25,6 +25,28 @@
...
+      )
+      _(input.to_hash).must_equal({
+        name: "test_input",
+        options: {
+        value: 12,
+        title: "Best input ever",
+        description: "important",
+        type: "Numeric",
+        required: true
+        }
```
zenspider: Indentation is incorrect for 39-45.

Also fails chefstyle.

zenspider (changes requested):

lib/inspec/input.rb
zenspider: Whole heartedly agree. This becomes an rdoc/ri comment on the `#to_hash` method.

I also don't think these improve navigability of code/files.

I'd rather we move towards having our public API have proper rdoc/ri comments.

lib/inspec/input.rb
clintoncwolfe: Well, I've modified it to `require "inspec/objects/input"`. Is that a lesser evil?

test/unit/inputs/input_test.rb
```diff
@@ -25,6 +25,27 @@
...
+    it "should be able to represent an Input as a Hash" do
+      input = Inspec::Input.new("test_input",
+        value: 12,
+        title: "Best input ever",
+        description: "important",
+        type: "Numeric",
+        required: true)
+      _(input.to_hash).must_equal({
+        name: "test_input",
+        options: {
```
miah, line 38: Alignment is a bit off here.

lib/inspec/input.rb
zenspider: It is A lesser evil... I'm not sure it is the right one.

My intent is to move `lib/inspec/objects/*` to inspec-scap. I know we use `to_hash` in inspec, but I don't see use in scap (there are 3 calls to to_hash in `or_test.rb` and `control.rb`, but I don't think they're really necessary).

I don't think the other to_hash methods are actually needed, but I'm double checking that now.

zenspider (changes requested):
Failing CI in a bizarre way?!?! That code is totally unrelated, is it not?

Also not sure if the move to use a require is the right one. See my comments above.

test/unit/inputs/input_test.rb
clintoncwolfe: I'm not sure what you're referring to - this was formatted by `chefstyle -a`, and passes `chefstyle` now. Could you maybe paste an example of what you mean?

miah: `options` starts a new hash, so the alignment needs to change.

```
_(input.to_hash).must_equal({
        name: "test_input",
        options: {
          value: 12,
          title: "Best input ever",
          description: "important",
          type: "Numeric",
          required: true,
          },
        })
```

lib/inspec/input.rb
clintoncwolfe: In discussion Friday October 4th, @zenspider felt that the least evil at this point would be to revert the code to have a duplicate `to_hash` implementation in both locations, and remove the `require "inspec/objects/input"`. I'll push that shortly.

zenspider (approved):
We should still do the namespacing... but that'll require synchronizing with @james-stocks.

miah (approved):

Issue #3749, opened 2019-01-24
Result from `ulimit -n` is empty


jerryaldrichiii:
### 🎛 Description

**Choose one:** is this a 🐛 bug report

The value of `ulimit -n` is empty but not when redirecting output. See:

```
[~]$ ulimit -n                                                                               
1024
[~]$ inspec shell                                                                               
Welcome to the interactive InSpec Shell
To find out how to use it, type: help

You are currently running on:

    Name:      arch
    Families:  linux, unix, os
    Release:   4.18.16-arch1-1-ARCH
    Arch:      x86_64

inspec> command('ulimit -n').result
=> #<struct Train::Extras::CommandResult stdout="", stderr="", exit_status=1>
inspec> command('ulimit -n | cat').result
=> #<struct Train::Extras::CommandResult stdout="1024\n", stderr="", exit_status=0>
inspec> command('ulimit -n 2>&1').result
=> #<struct Train::Extras::CommandResult stdout="1024\n", stderr="", exit_status=0>
```

Other shell built-in commands seem to work as expected. See:

```
inspec> command('printf example').result
=> #<struct Train::Extras::CommandResult stdout="example", stderr="", exit_status=0>
inspec> command('pwd').result
=> #<struct Train::Extras::CommandResult stdout="/home/jerry\n", stderr="", exit_status=0>
```

Also, it works when calling bash specifically see:

```
inspec> command('bash -c "ulimit -n"').result
=> #<struct Train::Extras::CommandResult stdout="1024\n", stderr="", exit_status=0>
```

chrisg-fastlane:
I've found these BASH builtins exhibit similar behaviour: _declare, dirs, enable, help, shopt, typeset, ulimit_.  I get the same behaviour in centos and mac_os_x (if the builtins exist).

    Name:      centos
    Families:  redhat, linux, unix, os
    Release:   7.4.1708
    Arch:      x86_64

    Name:      mac_os_x
    Families:  darwin, bsd, unix, os
    Release:   17.7.0
    Arch:      x86_64
```
control 'test builtins' do
  %w[declare dirs enable help shopt typeset ulimit].each do |cmd|
    describe command(cmd) do
      its(:stdout) { should_not cmp '' }
    end
    describe command("#{cmd} | cat") do
      its(:stdout) { should_not cmp '' }
    end
  end
end
```

clintoncwolfe:
[...]

PR #4548, opened 2019-10-01
Fix for invocation when calling npm under sudo

### LINKED ISSUE #4290: NPM Resource does not report on local package installations properly
The NPM resource allows a path to be set to find the node_modules in a project directory. The command that it uses works on the command line.

cd <PROJECT_DIR> && npm ls --json <PACKAGE_NAME>

```json
{
  "dependencies": {
    "sprintf-js": {
      "version": "1.1.2",
      "from": "sprintf-js@latest",
      "resolved": "https://registry.npmjs.org/sprintf-js/-/sprintf-js-1.1.2.tgz"
    }
  }
}
```

But when run in InSpec the test states that the package is not installed:

```ruby
describe npm('azure-storage', path: '/usr/local/camsa/statsd/azure-queue') do
   it { should be_installed }
end
```

```
  Npm Package azure-storage
     [FAIL]  should be installed
     expected that `Npm Package azure-storage` is installed
```

clintoncwolfe:

## Description

When calling `npm`, which happens when using the `npm` resource, it was possible for the invocation to fail due to `"sudo: cd: command not found\n"` when a path was provided to the `npm` resource.

The fix wraps the call in a `sh -c` wrapper.

Additionally, on windows, the resource never worked properly when a custom path was provided, since it was using `&&` to cd then check for the module. This PR now uses `;` to separate the cd command on Windows, which PowerShell seems to tolerate.

I've manually tested this on Windows; I have a very brittle kitchen / Vagrant setup that does everything but install the actual NPM modules, which had to be done manually.

Screen Shot 2019-11-07 at 17 29 54

## Related Issue

Fixes #4290 

## Types of changes

## Checklist:


zenspider (approved):

lib/inspec/resources/npm.rb
```diff
@@ -32,7 +32,7 @@ def info
         npm = "npm -g"
       end
 
-      cmd = inspec.command("#{npm} ls --json #{@package_name}")
+      cmd = inspec.command("sh -c '#{npm} ls --json #{@package_name}'")
```
miah: We should probably remove 'windows' as a supported platform.

miah (approved):

lib/inspec/resources/npm.rb
russellseymour: @miah I do not think that removing support for Windows is the way to go here.

Is it not be possible to detect the platform that Inspec is running on and choose if `sh -c` should be used for Linux or `cmd /c` for Windows?

clintoncwolfe: Thank you, on further reflection what I want to express is "there has never been correct support for windows if you use a custom path", but by removing it here we'd remove the npm resource for Windows all users, custom path or no. 

I agree, a custom Windows-specific invocation is needed.

miah:
related: https://github.com/inspec/inspec/issues/3749

lib/inspec/resources/npm.rb
clintoncwolfe: This is now in place and ready for review.

miah (approved):

lib/inspec/resources/npm.rb
```diff
@@ -27,12 +27,20 @@ def info
...
       else
-        npm = "npm -g"
+        invocation = "npm -g"
       end
 
-      cmd = inspec.command("#{npm} ls --json #{@package_name}")
+      invocation = "#{invocation} ls --json #{@package_name}"
+
+      # If on unix, wrap in sh -c to protect against sudo
+      unless inspec.os.platform?("windows")
```
zenspider, line 39: I generally eschew `unless` if there is a positive way to describe what I want:

```
# If on unix, wrap in sh -c to protect against sudo
unless inspec.os.platform?("windows")
```

vs

```
# Wrap in sh -c to protect against sudo
if inspec.os.platform?("unix")
```

Seems clearer to me.

lib/inspec/resources/npm.rb
```diff
@@ -27,12 +27,20 @@ def info
...
+        invocation = "sh -c '#{invocation}'"
+      end
+
+      cmd = inspec.command(invocation)
```
zenspider, line 43: Is this something that `command` (or train?) should be doing across the board? If on unix and sudo then wrap w/ `sh`?

PR #4550, opened 2019-10-02
Add no-op resource and switch Rule to use it for skipped resources.

### LINKED ISSUE #4541: InSpec 4.17.x breaks only_if on non-os platforms
# Describe the problem
I'm a contributor to the inspec-gcp resource pack. We frequently use only_if in integration tests to skip tests based on environment variables (if a user needs extra permissions to be able to run certain tests)

Since the 4.17.7 release my integration test pipeline has been broken with the following error message:
`Unsupported resource/backend combination: os / gcp. Exiting.`
This was introduced in this PR: https://github.com/inspec/inspec/pull/4214/files#diff-97cbc2a879b4ce388a6c777552c2b542

This seems to be caused by the resource defaulting to an os resource if a control is to be skipped in the following file:
https://github.com/inspec/inspec/blob/592f091d9571140e0eb272c42b8ddaca353986c3/lib/inspec/rule.rb#L232

I imagine this was silently failing the `check_supports` call and working as intended in previous versions until the error started to be thrown due to PR #4214 

I believe this would cause similar issues on other resource packs that do not support the OS platform and use the skip functionality.

## Possible Solution

I would suggest defaulting to a resource that does not need explicit platform support

zenspider:
Avoids the situation of using `os` for non-os transports and blowing up.

Fixes #4541.

### DIFF
--- a/lib/inspec/resources/noop.rb
+++ b/lib/inspec/resources/noop.rb
@@ -1,0 +1,9 @@
+module Inspec::Resources
+  class Noop < Inspec.resource(1)
+    name "noop"
+
+    def to_s
+      "No-op"
+    end
+  end
+end
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -227,9 +227,7 @@ module Inspec
         msg = "Skipped control due to #{skip_check[:type]} condition."
       end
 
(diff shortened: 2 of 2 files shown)


miah (approved):

clintoncwolfe (approved):
Thanks!

PR #4556, opened 2019-10-02
Don't use #at when you can use #[] for array (or hash) access.

zenspider:


miah (approved):

clintoncwolfe (approved):

stevendanna:
For this sort of thing it would be really helpful if the commit message explained a bit about why this change is what we want.

zenspider:
Reworded the commit message. Doesn't need retesting.

PR #4558, opened 2019-10-03
Add name property to interface resource

### LINKED ISSUE #3920: Interface resource doesn't implement the name property
### 🎛 Description

**Choose one:** 🐛 bug report

According to the [documentation](https://www.inspec.io/docs/reference/resources/interface/), there are two parameters for the interface resource, speed and name. Well, it turns out that there's no method to expose the name property, even though it is available in `@interface_provider` from both of the platform providers.

3.9.0-1.el7

### 🤔 Replication Case

This example from the documentation doesn't work, it'll result in an undefined method exception:

```ruby
describe interface('eth0') do
  it { should be_up }
  its('speed') { should eq 1000 }
  its('name') { should eq eth0 }
end
```

### 💁 Possible Solutions

Implement a name method for interface. There's probably a sketchy way for me to do this without modifying the source code of the resource... but I don't know how.

### 💻 Stacktrace

```
     ×  Interface eth0 name
     undefined method `name' for Interface eth0:#<Class:0x0000000003934a10>
```

miah:
Also some general cleanup.

Fixes #3920

### DIFF
--- a/lib/inspec/resources/interface.rb
+++ b/lib/inspec/resources/interface.rb
@@ -17,38 +17,34 @@ module Inspec::Resources
         its('ipv6_cidrs') { should include '::1/128' }
       end
     EXAMPLE
+
     def initialize(iface)
       @iface = iface
-
-      @interface_provider = nil
-      if inspec.os.linux?
-        @interface_provider = LinuxInterface.new(inspec)
-      elsif inspec.os.windows?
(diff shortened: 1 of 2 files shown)


zenspider (approved):
This looks good. In the future, please use multiple commits for multiple tasks/ideas. This way the "general cleanup" can be reviewed separately from the name property changes.

clintoncwolfe (approved):

PR #4517, opened 2019-09-25
Fix yum repo matcher for rhel-8 / centos-8

sspans:
Fix parsing for the final repo in yum repolist

## Description
The update yum / dnf tooling in rhel-8 / centos-8 omits the final empty line in "yum repolist" output.
This results in the final repository not being parsed.

CentOS 6 output:
```
Repo-id      : updates
Repo-name    : CentOS-6 - Updates
Repo-status  : enabled
Repo-revision: 1565879235
Repo-updated : Sun Aug 25 03:27:52 2019
Repo-pkgs    : 595
Repo-size    : 7.8 G
Repo-baseurl : https://foobar.tld/repo/
Repo-expire  : 21,600 second(s) (last: Wed Sep 25 11:18:39 2019)

repolist: 19,972
```

CentOS 8 output:
```
Repo-id      : sensu
Repo-name    : sensu monitoring
Repo-status  : enabled
Repo-revision: 1562883501
Repo-updated : Fri Jul 12 00:18:40 2019
Repo-pkgs    : 1
Repo-size    : 24 M
Repo-baseurl : http://repositories.sensuapp.org/yum/8/x86_64/
Repo-expire  : 21,600 second(s) (last: Wed Sep 25 11:18:39 2019)
Repo-excluded: 127

Repo-id      : updates
Repo-name    : CentOS-8 - Updates
Repo-status  : enabled
Repo-revision: 1565879235
Repo-updated : Sun Aug 25 03:27:52 2019
Repo-pkgs    : 595
Repo-size    : 7.8 G
Repo-baseurl : Repo-baseurl : https://foobar.tld/repo/
Repo-expire  : 21,600 second(s) (last: Wed Sep 25 11:18:39 2019)
```

In this case the 'updates' repo would not be parsed.

The easiest fix I could come up with was adding an additional echo.
I'm sure a cleaner fix is possible, but this was quick and easy.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.

PR #4566, opened 2019-10-05
Refactored and added an alternative test mock resource system.

zenspider:
test/unit/resources/json_test.rb had the start of a setup to directly
instantiate resources. I've refactored that up to helper.rb and
extended it with some helper methods so we can directly specify what a
resource command should respond with. For many/most of our tests, this
should be sufficient.

Finally, I switched our yum tests over to use the new setup. This will
allow us to address #4517 and extend the tests for centos8 output
differences with ease. Ease, folks... ease.

Cc: Miah Johnson <miah@chia-pet.org>


miah (approved):

miah:
I think we're also going to need a way to specify the platform.

This is a really nice start. I think we'll be able to simplify our unit tests moving forward, it will be easier for users to contribute and understand this compared to the MockLoader!

miah:
Also need to look at handling this: https://buildkite.com/chef-oss/inspec-inspec-master-verify/builds/677#fc29326a-8bfa-4064-8e27-f5c06a2c6e6f/648-697

clintoncwolfe (approved):
Nicely done - thanks!

zenspider:
This can't go in until it runs on windows properly... which does require setting a platform. Hrmph.

PR #4568, opened 2019-10-08
Improve testing for yum.repo in CentOS 8

ramereth:
The output for ``yum -v repolist all`` changes slightly in CentOS 8 and breaks
the current regex for parsing the output. There are two specific changes that
are breaking this.

1. EL8 is no longer trailing the repo name with release version and arch (i.e.
   base instead of base/7/x86)
2. EL8 no longer adds a trailing newline on the last repo and instead has a line
   with ``Total packages:``. This means the repo listed last will never show up.

This change updates the regex to account for both pre-CentOS 8 and CentOS 8 and
beyond.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/yum.rb
+++ b/lib/inspec/resources/yum.rb
@@ -59,7 +59,7 @@ module Inspec::Resources
         # detect repo start
         in_repo = true if line =~ /^\s*Repo-id\s*:\s*(.*)\b/
         # detect repo end
-        if line == "\n" && in_repo
+        if (line == "\n" || line =~ /\s*Total packages:/) && in_repo
           in_repo = false
           @cache.push(repo)
           repo = {}
--- a/test/helpers/mock_loader.rb
+++ b/test/helpers/mock_loader.rb
@@ -6,7 +6,7 @@ class MockLoader
     centos5: { name: "centos", family: "redhat", release: "5.11", arch: "x86_64" },
     centos6: { name: "centos", family: "redhat", release: "6.6", arch: "x86_64" },
     centos7: { name: "centos", family: "redhat", release: "7.1.1503", arch: "x86_64" },
-    centos8: { name: "centos", family: "redhat", release: "8.9.10", arch: "x86_64" },
+    centos8: { name: "centos", family: "redhat", release: "8.0.1905", arch: "x86_64" },
     cloudlinux: { name: "cloudlinux", family: "redhat", release: "7.4", arch: "x86_64" },
     coreos: { name: "coreos", family: "coreos", release: "1437.0.0", arch: "x86_64" },
     debian6: { name: "debian", family: "debian", release: "6", arch: "x86_64" },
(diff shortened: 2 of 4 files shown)


miah:
Can you verify if this is still busted on master? We may have just fixed this one.

https://github.com/inspec/inspec/pull/4566

ramereth:
@miah this is still busted on master and #4566 doesn't seem to address the two changes I mentioned above in the regex. Would you like me to rebase it once #4566 is merged?

miah:
Thanks for verifying. Lets get #4566 (probably tomorrow) in first and then this. =)

zenspider:
I believe the yum issue is fixed in #4566. Can you please verify?

ramereth:
@zenspider it does seem to be working however the tests still need some fixes it seems as it's only testing real ``yum repolist`` output from centos7 and not centos8. If you'd like. I can clean up this PR to remove my regex fix and just clean up the tests which I think would be a great idea to ensure its working properly.

zenspider:
@ramereth Can you look at https://github.com/inspec/inspec/blob/master/test/unit/mock/cmd/yum-centos8-repolist-all ?

AFAICT, the only real difference is that last line that you have a regexp for. The problem is that your PR literally _can't_ test for both yum on centos 7 and 8. If you could do a PR against the file above I'd appreciate it. I don't have access to a centos boxen atm.

ramereth:
@zenspider how does my PR not test both 7 and 8? I confirmed that it does. Can you please point out what needs to be fixed so that it does? Also the yum repolist output between CentOS 7 and 8 are vastly different due to the fact that 8 now uses DNF and also includes a different set of repositories that 7 doesn't include. What I have included in my PR shows what it really looks like.

Thanks!

miah:
Hi @ramereth I think this is finally fixed in master. I've validated it on my end, but could you double check? Thanks

ramereth:
@miah I think it's fixed however the tests need to be fixed IMHO. Please see what I stated above.

miah:
I agree, we should bring in the tests. Do you want to rebase and remove the change to the yum resource?

ramereth:
@miah yes I'll do that!

lib/inspec/resources/yum.rb
```diff
@@ -59,7 +59,7 @@ def repositories
         # detect repo start
         in_repo = true if line =~ /^\s*Repo-id\s*:\s*(.*)\b/
         # detect repo end
-        if line == "\n" && in_repo
+        if (line == "\n" || line =~ /\s*Total packages:/) && in_repo
```
ramereth, line 62: I discovered this was still needed because of how the new output from CentOS 8 changes. If you revert this line you will see how it fails. Good thing I improved the tests to catch this :smiley:

miah: Can you show me how its failing? I was testing master against my Centos 8 server last night but didn't see anything off. (i was admittedly quite tired though)

ramereth: It's likely not failing but adding extra data on the last repository that shouldn't be there. If you revert this line and run the test you'll see that it appends ``"Total packages" => "7,644"`` to the ``extras`` repository which shouldn't be there.

miah (approved):

miah:
Thanks @ramereth =)

PR #4581, opened 2019-10-10
Telemetry: wire up the options and env switches

clintoncwolfe:

## Description

This (Draft) PR makes the CLI option --enable-telemetry and ENV variables work with `chef-core/telemeter` to setup the Telemeter. It also kicks off the old session upload if enabled, so we're off to the races.

The logic behind whether to enable the telemetry system is a bit fraught. It bears review.

All enable logic is currently hidden behind a feature flag env var.

We'll need to add a call to the configuration method, `base_cli/configure_telemeter()` in each subcommand. This PR just does `shell`.

There are no tests for this code at the moment.

## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -6,16 +6,16 @@ expeditor:
       retry:
         automatic:
           limit: 1
-      
+
 steps:
 
-- label: run-tests-ruby-2.4
-  command:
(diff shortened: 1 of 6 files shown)


zenspider (approved):

clintoncwolfe:
Closing - we know we are going to do this in the chef-telemetry library, not here in the inspec library itself.

PR #4600, opened 2019-10-17
Add ruby 2.7 preview

### LINKED ISSUE #4588: Add Ruby 2.7-preview to CI
We need to start testing against Ruby 2.7 as it will be released in December and contains a huge change that will likely affect InSpec.

https://bugs.ruby-lang.org/issues/14183

miah:
closes: #4588

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -35,7 +35,7 @@ steps:
 
 - label:  run-tests-ruby-2.7-preview
   command:
-    - /workdir/.expeditor/buildkite/verify.sh
+    - /workdir/.expeditor/buildkite/verify.sh || true
   expeditor:
     executor:
       docker:
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -6,7 +6,7 @@ expeditor:
       retry:
         automatic:
           limit: 1
-      
+
 steps:
 
 - label: run-tests-ruby-2.4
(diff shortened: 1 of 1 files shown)


clintoncwolfe (commented):
The BK change looks fine, and certainly looks like it is doing the right thing according to its output. How do you want to handle the test fixes - on this PR?

zenspider (approved):
THANK YOU

zenspider:
Hrm... Can we make this a soft failure for now? We obviously need to get these fixed asap. A lot of them are probably easy fixes but I also suspect there's going to be some dependency hell to fix.

zenspider (approved):
Hrm... Can we make this a soft failure for now?

We obviously need to get these fixed asap (but not on this PR). A lot of them are probably easy fixes but I also suspect there's going to be some dependency hell to fix.

miah:
I'll tweak this so 2.7 is executed similar to our coverage test and allow failures.

.expeditor/verify.pipeline.yml
```diff
@@ -33,6 +33,14 @@ steps:
       docker:
         image: ruby:2.6-stretch
 
+- label:  run-tests-ruby-2.7-preview
+  command:
+    - /workdir/.expeditor/buildkite/verify.sh
```
zenspider: Probably should be `/workdir/.expeditor/buildkite/verify.sh || true` for now.

zenspider (changes requested):

.expeditor/verify.pipeline.yml
miah: Fixed =)

miah:
The 2.7 test will always pass for now.

We will remove the `|| true` when we get down to zero failures.

File issues and fixes to address the issues in the 2.7 bk step.

zenspider (approved):
Yay!

PR #4601, opened 2019-10-17
Re-organize test-kitchen files. Add BK Pipeline

miah:
The layout of the test kitchen files on the filesystem was somewhat confusing. This puts them all under `test/kitchen`.

This also adds a buildkite pipeline to start testing the core resources.


miah:
Dropped the WIP as all iteration on the pipeline will come after this is merged. This allows expeditor to set up the pipeline in the background and on the follow-up work will pick up the changes to `verify.resource.yml` and do stuff.

clintoncwolfe (commented):
I'm fine with the move

clintoncwolfe (approved):
Thanks!

zenspider (approved):

PR #4626, opened 2019-10-21
Create a user 'normal' and use it to run tests. Also bundle caching.

### LINKED ISSUE #4353: BuildKite runs as root
We definitely should NOT be running as root on our CI.

This issue should probably roll up for all of our BK infrastructure.

zenspider:
Fixes #4353.


miah (approved):
seems legit

clintoncwolfe (approved):
Great - thanks. Looking forward to more on the non-root user runs.

PR #4628, opened 2019-10-21
Fixed the rest of the minitest deprecations by removing FuncTestRunResult.

### LINKED ISSUE #4533: newest version of minitest has new deprecations that we abuse
```
DEPRECATED: global use of must_respond_to from /workdir/test/unit/utils/deprecation_test.rb:113. Use _(obj).must_respond_to instead. This will fail in Minitest 6.
```

we can pin back for a while to reduce noise, but we really should address this... preferably by not using minitest/spec at all, since it is a source of pain in general.

zenspider:
Folded the json & other payloads into plain ivars on the test.
Found and removed some unused code while I was there.

Fixes #4533.


miah (approved):
:fish:

clintoncwolfe (approved):

Issue #4570, opened 2019-10-08
Windows Resource Parity


zenspider:
This is an epic for ongoing work on our resources.

Many resources provide better support on non-Windows hosts; this Epic tracks work to level the playing field.

mhackethal:
May i highlight #4388 - for windows prarity? ;)

PR #4639, opened 2019-10-25
Add support for Windows to resource sys_info

mhackethal:
## Description

The resource sys_info was designed to support Linux and Windows OS.
The latest changes for (hostname) didn´t include the Windows support.
On Windows (OS) it Always returns that hostname is not supported for the OS yet.

In regards to #4570 i tried to add some of the Features for Windows.
This pull - is just a first guess. Maybe its better to re-design the whole resource to support
some kind of "initialize" method and have all data directly available?

Since most of the Windows "sys_info" are stored in the Win32_ComputerSystem i decided to create a "Cache" that holds all Information, instead of running a single query for every Attribute.

In case the functionality is welcome please provide me some hints how it should look like to fit into inspec. I will take care of These changes.

If its okay for you - the documentation and test updates will follow as soon as we have an allignment ?

## Related Issue

#4570 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/sys_info.rb
+++ b/lib/inspec/resources/sys_info.rb
@@ -9,7 +9,7 @@ module Inspec::Resources
     supports platform: "windows"
 
     desc "Use the sys_info resource to test for operating system properties."
-   example <<~EXAMPLE
+    example <<~EXAMPLE
       describe sys_info do
         its('hostname') { should eq 'hostname' }
       end
@@ -53,7 +53,7 @@ module Inspec::Resources
       @cache.info[:totalmemory_kb]
     end
 
-    def hostname 
+    def hostname
       @cache.info[:hostname]
     end
 
@@ -65,7 +65,7 @@ module Inspec::Resources
       "System Information"
     end
   end
-  
+
   class LinuxInfo < SysInfo
     attr_reader :inspec
     def initialize(inspec)
@@ -78,7 +78,7 @@ module Inspec::Resources
       ip = inspec.command("hostname -I").stdout.chomp
       short = inspec.command("hostname -s").stdout.chomp
       manufacturer = inspec.command("cat /sys/class/dmi/id/sys_vendor").stdout.chomp
-      model =  inspec.command("cat /sys/class/dmi/id/product_name").stdout.chomp
+      model = inspec.command("cat /sys/class/dmi/id/product_name").stdout.chomp
       totalmemory_kb = inspec.command("awk '/MemTotal/ {print $2}' /proc/meminfo").stdout.chomp
 
       {
--- a/lib/inspec/resources/sys_info.rb
+++ b/lib/inspec/resources/sys_info.rb
@@ -3,187 +3,140 @@ require "inspec/resources/powershell"
 
 module Inspec::Resources
   # this resource returns additional system informatio
-  class System < Inspec.resource(1)
+  class SysInfo < Inspec.resource(1)
     name "sys_info"
-    supports platform: "unix"
+    supports platform: "linux"
     supports platform: "windows"
 
-    desc "Use the user InSpec system resource to test for operating system properties."
-    example <<~EXAMPLE
+    desc "Use the sys_info resource to test for operating system properties."
+   example <<~EXAMPLE
       describe sys_info do
-        its('hostname') { should eq 'example.com' }
+        its('hostname') { should eq 'hostname' }
       end
 
       describe sys_info do
-        its('fqdn') { should eq 'user.example.com' }
+        its('fqdn') { should eq 'hostname.example.com' }
       end
 
     EXAMPLE
 
-    %w{ domain fqdn ip_address short }.each do |opt|
-      define_method(opt.to_sym) do
-        hostname(opt)
-      end
-    end
-    @cache = nil
-    
-    def win_cache
-      return {} if !inspec.os.windows?
-      cmd = inspec.command <<-EOF
-        Get-WmiObject Win32_ComputerSystem -Property * | ConvertTo-Json
-      EOF
-
-      return {} if cmd.stdout == ""
-      begin
-        sysinfo = JSON.parse(cmd.stdout)
-      rescue JSON::ParserError => each
-        raise Inspec::Exceptions::ResourceFailed,
-          "Failed to parse JSON from Powershell. " \
-          "Error: #{e}"
-      end
-      return sysinfo
-    end
-
-    # returns the hostname of the local system
-    def hostname(opt = nil)
+    def initialize
       os = inspec.os
+      @cache = nil
+
       if os.linux?
-        linux_hostname(opt)
-      elsif os.darwin?
-        mac_hostname(opt)
+        @cache = LinuxInfo.new(inspec)
       elsif os.windows?
-        opt = "s" if opt.nil?
-        windows_hostname(opt)
+        @cache = WindowsInfo.new(inspec)
       else
-        skip_resource "The `sys_info.hostname` resource is not supported on your OS yet."
-      end
-    end
-    
-    def windows_hostname(opt = nil)
-      if @cache.nil?
-        @cache = win_cache
-      end
-      if !opt.nil?
-        opt = case opt
-          when "f", "long", "fqdn", "full"
-            #https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-computersystem
-            return @cache["DNSHostName"] if (@cache["DomainRole"] == 2 || @cache["DomainRole"] == 0)
-            return "#{@cache["DNSHostName"]}.#{@cache["Domain"]}"
-          when "d", "domain"
-            return nil if (@cache["DomainRole"] == 2 || @cache["DomainRole"] == 0)
-            return @cache["Domain"]
-          when "i", "ip_address"
-            cmd = inspec.command("(Get-WmiObject Win32_NetWorkAdapterConfiguration | Where IPAddress -ne $null | select -Property ipaddress).IPAddress | ConvertTo-Json")
-            return {} if cmd.stdout == ""
-            begin
-              ip = JSON.parse(cmd.stdout)
-            rescue JSON::ParserError => each
-              raise Inspec::Exceptions::ResourceFailed,
-                "Failed to parse JSON from Powershell. " \
-                "Error: #{e}"
-            end
-            return ip
-          when "s", "short"
-            return @cache["DNSHostName"]
-          else "ERROR"
-        end
-      end
-      if opt == "ERROR"
-        skip_resource "The `sys_info.hostname` resource is not supported with that option on your OS."
+        return skip_resource "The `sys_info` resource is not supported on your OS yet."
       end
     end
 
-    def linux_hostname(opt = nil)
-      if !opt.nil?
-        opt = case opt
-              when "f", "long", "fqdn", "full"
-                " -f"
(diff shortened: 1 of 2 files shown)


lib/inspec/resources/sys_info.rb
```diff
@@ -106,11 +158,29 @@ def model
       elsif os.linux?
         inspec.command("cat /sys/class/dmi/id/product_name").stdout.chomp
       elsif os.windows?
-        inspec.powershell("Get-CimInstance -ClassName Win32_ComputerSystem | Select Model -ExpandProperty Model").stdout.chomp
+        if @cache.nil?
+          @cache = win_cache
```
miah: Instead of this conditional you can do:

`@cache ||= win_cache`

I'd look for similar chunks of code in this PR and see if you can apply similar logic. I'd also have one thing that sets @cache and just have everything else that needs the value for @cache to call it,

eg;

```
def cache
  @cache ||= win_cache
end
```

lib/inspec/resources/sys_info.rb
```diff
@@ -34,15 +52,46 @@ def hostname(opt = nil)
...
       else
         skip_resource "The `sys_info.hostname` resource is not supported on your OS yet."
       end
     end
+    
+    def windows_hostname(opt = nil)
+      if @cache.nil?
+        @cache = win_cache
+      end
+      if !opt.nil?
```
miah: `unless opt.nil?` is nicer than the negation. But since we're just checking that opt is set, `if opt` should suffice and the `else` can cover the case where its nil.

miah (changes requested):

mhackethal:
I will take care of your Suggestions and Implement it next week.
Thanks for the Good Feedback.

Am 28.10.2019 um 18:19 schrieb Miah Johnson <notifications@github.com>:

﻿

@miah requested changes on this pull request.

________________________________

In lib/inspec/resources/sys_info.rb<https://github.com/inspec/inspec/pull/4639#discussion_r339215393>:

> @@ -106,11 +158,29 @@ def model
       elsif os.linux?
         inspec.command("cat /sys/class/dmi/id/product_name").stdout.chomp
       elsif os.windows?
-        inspec.powershell("Get-CimInstance -ClassName Win32_ComputerSystem | Select Model -ExpandProperty Model").stdout.chomp
+        if @cache.nil?
+          @cache = win_cache


Instead of this conditional you can do:

@cache ||= win_cache

I'd look for similar chunks of code in this PR and see if you can apply similar logic. I'd also have one thing that sets @cache and just have everything else that needs the value for @cache to call it,

eg;

def cache
  @cache ||= win_cache
end


________________________________

In lib/inspec/resources/sys_info.rb<https://github.com/inspec/inspec/pull/4639#discussion_r339688366>:

>        else
         skip_resource "The `sys_info.hostname` resource is not supported on your OS yet."
       end
     end
+
+    def windows_hostname(opt = nil)
+      if @cache.nil?
+        @cache = win_cache
+      end
+      if !opt.nil?


unless opt.nil? is nicer than the negation. But since we're just checking that opt is set, if opt should suffice and the else can cover the case where its nil.

lib/inspec/resources/sys_info.rb
```diff
@@ -34,15 +52,46 @@ def hostname(opt = nil)
       elsif os.darwin?
         mac_hostname(opt)
       elsif os.windows?
-        if !opt.nil?
-          skip_resource "The `sys_info.hostname` resource is not supported with that option on your OS."
-        else
-          inspec.powershell("$env:computername").stdout.chomp
-        end
+        opt = "s" if opt.nil?
```
miah: `opt = "s" unless opt`

mhackethal:
sorry was quite busy last weeks and missed to update this :)
i wanted to try something else and re-created the resource. Maybe this fits a bit more into the current resource design?

lib/inspec/resources/sys_info.rb
```diff
@@ -3,117 +3,140 @@
...
-    def hostname(opt = nil)
+    def initialize
       os = inspec.os
+      @cache = nil
+
       if os.linux?
-        linux_hostname(opt)
-      elsif os.darwin?
-        mac_hostname(opt)
+        @cache = LinuxInfo.new(inspec)
```
clintoncwolfe: While you have a ivar named `@cache`, it actually isn't being used as a cache - you're repeatedly calling `info()` on it, which is expensive. The return value of info is what you want to cache.

Consider modifying this to:
```
@cache = LinuxInfo.new(inspec).info
```

and then (for example)

```
    def domain
      @cache[:domain]
    end
```

lib/inspec/resources/sys_info.rb
```diff
@@ -3,117 +3,140 @@
...
+    def fqdn
+      @cache.info[:fqdn]
     end
 
     def to_s
       "System Information"
     end
   end
+
+  class LinuxInfo < SysInfo
```
clintoncwolfe, line 82: I see a LinuxInfo and a new WindowsInfo, but the existing functionality supports MacOS as well. We'll need a support class for Mac.

lib/inspec/resources/sys_info.rb
```diff
@@ -3,117 +3,140 @@
...
+
+      if (info["DomainRole"] == 2 || info["DomainRole"] == 0)
+        fqdn = info["DNSHostName"]
+        domain = nil
+      else
+        fqdn = "#{info["DNSHostName"]}.#{info["Domain"]}"
+        domain = info["Domain"]
+      end
+
+      memory = info["TotalPhysicalMemory"] / 1024 rescue nil
```
clintoncwolfe: I'd prefer `info.fetch("TotalPhysicalMemory", 0) / 1024` to resorting to exception handling here.

clintoncwolfe (changes requested):
Thanks for the contribution, and for coming back to this!

Looks like we're missing some mac functionality, and we'll definitely need tests on this. A couple of other minor code changes requested. Thanks!

lib/inspec/resources/sys_info.rb
mhackethal: sorry for my stupid Questions. 
Will info.fetch("TotalPhysicalMemory",0) / 1024 not raise an divion by Zero error?
In case TotalPhysicalMemory can not be found as key it will Reply the value as 0 and then try to divide it through 1024?
Or do i missunderstand this function?

mhackethal:
I dont have Access to a MacOS so i Need some help to make all Methods available.
At least i re-added the equality as before. The only new method is:
- totalmemory_kb

mhackethal:
Sorry i have to get again into the resource. While cross-checking the "sys_info" documentation i saw that there is a Special function.
```
Options can be passed as arguments to hostname as well.
describe file('/path/to/some/file') do
  its('content') { should match sys_info.hostname('full') }
end
```
I was not aware of this and my change will break this. I try to resolve it

mhackethal:
dear all,

i did some Research on this "option" Topic which is stated in the documentation.
At least on all my Tests this Fails and is not executable.

```
You are currently running on:

    Name:      ubuntu
    Release:   16.04
    Arch:      x86_64
inspec> describe file('/root/sys_info') do
inspec>   its('content') { should match sys_info.hostname('s') }
inspec> end

Profile: inspec-shell
Version: (not specified)

  File /root/sys_info
     [FAIL]  content
     undefined local variable or method `sys_info' for #<RSpec::ExampleGroups::FileRootSysInfo::Content:0x0000000009e57078>

```

@clintoncwolfe  can you confirm this? And my question is how to continue right now.
Is it a good idea to trace down the reason and try to repair it? Or should i update the documentation and remove this feature?
Maybe a deprecation warning is also a suiteable solution.

At the moment i dont find any good usecase this behaviour. Another way to solve this could be a tests like this.
```

Control 'my control' do
var = inspec.command('hostname -XX') || sys_info.[fqdn/hostname/domain]
describe file (/foo/bar) do
  its('content') { should match var }
end
end
```

Please give me some advise here.
My prefered solution is to update the doc and Keep the resource to follow the design of the other resources. 

thx

mhackethal:
That should work - at least it should match the previous behaviour and it adds the missing functionality for Windows OS.

mhackethal:
@clintoncwolfe @miah :
Any update ? Feedback?
I am also fine to not implement it and leave the resource broken.

lib/inspec/resources/sys_info.rb
zenspider: 0/1 is always 0. 1/0 is undefined/error.

lib/inspec/resources/sys_info.rb
```diff
@@ -3,117 +3,182 @@
...
+
+      {
+        fqdn: fqdn,
+        domain: domain,
+        ip: ip,
+        hostname: short,
+        manufacturer: manufacturer,
+        model: model,
+        totalmemory_kb: totalmemory_kb,
+      }
```
miah, line 105: Can we construct this structure higher? Doing this in each SysInfo subclass is likely to cause issues when the structure changes in one, but not the other.

lib/inspec/resources/sys_info.rb
```diff
@@ -3,117 +3,182 @@
...
+    end
+
+    def info
+      fqdn = inspec.command("hostname -f").stdout.chomp
+      domain = inspec.command("hostname -d").stdout.chomp
+      ip = inspec.command("hostname -I").stdout.chomp
+      short = inspec.command("hostname -s").stdout.chomp
+      manufacturer = inspec.command("cat /sys/class/dmi/id/sys_vendor").stdout.chomp
+      model = inspec.command("cat /sys/class/dmi/id/product_name").stdout.chomp
+      totalmemory_kb = inspec.command("awk '/MemTotal/ {print $2}' /proc/meminfo").stdout.chomp
```
miah: I would probably create these as methods in the SysInfo subclasses, we'll get lots of bonus usability.

```
linux = LinuxInfo.new(...)
foo.fqdn
foo.ip
```

miah (changes requested):
Thanks for the contribution, just a few changes I'd like to see.

lib/inspec/resources/sys_info.rb
mhackethal: I think i got your point but moving this into the sys_info class would cause massives changes and will result in different layout as all other ressources: like "services","package". 
But i think i got your point. 
What about updating the methods in root class "sys_info" in this way.

```
    def foobar
      @cache[:foobar] || "The `sys_info.foobar` resource is not supported on your OS yet."
    end
```

In this case the method will return "unsupported" in case the information is not available by the OS subclass.
If this approach is followed by the next changes it will ensure that you can have different methods foreach OS. If a method is not available on a specific OS inspec will notify you.

lib/inspec/resources/sys_info.rb
mhackethal: yah my fault. already figured that out ;)

lib/inspec/resources/sys_info.rb
mhackethal: Here , i dont get your point. The resource have the methods within "sys_info" what is the advantage to have a copy of that foreach OS?
The subclasses foreach OS is only available inside sys_info ressource so its not usable directly inside the inspec tests.
Even if i change it in that way i would have to create a complex way of execution and caching for the windows side.
Or i have to split the 2 windows commands into 6 single powershell commands where i think this even dont bring any advantage.

mhackethal:
No response for Long time. I assume Not Required

PR #4648, opened 2019-10-29
Fixed problem with accessing backend in TestDslLazyLoader#method_missing.

### LINKED ISSUE #4524: Dynamic resource loading fails in certain RSpec extension blocks
# Describe the problem

The dynamic resource DSL-loading code introduced in #4365 altered the way that the resource DSL is imported into the control code space. User who use RSpec features like `before`/`after` blocks and `shared_example` blocks, which are not documented or supported Chef InSpec features, may experience errors if they reference dynamically loaded resources within such blocks.

First seen publicly in release 4.17.7 of Chef InSpec.

## Reproduction 

```
control "repro-1" do
  describe 'resource-in-before' do
    before(:all) do
      command('true')
    end
    it "required to do something" do
    end
  end
end
```

output:
```
Profile: tests from foo.rb (tests from foo.rb)
Version: (not specified)
Target:  local://

  ×  repro-1: resource-in-before
     ×  resource-in-before required to do something
     undefined method `command' for #<RSpec::ExampleGroups::ResourceInBefore before(:context) hook>

```

### LINKED ISSUE #4587: inspec 4.18.0 breaks `subject`-style tests
# Describe the problem
After upgrading this code breaks:

```
describe service('redis-server') do
  it 'is listening on port 6379' do
    expect(port(6379)).to be_listening
  end
end
```

```
     ×  is listening on port 6379
     undefined method `port' for #<RSpec::ExampleGroups::ServiceRedisServer_2:0x00007fc75d89b350>
```

This works with 4.17.17

## Possible Solution
?

zenspider:
Fixes #4587.
Fixes #4524.

### DIFF
--- a/test/unit/runner_test.rb
+++ b/test/unit/runner_test.rb
@@ -3,9 +3,40 @@
 require "helper"
 require "inspec/secrets"
 require "inspec/runner"
+require "fetchers/mock"
 
 describe Inspec::Runner do
   let(:runner) { Inspec::Runner.new({ command_runner: :generic }) }
+
+  it "bug #4524" do
+    file = <<-RUBY
+      describe "a thing" do
+        before(:all) { command("true") }
+        it("should pass") {}
+      end
+    RUBY
+    runner.add_target("bug4524.rb" => file)
+    runner.load
+
+    result = RSpec::Core::Runner.new(nil).run_specs(runner.test_collector.tests)
+
+    _(result).must_equal 0
+  end
+
+  it "bug #4587" do
+    file = <<-RUBY
+      describe "a thing" do
+        subject! { command("true") }
+        its("exit_status") { should eq 0 }
+      end
+    RUBY
+    runner.add_target("bug4587.rb" => file)
+    runner.load
+
+    result = RSpec::Core::Runner.new(nil).run_specs(runner.test_collector.tests)
+
+    _(result).must_equal 0
+  end
 
   # =============================================================== #
   #                         Reporter Options
--- a/test/unit/runner_test.rb
+++ b/test/unit/runner_test.rb
@@ -6,7 +6,7 @@ require "inspec/runner"
 require "fetchers/mock"
 
 describe Inspec::Runner do
-  let(:runner) { Inspec::Runner.new({ command_runner: :generic }) }
+  let(:runner) { Inspec::Runner.new({ command_runner: :generic, reporter: [] }) }
 
   it "bug #4524" do
     file = <<-RUBY
--- a/lib/inspec/rspec_extensions.rb
+++ b/lib/inspec/rspec_extensions.rb
@@ -50,8 +50,8 @@ module Inspec
     def method_missing(method_name, *arguments, &block)
       # see if it is a resource first
       begin
-        inspec = inspec if respond_to?(:inspec) # backend not available??
-        resource = Inspec::DSL.method_missing_resource(inspec, method_name, *arguments)
+        backend = inspec if respond_to?(:inspec) # backend not available??
+        resource = Inspec::DSL.method_missing_resource(backend, method_name, *arguments)
         return resource if resource
       rescue LoadError
         # pass through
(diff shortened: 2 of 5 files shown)


clintoncwolfe (changes requested):
Please add tests that exercise both bugs. In these cases, I find it clearest to simply add a functional test with a simple profile that clearly demonstrates the reproduction case.

We can later iterate on improving those tests (perhaps to move them out of the "functionals"), but for now confidence is key.

miah:
I'm +1 for tests, -1 for functionals. They are too slow.

zenspider:
They are too slow. I just wrote the equivalent functional and unit and the functional test is literally 5x slower. That's on a mac w/ SSD. It would be horrifically slower on windows.

test/unit/runner_test.rb
```diff
@@ -3,9 +3,20 @@
...
 require "inspec/secrets"
 require "inspec/runner"
+require "fetchers/mock"
 
 describe Inspec::Runner do
-  let(:runner) { Inspec::Runner.new({ command_runner: :generic }) }
+  let(:runner) { Inspec::Runner.new({ command_runner: :generic, reporter: [] }) }
+
+  it "bug #4524" do
+    file = 'describe "x" do before(:all) { command("y") }; it "z" do end; end'
```
clintoncwolfe: Please, for clarity, expand this into a heredoc, or place in an external file and make it look like more typical control code. I can help you with that if you would like.

A major advantage of having the control code in an external file is that it is easy to run it using `inspec exec` during testing and development.

I want to build trust and confidence with our users by showing that code that looks exactly like theirs works in our tests.

test/unit/runner_test.rb
```diff
@@ -3,9 +3,20 @@
...
+    runner.add_target("bug4524.rb" => file)
+    runner.load
+
+    result = RSpec::Core::Runner.new(nil).run_specs(runner.test_collector.tests)
```
clintoncwolfe, line 21: I suggest that these three lines be combined into a helper method.

test/unit/runner_test.rb
clintoncwolfe: I am not sure this test belongs in this file - perhaps a new file, runner_dsl_test.rb or similar. 

Once established, and helpers created, there are many functional tests that could likely be migrated to such a unit test file, which would be faster.

test/unit/runner_test.rb
```diff
@@ -3,9 +3,20 @@
...
 require "inspec/runner"
+require "fetchers/mock"
 
 describe Inspec::Runner do
-  let(:runner) { Inspec::Runner.new({ command_runner: :generic }) }
+  let(:runner) { Inspec::Runner.new({ command_runner: :generic, reporter: [] }) }
+
+  it "bug #4524" do
+    file = 'describe "x" do before(:all) { command("y") }; it "z" do end; end'
+    runner.add_target("bug4524.rb" => file)
```
miah, line 18: Think you forgot to add this file?

clintoncwolfe (changes requested):
I am glad that you have found a way to make this work using the much faster unit test approach, and your concern for the developers' test speed is addressed.

I'm still looking for test coverage - the test given covers the `before:` issue but does not provide a test case for the `subject` issue that was reported. 

I don't have full confidence in the unit testing mockup system here. I suggest a compromise in which the profile code under test is stored externally in files, able to be tested by the unit testing system but also able to be accessed manually for diagnosis and confirmation.

clintoncwolfe:
As this is a regression, and for the sake of getting this into a release, all I'm really concerned about for this is adding a test case for `subject`. If a tech debt issue were created to move the test fixtures into files and re-arrange / generalize the test setup, I'd be OK with this moving ahead.

test/unit/runner_test.rb
zenspider: It's the "mock" way of fetching. Totally undocumented but that's a separate issue.

miah (approved):

PR #4652, opened 2019-10-30
Fix building the website

### LINKED ISSUE #4613: Website is unbuildable
The readme hints that if there is a problem rendering CSS that it is because node is missing. I installed node, and still get the same error. There is a disconnect between what the Rakefile is doing and what needs to be available to do it.

btm:
- compass is now built in to middleman 4, including the old middleman-compass gem only creates issues with the dep graph
- deal with .css.scss files being ignored by splitting into separate .css and .scss files. (https://github.com/middleman/middleman-sprockets/issues/127)

Closes #4613


miah (approved):
omg. i would have never figured this out!

zenspider (approved):

Issue #3611, opened 2018-11-14
nginx_conf doesn't parse map sections with unquoted sources


KrisShannon:
### 🎛 Description

🐛 bug report

The map directive can contain both wildcard and dot-prefixed domains as well as unquoted regular expressions as the source.

This is not being matched by the `identifier` parsing which fails the parsing of the whole config.

#2278 was an earlier bug related to this which was fixed in #2292 by allowing identifiers to be quoted,  but nginx does not require the quoting unless there is whitespace or semicolons in the pattern.

### 🌍 InSpec and Platform Version

inspec version 3.0.0

```
Welcome to the interactive InSpec Shell
To find out how to use it, type: help

You are currently running on:

    Name:      debian
    Families:  debian, linux, unix, os
    Release:   8.5
    Arch:      x86_64
```

### 🤔 Replication Case

/etc/nginx/nginx.conf:
```
http {
    map $http_host $site_name {
        hostnames;
        default test;
        ~^www\..*$  production;
        ~^stage\..*$  staging;
        ~^test\..*$  test;
        ~^qa\..*$  qa;
        .example.com test;
        *.example.org qa;
    }
}
```

```
describe nginx_conf do
    its(:params) { should_not be_empty }
end
```

KrisShannon:
Looking at the [parser in the nginx source](https://github.com/nginx/nginx/blob/release-1.17.8/src/core/ngx_conf_file.c#L503-L817) I'm not sure why there is a separate `identifier` rule

An identifier is the same as any other value.

I have tried deleting `standard_identifier` and `quoted_identifier` and replacing identifier with

```ruby
  rule(:identifier) do
    value.as(:identifier)
  end
```

This is working for me.

Should I create a pull request with this change?

PR #4653, opened 2019-10-30
Nginx Resource: Add parsing support for wildcard, dot prefix, and regex

landychan:
My changes updates the nginx config parser to support wildcard, dot, and regex(~) prefixed keys. It fixes the issues #3611 has reported, but it does not support rows without a key/value pair.

Parses correctly
```
http {
    map $http_host $site_name {
        default test;
        ~^www\..*$  production;
        ~^stage\..*$  staging;
        ~^test\..*$  test;
        ~^qa\..*$  qa;
        .example.com test;
        *.example.org qa;
    }
}
```
Output 
```
{"http"=>
  [{"map"=>
     [{"_"=>["$http_host", "$site_name"],
       "default"=>[["test"]],
       "~^www\\..*$"=>[["production"]],
       "~^stage\\..*$"=>[["staging"]],
       "~^test\\..*$"=>[["test"]],
       "~^qa\\..*$"=>[["qa"]],
       ".example.com"=>[["test"]],
       "*.example.org"=>[["qa"]]}]}]}
```

Doesn't parse correctly.
```
http {
    map $http_host $site_name {
        hostnames;
        default test;
        ~^www\..*$  production;
        ~^stage\..*$  staging;
        ~^test\..*$  test;
        ~^qa\..*$  qa;
        .example.com test;
        *.example.org qa;
    }
}
```
Output
```
{"http"=>
  [{"map"=>
     [{"_"=>["$http_host", "$site_name"],
       "hostnames;"=>[["default", "test"]],
       "~^www\\..*$"=>[["production"]],
       "~^stage\\..*$"=>[["staging"]],
       "~^test\\..*$"=>[["test"]],
       "~^qa\\..*$"=>[["qa"]],
       ".example.com"=>[["test"]],
       "*.example.org"=>[["qa"]]}]}]}
```

Signed-off-by: Landy Chan Landychan92@gmail.com

## Related Issue
#3611

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/utils/nginx_parser.rb
+++ b/lib/inspec/utils/nginx_parser.rb
@@ -25,8 +25,12 @@ class NginxParser < Parslet::Parser
     str('"') >> (str('"').absent? >> any).repeat.as(:identifier) >> str('"') >> space.repeat
   end
 
+  rule(:regex_identifier) do
+    (str("~") >> match('\S').repeat).as(:identifier) >> space >> space.repeat
+  end
+
   rule(:identifier) do
-    standard_identifier | quoted_identifier
+    standard_identifier | quoted_identifier | regex_identifier
   end
 
   rule(:standard_value) do
--- a/lib/inspec/utils/nginx_parser.rb
+++ b/lib/inspec/utils/nginx_parser.rb
@@ -18,19 +18,15 @@ class NginxParser < Parslet::Parser
   end
 
   rule(:standard_identifier) do
-    (match("[a-zA-Z]") >> match('\S').repeat).as(:identifier) >> space >> space.repeat
+    (match("[a-zA-Z~*.]") >> match('\S').repeat).as(:identifier) >> space >> space.repeat
   end
 
   rule(:quoted_identifier) do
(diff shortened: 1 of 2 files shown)


miah (approved):
Thank you @landychan !

zenspider (approved):

clintoncwolfe (approved):
Thanks!

KrisShannon:
I think if you change the `match('\S')` to `match('[^\s;]')` and drop the extra `>> space` then even `hostnames;` will parse correctly.

KrisShannon:
Actually the logic in `rule(:standard_value)` is pretty much exactly correct.

Looking at the [parser in the nginx source](https://github.com/nginx/nginx/blob/release-1.17.8/src/core/ngx_conf_file.c#L503-L817) I'm not sure why you need a separate `identifier` rule

An identifier is the same as any other value.

Deleting `standard_identifier` and `quoted_identifier` and replacing identifier with

```ruby
  rule(:identifier) do
    value.as(:identifier)
  end
```

would be a minimal change that should work.

Issue #3655, opened 2018-12-07
oracledb_session escaping


philwinder:
### 🎛 Description

Bug

When using the `as_os_user` option, the queries don't appear to be escaped properly.

(This is in addition to #3594 #2360 and #2907 - I also have problems like these people, but I didn't want to duplicate issues) 

### 🌍 InSpec and Platform Version

`3.0.52`

### 🤔 Replication Case

I'm leaving my options in here, because they could be relevant. There's a lot of complexity in the code.

```
sql = oracledb_session(as_os_user: 'oracle', as_db_role: 'sysdba', service: 'OWSSWDEV', sqlplus_bin: '/home/oracle/ora01/app/oracle/product/12.2.0.1/db_1/bin/sqlplus')

describe sql.query('SELECT NAME FROM v$database;').row(0).column('name') do
  its('value') { should cmp 'OWSSWDEV' }
end
```

Results in:

```
Could not execute the sql query
SQL*Plus: Release 12.2.0.1.0 Production on Fri Dec 7 10:37:30 2018

Copyright (c) 1982, 2016, Oracle.  All rights reserved.


Connected to:
Oracle Database 12c Enterprise Edition Release 12.2.0.1.0 - 64bit Production

SQL> SQL&gt; <br>
SQL&gt; <br>
SQL&gt; <br>
<pre>
SELECT NAME FROM v
                 *
</pre>
ERROR at line 1:
<br>
ORA-00942: table or view does not exist
<br>


SQL&gt; <br>
```

### 💁 Possible Solutions

If I add a load of escapes infront of the `$`, it works:

```
describe sql.query('SELECT NAME FROM v\\\\$database;').row(0).column('name') do
  its('value') { should cmp 'OWSSWDEV' }
end
```

```
  SQL Column
     ✔  value should cmp == "OWSSWDEV"
```

aaronlippold:
In the postgres_session resource we used the `shellwords` gem to address this issue. https://github.com/inspec/inspec/blob/master/lib/resources/postgres_session.rb#L63 any reason that wouldn't work here?

Issue #3255, opened 2018-07-26
make oracledb_session platform independent


ehaselwanter:
### 🎛 Description

🐛 bug report

The documentation states the oracledb_session is platform independent, but it fails to run on windows.

### 🌍 InSpec and Platform Version

macOS

```
chef -v
Chef Development Kit Version: 3.1.0
chef-client version: 14.2.0
delivery version: master (6862f27aba89109a9630f0b6c6798efec56b4efe)
berks version: 7.0.4
kitchen version: 1.22.0
inspec version: 2.1.72
```

but with latest changes from https://github.com/inspec/inspec/commit/a16877f4273c2952270d698f7f8ae3cc5ff9915b

### 🤔 Replication Case

```
sql = oracledb_session(user: 'u', password: 'p', service: 'XE', sqlplus_bin: 'sqlplus', as_os_user: nil)

describe sql.query("select * from user_tablespaces where tablespace_name LIKE 'TEST'").row(0).column('tablespace_name') do
  its('value') { should eq 'TEST' }
end
```

### 💁 Possible Solutions

basically add powershell heredoc to https://github.com/inspec/inspec/blob/master/lib/resources/oracledb_session.rb#L69

```
      # use heredoc to ship the sql as a multiline string
      # we need to do it differently in powershell (prefix) and bash (postfix)
      if inspec.os.windows?
        sql_prefix = "@'\n#{opts}\n#{query}\nexit\n'@ | "
        sql_postfix = ''
      else
        sql_postfix = " <<EOC\n#{opts}\n#{query}\nEXIT\nEOC"
        sql_prefix = ''
      end

      if @db_role.nil? # rubocop:disable Style/ConditionalAssignment
        command = %(#{sql_prefix}#{bin} #{@user}/#{@password}@#{@host}:#{@port}/#{@service}#{sql_postfix})
      elsif @su_user.nil?
        command = %(#{sql_prefix}#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} as #{@db_role}#{sql_postfix})
      else
        command = %(su - #{@su_user} -c "env ORACLE_SID=#{@service} #{bin} / as #{@db_role}#{sql_postfix})
      end
```

### 💻 Stacktrace

none at hand

bytemarcz:
Hello, @ehaselwanter did you get this resource to work on a unix/linux system?

ehaselwanter:
[...]

PR #4654, opened 2019-10-31
oracledb_session fixes

miah:
- Fixes here documents on Windows. Windows uses a different format `@''@` and you need to pipe that to your app.
- Fixes here documents on UNIX. Bash will no longer interpolate SQL Queries into shell environment variables. You shouldn't need to use crazy escape techniques anymore! (This fix should apply to windows as well!)
- Added a bunch of tests for sqlplus, linux, and windows.
- Did _NOT_ test sqlcl / csv parsing. That will have to be separate work.

Fixes #4641 #3680 #3594 #3655 #3255

### DIFF
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -1,5 +1,4 @@
 require "inspec/resources/command"
-require "hashie/mash"
 require "inspec/utils/database_helpers"
 require "htmlentities"
 require "rexml/document"
@@ -21,8 +20,8 @@ module Inspec::Resources
       end
     EXAMPLE
 
-    attr_reader :user, :password, :host, :port, :service,
-                :db_role, :su_user, :bin
+    attr_reader :bin, :db_role, :host, :password, :port, :service,
+                :su_user, :user
 
     def initialize(opts = {})
       @user = opts[:user]
@@ -74,7 +73,7 @@ module Inspec::Resources
     # su, using a db_role
     def command_builder(format_options, query)
       verified_query = verify_query(query)
-      sql_prefix, sql_postfix = ""
+      sql_prefix, sql_postfix = "", ""
       if inspec.os.windows?
         sql_prefix = %{@'\n#{format_options}\n#{verified_query}\nEXIT\n'@ | }
       else
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -40,11 +40,7 @@ module Inspec::Resources
     end
 
     def query(q)
-      escaped_query = q.gsub(/\\/, '\\\\').gsub(/"/, '\\"')
-      # escape tables with $
-      escaped_query = escaped_query.gsub("$", '\\$')
 
-      p = nil
       # use sqlplus if sqlcl is not available
       if @sqlcl_bin && inspec.command(@sqlcl_bin).exist?
         bin = @sqlcl_bin
@@ -59,13 +55,6 @@ module Inspec::Resources
       command = command_builder(format_options, sql)
       cmd = inspec.command(command)
 
-      out = cmd.stdout + "\n" + cmd.stderr
-      if out.downcase =~ /^error/
-        # TODO: we need to throw an exception here
-        # change once https://github.com/chef/inspec/issues/1205 is in
-        warn "Could not execute the sql query #{out}"
-        DatabaseHelper::SQLQueryResult.new(cmd, Hashie::Mash.new({}))
-      end
       DatabaseHelper::SQLQueryResult.new(cmd, send(p, cmd.stdout))
     end
 
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -56,14 +56,7 @@ module Inspec::Resources
         p = :parse_html_result
       end
 
-      query = verify_query(escaped_query)
-      if @db_role.nil?
-        command = %{#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
-      elsif @su_user.nil?
-        command = %{#{bin} "#{@user}"/"#{@password}"@#{@host}:#{@port}/#{@service} as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC}
-      else
-        command = %{su - #{@su_user} -c "env ORACLE_SID=#{@service} #{bin} / as #{@db_role} <<EOC\n#{opts}\n#{query}\nEXIT\nEOC"}
-      end
+      command = command_builder(format_options, sql)
       cmd = inspec.command(command)
 
       out = cmd.stdout + "\n" + cmd.stderr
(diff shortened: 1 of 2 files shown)


lib/inspec/resources/oracledb_session.rb
```diff
@@ -39,9 +39,7 @@ def initialize(opts = {})
 
       @su_user = opts[:as_os_user]
       @db_role = opts[:as_db_role]
-
-      # we prefer sqlci although it is way slower than sqlplus, but it understands csv properly
-      @sqlcl_bin = "sql" unless opts.key?(:sqlplus_bin) # don't use it if user specified sqlplus_bin option
+      @sqlcl_bin = opts[:sqlcl_bin] || nil
```
zenspider, line 39: Is `opts[:sqlcl_bin]` going to be truthy, or is it going to be `"sql"`? It might be better as:

```ruby
@sqlcl_bin = "sql" if opts[:sqlcl_bin]
```

lib/inspec/resources/oracledb_session.rb
```diff
@@ -90,6 +83,11 @@ def to_s
 
     private
 
+    def pre_flight_check
+      return skip_resource "Option 'as_os_user' not available in Windows" if inspec.os.windows? && su_user
+      return fail_resource "Can't run Oracle checks without authentication" unless su_user && (user || password)
+      return fail_resource "You must provide a service name for the session" unless service
+    end
     def verify_query(query)
```
zenspider, line 92: blank line pls

lib/inspec/resources/oracledb_session.rb
```diff
@@ -89,8 +88,7 @@ def pre_flight_check
       return fail_resource "You must provide a service name for the session" unless service
     end
     def verify_query(query)
-      # ensure we have a ; at the end
-      query + ";" unless query.strip.end_with?(";")
+      query += ";" unless query.strip.end_with?(";")
```
zenspider, line 93: THANK YOU

lib/inspec/resources/oracledb_session.rb
```diff
@@ -87,6 +80,29 @@ def pre_flight_check
...
       return fail_resource "You must provide a service name for the session" unless service
     end
+
+    # 3 commands
+    # regular user password
+    # using a db_role
+    # su, using a db_role
+    def command_builder(format_options, query)
+      verified_query = verify_query(query)
+      sql_prefix, sql_postfix = ""
```
zenspider: should be `= "", ""`

lib/inspec/resources/oracledb_session.rb
```diff
@@ -87,6 +80,29 @@ def pre_flight_check
...
+      if inspec.os.windows?
+        sql_prefix = %{@'\n#{format_options}\n#{verified_query}\nEXIT\n'@ | }
+      else
+        sql_postfix = %{ <<'EOC'\n#{format_options}\n#{verified_query}\nEXIT\nEOC}
+      end
+
+      if @db_role.nil?
+        %{#{sql_prefix}#{bin} "#{user}"/"#{password}"@#{host}:#{port}/#{@service}#{sql_postfix}}
+      elsif @su_user.nil?
+        %{#{sql_prefix}#{bin} "#{user}"/"#{password}"@#{host}:#{port}/#{@service} as #{@db_role}#{sql_postfix}}
```
zenspider, line 86: I can't help but think that there are more opportunities to clean this up. Something like

```ruby
sql_postfix = " as #{@db_role}#{sql_postfix}" if @db_role
```

and then remove the first `if` entirely.

lib/inspec/resources/oracledb_session.rb
```diff
@@ -59,13 +55,6 @@ def query(q)
       command = command_builder(format_options, sql)
       cmd = inspec.command(command)
 
-      out = cmd.stdout + "\n" + cmd.stderr
-      if out.downcase =~ /^error/
-        # TODO: we need to throw an exception here
-        # change once https://github.com/chef/inspec/issues/1205 is in
-        warn "Could not execute the sql query #{out}"
-        DatabaseHelper::SQLQueryResult.new(cmd, Hashie::Mash.new({}))
-      end
```
zenspider, line 85: lovely. thank you. Down with Hashie.

lib/inspec/resources/oracledb_session.rb
```diff
@@ -69,9 +69,9 @@ def to_s
     private
 
     def pre_flight_check
-      return skip_resource "Option 'as_os_user' not available in Windows" if inspec.os.windows? && su_user
-      return fail_resource "Can't run Oracle checks without authentication" unless su_user && (user || password)
-      return fail_resource "You must provide a service name for the session" unless service
+      skip_resource "Option 'as_os_user' not available in Windows" if inspec.os.windows? && su_user
+      fail_resource "Can't run Oracle checks without authentication" unless su_user && (user || password)
+      fail_resource "You must provide a service name for the session" unless service
```
zenspider: I'm not entirely sure if it actually matters, but I think you want the `return`s so the first one wins rather than the last?

test/unit/resources/oracledb_session_test.rb
```diff
@@ -30,16 +31,42 @@
...
+    resource = quick_resource(:oracledb_session, :linux, user: "USER", password: "password", host: "localhost", service: "ORCL", as_db_role: "dbrole", as_os_user: "osuser")
+    _(resource.user).must_equal "USER"
+    _(resource.password).must_equal "password"
     _(resource.host).must_equal "localhost"
-    _(resource.service).must_equal "ORCL.localdomain"
+    _(resource.port).must_equal "1521"
+    _(resource.service).must_equal "ORCL"
+    _(resource.db_role).must_equal "dbrole"
+    _(resource.su_user).must_equal "osuser"
+    _(resource.bin).must_equal "sqlplus"
```
zenspider, line 70: fantastic

zenspider (approved):
I like this a lot. I gave some feedback with some changes suggested and some questions I wasn't sure about... but this is great. Thank you.

lib/inspec/resources/oracledb_session.rb
```diff
@@ -92,9 +68,30 @@ def to_s
 
     private
 
+    # 3 commands
+    # regular user password
+    # using a db_role
+    # su, using a db_role
```
clintoncwolfe, line 73: Could this be real RDoc (yes, it is private), or at least full English sentences? The comments don't clarify because they themselves are unclear.

lib/inspec/resources/oracledb_session.rb
```diff
@@ -21,69 +21,45 @@ class OracledbSession < Inspec.resource(1)
       end
     EXAMPLE
 
-    attr_reader :user, :password, :host, :service, :as_os_user, :as_db_role
-    # rubocop:disable Metrics/PerceivedComplexity,Metrics/CyclomaticComplexity
+    attr_reader :user, :password, :host, :port, :service,
+                :db_role, :su_user, :bin
```
clintoncwolfe: There are enough of these now that alphabetizing them would make them more manageable.

lib/inspec/resources/oracledb_session.rb
```diff
@@ -21,69 +21,45 @@ class OracledbSession < Inspec.resource(1)
...
-
-      out = cmd.stdout + "\n" + cmd.stderr
-      if out.downcase =~ /^error/
-        # TODO: we need to throw an exception here
-        # change once https://github.com/chef/inspec/issues/1205 is in
-        warn "Could not execute the sql query #{out}"
-        DatabaseHelper::SQLQueryResult.new(cmd, Hashie::Mash.new({}))
-      end
-      DatabaseHelper::SQLQueryResult.new(cmd, send(p, cmd.stdout))
+      command = command_builder(format_options, sql)
```
clintoncwolfe, line 57: 💯

clintoncwolfe (approved):
Two minor clarity concerns but overall this is fantastic - great work!

lib/inspec/resources/oracledb_session.rb
miah: Ya I was thinking about how to clean this up more too. I think thats a good suggestion.

lib/inspec/resources/oracledb_session.rb
miah: I dont think I need returns at all on these methods. _all_ they do is set some instance variables.

https://github.com/inspec/inspec/blob/master/lib/inspec/plugin/v1/plugin_types/resource.rb#L136-L148

The runner ultimately checks these values and if the resource isn't failed or skipped it gets added.

https://github.com/inspec/inspec/blob/master/lib/inspec/runner.rb#L265-L274

lib/inspec/resources/oracledb_session.rb
miah: Sure I can fix that.

lib/inspec/resources/oracledb_session.rb
miah: Ya I need to think about this one more. We're using the values of @sqlcl_bin and @sqlplus_bin to determine which the user wants to use.

PR #4665, opened 2019-11-03
Fix auditd resource processing of action and list

### LINKED ISSUE #4664: The auditd resource does not properly process lists and actions
Per auditctl(8), the `-a` option can have either `list,action` or `action,list`. However, the current implementation only allows for `action,list`.

## Possible Solution

PR incoming that matches on valid actions for `action` and adds the remainder to `list`.

trevor-vaughan:
Per auditctl(8), the -a option can have either list,action or
action,list. This PR matches against valid actions for the action field
and passed the remainder off to the list field.

Closes #4664

### DIFF
--- a/lib/inspec/resources/auditd.rb
+++ b/lib/inspec/resources/auditd.rb
@@ -187,8 +187,20 @@ module Inspec::Resources
       line.scan(/-S ([^ ]+)\s?/).flatten.first.split(",")
     end
 
+    # Processes the line and returns a pair of entries reflecting the 'action'
+    # and 'list' items.
+    #
+    # @return [Array[String,String]]
     def action_list_for(line)
-      line.scan(/-a ([^,]+),([^ ]+)\s?/).flatten
+      action_list = line.scan(/-a ([^,]+),([^ ]+)\s?/).flatten
+
+      # Actions and lists can be in either order
+      valid_actions = %w{never always}
+
+      [
+        (action_list & valid_actions).first,
+        (action_list - valid_actions).first,
+      ]
     end
 
     def key_for(line)
--- a/test/fixtures/cmd/auditctl
+++ b/test/fixtures/cmd/auditctl
@@ -2,7 +2,7 @@
 -a always,exit -F arch=b32 -S open,openat -F exit=-EPERM -F key=access
 -a always,exit -F arch=b32 -S chmod,fchmod,fchmodat -F auid>=500 f24!=0 -F key=perm_mod
 -a always,exit -S all -F path=/usr/bin/chage -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged
--a always,exit -S all -F path=/usr/bin/mount -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged
+-a exit,always -S all -F path=/usr/bin/mount -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged
 -w /etc/ssh/sshd_config -p rwxa -k CFG_sshd_config
 -w /etc/sudoers -p wa
 -w /etc/private-keys -p x
--- a/test/unit/resources/auditd_test.rb
+++ b/test/unit/resources/auditd_test.rb
@@ -10,11 +10,27 @@ describe "Inspec::Resources::AuditDaemon" do
       "-a always,exit -F arch=b32 -S open,openat -F exit=-EPERM -F key=access",
       "-a always,exit -F arch=b32 -S chmod,fchmod,fchmodat -F auid>=500 f24!=0 -F key=perm_mod",
       "-a always,exit -S all -F path=/usr/bin/chage -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged",
-      "-a always,exit -S all -F path=/usr/bin/mount -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged",
+      # Reversed for action_list validation
+      "-a exit,always -S all -F path=/usr/bin/mount -F perm=x -F auid>=1000 -F auid!=-1 -F key=privileged",
       "-w /etc/ssh/sshd_config -p rwxa -k CFG_sshd_config",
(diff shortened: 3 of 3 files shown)


clintoncwolfe (changes requested):
Hey, looks like the linter didn't like the array literal:
```

lib/inspec/resources/auditd.rb:107:23: C: Style/WordArray: Use %w or %W for an array of words.
--
  | valid_actions = ['never', 'always']
  | ^^^^^^^^^^^^^^^^^^^
  | lib/inspec/resources/auditd.rb:107:24: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  | valid_actions = ['never', 'always']
  | ^^^^^^^
```

`bundle exec chefstyle` should show the issue (and add -a for auto-correct).

aaronlippold (approved):
Looks good to me

lib/inspec/resources/auditd.rb
```diff
@@ -102,8 +102,14 @@ def parse_content
     end
 
     def file_syscall_syntax_rules_for(line)
+      # Actions and lists can be in either order
+      valid_actions = %w{never always}
+      action_list = action_list_for(line)
+
+      action = (action_list & valid_actions)
+      list = (action_list - valid_actions)
```
clintoncwolfe: Look like there is a failing test, because these both end up as Arrays, rather than the expected String (presumably the first element in the remaining Array).

```

1) Failure:
--
  | Inspec::Resources::AuditDaemon#test_0006_check auditd file watch with syscall syntax [/workdir/test/unit/resources/auditd_test.rb:50]:
  | expected | 0s
  | actual | 1s
  | @@ -1 +1 @@
  | -[{"file"=>"/usr/bin/chage", "list"=>"exit", "action"=>"always", "fields"=>["path=/usr/bin/chage", "perm=x", "auid>=1000", "auid!=-1", "key=privileged"], "permissions"=>["x"], "key"=>"privileged", "fields_nokey"=>["path=/usr/bin/chage", "perm=x", "auid>=1000", "auid!=-1"]}]
  | +[{"file"=>"/usr/bin/chage", "list"=>["exit"], "action"=>["always"], "fields"=>["path=/usr/bin/chage", "perm=x", "auid>=1000", "auid!=-1", "key=privileged"], "permissions"=>["x"], "key"=>"privileged", "fields_nokey"=>["path=/usr/bin/chage", "perm=x", "auid>=1000", "auid!=-1"]}]
  |  


```

clintoncwolfe (changes requested):
Looks like there is a failing test due to having the wrong structure.

Also, before we can accept this, I think we'll need a unit test showing it working with with list-first syntax - the unit test file is at https://github.com/inspec/inspec/blob/master/test/unit/resources/auditd_test.rb and the mock file is at https://github.com/inspec/inspec/blob/master/test/unit/mock/cmd/auditctl

lib/inspec/resources/auditd.rb
zenspider: `make_my_diffs_pretty!` really helps with these types of failures...

A change like this probably needs to be driven via a failing test. I'm also concerned that the change went into `file_syscall_syntax_rules_for` but not `syscall_rules_for`. I would assume that since they both call `action_list_for` that they'd both be affected.

zenspider (changes requested):

trevor-vaughan:
Ok, moving the split down to `action_list_for` and modifying all calls to do the right thing.

Will add a test.

trevor-vaughan:
Ok, so I looked at the testing structure but the functions aren't unit tested directly and I couldn't figure out how to do red/green testing with the existing code.

The current tests pass (no regressions) and it works on a real system.

clintoncwolfe (changes requested):
What I am looking for is additional lines added to the file at `test/unit/mock/cmd/auditctl` in the form that you are now exercising:

```
-a exit,always -F arch=b64 -S open,openat -F exit=-EACCES -F key=access
```

Ideally, we'd like to see a test added to `test/unit/resources/auditd_test.rb` explicitly exercising the list-first example(s).  Thanks!

lib/inspec/resources/auditd.rb
```diff
@@ -187,8 +187,20 @@ def syscalls_for(line)
...
+      action_list = line.scan(/-a ([^,]+),([^ ]+)\s?/).flatten
+
+      # Actions and lists can be in either order
+      valid_actions = %w{never always}
+
+      [
+        (action_list & valid_actions).first,
+        (action_list - valid_actions).first
+      ]
     end
```
zenspider, line 204: > Ok, so I looked at the testing structure but the functions aren't unit tested directly and I couldn't figure out how to do red/green testing with the existing code.

You can just write very plain unit tests for this that pass in the line and assert what you expect on the way out. Something like:

```ruby
resource = MockLoader.new(:centos7).load_resource("auditd") # which I don't like but it's there...
assert_equal ["a", "b"], resource.action_list_for("input line")
```

As long as you have covered both the positive and negative edge cases, then I think this is fine.

zenspider (changes requested):

clintoncwolfe (approved):
This looks good to me now - the one last thing I ask would be to leave parallization enabled on the lint task in the Rakefile, thanks!

trevor-vaughan:
@clintoncwolfe `--parallel` broke `test:lint:auto_correct`. Do you want to leave that broken and re-enable parallel?

clintoncwolfe:
Ahh, that's charming. That sounds like a separate issue to fix with a small amount of logic to detect the auto-correct flag - but I'm OK with you leaving it broken (with `--parallel`). I personally prefer separate changes on separate PRs.

trevor-vaughan:
@clintoncwolfe and @zenspider Any other issues that need to be resolved on this?

zenspider (approved):
This still has insufficient testing as pointed out previously... but I want this off my plate.

PR #4670, opened 2019-11-04
Pre-load the gem resource to override the global gem method.

zenspider:
This seems problematic to me in general. We should probably not
override global methods. Renaming this to rubygem and providing a
deprecated alias seems like the right thing to do.


miah (approved):
Can we file a issue to cover the TODO's?

clintoncwolfe (approved):
Regrettable but effective - thanks

PR #4674, opened 2019-11-06
Revamp www/doco setup

### LINKED ISSUE #4463: Website resource contrib documentation contains duplicated (aws) entries
# Describe the problem

The [AWS/Azure/GCP resources documentation](https://www.inspec.io/docs/reference/resources/#aws-resources) contains many duplicated entries.

See the screenshot below:
Screenshot from 2019-09-10 15-00-17
Almost of the entries are present twice: the text is the same and the target link is the same for every 2 entries.

## Possible Solution
I tracked down the problem to that [Rake docs task](https://github.com/inspec/inspec/blob/master/tasks/docs.rb#L302) where the call to `Dir.glob` seems to be made *after* the `.md.erb` files have been transformed into `.md` files. `Dir.glob` then returns both files which results in 2 duplicated entry.

### LINKED ISSUE #4609: Website build process is dangerously destructive and can't be automated
The latter part of the title is #3859.

But the former? The thing is doing this:

```ruby
    Dir.chdir(File.join(Dir.pwd, "..")) do
      dst_from_root = File.join("www", dst)
      sh("git rm -rf .")
      sh("git add #{dst_from_root}")
      sh("git clean -df")
```

which is to say, back up a directory (into your working tree for inspec), nuke EVERYTHING, then run `git clean` with force. 

## This. Is. Bad.

It also interactively asks questions which means expeditor can't do this stuff. I think this is doing WAY too much and could be simplified tremendously.

zenspider:
Fixes: #4463.
Fixes: #4609.


miah (approved):

miah (approved):

www/Rakefile
```diff
@@ -24,206 +7,64 @@ require "fastly"
...
-    end
-
-    if purge_result["status"] == "ok"
-      puts "Success! The Fastly cache is purged."
-    else
-      puts "ERROR: Fastly cache could not be purged: #{purge_result}"
+      sh("git worktree remove -f #{path}/gh-pages")
     end
   end
 end
```
clintoncwolfe: We are indeed behind a Fastly cache; I have a Fastly token and this code being removed here works to clear the cache. It looks like the default TTL is one hour. 

I think we should keep the Fastly clearing code.

clintoncwolfe (changes requested):
Looks good, but I think we need to retain the Fastly cache clearing code.

www/Rakefile
zenspider: I was watching it refresh immediately whenever I pushed. Not sure what the difference is.

miah: Looks like cache-control is setting the pages to expire after 10 minutes. But I bet if you hit a page that was updated recently github will know that too and sets the right headers so fastly updates the cache.

```
$ curl -svL http://inspec.io 2>&1 |grep cache
< X-Served-By: cache-pao17445-PAO
< cache-control: max-age=600
< x-proxy-cache: MISS
< x-served-by: cache-pao17421-PAO, cache-pao17430-PAO
< x-cache: MISS, HIT
< x-cache-hits: 0, 1
```

https://docs.fastly.com/en/guides/controlling-caching

That said, I'm +1 on ditching this code.

As this task will be moved into automation in buildkite, we'll end up with something like this: https://expeditor.chef.io/docs/reference/actions/#purge-packages-chef-io-fastly

zenspider: I've added it back but I'd rather ditch it and remove another dependency.

I don't think anything we do on the website is mission critical enough to warrant the complexity.

PR #4678, opened 2019-11-07
Add several tests to our artifact suite

miah:
To clarify what these are for.

When we build InSpec, we want to test the final built artifact. This is a simple set of tests to smoke test from the outside. We cover all the subcommands just to make sure we didn't miss anything in our prior tests.

### DIFF
--- a/test/artifact/artifact_helper.rb
+++ b/test/artifact/artifact_helper.rb
@@ -1,7 +1,7 @@
 require "minitest/autorun"
 require "open3"
 
-TEST_CLI_OPTS="--chef-license=accept-no-persist"
+TEST_CLI_OPTS = "--chef-license=accept-no-persist".freeze
 
 class ArtifactTest < Minitest::Test
   make_my_diffs_pretty!
@@ -10,7 +10,7 @@ class ArtifactTest < Minitest::Test
     command = "inspec #{inspec_command} #{TEST_CLI_OPTS}"
     stdout, stderr, status = Open3.capture3({ "PATH" => ENV["PATH"] },
                                             command,
-                                            { :chdir => ENV["project_root"]})
+                                            { chdir: ENV["project_root"] })
 
     assert_empty stderr.sub(/#< CLIXML\n/, "")
     assert stdout
--- a/test/artifact/artifact_helper.rb
+++ b/test/artifact/artifact_helper.rb
@@ -9,7 +9,8 @@ class ArtifactTest < Minitest::Test
   def assert_artifact(inspec_command)
     command = "inspec #{inspec_command} #{TEST_CLI_OPTS}"
     stdout, stderr, status = Open3.capture3({ "PATH" => ENV["PATH"] },
-                                            command)
+                                            command,
+                                            { :chdir => ENV["project_root"]})
 
     assert_empty stderr.sub(/#< CLIXML\n/, "")
     assert stdout
--- a/test/artifact/artifact_helper.rb
+++ b/test/artifact/artifact_helper.rb
@@ -6,8 +6,8 @@ TEST_CLI_OPTS="--chef-license=accept-no-persist"
 class ArtifactTest < Minitest::Test
   make_my_diffs_pretty!
 
-  def assert_artifact(command)
-    command = "inspec #{command} #{TEST_CLI_OPTS}"
+  def assert_artifact(inspec_command)
+    command = "inspec #{inspec_command} #{TEST_CLI_OPTS}"
     stdout, stderr, status = Open3.capture3({ "PATH" => ENV["PATH"] },
                                             command)
 
--- a/test/artifact/artifact_helper.rb
+++ b/test/artifact/artifact_helper.rb
@@ -1,0 +1,4 @@
+require "minitest/autorun"
+require "open3"
+
+TEST_CLI_OPTS="--chef-license=accept-no-persist"
--- a/test/artifact/artifact_helper.rb
+++ b/test/artifact/artifact_helper.rb
@@ -2,3 +2,17 @@ require "minitest/autorun"
 require "open3"
 
 TEST_CLI_OPTS="--chef-license=accept-no-persist"
+
+class ArtifactTest < Minitest::Test
+  make_my_diffs_pretty!
+
+  def assert_artifact(command)
+    command = "inspec #{command} #{TEST_CLI_OPTS}"
+    stdout, stderr, status = Open3.capture3({ "PATH" => ENV["PATH"] },
+                                            command)
+
+    assert_empty stderr.sub(/#< CLIXML\n/, "")
+    assert stdout
+    assert status
+  end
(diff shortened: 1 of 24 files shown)


test/artifact/artifact_installation_test.rb
```diff
@@ -0,0 +1,21 @@
...
+    assert_empty stderr
+    assert status
+  end
+
+  def test_inspec_exists_windows
+    skip unless windows?
+    stdout, stderr, status = Open3.capture3("Get-Command inspec")
+    assert_match(/inspec.bat/, stdout)
+    assert_empty stderr
+    assert status
```
zenspider, line 18: I think these should be merged and the conditional inside determines what command and regexp to use.

test/artifact/inspec_check_test.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "artifact_helper"
+
+class TestInspecCheck < Minitest::Test
+  parallelize_me!
```
zenspider: this isn't needed since there is only one test.

test/artifact/inspec_compliance_test.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "artifact_helper"
+
+class TestInspecCompliance < Minitest::Test
+  parallelize_me!
```
zenspider: same

test/artifact/inspec_detect_test.rb
```diff
@@ -1,24 +1,15 @@
-require "minitest/autorun"
+require_relative "artifact_helper"
+require "open3"
```
zenspider: nuke open3 since it is in artifact_helper.

test/artifact/inspec_env_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class TestInspecEnv < Minitest::Test
+  parallelize_me!
+  def test_env
+    command = "/bin/inspec env #{TEST_CLI_OPTS}"
+    stdout, stderr, status = Open3.capture3(command)
+
+    assert_empty stderr.sub(/#< CLIXML\n/, "")
+    assert stdout
+    assert status
```
zenspider: At this point, it seems like there's enough repetition that this should be refactored into its own test type and a compound assertion to clean these up.

test/artifact/inspec_plugin_test.rb
```diff
@@ -0,0 +1,15 @@
...
+class TestInspecPlugin < Minitest::Test
+  parallelize_me!
+
+  def test_plugin
+    flunk
+    command = "/bin/inspec plugin #{TEST_CLI_OPTS}"
+    stdout, stderr, status = Open3.capture3(command)
+
+    assert_empty stderr.sub(/#< CLIXML\n/, "")
+    assert stdout
```
zenspider: I suspect that this line can't fail because `capture3` always returns a String. I don't think we want to test the contents tho, other than maybe being non-empty?

test/artifact/inspec_supermarket_test.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "artifact_helper"
+
+class TestInspecSupermarket < Minitest::Test
+  parallelize_me!
+
+  def test_supermarket
+    flunk
```
zenspider: should probably be a skip... this is on a fresh pipeline, so even an `S` output will stand out and you'll want to fix them.

zenspider (changes requested):

test/artifact/inspec_plugin_test.rb
miah: Yes. In most cases I just want to make sure its not-empty. I generally don't care about the string contents here.

test/artifact/inspec_compliance_test.rb
```diff
@@ -0,0 +1,12 @@
+require_relative "artifact_helper"
+
+class TestInspecCompliance < Minitest::Test
+  def test_compliance_version
+    command = "/bin/inspec compliance version #{TEST_CLI_OPTS}"
```
clintoncwolfe: Note that `version` is a live-fire API command - it tries to fetch the *server* API version. If you want an innocuous command, you might use `help` or `logout`.

clintoncwolfe (changes requested):
Do you see these test scripts diverging over time? Right now it definitely seems that factoring them out to a shared testing function would make sense as they are all internally very similar. That would be my only major change.

test/artifact/artifact_installation_test.rb
```diff
@@ -0,0 +1,20 @@
...
+class TestArtifactInstallation < Minitest::Test
+  parallelize_me!
+
+  def test_inspec_exists
+    if windows?
+      stdout, stderr, status = Open3.capture3("Get-Command inspec")
+      assert_match(/inspec.bat/, stdout)
+    else
+      stdout, stderr, status = Open3.capture3({ "PATH" => ENV["PATH"] },
+                                              "/usr/bin/which inspec")
```
miah, line 12: Considering this is running _in hab_ i might be able to use which on both sides.

test/artifact/artifact_helper.rb
```diff
@@ -0,0 +1,18 @@
+require "minitest/autorun"
+require "open3"
+
+TEST_CLI_OPTS="--chef-license=accept-no-persist"
+
+class ArtifactTest < Minitest::Test
+  make_my_diffs_pretty!
+
```
miah, line 8: I should also ensure we have a `skip_windows!` method (https://github.com/inspec/inspec/blob/d9339eab024ab80eebb0f30a4d33eca0395a0ac4/test/helper.rb#L151)

test/artifact/inspec_env_test.rb
```diff
@@ -0,0 +1,7 @@
+require_relative "artifact_helper"
+
+class TestInspecEnv < ArtifactTest
+  def test_env
+    assert_artifact(:env)
```
miah: This fails on windows, so should be skipped for now and re-written to expect the failure.

zenspider (approved):

miah:
https://buildkite.com/chef/inspec-inspec-master-artifact-habitat/builds/107#_

https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/103

test/artifact/artifact_helper.rb
```diff
@@ -0,0 +1,19 @@
+require "minitest/autorun"
+require "open3"
+
+TEST_CLI_OPTS = "--chef-license=accept-no-persist".freeze
+
+class ArtifactTest < Minitest::Test
+  make_my_diffs_pretty!
+
+  def assert_artifact(inspec_command)
```
clintoncwolfe, line 9: `assert_artifact_invocation` might read a little more clearly

clintoncwolfe (approved):
Thanks!

PR #4722, opened 2019-11-13
Dropping chef-core from dependencies.

### LINKED ISSUE #4716: InSpec 4.18 introduces complex dependency chain via chef-core
# Describe the problem
InSpec 4.18 added telemetry support by using the chef-core gem. This gem was initially created by the workstation team to extract some logic in the workstation app out into a shared library. This library is no longer actively maintained and includes logic only used by the chef-apply command. By using it in InSpec we introduce 2 separate points where we depend on train-core (chef gem and chef-core gem).  Historically this sort of double dependency has created huge pains for us and made it difficult to ship products without flow charts detailing the ceremony involved. The chef-core gem really needs to go away since it's not supported and InSpec needs to depend on the telemetry gem directly instead. If that means moving logic into that gem and taking it out of chef-core then that work needs to be prioritized. Until then the current dependency setup of chef-core makes it so we can't bring new versions of InSpec into the chef package.

### LINKED ISSUE #4719: Remove chef-core dependency
In preparation for moving to chef-telemetry.

zenspider:
We'll be moving some code from chef-core to chef-telemetry and adding
that dependency in the next sprint.

Fixes #4716.
Fixes #4719.

### DIFF
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -21,7 +21,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "train-core", "~> 3.0"
   spec.add_dependency "license-acceptance", ">= 0.2.13", "< 2.0"
-  spec.add_dependency "chef-core", "~> 0.0"
   spec.add_dependency "thor", "~> 0.20"
   spec.add_dependency "json-schema", "~> 2.8"
   spec.add_dependency "method_source", "~> 0.8"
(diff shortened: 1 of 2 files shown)


miah (approved):

clintoncwolfe:
Reverts #4569

clintoncwolfe (approved):

PR #4724, opened 2019-11-14
Add last login date to WindowsUser

### LINKED ISSUE #4723: Windows Users - Last Logon
Provide the ability to get the Last Logon property of a Windows User Account.

## Detailed Description
I need to be able to detect inactive accounts on a Windows System, and currently, this is only possible via PowerShell, as the Users resource does not expose this property.

## Context
Allows a user to get all accounts that haven't been used in a specified number of days, and then verify that they are disabled.

## Possible Implementation
Recommend implementing this via Powershell over WinRM on the Users Resource.

mbaitelman:
Signed-off-by: Mendy Baitelman <mendy@baitelman.com>

## Related Issue
Closes #4723 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -636,7 +636,7 @@ module Inspec::Resources
         shell: res[:shell],
         domain: res[:domain],
         userflags: res[:userflags],
-        lastlogin: res[:lastlogin]
+        lastlogin: res[:lastlogin],
       }
     end
 
--- a/lib/inspec/resources/users.rb
+++ b/lib/inspec/resources/users.rb
@@ -223,6 +223,10 @@ module Inspec::Resources
       meta_info[:userflags] unless meta_info.nil?
     end
 
+    def lastlogin
+      meta_info[:lastlogin] unless meta_info.nil?
+    end
+
     # returns the minimum days between password changes
     def mindays
       credentials[:mindays] unless credentials.nil?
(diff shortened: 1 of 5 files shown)


mbaitelman:
WIP: Need to add tests and PS fixes

mbaitelman:
Im having trouble figuring out how to call `test\fixtures\cmd\adsiusers` from `test\helpers\mock_loader.rb#299`. 
Im assuming I need to update the hash but not sure which algorithm to use.

lib/inspec/resources/users.rb
```diff
@@ -715,6 +720,7 @@ def collect_user_details # rubocop:disable Metrics/MethodLength
             home = $_.HomeDirectory[0]
             shell = $null
             domain = $Computername
+            lastlogin = if($_.lastlogin.getType().Tostring() -eq "System.Management.Automation.PSMethod" ){ $null }else{[String]$_.lastlogin}
```
mbaitelman, line 723: There has to be a better way to return only a date or $null.

miah: maybe! I'm going to have to dig into this one =)

miah (approved):

miah (changes requested):
The changes look fine and I +1 that, but the tests are failing so something needs to be addressed here.

mbaitelman:
Miah can you help with https://github.com/inspec/inspec/pull/4724#issuecomment-553771154?

miah:
`Digest::SHA256.hexdigest(script)`

where `script` is the command being executed (in this case several lines of powershell).

zenspider (approved):

miah (approved):

PR #4734, opened 2019-11-18
Move lib/fetchers to lib/inspec/fetchers and re-namspace

### LINKED ISSUE #4708: Move lib/fetchers to lib/inspec/fetchers and fix namespacing

clintoncwolfe:
## Description

 * move lib/fetchers to lib/inspec/fetchers
 * rename a cache object `Inspec::Fetcher` to `Inspec::FetcherSystem`
 * move classes from `Fetcher::*` to `Inspec::Fetcher::*`

## Related Issue

Fixes #4708 

## Types of changes

## Checklist:


lib/inspec/fetcher.rb
```diff
@@ -26,7 +26,7 @@ def with_default_fetcher(target)
     end
   end
 
-  Fetcher = FetcherRegistry.new
+  FetcherSystem = FetcherRegistry.new
```
miah: I feel weird calling this FetcherSystem. It seems like it should be FetcherRegistry, but I see why you can't do that here.

miah (approved):

lib/inspec/fetcher.rb
zenspider: True... but why not `Registry`? It's already namespaced under `Inspec::Fetcher`.

lib/inspec/fetcher.rb
```diff
@@ -38,9 +38,9 @@ def self.fetcher(version)
 end
 
 # TODO: remove. require up, not down.
-require "fetchers/local"
-require "fetchers/url"
-require "fetchers/git"
+require "inspec/fetchers/local"
+require "inspec/fetchers/url"
+require "inspec/fetchers/git"
```
zenspider: Since the namespace is `Inspec::Fetcher`, then maybe the directory should be renamed to singular?

lib/inspec/fetcher.rb
zenspider: Oh... but it's not under `Inspec::Fetcher`... it should be.

zenspider (changes requested):

zenspider (approved):

PR #4736, opened 2019-11-20
Make fetchers fail consistently

### LINKED ISSUE #4726: Fetchers should fail gracefully
## Detailed Description

A **fetcher** is the component that knows how to obtain an InSpec profile. There are 4 common Fetchers - URL, Git, Local (aka Path) and Automate/Compliance (a fifth, Mock, exists for testing). 

Currently, when a fetcher fails for any reason, a stacktrace appears. This is undelightful.

 * No one likes a stacktrace
 * No useful corrective action for the user to take
 * Difficult for InSpec to handle the failures in a regular way, such as attempting a retry or falling back to a caching mechanism.

## Context

Refs #4519 
Refs #3297

clintoncwolfe:
## Description

 * Modifies the `local`, `url`, and `git` fetchers to throw an `Inspec::FetcherFailure` exception if the fetch operation fails. 
 * Since `inspec` intercepts errors in that inheritance tree, the UX changes to no longer stacktrace - instead issuing the human-friendly one-line exception message.
 * The `compliance` fetcher was found to already be using this facility.

Simple test fixtures (profiles with bad dependencies) are included. Functional tests are included.

## Related Issue

Fixes #4726

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/fetcher/git.rb
+++ b/lib/inspec/fetcher/git.rb
@@ -90,8 +90,8 @@ module Inspec::Fetcher
         # vendoring to something more complex, don't bother.
         FileUtils.rmdir(destination_path) if Dir.empty?(destination_path)
 
-        raise ArgumentError, "Cannot find relative path '#{@relative_path}' " \
-                             "within profile in git repo specified by '#{@remote_url}'"
+        raise Inspec::FetcherFailure.new("Cannot find relative path '#{@relative_path}' " \
+                                         "within profile in git repo specified by '#{@remote_url}'")
       end
       FileUtils.cp_r("#{working_dir}/#{@relative_path}", destination_path)
     end
@@ -129,11 +129,11 @@ module Inspec::Fetcher
     def resolve_ref(ref_name)
       command_string = "git ls-remote \"#{@remote_url}\" \"#{ref_name}*\""
       cmd = shellout(command_string)
-      raise "Error running '#{command_string}': #{cmd.stderr}" unless cmd.exitstatus == 0
+      raise Inspec::FetcherFailure.new("Profile git dependency failed for #{@remote_url} - error running '#{command_string}': #{cmd.stderr}") unless cmd.exitstatus == 0
 
       ref = parse_ls_remote(cmd.stdout, ref_name)
       unless ref
-        raise "Unable to resolve #{ref_name} to a specific git commit for #{@remote_url}"
+        raise Inspec::FetcherFailure.new("Profile git dependency failed - unable to resolve #{ref_name} to a specific git commit for #{@remote_url}")
       end
 
       ref
@@ -191,7 +191,7 @@ module Inspec::Fetcher
       cmd.error!
       cmd.status
     rescue Errno::ENOENT
-      raise "To use git sources, you must have git installed."
+      raise Inspec::FetcherFailure.new("Profile git dependency failed for #{@remote_url} - to use git sources, you must have git installed.")
     end
 
     def shellout(cmd, opts = {})
--- a/lib/inspec/fetcher/git.rb
+++ b/lib/inspec/fetcher/git.rb
@@ -90,8 +90,8 @@ module Inspec::Fetcher
         # vendoring to something more complex, don't bother.
         FileUtils.rmdir(destination_path) if Dir.empty?(destination_path)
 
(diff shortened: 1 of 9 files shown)


lib/inspec/cached_fetcher.rb
```diff
@@ -8,7 +8,7 @@ class CachedFetcher
     attr_reader :cache, :target, :fetcher
     def initialize(target, cache)
       @target = target
-      @fetcher = Inspec::Fetcher.resolve(target)
+      @fetcher = Inspec::FetcherSystem.resolve(target)
```
zenspider: Is this PR on top of the other fetcher PR? It looks like it is based off of master.

zenspider (changes requested):

test/fixtures/profiles/fetcher-failures/git-bad/inspec.yml
```diff
@@ -0,0 +1,9 @@
+name: git-bad
+license: Apache-2.0
+summary: A profile with a bad git dependency
+version: 0.1.0
+supports:
+  platform: os
+depends:
+  - name: nonesuch
+    git: http://localhost/no/such
```
miah: If `localhost` is intended to be invalid, it should be `localhost.invalid`. `localhost` is generally a valid resolvable host. There are a few other uses of localhost in this PR this applies to as well.

https://tools.ietf.org/html/rfc2606

miah (changes requested):

test/fixtures/profiles/fetcher-failures/git-bad/inspec.yml
clintoncwolfe: Thanks for the tip, I've modified the hostnames accordingly.

lib/inspec/fetcher/git.rb
```diff
@@ -90,8 +90,8 @@ def perform_relative_path_fetch(destination_path, working_dir)
         # vendoring to something more complex, don't bother.
         FileUtils.rmdir(destination_path) if Dir.empty?(destination_path)
 
-        raise ArgumentError, "Cannot find relative path '#{@relative_path}' " \
-                             "within profile in git repo specified by '#{@remote_url}'"
+        raise Inspec::FetcherFailure.new("Cannot find relative path '#{@relative_path}' " \
+                                         "within profile in git repo specified by '#{@remote_url}'")
```
zenspider: there's no need for `.new` in these flavors of exceptions. `raise` will handle it if it just takes a message arg.

lib/inspec/fetcher/local.rb
```diff
@@ -101,7 +101,13 @@ def sha256
...
 
     def perform_shasum(target)
-      @archive_shasum ||= OpenSSL::Digest::SHA256.digest(File.read(target)).unpack("H*")[0]
+      return @archive_shasum if @archive_shasum
+
+      if File.exist?(target)
+        @archive_shasum = OpenSSL::Digest::SHA256.digest(File.read(target)).unpack("H*")[0]
+      else
+        raise Inspec::FetcherFailure.new("Profile dependency local path '#{target}' does not exist")
+      end
```
zenspider: I'm a little surprised that this didn't come up in rubocop... the `raise` could be a standalone line flattening out the entire method:

```ruby
return @archive_shasum if @archive_shasum

raise Inspec::FetcherFailure.new("Profile dependency local path '#{target}' does not exist") unless File.exist?(target)

@archive_shasum = OpenSSL::Digest::SHA256.digest(File.read(target)).unpack("H*")[0]
```

I'm also not sure given the combos above... but `H*` is zero or more (I think) so it could be possible that `[0]` comes back with nothing and thus assigns nil and never caches? The `return` line might be better off checking `defined?`

lib/inspec/fetcher/git.rb
miah: So we can just do? : 

`raise Inspec::FetcherFailure "Cannot ..."`

neat

zenspider: comma in-between, but yes

zenspider (approved):

miah (approved):

miah (changes requested):
Tests are broken.

miah (approved):

PR #4742, opened 2019-11-25
Update expeditor to skip with new labels

zenspider:

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -89,8 +89,8 @@ changelog:
 merge_actions:
  - built_in:bump_version:
     ignore_labels:
-     - "Version: Skip Bump"
      - "Expeditor: Skip All"
+     - "Expeditor: Skip Version Bump"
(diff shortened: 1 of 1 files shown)


miah (approved):

.expeditor/config.yml
```diff
@@ -89,8 +89,8 @@ changelog:
 merge_actions:
  - built_in:bump_version:
     ignore_labels:
-     - "Version: Skip Bump"
-     - "Expeditor: Skip All"
+     - "Expeditor: Skip All",
+     - "Expeditor: Skip Version Bump",
```
clintoncwolfe: I don't think you want trailing commas here.

clintoncwolfe (changes requested):

.expeditor/config.yml
zenspider: oops!

clintoncwolfe (approved):

PR #4766, opened 2019-12-05
Fix warnings coming from ruby -w

zenspider:
This first pass fixes all our circular requires and duplicate method
warnings that come from:

```
ruby -w -Ilib:test test/functional/helper.rb
```

### DIFF
--- a/lib/inspec/resources/shadow.rb
+++ b/lib/inspec/resources/shadow.rb
@@ -52,12 +52,6 @@ module Inspec::Resources
       .register_column(:inactive_days, field: "inactive_days")
       .register_column(:expiry_dates, field: "expiry_date")
       .register_column(:reserved, field: "reserved")
-    # These are deprecated, but we need to "alias" them
-    filtertable
-      .register_custom_property(:user) { |table, value| table.resource.user(value) }
-      .register_custom_property(:password) { |table, value| table.resource.password(value) }
-      .register_custom_property(:last_change) { |table, value| table.resource.last_change(value) }
-      .register_custom_property(:expiry_date) { |table, value| table.resource.expiry_date(value) }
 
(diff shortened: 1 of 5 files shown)


miah (approved):

lib/inspec/resources/shadow.rb
```diff
@@ -52,12 +52,6 @@ def initialize(path = "/etc/shadow", opts = {})
       .register_column(:inactive_days, field: "inactive_days")
       .register_column(:expiry_dates, field: "expiry_date")
       .register_column(:reserved, field: "reserved")
-    # These are deprecated, but we need to "alias" them
-    filtertable
-      .register_custom_property(:user) { |table, value| table.resource.user(value) }
-      .register_custom_property(:password) { |table, value| table.resource.password(value) }
-      .register_custom_property(:last_change) { |table, value| table.resource.last_change(value) }
-      .register_custom_property(:expiry_date) { |table, value| table.resource.expiry_date(value) }
```
clintoncwolfe, line 60: @james-stocks can we get confirmation it is safe to remove these at this point?

clintoncwolfe (commented):
I'm happy with everything, but the `shadow` aliases were there to handle problems we had a while ago. Just need to make sure we don't need them anymore.

lib/inspec/resources/shadow.rb
james-stocks: These are what let you directly call like `shadow.user`?

All our supported profiles use `shadow.where` and we don't need anything other than `where`.

zenspider: There were no aliases. There were duplicates. I removed the former, not the latter. All the method names are still there. If I removed names, we'd have test failures (I hope).

clintoncwolfe (approved):

PR #4768, opened 2019-12-06
Wire up backend to rspec metadata, allowing for example groups to use resources

### LINKED ISSUE #628: Can't use os helper from example group context
It would be nice to map the `os` method helper in to all contexts. Currently you can't do this:

``` ruby
describe file('/app') do
  it { is_expected.to be_a_directory }
  it { is_expected.to be_owned_by 'root' } unless os.windows?
end
```

It fails with `undefined local variable or method 'os' for RSpec::ExampleGroups::FileApp:Class`. Pulling the `os.windows?` into a local variable from the top level fixes this, but is clunky. I think we can probably define an `os` method in the resource base class that leaves this as a valid alias.

zenspider:
Fixes #628.

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -23,11 +23,11 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.4"
 
-  spec.add_dependency "train", "~> 3.0" # Inspec 4 must have train 2+; 3+ if we include train-winrm
+  spec.add_dependency "train", "~> 3.0"
   # Train plugins we ship with InSpec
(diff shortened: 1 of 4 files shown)


zenspider:
These all pass now:

```ruby
# fails
describe "a string" do
  os.unix?
end

# passes
puts os.unix?

# passes
control "woot" do
  os.unix?
end

# fails
describe file('/etc') do
  it { is_expected.to be_a_directory }
  it { is_expected.to be_owned_by 'root' } unless os.windows?
end
```

clintoncwolfe (approved):

miah (approved):

PR #4778, opened 2019-12-11
Removed all metaprogramming from ControlEvalContext

zenspider:
+ Removed ControlEvalContext.rule_context entirely pushing into Rule
+ Gutted ControlEvalContext.create pushing into ControlEvalContext.
+ Added class ivar accessors for ControlEvalContext to bypass need for closures.
+ Converted all define_methods into defs.
+ Added resource_dsl to Rule#initialize

### DIFF
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -110,6 +110,11 @@ module Inspec
     end
 
     def add_resources(context)
+      # # TODO: write real unit tests for this and then make this change:
+      # dsl = context.to_resources_dsl
+      # self.class.include dsl
+      # Inspec::Rule.include dsl
+
       self.class.class_eval do
         include context.to_resources_dsl
       end
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -15,34 +15,13 @@ module Inspec
     include Inspec::DSL
     include Inspec::DSL::RequireOverride
 
-    class << self
-      attr_accessor :profile_context_owner
-      attr_accessor :profile_id
-      attr_accessor :resources_dsl
-    end
-
-    # Creates the heart of the control eval context:
-    #
-    # An instantiated object which has all resources registered to it
-    # and exposes them to the test file.
-    #
-    # @param profile_context [Inspec::ProfileContext]
-    # @param outer_dsl [OuterDSLClass]
-    # @return [ProfileContextClass]
-    def self.create(profile_context, resources_dsl)
-      klass = Class.new self
-      klass.include resources_dsl
-
-      klass.profile_context_owner = profile_context
-      klass.profile_id            = profile_context.profile_id
-      klass.resources_dsl         = resources_dsl
-
-      klass
-    end
-
     attr_accessor :skip_file
+    attr_accessor :profile_context
+    attr_accessor :resources_dsl
 
-    def initialize(backend, conf, dependencies, require_loader, skip_only_if_eval)
+    def initialize(profile_context, resources_dsl, backend, conf, dependencies, require_loader, skip_only_if_eval)
+      @profile_context = profile_context
+      @resources_dsl = resources_dsl
       @backend = backend
       @conf = conf
       @dependencies = dependencies
@@ -50,22 +29,18 @@ module Inspec
       @skip_file_message = nil
       @skip_file = false
       @skip_only_if_eval = skip_only_if_eval
+
+      extend resources_dsl # TODO: remove? push to method_missing?
+    end
+
+    alias profile_context_owner profile_context
+
+    def profile_id
+      profile_context.profile_id
     end
 
     def to_s
       "Control Evaluation Context (#{profile_name})"
-    end
-
-    def profile_context_owner
-      self.class.profile_context_owner
-    end
-
-    def profile_id
-      self.class.profile_id
-    end
-
-    def resources_dsl
-      self.class.resources_dsl
     end
 
     def title(arg)
@@ -134,12 +109,12 @@ module Inspec
 
       unless profile_context_owner.profile_supports_platform?
         platform = inspec.platform
-        msg = "Profile `#{profile_context_owner.profile_id}` is not supported on platform #{platform.name}/#{platform.release}."
+        msg = "Profile `#{profile_id}` is not supported on platform #{platform.name}/#{platform.release}."
         ::Inspec::Rule.set_skip_rule(control, true, msg)
       end
 
       unless profile_context_owner.profile_supports_inspec_version?
-        msg = "Profile `#{profile_context_owner.profile_id}` is not supported on InSpec version (#{Inspec::VERSION})."
+        msg = "Profile `#{profile_id}` is not supported on InSpec version (#{Inspec::VERSION})."
         ::Inspec::Rule.set_skip_rule(control, true, msg)
       end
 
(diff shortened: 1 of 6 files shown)


miah:
This was a fun adventure!

miah (approved):

lib/inspec/rule.rb
```diff
@@ -13,32 +13,21 @@ module Inspec
...
-      @resource_dsl
-    end
-
     attr_reader :__waiver_data
-    def initialize(id, profile_id, opts, &block)
+    attr_accessor :resource_dsl
+    attr_reader :__profile_id
+    alias profile_id __profile_id
+
+    def initialize(id, profile_id, resource_dsl, opts, &block)
```
clintoncwolfe, line 20: Could we have a rdoc block here describing the class and its constructor options?

lib/inspec/rule.rb
```diff
@@ -13,32 +13,21 @@ module Inspec
...
-
-    def self.resource_dsl # rubocop:disable Style/TrivialAccessors
-      @resource_dsl
-    end
-
     attr_reader :__waiver_data
-    def initialize(id, profile_id, opts, &block)
+    attr_accessor :resource_dsl
+    attr_reader :__profile_id
+    alias profile_id __profile_id
```
clintoncwolfe: My understanding is that individual `describe` blocks are `instance_eval`'d against an instance of this. Historically, inspec internal methods were prefixed with `__` in an attempt to keep them separate from anything a inspec profile author user might use. Do we have a strong reason for breaking that pattern?

lib/inspec/control_eval_context.rb
```diff
@@ -12,236 +12,189 @@ module Inspec
...
 
     # Creates the heart of the control eval context:
     #
     # An instantiated object which has all resources registered to it
-    # and exposes them to the a test file.
+    # and exposes them to the test file.
     #
     # @param profile_context [Inspec::ProfileContext]
     # @param outer_dsl [OuterDSLClass]
     # @return [ProfileContextClass]
```
clintoncwolfe: At this very moment, we have a lot more tribal knowledge about what is going on in this file than we usually do. Could you expand/update the in-code comments here to be more helpful to future contributors?

In particular, how does it "register" resources? There is no class named "OuterDSLClass", and what is returns is not a "ProfileContextClass".

lib/inspec/control_eval_context.rb
```diff
@@ -12,236 +12,189 @@ module Inspec
...
+        define_method name.to_sym do |*args|
+          new_res.new(@backend, name.to_s, *args)
         end
+      end
+    end
 
-        define_method :skip_control do |id|
-          profile_context_owner.unregister_rule(id)
-        end
+    def add_resources(context)
```
clintoncwolfe, line 87: It would be great here if this `context` variable could be renamed to be more clear as to what type of context it is referring to.

lib/inspec/profile_context.rb
```diff
@@ -56,10 +56,14 @@ def to_resources_dsl
...
-                                  ctx.new(@backend, @conf, dependencies, @require_loader, @skip_only_if_eval)
-                                end
+      @control_eval_context ||=
+        Inspec::ControlEvalContext.new(self,
+                                       to_resources_dsl,
+                                       @backend,
+                                       @conf,
+                                       dependencies,
+                                       @require_loader,
+                                       @skip_only_if_eval)
```
clintoncwolfe, line 66: 💯 for this change

clintoncwolfe (changes requested):
I think this is great. I would like to see the in-code comments and rdoc updated and expanded - we have an opportunity right now while the team has a clear understanding of how this area of the project works, and capturing that understanding in a way that other contributors can see would be worthwhile.

My other concern is that I'd like to see this branch run against some of our important content - the compliance profiles and the cloud resource packs, for example. I understand that we don't have automated testing setup against those yet, I'm just requesting a one-off run to verify there isn't anything in their codebases that depends on peculiar behavior, like isolating resources per-profile.

zenspider:
> like isolating resources per-profile

I didn't change any behavior. I only changed where things are stored. AFAIK, we have tests to verify that per-profile stuff works and they still pass.

zenspider:
> I think this is great. I would like to see the in-code comments and rdoc updated and expanded - we have an opportunity right now while the team has a clear understanding of how this area of the project works, and capturing that understanding in a way that other contributors can see would be worthwhile.

I simply refactored the code while trying to get to a place where I could hook up let cleanly (which failed miserably, but that's an issue around the overall design). It doesn't mean I understand anything about what I refactored. You don't need to understand a formula to algebraically manipulate it.

Cleaning up this file, renaming variables, and (re)documenting it feels out of scope for a simple refactoring and will make the diff more confusing. I will file issues for the concerns you raised.

lib/inspec/rule.rb
zenspider: Refactoring from `Class.new` + `define_method` to `class` + `def` means moving from closures (eg local variables within scope) to something else. In this specific case, I moved from local variable `profile_id` to accessor `profile_id`. 

I can switch them to `__profile_id` but I was trying to keep the diff as small as possible.

zenspider: Refactoring from `Class.new` + `define_method` to `class` + `def` means moving from closures (eg local variables within scope) to something else. In this specific case, I moved from local variable `profile_id` to accessor `profile_id`. 

I can switch them to `__profile_id` but I was trying to keep the diff as small as possible.

lib/inspec/control_eval_context.rb
zenspider: Refactoring from `Class.new` + `define_method` to `class` + `def` means moving from closures (eg local variables within scope) to something else. In this specific case, I moved from local variable `profile_id` to accessor `profile_id`. 

I can switch them to `__profile_id` but I was trying to keep the diff as small as possible.

lib/inspec/control_eval_context.rb
```diff
@@ -12,236 +12,169 @@ module Inspec
...
-          end
+    def add_resources(context)
+      # # TODO: write real unit tests for this and then make this change:
+      # dsl = context.to_resources_dsl
+      # self.class.include dsl
+      # Inspec::Rule.include dsl
 
-          rule_class.class_eval do
-            include context.to_resources_dsl
-          end
```
clintoncwolfe, line 162: Changing 
```
rule_class.class_eval do
  include context.to_resources_dsl
end
```
to
```
Inspec::Rule.class_eval do
  include context.to_resources_dsl
end
```

does not appear to be a not a refactoring, but a data change. Previously each control would end up a with a private copy of the DSL; now that is shared class data on Inspec::Rule.

I am in favor of the change, and removing a layer here is a good thing. I don't know what content out there would try to leverage having a private copy of the resources; but I want to at least have the assurance that some of our most important content won't break because we made what appears to be an innocuous change. 

So, that's why I'm asking for cross-product testing - check for impact on content, just to do due diligence.

zenspider: I'm experimenting with changing it to `extend`... but I have doubts we have any tests that fail if I remove this functionality entirely.

zenspider:
Removing Davy... I've already had Stocksy poke at this with a real compliance run and it checked out.

https://github.com/chef/compliance-profiles/pull/1494

clintoncwolfe (approved):
With it green against the compliance profiles I'm a lot happier, but I wish that it was feasible to run against the cloud resource packs.

Issue #4519, opened 2019-09-26
InSpec should use vendor directory to resolve dependencies for an archived profile


james-stocks:
InSpec 4.7.3 shipped with ChefDK 4.2.0

# Describe the problem

We have a "resource pack" InSpec profile in a Github repo.
We then create compliance profiles that depend on the resource pack. We `vendor` these and ship the tar.gz as part of Chef Automate.
Airgapped customers are unable to create wrapper profiles to customise the controls that run, because they do not have access to the source for the resource pack - even though the compliance profile is vendored and the resource code is available in the `vendor` directory.

```
➜  vendor-test inspec vendor my_wrapper_profile
Traceback (most recent call last):
	27: from /usr/local/bin/inspec:343:in `<main>'
	26: from /usr/local/bin/inspec:343:in `load'
	25: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/inspec-bin-4.7.3/bin/inspec:11:in `<top (required)>'
	24: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.7.3/lib/inspec/base_cli.rb:31:in `start'
	23: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
	22: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	21: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
	20: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
	19: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.7.3/lib/inspec/cli.rb:171:in `vendor'
	18: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.7.3/lib/inspec/base_cli.rb:274:in `vendor_deps'
	17: from /opt/chefdk/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.7.3/lib/inspec/profile_vendor.rb:13:in `vendor!'
[...]

PR #4806, opened 2019-12-19
Use an archive if possible when fetching fails and --airgap is provided

### LINKED ISSUE #4727: Modify Fetcher API to fallback to available archive if fetch attempt fails
## Detailed Description

If a fetcher fails to fetch a profile (see #4726), and an archive is available, check the archive for the dependency and use the copy of the profile from the archive.

## Context

See #4519 for extensive discussion.

## Possible implementation

It is hoped that this can be handled at a high level, not a per-fetcher basis.

clintoncwolfe:
## Description

Adds a new mode to the `inspec archive` command. Under the following conditions:

 * When the `--airgap` option is provided
 * and the top-level profile has one or more local dependencies on an archive (a zip file or tarball)
 * and fetching any other dependency in the tree fails for any of the local, git, or URL fetchers,

Instead of failing the operation, use the copy of the unreachable profile that is contained in the archive.

### Implementation Notes

This works by fooling the cache in a couple of ways. During an `archive` operation, the vendor cache is normally purged. However, if `--airgap` is provided, the cache is warmed by locating and extracting any archives that are at the top-level of the profile tree.

Next, when profiles fail to be fetched, it is usually because their cache key is dependent on information only available on the network. However, that same information is available in the lockfile of the archive. So, find the lockfile, find the cache key, and update the fetcher to make it believe that it already knows its cache key (which we know will be a "hit").

### Further Work

This PR currently provides implementations for the `local`, `url`, and `git` fetchers.  Currently no support is provided for the `supermarket` ~~or `compliance`~~ fetchers. UPDATE: the compliance fetcher is also included.

It's also unclear if `--airgap` is the right word for this. One idea is to use that same option on any command that is able to operate offline.  Is `--offline` a better word?

## Related Issue

Closes #4727

Refs #4519 

## Types of changes

## Checklist:


lib/inspec/fetcher/git.rb
```diff
@@ -112,6 +112,19 @@ def resolved_source
...
+        branch
+        tag
+        ref
+      }.each do |opt_name|
+        opt_changed = update_ivar_from_opt(opt_name, opts)
+        changed ||= opt_changed
+      end
+      changed
+    end
+
```
miah, line 118:
```ruby
%i{branch tag ref}.map { |opt_name| update_ivar_from_opt(opt_name, opts) }.any?
```

lib/inspec/fetcher/local.rb
```diff
@@ -100,6 +100,19 @@ def sha256
...
+      changed = false
+
+      old_val = @archive_shasum
+      new_val = opts[:sha256]
+      unless old_val == new_val
+        @archive_shasum = new_val
+        changed = true
+      end
+
+      changed
```
miah:
```ruby
changed = @archive_shasum != opts[:sha256]
@archive_shasum = opts[:sha256]
changed
```

lib/inspec/fetcher/url.rb
```diff
@@ -118,6 +118,19 @@ def to_s
...
+      changed = false
+
+      old_val = @archive_shasum
+      new_val = opts[:sha256]
+      unless old_val == new_val
+        @archive_shasum = new_val
+        changed = true
+      end
+
+      changed
```
miah: Identical code fetcher/local.rb#update_from_opts. Push up to `Inspec::Plugins::Fetcher`

lib/inspec/plugin/v1/plugin_types/fetcher.rb
```diff
@@ -65,6 +65,28 @@ def cache_key
...
+
+      # Helper for above.
+      def update_ivar_from_opt(opt_name, opts)
+        ivar_sym = "@#{opt_name}".to_sym
+        old_val = instance_variable_get(ivar_sym)
+        new_val = opts[opt_name]
+        return false if old_val == new_val
+
+        instance_variable_set(ivar_sym, new_val)
+        true
```
miah: Same pattern as above.

lib/inspec/profile_vendor.rb
```diff
@@ -51,8 +51,10 @@ def profile_opts
       }
     end
 
-    def vendor_dependencies
+    def vendor_dependencies(opts)
+      # This deletes any existing vendor/ directory
       delete_vendored_data
```
miah, line 56: Should this be a bang! method?

lib/inspec/profile_vendor.rb
```diff
@@ -80,5 +82,40 @@ def extract_archives
...
+      profile.metadata.dependencies.each do |dep_info|
+        next unless dep_info.key?(:path) # "Local"-type fetchers are identified by the :path key
+
+        # Locate local dep
+        dep_profile_path = File.expand_path(dep_info[:path], profile_cwd)
+        raise(Inspec::FetcherFailure, "Profile dependency local path '#{dep_profile_path}' does not exist") unless File.exist?(dep_profile_path)
+
+        # Determine if it is an archive
+        file_provider = FileProvider.for_path(dep_profile_path)
+        next unless file_provider.is_a?(ZipProvider) || file_provider.is_a?(TarProvider)
```
miah, line 105: If the fetcher system is plugable, then we have to update this list dynamically.

lib/inspec/profile_vendor.rb
```diff
@@ -80,5 +82,40 @@ def extract_archives
...
+        # Place the local dependency in the vendor cache.
+        # Fetchers are in charge of calculating cache keys.
+        fetcher = Inspec::Fetcher::Local.resolve(dep_profile_path)
+        dep_profile_cache_dir = "#{cache_path}/#{fetcher.cache_key}"
+        file_provider.extract(dep_profile_cache_dir)
+
+        # The archive (probably) contained a vendor cache of its own.
+        # We must flatten that, so that all vendor cache entries are top-level.
+        Dir["#{dep_profile_cache_dir}/vendor/*"].each do |sub_dep_cache_dir|
+          FileUtils.mv(sub_dep_cache_dir, cache_path)
```
miah, line 120: Collision if two dependencies have the same name and get moved to cache_path

test/functional/fetchers_test.rb
```diff
@@ -1,16 +1,16 @@
...
 
   let(:looks_like_a_stacktrace) { %r{lib/inspec/.+\.rb:\d+:in} }
   let(:invocation) { "exec #{path} --no-create-lockfile" }
   let(:run_result) { inspec(invocation) }
+  let(:fetcher_profiles) { "#{profile_path}/fetcher-failures" }
 
   # Refs #4726
   describe "when fetchers fetch a bad dependency" do
-    let(:fetcher_profiles) { "#{profile_path}/fetcher-failures" }
+    parallelize_me!
```
miah: We can't parallelize this whole suite?

test/functional/fetchers_test.rb
```diff
@@ -54,4 +54,52 @@ def assert_fetcher_failed_cleanly(run_result, error_regex, profile_location)
...
   end
+
+  # Refs #4727
+  describe "when a archive is available of an unfetchable profile with --airgap" do
+
+    let(:invocation) { "archive #{fetcher_profiles}/#{profile_name} --airgap" }
+
+    def teardown
+      FileUtils.rm_rf "#{fetcher_profiles}/#{profile_name}/vendor"
+      FileUtils.rm_f "#{profile_name}-0.1.0.tar.gz"
```
miah: *scary* we need to be really careful with rm_rf and rm_f especially since we're not using temp directories anywhere here (we should though).

miah (changes requested):

lib/inspec/fetcher/url.rb
```diff
@@ -127,8 +140,10 @@ def parse_uri(target)
     end
 
     def sha256
+      return @archive_shasum if @archive_shasum
+
       file = @archive_path || temp_archive_path
-      OpenSSL::Digest::SHA256.digest(File.read(file)).unpack("H*")[0]
+      @archive_shasum = OpenSSL::Digest::SHA256.digest(File.read(file)).unpack("H*")[0]
```
zenspider: This is cleaner (and gets us closer to `-w`) if it is in `||=`-style.

lib/inspec/profile_vendor.rb
zenspider: Agreed. Typechecks are a smell.

Both `ZipProvider` and `TarProvider` subclass from `FileProvider`... If they moved to subclassing `ArchiveProvider` (which subclassed `FileProvider`) then only one check needs to be made here and it is more resilient to change.

test/functional/fetchers_test.rb
```diff
@@ -54,4 +54,52 @@ def assert_fetcher_failed_cleanly(run_result, error_regex, profile_location)
...
+    end
+
+    def assert_archive_worked(run_result)
+      _(run_result.stderr).must_be_empty
+      _(run_result.stdout).must_include "Finished archive generation"
+      assert_exit_code(0, run_result)
+    end
```
zenspider, line 73: Please use `prepare_examples` or the `tmpdir` option in `run_inspec_process`.

zenspider (changes requested):

lib/inspec/fetcher/git.rb
clintoncwolfe: I don't trade conciseness for brevity.

clintoncwolfe: At first I grumbled (another @#$!% style opinion) but then I scrolled over and saw the `.any?` at the end and quite liked it. Thanks!

lib/inspec/fetcher/local.rb
clintoncwolfe: I think there are a lot of ways to write the same thing, and I will always prefer the longer, dumber, clearer way.

lib/inspec/fetcher/url.rb
clintoncwolfe: I hate to make "do something ugly and illogical" (having the opts not match the ivar name) the superclass implementation. To me, that obliges subclasses to override, which is not what you generally want.

lib/inspec/profile_vendor.rb
clintoncwolfe: Sure

lib/inspec/profile_vendor.rb
clintoncwolfe: This is the `FileProvider` system, not the `Fetcher`, as I understand it, and is used in other places to detect whether a file is an archive - in fact, just a few lines above at https://github.com/inspec/inspec/blob/8b5789004ac31f4b6e7210421b119af41f483e3e/lib/inspec/profile_vendor.rb#L74 . Good idea or not, "how do I detect an archive in InSpec" has at least one existing answer - and it is based on typechecking. Sounds like you'd like a  different approach, and you're like me to alter an unrelated class hierarchy to avoid a code smell.

Respectfully, I feel that is outside the scope of this PR. I'll open an issue. #4824

lib/inspec/profile_vendor.rb
clintoncwolfe: That's just going to be a limitation. We'll need to document that.

test/functional/fetchers_test.rb
clintoncwolfe: I believe we can, yes. That was a holdover from early work.

lib/inspec/fetcher/local.rb
zenspider: I might change the suggested code to the following:

```ruby
original = @archive_shasum
@archive_shasum = opts[:sha256]
original != opts[:sha256]
```

The only difference between those two versions is whether you want to express the intent of `changed` vs `original`... they're similar enough to me that I don't need to think about either of them for more than half a second.

Your original code is longer and "dumber", yes, but not clearer in any sense. It took me probably 20 seconds to fully comprehend after getting through these:

* The true/false literals are a smell.
* The `var = x; do stuff; var` pattern is a smell.
* The conditional requires thinking about.
* There's no reason NOT to assign if they're the same, so why the conditional?

lib/inspec/fetcher/git.rb
zenspider: Maybe you could not bias towards this attitude to begin with?

The code in `update_from_opts` is not clear:

* The false literal is a smell.
* The `var = x; collection.each { ... }; var` pattern is a smell and almost always a sign that Enumerable isn't being used to our advantage.
* The logical `||=` requires thinking about.

Whereas `collection.map { ... }.any?` expressly conveys what this method is trying to accomplish.

lib/inspec/fetcher/url.rb
zenspider: This doesn't make sense to me. You added the same named method to the superclass that returns `false` and thus "obliges subclasses to override". Can you clarify?

Having identical code in two sibling classes and pushing them up to the abstract superclass IS a very normal refactoring. This pushback doesn't make sense to me.

lib/inspec/fetcher/git.rb
clintoncwolfe: I made the change - no need to bully.

lib/inspec/fetcher/local.rb
clintoncwolfe: We should talk offline about subjective review.

lib/inspec/fetcher/url.rb
clintoncwolfe: First, thanks - you're right, I forgot I did provide a default superclass implementation that does no update and returns `false`. So any any subclass would be obliged to override if they wanted any behavior that wasn't that; but at least is it is a **sensible** default.

I'll cave on this one - I'll replace the sensible superclass implementation with the weird one from url/local. Any new fetcher would need to remember to override that with a sane updater, because it's not likely they'll happen to choose the same ivar name that local and url did. 

That trades a small x2 DRY problem for reduced clarity and increased extensibility cost.

test/functional/fetchers_test.rb
clintoncwolfe: Agreed

test/functional/fetchers_test.rb
clintoncwolfe: Agreed, I'll aim to rework the tests to use tmpdir option to `run_inspec_process`.

lib/inspec/cli.rb
```diff
@@ -173,6 +173,8 @@ def vendor(path = nil)
     desc: "Generates a tar.gz archive."
   option :overwrite, type: :boolean, default: false,
     desc: "Overwrite existing archive."
+  option :airgap, type: :string, default: false, banner: "", # Actually a boolean, but this suppresses the creation of a --no-airgap...
+    desc: "Prefer using existing local archives to fetching profiles."
```
james-stocks: Reading the PR description, it sounds like this option allows a fallback to local files when fetching fails; rather than preferring local files? So maybe `"Allow InSpec to use local archives when fetching dependencies"` ?

The term "local archives" could maybe be more explicitly written as "existing content in the vendor subdirectory"?

lib/inspec/dependencies/resolver.rb
```diff
@@ -82,5 +92,41 @@ def resolve(deps, top_level = true, seen_items = {}, path_string = "") # rubocop
       Inspec::Log.debug("Dependency traversal complete.") if top_level
       graph
     end
+
+    def fallback_to_archive_on_fetch_failure(dep)
```
james-stocks, line 96: Method name like `can_fallback_to_archive?` might be more expressive for what this method does

james-stocks: My suggestion there is wrong because this method makes some changes to the current state, it doesn't just return a boolean.

test/fixtures/profiles/fetcher-failures/local-dep-on-bad-auto-archive/inspec.yml
```diff
@@ -0,0 +1,8 @@
+name: local-dep-on-bad-auto-archive
+title: An inspec profile that has a local dep on an archive that has unresolvable deps
+version: 0.1.0
+supports:
+  platform: os
+depends:
+  - name: auto-dep-on-missing
+    path: ../auto-dep-on-missing-0.1.0.tar.gz
```
james-stocks, line 8: 👍 this appears to match the customer case - the tar.gz is available at the given path, but the dependency defined _inside_ that  .tar.gz is unreachable. Awesom!

lib/inspec/dependencies/resolver.rb
james-stocks: I would disagree with "very narrow use case" because it's the main use case for enterprise-scale customers (wrapper profile in an air gapped environment).
Maybe delete the current description paragraph and give 1-2 sentences that state what the method does and what the return condition is.

lib/inspec/plugin/v1/plugin_types/fetcher.rb
clintoncwolfe: Thanks, I've put that in.

lib/inspec/dependencies/resolver.rb
clintoncwolfe: Hrm, as written, that does read as editorializing. I'll revise. I don't mean that it is uncommon or unimportant, just that it is a very specific situation - a very special set of conditions have to be met in order for this feature to be able to work.

lib/inspec/cli.rb
clintoncwolfe: The behavior is to attempt to fetch first, then fallback (ie, reprogram the fetcher to point at the archive) and re-fetch. If the second fetch fails, give up. 

`Fallback to use local archives if fetching fails`?

Or should there be a `--airgap no-fetch` (or similar) value indicating no first attempt should be made?

james-stocks: I was agreeing with the functionality and just critiquing the `desc:` string. It uses the word "prefer" which implies it will fetch local first; which doesn't reflect the functionality.
I'm just suggesting changing this string.

At the same time, it would be awesome to add an option to indicate that InSpec should immediately use the local content first, and not attempt a remote fetch.

e.g.

```
--local-dependencies never
--local-dependencies allow
--local-dependencies always
```

clintoncwolfe: OK. Let's leave the functionality alone for now then, so we can see how customers like the feature as-is. Let's revise the usage string to clarify that it is a fallback operation.

lib/inspec/plugin/v1/plugin_types/fetcher.rb
```diff
@@ -65,6 +65,33 @@ def cache_key
         raise "Fetcher #{self} does not implement `cache_key()`. This is required for terminal fetchers."
       end
 
+      #
+      # This optional method may be used after a failed fetch. If the fetcher
+      # can be updated with information that might lead to a successful
+      # retrieval of alternative content, this method may be called.
+      #
+      #  Default implementation makes a peculiar assumption that the class has
```
miah, line 73: nitpick; extra space =)

miah (approved):
Thanks!

lib/inspec/profile_vendor.rb
```diff
@@ -107,14 +107,20 @@ def warm_vendor_cache_from_archives
         # Place the local dependency in the vendor cache.
         # Fetchers are in charge of calculating cache keys.
         fetcher = Inspec::Fetcher::Local.resolve(dep_profile_path)
-        dep_profile_cache_dir = "#{cache_path}/#{fetcher.cache_key}"
-        file_provider.extract(dep_profile_cache_dir)
+        # Use a shorter name here in hopes of dodging windows filesystems path length restrictions
+        actual_dep_profile_cache_dir = "#{cache_path}/#{fetcher.cache_key}"
+        short_dep_profile_cache_dir = "#{cache_path}/#{SecureRandom.hex[0..6]}"
```
zenspider: This doesn't have to be a random name, does it? Literally `"#{cache_path}/short"` would work. If the concern is collision, then mktmpdir might be a better bet.

lib/inspec/profile_vendor.rb
zenspider: doco doco? or comment here?

zenspider (approved):

PR #4813, opened 2019-12-30
DO NOT MERGE - Remove all deprecated aws and azure resources.

zenspider:
These have been available in inspec-aws and inspec-azure for about a
year now.


clintoncwolfe (changes requested):
This MUST be held until a major version bump, and we haven't discussed the timing or branching strategy for that, AFAIK. There are definitely people in the wild still using the core AWS resources.

Additionally, even if "removed", we'd need to leave behind deprecation stubs for one major version - so if you tried to use a resource that is now gone, you'd get a deprecation notice and further instructions, not an opaque error.

zenspider:
I don't think it needs to be held at all. We can release these resources as gems and add dependencies to those gems. Once we've made it through a major rev, final removal would involve removing the dependencies for them. This should be doable independent of any resource pack rework we want to do (and if anything should help that effort along).

miah:
> Additionally, even if "removed", we'd need to leave behind deprecation stubs for one major version - so if you tried to use a resource that is now gone, you'd get a deprecation notice and further instructions, not an opaque error.

https://github.com/inspec/inspec/blob/master/lib/resource_support/aws/aws_resource_mixin.rb#L3
https://github.com/inspec/inspec/blob/master/lib/resources/azure/azure_generic_resource.rb#L17

https://github.com/inspec/inspec/blob/master/etc/deprecations.json#L19-L22
https://github.com/inspec/inspec/blob/master/etc/deprecations.json#L75-L77

#3822 

I think we're all set on deprecation's, did we miss something?

miah (approved):

Issue #4811, opened 2019-12-28
lib/inspec/plugin/v1/plugin_types/resource.rb should not exist


zenspider:
We already have `Inspec::Resource`... there is no reason for this extra level of complexity.

zenspider:
Fixed in #4814.

kekaichinose:
Fixed in #4814

PR #4814, opened 2019-12-30
Removed ~50% of metaprogramming from LibraryEvalContext & Resource

zenspider:
This should probably be reviewed by commit.

A majority of that file count is BS as I nuked AWS tests just to get their noise out of the way. I can rebase that out and get them passing if we actually care. I also put up a PR to remove AWS + azure for funsies.


clintoncwolfe (dismissed):
Can we get a clean diff on this, without the unrelated AWS/Azure test deletions?

zenspider:
Fixes #4811.

miah (approved):
I paired with @zenspider for like 98% of this. :+1:

test/unit/profiles/profile_resource_exceptions_test.rb
```diff
@@ -6,107 +6,68 @@
   let(:profile) do
     profile = MockLoader.load_profile("profile-with-resource-exceptions")
     profile.load_libraries
-    profile.collect_tests
+    profile.collect_tests # TODO: bad name?
```
clintoncwolfe, line 9: Agreed, I've always thought that was a bad name. In our online docs, we have profile -> control -> describe block -> test (`its` block). I *think* here `collect_tests` is actually referring to `describe_blocks`/ExampleGroups.

lib/inspec/resources/platform.rb
```diff
@@ -71,31 +71,26 @@ def params
...
+            in_family?(v)
+          when :os, :platform then
+            platform?(v)
+          when :os_name, :platform_name then
+            name == v
+          when :release then
+            check_release(v)
+          end
+        }
+      } || supports.empty?
```
clintoncwolfe, line 89: The string/symbol normalization is both regrettable and an improvement :-)

But for the logic, I think this loses a lot of clarity - keep in mind other people who are not you may need to read and understand this. Imagine a junior engineer reading the original logic versus this. What is that dangling `||supports.empty?` doing? Easily missed. 

The original is much more straightforward and less clever.  I think we want code that is easy to maintain by a broad range of engineers.

lib/inspec/dsl_shared.rb
```diff
@@ -8,6 +8,26 @@ module RequireOverride
...
+      #     that has a require to a known file that is NOT loaded yet
+      #
+      # ProfileContext#initialize
+      #   -> library_eval_context
+      #
+      # ProfileContext#load_libraries autoload
+      #   -> load_library_file(@library_eval_context)
+      #
+      # probably most of this comment is useless, but it was hard to
+      # discover so I'm adding it for others.
```
clintoncwolfe, line 29: Thank you for this.

test/docker_test.rb
```diff
@@ -1,55 +0,0 @@
-require_relative "docker_run"
```
clintoncwolfe, line 1: We do still create a docker image during release - and some people are consuming it, apparently (we had a recent issue that mentioned it indirectly).

clintoncwolfe (commented):
I think this is wonderful - I especially appreciate all of the code comments reflecting what was learned along the way.

I don't have any requested changes - the only thing holding me back from an approval is that since this makes changes to the Resource base class as well as some of the logic behind the platform support, it would be advisable to do another run against the compliance-profile repo (this branch against their master).

clintoncwolfe (commented):
Hrm, see prior comment review.

lib/inspec/control_eval_context.rb
```diff
@@ -12,236 +12,168 @@ module Inspec
...
-          end
-        end
+      self.class.class_eval do
+        include context.to_resources_dsl
+      end
 
-        define_method :add_subcontext do |context|
-          profile_context_owner.add_subcontext(context)
-        end
+      # TODO: seriously consider getting rid of the NPM model
```
james-stocks: NPM?

james-stocks:
Ran this branch against our kitchen tests for compliance profiles, and it looks good!

james-stocks:
...also ran it against inspec-scap and it looks good with that too

zenspider:
## Complexity of all `lib/inspec/*context*.rb` files

### Before:

```
   527.9: flog total
    16.5: flog/method average

   236.4: Inspec::ControlEvalContext::create lib/inspec/control_eval_context.rb:88-242
    39.9: Inspec::ControlEvalContext::rule_context lib/inspec/control_eval_context.rb:20-73
    27.0: Inspec::ProfileContext#load_with_context lib/inspec/profile_context.rb:153-161
    25.8: Inspec::LibraryEvalContext::create lib/inspec/library_eval_context.rb:17-52
    23.8: Inspec::ProfileContext#load_libraries lib/inspec/profile_context.rb:118-139
```

### After

```
   412.4: flog total
     8.1: flog/method average

    27.0: Inspec::ProfileContext#load_with_context lib/inspec/profile_context.rb:162-170
    25.4: Inspec::ProfileContext::DomainSpecificLunacy::ClassMethods#add_registry_methods lib/inspec/profile_context.rb:246-256
    23.8: Inspec::ProfileContext#load_libraries lib/inspec/profile_context.rb:123-148
    22.5: Inspec::ControlEvalContext#register_control lib/inspec/control_eval_context.rb:105-121
    19.4: Inspec::ControlEvalContext#only_if lib/inspec/control_eval_context.rb:152-166
```

## Code Duplication

### Before:

```
Total score (lower is better) = 232

1) IDENTICAL code found in :iter (mass*2 = 156)
  lib/inspec/control_eval_context.rb:26
  lib/inspec/control_eval_context.rb:188

2) IDENTICAL code found in :iter (mass*2 = 76)
  lib/inspec/control_eval_context.rb:44
  lib/inspec/control_eval_context.rb:206
```

### After:

```
Total score (lower is better) = 0
```

test/docker_test.rb
zenspider: Sure. But we don't run any of these `test/docker*` files anywhere that I can find.

lib/inspec/resources/platform.rb
zenspider: > The original is much more straightforward and less clever

I wholeheartedly disagree. Using enumerable methods spells out intent much more clearly than managing logic via local variables w/ `next`, `break` and the like. As a result using higher level intention describing methods is more junior developer friendly than the original version. Objectively, the new code is 83% the complexity of the original (17.5/20.9).

As far as the hanging `empty?` clause, it is lamentable, but always I prefer to put the positive/main cases before the negative/edge cases whenever possible. I could try to format it slightly differently but I prefer this structure.

lib/inspec/resources/platform.rb
```diff
@@ -51,46 +56,41 @@ def in_family?(family)
...
-              check_release(v)
-            else
-              false
-            end
-
-          break if status == false
-        end
-        return true if status == true
-      end
+      raise ArgumentError, "`supports` is nil." unless supports
```
stevendanna, line 74: I'm mildly worried raising here rather than returning true will break someone.

stevendanna (commented):
Your perseverance on this is admirable.  I reviewed this simply because I try to keep a vague sense of how this code base works. A few general comments:

- There are a bunch of small cleanups here that are completely uncontroversial. I'd say if we can't merge this PR soon, it is worth spending a few minutes, pulling those out and getting them merged.

- I'll be honest, this code base is in a state where it is hard for me to judge whether this change to the metaprogramming is making things better or worse.  My guess is that I had just grown familiar to the old structure and that since it is still VERY complex the new structure means I lose the benefit of familiarity.  I trust y'all to do whatever seems like the right thing.

- I think the best thing we could do for ourselves is to write a big piece of documentation somewhere that covers: (1) The various evaluation contexts, what they are used for, and the key points of how they are constructed and (2) the invariants that we _think_ the current system provides with respect to resource registry "isolation"

clintoncwolfe (approved):
Thanks for all your work on this!

PR #4822, opened 2020-01-01
make command.exist? POSIX conform

PiQuer:
With the current implementation, on a Linux system without bash
command.exist? always returns false. 'sh -c' is guaranteed to exist on a
POSIX-conform system [1] whereas 'bash -c' only works if bash is
actually installed.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html

## Types of changes

## Checklist:


zenspider (approved):
Thanks!

miah (approved):

zenspider (changes requested):
This fails our tests hard because of the way we "mock" out our commands. Can you please make a corresponding change to `test/helpers/mock_loader.rb` so the tests go green?

PiQuer:
Yes, will try to get to it next weekend. Cheers!

PiQuer:
Tests are green now.

clintoncwolfe (approved):
Thanks!

zenspider (approved):

PR #4827, opened 2020-01-07
Move parsers under Utils

miah:
I'm more than happy to move these under a different namespace, but wanted to get these out of the top-level.

### DIFF
--- a/lib/inspec/resources/aide_conf.rb
+++ b/lib/inspec/resources/aide_conf.rb
@@ -24,7 +24,7 @@ module Inspec::Resources
 
     attr_reader :params
 
-    include CommentParser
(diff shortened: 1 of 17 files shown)


clintoncwolfe (approved):
Thanks!

zenspider (approved):

stevendanna (approved):
If you are opposed to Utils on aesthetic grounds, may I suggest Inspec::Parsers?

miah:
> If you are opposed to Utils on aesthetic grounds, may I suggest Inspec::Parsers?

Yes, then I could drop that word from all those Constants! I'll do this on the next revision.

PR #4833, opened 2020-01-13
Fixed duplicate groups on osx by merging members by gid.

zenspider:


clintoncwolfe (approved):
Thanks - I like the use of group_by in this.

miah (changes requested):
We should have some tests for this! =)

zenspider:
We have tests for this... and they're breaking for me... so they're not really mocked properly.

PR #4840, opened 2020-01-17
Start forming resource-core

miah:


kitchen.vagrant.yml
```diff
@@ -27,14 +29,31 @@ platforms:
...
+      attributes:
+        audit_attribute: 'Attribute Override!'
+      insecure: true
+      reporter: ['json-file','chef-automate']
+      fetcher: 'chef-automate'
+      json_file:
+        location: /tmp/json_export.json
+      profiles:
+        - name: integration
+          url: https://github.com/inspec/inspec-integration-profile/archive/master.zip
```
clintoncwolfe, line 52: This section looks a lot like the Kitchen file for the audit cookbook, which is complicated in order to exercise lots of its features. I hope this can be simplified in the future.

clintoncwolfe (approved):
Looks like a great start! If you want any help simplifying the kitchen suite, I can do that - I'm pretty familiar with it.

kitchen.vagrant.yml
miah: Yes! I kinda lifted our old config a bit just to get kitchen.vagrant working again for debugging train#458 with @zenspider. I have another update I'm going to push that drops the audit bits, but i'll do that in another PR along with a few actual 'resource-core' tests.

zenspider (approved):
🤷‍♀

Sure?

PR #4845, opened 2020-01-22
Add: certificate content x509_certificate

frezbo:
Support passing in certificate content to `x509_certificate` resource.

## Description
This change allows to read a certificate chain file having multiple certificates. The inspec `file` resource is used to read the combined pem, then split on the delimiter and pass the content to `x509_certificate` resource.

## Types of changes

## Checklist:
- [x] DCO signed

### DIFF
--- a/docs/resources/x509_certificate.md.erb
+++ b/docs/resources/x509_certificate.md.erb
@@ -33,6 +33,22 @@ An `x509_certificate` resource block declares a certificate `key file` to be tes
       its('validity_in_days') { should be > 30 }
     end
 
+The `filepath` property can also be used.
+
+    describe x509_certificate(filepath: 'mycertificate.pem') do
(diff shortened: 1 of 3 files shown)


frezbo:
An example usage of this addition:

```ruby
delimiter = "\n-----END CERTIFICATE-----\n"
file('/Users/ngeorgi/Downloads/rds-combined-ca-bundle.pem').content.split(delimiter).each do |blob|
  describe x509_certificate(content: blob+delimiter) do
    it { should be_valid }
  end
end
```

Amazon WebServices provides a bundle pem file that has multiple certificates, as a user I would like to validate the expiry of each certificate in the bundle during CI integrations, and have owners notfied before the certificate expires.

miah (approved):

clintoncwolfe (approved):
Very useful, thanks!

Issue #4810, opened 2019-12-28
Inspec Habitat Bin Wrapper Breaks custom SSL and cacerts verification


drakoswraith:
On Linux, the habitat bin wrapper script for inspec breaks the ability to use custom SSL cert verification by requiring the core/cacerts package, and forcing the value of the SSL_CERT_FILE variable. 

This breaks the ability to use "inspec archive" or to build effortless inspec packages where custom CA certs are required.

# Describe the problem
[In habitat/plan.sh, line 70](https://github.com/inspec/inspec/blob/dc278348b6a26304f1f3d3cdd7cc7b62deda4563/habitat/plan.sh#L70):
export SSL_CERT_FILE=$(pkg_path_for cacerts)/ssl/cert.pem

This line prevents custom cacerts packages from being used by not honoring any existing SSL_CERT_FILE values already set and forcing use of the core/cacerts package that chef/inspec depends on.

## Possible Solution
Check whether SSL_CERT_FILE already has a value, and only set if not set. 
Or
Move the value to a default.toml value so that a user config can override what file to use

zenspider:
There is probably a more elegant way to do this in bash (bash? there's no shebang):

```diff
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -67,7 +67,7 @@ wrap_inspec_bin() {
   build_line "Adding wrapper $bin to $real_bin"
   cat <<EOF > "$bin"
 #!$(pkg_path_for core/bash)/bin/bash
-export SSL_CERT_FILE=$(pkg_path_for cacerts)/ssl/cert.pem
+export SSL_CERT_FILE="${SSL_CERT_FILE:-$(pkg_path_for cacerts)/ssl/cert.pem}"
 set -e
```

TheLunaticScripter:
I don't think that will work I think this line needs to be removed from the plan so that users can specify the `SSL_CERT_FILE` independently.

drakoswraith:
Confirmed

PR #4848, opened 2020-01-23
Remove the references to the core/cacerts so customers can specify it

TheLunaticScripter:

Currently the InSpec habitat packages are built with a dependency on the `core/cacerts` package. This sets the environment variable `SSL_CERT_DIR` to a specific version of the core/cacerts package. If a user wants to use inspec to communicate, via ssl, to an Automate, Github, or other server that does not have a cert chain that is in the `core/cacerts` package they will get an ssl error. The workaround for this is to then break immutability of the `core/cacerts` package by in the users consuming plan adding their custom certs to the specific version of `core/cacerts` that inspec is packaged with. Removing this will allow the user to set a custom cacerts package and set the `SSL_CERT_FILE` and `SSL_CERT_DIR` in their consuming plan. An example of this can be found in the [habitat scaffolding for InSpec](https://github.com/chef/effortless/blob/master/scaffolding-chef-inspec/lib/scaffolding.sh#L107).

## Related Issue
#4810 

## Types of changes

## Checklist:

### DIFF
--- a/habitat/plan.ps1
+++ b/habitat/plan.ps1
@@ -10,7 +10,6 @@ $pkg_maintainer="The Chef Maintainers <humans@chef.io>"
 $pkg_license=('Apache-2.0')
 
 $pkg_deps=@(
-  "core/cacerts"
   "chef/ruby-plus-devkit"
 )
 $pkg_bin_dirs=@("bin"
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
Thanks!

miah (approved):
Thanks @TheLunaticScripter !

PR #4852, opened 2020-01-24
Missing end quote in oracledb_session

### LINKED ISSUE #4847: oracledb_session.rb missing closing " from line 83
# Describe the problem
Line 83 of file "inspec/lib/inspec/resources/oracledb_session.rb" is missing the closing " and therefore not working. It is also not throwing and errors or exceptions.

## Possible Solution
Change line to include closing double-quote:

```ruby
%{su - #{@su_user} -c "env ORACLE_SID=#{@service} #{bin} / as #{@db_role}#{sql_postfix}"}
```

miah:
Fixes https://github.com/inspec/inspec/issues/4847

### DIFF
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -86,7 +86,7 @@ module Inspec::Resources
       elsif @su_user.nil?
         %{#{sql_prefix}#{bin} "#{user}"/"#{password}"@#{host}:#{port}/#{@service} as #{@db_role}#{sql_postfix}}
       else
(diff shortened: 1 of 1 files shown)


zenspider (approved):
This bug fix should go in ASAP, obviously... but it also should have been caught, so this needs to be followed up with better unit tests that cover the 3 edge cases for this command.

Also, I think this code could be cleaned up but I don't want to try to do that until we actually have unit tests in place.

clintoncwolfe (approved):
Thanks!

PR #4854, opened 2020-01-24
Rework and unify gemspecs.

zenspider:
First step is to get them agreeing on dependencies.

There's still some oddities between train vs train-core and
dependencies for AWS and the like, but this brings them in line so
diffing between them is MUCH better.

Next is to gut inspec and push everything over to inspec-core.

### DIFF
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -15,38 +15,14 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = "~> 2.4"
 
-  # the gemfile and gemspec are necessary for appbundler so don't remove it
-  spec.files = %w{Gemfile inspec.gemspec README.md LICENSE} + Dir.glob(
-    "{bin,lib,etc}/**/*", File::FNM_DOTMATCH
-  ).reject { |f| File.directory?(f) }
+  # ONLY the aws/azure/gcp files. The rest will come in from inspec-core
+  # the gemspec is necessary for appbundler so don't remove it
+  spec.files =
+    Dir.glob("{{lib,etc}/**/*,inspec.gemspec}")
+      .grep(/aws|azure|gcp|gemspec/)
+      .reject { |f| File.directory?(f) }
 
-  spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
-    .reject { |f| File.directory?(f) || f =~ %r{lib/plugins/.*/test/} }
-
-  # Implementation dependencies
-  spec.add_dependency "chef-telemetry", "~> 1.0"
-  spec.add_dependency "license-acceptance", ">= 0.2.13", "< 2.0"
-  spec.add_dependency "thor", ">= 0.20", "< 2.0"
-  spec.add_dependency "json-schema", "~> 2.8"
-  spec.add_dependency "method_source", "~> 0.8"
-  spec.add_dependency "rubyzip", "~> 1.2", ">= 1.2.2"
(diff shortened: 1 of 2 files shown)


inspec.gemspec
```diff
@@ -11,48 +11,27 @@ Gem::Specification.new do |spec|
...
-  spec.require_paths = ["lib"]
+  # ONLY the aws/azure/gcp files. The rest will come in from inspec-core
+  # the gemspec is necessary for appbundler so don't remove it
+  spec.files =
+    Dir.glob("{{lib,etc}/**/*,inspec.gemspec}")
+      .grep(/aws|azure|gcp|gemspec/)
+      .reject { |f| File.directory?(f) }
 
-  spec.required_ruby_version = ">= 2.4"
+  spec.add_dependency "inspec-core", "= #{Inspec::VERSION}"
```
tas50, line 25: Doesn't this mean we always ship 2 gems?

tas50: Any reason not to do something like this? https://github.com/chef/chef/blob/master/chef-universal-mingw32.gemspec#L1

inspec.gemspec
```diff
@@ -15,38 +15,15 @@ Gem::Specification.new do |spec|
...
-  spec.add_dependency "tty-prompt", "~> 0.17"
-  spec.add_dependency "tomlrb", "~> 1.2"
-  spec.add_dependency "addressable", "~> 2.4"
-  spec.add_dependency "parslet", "~> 1.5"
-  spec.add_dependency "semverse"
-  spec.add_dependency "htmlentities"
-  spec.add_dependency "multipart-post"
-  spec.add_dependency "term-ansicolor"
+  spec.add_dependency "inspec-core", "= #{Inspec::VERSION}"
+  spec.add_dependency "inspec-bin",  "= #{Inspec::VERSION}"
```
clintoncwolfe: We're currently setup to do this dep in the other direction, with `inspec-bin` depending on `inspec`. We need that so the the `inspec` library (which is free to use) may be used by the community without requiring `inspec-bin` (which is license-encumbered). That was part of our commitment to the community in the business model change.

clintoncwolfe (changes requested):
Need to remove the dep on inspec-bin, then should be all set AFAICT.

inspec.gemspec
zenspider: I thought that was inspec-core now?

Otherwise, what's the point of inspec-core?

inspec.gemspec
zenspider: That's gross... I kinda like it.

zenspider: Also: 2 gems? We ship 3. 

https://rubygems.org/gems/inspec-bin
https://rubygems.org/gems/inspec-core
https://rubygems.org/gems/inspec

miah (approved):

inspec.gemspec
clintoncwolfe: We ship 4 gems. There is also inspec-core-bin.

inspec.gemspec
clintoncwolfe: inspec-core is a lighter-weight version of the library that does not contain anything that needs to be compiled, nor any of the "heavy" plugins such as AWS, Azure, etc. Along with the library, we provide a licensed binstub gem, inspec-core-bin.

clintoncwolfe (approved):
The linter is unhappy, but looks good now to me otherwise - thanks!

PR #4856, opened 2020-01-27
Detect install type via install path

### LINKED ISSUE #4585: Telemetry: detect installation type
## Detailed Description

Detect the installation type - omnibus, homebrew, rubygem, audit cookbook, etc, if possible. See configure_telemeter in base_cli.

clintoncwolfe:

## Description

Provides several new predicates and one classification method that guess at how InSpec was installed. Possible values include "chef-workstation", "chefdk", "docker", "habitat", "omnibus", "rubygem", and "source".

A unit test file is provided. This would be very impractical to test functionally, but I got the values in the test file (and in the implementation) by manually running each of the different scenarios. That said, I'm sure this could be easily fooled.

## Related Issue

Closes #4585

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/globals.rb
+++ b/lib/inspec/globals.rb
@@ -1,9 +1,15 @@
+require_relative "utils/install_context"
+
 module Inspec
+
+  extend Inspec::InstallContextHelpers
+
   def self.config_dir
     ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(Dir.home, ".inspec")
   end
 
   def self.src_root
-    File.expand_path(File.join(__FILE__, "..", "..", ".."))
+    @src_root ||= File.expand_path(File.join(__FILE__, "..", "..", ".."))
   end
+
 end
--- a/lib/inspec/globals.rb
+++ b/lib/inspec/globals.rb
@@ -9,7 +9,7 @@ module Inspec
   end
 
   def self.src_root
-    @src_root ||= File.expand_path(File.join(__FILE__, "..", "..", ".."))
+    @src_root ||= File.expand_path(File.join(__FILE__, "../../.."))
   end
 
 end
--- a/lib/inspec/utils/install_context.rb
+++ b/lib/inspec/utils/install_context.rb
@@ -54,13 +54,7 @@ module Inspec
     end
 
     def path_exist?(path)
-      # This exists for testing
-      if respond_to? :dummy_paths
-        dummy_paths.include? path
-      else
-        File.exist? path
-      end
+      File.exist? path
     end
-
   end
 end
--- a/lib/inspec/utils/install_context.rb
+++ b/lib/inspec/utils/install_context.rb
@@ -31,7 +31,7 @@ module Inspec
 
     def docker_install?
       # Our docker image is based on alpine. This could be easily fooled.
-      !!(rubygem_install? && File.exist?("/etc/alpine-release")) && File.exist?("/.dockerenv")
+      !!(rubygem_install? && path_exist?("/etc/alpine-release")) && path_exist?("/.dockerenv")
     end
 
     def habitat_install?
@@ -49,7 +49,16 @@ module Inspec
     def source_install?
       # These are a couple of examples of dirs removed during packaging
       %w{habitat test}.all? do |devdir|
-        Dir.exist?("#{src_root}/#{devdir}")
+        path_exist?("#{src_root}/#{devdir}")
+      end
+    end
+
+    def path_exist?(path)
+      # This exists for testing
+      if respond_to? :dummy_paths
+        dummy_paths.include? path
+      else
+        File.exist? path
       end
     end
 
--- a/lib/inspec/utils/install_context.rb
+++ b/lib/inspec/utils/install_context.rb
@@ -19,6 +19,8 @@ module Inspec
       "unknown"
     end
 
+    private
+
     def chef_workstation_install?
       !!(src_root.start_with?("/opt/chef-workstation") || src_root.start_with?("C:/opscode/chef-workstation"))
     end
@@ -29,7 +31,7 @@ module Inspec
 
     def docker_install?
       # Our docker image is based on alpine. This could be easily fooled.
-      !!(rubygem_install? && File.exist?("/etc/alpine-release"))
+      !!(rubygem_install? && File.exist?("/etc/alpine-release")) && File.exist?("/.dockerenv")
     end
 
     def habitat_install?
--- a/lib/inspec/utils/install_context.rb
+++ b/lib/inspec/utils/install_context.rb
@@ -1,0 +1,55 @@
+module Inspec
+
+  # Hueristics to determine how InSpec was installed.
+  module InstallContextHelpers
+
+    def guess_install_context
+      # These all work by simple path recognition
+      return "chef-workstation" if chef_workstation_install?
+      return "omnibus" if omnibus_install?
+      return "chefdk" if chefdk_install?
+      return "habitat" if habitat_install?
+
+      # Order matters here - gem mode is easily mistaken for one of the above
+      return "docker" if docker_install?
+      return "rubygem" if rubygem_install?
+
+      return "source" if source_install?
+
+      "unknown"
+    end
+
+    def chef_workstation_install?
+      !!(src_root.start_with?("/opt/chef-workstation") || src_root.start_with?("C:/opscode/chef-workstation"))
+    end
+
+    def chefdk_install?
+      !!(src_root.start_with?("/opt/chef-dk") || src_root.start_with?("C:/opscode/chef-dk"))
+    end
+
+    def docker_install?
+      # Our docker image is based on alpine. This could be easily fooled.
+      !!(rubygem_install? && File.exist?("/etc/alpine-release"))
+    end
+
+    def habitat_install?
+      !!src_root.match(%r{hab/pkgs/chef/inspec/\d+\.\d+\.\d+/\d{14}})
+    end
+
+    def omnibus_install?
+      !!(src_root.start_with?("/opt/inspec") || src_root.start_with?("C:/opscode/inspec"))
+    end
+
+    def rubygem_install?
+      !!src_root.match(%r{gems/inspec-\d+\.\d+\.\d+})
+    end
+
+    def source_install?
+      # These are a couple of examples of dirs removed during packaging
+      %w{habitat test}.all? do |devdir|
+        Dir.exist?("#{src_root}/#{devdir}")
+      end
+    end
+
+  end
+end
--- a/lib/inspec/utils/install_context.rb
+++ b/lib/inspec/utils/install_context.rb
@@ -1,6 +1,6 @@
 module Inspec
 
-  # Hueristics to determine how InSpec was installed.
+  # Heuristics to determine how InSpec was installed.
   module InstallContextHelpers
 
     def guess_install_context
--- a/test/unit/utils/install_context_test.rb
+++ b/test/unit/utils/install_context_test.rb
@@ -22,6 +22,7 @@ describe Inspec::InstallContextHelpers do
     before do
       Inspec.expects(:src_root).at_least_once.returns("/somewhere/gems/inspec-4.18.39")
       File.expects(:exist?).at_least_once.with("/etc/alpine-release").returns(true)
+      File.expects(:exist?).at_least_once.with("/.dockerenv").returns(true)
     end
 
     it "should properly detect a Docker install" do
--- a/test/unit/utils/install_context_test.rb
+++ b/test/unit/utils/install_context_test.rb
@@ -8,11 +8,11 @@ def assert_install_contexts(test_obj, test_expected_to_be_true, also_rubygem)
   should_be_false -= [test_expected_to_be_true]
   should_be_false = should_be_false.map { |m| "#{m}_install?".tr("-", "_").to_sym }
   should_be_false -= [:rubygem_install?] if also_rubygem
-  should_be_false.each { |m| expect(test_obj.send(m)).must_equal false }
+  should_be_false.each { |m| _(test_obj.send(m)).must_equal false }
 
   should_be_true   = ["#{test_expected_to_be_true}_install?".tr("-", "_").to_sym]
   should_be_true  += [:rubygem_install?] if also_rubygem
-  should_be_true.each { |m| expect(test_obj.send(m)).must_equal true }
+  should_be_true.each { |m| _(test_obj.send(m)).must_equal true }
 
   expect(test_obj.guess_install_context).must_equal test_expected_to_be_true
 end
--- a/test/unit/utils/install_context_test.rb
+++ b/test/unit/utils/install_context_test.rb
@@ -1,0 +1,86 @@
+require "helper"
+require "inspec/globals"
+require "inspec/utils/install_context"
+
+def check_install_contexts(test_expected_to_be_true, also_rubygem)
+  should_be_false = %w{chef-workstation chefdk docker
+                     habitat omnibus rubygem source}
+  should_be_false -= [test_expected_to_be_true]
+  should_be_false = should_be_false.map { |m| "#{m}_install?".tr("-", "_").to_sym }
+  should_be_false -= [:rubygem_install?] if also_rubygem
+  should_be_false.each { |m| expect(Inspec.method(m).call).must_equal false }
+
+  should_be_true   = ["#{test_expected_to_be_true}_install?".tr("-", "_").to_sym]
+  should_be_true  += [:rubygem_install?] if also_rubygem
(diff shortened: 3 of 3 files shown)


test/unit/utils/install_context_test.rb
```diff
@@ -0,0 +1,86 @@
...
+require "inspec/globals"
+require "inspec/utils/install_context"
+
+def check_install_contexts(test_expected_to_be_true, also_rubygem)
+  should_be_false = %w{chef-workstation chefdk docker
+                     habitat omnibus rubygem source}
+  should_be_false -= [test_expected_to_be_true]
+  should_be_false = should_be_false.map { |m| "#{m}_install?".tr("-", "_").to_sym }
+  should_be_false -= [:rubygem_install?] if also_rubygem
+  should_be_false.each { |m| expect(Inspec.method(m).call).must_equal false }
```
zenspider: 1. Just use `send`, it's faster and clearer and doesn't produce garbage (`method` is always a copy).

```
Comparison:
                send:  6857461.8 i/s
         method+call:  2475982.7 i/s - 2.77x  slower
```

2. Use `_` instead of `expect` to be consistent with the rest of the tests.

3. A failure here will say `expected false but got true` or something similar. Much clearer when it fails:

```ruby
_(Inspec).must_be m
```

That will say something like `expected Inspec to be : omnibus_install?`

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def rubygem_install?
+      !!src_root.match(%r{gems/inspec-\d+\.\d+\.\d+})
+    end
+
+    def source_install?
+      # These are a couple of examples of dirs removed during packaging
+      %w{habitat test}.all? do |devdir|
+        Dir.exist?("#{src_root}/#{devdir}")
```
zenspider: I honestly didn't know this method existed... Usually we use `File.directory?`.

test/unit/utils/install_context_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  should_be_true.each { |m| expect(Inspec.method(m).call).must_equal true }
+
+  expect(Inspec.guess_install_context).must_equal test_expected_to_be_true
+end
+
+describe Inspec::InstallContextHelpers do
+  describe "when it looks like a Docker installation" do
+    before do
+      Inspec.expects(:src_root).at_least_once.returns("/somewhere/gems/inspec-4.18.39")
+      File.expects(:exist?).at_least_once.with("/etc/alpine-release").returns(true)
```
zenspider: Does `at_least_once` actually do anything but take up space? I almost never mock but when I do I absolutely avoid quantifiers like the plague so I don't know.

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+      !!(src_root.start_with?("/opt/chef-workstation") || src_root.start_with?("C:/opscode/chef-workstation"))
+    end
+
+    def chefdk_install?
+      !!(src_root.start_with?("/opt/chef-dk") || src_root.start_with?("C:/opscode/chef-dk"))
+    end
+
+    def docker_install?
+      # Our docker image is based on alpine. This could be easily fooled.
+      !!(rubygem_install? && File.exist?("/etc/alpine-release"))
```
zenspider: You probably want `File.file?` here.

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def habitat_install?
+      !!src_root.match(%r{hab/pkgs/chef/inspec/\d+\.\d+\.\d+/\d{14}})
+    end
+
+    def omnibus_install?
+      !!(src_root.start_with?("/opt/inspec") || src_root.start_with?("C:/opscode/inspec"))
```
zenspider, line 42: This will fail on windows inside of vagrant... it's something stupid like `//srvsomething`

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
+module Inspec
+
+  # Hueristics to determine how InSpec was installed.
```
zenspider: "Heuristics"

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+  module InstallContextHelpers
```
zenspider, line 4: I'd like to avoid "Context" outside of our more "magical"/hairy areas if possible.

lib/inspec/globals.rb
```diff
@@ -1,9 +1,15 @@
+require_relative "utils/install_context"
+
 module Inspec
+
+  extend Inspec::InstallContextHelpers
```
zenspider, line 5: It does't feel like this is necessary. Why do we want more class methods on Inspec? Even if it comes in via a module it's still globals and more top-level stuff we need to manage.

lib/inspec/globals.rb
```diff
@@ -1,9 +1,15 @@
...
+
   def self.config_dir
     ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(Dir.home, ".inspec")
   end
 
   def self.src_root
-    File.expand_path(File.join(__FILE__, "..", "..", ".."))
+    @src_root ||= File.expand_path(File.join(__FILE__, "..", "..", ".."))
```
zenspider: Not your mess, but this can also be clearer as `File.expand_path(__FILE__, "../../..")`

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+    def habitat_install?
+      !!src_root.match(%r{hab/pkgs/chef/inspec/\d+\.\d+\.\d+/\d{14}})
+    end
+
+    def omnibus_install?
+      !!(src_root.start_with?("/opt/inspec") || src_root.start_with?("C:/opscode/inspec"))
+    end
+
+    def rubygem_install?
+      !!src_root.match(%r{gems/inspec-\d+\.\d+\.\d+})
```
zenspider, line 46: Mmm... `vendor` for bundler stuff? I'm thinking of the caching we use on BK.

test/unit/utils/install_context_test.rb
```diff
@@ -0,0 +1,86 @@
+require "helper"
+require "inspec/globals"
+require "inspec/utils/install_context"
+
+def check_install_contexts(test_expected_to_be_true, also_rubygem)
```
zenspider: If this is named to start with `assert` then a failure will show it down at the call instead of inside this method so it's easier to see context.

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
+module Inspec
+
+  # Hueristics to determine how InSpec was installed.
+  module InstallContextHelpers
+
+    def guess_install_context
```
zenspider, line 6: Where is this being used? I assume it is for telemetry, but I thought our current telemetry goal is about _how_ it's being run.

lib/inspec/utils/install_context.rb
```diff
@@ -0,0 +1,55 @@
...
+      return "docker" if docker_install?
+      return "rubygem" if rubygem_install?
+
+      return "source" if source_install?
+
+      "unknown"
+    end
+
+    def chef_workstation_install?
+      !!(src_root.start_with?("/opt/chef-workstation") || src_root.start_with?("C:/opscode/chef-workstation"))
```
zenspider, line 25: We don't really need `true/false` results, do we? Is this just for testing?

zenspider (changes requested):

test/unit/utils/install_context_test.rb
zenspider: One more thing... this design means the tests can't be run in parallel because they're mocking on the globals... but since this is a module, this need not be the case at all.

```ruby
class TestThingy
  include Inspec::InstallContextHelpers
  attr_accessor :src_root
  # ...
end

obj = TestThingy.new src_root: "/somewhere/gems/inspec-4.18.39"
check_install_contexts obj, "docker", true
```

etc

This redesign make it easy to use `parallelize_me!` and not jam up any other tests that use `File`, `Directory` or other `Inspec` globals.

zenspider (changes requested):

lib/inspec/utils/install_context.rb
clintoncwolfe: `install_context` is the term of art used throughout the [telemetry schema](https://github.com/chef/es-telemetry-pipeline/blob/master/schema/event.schema.json#L74).

lib/inspec/utils/install_context.rb
clintoncwolfe: For the first instrumentation point, which covers invocation, the return value of this will be one of the data fields. So a row will include fields like "a `check` invocation from a workstation installation using version 4.18.61 on windows2016 ..."

lib/inspec/utils/install_context.rb
clintoncwolfe: That's not the case that's being covered here, though - we're looking for an actual installed MSI omnibus package on the machine, and this is where it ends up.

lib/inspec/utils/install_context.rb
clintoncwolfe: I may need this explained to me.

When vendored, Bundler makes a directory `vendor/cache` full of gems - but the gems are not yet installed/expanded.  Presumably when Bundler / Rubygems activates a gem it then installs the gem somewhere? So... wouldn't the above regex work?

test/unit/utils/install_context_test.rb
clintoncwolfe: Thanks for the tip about `send` - while I didn't see that much of a gain overall, it is somewhat faster, and certainly clearer and briefer.

I really dislike `_`; I feel it is really unreadable and hostile to new contributors (imagine looking for docs on that). 

`_(Inspec).must_be m` doesn't express the same test at all - I want to run `Inspec.omnibus_install?` (whose output really should be true or false). I want a better, boolean-specific Expectation, or one in which I can provide a message.

test/unit/utils/install_context_test.rb
clintoncwolfe: `at_least_once` is required here by `mocha`; without it you are limited to exactly one invocation. I don't care how many times these are invoked.

clintoncwolfe: Re: the parallelization issue, I see how that can be addressed for src_root, but for File and Dir and I don't see how we can avoid mocking them.

test/unit/utils/install_context_test.rb
miah: I agree that `_` is weird, but finding docs on it wasn't too bad. I searched for `ruby minitest underscore method` on a private tab and this was the first result: https://chriskottom.com/blog/2015/04/unexpected-the-new-minitest-spec-syntax/

lib/inspec/utils/install_context.rb
miah: The way we determine if we're in docker in our virtualization resource is different as well.

This SO Answer might point us towards a better path for both the virtualization resource, and this method.

https://stackoverflow.com/a/23558932

miah (changes requested):
I left a few comments inline. I'm 99% approve on this, `#docker_install?` is my only worry.

lib/inspec/utils/install_context.rb
clintoncwolfe: So how about I simply rely on the virtualization resource here?

test/unit/utils/install_context_test.rb
clintoncwolfe: OK, I've put in a helper method to check for a path, and to look for paths in an optional ivar which will be set if it is under the test harness. That could probably be more graceful.

lib/inspec/utils/install_context.rb
clintoncwolfe: Relying on the virtualization resource turned out to be quite heavy; I've added a simple check for /.dockerenv . That's not foolproof either, but this isn't aiming for perfection.

test/unit/utils/install_context_test.rb
zenspider: official docs: http://docs.seattlerb.org/minitest/Minitest/Spec/DSL/InstanceMethods.html

zenspider: You might disagree with `_` here, but you already signed up for it when you accepted the PR flipping them all over. Consistency is important. We already have too fragmented a test codebase.

zenspider: Official doco: http://docs.seattlerb.org/minitest/Minitest/Spec/DSL/InstanceMethods.html

miah (approved):

lib/inspec/utils/install_context.rb
zenspider: I think when vendored, they're installed under `vendor` as well... Our BK "bundle install" now takes <20s because we've untarred the vendor directory from cache.

test/unit/utils/install_context_test.rb
```diff
@@ -2,17 +2,17 @@
...
   should_be_false -= [test_expected_to_be_true]
   should_be_false = should_be_false.map { |m| "#{m}_install?".tr("-", "_").to_sym }
   should_be_false -= [:rubygem_install?] if also_rubygem
-  should_be_false.each { |m| expect(Inspec.method(m).call).must_equal false }
+  should_be_false.each { |m| expect(Inspec.send(m)).must_equal false }
 
   should_be_true   = ["#{test_expected_to_be_true}_install?".tr("-", "_").to_sym]
   should_be_true  += [:rubygem_install?] if also_rubygem
-  should_be_true.each { |m| expect(Inspec.method(m).call).must_equal true }
+  should_be_true.each { |m| expect(Inspec.send(m)).must_equal true }
```
zenspider: How do you debug when all it says is`Expected true, got false` ?

Compare to `Expected Inspec to be docker` (these names aren't exactly adjectives... so it's a little clunky)

lib/inspec/utils/install_context.rb
```diff
@@ -49,7 +49,16 @@ def rubygem_install?
...
       %w{habitat test}.all? do |devdir|
-        Dir.exist?("#{src_root}/#{devdir}")
+        path_exist?("#{src_root}/#{devdir}")
+      end
+    end
+
+    def path_exist?(path)
+      # This exists for testing
+      if respond_to? :dummy_paths
+        dummy_paths.include? path
```
zenspider: This test code shouldn't be here. (see below)

test/unit/utils/install_context_test.rb
```diff
@@ -2,64 +2,65 @@
...
+  expect(test_obj.guess_install_context).must_equal test_expected_to_be_true
+end
+
+class InstallContextTester
+  include Inspec::InstallContextHelpers
+  attr_accessor :src_root, :dummy_paths
+  def initialize(src_root: "", dummy_paths: [])
+    @src_root = src_root
+    @dummy_paths = dummy_paths
+  end
```
zenspider, line 26: This class can have its own `path_exist?` override for testing.

test/unit/utils/install_context_test.rb
```diff
@@ -74,11 +75,9 @@ def assert_install_contexts(test_expected_to_be_true, also_rubygem)
         "omnibus" => "inspec",
       }.each do |inst_mode, inst_subdir|
         describe "when it looks like a #{inst_mode} installation" do
-          before do
-            Inspec.expects(:src_root).at_least_once.returns("#{inst_dir}/#{inst_subdir}/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.18.39")
-          end
           it "should properly detect a #{os_name} #{inst_mode} install" do
-            assert_install_contexts(inst_mode, true)
+            test_obj = InstallContextTester.new(src_root: "#{inst_dir}/#{inst_subdir}/embedded/lib/ruby/gems/2.6.0/gems/inspec-4.18.39")
+            assert_install_contexts(test_obj, inst_mode, true)
```
zenspider, line 88: nice

zenspider (changes requested):

test/unit/utils/install_context_test.rb
clintoncwolfe: As I described above, your suggestion is not an equivalent test. We are testing predicates with expected boolean results, not referencing a method on a module.

When a test fails, it's easy to see which one - I've forced a failure here:
```
Inspec::InstallContextHelpers::when it looks like a Habitat installation#test_0001_should properly detect a habitat install [/Users/cwolfe/sandbox/inspec/inspec-telemetry-2/test/unit/utils/install_context_test.rb:53]:
Expected: false
  Actual: true
```

I do wish I had a better expectation - perhaps one that could take a custom message, in which case I could provide user context. But this is not worthwhile at this point; it's a simple matter to add a pry statement should it ever break.

zenspider (approved):

PR #4862, opened 2020-01-28
Fix up the platform on etc_hosts, virtualization, ini, and xml resources to fix OS disparity.

### LINKED ISSUE #4857: First pass on resource disparity
Starting with:

```
apache_conf            debian linux             # debian+linux -> unix, windows???
etc_hosts              bsd    linux windows     # bsd+linux -> unix
filesystem             linux  windows           # linux -> unix
ini                    unix   windows           # -> os
postfix_conf           linux                    # linux -> unix, windows???
virtualization         linux                    # ??? unix? windows???
xml                    unix   windows           # -> os
```

See epic: https://github.com/inspec/inspec/issues/4782

zenspider:
Fixes #4857.

### DIFF
--- a/lib/inspec/resources/filesystem.rb
+++ b/lib/inspec/resources/filesystem.rb
@@ -3,7 +3,7 @@ require "inspec/resources/command"
 module Inspec::Resources
   class FileSystemResource < Inspec.resource(1)
     name "filesystem"
-    supports platform: "linux"
+    supports platform: "unix"
     supports platform: "windows"
     desc "Use the filesystem InSpec resource to test file system"
     example <<~EXAMPLE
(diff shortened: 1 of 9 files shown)


lib/inspec/resources/filesystem.rb
```diff
@@ -93,7 +93,7 @@ def initialize(inspec)
     end
   end
 
-  class LinuxFileSystemResource < FsManagement
+  class UnixFileSystemResource < FsManagement
     def info(partition)
       cmd = inspec.command("df #{partition} -PT")
```
clintoncwolfe, line 98: This sort-of doesn't work on FreeBSD 11:
```
[vagrant@bazinga ~]$ df / -PT
df: -PT: No such file or directory
Filesystem   1K-blocks   Used    Avail Capacity  Mounted on
/dev/ada0s1a  30450780 820904 27193816     3%    /
[vagrant@bazinga ~]$ echo $?
1
```

My Vagrantfile was simply:
```
Vagrant.configure("2") do |config|
  config.vm.box = "generic/freebsd11"
end
```

clintoncwolfe (changes requested):
This all looks good, but I think we need to add a special case for BSDs for `filesystem`. Otherwise great stuff!

lib/inspec/resources/filesystem.rb
zenspider: Filed https://github.com/inspec/inspec/issues/4868

PR #4869, opened 2020-01-29
Refactored crontab resource

zenspider:
Specifically the parsing was pretty unreadable. This refactoring
emphasizes using a default hash and merging in the differences for
what was parsed. Everything is funneled through one `merge_crontab`
method that does the actual parsing and the system vs user differences
are merged in after in their respective parse methods.


clintoncwolfe (approved):
Thanks - words > indices :-)

miah (approved):

PR #4898, opened 2020-02-07
Distinguish between user plugin gems and system gems and load system gems normally.

zenspider:


clintoncwolfe (approved):
Thanks

miah (approved):

PR #4913, opened 2020-02-12
Use require_relative instead of require where possible

tas50:
require_relative is faster than require

See https://github.com/rspec/rspec-expectations/pull/476#issuecomment-35848905

Aha! Link: https://chef.aha.io/features/SH-2170


zenspider (approved):
I'm not actually reviewing the content... just the idea.

clintoncwolfe (approved):

clintoncwolfe (commented):
We need to go through this and check the boundary between inspec and inspec-core to verify that there are no require_relatives being attempted across gem boundaries.

Schwad:
There's a nice six-year-long discussion on Ruby-lang around this issue here: https://bugs.ruby-lang.org/issues/8781

tas50:
Updated against the latest master

clintoncwolfe:
FWIW, I see wallclock time on `test/fixtures/profiles/complete-profile` from ~3.100s to ~2.950s using `inspec exec` under source with bundler. May be more dramatic under a real omnibus install.

`--version` dropped from 2.95 to 2.4, so that's good.

tas50:
Those are some pretty substantial gains for no actual code change.

james-stocks:
I've hesitated on this PR since Clinton mentioned the risk of requiring relative across gem boundaries.

It might be good to at least do https://github.com/inspec/inspec/issues/5106 first so we have per-commit confidence that InSpec's 4 gems all work. Though `test/integration` might not yet contain a test case that would catch relative requires that are outside the core gem's scope.

PR #4946, opened 2020-03-06
Create updated Release Process docs  

clintoncwolfe:
## Description

Describes the current release process, gotchas, and links to key expeditor/ buildkite configurations.

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -1,0 +1,161 @@
+# Releasing Chef InSpec
+
+This document describes the release process for Chef InSpec as used internally at Chef Software.
+
+## Promote Dependency Gems
+
+It's important that you promote dependency gems, and thus release them to rubygems.org, BEFORE the last PR that you intend to release merges into inspec/inspec. This is because during the `inspec` merge process, software artifacts, such as omnibus packages, Habitat packages, and Docker images, are created that embed the dependencies. During promotion, these artifacts are simply promoted, rather than rebuilt. If you need a later version of one of these dependencies and have no legitimate PR to merge, you may need a dummy PR.
+
+### Key Dependencies
+
+Any change to these gems must be promoted to rubygems.org prior to releasing Chef InSpec.
+
+#### `train` / `train-core`
+
+Train is the connectivity library for Chef InSpec, and is the most critical dependency of Chef InSpec.
+
+`train-core` is a stripped-down version of `train`, with no cloud provider plugins and no gems that require compilation during installation.
+
+You can promote `train` by running, in #inspec-notify:
+
+```
+/expeditor promote inspec/train:master 3.2.23
+```
+
+Note that the version you give to Expeditor is the plain version (not the tag, which includes a `v` prefix).
+
+#### `train-aws`, `train-winrm`, and `train-habitat`
+
+These gems are the transports, or connectivity drivers, that have been fully plugin-ized.
+
+`train-aws` in particular is critical because it carries the gem dependencies for all AWS libraries required by `inspec-aws`. When the gem dependencies change, you must update `train-aws`, and so must `train` and `inspec`.
+
+You can promote each of these gems by running in #inspec-notify:
+
+```
+/expeditor promote inspec/GEMNAME:master VERSION
+```
+
+## Merging Chef InSpec PRs
+
+At the time of writing, a merge typically takes about 30-40 minutes of test time.
+
+Watch the Slack channel #inspec-notify for messages about the success or failure of various steps.
+
+Connect to the Chef VPN to fetch Expeditor logs in the event of a failure.
+
+### Check Expeditor Labels
+
+Most, though not all, PRs should not have any Expeditor control labels. The patchlevel (4.18.X) will be automatically incremented by the [expeditor configuration](https://github.com/inspec/inspec/blob/44fe144732e1e0abb2594957a880c5f1821e7774/.expeditor/config.yml#L117) and the [version bump script](https://github.com/inspec/inspec/blob/master/.expeditor/update_version.sh).
+
+Here are the Expeditor control labels, and the circumstances under which they should be used:
+
+ * Expeditor: Bump Minor Version - Use when a significant new feature is being released.
+ * Expeditor: Skip Changelog - Should only be used for dummy PRs.
+ * Expeditor: Skip Version Bump - Use for non-code-change PRs, such as website or CI changes.
+ * Expeditor: Skip Omnibus - Use for non-code-change PRs, such as website or CI changes.
+ * Expeditor: Skip Habitat - Use for non-code-change PRs, such as website or CI changes.
+
+### Click Merge
+
+This task should be straightforward, assuming the big merge button is green.
+
+You will see a message in #inspec-notify from Expeditor that it "performed actions for merged_inspec/inspec#1234" in this case for PR 1234. Among the messages will be links to the Omnibus and Habitat builds.
+
+You'll see a message in #inspec-notify from Expeditor that it "performed actions for merged_inspec/inspec#1234" (for PR 1234, for example). Among the messages will be links to the Omnibus and Habitat builds.
+
+### Watch Omnibus Build
+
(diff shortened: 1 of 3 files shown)


RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+ * Expeditor: Skip Omnibus - Use for non-code-change PRs, such as website or CI changes.
+ * Expeditor: Skip Habitat - Use for non-code-change PRs, such as website or CI changes.
+
+### Click Merge
+
+This should be straightforward, assuming the big merge button is green.
+
+If the merge button is red, take a moment and reminisce about various regrets in your life, then make your choice. Merging a build that fails tests is counterproductive; we have further tests downstream during the post-merge process.
+
+You'll see a message in #inspec-notify from Expeditor that it "performed actions for merged_inspec/inspec#1234" (for PR 1234, for example). Among the messages will be links to the Omnibus and Habitat builds.
```
james-stocks: I'd remove the bit about there being a choice, and just point the reader to the expeditor alerts in #inspec-notify 

I know sometimes we will have to admin-force the merge, but it might be unclear if the docs suggest that it's OK _sometimes_.

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+
+The Habitat build creates Habitat .hart packages for Linux and Windows. The [Expeditor configuration](https://github.com/inspec/inspec/blob/44fe144732e1e0abb2594957a880c5f1821e7774/.expeditor/config.yml#L138) drives a [BuildKite configuration](https://github.com/inspec/inspec/blob/master/.expeditor/build.habitat.yml).
+
+The habitat build is unreliable, usually due to network timeouts on the Windows machines. It often requires manual retries by clicking the retry button in BuildKite (obtain the link from the message posted by Expeditor in #inspec-notify). If you do not retry, later steps will fail.
+
+When the hab build succeeds, the packages will be placed on the Hab builder in the `unstable` channel for public consumption.
+
+### Docker Image Built and Released
+
+We also release a Docker image, which contains an Alpine system and InSpec installed from a gem, with the EXEC of the Docker image being `inspec`. It's a simple way to ship the dependencies of inspec.
```
james-stocks: Would it be correct to point to https://github.com/inspec/inspec/blob/44fe144732e1e0abb2594957a880c5f1821e7774/.expeditor/config.yml#L150 at this line, to show where the docker build is configured to happen automatically?

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+ * inspec-core-bin contains an `inspec` executable and is encumbered by commercial licensing
+
+### Update Pending Release Notes
+
+As each pull request is merged, update the [Pending Release Notes](https://github.com/inspec/inspec/wiki/Pending-Release-Notes). Some guidelines:
+ * Don't include minor or non-customer visible changes, such as CI changes or test harness changes
+ * Do include bug fixes, features, etc - things that impact the user.
+ * Your words will be edited by the Docs team, but try help them along by keeping it human oriented; this isn't the technical-oriented CHANGELOG.
+ * Add the PR number as a reference for the Docs team. They will typically remove them, but they may need to get more context by looking up the originating PR.
+ * It's preferable to add notes as things merged, lest they be forgotten; sifting through a pile of merged PRs in a rush is a chore!
```
james-stocks: This is true.

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+
+The Omnibus build is generally reliable, if somewhat slow.
+
+When the omnibus build succeeds, omnitruck delivers the packages to various package repos in `unstable` channels for public consumption.
+
+### Watch Habitat Build
+
+The Habitat build creates Habitat .hart packages for Linux and Windows. The [Expeditor configuration](https://github.com/inspec/inspec/blob/44fe144732e1e0abb2594957a880c5f1821e7774/.expeditor/config.yml#L138) drives a [BuildKite configuration](https://github.com/inspec/inspec/blob/master/.expeditor/build.habitat.yml).
+
+The habitat build is unreliable, usually due to network timeouts on the Windows machines. It often requires manual retries by clicking the retry button in BuildKite (obtain the link from the message posted by Expeditor in #inspec-notify). If you do not retry, later steps will fail.
```
stevendanna: Do we have an issue open to investigate this unreliability?  It seems like we could add retries directly into the build process around core network dependencies.

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+train-core is a stripped-down version of train, with no cloud provider plugins and no gems that require compilation during installation.
+
+You can promote train by running, in #inspec-notify,
+```
+/expeditor promote inspec/train:master 3.2.23
+```
+
+Note that the version you give to Expeditor is the plain version (not the tag, which includes a `v` prefix).
+
+#### train-aws, train-winrm, and train-habitat
```
stevendanna: Do you promote these gems in the same way as you promote inspec/train above?

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Check Expeditor Labels
+
+Most PRs shouldn't have any Expeditor control labels. The patchlevel (4.18.X) will be automatically incremented by the [expeditor configuration](https://github.com/inspec/inspec/blob/44fe144732e1e0abb2594957a880c5f1821e7774/.expeditor/config.yml#L117) and the [version bump script](https://github.com/inspec/inspec/blob/master/.expeditor/update_version.sh).
+
+ * Expeditor: Bump Minor Version - Use when a significant new feature is being released.
+ * Expeditor: Skip Changelog - Should only be used for dummy PRs.
+ * Expeditor: Skip Version Bump - Use for non-code-change PRs, such as website or CI changes.
+ * Expeditor: Skip Omnibus - Use for non-code-change PRs, such as website or CI changes.
+ * Expeditor: Skip Habitat - Use for non-code-change PRs, such as website or CI changes.
```
stevendanna, line 57: It is a little confusing why we list these expeditor control labels where while the instruction says that we shouldn't use them.

Also, if we don't use these, can we remove them from the repository setup to avoid the risk completely?

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,151 @@
...
+
+The habitat build is unreliable, usually due to network timeouts on the Windows machines. It often requires manual retries by clicking the retry button in BuildKite (obtain the link from the message posted by Expeditor in #inspec-notify). If you do not retry, later steps will fail.
+
+When the hab build succeeds, the packages will be placed on the Hab builder in the `unstable` channel for public consumption.
+
+### Docker Image Built and Released
+
+We also release a Docker image, which contains an Alpine system and InSpec installed from a gem, with the EXEC of the Docker image being `inspec`. It's a simple way to ship the dependencies of inspec.
+
+When the Docker build succeeds, it is released directly to the `stable` channel.
```
stevendanna: This seems incongruent with the other actions. Is there a reason this is released directly to stable that we can include here?

stevendanna (commented):
I've left some inline questions as a person who hasn't done the inspec release.

RELEASE_PROCESS.md
clintoncwolfe: Yes, added.

RELEASE_PROCESS.md
clintoncwolfe: Clarified. It is an inconsistent process. The purpose of this document isn't  to make the process perfect, but to document the existing process.

RELEASE_PROCESS.md
clintoncwolfe: The instructions say *most* PRs should not use the labels. They certainly have legitimate uses.  I've added clarifying language.

RELEASE_PROCESS.md
clintoncwolfe: There is a single retry already in place.  The cause varies, and has been a source of several support investigations with releng.

kekaichinose (commented):
I'm good, pending @james-stocks review

james-stocks (approved):
This is great info, thanks @clintoncwolfe :)

RELEASE_PROCESS.md
stevendanna: @clintoncwolfe Is that retry at the buildkite level? I was referring to retries we could build into the underlying tooling.  The lower-level retry tends to be impactful in my experience.

stevendanna (approved):

james-stocks:
@clintoncwolfe can you squash this PRs commits before I merge it please?

PR #4949, opened 2020-03-09
Add support for alternate Gem source for plugin installation/search

thheinen:
## Description
In enterprise environments there is often no access to rubygems.org but an internal server is proxying the requests or even replacing it completely. This PR implements the --source option for InSpec plugins to allow referencing those internals servers instead of rubygems.org. It is added both for `plugin install` and `plugin search`

## Related Issue
Related to #3297 for more air-gap capabilities

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -3,6 +3,7 @@
 require "singleton"
 require "forwardable"
 require "fileutils"
+require "uri"
 
 # Gem extensions for doing unusual things - not loaded by Gem default
 require "rubygems/package"
@@ -56,6 +57,7 @@ module Inspec::Plugin::V2
     # @option opts [String] :gem_file Path to a local gem file to install from
     # @option opts [String] :path Path to a file to be used as the entry point for a path-based plugin
     # @option opts [String] :version Version constraint for remote gem installs
+    # @option opts [String] :source Alternate URL to use instead of rubygems.org
     def install(plugin_name, opts = {})
       # TODO: - check plugins.json for validity before trying anything that needs to modify it.
       validate_installation_opts(plugin_name, opts)
@@ -127,6 +129,11 @@ module Inspec::Plugin::V2
       validate_search_opts(plugin_query, opts)
 
       fetcher = Gem::SpecFetcher.fetcher
+      if opts[:source]
+        source_list = Gem::SourceList.from([opts[:source]])
+        fetcher = Gem::SpecFetcher.new(source_list)
+      end
+
       matched_tuples = []
       if opts[:exact]
         matched_tuples = fetcher.detect(opts[:scope]) { |tuple| tuple.name == plugin_query }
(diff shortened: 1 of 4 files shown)


lib/inspec/plugin/v2/installer.rb
```diff
@@ -284,8 +290,16 @@ def install_from_gem_file(requested_plugin_name, opts)
 
     def install_from_remote_gems(requested_plugin_name, opts)
       plugin_dependency = Gem::Dependency.new(requested_plugin_name, opts[:version] || "> 0")
-      # BestSet is rubygems.org API + indexing
-      install_gem_to_plugins_dir(plugin_dependency, [Gem::Resolver::BestSet.new], opts[:update_mode])
+
+      # BestSet is rubygems.org API + indexing, APISet is for custom sources
+      sources = Gem::Resolver::BestSet.new
+      sources = Gem::Resolver::APISet.new(opts[:source]) if opts[:source]
```
clintoncwolfe: It might be better to avoid double initializing `sources` - calling Gem::Resolver::BestSet.new may be expensive.

lib/inspec/plugin/v2/installer.rb
```diff
@@ -284,8 +290,16 @@ def install_from_gem_file(requested_plugin_name, opts)
...
+
+      install_gem_to_plugins_dir(plugin_dependency, [sources], opts[:update_mode])
+    rescue TypeError
+      # Retry with appending the dependencies endpoint
+      sources = Gem::Resolver::APISet.new(opts[:source] + "/api/v1/dependencies")
```
clintoncwolfe: Is this always a required part of the URL? If so, could it be checked for and appended if missing, rather than using trying and failing?

clintoncwolfe (commented):
Thanks for this @tecracer-theinen - I think this is a fantastic addition. I'm almost ready to approve, but I did have a question about the handling of a retry.  Thank you!

lib/inspec/plugin/v2/installer.rb
thheinen: From what I could see in the Rubygems API, they are placing that as a default which makes me assume it's a required part. Do you propose switching the two around or should I rather implement some sort of preflight check using Net::HTTP instead?

lib/inspec/plugin/v2/installer.rb
thheinen: Good point.

stevendanna: One nice way in ruby to implement @clintoncwolfe's suggestion is to do something like

```suggestion
      sources = if opts[:source]
                  Gem::Resolver::APISet.new(opts[:source]) 
                else
                  Gem::Resolver::BestSet.new
                end
```

thheinen:
Sorry for the mess. Typed the latest git command in my test container and... yeah.

lib/plugins/inspec-plugin-manager-cli/lib/inspec-plugin-manager-cli/cli_command.rb
```diff
@@ -102,16 +103,17 @@ def search(search_term) # rubocop: disable Metrics/AbcSize
       #==================================================================#
       #                       inspec plugin install
       #==================================================================#
-      desc "install [-v VERSION] PLUGIN", "Installs a plugin from rubygems.org, a gemfile, or a path to local source."
+      desc "install [options] PLUGIN", "Installs a plugin from rubygems.org, a gemfile, or a path to local source."
       long_desc <<~EOLD
-        PLUGIN may be the name of a gem on rubygems.org that begins with inspec- or train-.
+        PLUGIN may be the name of a gem on rubygems.org or an alternate source that begins with inspec- or train-.
```
james-stocks: It's possible that this could be confused to mean that the alternate source must be begin with `inspec-`

I suggest 

```
PLUGIN may be the name of a gem on rubygems.org (or an alternate source) that begins with `inspec-` or `train-`.
```

or simply remove mention of the source here and have

```
PLUGIN is the name of a gem that begins with `inspec-` or `train-`.
```

thheinen: I picked the first one, as both the `desc` and the following line in the `long_desc` create a tension betweet mentioning only rubygems.org and mentioning local installs. I think, leaving out the reference to external sources would confuse readers as well.

james-stocks:
This change looks good to me, thank you @tecracer-theinen 

There is one open question we should confirm with @clintoncwolfe https://github.com/inspec/inspec/pull/4949#discussion_r392909845

The commits need squashed to a single commit before we can merge.

I believe at the current time we ignore that [code-climate is failing](https://codeclimate.com/github/inspec/inspec/pull/4949)

lib/inspec/plugin/v2/installer.rb
clintoncwolfe: I meant doing a string check (likely by parsing it with URI) on the provided URL, and if it does not contain the right path, then append the correct path.

james-stocks (approved):
Awesome, thanks @tecracer-theinen !

PR #4951, opened 2020-03-11
Fix service not working correctly on windows machines

### LINKED ISSUE #4924: The "service" resource is not supported on your OS yet - Windows 2016
The service resource no longer works on Windows. I have used this resource quite extensively in the past (about 6-8 months ago)

```
C:\Users\Administrator>inspec detect

Name:      windows_server_2016_datacenter
Families:  windows, os
Release:   10.0.14393
Arch:      x86_64

C:\Users\Administrator>inspec version
4.18.85
```
inspec installed from: https://packages.chef.io/files/stable/inspec/4.18.85/windows/2016/inspec-4.18.85-1-x64.msi

Snippet of my test case

```
describe service('SepMasterService') do
  it { should be_installed }
  it { should be_enabled }
  it { should be_running }
end

describe service('CSFalconService') do
  it { should be_installed }
  it { should be_enabled }
  it { should be_running }
```

Snippet of execution test results

```
  Service SepMasterService
     [SKIP]  The `service` resource is not supported on your OS yet.
  Service CSFalconService
     [SKIP]  The `service` resource is not supported on your OS yet.
```

Axuba:
Signed-off-by: Albert Jubany Juàrez <ajubany@externos.itnow.es>

Fixes the service resource so it works on Windows machines.

## Description
The service resource when used on Windows machines it would appear as not supported platform. This was due to the change made some months ago of changing the original if/elsif/else structure to a switch statement. The elsif structure used the `os.windows?` to determine if it was a Windows host and the switch structure used a string comparation. Problem was that Windows machines have different names, rather than just "Windows". I changed it to a REGEX match.

## Related Issue
Fixes #4924

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -104,6 +104,8 @@ module Inspec::Resources
       os = inspec.os
       platform = os[:name]
 
+      return WindowsSrv.new(inspec) if os.windows?
+
       # Ubuntu
       # @see: https://wiki.ubuntu.com/SystemdForUpstartUsers
       # Ubuntu 15.04 : Systemd
@@ -154,8 +156,6 @@ module Inspec::Resources
         SysV.new(inspec, service_ctl)
       when "mac_os_x"
         LaunchCtl.new(inspec, service_ctl)
-      when "windows"
-        WindowsSrv.new(inspec)
       when "freebsd"
         BSDInit.new(inspec, service_ctl)
       when "arch"


james-stocks (approved):
It feels like a bug that train does not present `platform[:name]` as exactly `windows` on all releases of Windows.
Looking at https://github.com/inspec/train/blob/e6d0ccc83864a7fa24c4a236c93f5f5492d13f42/lib/train/platforms/detect/helpers/os_windows.rb it does appear that it queries the host to determine an OS name string.

I say merge this PR but also raise an issue on train suggesting that (to help resource coders) the OS name should just be `windows`

clintoncwolfe (approved):
Thanks! The regex is OK as is, but `os.windows?` would be safer.

lib/inspec/resources/service.rb
```diff
@@ -154,7 +154,7 @@ def select_service_mgmt # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCom
         SysV.new(inspec, service_ctl)
       when "mac_os_x"
         LaunchCtl.new(inspec, service_ctl)
-      when "windows"
+      when /^windows/
```
james-stocks: I wonder if there are other cases that this does not match.
For example a string starting `Microsoft` or a string with uppercase `Windows`.
Note that in train, the name for a Windows OS may be the result of a PowerShell command that we can't predict the exact result for.

For this resource, it may be safer to have this method rely on `os.windows?`

For example, you could put this line at line 106:

`return WindowsSrv.new(inspec) if os.windows?`

...and then delete this `when /^windows/` case.

Axuba: You're absolutely right. This prevents some Windows cases that may not match. I check it on my local lab and if it works I update the MR.

Axuba:
Changes done as suggested by @james-stocks , now it should work more consistently by using the `windows.os?` method that was used before.

stevendanna (approved):
Thanks, I think using `os.windows?` is the right move here rather than trying to maintain that code in two places.

clintoncwolfe (approved):
Thanks!

james-stocks (approved):
Thank you Albert

akr257:
Hi,
Has this issue been fixed? I'm using version 4.18.100 on a Windows 2016 server and the issue still persists.

Axuba:
> Hi,
> Has this issue been fixed? I'm using version 4.18.100 on a Windows 2016 server and the issue still persists.

Hey @akr257 , the fix will be out on the next InSpec release. The changes are not yet implemented on the 4.18.100 version.

james-stocks:
We just released 4.18.104 with this fix

Issue #2509, opened 2018-02-01
virtualization resource should detect vmware


mlcooper:
### Description

This is an enhancement request for the inspec virtualization resource to detect vmware.

### InSpec and Platform Version

1.49.2

### Replication Case
```
describe virtualization do
  its('system') { should include 'vmware' }
  its('role') { should include 'guest' }
end
```
results in the following error on a known good VMware guest (RHEL 7.4)
```
× Virtualization Detection system should eq "vmware"
expected: "vmware"
got: nil
(compared using ==)
× Virtualization Detection role should eq "guest"
expected: "guest"
got: nil
(compared using ==)
```

adamleff:
Thanks for your feature request, @mlcooper.

PopPopDevOps:
I think this could be as easy as the following:

```ruby
    # Detect VMware systems via /sys
    # guests will have the Virtual Machine Communication Interface
    def detect_vmware_from_sys
      return false unless inspec.file('/sys/devices/virtual/dmi/id/sys_vendor').exist?
      @virtualization_data[:system] = 'vmware'
        @virtualization_data[:role] = 'guest'
    end
```

(@zenspider: edited formatting)

clintoncwolfe:
#4954 implemented support for vmware detection, released in 4.18.104, closing

Issue #4782, opened 2019-12-11
Audit of platform support of all core resources


zenspider:
## These probably have problems

```
apache                 unix                     # windows???
bsd_service            unix                     # bsd only?
cpan                   unix                     # windows???
cran                   unix                     # windows???
docker                 unix                     # windows???
docker_container       unix                     # windows???
docker_image           unix                     # windows???
docker_plugin          unix                     # windows???
docker_service         unix                     # windows???
elasticsearch          unix                     # windows???
grub_conf              unix                     # -> linux
http                   unix                     # because curl?
kernel_module          unix                     # linux only?
kernel_parameter       unix                     # linux only?
launchd_service        unix                     # darwin only?
linux_kernel_parameter unix                     # -> linux
mount                  unix                     # windows??? no clue
mysql                  unix                     # windows???
nginx                  unix                     # windows???
nginx_conf             unix                     # windows???
postgres               unix                     # windows???
postgres_hba_conf      unix                     # windows???
postgres_ident_conf    unix                     # windows???
rabbitmq_conf          unix                     # windows???
rabbitmq_config        unix                     # windows???
ssh_config             unix                     # windows??? no?
sshd_config            unix                     # windows??? seems less likely
```

```
apache_conf            debian linux             # debian+linux -> unix, windows???
etc_hosts              bsd    linux windows     # bsd+linux -> unix
filesystem             linux  windows           # linux -> unix
ini                    unix   windows           # -> os
postfix_conf           linux                    # linux -> unix, windows???
virtualization         linux                    # ??? unix? windows???
xml                    unix   windows           # -> os
```

## These are probably fine:

```
csv                                     os
json                                    os
platform                                os
toml                                    os
yaml                                    os
```

```
command                                 unix   windows
directory                               unix   windows
file                                    unix   windows
gem                                     unix   windows
group                                   unix   windows
groups                                  unix   windows
host                                    unix   windows
interface                               unix   windows
key_rsa                                 unix   windows
mysql_conf                              unix   windows
mysql_session                           unix   windows
npm                                     unix   windows
oracledb_session                        unix   windows
os                                      unix   windows
os_env                                  unix   windows
package                                 unix   windows
parse_config                            unix   windows
parse_config_file                       unix   windows
pip                                     unix   windows
port                                    unix   windows
postgres_conf                           unix   windows
postgres_session                        unix   windows
powershell                              unix   windows
processes                               unix   windows
script                                  unix   windows
service                                 unix   windows
[...]

PR #4954, opened 2020-03-11
Improve virtualization resource

thheinen:
## Description
- Implemented VMware and Hyper-V detection on Linux systems
- Implemented VMware, Hyper-V, Virtualbox, KVM and Xen detection on Windows systems
- Added helpers `virtual_system?` and `physical_system?`

## Related Issue

Implements #2509

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/virtualization.rb
+++ b/lib/inspec/resources/virtualization.rb
@@ -1,9 +1,11 @@
 require "hashie/mash"
+require "inspec/resources/powershell"
 
 module Inspec::Resources
   class Virtualization < Inspec.resource(1)
     name "virtualization"
     supports platform: "unix"
+    supports platform: "windows"
     desc "Use the virtualization InSpec audit resource to test various virtualization platforms."
     example <<~EXAMPLE
       describe virtualization do
@@ -25,7 +27,15 @@ module Inspec::Resources
     def initialize
       # TODO: no need for hashie here... in fact, no reason for a hash at all
       @virtualization_data = Hashie::Mash.new
-      collect_data_linux
+
+      if inspec.os.linux?
+        collect_data_linux
+      elsif inspec.os.windows?
+        collect_data_windows
+      end
+
+      # Allows checking for non-virtualized systems as well
+      @virtualization_data[:physical] = @virtualization_data.empty?
     end
 
     # add helper methods for easy access of properties
@@ -34,6 +44,14 @@ module Inspec::Resources
       define_method(property) do
         @virtualization_data[property]
       end
+    end
+
+    def virtual_system?
+      @virtualization_data[:role] == "guest"
+    end
+
+    def physical_system?
+      @virtualization_data[:physical]
     end
 
     def params
@@ -236,6 +254,39 @@ module Inspec::Resources
       true
     end
 
+    # Detect VMware
+    def detect_vmware
+      return false unless inspec.file("/sys/devices/virtual/dmi/id/product_name").exist?
+
+      product_name = inspec.file("/sys/devices/virtual/dmi/id/product_name").content
+
+      if product_name =~ /^VMware/
+        @virtualization_data[:system] = "vmware"
+        @virtualization_data[:role] = "guest"
+      else
+        return false
+      end
+
+      true
+    end
+
+    # Detect Hyper-V
+    # @see https://gallery.technet.microsoft.com/scriptcenter/Get-MachineType-VM-or-ff43f3a9
+    def detect_hyperv
+      return false unless inspec.file("/sys/devices/virtual/dmi/id/product_name").exist?
+
+      product_name = inspec.file("/sys/devices/virtual/dmi/id/product_name").content
+
+      if product_name.rstrip == "Virtual Machine"
+        @virtualization_data[:system] = "hyper-v"
+        @virtualization_data[:role] = "guest"
+      else
+        return false
+      end
+
+      true
+    end
+
     def collect_data_linux
       # This avoids doing multiple detections in a single test
       return unless @virtualization_data.empty?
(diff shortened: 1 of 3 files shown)


lib/inspec/resources/virtualization.rb
```diff
@@ -36,6 +46,14 @@ def initialize
       end
     end
 
+    def virtual_system?
```
james-stocks, line 49: It would be great to update [the documentation for this resource](https://github.com/inspec/inspec/blob/master/docs/resources/virtualization.md.erb) with examples and descriptions for these new `virtual_system?` and `physical_system?` matchers.

lib/inspec/resources/virtualization.rb
```diff
@@ -236,6 +254,39 @@ def detect_lxd
       true
     end
 
+    # Detect VMware
```
james-stocks, line 257: These should be added to the list of possible virtualisation systems in the documentation

lib/inspec/resources/virtualization.rb
```diff
@@ -236,6 +254,39 @@ def detect_lxd
...
+    def detect_vmware
+      return false unless inspec.file("/sys/devices/virtual/dmi/id/product_name").exist?
+
+      product_name = inspec.file("/sys/devices/virtual/dmi/id/product_name").content
+
+      if product_name =~ /^VMware/
+        @virtualization_data[:system] = "vmware"
+        @virtualization_data[:role] = "guest"
```
james-stocks, line 265: I think it's fine to merge this as is, but it would be possible to save having the `else` branch by returning `return true` here then returning `false` if we reach the end of the method.

lib/inspec/resources/virtualization.rb
```diff
@@ -253,7 +304,40 @@ def collect_data_linux
...
       return if detect_openstack
       return if detect_parallels
-      # TODO: where is vmware?
+      return if detect_vmware
+      return if detect_hyperv
+    end
+
+    def windows_computer_system
+      script = "Get-WmiObject -Class Win32_ComputerSystem -ErrorAction Stop | ConvertTo-Json"
+      cmd = inspec.powershell(script)
```
james-stocks, line 313: It's possible to raise ResourceFailed here with the contents of `cmd.stderr` if there is any error output.

lib/inspec/resources/virtualization.rb
```diff
@@ -253,7 +304,40 @@ def collect_data_linux
...
+
+      JSON.parse(cmd.stdout)
+    rescue JSON::ParserError => _e
+      nil
+    end
+
+    def collect_data_windows
```
james-stocks, line 320: I believe this will cause the system to return true for the `physical_system?` method if it is virtualised but is not in the list of hypervisors we check for below.
I think this is acceptable, but we might have to help others in future who ask why InSpec tells them their VM is physical.

test/unit/resources/virtualization_test.rb
```diff
@@ -12,17 +12,7 @@ def mock_proc(mocked_files)
...
-    mock_loader = MockLoader.new(:windows)
-
-    mock_loader.backend.stub :file, mock_proc(mocked_files) do
-      mock_resource = mock_loader.load_resource("virtualization")
-      _(mock_resource.resource_failed?).must_equal true
-    end
-  end
-
-  it "returns nil for all properties if no virutalization platform is found" do
+  it "returns nil for all properties if no virtualization platform is found" do
```
james-stocks, line 15: The full set of tests here would be testing Linux and Windows; for each of the virtualisation systems (at least each that has a distinct execution flow), plus a test for detecting physical on both Linux and Windows.

That is quite a lot of work and involves creating new data for the mock loader. 
I think it is acceptable to merge this as-is, but I will update the resource improvement epic https://github.com/inspec/inspec/issues/4782 and comment that this resource needs testing filled out.

james-stocks (changes requested):
Thank you very much @tecracer-theinen 
Can you please update the documentation for the resource with your new additions?

I left a couple of other comments on the code and the tests, but nothing that stops us merging this change as-is

Thanks!

test/unit/resources/virtualization_test.rb
thheinen: Actually I tried figuring out how to mock file contents and such and had to give up after a while.

Maybe some documentation on how to properly write unit tests and the different patterns would be an idea, showing examples for the use cases instead of needing to dig through the literally thousands of tests (respect!). I suspect that this is mostly straighforward for those who wrote those tests.

lib/inspec/resources/virtualization.rb
thheinen: Just been following the pattern of other helpers in that file :)

lib/inspec/resources/virtualization.rb
james-stocks: I see this is now in the documentation, thanks!

test/unit/resources/virtualization_test.rb
james-stocks: I agree, we need to improve this.
Documenting the patterns and hierarchy of the current tests is a great idea. Or refactor them so that the folder hierarchy and file and test names make the tests self-describing.

james-stocks (approved):
Thank you!

lib/inspec/resources/virtualization.rb
```diff
@@ -253,7 +304,40 @@ def collect_data_linux
...
+      # This avoids doing multiple detections in a single test
+      return unless @virtualization_data.empty?
+
+      case windows_computer_system["Model"]
+      when "Virtual Machine"
+        @virtualization_data[:system] = "hyper-v"
```
tas50, line 326: You may want to look at pulling the full DMI decode logic out of Ohai for use here. It would give you pretty robust cross platform virtualization detection.

PR #4957, opened 2020-03-13
Document describe.one usage

james-stocks:
As suggested by @aaronlippold in community Slack

Fixed section indenting in the same commit.

### DIFF
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -59,7 +59,9 @@ where
 
 ## Advanced concepts
 
-With Chef InSpec it is possible to check if at least one of a collection of checks is true. For example: If a setting is configured in two different locations, you may want to test if either configuration A or configuration B have been set. This is accomplished via `describe.one`. It defines a block of tests with at least one valid check.
+### Checking at least one condition passes with `describe.one`
+
+With Chef InSpec it is possible to check if at least one of a collection of checks is true. For example: If a setting is configured in two different locations, you may want to test if either configuration A or configuration B have been set. This is accomplished via `describe.one`. It defines a set of `describe` blocks, of which only one needs to pass.
 
 ```ruby
 describe.one do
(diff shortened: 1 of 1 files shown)


docs/dsl_inspec.md
```diff
@@ -73,6 +75,12 @@ describe.one do
 end
 ```
 
+#### `describe.one` usage notes
+
+* InSpec considers a `describe.one` block as passing if a `describe` nested within has all assertions passing. That is, it needs an entire `describe` block to pass and not just single assertions.
```
clintoncwolfe: "not just a single assertion"

clintoncwolfe (changes requested):
Thanks, this is much better! I caught one minor grammo.

docs/dsl_inspec.md
```diff
@@ -73,6 +75,12 @@ describe.one do
 end
 ```
 
+#### `describe.one` usage notes
+
+* InSpec considers a `describe.one` block as passing if a `describe` nested within has all assertions passing. That is, it needs an entire `describe` block to pass and not just a single assertion.
```
stevendanna: It may be the result of too little coffee, but I found this and some of the following sentences hard to follow on first reading.  I think some of it might be the result of the tense of this sentence:

```suggestion
* A `describe.one` block passes if one of its nested `describe` blocks have all assertions passing. That is, it needs an entire `describe` block to pass and not just a single assertion.
```

stevendanna (approved):
Nice improvements.  I left a minor comment on the one bullet point that I found a bit difficult to parse, but feel free to take or leave the suggestion as I'm not sure I've actually made it better.

docs/dsl_inspec.md
james-stocks: Thanks Steven, that is a neater way to write it. I think `has all assertions` is correct because a singular block is the subject.

stevendanna: @james-stocks Yes, sorry, I was playing around with a few different ideas and let that error creep in. `has` is definitely correct here.

clintoncwolfe (approved):
Thanks!

mjingle (approved):

PR #4966, opened 2020-04-02
Revert to Github for fetching ruby devkit

james-stocks:
The S3 bucket URL (https://public-cd-buildkite-cache.s3-us-west-2.amazonaws.com/rubyinstaller-devkit-2.6.5-1-x64.exe) no longer works.

### DIFF
--- a/.expeditor/buildkite/artifact.habitat.test.ps1
+++ b/.expeditor/buildkite/artifact.habitat.test.ps1
@@ -39,8 +39,7 @@ Write-Host "--- Installing $pkg_ident/$pkg_artifact"
 hab pkg install -b $project_root/results/$pkg_artifact
 
 Write-Host "--- Downloading Ruby + DevKit"
-[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-(New-Object System.Net.WebClient).DownloadFile('https://public-cd-buildkite-cache.s3-us-west-2.amazonaws.com/rubyinstaller-devkit-2.6.5-1-x64.exe', 'c:\\rubyinstaller-devkit-2.6.5-1-x64.exe')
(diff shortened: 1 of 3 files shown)


james-stocks:
I've asked #releng-support whether the S3 bucket should still currently be working or if there's another bucket we should be using. 
This PR is a fallback to keep CI rolling.

.expeditor/buildkite/artifact.habitat.test.ps1
```diff
@@ -40,7 +40,7 @@ hab pkg install -b $project_root/results/$pkg_artifact
 
 Write-Host "--- Downloading Ruby + DevKit"
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-(New-Object System.Net.WebClient).DownloadFile('https://public-cd-buildkite-cache.s3-us-west-2.amazonaws.com/rubyinstaller-devkit-2.6.5-1-x64.exe', 'c:\\rubyinstaller-devkit-2.6.5-1-x64.exe')
+(New-Object System.Net.WebClient).DownloadFile('https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.6.5-1/rubyinstaller-devkit-2.6.5-1-x64.exe')
```
stevendanna: I think this function still needs two arguments, the URL and the path on disk you want:

https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient.downloadfile?view=netframework-4.8

james-stocks: This was a mistake, good catch.

james-stocks:
Looking at this test script, it possibly should not be installing Ruby and devkit but instead be running `hab exec $pkg_ident rake` to re-use the Ruby built into the hab package

james-stocks:
Windows tests are failing because this time limited test skip has ended: 

https://github.com/inspec/inspec/blob/master/test/helper.rb#L151

stevendanna (approved):

PR #4967, opened 2020-04-06
Switch to public URL for Ruby devkit for Omnibus

james-stocks:
Omnibus pipelines do not currently have a recommended S3 bucket to use, so switch back to the public URL for Ruby+DevKit for now.

Omnibus AdHoc build showing this working:
https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/117#51b85260-4a0e-4d54-91a9-6900569d5cda

Signed-off-by: James Stocks <jstocks@chef.io>

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [ X] I have read the **CONTRIBUTING** document.

PR #4968, opened 2020-04-06
Add Windows 10 support and fix Omnibus tests to use embedded Ruby

snapp:
## Description

NOTE: This PR conflicts with #4967 as it addresses the current issue installing Ruby during Omnibus testing in a different way.

In addition to adding Windows 10 support, this PR skips the install of Ruby during Windows Omnibus testers in favor of running the tests using the embedded Ruby similar to how Chef Infra Client performs its tests.

An adhoc build showing this working can be found here:

https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/120

## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -31,3 +31,4 @@ builder-to-testers-map:
     - windows-2012r2-x86_64
     - windows-2016-x86_64
     - windows-2019-x86_64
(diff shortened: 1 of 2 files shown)


james-stocks (approved):
This is better than my URL change, thanks very much @christopher-snapp

Issue #3637, opened 2018-11-29
How shall profiles rely on gems?


clintoncwolfe:
### 🎛 Description

🙋 feature request

The InSpec project suggests packaging custom resources in special profiles called [resource packs](https://www.inspec.io/docs/reference/glossary/#resource-pack) which typically contain no controls, but define new controls in the `libraries/` folder.  

Often, those custom controls need (or want) to rely on specific Ruby Gems - not inspec plugins, but simply gems that provide some element of functionality.

How should that be handled?

### 🌍 InSpec and Platform Version

3.0.57

#### Do Nothing

Under this approach, each custom resource says `require 'some-gem'` at the top of the file, and it is up to the user to install the gem in a place InSpec can find it.  That works, but is poor UX.

#### Hide it in a Train plugin

In some cases, the desired library provides both connectivity as well as the object library that your custom resources want to work with.  A clear example of this is the [train-digitalocean](https://github.com/inspec/train-digitalocean) Train plugin and [inspec-digitalocean](https://github.com/inspec/inspec-digitalocean) InSpec resource pack.  The train plugin has a gem dependency in the gemspec on `droplet_kit`; so when you install it, you get droplet_kit as well; then when you run InSpec with a DO target, that gem loads, and the resource pack can use the included classes that represent the DigitalOcean objects.

#### Hide it in an InSpec plugin

If #3636 is implemented, we'll be able to declare that a profile requires a particular plugin.  So, we could have inspec plugins that rely on your needed gem.  That might lead to a proliferation of shell plugins, that only exist to depend on a non-plugin gem.

#### Make an inspec plugin that manages loose gems

We could create a plugin, `inspec-jeweler`, that handles loose gems.  Its gemspec could be dynamic, so that its apparent dependencies would change, depending on what it was told to depend on.  A user could run `inspec jeweler install some-gem` and now it would be installed.  This is probably a bad idea.

#### Add a metadata entry for gems

This is most straightforward to the user, but may have impacts we have not considered.

#### Make profiles be gems

This suggestion came from @zenspider.

@chris-rock @arlimus @jquick any thoughts?

aaronlippold:
Well implementing this capability please keep in mind both the connected and disconnected operation news cases. If a plug-in is required I should be able to vendor it and package all needed dependencies so that I can ship the profile or resource pack for offline operation. Thanks

frezbo:
[...]

PR #4971, opened 2020-04-07
Dist runner

bobchaos:
## Description
This switches from dist constants to use generic terms in runner logs to keep the core gem distro-agnostic.

It also adds (in a separate commit) 3rd party distro support to the install_context helpers

## Related Issue
related issue from cinc: https://gitlab.com/cinc-project/distribution/auditor/-/issues/5
related issue from inspec core: https://github.com/inspec/inspec/issues/3637
related issue from inspec AWS https://github.com/inspec/inspec-aws/issues/166

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) I guess?

## Checklist:

### DIFF
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -8,7 +8,6 @@ require "inspec/profile"
 require "inspec/metadata"
 require "inspec/config"
 require "inspec/dependencies/cache"
-require "inspec/dist"
 require "inspec/reporters"
 require "inspec/runner_rspec"
 # spec requirements
@@ -201,9 +200,9 @@ module Inspec
 
     def supports_profile?(profile)
       unless profile.supports_runtime?
-        raise "This profile requires #{Inspec::Dist::PRODUCT_NAME} version "\
+        raise "This profile requires product version "\
              "#{profile.metadata.inspec_requirement}. You are running "\
-             "#{Inspec::Dist::PRODUCT_NAME} v#{Inspec::VERSION}.\n"
+             "product v#{Inspec::VERSION}.\n"
       end
 
       true
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -200,7 +200,7 @@ module Inspec
 
     def supports_profile?(profile)
       unless profile.supports_runtime?
-        raise "This profile requires product version "\
+        raise "This profile requires the inspec-core gem version "\
              "#{profile.metadata.inspec_requirement}. You are running "\
              "product v#{Inspec::VERSION}.\n"
       end
--- a/lib/inspec/dist.rb
+++ b/lib/inspec/dist.rb
@@ -16,5 +16,21 @@ module Inspec
 
     # name of the compliance product
     COMPLIANCE_PRODUCT_NAME = "Chef Compliance"
+
+    # The suffix for workstation's eponymous folders, like /opt/workstation
+    # or C:/<LEGACY_CONF_DIR>/workstation
+    WORKSTATION_DIR_SUFFIX = "chef-workstation"
+
+    # The suffix for ChefDK's eponymous folders, like /opt/chef-dk
+    # or C:/<LEGACY_CONF_DIR>/chef-dk
+    CHEFDK_DIR_SUFFIX = "chef-dk"
+
+    # The suffix for Inspec's eponymous folders, like /opt/inspec
+    # or C:/<LEGACY_CONF_DIR>/inspec
+    INSPEC_DIR_SUFFIX = "inspec"
+
+    # The legacy conf folder: C:/opscode/chef*. Specifically the "opscode" part
+    # relevant *_DIR_SUFFIX is appended to it in code when needed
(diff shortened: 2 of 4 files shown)


bobchaos:
it just occurred to me "inspec-core rubygem" may be better generic wording than "product", which is so generic it may lead to confusion. thoughts?

lib/inspec/runner.rb
```diff
@@ -201,9 +200,9 @@ def add_target(target, _opts = [])
 
     def supports_profile?(profile)
       unless profile.supports_runtime?
-        raise "This profile requires #{Inspec::Dist::PRODUCT_NAME} version "\
+        raise "This profile requires product version "\
```
clintoncwolfe: I think this phrasing will lead to confusion - when a user hits this error, they need to know specifically what component has the faulty version (workstation? knife? kitchen? audit cookbook? inspec/cinc?). It seems to me that the product name needs to remain a constant reference here. 

I'd like to understand why that is problematic for the cinc project, and look for alternatives.

bobchaos: the runner is used in the core gem, which is in turn used in other 3rd party products like kitchen-terraform. When the runner executes in that context, it outputs a Chef wordmark. Changing it for a generic word ensures compliance of 3rd party projects that use inspec-core as a library.

I reformulated to "the inspec-core gem" for clarity but I'm not sure it's accurate.

bobchaos: Is there a way to detect if it's executing in the context of Chef binstubs maybe?

lib/inspec/runner.rb
```diff
@@ -201,9 +200,9 @@ def add_target(target, _opts = [])
 
     def supports_profile?(profile)
       unless profile.supports_runtime?
-        raise "This profile requires #{Inspec::Dist::PRODUCT_NAME} version "\
+        raise "This profile requires the inspec-core gem version "\
```
james-stocks, line 203: InSpec should be allowed to refer to itself by product name. It would be better to tell the user that they need "Chef InSpec" or the community package at a particular version; as that's the item they will need to download. Referring to the gem might mislead the user that they need to adjust gems in their current product installation.

More generally Chef InSpec should be allowed to refer to itself anywhere in its code. It shouldn't need to "unbrand". Is it possible for the community distribution to manage a modified version of `dist.rb` so that `Inspec::Dist::PRODUCT_NAME` and the other dist consts refer to the community distribution names?

I'm available on the community slack if I'm missing some of the context here and you'd like to discuss

bobchaos: Lemme try to reformulate the problem: I use kitchen-terraform, which uses the Inspec lib under the hood. That line i'm changing is the _only_ direct reference to Chef Inspec in the library proper.

 Removing it means tools like kitchen-terraform are trademark compliant. Leaving it in puts a lot of the community in a really weird place; Do projects like Cinc need to redistribute kitchen-terraform to make it compliant? Or does the burden fall on the authors to publish and use edited inspec-core gems? Should we have 2 versions of every such product, one with Inspec and one with a modified lib?

As an alternative to this PR, perhaps Chef Legal could issue a statement here or somewhere else official stating that this particular occurence of "Chef Inspec" is cool with them (It's a pretty discreet one, all things considered). Something I can reference when folks raise this particular issue.

What's your handle on Slack? We can pick this up over there this week sometime if this still doesn't convey the perceived issue correctly.

james-stocks: Hi @bobchaos -

I'm `Stocksy` in the Chef community Slack. We can ask Chef Legal about this , but I wanted to clarify:
Does Cinc include kitchen-terraform in its distribution? Or do Cinc users download and install kitchen-terraform separately from Cinc?
I believe you mean the latter; but we wanted to confirm before asking Legal.

Thanks

bobchaos: It is indeed the latter scenario at this time; we're asking because, as the Cinc Project, we want to know if we should align on the former in the future (or find someone willing to do that, whether or not kitchen-terraform is in the project's scope is up for debate)

You could say I'm asking more as a service to the community than as a directly interested party

james-stocks: Hi @bobchaos - 

Sorry for slow response.

I asked Chef Legal about Chef trademarks appearing in CINC logging due to kitchen-terraform being added to the CINC distribution. I got this statement:

> CINC (or any distribution) is permitted to include/reference 3rd party components that ultimately depend on Chef's original components. This is not a trademark violation, just a side effect of a highly connected OSS ecosystem.

I hope this statement is useful.

Regarding this PR specifically, I think it's a good idea to remove hardcoded distribution details from inspec-core, but they should still be accessible to the runner class. Clinton made a good point that when the InSpec run is in the middle of your full automation stack, it's super unhelpful for error messages to not hint at the component that raised the error. And Chef InSpec should be able to continue to refer to itself as Chef InSpec.

bobchaos: This 100% resolves the issue as far as I'm concerned, thanks! I'ma go ahead and close this.

PR #4975, opened 2020-04-08
Use Expeditor's built in Bash helper to post release announcement

schisamo:
The Discourse API has recently changed its authentication method and the underlying issue was fixed in Expeditor's `post_discourse_release_announcement` helper function: https://github.com/chef/expeditor/pull/1240


james-stocks (approved):
The bash helper makes it a lot simpler for us, and I see that we now fail if the curl POST fails, thanks very much @schisamo !

PR #4980, opened 2020-04-09
Fixed inputs example to display correct output when the same code is …

paul1994:
…ran.

updated output example

Signed-off-by: Paul Bradford <pbradford@chef.io>

Fixed input documentation example
## Description

The current documentation for inputs displays the wrong `expected` and `got` values causing confusion.
Also the `eq` comparison in the code does not return a proper output for a successful control check. I changed the code to use `cmp` so that the test will run and be successful as designed.
## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #4992, opened 2020-04-21
Bumps method_source dependency version

Schwad:
The `method_source` dependency for inspec-core has not been bumped in
five years, and it is indirectly holding `pry` back elsewhere in the
ecosystem which is causing issues for ruby 2.7+ support.

## Related Issue

This was brought to our attention by @tas50 , I do not have references for related issues but pertinent history can be found here:

* https://github.com/inspec/inspec/commit/f6509b7f8136d7cfea8270b3909f60d6bbe9a91f
* https://github.com/banister/method_source/commits/master

## Types of changes

## Checklist:

### DIFF
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "license-acceptance", ">= 0.2.13", "< 2.0"
   spec.add_dependency "thor",               ">= 0.20", "< 2.0"
   spec.add_dependency "json_schemer",       "~> 0.2.1"
-  spec.add_dependency "method_source",      "~> 1.0"
+  spec.add_dependency "method_source",      ">= 0.8", "< 2.0"
   spec.add_dependency "rubyzip",            "~> 1.2", ">= 1.2.2"
   spec.add_dependency "rspec",              "~> 3.9"
   spec.add_dependency "rspec-its",          "~> 1.2"
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "rubyzip",            "~> 1.2", ">= 1.2.2"
   spec.add_dependency "rspec",              "~> 3.9"
   spec.add_dependency "rspec-its",          "~> 1.2"
-  spec.add_dependency "pry",                "~> 0"
+  spec.add_dependency "pry",                "~> 0.13"
   spec.add_dependency "hashie",             "~> 3.4"
   spec.add_dependency "mixlib-log",         "~> 3.0"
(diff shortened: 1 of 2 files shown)


inspec-core.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "license-acceptance", ">= 0.2.13", "< 2.0"
   spec.add_dependency "thor",               ">= 0.20", "< 2.0"
   spec.add_dependency "json_schemer",       "~> 0.2.1"
-  spec.add_dependency "method_source",      "~> 0.8"
```
tas50, line 30: you might want to make this >= 0.8 and < 2 so we can pull in any known working version

clintoncwolfe:
913dcac should get `functional/inspec_shell_test.rb` green again; however we have a user-visible deprecation warning when running `inspec shell`:
```
warning: setting prompt with help of `Pry.config.prompt = [proc {}, proc {}]` is deprecated. Use Pry::Prompt API instead
```

which comes from [here](https://github.com/inspec/inspec/blob/a3ea5acb3f38e6506cd1755a4ea1bd0523d4a533/lib/inspec/shell.rb#L40) - if I have time I'll poke at that some more.

UPDATE: I've made a push on ad75691 that fixes the deprecation warning.

Schwad:
@clintoncwolfe this looks great - thanks for solving the `Pry::Terminal` bug and also for going the extra mile on the `Pry.config.prompt` deprecation (and version tweak)! 👍  :)

lib/inspec/shell.rb
```diff
@@ -36,8 +36,13 @@ def configure_pry # rubocop:disable Metrics/AbcSize
       end
 
       # configure pry shell prompt
-      Pry.config.prompt_name = "inspec"
-      Pry.prompt = [proc { "#{readline_ignore("\e[1m\e[32m")}#{Pry.config.prompt_name}> #{readline_ignore("\e[0m")}" }]
+      Pry::Prompt.add(
```
james-stocks, line 39: The previous line was causing a deprecation warning. Does this new line work with the lowest `">= 0.8"` constraint for method_source ?

Schwad: Nice catch. It appears the `#add` API was added in `pry` 0.13.1 https://github.com/pry/pry/commit/3da930f908ff73f7638bd0471c25d47c28320df3

stevendanna (approved):
Everything looks good when I run this locally.  Thanks for sorting this out.

lib/inspec/shell.rb
james-stocks: And we now require pry 0.13, cool

james-stocks (approved):
Awesome, thanks @Schwad

PR #4994, opened 2020-04-24
Reporter options to truncate message and suppress stacktrace

james-stocks:

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -360,6 +360,13 @@ module Inspec
       end
 
       raise ArgumentError, "The option --reporter can only have a single report outputting to stdout." if stdout_reporters > 1
+
+      # reporter_message_truncation needs to either be the string "ALL", an Integer, or a string representing an integer
+      if (truncation = @merged_options["reporter_message_truncation"])
+        unless truncation == "ALL" || truncation.is_a?(Integer) || truncation.to_i.to_s == truncation
+          raise ArgumentError, "reporter_message_truncation is set to #{truncation}. It must be set to an integer value or ALL to indicate no truncation."
+        end
+      end
     end
 
     def validate_plugins!
--- a/lib/inspec/reporters/base.rb
+++ b/lib/inspec/reporters/base.rb
@@ -5,22 +5,17 @@ module Inspec::Reporters
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
-      apply_options_to_run_data
+      apply_report_resize_options unless @run_data.nil?
       @output = ""
     end
 
     # Apply options such as message truncation and removal of backtraces
-    def apply_options_to_run_data
-      message_truncation = @config[:message_truncation] || "ALL"
-      trunc = -1
-      if message_truncation != "ALL"
-        if message_truncation.to_i.to_s != message_truncation
-          Inspec::Log.warn("Messages will not be truncated because #{message_truncation} is not an integer value")
-        else
-          trunc = message_truncation.to_i
-        end
-      end
-      include_backtrace = @config[:include_backtrace].nil? ? true : @config[:include_backtrace]
+    def apply_report_resize_options
+      runtime_config = Inspec::Config.cached.respond_to?(:final_options) ? Inspec::Config.cached.final_options : {}
+
+      message_truncation = runtime_config[:reporter_message_truncation] || "ALL"
+      trunc = message_truncation == "ALL" ? -1 : message_truncation.to_i
+      include_backtrace = runtime_config[:reporter_backtrace_inclusion].nil? ? true : runtime_config[:reporter_backtrace_inclusion]
 
       @run_data[:profiles]&.each do |p|
         p[:controls].each do |c|
--- a/lib/inspec/reporters/base.rb
+++ b/lib/inspec/reporters/base.rb
@@ -5,9 +5,34 @@ module Inspec::Reporters
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
-      @message_truncation = config[:message_truncation] || 'ALL'
-      @include_backtrace = config[:include_backtrace].nil? ? true : config[:include_backtrace]
+      apply_options_to_run_data
       @output = ""
+    end
+
+    # Apply options such as message truncation and removal of backtraces
+    def apply_options_to_run_data
+      message_truncation = @config[:message_truncation] || "ALL"
+      trunc = -1
+      if message_truncation != "ALL"
+        if message_truncation.to_i.to_s != message_truncation
+          Inspec::Log.warn("Messages will not be truncated because #{message_truncation} is not an integer value")
+        else
+          trunc = message_truncation.to_i
+        end
+      end
+      include_backtrace = @config[:include_backtrace].nil? ? true : @config[:include_backtrace]
+
+      @run_data[:profiles]&.each do |p|
+        p[:controls].each do |c|
+          c[:results]&.map! do |r|
+            r.delete(:backtrace) unless include_backtrace
+            if r.key?(:message) && r[:message] != "" && trunc > -1
+              r[:message] = r[:message][0...trunc] + "[Truncated to #{trunc} characters]"
+            end
+            r
+          end
+        end
+      end
     end
 
     def output(str, newline = true)
--- a/lib/inspec/reporters/base.rb
+++ b/lib/inspec/reporters/base.rb
@@ -5,6 +5,8 @@ module Inspec::Reporters
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
+      @message_truncation = config[:message_truncation] || 'ALL'
+      @include_backtrace = config[:include_backtrace].nil? ? true : config[:include_backtrace]
       @output = ""
     end
 
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -134,6 +134,8 @@ module Inspec
       return if @conf["reporter"].nil?
 
       @conf["reporter"].each do |reporter|
+        reporter[1][:include_backtrace] = @conf["reporter_backtrace_inclusion"] if @conf.key?(:reporter_backtrace_inclusion)
+        reporter[1][:message_truncation] = @conf["reporter_message_truncation"] if @conf.key?(:reporter_message_truncation)
         result = Inspec::Reporters.render(reporter, run_data)
         raise Inspec::ReporterError, "Error generating reporter '#{reporter[0]}'" if result == false
       end
--- a/lib/inspec/runner.rb
+++ b/lib/inspec/runner.rb
@@ -134,8 +134,6 @@ module Inspec
       return if @conf["reporter"].nil?
 
       @conf["reporter"].each do |reporter|
-        reporter[1][:include_backtrace] = @conf["reporter_backtrace_inclusion"] if @conf.key?(:reporter_backtrace_inclusion)
-        reporter[1][:message_truncation] = @conf["reporter_message_truncation"] if @conf.key?(:reporter_message_truncation)
         result = Inspec::Reporters.render(reporter, run_data)
         raise Inspec::ReporterError, "Error generating reporter '#{reporter[0]}'" if result == false
       end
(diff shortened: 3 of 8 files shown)


lib/inspec/runner.rb
```diff
@@ -134,7 +134,18 @@ def render_output(run_data)
       return if @conf["reporter"].nil?
 
       @conf["reporter"].each do |reporter|
-        result = Inspec::Reporters.render(reporter, run_data)
+        reporter_opts = {}
```
james-stocks: Optionally the `run_data` could be directly modified here instead of passing options to the reporters. But I think there should be a distinction between the data that actually happened and the data that's reported

lib/inspec/reporters.rb
```diff
@@ -9,26 +9,26 @@
 
 module Inspec::Reporters
   # rubocop:disable Metrics/CyclomaticComplexity
-  def self.render(reporter, run_data)
+  def self.render(reporter, run_data, opts = {})
```
Schwad: I think what's interesting here, is wouldn't this setup give us the power to use options on any of our reporters in the future? (With the first two being these specific ones relating to JSON)

Though I do see where we could be increasing data coupling (passing data to the reporters) -  it is with a single options hash that opens up these future feature possibilities.

james-stocks: At this point, the `reporter` passed to this method is `["json", {"stdout"=>true}]`
That might be an existing hash that we can and should use to pass options for the reporter

lib/inspec/runner.rb
```diff
@@ -134,7 +134,18 @@ def render_output(run_data)
...
+        if @conf.key?(:reporter_backtrace_inclusion)
+          reporter_opts[:include_backtrace] = @conf['reporter_backtrace_inclusion']
+        elsif reporter[0] == 'json-automate'
+          reporter_opts[:include_backtrace] = false
+        end
+        if @conf.key?(:reporter_message_truncation)
+          reporter_opts[:message_truncation] = @conf['reporter_message_truncation']
+        elsif reporter[0] == 'json-automate'
+          reporter_opts[:message_truncation] = '200'
```
james-stocks: From chatting to @kekaichinose the default for Automate should maybe be to not truncate. On the principle that for compliance, all data should be reported. 
Then Automate users will need to opt in for the reduced report size.

alexpop:
Tried running inspec from this branch but getting this error:

```bash
[06:20:24 ~/git/inspec-stocksy (js/json)]$ bundle exec inspec exec ~/git/mycompliance-profile/myprofile1 --reporter json-automate
wrong number of arguments (given 2, expected 1)
```

If I switch to the system inspec, same profile runs:
```bash
[06:20:56 ~/git/inspec-stocksy (js/json)]$ inspec exec ~/git/mycompliance-profile/myprofile1 --reporter json-automate
{"platform":{"name":"mac_os_x","release":"19.4.0"},"profiles":[ ...
```

james-stocks:
@alexpop @Schwad I changed the default behaviour to not change the output at all by default, and added some tests.
The remaining bit I need to do is either rename the new options to refer to JSON, or implement the option in all other reporters.

Schwad:
Sorry I just saw the ready for review marking - I'll have a nice fresh look at this Monday if that works? Really excited for this feature! :)

alexpop:
Thanks @james-stocks 

I tested the change and I can see it working if I run inspec like this:
```
[05:59:54 ~/git/inspec-stocksy (js/json)]$ bundle exec inspec exec ~/git/mycompliance-profile/myprofile1 --reporter_message_truncation 100 --reporter_backtrace_inclusion false --reporter json-automate
```

But passing the options via json-config does not truncate:
```
[05:59:54 ~/git/inspec-stocksy (js/json)]$ echo '{"reporter_backtrace_inclusion": false, "reporter_message_truncation": 100, "reporter":["json-automate"], "backend_cache":true, "chef_license": "accept"}' | bundle exec inspec exec ~/git/mycompliance-profile/myprofile1 --json-config=-
```

lib/inspec/runner.rb
```diff
@@ -134,6 +134,8 @@ def render_output(run_data)
       return if @conf["reporter"].nil?
 
       @conf["reporter"].each do |reporter|
+        reporter[1][:include_backtrace] = @conf["reporter_backtrace_inclusion"] if @conf.key?(:reporter_backtrace_inclusion)
```
clintoncwolfe: This seems odd on first reading - check for a Symbol key, then set based on a String key?

clintoncwolfe: These options are global to the run, not per-reporter. Inspec::Config (that is, `@conf`) is global, too - no need to make changes to `runner.rb` at all. You could just call it directly in `reporters/base.rb`

lib/inspec/reporters/base.rb
```diff
@@ -5,9 +5,36 @@ class Base
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
+      apply_options_to_run_data
       @output = ""
     end
 
+    # Apply options such as message truncation and removal of backtraces
+    def apply_options_to_run_data
+      message_truncation = @config[:message_truncation] || "ALL"
```
clintoncwolfe: Right here, you could call `runtime_config = Inspec::Config.cached` and get an instance of  the runtime config with access to the CLI options or JSON config - no need to sneak it into the per-reporter configs.

lib/inspec/reporters/base.rb
```diff
@@ -5,9 +5,36 @@ class Base
...
+      trunc = -1
+      if message_truncation != "ALL"
+        if message_truncation.to_i.to_s != message_truncation
+          Inspec::Log.warn("Messages will not be truncated because #{message_truncation} is not an integer value")
```
clintoncwolfe: Consider making this an error - if they provided the option but it is unusable, halt and show usage of the bad option is a reasonable thing to do. Continuing and generating too much data might be worse.

lib/inspec/reporters/base.rb
```diff
@@ -5,9 +5,36 @@ class Base
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
+      apply_options_to_run_data
```
clintoncwolfe: I know naming things is hard and no one will ever agree, but perhaps "apply_message_size_limits"? If there are future options to implement, they would go in their own methods.

clintoncwolfe (commented):
The algorithm seems fine, main thing I note is there is no need stash the CLI options on the per-reporter configs. Few other suggestions, most minor.

lib/inspec/reporters/base.rb
james-stocks: I was foreseeing the opposite - that this method _would_ handle any future options. I can rename the method like you suggest so that it is named specific to the thing it _currently_ does, and we can discuss handling future options when we get to that point.

lib/inspec/reporters/base.rb
james-stocks: Thanks, that sounds good! I think that solves the problem Alex mentioned where these options are not working from STDIN config.
It would make it a little odd that the class is instantiated with a `config` hash but we don't use that for config - but it does mean I can revert my changes to the runner.

lib/inspec/reporters/base.rb
james-stocks: My thinking for making this a warning is that:
 * The CLI, not this class, is responsible for config validation (even though it's not doing any currently...). 
 * At this point, the run has completed. It possibly took a long time and burned through billable compute resource. The reporter should generally never halt.

Possibly I could make the option strictly numeric and throw away the `ALL` special case - I did it this way so that those using InSpec in CI could have a variable driven run command where `message-truncation` is always set at the CLI and the value is defined by a CI variable.

lib/inspec/runner.rb
james-stocks: I pushed another commit that leaves runner.rb unchanged

lib/inspec/reporters/base.rb
clintoncwolfe: Good point about the validation happening post-run - perhaps the validation should be moved to be here - https://github.com/inspec/inspec/blob/2e887a94afcca819da781d4774aa2a5a0b56785e/lib/inspec/config.rb#L328

james-stocks: Thanks, that looks like the right spot for validation.
This warning was meant to supplementary to proper validation - but I think if there's validation in `config.rb` then this class also doing a check is overkill - so I'll just remove this warning.

clintoncwolfe (approved):
This looks good, thanks!

lib/inspec/reporters/base.rb
```diff
@@ -5,9 +5,31 @@ class Base
     def initialize(config)
       @config = config
       @run_data = config[:run_data]
+      apply_report_resize_options unless @run_data.nil?
       @output = ""
     end
 
+    # Apply options such as message truncation and removal of backtraces
+    def apply_report_resize_options
+      runtime_config = Inspec::Config.cached.respond_to?(:final_options) ? Inspec::Config.cached.final_options : {}
```
Schwad, line 14: This is good to know for future

lib/inspec/reporters/base.rb
```diff
@@ -5,9 +5,31 @@ class Base
...
+
+      message_truncation = runtime_config[:reporter_message_truncation] || "ALL"
+      trunc = message_truncation == "ALL" ? -1 : message_truncation.to_i
+      include_backtrace = runtime_config[:reporter_backtrace_inclusion].nil? ? true : runtime_config[:reporter_backtrace_inclusion]
+
+      @run_data[:profiles]&.each do |p|
```
Schwad, line 20: I had a quick look with `benchmark/ips` in the Terminal with an identical data structure to see if I could find another way to make this change without the loops but as far as I could tell this is the best approach!

Schwad (approved):
Looking great, thanks! 👍

lib/inspec/config.rb
```diff
@@ -360,6 +360,12 @@ def validate_reporters!(reporters)
       end
 
       raise ArgumentError, "The option --reporter can only have a single report outputting to stdout." if stdout_reporters > 1
+
+      if (truncation = @merged_options["reporter_message_truncation"])
+        unless truncation == "ALL" || truncation.to_i.to_s == truncation
```
james-stocks: This is wrong if the value is already an Integer. Usually it will be a string (via the CLI) but can be an Integer via JSON config.
@clintoncwolfe 's PR for typing inputs will also cause this to be wrong, as that PR causes inputs to be typed where possible

Schwad (changes requested):
Look to handle actual integer via JSON:

````
echo '{"reporter_backtrace_inclusion": false, "reporter_message_truncation": 100, "reporter":["json-automate"]}' | be inspec exec ~/git/mycompliance-profile/myprofile1 --json-config=-
#=> reporter_message_truncation is set to 100. It must be set to an integer value or ALL to indicate no truncation.
```

lib/inspec/reporters/base.rb
james-stocks: Yes I don't much like these nested loops.
Note the cautious  `&.` because at each level the report might be missing any occurrences of the nested elements.

james-stocks:
Please don't merge this before @alexpop approves

lib/inspec/config.rb
clintoncwolfe: #5011 will only impact the `--input` CLI option

alexpop:
Thank you everybody, great team effort!
```
$ echo '{"reporter_backtrace_inclusion": false, "reporter_message_truncation": 100, "reporter":["json-automate"]}' | bundle exec inspec exec ~/git/mycompliance-profile/myprofile1 --json-config=- | jq .
...
          "waiver_data": {},
          "results": [
            {
              "status": "failed",
              "code_desc": "Command: `ps aux` stdout is expected to include \"No packages needed for security\"",
              "run_time": 0.70555,
              "start_time": "2020-05-06T22:07:55+01:00",
              "message": "expected \"USER               PID  %CPU %MEM      VSZ    RSS   TT  STAT STARTED      TIME COMMAND\\nap[Truncated to 100 characters]"
            }
          ]
        }
      ],
      "status": "loaded"
    }
  ],
  "statistics": {
    "duration": 0.711616
  },
  "version": "4.18.109"
}
```

Same result with:
```
bundle exec inspec exec ~/git/mycompliance-profile/myprofile1 --reporter_message_truncation 100 --reporter json-automate | jq .
```

alexpop (approved):
Cheers 🚀 🚀 🚀

Schwad (approved):

lib/inspec/config.rb
james-stocks: D'oh, good point!

PR #4998, opened 2020-04-28
(WIP) Re-enable Windows tests

james-stocks:
Initially just enable Windows tests again to see what fails.


Schwad:
To enable tests again would changing the method to:

`def skip_windows!; end` work?

Or deleting all the `skip_windows!` calls in the tests etc. 

It'd be really interesting to see if any of those 74 tests *do* actually end up passing

Issue #4936, opened 2020-03-02
Namespace clash between control name and input


lhasadreams:
InSpec V4.16.0 and before execute a control without error
Raised on behalf of Paul Burnet (Lloyds Banking Group)

# Describe the problem
This control use to work fine up to InSpec v4.16.0, using the latest InSpec (v4.18.97) gives an error.
```
control 'my_control' do
only_if {input('my_control', value: 'false') == 'false' }
 <ommitted>
end
```

When executing the control InSpec gives the following error:
```
 ×  Control Source Code Error ./controls/IAM/iam_benchmark_1_57_27_58_cloudsql_mysql_restrict_stop_delete.rb:7
     string not matched
```

This can be traced to a namespace error between the control name and the inout name - this a regression.
Changing the control to:
```
control 'my_control' do
only_if {input('skip_my_control', value: 'false') == 'false' }
 <ommitted>
end
```
Removes the error.

The latest version of InSpec should execute the above control without error. The customer doesn't really want to change the code to work around the regression.

## Possible Solution
(1) Fix the code regression, @clintoncwolfe and @TheLunaticScripter indicated that it may be a side effect of the Waiver work.
(2) Pin to the earlier version of InSpec - not an option as later functionality required.
(3) Change all the input names - customer would prefer not to.
(3) Change completely and use Waivers instead - I am testing this with the customer, however Waivers are beta at the moment and InSpec is being used in a security critical area.

Schwad:
I can confirm that I've run the tests/output laid out above and replicated the error. 

```
control 'my_control' do
only_if {input('my_control', value: 'false') == 'false' }
 <ommitted>
end
```

Fails current inspec, passes 4.16.0. Will try and find the breaking commit 👍

Schwad:
Having reviewed our history here, confirmed that the commit that causes the above syntax to fail is: https://github.com/inspec/inspec/commit/43bc7e034387eccaff1b889c02dc7c57aae33909

To reproduce, cd into local copy of repo and run:

```
git checkout 43bc7e034387eccaff1b889c02dc7c57aae33909
bundle install
bundle exec inspec-bin/bin/inspec exec path/to/profile/with/above/control -t <my_target>
#=> ×  Control Source Code Error schwad_sandbox/controls/example.rb:22 
     undefined method `key?' for "false":String
```

Then go to prior commit

```
git checkout aab45b507fe625d81a2104f6f9b9362471e566ed
rm Gemfile.lock
bundle install
[...]

PR #4995, opened 2020-04-24
Fix for warning when input is provided

clintoncwolfe:
## Description

Fixes some cases in which a profile may require inputs in its metadata but not provide a default value (a reasonable thing to do), causing InSpec to emit noisy warnings like:

```
[2020-02-07T16:33:50-05:00] WARN: Input 'hostname' does not have a value. Use --input-file to provide a value for 'hostname' or specify a value with attribute('hostname', value: 'somevalue', ...).
```

These warnings would be emitted even if the value was provided by `--input-file`.

The cause was traced to the validation routines requesting a value; the act of requesting a value would cause a warning to be emitted if one was not known. A parameter was added to make the check silent if it was only checking for a value, not trying to use one.

Additionally, the message emitted referred to inputs as attributes, which is deprecated; the message was updated.

A new test fixture was created, along with a pair of functional tests.

## Related Issue

Fixes #4769
Fixes #4445

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.

PR #5001, opened 2020-04-29
Allows input and control to have the same name

Schwad:
# Background

In https://github.com/inspec/inspec/issues/4936 the issue was reported that naming an input the same as a control caused an unexpected failure.

In that particular case, the naming was a result of a pre-waivers workaround which is no longer necessary, but ultimately a breakage of that name clash is an unexpected occurrance.

# The error

Due to how inputs are named and registered, `__apply_waivers` thinks that an object is a waiver that is not a waiver and tries to process it. On the micro level, it breaks when trying to pass a variable to a string as if it were a Hash.

# The fix

It is imperative that we preserve 100% of the current featureset, pass our tests, and fix this edge case along with new test coverage for the failure.

This PR updates the code to do a small ‘waiver check’ to stop the namespace clash from breaking the code.

## Related Issue

#4936 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -332,7 +332,7 @@ module Inspec
       input_name = @__rule_id # TODO: control ID slugging
       registry = Inspec::InputRegistry.instance
       input = registry.inputs_by_profile.dig(__profile_id, input_name)
-      return unless input
+      return unless input && input.has_value? && input.value.is_a?(Hash)
 
       # An InSpec Input is a datastructure that tracks a profile parameter
       # over time. Its value can be set by many sources, and it keeps a
--- a/test/functional/waivers_test.rb
+++ b/test/functional/waivers_test.rb
@@ -10,6 +10,23 @@ describe "waivers" do
   let(:controls_by_id)        { run_result; @json.dig("profiles", 0, "controls").map { |c| [c["id"], c] }.to_h }
   let(:cmd)                   { "exec #{waivers_profiles_path}/#{profile_name} --input-file #{waivers_profiles_path}/#{profile_name}/files/#{waiver_file}" }
 
+  attr_accessor :out
+
+  def inspec(commandline, prefix = nil)
+    @stdout = @stderr = nil
+    self.out = super
+  end
+
+  def stdout
+    @stdout ||= out.stdout
+      .force_encoding(Encoding::UTF_8)
+  end
+
+  def stderr
+    @stderr ||= out.stderr
+      .force_encoding(Encoding::UTF_8)
+  end
+
   def assert_test_outcome(expected, control_id)
     assert_equal expected, controls_by_id.dig(control_id, "results", 0, "status")
   end
@@ -85,6 +102,19 @@ describe "waivers" do
           refute_waiver_annotation control_id
         end
       end
+    end
+  end
+
+  describe "an input and control with the same name" do
+    # This is a test for a regression articulated here:
+    # https://github.com/inspec/inspec/issues/4936
+    it "can execute when control namespace clashes with input" do
+      inspec("exec " + "#{waivers_profiles_path}/namespace-clash" + " --no-create-lockfile" + " --no-color")
+
+      _(stdout).wont_include("Control Source Code Error")
+      _(stdout).must_include "\nProfile Summary: 1 successful control, 0 control failures, 0 controls skipped\n"
+      _(stderr).must_equal ""
+      assert_exit_code 0, out
     end
   end
 
--- a/test/fixtures/profiles/waivers/namespace-clash/controls/namespace_clash.rb
+++ b/test/fixtures/profiles/waivers/namespace-clash/controls/namespace_clash.rb
@@ -1,0 +1,7 @@
+# This fixture tests for a regression found here: https://github.com/inspec/inspec/issues/4936
+control '01_my_control' do
+only_if { input('01_my_control', value: 'false') == 'false' }
+  describe true do
+    it { should eq true }
+  end
+end
--- a/test/fixtures/profiles/waivers/namespace-clash/inspec.yml
+++ b/test/fixtures/profiles/waivers/namespace-clash/inspec.yml
@@ -1,0 +1,5 @@
+name: namespace-clash
+summary: Verifies input and control namespace can safely clash
+version: 0.1.0
+supports:
+  platform: os
--- a/test/functional/inspec_exec_json_test.rb
+++ b/test/functional/inspec_exec_json_test.rb
@@ -176,8 +176,8 @@ describe "inspec exec with json formatter" do
         "summary" => "Demonstrates the use of InSpec Compliance Profile",
         "version" => "1.0.0",
         "supports" => [{ "platform-family" => "unix" }, { "platform-family" => "windows" }],
-        "status" => "loaded",
         "attributes" => [],
+        "status" => "loaded",
       })
 
       _(groups.sort_by { |x| x["id"] }).must_equal([
--- a/test/functional/inspec_exec_junit_test.rb
+++ b/test/functional/inspec_exec_junit_test.rb
@@ -42,11 +42,11 @@ describe "inspec exec with junit formatter" do
     describe "the test suite" do
       let(:suite) { doc.elements.to_a("//testsuites/testsuite").first }
 
-      it "must have 6 testcase children" do
+      it "must have 4 testcase children" do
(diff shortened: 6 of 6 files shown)


lib/inspec/rule.rb
```diff
@@ -373,6 +373,10 @@ def __apply_waivers
       __waiver_data["skipped_due_to_waiver"] = true
     end
 
+    def is_waiver?(input)
+      input && input.value.is_a?(Hash)
```
james-stocks: I've not seen it in action but I think it's possible for an input to be a hash (see https://github.com/inspec/inspec/blob/master/lib/inspec/input.rb#L155 )

This PR makes things a little better as it lets users continue to use the control name as an input name as long as they aren't trying to use a hash input.

I think this can only be fixed fully by having InSpec load waivers separately from inputs.

Schwad: I think that's a really good point and we would be warranted to file a separate larger issue for waivers/input loading. 

NOTE: `value` is an attribute (well, actually a method but you know what I mean) on input. So each input has a `value` piece of information you can access. Which is sometimes a string, hash, etc. However, when the input is a waiver, it's *always* a Hash for value. (As __apply_waivers interacts with it like a hash)

The name clash has an unwanted side affect of passing through an input and breaking on this line: https://github.com/inspec/inspec/blob/0d44b0e2818a5ff595bcc2bd9e166139bd0e4d6f/lib/inspec/rule.rb#L343 (because it assumes a Hash for `value` and non-waivers have strings on Value sometimes)

This four-liner blocks non-waiver inputs from trying to be applied. 

I do agree that longer term I would like some more separation, even if there was a type on inputs so I could just call something like `is_waiver?` *directly* on the input :)

I believe this does work for the customer though

Schwad: To make it more straightforward I could pull this method out and we could change the line

`return unless input` (original)

to:

`return unless input && input.value.is_a?(Hash)` 

Which is really what I'm trying to accomplish here. (I just like to name compound conditionals)

stevendanna: I think you clearly have investigated this code path more than I have.  

I too typically like to name compound conditionals, but in this case it could introduce some confusion, so let's avoid it.

Naively, I think my order of preference would be:

1) Change waivers to be a specific class we can look for and then look for it here
2) Change the waivers hash to have some sentinel `_is_waiver` key that we can look for if we have a hash here.
3) Use `return unless input && input.value.is_a?(Hash)` with a comments about the situation.

I'm completely cool with (3) for now if (1) and (2) are a bit more than we want to chew off to get this fixed or if they are not good ideas for some other reason.

Schwad: Thank you @stevendanna and @james-stocks for looking over this and your (very helpful) feedback. 

Now that I've had a think on it with this feedback, I think I might revert to no.3, and if it's okay file an issue to move forward on 1&2?

Perhaps working on this issue has helped us uncover the need to class-ify or otherwise identify waivers in the wild which I definitely agree would have made this easier to debug! :) 

If it passes and everyone's happy then I could write up that issue tomorrow, let folks in 4936 know, and squash. (also doing a few small spec updates)

james-stocks: Sounds good @Schwad - I think we want to do 1 or 2 at some point because currently inputs and waivers aren't very extensible

test/fixtures/profiles/old-examples/profile/controls/example-namespace-clash.rb
```diff
@@ -0,0 +1,8 @@
+title "Namespace clashing control"
+
+control 'my_control' do
+only_if { input('my_control', value: 'false') == 'false' }
+  describe file("/tmp") do
+    it { should be_directory }
+  end
+end
```
clintoncwolfe: I'd suggest avoid making any changes to the old-examples fixtures - those are a relic from when user examples were used as test fixtures. This might be better suited under `fixtures/profiles/waivers` or `fixtures/profiles/inputs` . If you need to make a new profile, use `inspec init profile`. 

Placing it here impacts a lot of tightly coupled things like test counts, which could be avoided.

test/functional/inspec_exec_test.rb
```diff
@@ -214,6 +214,15 @@ def stderr
...
   end
 
+  it "can execute when control namespace clashes with input" do
+    inspec("exec " + namespace_clash_control + " --no-create-lockfile")
+
+    _(stdout).wont_include("Control Source Code Error")
+    _(stdout).must_include "\nProfile Summary: 1 successful control, 0 control failures, 0 controls skipped\n"
+    _(stderr).must_equal ""
+    assert_exit_code 0, out
+  end
```
clintoncwolfe: This might be better in test/functional/waivers_test, as it is a fix for a regression related to the waivers system. As it is a regression test might also add a link to the issue.

lib/inspec/rule.rb
```diff
@@ -332,7 +332,7 @@ def __apply_waivers
       input_name = @__rule_id # TODO: control ID slugging
       registry = Inspec::InputRegistry.instance
       input = registry.inputs_by_profile.dig(__profile_id, input_name)
-      return unless input
+      return unless input && input.value.is_a?(Hash)
```
clintoncwolfe: I can confirm that inputs can indeed have Hash values; but here they'd have to have a hash-valued Input with the same name as a control, which is increasingly unlikely. Has anyone asked the affected users if they use Hashes?

One note, input.value may throw a WARNing just by calling it if the input has no value; you may want to add:

`input && input.has_value? && input.value.is_a?(Hash)`

which will be quieter. #4995 addresses this a bit more.

clintoncwolfe (commented):
Some grumblings about where the tests are located, and a suggestion to add another predicate to the check. I like the idea of adding a sentinel value check to the Hash. Reading the user docs at https://www.inspec.io/docs/reference/waivers/ , it looks like `justification` might be a required field, but I can't find anything in the code to enforce that.

lib/inspec/rule.rb
Schwad: Thanks for this! :) 

I know that if you `pry` into this with a control that matches the affected user's, line 334:

`input = registry.inputs_by_profile.dig(__profile_id, input_name)`

Does return an `input` where `value` is the string `false`, not a hash. Which will break https://github.com/inspec/inspec/blob/7242da8bf557b6a76ac32d698a59d1d43a98ad04/lib/inspec/rule.rb#L343

I wasn't able to dig deep enough into the internals to figure out why clashing controls and input namespace does this, though I have some half-baked theories. I'll update per your suggestion & today writing a wider issue to look into this which I would love your feedback on! 👍

test/fixtures/profiles/old-examples/profile/controls/example-namespace-clash.rb
Schwad: Thanks for the feedback on this. I'm still building up my internal knowledge on how to test things here so that type of input is very much appreciated. I'll move this all over to `fixtures/profiles/waivers` to keep in flow 👍

test/functional/inspec_exec_test.rb
Schwad: Great point, I wondered if I could house this in a better location. I'll move it and link to the regression issue.

lib/inspec/rule.rb
Schwad: p.s. I don't resolve to say the conversation is over, just to note that I've achieved the changes so feel free to reopen 👍

Schwad:
Wider issue pertinent to other discussions opened here: https://github.com/inspec/inspec/issues/5005

I will check after lunch but think I will have to tweak some specs now that I've moved things around 👍

clintoncwolfe (approved):
Approved pending getting CI to green ;-)

test/functional/waivers_test.rb
```diff
@@ -10,6 +10,25 @@
...
+
+  def inspec(commandline, prefix = nil)
+    @stdout = @stderr = nil
+    self.out = super
+  end
+
+  def stdout
+    @stdout ||= out.stdout
+      .force_encoding(Encoding::UTF_8)
+      .gsub(/\e\[(\d+)(;\d+)*m/, "") # strip ANSI color codes
```
james-stocks: There's a `--no-color` option you can pass if you don't want color characters

Schwad: Nice! I was just using some setup code like it had been used in `inspec_exec_test`; might be worth extracting and unifying at some point but I didn't want to go too far with refactoring in this issue

Issue #4144, opened 2019-05-25
Report streaming capability


mikemol:
## Detailed Description
Right now, inspect runs are completed in full, and then their results are passed in bulk. It would be helpful if individual tests could be streamed to reporters, and those reporters could pass data onward as soon as they're available, or accumulate them into batches themselves if desired.

## Context
Real-time accumulation is important when using live logging data to measure the progress of a process, and permits more rapid presentation and response to error conditions; Inspec could be used as part of a linter, for example, if response latencies are sufficiently low. Likewise, if various profiles take a significant percentage of a process's runtime, being able to monitor their activities and consume their results in realtime makes for a significantly more responsive DevOps cycle.

## Possible Implementation

Reporters might expose a handful of callback functions for opening a test set, noting the start of a test, the end of a test (and its disposition, pass, fail, skip), and the close of a test set.

Closures calling each enabled reporter's callback function could then be passed to the test runner.

clintoncwolfe:
Refs #3667

PR #5007, opened 2020-05-02
Reporters as Plugins

### LINKED ISSUE #3667: Convert Reporters to a v2-API Plugin Type

clintoncwolfe:
## Description

Makes it possible to create a custom reporter and use it with InSpec. Includes documentation and tests.

Notably, this does not introduce a "streaming" API for reporters, per #4144 - that has some fundamental barriers related to how InSpec reconciles RSpecExamples back into Controls, not to mention the sequencing of when reporters are called at all.  That could be introduced later as an optional extension API.

I have not yet converted the  existing reporters into plugins, though that should be straightforward. I was holding off on doing that for at least the Automate ones, as I believe they are being worked on.

## Related Issue

Partially implements #3667 

## Types of changes

## Checklist:


lib/inspec/config.rb
```diff
@@ -328,21 +328,35 @@ def validate_config_file_contents!
...
+      # json-rspec is our alias for RSpec's json formatter.
+      rspec_built_in_formatters = %w{
         documentation
         html
+        json-rspec
+        progress
+      }
+
+      # These are true reporters, but have not been migrated to be plugins yet.
+      unmigrated_reporters = %w{
```
james-stocks: In the spirit of naming things to reflect current implementation, this could be named `inspec_reporters`.
Naming these `unmigrated`  might suggest that migration is a thing that happens somewhere inside the current InSpec code. 
The comment on the line above leaves a record that these need migrated at some point (we should make sure there's a ticket for doing this as well).

james-stocks:
I'm curious what happens if a user's plugin name clashes with one we ship (after we migrate current reporters to plugins).
Like if a user wants a YAML reporter (lol) named `yaml` and writes one for themself, then the next version of InSpec happens to ship a `yaml` reporter plugin. If this just results in an error, it may be reasonable to ask users to rename their plugins e.g. call theirs `yaml_company`

lib/inspec/reporters.rb
```diff
@@ -30,7 +30,10 @@ def self.render(reporter, run_data)
     when "yaml"
       reporter = Inspec::Reporters::Yaml.new(config)
     else
-      raise NotImplementedError, "'#{name}' is not a valid reporter type."
+      # If we made it here, it must be a plugin, and we know it exists (because we validated it in config.rb)
+      activator = Inspec::Plugin::V2::Registry.instance.find_activator(plugin_type: :reporter, activator_name: name.to_sym)
```
clintoncwolfe, line 34: A plugin naming collision will throw an error if there is a non-unique name loaded. So if a user installs an `inspec-my-yaml-reporter` that implements a `reporter: yaml` activator (and we've migrated our own YAML reporter), then this line will throw an exception because it expects to find a unique result.

james-stocks: Cool, thanks. It seems fair that we require users to rename their plugin (or move over to using the one we ship, if they do the same thing) :+1:

lib/inspec/plugin/v2/plugin_types/reporter.rb
```diff
@@ -0,0 +1,27 @@
...
+      @output << "\n" if newline
+    end
+
+    def rendered_output
+      @output
+    end
+
+    # each reporter must implement #render
+    def render
+      raise NotImplementedError, "#{self.class} must implement a `#render` method to format its output."
```
Schwad, line 61: Nice way to enforce render

Schwad (commented):
I've had a good read through, found this an informative read. I'm new enough and this PR is deep enough that I can't provide a tremendous amount of architectural feedback (yet). However tomorrow AM I'm planning on having a crack at building my own reporter against this branch to test things out as well. Thanks for this!

james-stocks:
@mjingle - this PR has some docs changes, can you review please?

test/helper.rb
```diff
@@ -148,7 +148,7 @@ def skip_until(y, m, d, msg)
   end
 
   def skip_windows!
-    skip_until 2020, 5, 1, "These have never passed" if windows?
+    skip_until 2020, 7, 1, "These have never passed" if windows?
```
james-stocks: This got merged into master already so can be removed from this PR

test/unit/plugin/v2/api_reporter_test.rb
```diff
@@ -0,0 +1,39 @@
+require "helper"
+
+require "inspec/plugin/v2"
+
+describe "Reporter plugin type" do
+  describe "when registering the plugin type superclass" do
+    it "returns the superclass when calling the global defintion method" do
```
james-stocks: Small typo `definition`

test/unit/plugin/v2/api_reporter_test.rb
```diff
@@ -0,0 +1,39 @@
...
+      klass = Inspec.plugin(2, :reporter)
+      _(klass).must_be_kind_of Class
+      _(klass).must_equal Inspec::Plugin::V2::PluginType::Reporter
+    end
+
+    it "returns the superclass when referenced by alias" do
```
james-stocks, line 13: I'm maybe not thinking in Ruby here, but isn't `Inspec::Plugin::V2::PluginBase` the superclass and `Inspec::Plugin::V2::PluginType::Reporter` the subclass?

james-stocks:
I wanted to test this manually before approving.
When I try to manually create a plugin in my home folder it doesn't seem to get detected.

The folder I created was:

```
jstocks@phobos inspec % tree ~/.inspec/plugins
/Users/jstocks/.inspec/plugins
└── inspec-jimmy
    └── lib
        └── inspec-jimmy
            ├── plugin.rb
            └── reporter.rb
```

`plugin.rb` contains:

```
module InspecPlugins
  module JimmyReporter
    class Plugin < ::Inspec.plugin(2)
      plugin_name :'inspec-jimmy-reporter'
      reporter :"jimmy" do
        require_relative "reporter.rb"
        InspecPlugins::JimmyReporter::ReporterImplementation
      end
    end
  end
end
```

The exception is like it's not even detected at all:

```
jstocks@phobos inspec % bundle exec inspec exec ../../profiles/inputs --reporter=jimmy
bundler: failed to load command: inspec (/Users/jstocks/code/bundles/inspec/ruby/2.6.0/bin/inspec)
NotImplementedError: 'jimmy' is not a valid reporter type.
  /Users/jstocks/code/inspec/lib/inspec/config.rb:362:in `block in validate_reporters!'
  /Users/jstocks/code/inspec/lib/inspec/config.rb:361:in `each'
  /Users/jstocks/code/inspec/lib/inspec/config.rb:361:in `validate_reporters!'
  /Users/jstocks/code/inspec/lib/inspec/config.rb:474:in `finalize_parse_reporters'
  /Users/jstocks/code/inspec/lib/inspec/config.rb:430:in `finalize_options'
  /Users/jstocks/code/inspec/lib/inspec/config.rb:65:in `initialize'
  /Users/jstocks/code/inspec/lib/inspec/base_cli.rb:249:in `new'
  /Users/jstocks/code/inspec/lib/inspec/base_cli.rb:249:in `config'
  /Users/jstocks/code/inspec/lib/inspec/cli.rb:282:in `exec'

```

test/unit/plugin/v2/api_reporter_test.rb
clintoncwolfe: "The superclass" here meaning "the superclass we are creating for the new plugin type" - that is, the class under test at line 6.

docs/dev/plugins.md
```diff
@@ -328,6 +328,62 @@ no_command do
 end
 ```
 
+## Implementing Reporter Plugins
```
mjingle, line 331: @cwolfe This new informative documentation is in a dev doc. Does this need to be moved to something more public-facing?

aaronlippold: Yes, it will land in the Plugins section I would guess

clintoncwolfe: It is documentation for developers, and is linked from the public-facing Plugins page. I wasn't planning on moving it, this is just adding a section to an existing doc.

mjingle: @clintoncwolfe Oh! I had forgotten/assumed dev docs were internal only similar to other chef repos. (That link is good to know about as we migrate the docs.) Thank you for clarifying. With that context, this new content should stay where it is. :) Copyedits will appear soon.

mjingle (approved):
This PR is good from a docs perspective. Thank you for the documentation, Clinton! :D

docs/dev/plugins.md
```diff
@@ -328,6 +328,62 @@ no_command do
...
+
+### Implementing your Reporter
+
+#### Implement render()
+
+The primary responsibility you must fulfill is to implement render. Typically, you will examine the `run_data` Hash, which is provided as an accessor. Call `output(String, newline_wanted = true)` to send output.
+
+#### The run_data structure
+
+The `run_data` object contains all data from the Chef InSpec run. This object is a Hash, but includes many fields and is often large. No specific documentation exists for the `run_data` object. See [the legacy JSON reporter](https://github.com/inspec/inspec/blob/2e887a94afcca819da781d4774aa2a5a0b56785e/lib/inspec/reporters/json.rb#L10) for one example of how to iterate over the object.
```
stevendanna: If this is intended as a user-facing feature, one worry here is that this effectively makes `run_data` and every object reachable from `run_data` part of the public facing API. Are we OK with that? I took a peek of the surface area there and it doesn't seem _too_ bad on the surface, but my guess is that exposing `Inspec::Profile` probably means most of our major data structures are somehow reachable from run_data.

To completely fix the problem probably involves boiling the ocean, but I'm wondering if we want to do a few things:

1) Wrap the run_data hash in a class with accessors before passing it to reporters. 
2) Extend some of this document with the core bits of this structure that we think we can keep stable over time.
2) Add a warning that accessing data other than those documented might result in their reporter breaking in the future?

(note, this is an actual question, I yield to y'alls judgement)

Schwad: Are you concerned that, say for example, someone builds a very involved/intricate reporter plugin that expects `run_data` to be a certain shape and format? And then if we modify it at any point in the future we risk breaking the reporter?

stevendanna: @Schwad Yup

Schwad:
This was my first time trying to compose an `inspec-plugin`, but I think this issue was a good opportunity. 

There may be edge cases but @clintoncwolfe I can confirm that I followed the guides and using this branch was able to publish an `inspec-pdf` reporter: https://github.com/Schwad/inspec-pdf . I then installed the plugin.

Against the branch `bundle exec inspec exec my_profile --reporter pdf` worked perfectly and generated this file: 
[inspec-report-152732809.pdf](https://github.com/inspec/inspec/files/4587387/inspec-report-152732809.pdf)

👍

docs/dev/plugins.md
james-stocks: This is a good point @stevendanna 
I think the more general problem is that there is no API into InSpec for plugins to use.

I believe that currently plugin maintainers are responsible for keeping their plugins working if they want to change InSpec version.
That is a. Unfriendly and b. Possibly very risky when it comes to reporting - their reporter plugin may continue to complete but start to miss bits of result data.

Another option is to maintain a versioned schema for run_data's structure and have reporter plugins need to declare what schema version they are intended for. It will still need plugin developers to keep up to date with InSpec changes, but at least their runs will outright fail instead of resulting in bad report data

clintoncwolfe: I'm put some time into making an object model for run_data and documenting as I go. I'm keeping it backwards compatible with Hash access so the existing reports should still work with it.

docs/dev/plugins.md
```diff
@@ -378,11 +381,105 @@ end
...
+Prior to version 1.0.0, the API is considered unstable.
 
 #### The run_data structure
 
-The `run_data` object contains all data from the Chef InSpec run. This object is a Hash, but includes many fields and is often large. No specific documentation exists for the `run_data` object. See [the legacy JSON reporter](https://github.com/inspec/inspec/blob/2e887a94afcca819da781d4774aa2a5a0b56785e/lib/inspec/reporters/json.rb#L10) for one example of how to iterate over the object.
+The `run_data` object contains all data from the Chef InSpec run. Here is an overview of this object:
+
+| Field | Description |
+|-------|-------------|
+|`run_data.control`| Array of Control records. Flattened copy of those found in |`run_data.profiles[*].controls[*]`. See there for details.|
```
Schwad: Thanks for taking the time to break out `run_data` in a way that can be documented and versioned as below, I find this really handy

test/unit/plugin/v2/api_reporter_test.rb
james-stocks: This example says that it `returns the superclass` but the assertion is `_(klass).must_equal Inspec::Plugin::V2::PluginType::Reporter`

In inheritance terminology (e.g. http://rubylearning.com/satishtalim/ruby_inheritance.html ) `Inspec::Plugin::V2::PluginType::Reporter` is the subclass not the superclass

lib/inspec/plugin/v2/plugin_types/reporter.rb
```diff
@@ -0,0 +1,38 @@
+require_relative "../../../run_data"
+
+module Inspec::Plugin::V2::PluginType
+  class Reporter < Inspec::Plugin::V2::PluginBase
```
james-stocks, line 4: Since this PR was raised, the base reporter class had some option handling added to it; it would be great to copy that here so that the options will also work with plugin reporters

https://github.com/inspec/inspec/commit/2103b52dbea8224f0c96d02912ec6cc940d27520#diff-bc3d043c1864e7c4d39d9aec04707b34R13

docs/dev/plugins.md
stevendanna: 💯 You went far beyond what I expected here. Thanks.

lib/inspec/run_data.rb
```diff
@@ -0,0 +1,60 @@
...
+      self.statistics = Inspec::RunData::Statistics.new(raw_run_data[:statistics])
+      self.platform   = Inspec::RunData::Platform.new(raw_run_data[:platform])
+      self.version    = raw_run_data[:version]
+    end
+  end
+
+  class RunData
+
+    # This is the data layout version of RunData.
+    # We plan to follow a data-oriented version of semver:
```
james-stocks, line 34: Awesome 👍

lib/inspec/run_data.rb
```diff
@@ -0,0 +1,60 @@
...
+    #  patch: fixing a bug in the provenance or description of a data element, no key changes
+    #  minor: adding new data elements
+    #  major: deleting or renaming data elements
+    # Less than major version 1.0.0, the API is considered unstable.
```
james-stocks, line 38: This says that < 1.0.0 is unstable, but the PR is starting with version 0.1.0. Do you intend to bump this to 1.0.0 before we merge this PR?

clintoncwolfe: @james-stocks and I discussed this a bit, and my thoughts are basically that I feel the API is indeed a bit unstable. This is largely because the source run_data Hash is populated variously depending on the profile being ran, and examined variously depending on the core reporter being used. So, it has been a case of discovery on both ends.

Until the core reporters are adapted to use the RunData object, I'm less confident that there won't be needed changes. That may be non-data changes - like adding more Hash-compatibility features.

That doesn't necessitate blocking this PR on converting the core reporters, though. It does mean that we may want to declare the RunData API as experimental - which it is - and that is a [totally normal thing to do with SemVer](https://semver.org/#spec-item-4).

Alternatively, the RunData API work could be split off, that work incorporated first as an internal refactoring with the core Reporters, then this PR could be revisited with a "mature" API to run against.

james-stocks: I'm thinking we can accept this PR as is, with just a little extra explanation for what 0.1.0 means and why we're not 1.0 (yet)

clintoncwolfe: Added an explanation.

james-stocks:
Let's have expeditor bump to 4.19 when this merges

james-stocks (approved):
Awesome, thanks @clintoncwolfe !

kmf:
watching

Issue #2821, opened 2018-03-14
inspec artifact workflow is not documented on the docs site


burtlo:
### 🎛 Description

This a 🐛 bug report or a 🙋 feature request. Based on how you view the world and the importance of a user being able to use a tool through the documentation.

The `inspec artifact` subcommand enables you to generate asymmetrical keys that you can use to encrypt a profile. You can then send the public key and an encrypted profile (independently) to another party and they can extract that profile with the given public key.

None of this workflow is described in the documentation.

### 🌍 InSpec and Platform Version

InSpec 2.0.32, Mac OSX 10.13.3

#### Profile Author

```shell
# generate your keys
$ inspec artifact generate --keyname myprofilekeys
# send the public key to the intended receiver of profiles
# Sign the profile at the given path
$ inspec artifact sign-profile --profile myprofile --keyname myprofilekeys
# myprofile.iaf is created
# send myprofile.iaf to the intended receiver with the public key
```

#### Profile Receiver

```shell
# The intended receiver of the profile needs to copy the key into the same working directory where they execute the command. The packaged profile path is specified so that can be anywhere.
$ inspec artifact install-profile --infile myprofile-0.1.0.iaf --destdir /tmp/myprofile
```

### 💁 Possible Solutions

* Document it.

* Remove it.

jquick:
Hey @burtlo - `inspec artifact` is only experimental at this point and should not be used. We will keep this undocumented until it matures and is ready for use. Thanks!

Issue #4777, opened 2019-12-11
Logging and Errors should go to STDERR always.


miah:
When anything prints to STDOUT, it corrupts tests that parse it, eg all the json tests.

Schwad:
Note: It appears any invocation of `Inspec::Log` does currently go to STDERR. However any `puts` statements would go to STDOUT. So a good rule moving forward will to keep an eye out for `puts` statements and clean up where possible (I ran into this in the above issue)

PR #5017, opened 2020-05-07
Stores output of `inspec json` to archive and artifact

### LINKED ISSUE #4973: include the JSON summary of a profile in any artifact produced
The JSON data produced by the `inspec json` command is a more complete record of a profile and its capabilities. A lesson learned fairly early with cookbook artifacts was to include a metadata record that was as complete as possible in the artifact itself. The publishing client performs all the interpretive work to determine the content's complete record of metadata and then stores that complete record within the artifact itself in a data-only format. That data is then available to any system that needs it as the artifact is passed around between systems.

1. The data should be 👍 All Correct 👍 at the time of artifact creation prior to publication.
2. The data does not change for a particular version after version publication.
3. The data is usually the product of interpretation, so for consistency and security's sake, only interpret once by the author/publisher.
4. All downstream systems that use the artifact can rely on a metadata record within the artifact to determine the artifact's capabilities. 

If the resultant JSON includes everything from the profile's YAML, I will go so far as to suggest that eventually tools and services that handle, consume, or use profiles should switch to treating the JSON record as the profile's metadata and _not_ use the YAML.

## Context

The JSON is already able to be produced by `inspec`. The `controls` data within the JSON can be used by downstream tools to inform users of a profile's capabilities and even to provide a list of controls to select from for customized reports.

## Possible Implementation

At a very high level, it seems possible to use the existing logic behind `inspec json` to produce the JSON file during `inspec archive` and `inspec artifact` for inclusion in the resultant tarball or signed IAF.

Schwad:
## Description

This was a feature request articulated in the issue below. 

The 30,000 foot version is that this PR takes the output of `inspec json` and stores it in the file `inspec.json` when running `inspec artifact` or `inspec archive`, within the actual archived file itself. It also includes a small refactoring to make the `inspec json` functionality portable throughout InSpec.

For more on the reasonings and benefits of this feature, see below.

## Related Issue

Fixes #4973 

If you are not familiar with `inspec artifact`, it appears to be a currently undocumented feature that is not recommended to be used. If you want to test it out or add to the discussion around artifact some useful info can be found here: https://github.com/inspec/inspec/issues/2821

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -12,6 +12,7 @@ require "inspec/method_source"
 require "inspec/dependencies/cache"
 require "inspec/dependencies/lockfile"
 require "inspec/dependencies/dependency_set"
+require "inspec/utils/json_profile_summary"
 
 module Inspec
   class Profile
@@ -465,27 +466,38 @@ module Inspec
       end
 
       # remove existing archive
-      File.delete(dst) if dst.exist?
+      FileUtils.rm_f(dst) if dst.exist?
       @logger.info "Generate archive #{dst}."
 
       # filter files that should not be part of the profile
       # TODO ignore all .files, but add the files to debug output
 
+      # Generate temporary inspec.json for archive
+      Inspec::Utils::JsonProfileSummary.produce_json(
+        info: info,
+        write_path: "#{root_path}inspec.json",
+        suppress_output: true
+      )
+
       # display all files that will be part of the archive
       @logger.debug "Add the following files to archive:"
       files.each { |f| @logger.debug "    " + f }
+      @logger.debug "    inspec.json"
 
       if opts[:zip]
         # generate zip archive
         require "inspec/archive/zip"
         zag = Inspec::Archive::ZipArchiveGenerator.new
-        zag.archive(root_path, files, dst)
+        zag.archive(root_path, files.push("inspec.json"), dst)
       else
         # generate tar archive
         require "inspec/archive/tar"
         tag = Inspec::Archive::TarArchiveGenerator.new
-        tag.archive(root_path, files, dst)
+        tag.archive(root_path, files.push("inspec.json"), dst)
       end
+
+      # Cleanup
+      FileUtils.rm_f("#{root_path}inspec.json")
 
       @logger.info "Finished archive generation."
       true
--- a/lib/inspec/utils/json_profile_summary.rb
+++ b/lib/inspec/utils/json_profile_summary.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Inspec
+  module Utils
+    #
+    # Inspec::Utils::JsonProfileSummary takes in certain information to identify a
+    # profile and then produces a JSON-formatted summary of that profile. It can
+    # return the results to STDOUT or a file. It is currently used in several
+    # places in the CLI such as `json`, `archive` and `artifact`.
+    #
+    #
+    module JsonProfileSummary
+      def self.produce_json(info:, write_path: "", suppress_output: false)
+        # add in inspec version
+        info[:generator] = {
+          name: "inspec",
+          version: Inspec::VERSION,
+        }
+        if write_path.empty?
+          puts JSON.dump(info)
+        else
+          unless suppress_output
+            if File.exist? write_path
+              Inspec::Log.info "----> updating #{write_path}"
+            else
+              Inspec::Log.info "----> creating #{write_path}"
+            end
+          end
+          full_write_path = File.expand_path(write_path)
+          File.write(full_write_path, JSON.dump(info))
+        end
+      end
+    end
+  end
+end
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -4,6 +4,7 @@ require "inspec/utils/deprecation/deprecator"
 require "inspec/dist"
 require "inspec/backend"
 require "inspec/dependencies/cache"
+require "inspec/utils/json_profile_summary"
 
 module Inspec # TODO: move this somewhere "better"?
   autoload :BaseCLI,       "inspec/base_cli"
@@ -77,24 +78,13 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     o[:vendor_cache] = Inspec::Cache.new(o[:vendor_cache])
 
     profile = Inspec::Profile.for_target(target, o)
-    info = profile.info
-    # add in inspec version
-    info[:generator] = {
-      name: "inspec",
-      version: Inspec::VERSION,
-    }
     dst = o[:output].to_s
-    if dst.empty?
-      puts JSON.dump(info)
-    else
-      if File.exist? dst
-        puts "----> updating #{dst}"
-      else
-        puts "----> creating #{dst}"
-      end
-      fdst = File.expand_path(dst)
-      File.write(fdst, JSON.dump(info))
-    end
+
+    # Write JSON
+    Inspec::Utils::JsonProfileSummary.produce_json(
+      info: profile.info,
+      write_path: dst
+    )
   rescue StandardError => e
     pretty_handle_exception(e)
   end
--- a/lib/plugins/inspec-artifact/lib/inspec-artifact/base.rb
+++ b/lib/plugins/inspec-artifact/lib/inspec-artifact/base.rb
@@ -5,6 +5,7 @@ require "set"
 require "tempfile"
 require "yaml"
 require "inspec/dist"
+require "inspec/utils/json_profile_summary"
 
 module InspecPlugins
   module Artifact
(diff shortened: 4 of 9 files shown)


Schwad:
NB: This feature still requires a decent chunk of tests but is passing my 'manual tryout' and I wanted to have it up and out before the UK bank holiday weekend so folks could try it.

Schwad:
note: will squash before merge

Schwad:
@TheLunaticScripter Don't know why but review request is hating me today so I can't request you. Just wanted to let you know I've got this running for you and @robbkidd if you wanted to try out this branch while the PR is still open for development. 

More than happy to treat this PR as our initial approach then if you think you need versioning for the future we can look into that too. Cheers! Really enjoyed working on this :)

TheLunaticScripter:
Sweet yeah I'll take a look. Super excited about this.

lib/inspec/json_profile_summary.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module Inspec
+  #
+  # Inspec::JsonProfileSummary takes in certain information to identify a
+  # profile and then produces a JSON-formatted summary of that profile. It can
+  # return the results to STDOUT or a file. It is currently used in several
+  # places in the CLI such as `json`, `archive` and `artifact`.
+  #
+  #
+  module JsonProfileSummary
```
james-stocks: This is a pretty specific thing to put directly into `Inspec::`. It's not wrong by any means, but much of the code has modules arranged like `Inspec::Category::Thing` 
There is a `Util` module that this could go inside.

lib/inspec/json_profile_summary.rb
```diff
@@ -0,0 +1,33 @@
...
+      info[:generator] = {
+        name: "inspec",
+        version: Inspec::VERSION,
+      }
+      if dst.empty?
+        puts JSON.dump(info)
+      else
+        unless suppress_output
+          if File.exist? dst
+            puts "----> updating #{dst}"
```
james-stocks: Suggest using InSpec logger calls instead of `puts`

lib/inspec/json_profile_summary.rb
```diff
@@ -0,0 +1,33 @@
...
+module Inspec
+  #
+  # Inspec::JsonProfileSummary takes in certain information to identify a
+  # profile and then produces a JSON-formatted summary of that profile. It can
+  # return the results to STDOUT or a file. It is currently used in several
+  # places in the CLI such as `json`, `archive` and `artifact`.
+  #
+  #
+  module JsonProfileSummary
+    def self.produce_json(info:, dst: [], suppress_output: false)
```
james-stocks: Could `dst` be documented or renamed? 
It looks like it can be an array or a filename; and the cases for empty array and a filename are implemented - what happens to a non-empty array?

lib/inspec/profile.rb
```diff
@@ -465,28 +466,39 @@ def archive(opts)
...
+      FileUtils.rm_f(dst) if dst.exist?
       @logger.info "Generate archive #{dst}."
 
       # filter files that should not be part of the profile
       # TODO ignore all .files, but add the files to debug output
 
+      # Generate temporary inspec.json for archive
+      Inspec::JsonProfileSummary.produce_json(
+        info: info,
+        dst: "#{root_path}inspec.json",
```
james-stocks: I'm suspicious there's a `/` missing from this, but I don't see where `root_path` is defined to check if it has a trailing `/` or not

lib/inspec/json_profile_summary.rb
Schwad: Hm, it looks like it should actually be an empty string default based on where it originally came from. Especially since being extracted it could be better documented.

lib/inspec/profile.rb
Schwad: This caught me out on the first implementation, which was `"#{root_path}/inspec.json"`, when I got the error of no path `something/something//inspec.json`

Definition:   https://github.com/inspec/inspec/blob/d478064973639d7fda36e0fb10ef8d3eb9f97333/lib/inspec/profile.rb#L518

I'll double check with pry though to be sure

lib/inspec/utils/json_profile_summary.rb
```diff
@@ -0,0 +1,36 @@
...
+    #
+    # Inspec::Utils::JsonProfileSummary takes in certain information to identify a
+    # profile and then produces a JSON-formatted summary of that profile. It can
+    # return the results to STDOUT or a file. It is currently used in several
+    # places in the CLI such as `json`, `archive` and `artifact`.
+    #
+    #
+    module JsonProfileSummary
+      def self.produce_json(info:, write_path: "", suppress_output: false, logger:)
+        @logger = logger || Logger.new
```
clintoncwolfe: InSpec has a globally available object, `Inspec::Log`, which has been configured by the CLI options. You don't need to pass around a logger or create one of your own, nor keep a reference to it.
Just use it like this:
`Inspec::Log.info "Something happened"`

clintoncwolfe (commented):
Great work, just a note about hey-we-already-have-a-logger.

lib/inspec/utils/json_profile_summary.rb
Schwad: Thanks for this! Very good to know and I've updated the repo :)

Schwad:
Reminder to self:: It appears Inspec::Logger knocked out a handful of `assert_output` tests, need to resolve

Schwad:
@clintoncwolfe quick question - updating from `puts` to `Inspec::Log.info` appears to have written to `stderr` in the `Train::Extras::CommandResult`.

Specifically this assertion:  https://github.com/inspec/inspec/blob/5cf742055b09781164b52cfed117e5114eab2284/test/functional/inspec_json_profile_test.rb#L113

<hr>

`Inspec::Log.info` produces this for  `out`:

```
#<struct Train::Extras::CommandResult
 stdout="",
 stderr=
  "[2020-05-18T10:45:27+01:00] INFO: ----> creating /var/folders/1c/y4lz6yt92n5cd1fj09tlydmm0000gn/T/inspec-shred20200518-28176-h8ap85\n",
 exit_status=0>
```

<hr>

The result of `puts "----> creating #{write_path}"` is:

```
#<struct Train::Extras::CommandResult
 stdout="----> creating /var/folders/1c/y4lz6yt92n5cd1fj09tlydmm0000gn/T/inspec-shred20200518-29052-18bulln\n",
 stderr="",
 exit_status=0>
```
<hr>

And for note, `Logger.new(STDOUT).info "----> creating #{write_path}"` gives:

```
#<struct Train::Extras::CommandResult
 stdout=
  "I, [2020-05-18T11:48:27.758109 #29252]  INFO -- : ----> creating /var/folders/1c/y4lz6yt92n5cd1fj09tlydmm0000gn/T/inspec-shred20200518-29235-1ec4s60\n",
 stderr="",
 exit_status=0>
```

I figured out that we use https://github.com/chef/mixlib-log but couldn't find any configs there that I think need changing...

Do I need to update [how we write Inspec::Log to ::CommandResult in train](https://github.com/inspec/train/search?q=CommandResult&unscoped_q=CommandResult) or is there something simpler I'm missing? 😅 

Thanks! 🙏

Schwad:
I think I'll hold onto this until at least next week release. Though it was unrelated to the initial PR (the `puts` logic was already there before) I'd love to resolve this in a way that I can use `Inspec::Log.info` and have it treated as `stdout` internally.

More digging, this with local transport. Looks like we're directly setting `CommandResult` from the outputs of `Mixlib.shellout.new(cmd)` in train:

https://github.com/inspec/train/blob/6cdb9f874cef373c22524758e6fb3fed3a2f6c25/lib/train/transports/local.rb#L74-L86

Which we set way over here, and may require more digging: 

https://github.com/chef/mixlib-shellout/blob/9b2ee29d50a44f309de456765f6fc76d398e7897/lib/mixlib/shellout.rb#L169-L188

I wonder if any of my other cross-app Foundation Friends ( @stevendanna , @tas50 ) who've played with `Mixlib::Shellout` and `Mixlib::Log` (which we directly use in `InSpec::Log`) might know why an `info` output is registering as a `stderr`? 🤔

james-stocks:
I believe logging is kept on stderr to keep stdout purely for result data. If stdout mixed results and log entries then stdout couldn't be piped to another program.
You may want to drop the log severity for your new log entries from info to debug - under normal usage users might not need to be informed about what's happening at this point.

Schwad:
Hey @stocksy - thanks for the background! That's really helpful and I was
suspicious I was diving a bit too deep here.

Currently I suppress output on `inspec archive` & `inspec artifact` for the
JSON file creation, the old flow on `inspec json` (before this PR) was to
output to STDOUT:

`---> creating x-file`

or

`---> updating x-file`

So it seems I have two options:

1. `inspec json PATH` no longer returns any confirmation output (lowering
to debug)
2. Update the three failing specs that expect `_(out.stderr).to eq ''` to
just equal or include the new output

I am tempted by number 2, but wanted to be sure that I properly understood
our `stderr` usage before doing so.

clintoncwolfe:
Broadly speaking, the *goal* was that inspec should keep data output on stdout, and logging/messaging on stderr (#4777). This is especially critical when outputting JSON as output. Existing code, however, doesn't always follow that guideline, and existing tests may enforce the bad behavior.

I'd suggest changing the tests to expect output to stderr.

Schwad:
Have implemented those changes, thanks for the message, that totally makes
sense now. :)

clintoncwolfe (approved):

james-stocks (approved):
Awesome!

Schwad:
🎉  😄

PR #5020, opened 2020-05-08
No longer need to include rubygems or bundler software dep

tyler-ball:
## Description
Recent versions of Ruby have included Rubygems and Bundler, which they
did not always do. We also recently fixed some issues in Appbundler so
these software dependencies are no longer needed.

## Types of changes

## Checklist:

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -4,8 +4,6 @@
 #
 # NOTE: You MUST update omnibus-software when adding new versions of
 # software here: bundle exec rake dependencies:update_omnibus_gemfile_lock
-override :rubygems, version: "3.0.3"
-override :bundler, version: "1.17.2" # currently pinned to what ships in Ruby to prevent double bundle
 # grab the current train release from rubygems.org
 train_stable = /^train \((.*)\)/.match(`gem list ^train$ --remote`)[1]
 override "train", version: "v#{train_stable}"
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -1,9 +1,5 @@
 # THIS IS NOW HAND MANAGED, JUST EDIT THE THING
-# .travis.yml and appveyor.yml consume this,
-# try to keep it machine-parsable.
-#
-# NOTE: You MUST update omnibus-software when adding new versions of
-# software here: bundle exec rake dependencies:update_omnibus_gemfile_lock
+
 # grab the current train release from rubygems.org
 train_stable = /^train \((.*)\)/.match(`gem list ^train$ --remote`)[1]
 override "train", version: "v#{train_stable}"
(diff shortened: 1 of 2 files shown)


tyler-ball:
I kicked off an ad-hoc build for this. If it builds correctly, can someone install the package and manually verify it has rubygems and bundler in the gem list? It shouldn't be able to build without those but I think it is worth doing some due diligence

omnibus_overrides.rb
```diff
@@ -4,8 +4,6 @@
 #
 # NOTE: You MUST update omnibus-software when adding new versions of
 # software here: bundle exec rake dependencies:update_omnibus_gemfile_lock
```
james-stocks: It may be helpful to insert a blank line to keep the comment about train separate from the comment block above

james-stocks:
Passing ad-hoc build: https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/129#35c0e285-884a-4206-b440-3813f9df4a8a

Schwad:
Nice catch @tyler-ball - thanks!

james-stocks:
@tyler-ball - I've installed http://artifactory.chef.co/simple/omnibus-unstable-local/com/getchef/inspec/4.18.113+20200508234600/windows/10/ and it comes with bundler 1.17.2 as a gem and C:\opscode\inspec\embedded\lib\ruby\2.6.0\rubygems is included in the install

A quick test of `inspec shell` shows the install as working. Anything else to check?

omnibus_overrides.rb
btm: I added a commit to this PR removing that comment altogether. It's actually not relevant in this project.

btm (approved):

btm:
@james-stocks we've been making this change in all of our projects recently, so confidence is high this just works and I think you've adequately checked everything.

james-stocks (approved):

omnibus_overrides.rb
```diff
@@ -1,11 +1,6 @@
 # THIS IS NOW HAND MANAGED, JUST EDIT THE THING
-# .travis.yml and appveyor.yml consume this,
-# try to keep it machine-parsable.
+# .travis.yml and appveyor.yml consume this, try to keep it machine-parsable.
```
tas50: This comment is no longer valid. We used it previously to install bundler / rubygems

lamont-granquist: technically its not a bad thing to keep it machine-parseable though.

tas50: I meant just the bit about travis / appveyor

btm: I removed the whole line. I was going to say "Please try to keep it machine-parsable" but then the obvious question was "for what? how do I know if the machine can parse this or not" and it isn't really doing anyone favors.

PR #5026, opened 2020-05-14
CLI Option to silence deprecations

### LINKED ISSUE #4228: Toggle for Deprecation Warnings
As a user I would like a way to toggle deprecation warnings off. This could be performed by setting an Environment variable, passing a CLI argument, or setting a configuration option.

clintoncwolfe:
## Description

Adds a CLI option to `exec`, `--silence-deprecations` which changes the action taken for specified deprecation groups from either `warn` or `fail_control` to `ignore`, effectively silencing them.

The user may name multiple deprecation groups separated by spaces, or in a feature I am sure will be used frequently, simply say `--silence-deprecations all` to suppress all such deprecations. 

Deprecations whose actions are `exit` cannot be suppressed, as they are intended to be the warning after functionality has *already* been removed.

## Related Issue

Closes #4228

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -155,6 +155,9 @@ module Inspec
         desc: "Show progress while executing tests."
       option :distinct_exit, type: :boolean, default: true,
         desc: "Exit with code 101 if any tests fail, and 100 if any are skipped (default).  If disabled, exit 0 on skips and 1 for failures."
+      option :silence_deprecations, type: :array,
+        banner: "[all]|[GROUP GROUP...]",
+        desc: "Suppress deprecation warnings. See install_dir/etc/deprecations.json for list of GROUPs or use 'all'."
     end
 
     def self.format_platform_info(params: {}, indent: 0, color: 39)
--- a/lib/inspec/utils/deprecation/config_file.rb
+++ b/lib/inspec/utils/deprecation/config_file.rb
@@ -1,6 +1,7 @@
 require "stringio"
 require "json"
 require "inspec/globals"
+require "inspec/config"
 
 module Inspec
   module Deprecation
@@ -32,6 +33,7 @@ module Inspec
         @groups = {}
         @unknown_group_action = :warn
         validate!
+        silence_deprecations_from_cli
       end
 
       private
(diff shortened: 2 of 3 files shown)


Schwad:
This looks awesome, thanks! :) 

I'll try this out locally tomorrow morning but otherwise looking pretty good

james-stocks:
This may be a bad thing if users start using the option constantly and then complain when things break in InSpec 6 "without warning".

Generally (not just for InSpec) users should address deprecation warnings by moving away from the deprecated functions.
I believe there's an issue open where we fire a warning for the core AWS resources even if the user is depending on the resource packs - that's a bug that's on us to fix.

Maybe there would be a way to have this CLI option be specific to InSpec 4 so when InSpec 5 releases, users who use the option constantly start to see new deprecation warnings. Or maybe we need to only ignore very specific warnings and not give an `all` option.

clintoncwolfe:
Thanks for your thoughts. 

I disagree about making this version specific. At major version change, the deprecation groups are supposed to change - making things that warn into fatalities, and adding new deprecation groups as needed. That's how inspec makes deprecation changes.

I agree this is potentially dangerous, but the user must actively ask to do something dangerous. This is not default behavior; and the option is inconveniently long. Take note of how many people are wanting this feature on #4228; people want to be able to squelch deprecations, despite the danger.

It sounds like most of your objections are around the 'all' feature. If you'd like that removed, I'm happy to do so, though I think it makes the feature less useful in the real world. It is very difficult to locate the installation source which contains the deprecations JSON file to determine the name of the deprecation group to use.

We could add the deprecation group name to the deprecation message to make it easier to determine which group to suppress.

james-stocks:
> At major version change, the deprecation groups are supposed to change

Ah, I didn't know that, thanks.

I think we should do _something_ to address the risk of `all` causing users to miss deprecation guidance after upgrading to a new major version. Options might be to not have `all`; or to rename it something that's tied to version and lets new deprecation warnings through; or keep it and put a clear warning in docs about the risk.
Any ideas @Schwad ?

Schwad:
Hi all! 

I think it's always interesting to try and figure out where the sharp knives should be handed out to end users and where they should be blunted. (As we know well, Ruby is a *very* sharp knife!) :) 

My thoughts: 

1. I like that it’s enforced as a flag at the CLI. I wouldn’t like a generic YAML config that would silence deprecations. That way every time you run exec you’re giving your consent to silence deprecations (other Ruby libraries, like Rails, would let you do this in the config which could absolutely trip you up years later!) - That requirement is probably enough to assuage fears I’d have around `all` otherwise. For our initial InSpec 5 upgrade guide we could stress the importance of running and squashing deprecations prior to upgrade. Kind of like has been done with these steps: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-upgrade-process

2. Also looking forward, if we are able to push out an InSpec Linter by we’ll be able to stop many deprecations before they ever get written (ideally)

james-stocks:
Sounds we should leave `all` is it currently is then!

@clintoncwolfe are doc changes needed for this option or do you want it intentionally undocumented? (Anyone actively looking for this option should find it the CLI description)

clintoncwolfe:
I think leaving it in the CLI usage docs is enough. Those get automatically converted to the website docs anyway.

Schwad (approved):
Had another look at this one and run locally. A whole run against `typical` profile matches expectations:

`bundle exec inspec exec test/fixtures/profiles/deprecation/typical --silence_deprecations all`

```
[2020-06-03T12:24:18+01:00] ERROR: DEPRECATION: This should exit (used at test/fixtures/profiles/deprecation/typical/controls/typical.rb:27)
```

(exit deprecations not suppressed as articulated above)

james-stocks (approved):
Brilliant, thanks @clintoncwolfe

PR #5034, opened 2020-05-18
Add placeholder verify pipeline

james-stocks:
The pipeline won't exist until it merges onto the 3-stable branch.
This commit adds the pipeline with a placeholder step - the next PR will contain the real verification steps.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -19,6 +19,14 @@ pipelines:
       definition: .expeditor/release.omnibus.yml
       env:
         - ADHOC: true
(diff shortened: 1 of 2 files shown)


james-stocks:
This PR is waiting on the `buildkite/inspec-inspec-master-verify` check - this is incorrectly configured and will never return a status.

After merging this, the branch protection options for this repo need changed so that this branch is protected by the `buildkite/inspec-inspec-3-stable-verify` check instead.

Schwad (approved):

Issue #4371, opened 2019-08-07
serverspec migration docs needs to be updated


zenspider:
https://www.inspec.io/docs/reference/migration/

this lists the resources not available in serverspec and is clearly out of date (we're up to about 100 resources)

Aha! Link: https://chef.aha.io/features/SH-2175

miah:
I dont think we've tried to maintain compatibility with serverspec for some time. I wonder if we should do away with this entire section of the site.

james-stocks:
There's also a section of the bottom of https://www.inspec.io/docs/reference/inspec_and_friends/ that needs updated. 
Suggest making InSpec and Friends change mention serverspec in the past tense, as part of InSpec's past. And perhaps just delete the migration guide; or also make clear it's now historical.

PR #5036, opened 2020-05-19
Additions to the interface resource

### LINKED ISSUE #722: Return IP address of tested node
### Description

I would expect, that inspec is providing an easy way to return the node ip address. This was reported by @cwebberOps 

InSpec 0.20.1
### Possible Solutions

At this point, a solution is to run a native command:

```
ip = command("ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/'").stdout.strip
```

Serverspec is offering this via the `interface` resource:

```
describe interface('eth0') do
  its(:ipv4_address) { should eq '1.2.3.4/1' }
  its(:ipv6_address) { should eq '2001:db8::1234/1' }
end
```

I would expect the same for InSpec.

Aha! Link: https://chef.aha.io/features/SH-2392

clintoncwolfe:
## Description

Several enhancements to the `interface` resource, and a new plural `interfaces` resource. Inspired by #722, currently the most upvoted issue.

 * some minor doc updates
 * Adds MacOS/ BSD support to the interface resource
 * Adds a pair of properties to `interface` for ServerSpec compatibility and convenience, which return the first IPv4 or IPv6 address in CIDR form on a named interface
 * However, knowing the names of your interfaces wasn't easy, so I added `interfaces`, which simply lists names of interfaces, on Linux, Windows, and MacOs/BSD,
 * and then I added a property to `interfaces`, to return the "first" IP it could find, which is what I think people were really after on on #722 .

Fun new areas to explore here include FilterTable and the MockLoader command loader system.

## Related Issue

Closes #722

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2391


lib/inspec/resources/interfaces.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # Look for progressively "better" IP addresses
+
+      # Reject anything that looks loopback-ish
+      non_loopback_addrs = all_addrs.reject { |a| a =~ /^127/ }
+      return all_addrs.first if non_loopback_addrs.empty?
+
+      # OK, we have something that isn't loopbackish.
+      # Try to filter out management networks.
+      non_management_addrs = non_loopback_addrs.reject { |a| a =~ /^10\./ }
```
james-stocks: I think `10.0.0.0/8` is just a private IP address range alongside `172.16.0.0/12` and `192.168.0.0/16` 

Maybe 10.x is used for management layers but I've also seen it used for workstation IP addresses.

I think these 3 ranges should be treated equally (some other IP address that _might_ be a public IP is preferable). 

https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses

docs/resources/interface.md.erb
```diff
@@ -30,61 +31,71 @@ An `interface` resource block declares network interface properties to be tested
...
 
-    its('name') { should eq eth0 }
+Provided for compatibility with ServerSpec. Returns the first `ipv4_cidrs` entry.
 
-### speed
-
-The `speed` property tests the speed of the network interface, in MB/sec:
-
-    its('speed') { should eq 1000 }
+    its('ipv4_address') { should eq '10.0.0.5/8' }
```
james-stocks: How come the IP addresses appear in CIDR form? Is the mask number `8` just redundant here?

clintoncwolfe: The `ipv4_address` property is provided for compatibility with ServerSpec, and that is how ServerSpec returns the value. Compatibility was requested on #722.

https://serverspec.org/resource_types.html#interface

lib/inspec/resources/interfaces.rb
clintoncwolfe: OK, I've reworked this section to clarify what is happening and remove the variable names which seem to be objectionable.  I've added 172.16/12 as well.

Schwad:
This looks pretty good to me but will petition for someone else's approval as I'm still in early days with resource composition & traversing the domain this resource is covering. Thanks for putting this together

docs/resources/interface.md.erb
```diff
@@ -30,61 +31,71 @@ An `interface` resource block declares network interface properties to be tested
...
 
 ### ipv4_addresses
 
-The `ipv4_addresses` property tests if the specified address exists on the named network interface:
+The `ipv4_addresses` property returns an Array of IPv4 addresses as Strings. You may then test if the specified address exists on the named network interface:
 
     its('ipv4_addresses') { should include '127.0.0.1' }
```
mjingle, line 56: @clintoncwolfe What type of interface (e.g. bash) is this supposed to be? Or is this supposed to be for a yml file?

clintoncwolfe: Network interface - a connection from a computer to a network. Could be wifi, ethernet, fiber optic, or something more exotic.

docs/resources/interface.md.erb
james-stocks: #722 seems concerned with getting the IP address, and gives the serverspec form as an example. From reading through the issue, I think the `8` was just a detail in the serverspec example but it's not beneficial for us to include. 

I imagine that people writing controls for IP addresses will want to work with just the IP address and needing to also include the mask number in their expected IP address string is just extra work for them.
Also `its('ipv4_addresses') { should include '10.0.0.5' }` uses IP addresses without mask bits.

We're also implementing `ipv4_addresses_netmask` and that lets users test IP address and mask in cases where they really do want to test IP address and mask.

lib/inspec/resources/interfaces.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+
+      addrs = interfaces.map(&:ipv4_addresses).flatten.map { |a| IPAddr.new(a) }
+
+      # Look for progressively "better" IP addresses
+      [
+        # Loopback and private IP ranges
+        IPAddr.new("127.0.0.0/8"),
+        IPAddr.new("192.168.0.0/16"),
+        IPAddr.new("172.16.0.0/12"),
```
james-stocks, line 42: Nice!  👍

docs/resources/interface.md.erb
james-stocks: As discussed, please have this property only return the IP address, and drop the mask number.

I said I would raise an issue to change our documentation about serverspec compatibility - I see there's already https://github.com/inspec/inspec/issues/4371 for this

docs/resources/interfaces.md.erb
```diff
@@ -0,0 +1,76 @@
...
+platform: os
+---
+
+# interfaces
+
+Use the `interfaces` Chef InSpec audit resource to test the properties of multiple network interfaces on the system.
+
+## Version
+
+This resource first became available in v??? of InSpec.
```
mjingle: @clintoncwolfe Is there available version information for this?

james-stocks (approved):
Fantastic, thanks Clinton!

docs/resources/interfaces.md.erb
mjingle: @clintoncwolfe Unless told otherwise, I'm going to modify the ??? to have an approximation a la 4.20.x

mjingle: Just removed this entirely per discussion. :)

mjingle (approved):
Docs are good. Thank you again, @clintoncwolfe!

james-stocks:
Removing the Bump Version label as https://github.com/inspec/inspec/pull/5045 just merged and will bump the minor version since the last release already.

PR #5038, opened 2020-05-19
Expand docs and tests for only_if

### LINKED ISSUE #760: document only_if for controls
### Description

With https://github.com/chef/inspec/pull/619 we introduced `only_if` for controls, but we have not documented it yet.

```
control 'hello' do
  only_if { os[:family] != 'arch' }
  describe 1 do
    it { should eq 2 }
  end
end
```

0.22.0

Aha! Link: https://chef.aha.io/features/SH-2193

clintoncwolfe:
## Description

#760 is currently the third-most 👍 issue in the repo, and is a simple docs request. Turns out not so simple; the behavior of `only_if` is poorly defined, because we have no tests on it.

I added tests and test fixtures.  I added more detailed docs. 

I changed the heading of the doc section to include the word `only_if`. This improves the eye-catching ability and likely search results, but also breaks links.

## Related Issue

Closes #760

## Types of changes

## Checklist:

### DIFF
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -215,10 +215,10 @@ control 'windows-base-101' do
 end
 ```
 
-### Use `only_if` to exclude a specific test
+### Use `only_if` to exclude a specific control
 
-This shows how to allow skipping certain tests if conditions are not met, by using `only_if`.
-In this example the test will not be performed if `redis-cli` command does not exist. A optional
+This shows how to allow skipping certain controls if conditions are not met, by using `only_if`.
+In this example the control will not be performed if `redis-cli` command does not exist. A optional
 message can say why it was skipped.
 
 ```ruby
@@ -237,7 +237,7 @@ control 'nutcracker-connect-redis-001' do
 end
 ```
 
-Mixing this with other conditionals (like checking existence of the files etc.) can help to test different test paths using InSpec. This way you can skip certain tests which would 100% fail due to the way servers are prepared, but you know that the same test suites are reused later in different circumstances by different teams.
+Mixing this with other conditionals (like checking existence of the files etc.) can help to test different test paths using InSpec. This way you can skip certain controls which would 100% fail due to the way servers are prepared, but you know that the same control suites are reused later in different circumstances by different teams.
 
 Some notes about `only_if`:
 
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -215,7 +215,7 @@ control 'windows-base-101' do
 end
 ```
 
-### Exclude specific test
+### Use `only_if` to exclude a specific test
 
 This shows how to allow skipping certain tests if conditions are not met, by using `only_if`.
 In this example the test will not be performed if `redis-cli` command does not exist. A optional
@@ -238,6 +238,43 @@ end
 ```
 
 Mixing this with other conditionals (like checking existence of the files etc.) can help to test different test paths using InSpec. This way you can skip certain tests which would 100% fail due to the way servers are prepared, but you know that the same test suites are reused later in different circumstances by different teams.
+
+Some notes about `only_if`:
+
+ * `only_if` applies to the entire `control`. If the results of the `only_if` block evaluate to false, the contents of the describe blocks will not be run. However, the describe block resources (`command('redis-cli SET test_inspec "HELLO"')` in the above example) that preceded the only_if WILL run. For this reason, `only_if` should usually come first in the control, prior to any `describe` blocks.
+
+```ruby
+control "don't do this" do
+  describe command("something-dangerous") do
+    # ...
+  end
+  # command("something-dangerous") already ran!!!
+  only_if { its_safe }
+end
+```
+
+Instead, do this:
+```ruby
+control "do this instead" do
+  only_if { its_safe }
+  # command("something-dangerous") won't run unless it is safe
+  describe command("something-dangerous") do
+    # ...
+  end
+end
+```
+
+ * `only_if` results in a skipped result for the control and all tests in the control, and is the only supported way of skipping an InSpec control. Other methods relying on RSpec-specific features such as `before` and `skip` may not work in the future.
+ * Only one `only_if` is permitted per `control` block; if multiple `only_if` blocks are present, only the last one will be honored.
+ * `only_if` may also be used outside a control block. In that case, it will be used to skip all controls in the current file.
+ * To implement complex logic, use Ruby 'or' (`||`) and 'and' (`&&`) inside your `only_if` block:
+
+```ruby
+  only_if('ready for launch') do
+    rocket_is_ready && weather_is_clear
+  end
+```
+
 
 ### Additional metadata for controls
 
--- a/docs/dsl_inspec.md
+++ b/docs/dsl_inspec.md
@@ -217,9 +217,8 @@ end
 
 ### Use `only_if` to exclude a specific control
 
-This shows how to allow skipping certain controls if conditions are not met, by using `only_if`.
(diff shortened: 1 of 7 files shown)


docs/dsl_inspec.md
```diff
@@ -239,6 +239,43 @@ end
...
+  # command("something-dangerous") won't run unless it is safe
+  describe command("something-dangerous") do
+    # ...
+  end
+end
+```
+
+ * `only_if` results in a skipped result for the control and all tests in the control, and is the only supported way of skipping an InSpec control. Other methods relying on RSpec-specific features such as `before` and `skip` may not work in the future.
+ * Only one `only_if` is permitted per `control` block; if multiple `only_if` blocks are present, only the last one will be honored.
+ * `only_if` may also be used outside a control block. In that case, it will be used to skip all controls in the current file.
```
Schwad: Is this similar as above where if you put the `only_if` in the middle of the file, it will only skip the following controls? & Only one only_if will run in the scope?

Schwad (commented):
This is awesome - thank you! Is this the sort of PR that we'd need to cc in the docs team for a docs review? 

I can say from the perspective of a developer who hasn't used `only_if` much in the wild yet (target audience?) I find the docs very clear and informative. And always appreciate the additional testing!

docs/dsl_inspec.md
clintoncwolfe: My testing showed it skipped all controls in the file, including those that preceded it.

docs/dsl_inspec.md
```diff
@@ -215,7 +215,7 @@ control 'windows-base-101' do
 end
 ```
 
-### Exclude specific test
+### Use `only_if` to exclude a specific test
```
james-stocks: In this whole section, the term `test` should possibly be `control`

I would use the term `test` to refer to refer to individual `it` assertions... but maybe I'm wrong on that.

docs/dsl_inspec.md
```diff
@@ -239,6 +239,43 @@ end
 
 Mixing this with other conditionals (like checking existence of the files etc.) can help to test different test paths using InSpec. This way you can skip certain tests which would 100% fail due to the way servers are prepared, but you know that the same test suites are reused later in different circumstances by different teams.
 
+Some notes about `only_if`:
+
+ * `only_if` applies to the entire `control`. If the results of the `only_if` block evaluate to false, the contents of the describe blocks will not be run. However, the describe block resources (`command('redis-cli SET test_inspec "HELLO"')` in the above example) that preceded the only_if WILL run. For this reason, `only_if` should usually come first in the control, prior to any `describe` blocks.
```
james-stocks: I don't think this is true.
I wrote this test control - 

```
control 'only_if test' do

  describe command('echo \'AFTER\' > /tmp/AFTER') do
  end

  only_if { false }

end
```

The command causes the file `/tmp/AFTER` to exist. When I run this, the control is skipped and the file does not exist. If I change the `false` to `true` then the control runs and the file is created.

A case where this is true is where Ruby statements are placed outside of InSpec's describe/its DSL:

```
control 'only_if test' do

  result = command('echo \'AFTER\' > /tmp/AFTER').exit_status

  describe result do
    it { should cmp 0 }
  end

  only_if { false }

end
```

For this control, the tmp file WILL get created.

Keka, @Schwad and myself have been discussing how Ruby statements like this are generally bad practice because they execute regardless of skip/waiver/exclude_control etc etc. If warning the user here about Ruby statements it would be good to make clear that they are generally not good practice (not just for this specific case).

docs/dsl_inspec.md
```diff
@@ -239,6 +239,43 @@ end
...
+control "do this instead" do
+  only_if { its_safe }
+  # command("something-dangerous") won't run unless it is safe
+  describe command("something-dangerous") do
+    # ...
+  end
+end
+```
+
+ * `only_if` results in a skipped result for the control and all tests in the control, and is the only supported way of skipping an InSpec control. Other methods relying on RSpec-specific features such as `before` and `skip` may not work in the future.
```
james-stocks: `skip` is used in our compliance content; primarily for compliance checks that are specified in a standard but  can't be automated.
We could suggest changing these to `only_if('This cannot be automated') { false }` but that's a lot less readable. 
I would lean towards considering `skip` supported and adding a test that proves that it works.

docs/dsl_inspec.md
james-stocks: That's cool, I didn't know that. Quickly testing it, I see you can also have a per-control skip as well as the entire-file skip. That's interesting because if you have one control per file then you have a way to have two separate conditions that don't share the same line of code (something we've wanted more than once in the past).

docs/dsl_inspec.md
james-stocks: I see the test case is doing `describe 1/0 do`
There, the `1/0` is a Ruby evaluation.

test/fixtures/profiles/only_if/skip-control/controls/skip-control.rb
```diff
@@ -0,0 +1,56 @@
...
+  desc "This control should demo that following test resources do not get evaluated"
+  only_if { false }
+  describe 1/0 do # does not error!
+    it { should cmp 1/0 }
+  end
+end
+
+control "control-skip-test-outer-error-test-first" do
+  desc "This control should demo that preceding test resources DO get evaluated"
+  describe 1/0 do # does error!
```
james-stocks: At the moment there are a number of InSpec features that don't work well with Ruby statements that InSpec will need to evaluate (like `1/0` or `1+1`). I'd prefer not to put a test like this into our supported tests.

The approach I'd recommend to people is to keep *all* code inside of a resource; even if it means needing custom resources that boil down to a single line of functional code.

It's something we're chatting with @kekaichinose about, to decide how to improve on this.

docs/dsl_inspec.md
clintoncwolfe: Where is `skip` documented? I'm documenting and testing `only_if`, which is an `inspec` feature. Is `skip` an RSpec feature? Where is it tested and documented?

docs/dsl_inspec.md
clintoncwolfe: Well, you're right-ish. It's at least sort of true :-) The fact that 1/0 is evaluated in a preceding control, and not in a subsequent control, indicates that there is a difference.

I have no idea how that works.

But I tried it with a command(echo) and I got the same result as you. I'll edit accordingly and treat the ruby expression case as an undocumented edge case.

test/fixtures/profiles/only_if/skip-control/controls/skip-control.rb
clintoncwolfe: It appears the action for me the contributor is to remove the test, which is what I have done.

docs/dsl_inspec.md
clintoncwolfe: OK, I tracked it down - it's an RSpec core feature, documented here - https://relishapp.com/rspec/rspec-core/docs/pending-and-skipped-examples/skip-examples#skipping-using-%60skip%60 . 

So, this PR is about `only_if`. I've removed the references to `skip`  (I had mentioned it because a commenter on #760 had asked specifically about it). If you feel documentation for `skip` should be added to InSpec, or even tests for it, that should probably be done on another issue / PR.

james-stocks (approved):
Thanks very much @clintoncwolfe ! 
Need Mary's review before merging.
At some point we should write up docs about InSpec code evaluation, as it catches most users out.

Schwad (approved):
Looking great 👍

mjingle (approved):
LGTM from a docs perspective. Thank you, @clintoncwolfe!

PR #5039, opened 2020-05-20
Allow for waiver time as well as date

### LINKED ISSUE #5037: Waiver expiration dates do not work with (zeroed) time accuracy
The current code for waiver expiration depends on the value being a Date

See https://github.com/inspec/inspec/blob/d9e5b65e3d5c710b479499b8c3946905b5d51834/lib/inspec/rule.rb#L356

```
      expiry = __waiver_data["expiration_date"]
      if expiry
        if expiry.is_a?(Date)
          # It appears that yaml.rb automagically parses dates for us
          if expiry < Date.today # If the waiver expired, return - no skip applied
            __waiver_data["message"] = "Waiver expired on #{expiry}, evaluating control normally"
            return
          end
        else
          ui = Inspec::UI.new
          ui.error("Unable to parse waiver expiration date '#{expiry}' for control #{@__rule_id}")
          ui.exit(:usage_error)
        end
      end
```

If the date is expressed in a time format with higher precision (examples: '2021-11-10T00:00:00Z', '2020-11-04T00:00:00' ), ruby YAML will provide a value with class Time instead. This fails `if expiry.is_a?(Date)` and throws the "unable to parse" error.

Time-precise values should be supported even if only the DD/MM/YYYY elements are used. This bug was noticed because another bit of code that uses Go's YAML library which always uses nano precision when writing out dates or times to YAML.
It may potentially hit users who copy the expiration date over from another system.

james-stocks:
Fixes #5037

The YAML parser may parse a waiver timestamp as a Time rather than a Date. Even when the user doesn't care about time, they may be using a tool that outputs YAML with trailing zeroes for hour, minutes, seconds etc.

TODO: confirm with @kekaichinose if allowing for time accuracy is desirable. We could also keep working with dates only and warn the user if they specify non-zero time values that will be ignored.

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -353,9 +353,12 @@ module Inspec
       # if so, is it in the future?
       expiry = __waiver_data["expiration_date"]
       if expiry
-        # YAML will automagically give us a Date or a Time
-        if [Date, Time].include?(expiry.class)
+        # YAML will automagically give us a Date or a Time.
+        # If transcoding YAML between languages (e.g. Go) the date might have also ended up as a String.
+        # A string that does not represent a valid time results in the date 0000-01-01.
+        if [Date, Time].include?(expiry.class) || (expiry.is_a?(String) && Time.new(expiry).year != 0)
           expiry = expiry.to_time if expiry.is_a? Date
+          expiry = Time.new(expiry) if expiry.is_a? String
           if expiry < Time.now # If the waiver expired, return - no skip applied
             __waiver_data["message"] = "Waiver expired on #{expiry}, evaluating control normally"
             return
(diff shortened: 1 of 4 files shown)


james-stocks:
cc @TheLunaticScripter

james-stocks:
Chatting to @TheLunaticScripter , the problem interacting with Go is actually two-fold:
1. Go doesn't recognise YAML `expiration_date: 2014-04-04` as a date, it will load it in as a string.
2. Go writes dates to YAML as a full timestamp; which InSpec subsequently considers a Time (not a Date) when it reads the YAML.

clintoncwolfe (approved):
Thanks, this looks like it will make the waivers expiry system much more robust!

TheLunaticScripter (approved):
This is awesome thanks so much for getting this in so quickly.

Issue #5042, opened 2020-05-21
warnings from pry-debug with inspec shell under chef-workstation


clintoncwolfe:
# Describe the problem

Install the latest build of chef-workstation (available from Current channel on the downloads page). Run `eval "$(chef shell-init bash)"` to set your environment to run from Chef workstation. When you run `inspec shell`, you get:

```
[cwolfe@lodi temp]$ inspec shell
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-byebug-3.9.0/lib/pry-byebug/pry_ext.rb:13: warning: __FILE__ in eval may not return location in binding; use Binding#source_location instead
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-stack_explorer-0.4.9.3/lib/pry-stack_explorer/when_started_hook.rb:63: warning: in `eval'
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-0.13.1/lib/pry/core_extensions.rb:45: warning: __FILE__ in eval may not return location in binding; use Binding#source_location instead
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-stack_explorer-0.4.9.3/lib/pry-stack_explorer/when_started_hook.rb:63: warning: in `eval'
Welcome to the interactive InSpec Shell
To find out how to use it, type: help

You are currently running on:

    Name:      mac_os_x
    Families:  darwin, bsd, unix, os
    Release:   18.7.0
    Arch:      x86_64

inspec> exit
[cwolfe@lodi temp]$ chef --version
Chef Workstation version: 20.5.45
Chef Infra Client version: 16.1.0
Chef InSpec version: 4.18.114
Chef CLI version: 3.0.1
Test Kitchen version: 2.5.1
Cookstyle version: 6.5.3
```

## Possible Solution

Aha! Link: https://chef.aha.io/features/SH-2176

james-stocks:
This is already fixed upstream

```
jstocks@phobos inspec % ./inspec-bin/bin/inspec shell 
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-byebug-3.9.0/lib/pry-byebug/pry_ext.rb:13: warning: __FILE__ in eval may not return location in binding; use Binding#source_location instead
[...]

PR #5043, opened 2020-05-22
Use forked copy of pry-stack_explorer to avoid ruby 2.7 warning

clintoncwolfe:
Makes a small modification to the Gemfile to cause the omnibus distribution of InSpec to use a patched version of pry-stack_explorer.

This is a temporary measure until a new, fixed gem is properly released. A PR has been created upstream. Until it is accepted and a new gem is released, the fix will not be present in gem-based distributions of Chef InSpec.

https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/133 is an adhoc omnibus build on the branch.

## Related Issue

Partial fix for #5042 
https://github.com/chef/chef-workstation/pull/1225 similar PR in Chef Workstation to use fork in CW distribution

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,10 @@ gem "inspec-bin", path: "./inspec-bin"
 
 gem "ffi", ">= 1.9.14"
 
+# pry-stack_explorer 0.4.9.3 emits warnings under 2.7 under inspec shell
+# https://github.com/pry/pry-stack_explorer/issues/43
+gem "pry-stack_explorer", git: "https://github.com/chef/pry-stack_explorer.git"
+
 group :omnibus do
   gem "rb-readline"
   gem "appbundler"


james-stocks:
There's already a related fix upstream, but it discusses fixing the issue for 2.7 and leaving 2.5 behind
https://github.com/pry/pry-stack_explorer/pull/49

james-stocks:
I noticed while reproducing the issue to test this fix that InSpec does not currently depend on `pry-stack_explorer`- `bundle info pry-stack_explorer` doesn't list it as a bundled gems. The issue only happens when running InSpec inside an environment with the gem installed. 
Can the problem be fixed by switching to the Chef fork for the gem in the project(s) that do need the gem as a dependency?

clintoncwolfe:
Agreed, it's not in the inspec omnibus build nor in the gemfile.lock. It's just in the workstation build. I'm closing this.

PR #5045, opened 2020-05-23
Improved HTML Reporter

### LINKED ISSUE #4178: HTML reporter needs re-styling to remove RSpec branding, improve UX
Just saw my first screenshot of someone's inspec run and right at the top is "rspec" emblazoned across the top. We should probably wrap that up w/ our own, and restyle it while we're in there. There was also an issue about not finding the right snippet, which was what was originally coming up as it looks like an error w/ no details (because it basically is):

unnamed

Aha! Link: https://chef.aha.io/features/SH-2390

clintoncwolfe:
## Description

This is the html2 reporter plugin demo'd in Friday May 22nd's internal company demo.

To try it out, check out the branch, then try:
```
bundle exec inspec exec test/fixtures/profiles/old-examples/profile --reporter html2:out.html
```

Then open `out.html` in your browser. Compare results vs using the `html` reporter.

Try it with different profiles: those that inherit; those with many controls; those with few controls but many tests per control. See if it gets unclear.

~It is not done, but I had several requests from people to either play with it or start looking at the web design, so a Draft PR it is.~ Feels done, except for UX/UI work, which is not scheduled.

Remaining work prior to merge:
 * ~There are no tests~
 * ~There are no docs~
 * ~I've started but not finished a mechanism by which a user may, via the inspec config file, replace the CSS and/or JavaScript~
 * ~We should probably put in conditionals on the branding to allow the community distribution to use it~
 * ~I'd like to add an interactive feature by which child profiles may be shown/hidden, with them starting off initially hidden.~
 * People will probably have opinions about what data I chose to include or not include, and how I chose to represent it. Make your case :-)

Remaining work that I imagine may occur on a later PR, or could occur on this one:
 * Consultation with and contributions from the UX/UI team. This work was created as a proof of concept; the general idea is there but I am no web designer, not having worked in CSS for nearly a decade. There is much that could be done. Note however that the output must be a single HTML file with no external references (no image loads, no JS files, no CSS) - this must be airgap-compatible. So if you want fancy libraries, you'll have to embed them.

## Related Issue

Closes #4178 

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2369


Schwad (approved):
I just had a careful look through and also ran this against my machine. 

I'd like to approve this and am open to further feedback.

It would be nice to get this into users' hands, & if they have strong opinions about changing data or styling we can have these discussions in issues & InSpec. It would be fun to have our community using this and offering their feedback before it supercedes the original `html` (which I'm assuming will happen at some release in future).

I personally am a fan of this layout because:

1. It contains needed data in such a way that already puts us lightyears ahead of legacy `html` (`show source` and checkbox filtering is a great addition too)
2. It's incredibly lightweight - my test run total size was `17kb`. When lots of styling comes in this can quickly balloon.

In future if there were desire for a highly stylized layout, I would even advocate keeping an additional lightweight plugin in this style.

Thanks so much for your hard work on this @clintoncwolfe - really slick! 👍  😄

lib/plugins/inspec-reporter-html2/templates/control.html.erb
```diff
@@ -0,0 +1,77 @@
...
+      </tr>
+    <% end %>
+    <tr class="code">
+      <th>Source Code:</th>
+      <td>
+        <input type="button" class="show-source-code" id="show-code-<%= slugged_id %>" value="Show Source"/>
+        <input type="button" class="hide-source-code hidden" id="hide-code-<%= slugged_id %>" value="Hide Source"/>
+        <pre class="source-code hidden" id="source-code-<%= slugged_id %>">
+          <code>
+<%= control.code %>
```
james-stocks, line 64: I don't think it needs fixed but I noticed that when a test exists outside of a control e.g. for our `simple-inheritence` test fixture profile:

```
include_controls 'failures'

describe file('/') do
  it { should be_directory }
end
```

...the source code panel is blank.
Having describes outside controls is supported but I think it's reasonable to require that anyone who wants good quality HTML reports will also follow best practices and put all their tests inside `control`s

james-stocks (approved):
This is superb, thank you @clintoncwolfe !
I left one comment about missing source for a describe outside a control, but I don't think it's essential to do anything about.

Have added Mary as reviewer to look over the docs changes.

lib/plugins/inspec-reporter-html2/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+you@machine $ inspec exec some_profile --reporter html2:report.html
+```
+
+Note the `2` in the reporter name; if you omit it and run `--reporter html` instead, you will run the legacy RSpec HTML reporter.
+
+## Configuring the Plugin
+
+No configuration is required for the plugin to function. However, two options are available for customization. The options are set in the JSON-formatted configuration file that Chef InSpec consumes. For details on this file, see [the documentation](https://www.inspec.io/docs/reference/config/).
```
mjingle: @clintoncwolfe Are the two options you mention in the second sentence here named on lines 43 and 47?

clintoncwolfe: Yes, `alternate_css_file` and `alternate_js_file`

mjingle (approved):
Good to go from a docs perspective! :D Thank you @clintoncwolfe!!!

james-stocks:
This is good to merge, I'd like to get this and https://github.com/inspec/train/pull/613 released this week - so pausing merging this until that train update is available.

Issue #4972, opened 2020-04-08
InSpec Reporter | HTML format append error message


amitchettri:
# Describe the problem
I am using InSpec Reporter to generate the patch scan report in HTML format but in the report I am getting error message as show in the snapshot. Can i get some help overcome this?
![image](https://user-images.githubusercontent.com/46727623/78772074-d1d6b100-79ae-11ea-8f1c-ba357e2c5d7f.png)

I have used the devsec community profile from the supermarket to get this report and below is the command executed
`inspec supermarket exec dev-sec/linux-patch-baseline -t ssh://root:password@192.168.4.13 --reporter html:patch.htm`

When i don't generate the report and get the output on screen the output is normal as below

```
Profile: DevSec Linux Patch Benchmark (linux-patch-baseline)
Version: 0.5.0
Target:  ssh://root@192.168.4.13:22

  ×  verify-patches: Operating system is up-to-date
     ×  Linux Update is expected to be uptodate
     expected `Linux Update.uptodate?` to return true, got false
  ×  patches: All operating system package updates are installed (480 failed)
     ×  System Package device-mapper-persistent-data version is expected to eq "0.8.5-1.el7"

     expected: "0.8.5-1.el7"
          got: "0.7.3-3.el7"

     (compared using ==)

     ×  System Package gnome-settings-daemon version is expected to eq "3.28.1-5.el7"

     expected: "3.28.1-5.el7"
          got: "3.28.1-2.el7"

     (compared using ==)
```

Aha! Link: https://chef.aha.io/features/SH-2529

amitchettri:
for all the profile when --reporter html:name.html CLI flag is used. the generated page have below error's added on it 

below is from a sample from nginx-baseline profile. And a sample snapshot from patch and linux basline profile
```
Failure/Error: [0m[1;34;4mDEFAULT_FAILURE_NOTIFIER[0m = lambda { |failure, _opts| raise failure }

# nginx-baseline-master/controls/nginx_spec.rb:133:in `block (3 levels) in load_with_context'
nginx-baseline-master/controls/nginx_spec.rb:133:in `block (3 levels) in load_with_context'
131# Couldn't get snippet for nginx-baseline-master/controls/nginx_spec.rb
```


<img width="639" alt="linux-basline" src="https://user-images.githubusercontent.com/46727623/78927744-06d12980-7abd-11ea-93ac-907967b69a40.png">
[...]

PR #5054, opened 2020-05-28
RunData Filters; sensitive inputs; suppress diff; sorted output

### LINKED ISSUE #1535: Option to disable Diff output within exec
### Description

When running `inspec exec`, I'd like the ability to disable the Diff output on failed tests.

Inspec 1.7.1 on Mac OS X 10.9.5

### Replication Case

    /opt/inspec/bin/inspec exec cis_oracle_db_12c_test.rb -t ssh://oracle@dbserver01.example.com

    Target:  ssh://oracle@dbserver01.example.com:22

    ×  cis-oracle-db-12c-3-4: 3.4 Ensure 'PASSWORD_REUSE_MAX' Is Greater than or Equal to '20' (Scored) (
       expected: ""
          got: "DEFAULT\nPASSWORD_REUSE_MAX\nUNLIMITED\n\nORA_STIG_PROFILE\nPASSWORD_REUSE_MAX\n10"
     
     (compared using ==)
     
     Diff:
     @@ -1 +1,8 @@
     +DEFAULT
     +PASSWORD_REUSE_MAX
     +UNLIMITED
     +
     +ORA_STIG_PROFILE
     +PASSWORD_REUSE_MAX
     +10
     )
     ×  Oracle SQLPlus query: SELECT PROFILE, RESOURCE_NAME, LIMIT
     FROM DBA_PROFILES
     WHERE RESOURCE_NAME='PASSWORD_REUSE_MAX'
     AND
     (
     LIMIT = 'DEFAULT'
     OR LIMIT = 'UNLIMITED'
     OR LIMIT < 20
     ); output should eq ""
     
     expected: ""
          got: "DEFAULT\nPASSWORD_REUSE_MAX\nUNLIMITED\n\nORA_STIG_PROFILE\nPASSWORD_REUSE_MAX\n10"
     
     (compared using ==)
     
     Diff:
     @@ -1 +1,8 @@
     +DEFAULT
     +PASSWORD_REUSE_MAX
     +UNLIMITED
     +
     +ORA_STIG_PROFILE
     +PASSWORD_REUSE_MAX
     +10
     
    ✔  cis-oracle-db-12c-3-7: 3.7 Ensure 'DBA_USERS.PASSWORD' Is Not Set to 'EXTERNAL' for Any User (Scored)
      ✔  Oracle SQLPlus query: SELECT USERNAME FROM DBA_USERS WHERE PASSWORD='EXTERNAL'; output should eq ""

    Profile Summary: 1 successful, 1 failures, 0 skipped

### Possible Solutions

    /opt/inspec/bin/inspec exec cis_oracle_db_12c_test.rb --no-diff -t ssh://oracle@dbserver01.example.com

    Target:  ssh://oracle@dbserver01.example.com:22

    ×  cis-oracle-db-12c-3-4: 3.4 Ensure 'PASSWORD_REUSE_MAX' Is Greater than or Equal to '20' (Scored) (
     ×  Oracle SQLPlus query: SELECT PROFILE, RESOURCE_NAME, LIMIT
     FROM DBA_PROFILES
     WHERE RESOURCE_NAME='PASSWORD_REUSE_MAX'
     AND
     (
     LIMIT = 'DEFAULT'
     OR LIMIT = 'UNLIMITED'
     OR LIMIT < 20
     ); output should eq ""
     
     expected: ""
          got: "DEFAULT\nPASSWORD_REUSE_MAX\nUNLIMITED\n\nORA_STIG_PROFILE\nPASSWORD_REUSE_MAX\n10"

    ✔  cis-oracle-db-12c-3-7: 3.7 Ensure 'DBA_USERS.PASSWORD' Is Not Set to 'EXTERNAL' for Any User (Scored)
      ✔  Oracle SQLPlus query: SELECT USERNAME FROM DBA_USERS WHERE PASSWORD='EXTERNAL'; output should eq ""

    Profile Summary: 1 successful, 1 failures, 0 skipped

### LINKED ISSUE #3797: Make reporting consistent/predictable

### LINKED ISSUE #4229: Ability to mark inputs as sensitive
As an end user, I would like to be able to label an input As sensitive data. Much like the current implementation in controls. I would like to ensure that inputs marked as sensitive will not be rendered in CLI output or the other reporters. this would probably also necessitate that we add a - - show-sensitive flag for possible debugging.

### LINKED ISSUE #4997: Order/Sequence of Result
## Detailed Description
Hi,

I'm currently having an issue with the order of the result. Sometimes when I execute the profile, my result will be displayed in an orderly manner based on my controls naming and sometimes it does not. The result will be randomised and not displayed according to the naming of my controls. I tried to change the creation time of the controls file but it does not seem to work either. 

Thus, I would like to ask if there is any way I can make sure the results are always consistent and always display in the naming order of my control files
Example:
```
test-1.rb
test-2.rb
test-3.rb
```
Expecting result:
```
 ✔  test-1
 ✔  test-2
 ✔  test-3
```

## Context
I'm using this for my CIS compliance and it would be easier for me to go thru the result if the result is in sequence order.

clintoncwolfe:
## Description

Formalizes the idea of filtering the Run Data before sending it to the reporters. Then implements 3 simple requested features using the concept. These are 4 clean commits, so if there are objections to any one feature, it can be rejected cleanly. Best to walk the commits on review, as otherwise this may seem nonsensical.

 * Refactors the message limiting features (truncation and backtrace squelch) into a mixin class, so that both core reporters and plugin reporters benefit without duplicating code.
 * Implements a mechanism to mark inputs as sensitive, replacing their values with "***"
 * Implements a CLI option to suppress Diff output for textual tests
 * Implements a CLI option to control order of controls in output (not execution order)

#4972 is also easily addressable using this technique (at least for plugin-based reporters; the RSpec html reporter is a lost cause, here), but I am unable to reproduce the issue. If that changes I may add a filter to intercept ugly internals in the test messaging.

## Related Issue

Closes #4229, sensitive inputs
Closes #1535, suppress diff
Closes #4997 and Closes #3797 output ordering of controls

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2472

### DIFF
--- a/docs/dev/plugins.md
+++ b/docs/dev/plugins.md
@@ -459,6 +459,7 @@ The `run_data` object contains all data from the Chef InSpec run. Here is an ove
 |`run_data.profiles[0].inputs`| Array of Input records describing inputs present in profile|
 |`run_data.profiles[0].inputs[0].name`| String name of Input|
 |`run_data.profiles[0].inputs[0].options.required`| Boolean, whether input was required.|
+|`run_data.profiles[0].inputs[0].options.sensitive`| Boolean, true if the value of the input has been replaced with "***" |
 |`run_data.profiles[0].inputs[0].options.type`| String, type constraint on input, if any|
 |`run_data.profiles[0].inputs[0].options.value`| Value of Input|
 |`run_data.profiles[0].license`| String, name of license for the profile|
@@ -480,6 +481,11 @@ The `run_data` object contains all data from the Chef InSpec run. Here is an ove
 |`run_data.statistics.controls.total`| Integer, total count of controls|
 |`run_data.statistics.duration`| Float, time in seconds for the execution of Resources.|
 |`run_data.version`| A String, such as "4.18.108" representing the Chef InSpec version.|
+
+#### RunData Structure ChangeLog
+
+v0.1.0 - Initial version
+v0.2.0 - added `run_data.profiles[0].inputs[0].options.sensitive`
 
 ## Implementing Input Plugins
 
--- a/lib/inspec/input.rb
+++ b/lib/inspec/input.rb
@@ -171,7 +171,7 @@ module Inspec
     # are free to go higher.
     DEFAULT_PRIORITY_FOR_VALUE_SET = 60
 
-    attr_reader :description, :events, :identifier, :name, :required, :title, :type
+    attr_reader :description, :events, :identifier, :name, :required, :sensitive, :title, :type
 
     def initialize(name, options = {})
       @name = name
@@ -264,6 +264,7 @@ module Inspec
       @required = options[:required] if options.key?(:required)
       @identifier = options[:identifier] if options.key?(:identifier) # TODO: determine if this is ever used
       @type = options[:type] if options.key?(:type)
+      @sensitive = options[:sensitive] if options.key?(:sensitive)
     end
 
     def make_creation_event(options)
@@ -320,7 +321,7 @@ module Inspec
 
     def to_hash
       as_hash = { name: name, options: {} }
-      %i{description title identifier type required value}.each do |field|
+      %i{description title identifier type required value sensitive}.each do |field|
         val = send(field)
         next if val.nil?
 
@@ -334,7 +335,7 @@ module Inspec
     #--------------------------------------------------------------------------#
 
     def to_s
-      "Input #{name} with #{current_value}"
+      "Input #{name} with " + (sensitive ? "***" : "#{current_value}")
     end
 
     #--------------------------------------------------------------------------#
--- a/lib/inspec/input.rb
+++ b/lib/inspec/input.rb
@@ -335,7 +335,7 @@ module Inspec
     #--------------------------------------------------------------------------#
 
     def to_s
-      "Input #{name} with " + (sensitive ? "***" : "#{current_value}")
+      "Input #{name} with value " + (sensitive ? "*** (senstive)" : "#{current_value}")
     end
 
     #--------------------------------------------------------------------------#
--- a/lib/inspec/utils/run_data_filters.rb
+++ b/lib/inspec/utils/run_data_filters.rb
@@ -14,6 +14,7 @@ module Inspec
         @config[:runtime_config] = Inspec::Config.cached || {}
         apply_report_resize_options
         redact_sensitive_inputs
+        suppress_diff_output
       end
 
       # Apply options such as message truncation and removal of backtraces
@@ -41,6 +42,24 @@ module Inspec
             next unless i[:options][:sensitive]
 
             i[:options][:value] = "***"
+          end
+        end
+      end
+
+      # Optionally suppress diff output in the message field
+      def suppress_diff_output
+        return if @config[:runtime_config][:diff]
+
+        @run_data[:profiles]&.each do |p|
+          p[:controls]&.each do |c|
+            c[:results]&.each do |r|
+              next unless r[:message] # :message only set on failure
+
+              pos = r[:message].index("\n\nDiff:")
+              next unless pos # Only textual tests get Diffs
+
+              r[:message] = r[:message].slice(0, pos)
+            end
           end
         end
       end
--- a/lib/inspec/utils/run_data_filters.rb
+++ b/lib/inspec/utils/run_data_filters.rb
@@ -1,0 +1,46 @@
+module Inspec
+  module Utils
+    #   RunDataFilters is a mixin for core Reporters and plugin reporters.
+    # The methods operate on the run_data Hash (prior to any conversion to a
+    # full RunData object).
+    #   All methods here operate using the run_data accessor and modify
+    # its contents in place (if needed).
+    module RunDataFilters
+
+      # Long name, but we want to be clear this operates on the Hash
+      # This is the only method that client libraries need to call; any future
+      # feature growth should be handled internally here.
+      def apply_run_data_filters_to_hash
+        @config[:runtime_config] = Inspec::Config.cached || {}
+        apply_report_resize_options
+      end
+
+      # Apply options such as message truncation and removal of backtraces
+      def apply_report_resize_options
+        runtime_config = @config[:runtime_config]
+
+        message_truncation = runtime_config[:reporter_message_truncation] || "ALL"
+        @trunc = message_truncation == "ALL" ? -1 : message_truncation.to_i
+        include_backtrace = runtime_config[:reporter_backtrace_inclusion].nil? ? true : runtime_config[:reporter_backtrace_inclusion]
+
+        @run_data[:profiles]&.each do |p|
+          p[:controls].each do |c|
+            c[:results]&.map! do |r|
+              r.delete(:backtrace) unless include_backtrace
+              process_message_truncation(r)
+            end
+          end
+        end
+      end
+
+      private
+
+      def process_message_truncation(result)
+        if result.key?(:message) && result[:message] != "" && @trunc > -1 && result[:message].length > @trunc
+          result[:message] = result[:message][0...@trunc] + "[Truncated to #{@trunc} characters]"
+        end
+        result
+      end
+    end
+  end
+end
--- a/lib/inspec/utils/run_data_filters.rb
+++ b/lib/inspec/utils/run_data_filters.rb
@@ -68,6 +68,7 @@ module Inspec
       # Optionally sort controls within each profile in report
       def sort_controls
         sort_type = @config[:runtime_config][:sort_results_by]
+        return unless sort_type
         return if sort_type == "none"
 
         @run_data[:profiles]&.each do |p|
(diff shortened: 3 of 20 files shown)


lib/inspec/input.rb
```diff
@@ -334,7 +335,7 @@ def to_hash
     #--------------------------------------------------------------------------#
 
     def to_s
-      "Input #{name} with #{current_value}"
+      "Input #{name} with " + (sensitive ? "***" : "#{current_value}")
```
james-stocks: The grammar here was already a little terse, it should have said `with value` instead of just `with`.
Now we handle the sensitive case it would be helpful to make this clearer.

Suggest that for non-sensitive it should look like `Input php_username with value php_admin` and for the sensitive case `Input php_password with value *** (sensitive)`

I was also wondering if it would be helpful to handle the case where the input is sensitive by has no value. In that case it would be misleading to show *** as that suggests there is a value but it's been redacted.
However I'm not sure on the use case for that because if I try to include a value-less input in a control, the control gets skipped.

lib/inspec/utils/run_data_filters.rb
```diff
@@ -0,0 +1,97 @@
+module Inspec
+  module Utils
+    #   RunDataFilters is a mixin for core Reporters and plugin reporters.
```
james-stocks, line 3: This is nice!

lib/inspec/input.rb
clintoncwolfe: OK, I've added "with value" and "(sensitive)".

lib/inspec/utils/run_data_filters.rb
```diff
@@ -0,0 +1,97 @@
...
+                r[:message] = r[:message][0...trunc] + "[Truncated to #{trunc} characters]"
+              end
+              r
+            end
+          end
+        end
+      end
+
+      # Find any inputs with :sensitive = true and replace their values with "***"
+      def redact_sensitive_inputs
```
james-stocks, line 40: I was thinking that this replacement happens before a user's plugin `render` method runs. That means that a user's custom reporter has no opportunity to work with the sensitive value.
I think this is OK - I can't think of a sensible use case where someone would want their custom reporter logic to depend on the non-redacted sensitive value.

docs/dev/plugins.md
```diff
@@ -459,6 +459,7 @@ The `run_data` object contains all data from the Chef InSpec run. Here is an ove
 |`run_data.profiles[0].inputs`| Array of Input records describing inputs present in profile|
 |`run_data.profiles[0].inputs[0].name`| String name of Input|
 |`run_data.profiles[0].inputs[0].options.required`| Boolean, whether input was required.|
+|`run_data.profiles[0].inputs[0].options.sensitive`| Boolean, true if the value of the input has been replaced with "***" |
```
james-stocks, line 462: This description might be confusing because the run_data from the runner has not yet redacted the sensitive value. That only happens in the base reporter.
So for someone developing reporter plugins, this description is 100% true; but if the reader is interested in the run_data API itself, or some other part of InSpec, this may be confusing because it says the value has been replaced with *** but they may see the real value in memory.

clintoncwolfe: Filtering happens to the Hash, then the Hash is converted to the RunData object. By the time the RunData object sees this value, it has indeed already been replaced.

This documentation is for the RunData object as seen by the reporter plugin system. The docs are in the file `docs/dev/plugins.md`, in the section regarding how to write a reporter plugin.

Schwad:
This is looking pretty good from my end, thanks much for this @clintoncwolfe ! 😄  Will request a docs review as we do touch docs with `plugin.md` and `inputs.md` just to make sure we're good to go.

lib/inspec/utils/run_data_filters.rb
```diff
@@ -0,0 +1,97 @@
...
+              r[:message] = r[:message].slice(0, pos)
+            end
+          end
+        end
+      end
+
+      # Optionally sort controls within each profile in report
+      def sort_controls
+        sort_type = @config[:runtime_config][:sort_results_by]
+        return if sort_type == "none"
```
james-stocks, line 72: Might want a `return unless sort_type` in case this option isn't set. (If InSpec is running in kitchen or audit cookbook, I _think_ it bypasses the default options defined for the CLI)

lib/inspec/utils/run_data_filters.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          case sort_type
+          when "control"
+            p[:controls].sort_by! { |c| c[:id] }
+          when "random"
+            p[:controls].shuffle!
+          when "file"
+            # Sort the controls by file, but preserve order within the file.
+            # Files are called "groups" in the run_data, and the filename is in the id.
+            sorted_control_ids = p[:groups].sort_by { |g| g[:id] }.map { |g| g[:controls] }.flatten
```
james-stocks, line 86: I'd like to test this out (will assign PR to myself to do so) because I think what some have been asking for is InSpec to give results in the order that they wrote their tests, without jumbling them. 
If that's what this option does, then users might expect sort `none` to _not_ jumble results in the way it currently does. 

Sort by `control` won't work for benchmarks like CIS because they are ordered by section number but those come out wrong when sorted alphanumerically

```
 %w{cis_1.10_ssh_options cis_1.9_firewall_settings cis_1.8_login_settings}.sort
=> ["cis_1.10_ssh_options", "cis_1.8_login_settings", "cis_1.9_firewall_settings"]
```

mjingle (approved):
This PR is good to go from a docs perspective. Thanks @clintoncwolfe! :)

lib/inspec/utils/run_data_filters.rb
james-stocks: Beg pardon, InSpec does NOT currently jumble results inside one file. I think the case that bugs customers is when they have multiple files (commonly one per control) - and that's exactly what you're fixing here.

Would it be possible to make `file` the default, so users get this desired behaviour for free? 

This does still have the problem mentioned in my last comment - control `1.11` from `cis-1.11.rb`will appear before control `1.2` from `cis-1.2.rb` - I don't think that should be fixed in this PR. Unless maybe to document that alphanumeric ordering isn't magic and InSpec can only put 1.11 before 1.2.

aaronlippold:
Any thoughts on when this will merge?

lib/inspec/utils/run_data_filters.rb
```diff
@@ -0,0 +1,97 @@
...
+        @run_data[:profiles]&.each do |p|
+          p[:controls] ||= []
+          p[:groups] ||= []
+
+          case sort_type
+          when "control"
+            p[:controls].sort_by! { |c| c[:id] }
+          when "random"
+            p[:controls].shuffle!
+          when "file"
```
Schwad, line 83:
```suggestion
          else
```

Schwad (approved):
@clintoncwolfe sorry this took so long.

I believe this is looking good and we can move forward. 

I put in one suggestion which I think would implement Stocksy's recommendation to make `"file"` the default. Also I think a rebase would fix the testing failures on 
buildkite/inspec-inspec-master-verify because they all seem to be hitting one issue. Finally would you like to take a quick peek at the merge conflicts? Otherwise I can!

Thanks much again for this and looking forward to getting it in!

lib/inspec/utils/run_data_filters.rb
Schwad: This ☝️ might be what's needed to accomplish this

> Would it be possible to make file the default, so users get this desired behaviour for free?

https://github.com/inspec/inspec/pull/5054#discussion_r454275121

lib/inspec/utils/run_data_filters.rb
clintoncwolfe: Adding, thanks!

lib/inspec/utils/run_data_filters.rb
clintoncwolfe: I've made `file` the default by telling Thor that it should be the default value.

Schwad:
Sorry this took so long, thanks much for your work @clintoncwolfe ! :)

PR #5062, opened 2020-06-02
Make the JSON-min Reporter into Plugin

clintoncwolfe:
## Description

* Converts the json-min reporter from a core reporter into a plugin reporter.
* Modifies the HashLike interface to RunData to hide nil-valued keys
* Leaves the json-min functional test in the main core area, though it could be moved

Note that the json-min reporter is unrelated to the JSON / Automate reporters; changing it does not impact them.

## Related Issue

Partial fix for #3667

## Types of changes

## Checklist:


james-stocks (approved):
This is great, thanks @clintoncwolfe 
Leaving the tests where they are seems reasonable.

PR #5063, opened 2020-06-03
Make progress on `skip_windows!`

Schwad:
72 `skip_windows` needing resolution OR alternative documentation upon investigation.

When `pry`'ing in a windows OS this did pass, checking against CI.

Aha! Link: https://chef.aha.io/features/SH-2149

### DIFF
--- a/test/functional/git_fetcher_test.rb
+++ b/test/functional/git_fetcher_test.rb
@@ -12,7 +12,7 @@ describe "running profiles with git-based dependencies" do
   fixture_repos = %w{basic-local git-repo-01}
 
   before(:all) do
-    skip_windows! # Right now, this is due to symlinking
+    skip_windows! # Right now, this is due to symlinking, break executes on L24 <nickchecked>
 
     # We need a git repo for some of the profile test fixtures,
     # but we can't store those directly in git.
--- a/test/functional/inspec_exec_json_test.rb
+++ b/test/functional/inspec_exec_json_test.rb
@@ -15,8 +15,6 @@ describe "inspec exec with json formatter" do
     _(schema.validate(data).to_a).must_equal []
 
     _(out.stderr).must_be_empty
-
-    skip_windows!
     assert_exit_code 0, out
   end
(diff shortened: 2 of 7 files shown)


Schwad:
I'm going to leave this open and have a look at a few more windows tests with this branch

Schwad:
(now that my local instance is working)

clintoncwolfe:
Oddly, that's often been my experience with these as well - the skips don't seem to actually be protecting a failing test.  I'm not saying to remove them thoughtlessly, but there might be fewer skips to fix than one might think.

Schwad:
> Oddly, that's often been my experience with these as well - the skips don't seem to actually be protecting a failing test. I'm not saying to remove them thoughtlessly, but there might be fewer skips to fix than one might think.

My hope is to work on this where I can and *hopefully* document the hairy ones and slowly but surely remove the unneeded ones... And hopefully get us to a smaller number over the next two weeks.

I'm not going carte blanche as I want to ensure a few at a time pass on my local instance then also ensure they pass all our CI checks as well (as I'm only doing one windows locally)

Schwad:
Updated this to WIP so it's not merged. Going to work through `skip_windows` one at a time with pry in the CLI, find any removals and quick fixes, document rest, and request review when I think we can close off a bunch.

james-stocks (approved):
@Schwad this is a great improvement, thanks!

It looks like the branch just needs rebased to a. avoid the SSL test fails b. pass the DCO check (also if you think the 6 commits could be squashed - no harm in keeping them separate if you like).

Schwad:
Thanks! I had just squashed on the first rebase but repushed with master rebased in also to catch our SSL test fails too 👍 

Will merge when it passes unless I'm missing something

Schwad:
Post-merge note for future work:

Specs that pass on our local windows machines but complain in CI should be flagged and handled separately to others, as they are failures in the pipeline process. (Plus could be long and painful to debug ourselves)

PR #5069, opened 2020-06-04
Make CLI reporter into a plugin

clintoncwolfe:
## Description

This just makes the CLI reporter into a plugin. Some mild refactoring on a helper class, too. I made no attempt to move the functional test for this - it's `inspec_exec_test` and that has sooooo many things mixed up in it.

There is a cherry-picked commit from the JSON-min PR regarding the rundata unit test fixture. if that turns out to be needed on all of these, I'll split that out and maybe that can merge first.

## Related Issue

Partial fix for #3667

## Types of changes

## Checklist:


lib/inspec/reporters.rb
```diff
@@ -13,8 +12,6 @@ def self.render(reporter, run_data)
     name, config = reporter.dup
     config[:run_data] = run_data
     case name
-    when "cli"
-      reporter = Inspec::Reporters::CLI.new(config)
     when "json"
```
Schwad, line 15: Would part of the long term goal here see the entire case statement replaced by the line we're currently catchingn in the else? https://github.com/inspec/inspec/blob/afdcc9d2a1b76a4f0051e527015e2029095e458a/lib/inspec/reporters.rb#L31

lib/plugins/inspec-reporter-cli/README.md
```diff
@@ -0,0 +1,12 @@
+# cli reporter
```
Schwad, line 1: Looks good, do these sync up with our public `www` documentation? If so, do we need docs review or is this small enough not to need it?

lib/plugins/inspec-reporter-cli/lib/inspec-reporter-cli.rb
```diff
@@ -0,0 +1,13 @@
+require_relative "inspec-reporter-cli/version"
+
+module InspecPlugins
+  module CliReporter
+    class Plugin < ::Inspec.plugin(2)
+      plugin_name :'inspec-reporter-cli'
+      reporter :cli do
+        require_relative "inspec-reporter-cli/reporter"
+        InspecPlugins::CliReporter::Reporter
```
Schwad, line 9: One thing I reallly like about getting this conversion done is it'll help give further examples for developers building their own InSpec reporters in the future (including me! 😄 )

Schwad (approved):
Overall this looks great to me @clintoncwolfe just a couple comments, it's exciting watching these reporters get plugin-itized! 😄  👍

lib/inspec/reporters.rb
clintoncwolfe: Yes, that's right.

lib/plugins/inspec-reporter-cli/README.md
clintoncwolfe: This *should* be simply an internal refactor - no change to the behavior of Chef InSpec and no change to www docs. I simply added a README because there is usually a README in each plugin directory.

james-stocks (approved):
Sorry for taking a while to look at this; this is awesome, thank you @clintoncwolfe :)

PR #5073, opened 2020-06-04
Update installer license text

james-stocks:
Copied the correct license text over from chef/chef

Fixes 5066

Kicked off an adhoc build to test the installers from this branch: https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/140#eb343d2c-9fea-42da-adba-222c6bf0831a

Aha! Link: https://chef.aha.io/features/SH-2160

### DIFF
--- a/omnibus/resources/inspec/msi/assets/LICENSE.rtf
+++ b/omnibus/resources/inspec/msi/assets/LICENSE.rtf
@@ -1,197 +1,128 @@
 {\rtf1\ansi\deff0{\fonttbl{\f0\fnil\fcharset0 Courier New;}}
 
-{\*\generator Msftedit 5.41.21.2500;}\viewkind4\uc1\pard\qc\lang1033\f0\fs20 Apache License\par
+{\*\generator Msftedit 5.41.21.2500;}\viewkind4\uc1\pard\qc\lang1033\f0\fs20 Software End User License Agreement\par
 
-Version 2.0, January 2004\par
+(Personal, Non-Commercial, Experimental)\par
 
-http://www.apache.org/licenses/\par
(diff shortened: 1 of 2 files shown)


clintoncwolfe (commented):
I think correcting the license displayed in the GUI installer is absolutely the correct thing to do. I wish we had caught this last April.

I think we need to consider whether accepting the Personal-Use license in the GUI installer should also write the license-accepted marker file. If not, the user will be prompted again on first invocation, which is bad UX.

Second, is this the right license? For our commercial customers, my understanding is they are bound by a different, commercial agreement. Do we need to offer a choice between a personal use license and a commercial license? Do we need to validate the commercial license....?

Third, I'm pretty sure we need to have Legal review this before we make a change. We don't want to change to the wrong thing, and then have to change again, further eroding trust.

james-stocks:
@clintoncwolfe -

1. This is a great idea. I imagine we can get Omnibus to do this for us, but I'd prefer to do it outside this PR.

2. This is the same license included with chef/chef and I don't think omnibus supports offering multiple licenses.

3. good idea to check with Legal, I have CC'ed you on a query with them.

Schwad (dismissed):
I'll defer to y'alls judgment on the best ways to approach licensing but cheers for getting this updated, following legal check looks good to me 👍  <edit: approval removed>

james-stocks:
@clintoncwolfe I just had a call with Legal - they confirmed that the current text is correct

james-stocks:
@clintoncwolfe I've also tested the OSX and Windows installers from the ad-hoc build; license is definitely updated

clintoncwolfe (approved):

PR #5078, opened 2020-06-10
Change ffi pinning

james-stocks:
ffi 1.13.1 reverts the breaking change in 1.13.0
We can now use any version other than specifically 1.13.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem "inspec", path: "."
 # in it in order to package the executable. Hence the odd backwards dependency.
 gem "inspec-bin", path: "./inspec-bin"
 
-gem "ffi", [">= 1.9.14", "< 1.13"] # 1.13 does not work on Windows: https://github.com/ffi/ffi/issues/784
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -9,7 +9,7 @@ gem "inspec", path: "."
 # in it in order to package the executable. Hence the odd backwards dependency.
 gem "inspec-bin", path: "./inspec-bin"
 
-gem "ffi", [">= 1.9.14", "< 1.13"] # 1.13 does not work on Windows: https://github.com/ffi/ffi/issues/784
+gem "ffi", ">= 1.9.14", "!= 1.13.0"
```
james-stocks, line 12: The 1.9.14 seems to be there for historical reasons. If 1.13.1 is known to work then I think we should just change this to `"ffi", ">= 1.13.1", "< 2.0"`

Schwad (approved):
Looks great, thanks for updating this so quick after they got the patch

PR #5084, opened 2020-06-10
Convert JUnit Reporter to a Plugin

clintoncwolfe:
## Description

Converts the JUnit reporter to a plugin, including the RunData accessors to be method-based. Does nothing to fix any problems with the JUnit reporter; ~there may be a separate PR for that.~ see #5085

## Related Issue

Partial implementation of #3667 

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2332

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -344,7 +344,6 @@ module Inspec
         cli
         json
         json-automate
-        junit
         yaml
       }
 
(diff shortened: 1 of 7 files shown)


Schwad:
This looks good to me, thanks @clintoncwolfe ! I'm not sure if we should hold fire on merging plugin PRs until we resolve https://github.com/inspec/inspec/issues/5102

clintoncwolfe:
JUnit is not a default reporter - you must explicitly request it, so it should not have the problem that CLI had on #5102.

Schwad (approved):
@clintoncwolfe sounds good, in that case thank you very much!

PR #5098, opened 2020-06-17
Stop building habitat for InSpec 3

james-stocks:
A habitat project can only track one major version at a time and `chef/inspec` is now InSpec 4.
The habitat config for this branch needs to build `chef/inspec3` instead, and we don't need to build/release it while no-one needs it.


Schwad (approved):
Looks great - thanks for your work on 3! 👍

Issue #5099, opened 2020-06-18
Upgrade to inspec/inspec-core 4.20.6 brakes our CI 


crazymind1337:
After our CI automatically upgraded the Gems inspec and inspec-core it fails with an error:

```sh
Installing inspec-core 4.20.6 (was 4.20.2)
Installing inspec 4.20.6 (was 4.20.2)
```

```log
Caused by:
Inspec::Plugin::V2::LoadError: Plugin hooks search returned zero results for filter {:plugin_type=>:reporter, :activator_name=>:cli}
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/plugin/v2/registry.rb:77:in `find_activator'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/reporters.rb:28:in `render'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/runner.rb:137:in `block in render_output'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/runner.rb:136:in `each'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/runner.rb:136:in `render_output'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/runner.rb:161:in `run_tests'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/inspec-core-4.20.6/lib/inspec/runner.rb:130:in `run'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/kitchen-inspec-2.0.0/lib/kitchen/verifier/inspec.rb:105:in `call'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/test-kitchen-2.5.2/lib/kitchen/instance.rb:459:in `block in verify_action'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/test-kitchen-2.5.2/lib/kitchen/instance.rb:552:in `synchronize_or_call'
/home/someuser/.rvm/gems/ruby-2.5.5/gems/test-kitchen-2.5.2/lib/kitchen/instance.rb:514:in `block in action'
[...]

PR #5100, opened 2020-06-18
Revert "Make CLI reporter into a plugin"

Schwad:
Reverts inspec/inspec#5069

Currently debugging https://github.com/inspec/inspec/issues/5099, if no hotfix found this will revert


james-stocks (approved):
Thanks - will discuss in Slack what we need to do to get this implemented again.

PR #5124, opened 2020-06-29
feat(sensitive): add sensitive output to mysql and postgres sessions

fzipi:
Signed-off-by: Felipe Zipitria <fzipitria@perceptyx.com>

## Description

Sessions in postgres or mysql end in the `command` resource. While the resource supports using regex to prevent showing sensitive information, none of the resources for databases prevent showing sensitive information.

This PR adds an additional `sensitive` parameter, that uses a proper (I hope) regex to show `REDACTED` in the command output.

## Related Issue

Related to #3185.

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2448

### DIFF
--- a/lib/inspec/resources/mysql_session.rb
+++ b/lib/inspec/resources/mysql_session.rb
@@ -4,6 +4,23 @@ require "inspec/resources/command"
 require "shellwords"
 
 module Inspec::Resources
+  class Lines
+    attr_reader :output
+
+    def initialize(raw, desc)
+      @output = raw
+      @desc = desc
+    end
+
+    def lines
+      output.split("\n")
+    end
+
+    def to_s
+      @desc
+    end
+  end
+
   class MysqlSession < Inspec.resource(1)
     name "mysql_session"
     supports platform: "unix"
@@ -28,15 +45,17 @@ module Inspec::Resources
 
     def query(q, db = "")
       mysql_cmd = create_mysql_cmd(q, db)
-      cmd = inspec.command(mysql_cmd)
+      cmd = if !@pass.nil?
+              inspec.command(mysql_cmd, redact_regex: /(mysql -u\w+ -p).+(\s-(h|S).*)/)
+            else
+              inspec.command(mysql_cmd)
+            end
       out = cmd.stdout + "\n" + cmd.stderr
-      if out =~ /Can't connect to .* MySQL server/ || out.downcase =~ /^error /
-        # skip this test if the server can't run the query
-        warn("Can't connect to MySQL instance for SQL checks.")
+      if cmd.exit_status != 0 || out =~ /Can't connect to .* MySQL server/ || out.downcase =~ /^error:.*/
+        Lines.new(out, "MySQL query with errors: #{q}")
+      else
+        Lines.new(cmd.stdout.strip, "MySQL query: #{q}")
       end
-
-      # return the raw command output
-      cmd
     end
 
     def to_s
--- a/test/unit/resources/mysql_session_test.rb
+++ b/test/unit/resources/mysql_session_test.rb
@@ -1,6 +1,7 @@
 require "helper"
 require "inspec/resource"
 require "inspec/resources/mysql_session"
+require "inspec/resources/command"
 
 describe "Inspec::Resources::MysqlSession" do
   it "verify mysql_session escaped login details with single quotes correctly" do
@@ -16,5 +17,13 @@ describe "Inspec::Resources::MysqlSession" do
 
     _(resource.send(:create_mysql_cmd, "SELECT 1 FROM DUAL;"))
       .must_equal('mysql -h localhost -s -e "SELECT 1 FROM DUAL;"')
+  end
+  it "verify mysql_session redacts output" do
+    cmd = %q{mysql -uroot -p\'\%\"\'\"\&\^\*\&\(\)\'\*\% -h localhost -s -e "SELECT 1 FROM DUAL;"}
+    options = { redact_regex: /(mysql -u\w+ -p).+(\s-(h|S).*)/ }
+    resource = load_resource("command", cmd, options)
+
+    expected_to_s = %q{Command: `mysql -uroot -pREDACTED -h localhost -s -e "SELECT 1 FROM DUAL;"`}
+    _(resource.to_s).must_equal(expected_to_s)
   end
 end
--- a/test/unit/resources/postgres_session_test.rb
+++ b/test/unit/resources/postgres_session_test.rb
@@ -1,6 +1,7 @@
 require "helper"
 require "inspec/resource"
 require "inspec/resources/postgres_session"
+require "inspec/resources/command"
 
 describe "Inspec::Resources::PostgresSession" do
   it "verify postgres_session create_psql_cmd with a basic query" do
@@ -10,5 +11,13 @@ describe "Inspec::Resources::PostgresSession" do
   it "verify postgres_session escaped_query with a complex query" do
     resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1")
     _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
+  end
+  it "verify postgres_session redacts output" do
+    cmd = %q{PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c "SELECT current_setting('client_min_messages')"}
+    options = { redact_regex: /(PGPASSWORD=').+(' psql .*)/ }
+    resource = load_resource("command", cmd, options)
+
+    expected_to_s = %q{Command: `PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c "SELECT current_setting('client_min_messages')"`}
+    _(resource.to_s).must_equal(expected_to_s)
   end
 end
--- a/lib/inspec/resources/postgres_session.rb
+++ b/lib/inspec/resources/postgres_session.rb
@@ -47,7 +47,7 @@ module Inspec::Resources
 
     def query(query, db = [])
       psql_cmd = create_psql_cmd(query, db)
-      cmd = inspec.command(psql_cmd)
+      cmd = inspec.command(psql_cmd, redact_regex: /(PGPASSWORD=').+(' psql .*)/)
       out = cmd.stdout + "\n" + cmd.stderr
       if cmd.exit_status != 0 || out =~ /could not connect to .*/ || out.downcase =~ /^error:.*/
         Lines.new(out, "PostgreSQL query with errors: #{query}")


lib/inspec/resources/mysql_session.rb
```diff
@@ -16,19 +16,24 @@ class MysqlSession < Inspec.resource(1)
       end
     EXAMPLE
 
-    def initialize(user = nil, pass = nil, host = "localhost", port = nil, socket = nil)
+    def initialize(user = nil, pass = nil, host = "localhost", port = nil, socket = nil, sensitive: false)
```
Schwad: Hello! 👋  Thanks for the PR!

This is an interesting proposition. If we were to extend the functionality that we enjoy from the `command` resource to the `mysql_session` resource, would it be better to keep the DSL the same?

i.e. Instead of adding a new `sensitive: true` option, should we just make it the same as the `command` resource with `redact_regex`? ( Looking at the docs here: https://www.inspec.io/docs/reference/resources/command/ )

So maybe your code could look like `sql = mysql_session('my_user','password','db.example.com',  redact_regex: /(-p ).*( -d)/)`

I do think longer term it might be nice to support both a `redact_regex` (which gives the user lots of redacting power) as well as something to simply blocklist showing passwords, on both `command` and `mysql_session` and potentially other resources.

Side note: I had a look at your CI run/specs and I'm not sure if it actually is having the output you expected? https://buildkite.com/chef-oss/inspec-inspec-master-verify/builds/2000#5ef65086-d1db-43de-9ab9-d6deabc86869

Thanks again so much for sending this in, really appreciate it, and look forward to hearing from you! 🙏

test/unit/resources/mysql_session_test.rb
```diff
@@ -17,4 +17,13 @@
     _(resource.send(:create_mysql_cmd, "SELECT 1 FROM DUAL;"))
       .must_equal('mysql -h localhost -s -e "SELECT 1 FROM DUAL;"')
   end
+  it "verify mysql_session correctly redacts sensitive passwords" do
+    resource = load_resource("mysql_session",
+                             "root",
+                             "MYPASSWORD",
+                             sentitive: true)
```
Schwad: I think this one might pass but just need to catch a quick typo

test/unit/resources/postgres_session_test.rb
```diff
@@ -11,4 +11,8 @@
     resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1")
     _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
   end
+  it "verify postgres_session create_psql_cmd using sensitive argument" do
+    resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1", sensitive: true)
+    _(resource.send(:create_psql_cmd, "SELECT * FROM STUDENTS;", ["testdb"])).must_equal "PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ STUDENTS\\;"
```
Schwad: For some reason the spec isn't quite redacting here 🤔 

```

1) Failure:
--
  | Inspec::Resources::PostgresSession#test_0003_verify postgres_session create_psql_cmd using sensitive argument [/workdir/test/unit/resources/postgres_session_test.rb:16]:
  | expected | 0s
  | actual | 1s
  | @@ -1 +1 @@
  | -"PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ STUDENTS\\;"
  | +"PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ STUDENTS\\;"
  |  
```

test/unit/resources/mysql_session_test.rb
fzipi: Definitely! Good catch.

lib/inspec/resources/mysql_session.rb
fzipi: I don't foresee anything sensitive other than password for the time being. So maybe it makes sense to just use a proper regex to do that and don't let the user mess with it. Another option was to use the syntax below, but sadly I don't know much about how to get there:
```
describe mysql_session('user','password','host'), :sensitive
```
For the time being, your proposal is very straightforward. Maybe it makes sense at first let the `redact_regex` be the initial available, and then we can think for an additional syntax that can be used like the `:sensitive` here.

test/unit/resources/postgres_session_test.rb
fzipi: Yeah, I tested it using https://rubular.com/, but seems like I still need to change something there.

lib/inspec/resources/mysql_session.rb
```diff
@@ -16,19 +16,24 @@ class MysqlSession < Inspec.resource(1)
...
       @socket = socket
+      @sensitive = sensitive
       init_fallback if user.nil? || pass.nil?
       skip_resource("Can't run MySQL SQL checks without authentication") if @user.nil? || @pass.nil?
     end
 
     def query(q, db = "")
       mysql_cmd = create_mysql_cmd(q, db)
-      cmd = inspec.command(mysql_cmd)
+      cmd = if @sensitive && !@pass.nil?
```
james-stocks: We could consider always passing the redact_regex and not exposing an option to the user. I can't imagine a case where the user will need to see their password echoed back to them.

fzipi: @james-stocks I like that.

My only use case was just debug to see if you have the right pass (which you should, but you know...). But I guess if you actually use `-l debug` you are good to go?

Schwad: That is an interesting option @james-stocks ! 🤔  I would definitely support always using `redact_regex` in the cases you've identified @fzipi if that's something you wouldn't mind doing.

I would be very surprised if someone were to come back in the future saying "hey, why can't I see my password in the results?" 😅

test/unit/resources/postgres_session_test.rb
Schwad: Screenshot 2020-07-01 at 12 42 10

This is very interesting.... I just played around on rubular as well and it seems like you're right on here.. 🤔  I wonder if it might be worth actually prying in and seeing why it doesn't like that regex.

lib/inspec/resources/mysql_session.rb
fzipi: Sounds good to me! So I'm removing the extra parameter, and always redacting the password.

fzipi:
After diving a little bit more into both `postgres_session` and `mysql_session`, I think postgres manages output better than mysql by using a helper class `Lines` that only deals with output. `mysql_session` includes the full command on its output, but I think it is more elegant to just show the query like postgres does.

```
❯ bundle exec inspec exec test_mysql_session.rb

Profile: tests from test_mysql_session.rb (tests from test_mysql_session.rb)
Version: (not specified)
Target:  local://

  ✔  check-mysql-session: Check mysql_session redacted
     ✔  MySQL query: show databases like 'demoa'; output is expected to eq "demoa"


```

vs.

```
  ✔  check-mysql-session: Check mysql_session redacted
     ✔  Command: `mysql -uinspec -pREDACTED -h 127.0.0.1 -s -e "show databases like 'demoa';"` output
```

So I reformatted that one using the same `Lines` from postgres_session.

fzipi:
Regarding tests in postgres, using this file:
```
control 'check-postgres-session' do
  impact 0.3
  title 'Check postgres_session redacted'
  sql = postgres_session('inspec', 'inspec', '127.0.0.1')

  describe sql.query('SELECT * FROM actor LIMIT 1;', ['inspec']) do
    #require 'pry'; binding.pry;
    its('output') { should eq('1|Penelope|Guiness|2013-05-26 14:47:57.62') }
  end
end
```

This is what I see using `pry`:
```
From: /Workspace/inspec/lib/inspec/resources/postgres_session.rb:52 postgres_session#query:

    48: def query(query, db = [])
    49:   psql_cmd = create_psql_cmd(query, db)
    50:   cmd = inspec.command(psql_cmd, redact_regex: /(PGPASSWORD=').+(' psql .*)/)
    51:   require 'pry'; binding.pry;
 => 52:   out = cmd.stdout + "\n" + cmd.stderr
    53:   if cmd.exit_status != 0 || out =~ /could not connect to .*/ || out.downcase =~ /^error:.*/
    54:     Lines.new(out, "PostgreSQL query with errors: #{query}")
    55:   else
    56:     Lines.new(cmd.stdout.strip, "PostgreSQL query: #{query}")
    57:   end
    58: end

[1] pry(#<#<Class:0x00000000048a7dd0>>)> cmd.to_s
=> "Command: `PGPASSWORD='REDACTED' psql -U inspec -d inspec -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ actor\\ LIMIT\\ 1\\;`"

So this looks like it is doing things properly.

**Comment by fzipi** [comment]
Don't know why the user test for freebsd is failing :thinking:

**Comment by fzipi** [comment]
Hey @james-stocks @Schwad ! Can you take a look at this now?

File: test/unit/resources/postgres_session_test.rb [hunk 5]
```diff
@@ -5,10 +5,14 @@
...
   end
   it "verify postgres_session escaped_query with a complex query" do
     resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1")
-    _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
+    _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
+  end
+  it "verify postgres_session create_psql_cmd using password shows REDACTED" do
+    resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1", sensitive: true)
```
Schwad: Since we no longer use the `sensitive: true` approach this test is failing with: `ArgumentError: wrong number of arguments (given 4, expected 2..3)`

test/unit/resources/postgres_session_test.rb
```diff
@@ -5,10 +5,14 @@
...
   it "verify postgres_session create_psql_cmd with a basic query" do
     resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1")
-    _(resource.send(:create_psql_cmd, "SELECT * FROM STUDENTS;", ["testdb"])).must_equal "PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ STUDENTS\\;"
+    _(resource.send(:create_psql_cmd, "SELECT * FROM STUDENTS;", ["testdb"])).must_equal "PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ \\*\\ FROM\\ STUDENTS\\;"
   end
   it "verify postgres_session escaped_query with a complex query" do
     resource = load_resource("postgres_session", "myuser", "mypass", "127.0.0.1")
-    _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='mypass' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
+    _(resource.send(:create_psql_cmd, "SELECT current_setting('client_min_messages')", ["testdb"])).must_equal "PGPASSWORD='REDACTED' psql -U myuser -d testdb -h 127.0.0.1 -A -t -c SELECT\\ current_setting\\(\\'client_min_messages\\'\\)"
```
Schwad: These and the other tests are failing at the moment because the password is still getting sent out. I believe your changes will work on the #query method which isn't tested here, but on the #create_psql_cmd method we still produce the plaintext password which fails this exact test.

I would recommend changing this line of the resource to `PGPASSWORD='REDACTED'`

https://github.com/inspec/inspec/blob/8b2c660a474b008b611dd161ad30c6f789372956/lib/inspec/resources/postgres_session.rb#L67

test/unit/resources/mysql_session_test.rb
```diff
@@ -9,12 +9,20 @@
                              %q{'%"'"&^*&()'*%})
 
     _(resource.send(:create_mysql_cmd, "SELECT 1 FROM DUAL;"))
-      .must_equal(%q{mysql -uroot -p\'\%\"\'\"\&\^\*\&\(\)\'\*\% -h localhost -s -e "SELECT 1 FROM DUAL;"})
+      .must_equal(%q{mysql -uroot -pREDACTED -h localhost -s -e "SELECT 1 FROM DUAL;"})
```
Schwad: This is also failing spec as noted in another comment. Since this is hitting the `create_mysql_cmd` method, the password is not being escaped. I might recommend using `command += " -pREDACTED" unless @pass.nil?`at this line: https://github.com/Perceptyx/inspec/blob/8b2c660a474b008b611dd161ad30c6f789372956/lib/inspec/resources/mysql_session.rb#L79

Schwad (dismissed):
Thanks again for putting this together for us, @fzipi ! It's looking really good and hopefully not too far from merging! 

I cloned a copy of your code to my machine today and walked through some of the failing specs to try and help you get this across the line. It appears that your failing tests are due to some other areas still producing a plain text password. I have included my suggestions on how I think best to solve these tests. 

We should be close to ready once those changes come in and we pass the tests! 👍

test/unit/resources/postgres_session_test.rb
fzipi: But we need that password, right? Otherwise the actual `psql` command won't work....

I guess we can test both, so we are sure the password is included in the query, but it is actually being REDACTED in the query.

Schwad:
Aha - good point. Thank you.

In that case, should we revert those tests where the password *shouldn't*
be redacted and create some new tests that hit the `#query` method instead
to ensure they're redacting there? It appears this assertion is against the
`#create_psql_cmd` method where we aren't doing the redacting. I think a
functional test might be able to cover this? 🤔

Schwad:
I'm 100% happy whatever you think is the best approach too, as long as we
get the pipeline green too. I'll make sure to get full builds going as well
(there's part of it I have to manually trigger as well)

test/unit/resources/mysql_session_test.rb
fzipi: Yeah, again, that is the actual password :smile:  We should not remove it.

I will do separate tests in both the create cmd, and the query.

fzipi:
Did a rebase and pushed again to see if Windows tests pass now.

Schwad:
@fzipi cheers it looks like you are passing your tests, BUT we had a breaking change introduced with a dependency update (`json_schemer`) that released on July 12th and breaks our support for Ruby 2.4.

I have a PR in to resolve this (which is also hitting an external issue) https://github.com/inspec/inspec/pull/5144 but as soon as we can get this merged one final rebase should solve this. Sorry for this, it seems when it rains it pours! 🙏  👍

Schwad:
Okay - now we should be good, the changes fixing ruby 2.4 are in master!

fzipi:
Hey @Schwad ! Any news on this?

Schwad:
Thank you @fzipi ! Apologies for the delay on my response there.

Looks awesome two small points (and I swear I'll be done!)

First - I was thinking it would be nice to have *some* test along with this. So we at least document how you'd like the string to work would you be able to write a test like this:

https://github.com/inspec/inspec/blob/138935f9bbd22fe33e97c24f8e4423ce764797b0/test/unit/resources/command_test.rb#L49

(literally just against the command input and expected output), and put one each in the `mysql` and `postgres` test files? It could literally even be a copy paste but updated for the `redact_regex` we're using here. 

Last - I pointed to this in another issue, would you be able to squash this PR then into one commit for merging? 😄 

Cheers!

fzipi:
@Schwad Added those tests! :)

Schwad:
@fzipi fantastic!

Only one small linting issue, part of the Chef styleguide (which is unique as most other ruby repos don't do this) is preferring double quotes over single:


https://buildkite.com/chef-oss/inspec-inspec-master-verify/builds/2101#48761ae6-76cc-4c08-a007-d4175441a90e

```
it 'verify mysql_session redacts output' do
--
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | test/unit/resources/mysql_session_test.rb:24:30: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  | resource = load_resource('command', cmd, options)
  | ^^^^^^^^^
  | test/unit/resources/postgres_session_test.rb:15:6: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  | it 'verify postgres_session redacts output' do
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | test/unit/resources/postgres_session_test.rb:18:30: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  | resource = load_resource('command', cmd, options)
  | ^^^^^^^^^


```

fzipi:
Sure, np. I did run `rubocop` and that changed lots of other stuff.

BTW, shouldn't this be configured in the `.rubocop.yml`?

Schwad:
Thanks and good question. We have a lot of repositories we look after and have our pinned rubocop rules defined in Chefstyle here: https://github.com/chef/chefstyle 

(Note: we use this for *internal* Chef coding like InSpec, for external people using chef we point them to `cookstyle` 😄 )

Schwad (approved):
I'm approving this. 

@fzipi - thank you for your work on this. We're awaiting an internal resolution from our internal engineering to resolve the failure on this post-build process step: https://buildkite.com/chef/inspec-inspec-master-habitat-build/builds/903#bf77217c-4eeb-4322-bec8-a8b8868172f3

Just so you know @fzipi - if you aren't already signed up, every Thursday around 9AM U.S. Pacific time there's a community meeting where we give an InSpec update to the Chef community: https://chefcommunity.slack.com/archives/C1P99AXGV

When this is resolved and your contribution is released, we'll be giving you a mention in the meeting to say thanks! 🎉

fzipi:
Cool! I'm not signed up, I think you need a chef.io email address :)

Schwad:
Sorry I should have shared this link! https://community-slack.chef.io/ is where you get your invite (there is an internal chef slack but this one is the community one where we have the community meeting )

Schwad:
@fzipi on this and the other PR we had internal issues with our pipeline unrelated to your code so you're good! Be sure to ping me (@schwad) when you join our community slack there so I can be sure to tag you when we announce this in the next release! 🎉  Thank you!

aaronlippold:
It seems we would also want to add this to Oracle and MSSQL and perhaps the RDS resources as well

fzipi:
Makes total sense.

PR #5126, opened 2020-06-30
Log to stderr when using html2 reporter

### LINKED ISSUE #5125: html2 reporter: InSpec logging mixed in stdout
If you use the `html2` reporter and InSpec produces any log entries; those log entries are on stdout (not stderr) and end up in the HTML doc.

Might be this line:
https://github.com/inspec/inspec/blob/master/lib/inspec/base_cli.rb#L234

I'll fix this while I'm in this bit of the code...

Aha! Link: https://chef.aha.io/features/SH-2454

james-stocks:
Fixes #5125

Aha! Link: https://chef.aha.io/features/SH-2455

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -231,7 +231,7 @@ module Inspec
 
     private
 
-    ALL_OF_OUR_REPORTERS = %w{json json-min json-rspec json-automate junit html yaml documentation progress}.freeze # BUT WHY?!?!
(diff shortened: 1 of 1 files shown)


Schwad (approved):
Nice catch!! 👍 

I wonder if this has any implications when logging with a third party reporter plugin.... Probably not, right? 🤔

james-stocks:
@Schwad I believe you're right. I'll file an issue for that.

PR #5128, opened 2020-06-30
Report faulty profiles instead of crashing

### LINKED ISSUE #5107: Report a profile that fails to load as "failed"
If you have a profile that raise an exception outside a control e.g. 

```
raise 'oh no!'

control "tmp-1.0" do                        # A unique ID for this control
  impact 0.7                                # The criticality, if this control fails.
  title "Create /tmp directory"             # A human-readable title
  desc "An optional description..."
  describe file("/tmp") do                  # The actual test
    it { should be_directory }
  end
end
```

Then run `inspec exec crash_profile working_profile --reporter=json`...
...currently InSpec will just crash on the invalid code of `crash_profile`, the `working_profile` does not run, and no JSON report is generated.

It would be better than we catch exceptions and report the profile as `errored` (similar to how unsupported profiles still generate a report, with the profile reported as `skipped` )

This should work with all reporters (at least, no reporter should default to glossing over the errored profile).
There should be a non-zero exit code (tbc) for this case.

Aha! Link: https://chef.aha.io/features/SH-136

james-stocks:
Fixes #5107 

When InSpec cannot load a profile, it should log the failed profile instead of terminating InSpec entirely.

Aha! Link: https://chef.aha.io/features/SH-2553

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -265,7 +265,7 @@ module Inspec
           metadata.dependencies[i][:status] = "skipped"
           msg = "Skipping profile: '#{d.name}' on unsupported platform: '#{d.backend.platform.name}/#{d.backend.platform.release}'."
           metadata.dependencies[i][:status_message] = msg
-          metadata.dependencies[i][:skip_message] = msg  # Repeat as skip_message for backward compatibility
+          metadata.dependencies[i][:skip_message] = msg # Repeat as skip_message for backward compatibility
           next
         elsif metadata.dependencies[i]
           # Currently wrapper profiles will load all dependencies, and then we
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -264,7 +264,8 @@ module Inspec
           # TODO: NO! this is a violation of encapsulation to an extreme
           metadata.dependencies[i][:status] = "skipped"
           msg = "Skipping profile: '#{d.name}' on unsupported platform: '#{d.backend.platform.name}/#{d.backend.platform.release}'."
-          metadata.dependencies[i][:skip_message] = msg
+          metadata.dependencies[i][:status_message] = msg
+          metadata.dependencies[i][:skip_message] = msg  # Repeat as skip_message for backward compatibility
           next
         elsif metadata.dependencies[i]
           # Currently wrapper profiles will load all dependencies, and then we
@@ -337,8 +338,9 @@ module Inspec
       if !supports_platform?
         res[:status] = "skipped"
         msg = "Skipping profile: '#{name}' on unsupported platform: '#{backend.platform.name}/#{backend.platform.release}'."
-        res[:skip_message] = msg
+        res[:status_message] = msg
       else
+        res[:status_message] = @status_message || ""
         res[:status] = failed? ? "failed" : "loaded"
       end
 
@@ -463,6 +465,10 @@ module Inspec
 
     def controls_count
       params[:controls].values.length
+    end
+
+    def set_status_message(msg)
+      @status_message = msg.to_s
     end
 
     # generates a archive of a folder profile
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -94,7 +94,7 @@ module Inspec
       @input_values = options[:inputs]
       @tests_collected = false
       @libraries_loaded = false
-      @failed = false
+      @state = :loaded
       @check_mode = options[:check_mode] || false
       @parent_profile = options[:parent_profile]
       @legacy_profile_path = options[:profiles_path] || false
@@ -147,7 +147,12 @@ module Inspec
         options[:profile_context] ||
         Inspec::ProfileContext.for_profile(self, @backend)
 
-      @supports_platform = metadata.supports_platform?(@backend)
+      if metadata.supports_platform?(@backend)
+        @supports_platform = true
+      else
+        @supports_platform = false
+        @state = :skipped
+      end
       @supports_runtime = metadata.supports_runtime?
     end
 
@@ -164,7 +169,7 @@ module Inspec
     end
 
     def failed?
-      @failed
+      @state == :failed
     end
 
     #
@@ -214,7 +219,7 @@ module Inspec
           begin
             @runner_context.load_control_file(content, abs_path, nil)
           rescue => e
-            @failed = true
+            @state = :failed
             raise Inspec::Exceptions::ProfileLoadFailed, "Failed to load source for #{path}: #{e}"
           end
         end
@@ -259,7 +264,7 @@ module Inspec
         d = dep.profile
         # this will force a dependent profile load so we are only going to add
         # this metadata if the parent profile is supported.
-        if supports_platform? && !d.supports_platform?
+        if @supports_platform && !d.supports_platform?
           # since ruby 1.9 hashes are ordered so we can just use index values here
           # TODO: NO! this is a violation of encapsulation to an extreme
           metadata.dependencies[i][:status] = "skipped"
@@ -335,13 +340,13 @@ module Inspec
       res[:sha256] = sha256
       res[:parent_profile] = parent_profile unless parent_profile.nil?
 
-      if !supports_platform?
+      if @supports_platform
+        res[:status_message] = @status_message || ""
+        res[:status] = failed? ? "failed" : "loaded"
+      else
         res[:status] = "skipped"
         msg = "Skipping profile: '#{name}' on unsupported platform: '#{backend.platform.name}/#{backend.platform.release}'."
         res[:status_message] = msg
-      else
-        res[:status_message] = @status_message || ""
-        res[:status] = failed? ? "failed" : "loaded"
       end
 
       # convert legacy os-* supports to their platform counterpart
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -94,6 +94,7 @@ module Inspec
       @input_values = options[:inputs]
       @tests_collected = false
       @libraries_loaded = false
+      @failed = false
       @check_mode = options[:check_mode] || false
       @parent_profile = options[:parent_profile]
       @legacy_profile_path = options[:profiles_path] || false
@@ -162,6 +163,10 @@ module Inspec
       @writable
     end
 
+    def failed?
+      @failed
+    end
+
     #
     # Is this profile is supported on the current platform of the
     # backend machine and the current inspec version.
@@ -197,7 +202,7 @@ module Inspec
     end
 
     def collect_tests(include_list = @controls)
-      unless @tests_collected
+      unless @tests_collected || failed?
         return unless supports_platform?
 
         locked_dependencies.each(&:collect_tests)
@@ -206,7 +211,12 @@ module Inspec
           next if content.nil? || content.empty?
 
           abs_path = source_reader.target.abs_path(path)
-          @runner_context.load_control_file(content, abs_path, nil)
+          begin
+            @runner_context.load_control_file(content, abs_path, nil)
+          rescue => e
+            @failed = true
+            raise Inspec::Exceptions::ProfileLoadFailed, "Failed to load source for #{path}: #{e}"
+          end
         end
         @tests_collected = true
       end
@@ -329,7 +339,7 @@ module Inspec
         msg = "Skipping profile: '#{name}' on unsupported platform: '#{backend.platform.name}/#{backend.platform.release}'."
         res[:skip_message] = msg
       else
-        res[:status] = "loaded"
+        res[:status] = failed? ? "failed" : "loaded"
       end
 
       # convert legacy os-* supports to their platform counterpart
--- a/lib/inspec/reporters/json.rb
+++ b/lib/inspec/reporters/json.rb
@@ -69,7 +69,7 @@ module Inspec::Reporters
         }.reject { |_k, v| v.nil? }
 
         # For backwards compatibility
-        res[:skip_message] = res[:status_message] if res[:status] == 'skipped'
+        res[:skip_message] = res[:status_message] if res[:status] == "skipped"
 
         res
       end
--- a/lib/inspec/reporters/json.rb
+++ b/lib/inspec/reporters/json.rb
@@ -45,8 +45,8 @@ module Inspec::Reporters
     end
 
     def profiles
-      run_data[:profiles].map { |p|
-        {
+      run_data[:profiles].map do |p|
+        res = {
           name:            p[:name],
           version:         p[:version],
           sha256:          p[:sha256],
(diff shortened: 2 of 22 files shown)


lib/inspec/runner_rspec.rb
```diff
@@ -90,9 +90,12 @@ def exit_code
       return @rspec_exit_code if @formatter.results.empty?
 
       stats = @formatter.results[:statistics][:controls]
+      load_failures = @formatter.results[:profiles]&.select { |p| p[:status] == "failed" }&.any?
       skipped = @formatter.results.dig(:profiles, 0, :status) == "skipped"
```
james-stocks, line 94: I noticed this line _might_ be wrong, as it requires the first (`0`) profile to be skipped; when the run might have a passing profile ahead of the skipped profile.

lib/inspec/reporters/json.rb
```diff
@@ -31,22 +31,22 @@ def platform
...
-          status:       r[:status],
-          code_desc:    r[:code_desc],
-          run_time:     r[:run_time],
-          start_time:   r[:start_time],
-          resource:     r[:resource],
-          skip_message: r[:skip_message],
-          message:      r[:message],
-          exception:    r[:exception],
-          backtrace:    r[:backtrace],
+          status:         r[:status],
```
james-stocks: I need to back this out, this spacing change is a mistake.

test/functional/inspec_exec_test.rb
```diff
@@ -33,11 +33,12 @@ def stderr
   end
 
   it "cleanly fails if mixing incompatible resource and transports" do
-    # TODO: I do not know how to test this more directly. It should be possible.
+    # TODO: It should be possible to test this more directly.
     inspec "exec -t aws:// #{profile_path}/incompatible_resource_for_transport.rb"
 
-    _(stdout).must_be_empty
-    _(stderr).must_include "Unsupported resource/backend combination: file / aws. Exiting."
+    _(stderr).must_be_empty
```
james-stocks, line 39: This test is being changed because the previous version would now be counted as a "failed to load" profile. And that's not the point of this test. It now asserts that the expected error message appears on stdout (as it does as a test failure with the CLI reporter).

lib/inspec/profile.rb
```diff
@@ -197,7 +202,7 @@ def params
     end
 
     def collect_tests(include_list = @controls)
-      unless @tests_collected
+      unless @tests_collected || @failed
```
clintoncwolfe: Maybe use `failed?` here

lib/inspec/profile.rb
```diff
@@ -94,6 +94,7 @@ def initialize(source_reader, options = {})
       @input_values = options[:inputs]
       @tests_collected = false
       @libraries_loaded = false
+      @failed = false
```
clintoncwolfe: It appears that status (used for skipped) is a property of metadata, while failed status is being represented as an instance variable directly on Profile. Could failed be a value of the metadata property?

lib/inspec/run_data.rb
```diff
@@ -47,7 +47,7 @@ class RunData
     # core reporters have been migrated to plugins. It is probable that new data elements
     # and new Hash compatibility behavior will be added during the core reporter plugin
     # conversion process.
-    SCHEMA_VERSION = "0.1.0".freeze
+    SCHEMA_VERSION = "0.2.0".freeze
```
clintoncwolfe, line 50: The rundata docs need to be updated - https://github.com/inspec/inspec/blob/94c30647ed8a3374f576379553a73ae7e1587a4c/docs/dev/plugins.md#L449

lib/inspec/schema/exec_json.rb
```diff
@@ -100,10 +100,10 @@ module ExecJson
           "description" => Primitives::STRING,
           "inspec_version" => Primitives::STRING,
 
-          # These are generated at runtime, and all except skip_message are guaranteed
+          # These are generated at runtime, and all except status_message are guaranteed
           "sha256" => Primitives::STRING,
           "status" => Primitives::STRING,
-          "skip_message" => Primitives::STRING, # If skipped, why
```
clintoncwolfe, line 106: Shouldn't this be retained for backwards compatibility? My understanding is that on a skip, both skip_message and status_message will be present.

lib/inspec/schema.rb
```diff
@@ -147,7 +147,7 @@ class Schema
         "license" => { "type" => "string", "optional" => true },
         "summary" => { "type" => "string", "optional" => true },
         "status" => { "type" => "string", "optional" => false },
-        "skip_message" => { "type" => "string", "optional" => true },
```
clintoncwolfe, line 150: I suspect this should be retained; see other comment.

clintoncwolfe (commented):
This looks great overall, I especially like the testing. I'm not sure the failed status is being stored in the right place, and there is a missing doc entry, but I think this is pretty close.

lib/inspec/schema/exec_json.rb
james-stocks: I'm taking advantage of the API being pre-1.0 to just drop it. `status_message` should be used. I'm adding `skip_message` in the reporters until 5.0 when they can be cut out.

lib/inspec/profile.rb
```diff
@@ -206,7 +211,12 @@ def collect_tests(include_list = @controls)
           next if content.nil? || content.empty?
 
           abs_path = source_reader.target.abs_path(path)
-          @runner_context.load_control_file(content, abs_path, nil)
+          begin
+            @runner_context.load_control_file(content, abs_path, nil)
```
Schwad, line 220: Thanks for doing this! So I'm guessing the flow here is during the `collect_tests` phase for the profile object, this is where you check and kill things if `load_control_file` fails, to isolate and document that failure against the Profile

lib/inspec/runner.rb
```diff
@@ -115,8 +115,14 @@ def load
           @test_collector.add_profile(requirement.profile)
         end
 
-        tests = profile.collect_tests
-        all_controls += tests unless tests.nil?
+        begin
+          tests = profile.collect_tests
+          all_controls += tests unless tests.nil?
+        rescue Inspec::Exceptions::ProfileLoadFailed => e
+          Inspec::Log.error "Failed to load profile #{profile.name}: #{e}"
```
Schwad, line 122: ....and here's where it seems we catch the `raise` referenced earlier and set our message. Nice! 👍

test/fixtures/profiles/raise_outside_control/controls/raises.rb
```diff
@@ -0,0 +1,11 @@
+# Ref https://github.com/inspec/inspec/issues/5107
+# Users may add code outside controls without exception handling.
+# While this is a user mistake, users may not know about it if we don't report it...
+
+raise StandardError, 'Something unforeseen...'
+
```
Schwad, line 6: Would this feature also work for a `raise` inside of a control block but outside of a `describe` block? I feel like that code also gets evaluated but it also might be a case you're not looking to solve with this issue.

Schwad (commented):
Apologies, my two earlier comments I mashed the "comment" button instead of bundling them into a review so now you've probably got extra notifications. Those two comments were just me making sure I understood the flow for solving this correctly. 

Great stuff and looks good to me, thanks for pulling this together! 👍  I also like how you created a home for `status_message` which can be used in new ways.

test/fixtures/profiles/raise_outside_control/controls/raises.rb
james-stocks: See the issue: https://github.com/inspec/inspec/issues/5107#issuecomment-648726990

There's at least one other case that needs a different fix, it will be filed as a separate issue.

lib/inspec/run_data.rb
james-stocks: Good call, thanks, I pushed an update to this file.

lib/inspec/profile.rb
james-stocks: I think you're right that having `failed` as an instance variable might not be the right place.
However, I'm not seeing an existing `status` in the metadata. The `skipped` status is instead inferred by the profile by calling [`supports_platform?`](https://github.com/inspec/inspec/blob/master/lib/inspec/profile.rb#L327) on the metadata instance.

How would you see this new `failed` property neatly fitting into the metadata? It could be a new `status` metadata property, but I'd need to rework how skipped profiles work as well.

lib/inspec/profile.rb
james-stocks: Good call, done.

lib/inspec/schema/exec_json.rb
clintoncwolfe: This isn't part of the (internal) pre-1.0 rundata API though; this is part of the published schema for the (external) output of the JSON reporter. This will impact people.

lib/inspec/profile.rb
clintoncwolfe: This is an awkward location, but there is a status on the dependency structure, which is set in a really weird way. 
https://github.com/inspec/inspec/pull/5128/files#diff-d1eede4085683a605d9b8892a198f642R265

james-stocks: I did see that, but that is metadata for dependency profiles, not metadata for the immediate profile, no? Or can it be used to represent the metadata for the current profile?
At this point I'd rather just have the instance variable than a metadata entry that behaves weirdly.

lib/inspec/schema/exec_json.rb
james-stocks: Ah, people running JSON profiles generated by older InSpec versions? I'll put `skip_message` back in then

lib/inspec/schema.rb
james-stocks: I trust you're right and I'll put it back, but I'd like to know - what's the use case where this breaks for users?

lib/inspec/profile.rb
clintoncwolfe: I agree the existing approach on Metadata's dependency structure is inappropriate.  

What I'm saying is that the responsibility for knowing the status of the profile is currently on the Metadata object. This PR now divides that responsibility between the Profile and Metadata objects, which creates a lot of potential for bugs.  An implementation could be chosen - an instance variable on Metadata, perhaps - that keeps the responsibility on one class. 

You could still have `Profile#failed?` - it can just call into metadata.

lib/inspec/schema.rb
clintoncwolfe: (This has nothing to do with the internal RunData API)

This is part of the schema components used to construct the schemae for the JSON output of InSpec - both the JSON profile export and the JSON reporter. We don't know how much the schema are used, but since we are retaining the fields in the output in the JSON reporter when the profile is skipped, if we drop the skip_message field from the schema, the schema will definitely be wrong.

People definitely use the JSON profile export and JSON reporter; we don't know if users are validating against the schema but breaking that would be bad.

james-stocks: Thanks very much for explaining - I need to learn more about these bits of InSpec.

I put `skip_message` back in with a comment that `status_message` should now be used for skip reasons.

lib/inspec/schema/exec_json.rb
james-stocks: I put it back but changed the comment.

lib/inspec/profile.rb
james-stocks: I'll have a go at this, but I might need to leave a backlogged issue to fix how skipping works so that it is also a real status and not inferred indirectly.

james-stocks: Looking at this a second time, `Metadata` is intended to represent the static metadata stored in `my_inspec_profile/metadata.rb`

https://github.com/inspec/inspec/blob/master/lib/inspec/metadata.rb#L12

It does not appear to be the right place to represent an instance of the profile loaded by the runner - this file (`lib/inspec/profile.rb`) is the correct place for that.

So I would like to keep the change here. What should change is to have the instance variable be `@state` not `@failed` and have it store whether this instance of the profile is `:loaded` `:failed` or `:skipped` (the latter would mean directly storing here if this instance of the profile is skipped instead of calling into the metadata).

@clintoncwolfe if you don't agree with me, can we at least agree on a refactoring issue to get the responsibility of Metadata fixed up and documented properly outside of this PR?

clintoncwolfe: Your approach sounds reasonable - having an ivar named state on Profile seems fine, and (ideally) moving the skipped status to be indicated using that ivar sounds good. Adding a comment to the effect that Metadata represents metadata and Profile represents state might be helpful, as they have some overlap. Up to you if you want that on a separate PR.

james-stocks:
@clintoncwolfe I pushed one more commit that adds the `@state` ivar and adds more doc lines to metadata to make its remit clearer

lib/inspec/profile.rb
Schwad: Just had a look at this and appreciate the thought gone into this issue on the backscroll. Looking at this from a newer set of eyes I do appreciate the clarifying comment on metadata (on an initial reading of the object class my interpretation would have been that it solely holds responsibility for `metadata.rb`). 

I also found it useful to clarify that `Profile` is the runtime state representation, and the use of `@state` seems fitting to me. (Naming this is hard, no?) Thanks for pulling this together.

lib/inspec/schema/exec_json.rb
Schwad: If this was re-added, would we need it to be added in `plugins.md` or left out due to deprecation?

lib/inspec/schema.rb
Schwad: Status_message is a great approach but definitely appreciate Clinton's comments on how this affects downstream users. Would this be a candidate for an InSpec 5 deprecation? (either one to add then or add now for future removal?) 

But before going that far, is there definitely no edge case in which we'd want both a skip_message and a status_message? 🤔  my initial reaction is no; because if something is skipped in theory there ought not to be anymore to say on the subject.

Schwad (approved):
Thank you for doing this! 🚀  LGTM 👍

lib/inspec/schema.rb
james-stocks: I don't think there's an opportunity here to add a deprecation warning since this is just a schema. We can just add it to our list of things to take care of when cutting InSpec 5.0.

I don't foresee a case where 2 statuses are needed. At some point we potentially might need to add code to decide which of 2 statuses trumps the other. But currently once it reaches `failed` or `skipped` state, that should be the profile's final outcome.

lib/inspec/schema/exec_json.rb
james-stocks: `plugins.md` documents the run data API and this isn't that.

clintoncwolfe (approved):
Thanks!

PR #5129, opened 2020-07-01
Empty commit to rebuild InSpec for release

Schwad:
Sometimes we need to do another PR to rebuild InSpec when something happens downstream or we need to capture changes elsewhere. This is to capture the latest version of `train`.

Aha! Link: https://chef.aha.io/features/SH-2473


Schwad:
Killing as we've changed directions on rebuild and pushing train next week 👍

PR #5130, opened 2020-07-01
Add the generic 'darwin' to the service resource

btm:
macOS 11 Big Sur will be released later this year. Current beta versions
return 10.16 as the version, but the product name has changed from 'Mac
OS X' to 'macOS'. Train probably needs to be modified to deprecate
'mac_os_x' as a platform in favor of 'macos' but that would be a
significant downstream change. Train does fall back to 'darwin' on macOS
10.16, so by adding darwin to the list of platform names for the service
resource we are able to work around this for the moment.

This is the only location where mac_os_x is currently being used in
InSpec. Because we're in a case statement on platform rather than the
more generic platform family, we can't simply remove mac_os_x in favor
of darwin.

Aha! Link: https://chef.aha.io/features/SH-2490


james-stocks (approved):
LGTM, thanks @btm 

Verify is failing due to the Windows test waiver expiring - we're going to push that back out again today.

james-stocks:
Opened https://github.com/inspec/train/issues/617 to track fixing this in train.

Schwad (approved):
Hi @btm ! 👋  thanks for this, looks good

> Verify is failing due to the Windows test waiver expiring - we're going to push that back out again today.

Just merged #5063 which handles the expiration and some of the `skip_windows` invocations, so all should be green now 👍

PR #5135, opened 2020-07-07
Allow ‘=‘ character in CLI inputs

### LINKED ISSUE #5131: CLI inputs do not handle = (equals) character
Present from 4.18 onwards. 

To reproduce:

```
control 'Check an inputs value' do
  describe input(:my_input) do
    it { should eq 'ab=cde' }
  end
end
```

```
jstocks@phobos profiles % inspec exec my_profile --input my_input='ab=cde'

Profile: InSpec Profile (my_profile)
Version: 0.1.0
Target:  local://

  ×  Check an inputs value: ab
     ×  ab is expected to eq "ab=cde"
     
     expected: "ab=cde"
          got: "ab"
     
     (compared using ==)



```

It seems to terminate the string on the `=` character.

In 4.20 onwards since we have parsing of inputs to hashes (I thought this might offer a quick workaround), this also does not work - InSpec still treats the YAML/JSON expression as a `=`-terminated string.

Aha! Link: https://chef.aha.io/features/SH-2510

Schwad:
Fixes #5131

Due to the use of `#split(‘=‘)` against inputs supplied via the CLI we had an edge case where inputs with `’=‘` in the value would cause a breakage.

This PR supplies a test for the regression and fixes the bug with a regex solution.

Aha! Link: https://chef.aha.io/features/SH-2552

### DIFF
--- a/lib/inspec/input_registry.rb
+++ b/lib/inspec/input_registry.rb
@@ -165,7 +165,8 @@ module Inspec
             raise ArgumentError, "ERROR: An '=' is required when using --input. Usage: --input input_name1=input_value1 input2=value2"
           end
         end
-        input_name, input_value = pair.split("=")
+        pair = pair.match(/(.*?)=(.*)/)
+        input_name, input_value = pair[1], pair[2]
         input_value = parse_cli_input_value(input_name, input_value)
         evt = Inspec::Input::Event.new(
           value: input_value,
--- a/test/fixtures/profiles/inputs/with_various_values/controls/input_equals.rb
+++ b/test/fixtures/profiles/inputs/with_various_values/controls/input_equals.rb
@@ -1,0 +1,7 @@
+title 'With various values'
+
+control 'with equals sign' do
+  describe input(:my_input) do
+    it { should eq 'ab=cde' }
+  end
+end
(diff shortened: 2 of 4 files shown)


clintoncwolfe (approved):
Thanks! I like where you put the test fixture profile.

Schwad:
> Thanks! I like where you put the test fixture profile.

@clintoncwolfe Cheers! 😄  I figured that might be a useful pattern for future PRs like this. (Plus, when you're scanning fixtures in future hopefully it makes things clear what the genesis of that profile was)

test/fixtures/profiles/regressions/5131_input_equals/controls/input_equals.rb
```diff
@@ -0,0 +1,7 @@
+title 'Regressions for 5131'
```
james-stocks: It's only a regression if it previously worked and we broke it. Did this case previously work OK in the past?

james-stocks: When I said `Present from 4.18 onwards.` I meant it's known to be a problem that far back. I didn't mean that it worked previous to that version, sorry.

Schwad: Good point! I've reverted and tried to put this fixture in a more aptly named location. If the pipeline passes I'll give a rebase squash so we're clean again 👍

james-stocks: Thanks. It would be fantastic to have a `regressions` directory for cases that are regressions. Sorry if my issue description was misleading

james-stocks (approved):
Thank you @Schwad :)

Issue #5133, opened 2020-07-06
FilterTable resources are fully loaded before waivers/skips/excludes apply


james-stocks:
*Expected behaviour*: Waiving or skipping a control or excluding it in an inheriting profile will skip all execution of the control. This allows slow or broken controls to be removed from execution.

While there are a number of existing issues around mixing Ruby code with InSpec DSL, this is one case where it doesn't work within the bounds of our DSL.

*To reproduce*:

Create a filter table resource e.g. `libraries/Animals.rb` that outputs a message when it loads all test data (in a real world example, this would be reading in _all_ data from a host or API)

```
class Animals < Inspec.resource(1)
  name 'animals'

  # FilterTable setup
  filter_table_config = FilterTable.create
  filter_table_config.register_column(:names, field: :name)
  filter_table_config.register_column(:species, field: :species)
  filter_table_config.register_column(:weights, field: :weight)
  filter_table_config.install_filter_methods_on_resource(self, :data)

private
  @data = nil

  def data
    puts 'Filter table is doing all its work'
    [{name: 'Alice', species: 'Aardvark', weight: 20},
     {name: 'Bob', species: 'Badger', weight: 40},
     {name: 'Colin', species: 'Aardvark', weight: 10}]
  end
end
```

And some control code like:

```
control 'Check my animals' do
  impact 1.0
  title 'Check my animals'
  describe animals.where(name: 'Alice') do
    it { should exist }
    its('species') { should eq ['Aardvark'] }
  end

  describe animals.where(name: 'Barry') do
    it { should_not exist }
  end
end
```

When this runs with a waiver, the output indicates that the filter table gets fully loaded despite the control being skipped:

```
jstocks@phobos inspec-filtertable % inspec exec . --waiver-file waivers.yaml
Filter table is doing all its work
Filter table is doing all its work

Profile: InSpec Profile (inspec-filtertable)
Version: 0.1.0
Target:  local://

  ↺  Check my animals: Check my animals
     ↺  Skipped control due to waiver condition: I don't care


Profile Summary: 0 successful controls, 0 control failures, 1 control skipped
Test Summary: 0 successful, 0 failures, 1 skipped
jstocks@phobos inspec-filtertable % 
```

Aha! Link: https://chef.aha.io/features/SH-2515

Issue #5068, opened 2020-06-03
Fix skipping of controls to actually do what it says


kekaichinose:
## What problem are we solving?

_Should be a few sentences. Describe the user stories and the target market._

As a **Chef customer**, I want to be able to skip controls that are not applicable for my needs

Today, though InSpec claims to allow users to "skip" controls, on the back end, its still executing the control, but just reporting the result as "skipped". This:
* misleads the user as they were told a control was skipped, but was in fact run
* wastes cycles on evaluating/running a control that the user does not care about

We want to restore the customer's trust in InSpec's skip (and even waiver) functionalities by ensuring that a control that is skipped is actually skipped.

## **Objective and Key Results**

_What's the ideal outcome? How will we measure that we're successful?_

**Objective** | **Key Results**
-- | --
Ensure that controls that a user wants to skip are actually skipped (i.e. not run) | • &nbsp; Customer can trust that a control has been skipped during an audit run <br>• &nbsp; (secondary) customer will notice performance improvements as non-running controls will save execution cycles

## **Acceptance Criteria**

_How will we know when we're done?_

* Skipped controls are not actually evaluated/run
* Waived controls that are `run: false` are actually not evaluated/run
* Profiles that had controls that were previously skipped but still running are now properly not evaluated/run

## **Dependencies**
_What other work needs to come first? What will block us from success?_

Spike of RSpec test aggregation - understanding what it will take to pull aggregation out of RSpec

## **Out of Scope**
_What problems are not included in this epic?_

* Deprecating skip_control for waivers
  * The `skip_control` feature was good to address an initial need from customers to skip non-applicable controls in a given profile. However, as the compliance journey evolved, we also recognized there needed to be auditing around why a control was skipped, by who, whether it was temporary or permanent, etc.
  * Waivers launched to address this need back in Q1 2020
  * As a result, we should deprecate the `skip_control` feature in favor of waivers

## **Open Questions**

* What questions still need to be addressed?
* Why was this missed during initial implementation?
* How will fixing control skipping impact existing customer use cases? Will customers that are using InSpec's skip control functionality as-is today be broken once we address this issue?

## **GitHub Epic**
_Aha! epics should be tracked in GitHub and attached to tactical issues and tasks. What is the link to that epic?_

https://github.com/inspec/inspec/issues/5068

## **Other Links**
_List links to other documents or artifacts that are mentioned/inform this epic_

* Fixes #4886 
* Fixes #5050

[...]

Issue #5005, opened 2020-05-01
Resolve Waiver Extraction


Schwad:
## Detailed Description

**RFC: Expanding/refining this request description**

On the fly, it does not appear that there is a straightforward way to identify and manipulate waivers while prying into the codebase. This was captured in issue #5001 where we needed to identify if an input was a waiver, in order to pass it along through the script correctly. 

There was a regression fix put in the PR (simply identifying elements of an input that are `waiveresque`, but more as a bugfix which started this larger discussion as pointed out by @stevendanna : https://github.com/inspec/inspec/pull/5001#discussion_r417934703

## Context

This would allow contributors to be more agile in handling and assisting with future waivers-related development.

## Possible Implementation

I think we should consider implementing here the options @stevendanna  outlined:

```
1. Change waivers to be a specific class we can look for and then look for it here
2. Change the waivers hash to have some sentinel _is_waiver key that we can look for if we have a hash here.
```

Schwad:
I'm hopeful that this can help expand the discussion started in #5001 

I would also appreciate some support on labeling as I'm unsure which would be most appropriate for this issue

james-stocks:
Thanks @Schwad 
I think this is technical debt that will bite us in future if we want to extend either inputs or waivers

clintoncwolfe:
The field `justification` may already be an implicit sentinel value - the docs at https://www.inspec.io/docs/reference/waivers/ seem to imply that it is a required field. If so, simply verifying that an input is a control-named Hash with a key named `justification` would be one option.

Justification is unconditionally used here: https://github.com/inspec/inspec/blob/8b5789004ac31f4b6e7210421b119af41f483e3e/lib/inspec/rule.rb#L372 though there is no safety code to error-check whether it is present in the input.

clintoncwolfe:
This is not a technical priority, I'm going to close this.

Issue #4959, opened 2020-03-19
controls are still executed with waivers and if they time out they are recorded as failures


mattray:
InSpec 4.18.85, audit cookbook 9.1.0, Chef 15.8.23

I have a waiver for a command that fails because it times out, but it still shows up as failed. Other machines with the same waiver don't have the command failure, so they report as passing and the control reports as "Skipped control due to waiver condition".

Related to #4886 for sure.

Failure
```
Command timed out after 600s:
Command exceeded allowed execution time, process terminated
...
```

clintoncwolfe:
Refs #5327 , #5273

clintoncwolfe:
Command timeouts were implemented on #5273, so that may help. You can now specify a timeout on a command on certain transports by specifying 
```
describe command("some invocation", timeout: 30) do
 # Will timeout after 30 seconds
end
```

PR #5142, opened 2020-07-15
Ensure only_if is evaluated when control is run

Schwad:
This is a small change that affects a few issues around expected
behavior and actual behavior. This is best illustrated with a few
examples.

Say within a control I have:

```ruby
control "01_test_control" do
  #... Some other code and describe blocks
  only_if { some_requirement_needed_to_proceed_like_an_unreliable_connection }
  describe(true) { it { should eq true } }
end
```

As I drafted this `only_if` I expect it to run.

Then, I am given a requirement to waive this control. I draft the
following waiver file:

```yaml
01_test_control:
  justification: At the moment we are not running this waiver.
  run: false
```

The control stops running with the following output and all is good:

```
 ↺  01_test_control No-op
     ↺  Skipped control due to waiver condition: At the moment we are
     not running this waiver.
```

Things change. Several months later I am asked to run that control. But
for documentation purposes we do keep it in waivers.yml as such:

```yaml
01_test_control:
  justification: At the moment we are not running this waiver.
  run: true
```

My __expectation__ is that the entire control block would evaluate at
this point, including my business logic within the only_if . This works
because if it does not run I’ll receive the different output of:

```
 ↺  01_test_control No-op
     ↺  Skipped control due to only_if condition.
```

However, that is not what happens in practice. `run: true` in waivers
will stop application of any only_ifs. I do not think this is expected
behavior. Feel free to use above code for reproduction cases.

**Why does this matter**

This also affects:

- https://github.com/inspec/inspec/issues/5068
- Customer issues around `skip_controls and waiver loading`
- https://github.com/inspec/inspec/issues/4959
- https://github.com/inspec/inspec/issues/5005
- https://github.com/inspec/inspec/issues/3760
- https://github.com/inspec/inspec/issues/4886
- https://github.com/inspec/inspec/issues/5133

This PR also prevents failures within `control` blocks from breaking when 
waived. It ensures that they remain skipped, which does not 
currently happen. ( NOTE: this applies only to failures, you can see 
that `puts 'hello'` or `sleep 20` will still evaluate and need a separate solution )

To reproduce:

```ruby
control "01_test_control" do
  fail
  describe(true) { it { should eq true } }
end
```

with:

```yaml
01_test_control:
  justification: At the moment we are not running this waiver.
  run: false
```

Running against master you get:

```
Profile: tests from testy.rb (tests from testy.rb)
Version: (not specified)
Target:  local://

  ×  01_test_control: testy.rb:1
     ×  Control Source Code Error testy.rb:1
```

Against this branch you get the expected behavior of:

```
Profile: tests from testy.rb (tests from testy.rb)
Version: (not specified)
Target:  local://

  ↺  01_test_control: No-op
     ↺  Skipped control due to waiver condition: At the moment we are
     not running this waiver.
```

Aha! Link: https://chef.aha.io/features/SH-2616

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -26,7 +26,6 @@ module Inspec
 
       @resource_dsl = resource_dsl
       extend resource_dsl # TODO: remove! do it via method_missing
-
       # not changeable by the user:
       @__code = nil
       @__block = block
@@ -43,11 +42,11 @@ module Inspec
       return unless block_given?
 
       begin
-        instance_eval(&block)
 
-        # By applying waivers *after* the instance eval, we assure that
-        # waivers have higher precedence than only_if.
+        # This ensures we do not load and run control code that is waived.
         __apply_waivers
+
+        instance_eval(&block)
 
       rescue SystemStackError, StandardError => e
         # We've encountered an exception while trying to eval the code inside the


james-stocks:
>  `run: true` in waivers will stop application of any `only_ifs`. I do not think this is expected
behavior. 

Just to be clear, what does happen in this case? Does the control report the result of `describe(true) { it { should eq true } }` ?

Schwad:
> Just to be clear, what does happen in this case? Does the control report the result of `describe(true) { it { should eq true } }` ?

Thanks for asking this question. Upon looking again a few more cases, I've found some different outcomes that I think makes this PR untenable. 

It appears that `only_if` *is* evaluated, but depending on how it is composed, it may actually violate the original hierarchy problem that had `__apply_waivers` later. 

So initially it appears that the only benefit here is that this PR would block failures in waived control blocks from percolating to the surface. (And just report the waiver)

Where this PR falls flat on its face is something I just discovered in manual testing. 

1. When running an `only_if { false }` this PR behaves exactly as expected.
2. When running an `only_if("some_text") { false }` this PR does suffer the original problem of `only_if`'s overriding `waivers`, which we definitely don't want. (i.e. the only_if message would be what's output even if control was waived).

Prying deeper into this would definitely be a journey into __there be dragons__ territory and I'm unsure if it would provide tremendous value. 

I'm going to close this out but hopefully it'll serve as helpful on future waiver work/considerations.

james-stocks:
Thanks for digging into it @Schwad . It does feel like this place in the code (or somewhere directly below it in the call stack) is where we need to untangle skips and waivers.
I suspect waivers will need to be processed twice - running first to exclude any controls that are waived and not meant to run; then also running last and overwriting any controls that did run but should be marked waived.

Schwad:
> I suspect waivers will need to be processed twice - running first to exclude any controls that are waived and not meant to run; then also running last and overwriting any controls that did run but should be marked waived.

That's an interesting point @james-stocks - kicking around my head I'm wondering if we see something like:

```ruby
# To help with naming so it doesn't look like we're doing the same method twice for no reason
alias_method :__scrub_waived_controls, :__apply_waivers

__scrub_waived_controls

instance_eval(our_code)

__apply_waivers
```

Which could *possibly* give us an immediate solution to ensure breakages on the control level are still reported as waivers. 

If I remember right on my manual tests it __still__ seemed to evaluate non-breaking code though (as tested with `sleep` and `puts`)... it just seems to protect us from failures. 

The long term solution may end up being a method that takes inspiration from the `__apply_waivers` method and runs before the `instance_eval` method... but is slightly different to focus on halting eval. 

Cheers for the thoughts, even though this isn't the solution I think we're forming some healthy learnings while digging around.

PR #5148, opened 2020-07-17
Habitat should not expand $PATH at buildtime

james-stocks:
Habitat should have everything it needs at runtime as a habitat dependency and should prefer habitat dependency bins over native bins.

Aha! Link: https://chef.aha.io/features/SH-2628

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -70,7 +70,7 @@ wrap_inspec_bin() {
 set -e
 
 # Set binary path that allows InSpec to use non-Hab pkg binaries
-export PATH="/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:$PATH"
+export PATH="/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:\$PATH"
 
 # Set Ruby paths defined from 'do_setup_environment()'
 export GEM_HOME="$GEM_HOME"
--- a/.expeditor/buildkite/artifact.habitat.test.sh
+++ b/.expeditor/buildkite/artifact.habitat.test.sh
@@ -48,6 +48,9 @@ fi
 echo "+++ Installing ${pkg_ident:?is undefined}"
 hab pkg install -b "${project_root:?is undefined}/results/${pkg_artifact:?is undefined}"
 
+echo "--- Removing world readability from /usr/local/bundle"
+chmod go-w /usr/local/bundle
+
 echo "+++ Testing $PLAN"
 
 PATH="$(hab pkg path ci/inspec)/bin:$PATH"


james-stocks:
@TheLunaticScripter I can't add you as reviewer, but could you review this please?

habitat/plan.sh
```diff
@@ -69,9 +69,6 @@ wrap_inspec_bin() {
 #!$(pkg_path_for core/bash)/bin/bash
 set -e
 
-# Set binary path that allows InSpec to use non-Hab pkg binaries
-export PATH="/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:$PATH"
```
stevendanna, line 73: Another point of interest is that this `$PATH` will actually be evaluated at build-time and will include the build desp rather than just the runtime deps.

stevendanna (commented):
I wonder if instead of completely removing this, we want to setup a PATH that would be similar to what `hab pkg exec` builds such that people who are using this package via binstubs get consistent behavior.  Namely, as of hab 1.5.96+, `hab pkg exec` will construct a PATH consisting of the paths for the package's runtime dep tree plus the incoming using environment's PATH.

The binstubs, however, are simple links into the package and do not get a hab controlled path.

We could potentially create a more consistent experience in our wrapper by doing something like (warning, code written in comment box):

```shell
USER_PATH=\$PATH
. ${pkg_prefix}/RUNTIME_ENVIRONMENT
export PATH=\$PATH:\$USER_PATH
```

One note that we may want to put in the release notes or something is that this may break people who are using `hab pkg exec` to execute hab on versions older than 1.5.96 since then they will not have access to system commands and since`inspec` has a number of resources that require a dependencies not currently in our `pkg_deps`.  (I believe this was done to avoid `inspec` depending a very large dependency tree of hab packages. For example the postgres_session resource depends on `psql`: https://github.com/inspec/inspec/blob/master/lib/inspec/resources/postgres_session.rb#L67)

james-stocks:
I was suspicious that this will affect InSpec resources that assume programs are on path, such as `http`, `port` and `host`.

Testing this change, `http` works fine. `port` and `host` appear to not work at all, before or after this change. `port` seems to claim that no ports are present on the system, and `host` informs that `netcat` is not available...

```
inspec> port.where { true }.raw_data
=> []
inspec> describe host('example.com', port: 80, protocol: 'tcp') do
inspec>   it { should be_reachable }  
inspec>   it { should be_resolvable }  
inspec>   its('ipaddress') { should include '12.34.56.78' }  
inspec> end  

Profile: inspec-shell
Version: (not specified)

  Host example.com port 80 proto tcp
     ↺  The following requirements are not met for this resource: netcat must be installed

inspec> 
```

habitat/plan.sh
james-stocks: I would have assumed this only affected build-time but I tested building an artifact with and without this line and it does change the `$PATH` when I'm running `hab pkg exec`

stevendanna: Yes, sorry, what I mean is that `PATH` here is interpreted at build-time not run time. which is why you are getting all of the Hab binaries in your path.

stevendanna: As in, this script this is rendering is called at runtime, but the script that is rendered doesn't contain `$PATH` it contains the build-time expansion of that environment variable.

TheLunaticScripter:
Yep, this is just a build time constraint and this is a good move. To address some of the concerns with `hab pkg exec` and the potential need for binaries at run time the way this has been addressed in the chef-client plan has been to expose the ability to set the ruby `ENV["PATH"]` variable in the client config which we expose to the user in the effortless scaffolding. If that was an option in InSpec as well that would be nice but probably should be scoped under a different PR IMO.

james-stocks:
We should consider merging this PR, but it does not help the related customer issue. Their issue is with Habitat adding all the runtime deps `bin` directories to PATH. When you try to check the CIS recommendation `xccdf_org.cisecurity.benchmarks_rule_13.6_Ensure_root_PATH_Integrity` the habitat bin directories get checked by this control and they are non-compliant.

habitat/plan.sh
james-stocks: _puts on glasses_

Oh this is rendering the bin script for InSpec. And `$PATH` gets expanded at build time. So this was extra buggy.

stevendanna:
> We should consider merging this PR, but it does not help the related customer issue. Their issue is with Habitat adding all the runtime deps bin directories to PATH. When you try to check the CIS recommendation xccdf_org.cisecurity.benchmarks_rule_13.6_Ensure_root_PATH_Integrity the habitat bin directories get checked by this control and they are non-compliant.

Well, it _may_. If they are running via a binlink then no Habitat paths will be added to their path.  If they are running via `hab pkg exec` then the runtime dep's of the package will be added to PATH which at least have a chance of existing, unlike some of the build-deps.

habitat/plan.sh
james-stocks: @stevendanna reviewing this with @TheLunaticScripter , we currently need to put native bin dirs first. But we need to fix the $PATH containing build deps. I changed the commit to simply escape the $PATH variable at build time so it expands at runtime.

stevendanna (approved):

james-stocks:
Testing this change, the only package that gets excluded is gcc. It also avoids a number of duplicate paths in PATH. 
I think the larger part of this fix is that the paths will not be hardcoded at the package idents from build time; they will be the package idents for the current Habitat runtime.

stevendanna:
@james-stocks I assume in your testing you are executing this with `hab pkg exec`?  I would have assume that if you use a binlink you should see a much smaller set of entries in your PATH.

james-stocks:
Good call @stevendanna - if I binlink the pkg install then execute `/bin/inspec shell`, that additionally removes from the path:

core/curl
core/db
core/expat
core/gdbm
core/gettext
core/git
core/libtool
core/openssh
core/perl
core/ruby/2.5.7

(some of those look suspicious, I'm going to file some follow-on issues, as [previously commented](https://github.com/inspec/inspec/pull/5148#issuecomment-660034495) it seems some InSpec core resources are not working via Habitat)

james-stocks:
Admin-merging to get around `expeditor/config-validation` not completing.

PR #5149, opened 2020-07-17
WIP: Implement optional InSpecStyle in `inspec check`

Schwad:
With the `--inspecstyle` flag after `inspec check`, this will execute
a full run of the `inspecstyle` gem against the target file or
directory. As the aim is to have this default in InSpec 5.

It also includes additional information including how to get involved in
the discussion around future InSpecStyle rules; and that there are other
InSpecStyle tools available.

Aha! Link: https://chef.aha.io/features/SH-2629

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -382,8 +382,9 @@ module Inspec
       # Capture InSpecStyle run
       runtime_config = Inspec::Config.cached.respond_to?(:final_options) ? Inspec::Config.cached.final_options : {}
 
+      autocorrect = runtime_config[:inspecstyle_autocorrect] ? "-a" : nil
 
-      if runtime_config[:inspecstyle]
+      if runtime_config[:inspecstyle] || runtime_config[:inspecstyle_autocorrect]
         inspecstyle_target = if File.directory?(@target)
                            File.join(@target, "**/controls/*.rb")
                          else
@@ -393,7 +394,14 @@ module Inspec
         output = capture_stdout do
           require 'rubocop'
           ::RuboCop::CLI.new.run(
-            [inspecstyle_target, "-r", "inspecstyle", "--only", "InSpecStyle"]
+            [
+              inspecstyle_target,
+              "-r",
+              "inspecstyle",
+              "--only",
+              "InSpecStyle",
+              autocorrect
+            ]
           )
         end
         result[:inspecstyle] = output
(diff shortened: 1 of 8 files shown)


Schwad:
This needs a few more things, but we're getting there. This is the core of what I want at the moment.

TODO: 

- [ ] Tests for output and flag
- [ ] Rebase
- [ ] Confirm best location for gem
- [ ] Confirm "style" and copy of output. (This will be part of the team review and maybe docs?)

lib/inspec/ui.rb
```diff
@@ -75,6 +75,14 @@ def bold(str, opts = { print: true })
       end
     end
 
+    # Allows bold and color
+    ANSI_CODES[:color].keys.each do |color|
```
james-stocks, line 79: Nice!

lib/inspec/profile.rb
```diff
@@ -362,6 +362,22 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
...
+
+      if runtime_config[:inspecstyle]
+        inspecstyle_target = if File.directory?(@target)
+                           File.join(@target, "**/*.rb")
+                         else
+                           @target
+                         end
+
+        cmd = Mixlib::ShellOut.new(
+          "bundle exec rubocop #{inspecstyle_target} -r inspecstyle --only InSpecStyle"
```
james-stocks: Can this be changed from a ShellOut to instantiating an instance of InSpecStyle or rubocop within Ruby?
Not shelling out is preferable (though it's more work for us :( )
I believe the use of `bundle exec` means this won't work for people on package installs or running in Habitat.

Schwad: Oooh yeah that would be better. I'll see if I can get our report generated in PureRuby™️

Schwad:
Not 100% sure what to do with PRs that are valid/draft/experimental whilst unlikely to get picked up this quarter. I feel like this information should stay github-side, but it may give the false representation about how many open/running PRs there are. 

Possibly internal tooling/boards should remain the source of truth on this and I shouldn't be so paranoid about the PR number

kmf:
@Schwad please keep it on GitHub. 
Perhaps add a Label to show the intentions. As a community member I don't have access to internal boards.

Schwad:
Sounds good to me, @kmf 👍  :)

Schwad:
NOTE: This is not receiving development right now but the intent is to do so in the future.

clintoncwolfe:
This has been superceded by #5618.

PR #5151, opened 2020-07-17
changed inspec.yml example to use platform-family instead of os-family, since that's the recommended property name

bcwood:
Aha! Link: https://chef.aha.io/features/SH-2631


Schwad (approved):
Sorry for the delay on this @bcwood - looks good to me - thanks!

clintoncwolfe (approved):

PR #5164, opened 2020-07-24
Supports Scientific Linux in service resource

Schwad:
Currently when using Scientific Linux the user will be prompted with the
unsupported OS error message. As it is based on RHEL, it should be able
to be supported inn the same way.

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-2701

### DIFF
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -141,7 +141,7 @@ module Inspec::Resources
         elsif version > 0
           SysV.new(inspec, service_ctl || "/usr/sbin/service")
         end
-      when "redhat", "fedora", "centos", "oracle", "cloudlinux"
+      when "redhat", "fedora", "centos", "oracle", "cloudlinux", "scientific"
         version = os[:release].to_i
 
         systemd = ((platform != "fedora" && version >= 7) ||


Schwad:
Train identifies Scientific Linux as `scientific` so this should achieve what we're looking for here. We're running into issues with config-validator and the general build which are being addressed separately.

lib/inspec/resources/service.rb
```diff
@@ -141,7 +141,7 @@ def select_service_mgmt # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCom
         elsif version > 0
           SysV.new(inspec, service_ctl || "/usr/sbin/service")
         end
-      when "redhat", "fedora", "centos", "oracle", "cloudlinux"
+      when "redhat", "fedora", "centos", "oracle", "cloudlinux", "scientific"
```
james-stocks, line 144: I'm not at a laptop to check, but I think the `os` hash will contain the "families" the OS is part of, and a more generic solution might be to make this switch block work by family and not specific OSs.

To avoid complicating things, it'd also be fine to just add `scientific` for now

Schwad:
That's a good point! I'll check this going the family route so we do not
have to manually add individual OS's

Schwad:
I spent some time on this this afternoon, first to confirm it works with Scientific Linux and confirm the possibility to switch this over to family. 

It does work for me, and we do have access to family. I also agree that this should be refactored to family and will file an issue to ensure that gets addressed. 

@gsreynolds does this look good to you?

lib/inspec/resources/service.rb
tas50: Does inspec have a concept for platform_family like chef? We bring in scientific and oracle for free in most of our logic by just saying 'rhel' which includes redhat and all the direct clones.

Schwad: I believe we support both the `os[:family]` and `platform_family` options but will need to double confirm next time I'm pried in

PR #5167, opened 2020-07-24
Fixes expeditor-validate violations

Schwad:

Aha! Link: https://chef.aha.io/features/SH-2707


Schwad:
I have been advised that to see fixes in expeditor applied they must first be merged into master. In some cases. In particular here expeditor-validate must be fixed first, and that fix should percolate to the larger fix in general. This will result in a followup build to check that this is working.

PR #5169, opened 2020-07-28
Update docs for known expeditor issues and resolutions

### LINKED ISSUE #5155: Update RELEASE_PROCESS.md to cover expeditor retry cases
# Describe the problem

Currently, we have a 10-step workload on expeditor when running `/expeditor promote inspec/inspec:master x.x.x`. 

If any of those steps fails, we are given a "retry" button. However, the workload is not inherently "retry-supported". For example, if any step after `built_in:publish_rubygems` fails, then a retry will certainly fail the rubygems step saying that the gem has already been released and break the rest of the workload. Additionally, the steps such as `bash:./expeditor/publish-release-notes.sh` are unlikely to have a case to run twice. 

The current advice is if the workload fails, once we are able to retry (say an external outage is resolved) to make a dummy PR commenting out the succeeding steps and merge to master, then click the retry button. In theory expeditor should see this and only start at the uncommented out steps. In practice the retry still would execute each step, dying on an earlier step like `built_in:publish_rubygems`. It is possible that the advice should be 1. merge dummy pr 2. wait for dummy pr master builds to complete-- but this is unconfirmed due to the difficult nature of testing this in normal uptime.

Without a retry process, the only solution would be __manually__ executing these steps either from support or locally, which is not what it was built for and unsustainable. 

Handling this case with a retry is also undocumented in `RELEASE_PROCESS.md` and could be a valuable addition.

## Possible Solution

Obtain guidance on a working approach for promotion retries. If in any way possible, test it (this will take some thinking) manually at least once. Include guidance in `RELEASE_PROCESS.md`

Aha! Link: https://chef.aha.io/features/SH-2645

Schwad:
In the past 1-2 months we have run into several issues with releases
while working with expeditor. The goal of this PR is to document
and define internal processes around these expeditor issues.

Fixes #5155 

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -106,6 +106,16 @@ As you merge each pull request, update the [Pending Release Notes](https://githu
  * Add the PR number as a reference for the Docs team. They will typically remove them, but they may need to get more context by looking up the originating PR.
  * It's preferable to add notes as things merge, lest they be forgotten; sifting through a pile of merged PRs in a rush is a chore!
 
+### Known Expeditor issues and resolutions
+
+Generally, expeditor issues and failures that appear unrelated to the InSpec PR should be flagged internally to `#releng-support` or `#expeditor-support`.
+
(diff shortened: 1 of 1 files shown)


Schwad:
I am unsure whether we ping docs for certain internal-facing documentation such as `RELEASE_PROCESS`. Happy to ping if that's required. 

For background, we've faced a few tricky points with releases and builds in the last few months and this PR is to help make note of how we handled those situations and what we learned for future reference.

PR #5171, opened 2020-07-28
fix error in passwd example

chrisbirster:
## Description
Fixes a spelling error in the passwd example. If you try to run the example provided you will get an error message:

NoMethodError: undefined method `uid' for /etc/passwd:#<Class:0x00000000058e7a60>
from (pry):61:in `load_with_context'

## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #5175, opened 2020-07-30
Ensure Darwin is always detected in the install scripts

tas50:
This is more reliable if the output changes

### DIFF
--- a/omnibus/package-scripts/inspec/postinst
+++ b/omnibus/package-scripts/inspec/postinst
@@ -17,7 +17,7 @@ error_exit()
 
 is_darwin()
 {
-  uname -v | grep "^Darwin" 2>&1 >/dev/null
+  uname -a | grep "^Darwin" 2>&1 >/dev/null
 }
 
 if is_darwin; then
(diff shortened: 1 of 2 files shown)


omnibus/package-scripts/inspec/postinst
```diff
@@ -17,7 +17,7 @@ error_exit()
 
 is_darwin()
 {
-  uname -v | grep "^Darwin" 2>&1 >/dev/null
+  uname -a | grep "^Darwin" 2>&1 >/dev/null
```
Schwad, line 20: Thanks for this! Would there be a case in which `uname -s` wouldn't work? (It seems that `-a` ensures that the `-s` value is supplied and `-v` is risky if the `-s` default is removed.

omnibus/package-scripts/inspec/postinst
```diff
@@ -41,6 +41,6 @@ done
 # be manually fixed.
 chown -Rh 0:0 $INSTALLER_DIR
 
-echo "Thank you for installing InSpec!"
+echo "Thank you for installing Chef InSpec!"
```
Schwad, line 44: This is interesting to know - do we use this lingo elsewhere or is the party line to brand "Chef InSpec" where possible? (I support this change just interested in the background)

Schwad (approved):
Thanks!

PR #5181, opened 2020-08-04
Bump faraday dependencies

### LINKED ISSUE #4234: Update faraday gem to support additional HTTP methods
Only some HTTP methods can be used in the http resource. `TRACE` and `CONNECT` are not supported at the moment so invoking the http resource with them fails.

```
inspec> http('https://www.example.org', method: 'TRACE').headers
ArgumentError: unknown http method: trace
from /opt/inspec/embedded/lib/ruby/gems/2.6.0/gems/faraday-0.15.4/lib/faraday/connection.rb:373:in `run_request'
```

## Detailed Description
Update the faraday gem once a new version has been released that includes the additional HTTP methods. The latest version 0.15.4 was released 27th Nov 2018.
 
## Context
I want to validate web server hardening measures. One of the typical steps is to verify that the TRACE method is not allowed. This should be possible with the built-in http resource instead of having to rely on invoking the curl binary.

## Possible Implementation
I've asked the faraday maintainers about a new release that includes the change for the HTTP methods so that the required functionality can be made available in InSpec: https://github.com/lostisland/faraday/issues/462

Schwad:
There are several outstanding issues about bumping faraday and
faraday_middleware. This gives us access to additional HTTP methods and
the versions we are pinned to are quite old.

Fixes #4234

This also helps us test out whether this bump passes our CI.

### DIFF
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -36,7 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "mixlib-log",         "~> 3.0"
   spec.add_dependency "sslshake",           "~> 1.2"
   spec.add_dependency "parallel",           "~> 1.9"
-  spec.add_dependency "faraday",            ">= 0.9.0"
+  spec.add_dependency "faraday",            ">= 0.9.0", "< 1.1"
   spec.add_dependency "tty-table",          "~> 0.10"
   spec.add_dependency "tty-prompt",         "~> 0.17"
   spec.add_dependency "tomlrb",             "~> 1.2.0"
--- a/inspec.gemspec
+++ b/inspec.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "train", "~> 3.0"
 
   # Used for Azure profile until integrated into train
-  spec.add_dependency "faraday_middleware", "~> 0.12.2"
+  spec.add_dependency "faraday_middleware", ">= 0.12.2", "< 1.1"
 
   # Train plugins we ship with InSpec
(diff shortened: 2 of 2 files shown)


Schwad:
CC @rmoles for our earlier discussion
CC @inspec/foundation-team-reviewers in case there are issues with faraday pinning that I am unaware of (I did have a quick search before pushing this issue, but don'e know if there's a Ruby 2.4 issue).

rmoles:
Thanks @Schwad 
CC: @omerdemirok

Schwad:
Looks like pipelines are complaining but unsure if it's faraday. What I might do is get these as green as I can, and then manually ping foundation to ensure there's no undocumented problem with touching faraday

omerdemirok:
Can `faraday` start from version `1.0`?
Some Ruby 2.7 related issues have been fixed in the major change, e.g. https://github.com/lostisland/faraday/issues/989.

Schwad:
We're hopeful to be able to support 1.0, I'm not sure if forcing >1.0 will
cause breaking changes for someone 🤔also need to double check that it
supports ruby 2.4.

I haven't anything coming flying in from foundation-team about this pinning
so fingers crossed there are no known issues

Schwad:
Rebuilding due to pipeline issue from last week that should hopefully be resolved

tas50 (approved):

Schwad:
Rebuilding this after wild (unrelated?) failures to see if we can get this merged

Schwad:
This died an unceremonious death when we had hab issues. @rmoles @omerdemirok sorry for the delay I'll see if I can reignite these builds

PR #5189, opened 2020-08-06
Resets expeditor to original settings

Schwad:
We have been trying to retry the failed promotion of InSpec 4.22.8. We have failed in that endeavor, so the final three steps will again be executed manually.

We are trying to put together a PR to document the right retry path so (we hope) this works in the future https://github.com/inspec/inspec/pull/5169


stevendanna (commented):
Is the submodule update intentional here?

Schwad:
> Is the submodule update intentional here?

Nice catch. That was something that shouldn't have gone in and has been cleared out

Issue #5003, opened 2020-04-30
Add benchmarking within InSpec Tests


Schwad:
## 🎛 Description

**Choose one**: 🙋 feature request

Something we will look at and continue to look at is performance. Not only is it an important element for the development team, but is a critical component of improving scaling support. There are elements of this gem that may get called thousands or millions of times on larger runs, so micro-improvements of 1 or 2 milliseconds runtime could be a game changer.

RSpec and other libraries actually maintain full benchmark records in their core repository, some for various methods and others for running their core library elements at varying scales.

A similar approach would be beneficial to seeing our growth and weak points moving forward for InSpec.

This is not an automation issue, but a foundation for us to try this new framework. If implemented and used, we can revisit automation benefits in future (6-12+ months).

## 👍 Acceptance Criteria
* initialize new `benchmarks` directory
* include a descriptive readme explaining benchmarks
* start with a view basic benchmarks, as runnable ruby files that include text data on prior benchmarking runs
* add new benchmarks when putting together new performance improvement issues, so that we can measure our performance improvement gains

## 🔗 Related Links

* RSpec Benchmarks https://github.com/rspec/rspec-core/tree/master/benchmarks

Aha! Link: https://chef.aha.io/features/SH-2357

james-stocks:
This is an awesome idea. I put the `Continuous Delivery` label on this to try to express that this is to give us benchmarking for our own dev/test/release and that it is not adding a benchmarking feature to InSpec

Schwad:
Thanks! Still trying to get a handle on all the labels

james-stocks:
One possible plan is to have expeditor run a job that does a couple of small invocations of InSpec e.g. `inspec --version` and `inspec exec test/fixture/profile/...` then send details of the timings, OS we ran on, and InSpec git ref to Chef's internal data store.

We could also look at a telemetry point that for `inspec exec`, but that is not this ticket.

Schwad:
So although a benchmarking folder is not off the table for internals (say individual components of InSpec for development, etc.), it looks like our first step here will be as @james-stocks indicated to do this capture during CI run and send to Chef internal datastore. 

Having a think this morning on what the logical approach might be, and drew something up. Open to comments if others have a different preferable approach. 

I drew this up (poorly) on my machine but I'm thinking first, we simply run `time` against `inspec exec benchmark_profile.rb`. That profile being as simple as possible. Perhaps we run this five times and take the averages to avoid outlier runs.

[...]

PR #5199, opened 2020-08-12
Optimize requires for non-omnibus installs

tas50:
require is quite slow in Ruby and doing requires for things you've already required is also slow. We've used this simple hack in Chef to speed up our requires. In the omnibus installs we patch how rubygems works to make this somewhat pointless, but this will help non-omnibus installs of InSpec.

### DIFF
--- a/lib/bundles/inspec-supermarket/api.rb
+++ b/lib/bundles/inspec-supermarket/api.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require "net/http"
-require "addressable/uri"
+require "net/http" unless defined?(Net::HTTP)
+require "addressable/uri" unless defined?(Addressable::URI)
 
 module Supermarket
   class API
--- a/lib/bundles/inspec-supermarket/target.rb
+++ b/lib/bundles/inspec-supermarket/target.rb
@@ -1,4 +1,4 @@
-require "uri"
+require "uri" unless defined?(URI)
 require "inspec/fetcher"
 require "inspec/fetcher/url"
 
--- a/lib/inspec/archive/zip.rb
+++ b/lib/inspec/archive/zip.rb
@@ -1,6 +1,6 @@
-require "rubygems"
+require "rubygems" unless defined?(Gem)
 require "zip"
-require "pathname"
+require "pathname" unless defined?(Pathname)
 
 module Inspec::Archive
   class ZipArchiveGenerator
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -1,4 +1,4 @@
-require "thor"
+require "thor" unless defined?(Thor)
 require "inspec/log"
 require "inspec/ui"
 require "inspec/config"
--- a/lib/inspec/cached_fetcher.rb
+++ b/lib/inspec/cached_fetcher.rb
@@ -1,5 +1,5 @@
 require "inspec/fetcher"
-require "forwardable"
+require "forwardable" unless defined?(Forwardable)
 
 module Inspec
   class CachedFetcher
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -66,7 +66,7 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: "A list of controls to include. Ignore all other tests."
   profile_options
   def json(target)
-    require "json"
+    require "json" unless defined?(JSON)
 
     o = config
     diagnose(o)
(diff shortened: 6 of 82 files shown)


lib/bundles/inspec-supermarket/api.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require "net/http"
-require "addressable/uri"
+require "net/http" unless defined?(Net::HTTP)
```
james-stocks, line 3: C compiler directive vibes :)

james-stocks:
This looks good to me, but might be wisest to merge after https://github.com/inspec/inspec/pull/5204

Schwad:
Thank you for this @tas50 ! 🙏 

Great point @james-stocks . This would be a fantastic candidate to try out after we get 

* #5204 (all the tests)
* And an answer on releng's https://github.com/chef/release-engineering/issues/1236 to resolve https://github.com/inspec/inspec/issues/5003 

-- so we could have documented performance/improvement on this issue. 

For those unsure about the extra time lost doing unnecessary requires, here's an example you can fire off today in your terminal:

```
require 'benchmark/ips'


Benchmark.ips do |x|

  x.config(time: 10, warmup: 4)

  x.report("launchy1") do
    require "launchy"
  end

  x.report("launchy10") do
    10.times do
      require "launchy"
    end
  end

  x.report("launchy100") do
    100.times do
      require "launchy"
    end
  end

  x.report("launchy1000") do
    1000.times do
      require "launchy"
    end
  end

  x.compare!
end

Calculating -------------------------------------
            launchy1      1.361M (± 1.9%) i/s -     13.698M in  10.071021s
           launchy10    131.324k (± 2.3%) i/s -      1.319M in  10.050078s
          launchy100     13.034k (± 2.3%) i/s -    130.391k in  10.010020s
         launchy1000      1.303k (± 3.3%) i/s -     13.029k in  10.012206s

Comparison:
            launchy1:  1360698.2 i/s
           launchy10:   131323.6 i/s - 10.36x  slower
          launchy100:    13033.6 i/s - 104.40x  slower
         launchy1000:     1302.8 i/s - 1044.44x  slower
```

And a comparison with @tas50 's solution here:

```
require 'benchmark/ips'


Benchmark.ips do |x|

  x.config(time: 10, warmup: 4)

  x.report("launchy") do
    10.times do
      require "launchy"
    end
  end

  x.report("launchy unless defined") do
    10.times do
      require "launchy" unless defined?(Launchy)
    end
  end

  x.compare!
end

Calculating -------------------------------------
             launchy    133.487k (± 2.2%) i/s -      1.341M in  10.051712s
launchy unless defined
                          1.871M (± 2.4%) i/s -     18.787M in  10.045053s

Comparison:
launchy unless defined:  1871374.2 i/s
             launchy:   133487.3 i/s - 14.02x  slower
```

Traditionally I would've expected in a gem to utilize as many of our requires as possible in https://github.com/inspec/inspec/blob/master/lib/inspec.rb but with a gem this size I'm sure we get to a point where the requiring often happens on a file-level.

Schwad:
#5204 will likely be getting merged in the near future, which moves this PR up. I wonder if this would also work using ruby's `autoload` or if that would make any difference. 🤔

tas50:
Some of these we'd want to later convert to autoload, but some you just can't really autoload so these are still helpful.

lamont-granquist:
@Schwad where autoload can be used it is even better than this.  That will prevent the gems from being even loaded once if they're never used, which really can't be beat.

Schwad:
@lamont-granquist cool. I may fire that up as a test PR to see if that approach burns everything down. If not, it would be neat to run a quick manual benchmark of those two PRs vs. master on a compliance-profile-level run. (&& also a simple small run with one control)

lamont-granquist:
Autoload isn't perfectly straightforwards text editing like this is.  You'll be better off shipping this as an incremental fix, then working on autoload.  We've autoloaded chef/chef and it just takes a week of beating on it or so working around the quirks of autoload.

lamont-granquist:
(and you'll still wind up needing to preserve some of these)

Schwad:
@lamont-granquist I do see what you mean. That test PR has required a bit of beating around but I might keep rolling with it.  Tentatively in my eyes this is "approved" pending 5204. Thanks all for the help and feedback

james-stocks (approved):
I believe the one risk that we're missing tests for (either currently or after #5204 merges) is that this repo breaks into 4 gems, and `require`s need to work after that split happens. The changes in this PR don't seem to overlap with that risk though, so this looks good to me!

PR #5202, opened 2020-08-13
Don't ship the readme in the gem

tas50:
We're just deleting this from chef and workstation anyways. It's buried deep in the install directory and doesn't do anyone any good.


james-stocks (approved):
Looks good to me, agreed that it's unlikely that someone would look in the gem dir for documentation.
Not sure what went wrong with CI, just clicked retry on it...

Schwad:
Looks great. Only case I could think of this affecting is someone running `bundle open` to try and see our README and having nothing. I think for us this is a vastly unlikely case and we expect our documentation to be accessed via the web.

Issue #5106, opened 2020-06-18
Run test/integration in CI


james-stocks:
We have a [`test/integration`](https://github.com/inspec/inspec/tree/master/test/integration) directory that we don't run in CI - this could and should be added to BuildKite verification

Acceptance Criteria:
* Get one of the tests within [`test/integration`](https://github.com/inspec/inspec/tree/master/test/integration) working
  * run-context-tk
  * run-context-after-audit
* Cut follow-up issue to get the other one green

Aha! Link: https://chef.aha.io/features/SH-2335

james-stocks:
During or after doing this, take a look at https://github.com/chef/chef/tree/master/kitchen-tests - work has gone into running a kitchen suite across a selection of OS images

clintoncwolfe:
As we a now have kitchen testing in #5204, I'm closing this.

PR #5204, opened 2020-08-13
Reinstate resource testing on supported platforms using Test-Kitchen

clintoncwolfe:
## Description

Re-enables basic testing of many InSpec resources on many supported target platforms by running tests using Test Kitchen and the dokken driver. The word "all" was not used in the previous sentence.

Updates the Kitchenfile:
 * Remove the usage of Audit Cookbook. Modern Audit cookbook insists on using its copy of InSpec, but we need to test InSpec from source, so we'll never be able to test using audit cookbook here. Audit cookbook is tested with InSpec in it own repo anyway; here we're trying to test InSpec, not Audit cookbook. It just added confusion and slowed the process, since by using audit cookbook we ran inspec twice - during converge and verify steps.
 * Remove configuration for a connection to Automate. AFAIK this was never enabled, and certainly is not present now. If we want to do so later, that should be a separate integration test unrelated to verifiying core resource support on supported platforms.
 * Rename the Kitchenfile from `kitchen.yml` to `kitchen.dokken.yml` to clarify that this is all the tests that can be run on the Dokken driver, and that is why they are grouped together. This file isn't any more special than the others; it should not be considered the default.
 * Update the list of platforms to the current offering. 

Concurrency is currently set to 5, which may need to be turned down if we see crashes, or up if we see timeouts. If we see both we'll need to turn it down and split the Kitchenfile.

Also, the platform oraclelinux-8 was found to have a defective image. I'll be opening an issue in the dokken-images repo.

## Related Issue

Partially implements #5106

## Types of changes

## Checklist:


kitchen.dokken.yml
```diff
@@ -39,47 +41,52 @@ platforms:
...
-
 - name: debian-10
   driver:
     image: dokken/debian-10
     pid_one_command: /bin/systemd
     intermediate_instructions:
       - RUN /usr/bin/apt-get update -y
 
-- name: fedora-29
+- name: fedora-30
```
tas50: You might want to just use fedora-latest here to avoid needing to update this

kitchen.dokken.yml
```diff
@@ -39,47 +41,52 @@ platforms:
...
     pid_one_command: /usr/lib/systemd/systemd
+# [2020-08-12T17:38:38+00:00] FATAL: RuntimeError: dnf_package[openssh-server] (ssh-hardening::server line 47) had an error: RuntimeError: dnf-helper.py had stderr output:
+# Errors during downloading metadata for repository 'ol8_baseos_latest':
+#   - Curl error (6): Couldn't resolve host name for https://yum$ociregion.oracle.com/repo/OracleLinux/OL8/baseos/latest/x86_64/repodata/repomd.xml [Could not resolve host: yum$ociregion.oracle.com]
+# - name: oraclelinux-8
+#   driver:
+#     image: dokken/oraclelinux-8
+#     pid_one_command: /usr/lib/systemd/systemd
 
 - name: opensuse-leap
```
tas50, line 64: this needs to get updated to the opensuseleap 15 box. 42 is EOL

test/kitchen/cookbooks/os_prepare/recipes/iptables.rb
```diff
@@ -1,8 +1,10 @@
 if platform_family?("rhel", "debian", "fedora", "amazon", "suse")
-  package value_for_platform_family(
```
tas50, line 2: This is the preferred way to do it, but the platform family should be rhel not centos and oracle which aren't valid platform families

james-stocks:
Thanks a million @clintoncwolfe :)

This seems to largely address https://github.com/inspec/inspec/issues/5106
The one thing we specifically need (and need to keep an issue open if this PR isn't doing it) is a test that requires InSpec as a gem and instantiates the runner class. I think since we're using kitchen here, kitchen-inspec will be doing that?

kitchen.dokken.yml
```diff
@@ -39,47 +41,52 @@ platforms:
...
 - name: oraclelinux-6
   driver:
     image: dokken/oraclelinux-6
     pid_one_command: /sbin/init
-
 - name: oraclelinux-7
   driver:
     image: dokken/oraclelinux-7
     pid_one_command: /usr/lib/systemd/systemd
+# [2020-08-12T17:38:38+00:00] FATAL: RuntimeError: dnf_package[openssh-server] (ssh-hardening::server line 47) had an error: RuntimeError: dnf-helper.py had stderr output:
```
james-stocks: Is this just leaving a note for why Oracle 8 isn't enabled? Suggest adding a TODO line to make that clear

james-stocks:
@Schwad I think the tests haven't run for this PR because buildkite runs from the config on master, not the PR branch. We should merge this at a time when we're happy to troubleshoot any stability issues afterwards.

Schwad:
Thank you so much for this @clintoncwolfe ! 🙏 

I'll be sure to pull it down and have a look today 😄 Good point @james-stocks 👍 

I'm almost certain the answer is "absolutely not", but I'll ping a question to releng to see if there's any sneaky path that those with repo-write access can trigger a build/expeditor/buildkite against a PR. For PRs like this that would be quite handy.

Schwad:
Is the path for local running still `bundle exec kitchen test oraclelinux-7` or `bundle exec kitchen test` for all? 🤔 

Looks relatively happy. Putting the question now to releng

Schwad:
I'm linking an internal message here just so I can find it when I'm back Monday - but looks like Expeditor does have a great pattern for implementing stuff like this with "release branches" which can help us accomplish this. I'll try to get this set up so we can see how we're looking on the full expeditor run 🎉  https://chefio.slack.com/archives/CB71Y7BQS/p1597762475099900

Example doc: https://expeditor.chef.io/docs/patterns/version-management/#release-branches

Schwad:
I'm thinking an `expeditor-staging` release branch would be nifty for this and in the future, will have a crack at it monday

Schwad:
Another flow discussed from Releng here: https://expeditor.chef.io/docs/pipelines/general/#creating-an-empty-pipeline

(internal ref: https://chefio.slack.com/archives/CB71Y7BQS/p1597769565104900?thread_ts=1597762140.096600&cid=CB71Y7BQS )

james-stocks:
It's not essential to change how expeditor works - we can merge this now (conflicts aside); we just need to do it at a time when we're OK to spend any time needed to fix any teething problems. i.e. not on a Friday :D

kitchen.dokken.yml
```diff
@@ -1,35 +1,37 @@
...
-  chef_version: 14.12.9
+  chef_version: :latest
   privileged: true # because Docker and SystemD/Upstart
 
 transport:
   name: dokken
 
 lifecycle:
+  # The purpose of this code is to build the InSpec gems from source and then make them available
+  # to the install_inspec cookbook, which will install the gems. This means that the inspec used
```
james-stocks: I'm confused how this works - these commands are setting up for the `install_inspec` recipe to install the current InSpec on the test instance. But kitchen's inspec verifier runs the local copy of InSpec (i.e. the one in the current bundle) against the test instance as a target.

Am I missing some kitchen magic here where the InSpec installed on the instance is used for the verify phase instead of the local one?

I can't quickly test this locally because I'm getting this error (which I suspect is something dumb/untidy in my local environment)

```
-----> Converging <resources-ubuntu-2004>...
       Creating kitchen sandbox in /Users/jstocks/.dokken/kitchen_sandbox/fe0834e046-resources-ubuntu-2004
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 7.0.10...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
/opt/chef/embedded/lib/ruby/2.7.0/rubygems/dependency.rb:313:in `to_specs': Could not find 'inspec-core' (= 4.23.4) - did find: [inspec-core-4.23.10] (Gem::MissingSpecVersionError)
Checked in 'GEM_PATH=/root/.gem/ruby/2.7.0:/opt/chef/embedded/lib/ruby/gems/2.7.0', execute `gem env` for more information
	from /opt/chef/embedded/lib/ruby/2.7.0/rubygems/dependency.rb:323:in `to_spec'
	from /opt/chef/embedded/lib/ruby/2.7.0/rubygems/core_ext/kernel_gem.rb:62:in `gem'
	from /opt/chef/bin/chef-client:63:in `<main>'

```

.expeditor/integration.resources.yml
```diff
@@ -1 +1,20 @@
...
+steps:
+  - label: integration-test-kitchen
+    command:
+      - RAKE_TASK=test:integration /workdir/.expeditor/buildkite/verify.sh
+    expeditor:
+      executor:
+        docker:
+          environment:
+            - CONCURRENCY: 5
+            - DOCKER: 1
```
james-stocks, line 19: I think we need to set `KITCHEN_YAML=./kitchen.dokken.yml` here to avoid kitchen trying to run the default `kitchen.yml`

clintoncwolfe:
At this point this runs locally OK for me with `-c 3`; make sure you have the latest test-kitchen and kitchen-inspec in your bundle, as the most recent versions have changes that affect concurrency. 
It is rather slow locally for me - about 35 minutes. We could break that up into separate pipelines and have manual parallelization.

james-stocks:
The last time I tried this branch I was still getting the `Gem::MissingSpecVersionError` error similar to 

```
/opt/chef/embedded/lib/ruby/2.7.0/rubygems/dependency.rb:313:in `to_specs': Could not find 'inspec-core' (= 4.23.4) - did find: [inspec-core-4.23.10] (Gem::MissingSpecVersionError)
```

I assume this error originates from `/opt/chef` on the dokken image, not `/opt/chef` on my laptop. I'm confused why the Chef run is hung up on wanting InSpec `= 4.23.4` when nothing in this branch specifies that version.

I'm running:
`CONCURRENCY=3 DOCKER=1 KITCHEN_YAML=./kitchen.dokken.yml bundle exec rake test:kitchen`

Issue #5206, opened 2020-08-14
SSL client certificate for the http resource


micheelengronne:
I would like to use the http resource on an endpoint that requires a ssl client certificate.

## Detailed Description
I propose to add 3 parameters to the http resource:

* ssl_client_cert
* ssl_client_key
* ssl_client_cacert

## Context
It would make the http resource available to test strongly encrypted endpoints.

brandon-welsch:
Hi, i need the same feature. And i'm disappointed to see that a well-known used feature is still missing on this resource.

Can an update be done, on the related PR? https://github.com/inspec/inspec/pull/5207

PR #5207, opened 2020-08-14
WIP: Add ssl client parameters to the http resource

micheelengronne:
## Description
The http resource can now be used on endpoints requiring a TL client certificate.

## Related Issue
https://github.com/inspec/inspec/issues/5206

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/http.rb
+++ b/lib/inspec/resources/http.rb
@@ -115,6 +115,18 @@ module Inspec::Resources
         def ssl_verify?
           opts.fetch(:ssl_verify, true)
         end
+        
+        def ssl_client_cert?
+          opts.fetch(:ssl_client_cert, nil)
+        end
+        
+        def ssl_client_key?
+          opts.fetch(:ssl_client_key, nil)
+        end
+        
+        def ssl_client_cacert?
+          opts.fetch(:ssl_client_cacert, nil)
+        end
 
         def max_redirects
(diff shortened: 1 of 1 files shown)


micheelengronne:
In order to write unit tests, I need to create a TLS CA, certificate and key. How do you handle that ?

Schwad:
Hi @micheelengronne - sorry for the delay here. I'm not 100% on the best answer to your question, it may be something @clintoncwolfe has thoughts on? 🤔

clintoncwolfe:
letsencrypt.org ?

micheelengronne:
Is it not a bit too much to create a Let'sEncrypt signature in order to mock an endpoint with a TLS client certificate ?

I thought to use a generated certificate/key from a local CA certificate. Put that in a sort of `tests/assets` folder and use them to imitate an https endpoint. But I have no clue on how to handle that.

Schwad:
@micheelengronne all of our testing around the http resource is currently covered with webmock tooling, with which SSL/TLS testing is another layer removed: https://github.com/bblimke/webmock/issues/757

However it appears that @notpace had a crack at something similar to doing some TLS testing while using webmock which might be a good inspiration for some unit tests: https://github.com/onc-healthit/inferno/pull/32

aaronlippold:
It's seems this may be useful still and given we just did some other fixing on this response should this addition be pulled in as well

PR #5208, opened 2020-08-14
Enable profile-bundled plugins

thheinen:
## Description

_PR for discussion about this topic, following an example implementation_

In some bigger profiles, we encounter the need to add profile specific DSL methods as the controls themselves are handled by non-programmers. As these specifics are only valid in this one profile, we would like to bundle a plugin in the `libraries` path instead of spamming Gem files in 1:1 relations.

So far, this seems to trigger some test-related code in [`loader.rb`](https://github.com/inspec/inspec/blob/7888738cc096e92a9822cb8a8a17c21d1dc2968e/lib/inspec/plugin/v2/loader.rb#L64-L68) and result in a crash:

```
-----> Verifying <two-amazon2>...
[2020-08-14T14:38:52+00:00] ERROR: Could not load plugin profile-bundled: cannot load such file -- inline.rb
[2020-08-14T14:38:52+00:00] ERROR: Errors were encountered while loading plugins...
[2020-08-14T14:38:52+00:00] ERROR: Plugin name: profile-bundled
[2020-08-14T14:38:52+00:00] ERROR: Error: cannot load such file -- inline.rb
[2020-08-14T14:38:52+00:00] ERROR: Run again with --debug for a stacktrace.
```

The circumstances are pretty special: the run must be from within kitchen (`kitchen verify`) with more than one instance under test. The second instance under test then throws the error (maybe related to Singleton usage in Registry?).

My proposition is to skip the code in question for "profile-bundled" plugins to enable this sort of extension. My implementation is only looking for a name prefix so far, but maybe a way to set a plugin installation method to `:profile_bundled` would be cleaner. 

## Related Issue

No filed issue yet

## Types of changes

## Checklist:

### DIFF
--- a/docs/dev/plugins.md
+++ b/docs/dev/plugins.md
@@ -30,6 +30,12 @@ The normal distribution and installation method is via gems, handled by the `ins
 
 For more on the `plugin` CLI command, run `inspec plugin help`.
 
+### Plugins can be bundled with a profile
+
+If plugin code is tightly coupled to one profile only, you can bundle it into the profile by placing it into the `libraries/` directory.
+
+The plugin name must be prefixed with "profile" to handle loading properly.
+
 ### Plugins may also be found by path to a source tree
 
 For local development or site-specific installations, you can also 'install' a plugin by path using `inspec plugin`, or edit `~/.inspec/plugins.json` directly to add a plugin.
--- a/lib/inspec/plugin/v2/registry.rb
+++ b/lib/inspec/plugin/v2/registry.rb
@@ -50,6 +50,12 @@ module Inspec::Plugin::V2
       known_plugin?(name.to_sym) && registry[name.to_sym].installation_type == :path
     end
 
+    def profile_bundled_plugin?(name)
+      return unless known_plugin?(name.to_sym)
+
+      registry[name.to_sym].installation_type == :profile_bundled || name.start_with?("profile")
+    end
+
     def find_status_by_class(klass)
       registry.values.detect { |status| status.plugin_class == klass }
     end
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -49,6 +49,9 @@ module Inspec::Plugin::V2
       # Be careful not to actually iterate directly over the registry here;
       # we want to allow "sidecar loading", in which case a plugin may add an entry to the registry.
       registry.plugin_names.dup.each do |plugin_name|
+        # Skip plugins bundled with profiles, which get loaded automatically
+        next if registry.profile_bundled_plugin?(plugin_name)
+
         plugin_details = registry[plugin_name]
         # We want to capture literally any possible exception here, since we are storing them.
         # rubocop: disable Lint/RescueException


Schwad:
Quick aside - I do not know why `config-validation` is failing this PR as it does not touch expeditor. My suspicion is that there may have been new expeditor validations added so will check that out

Schwad:
Actually there are no changes - is this PR branched off of inspec:master? I only ask because we did have an expeditor issue recently. If not, might need a master rebase. 

Thanks for the PR and for kicking off the discussion! 😄 I am interested in what some of the other inspec-core and @inspec/foundation-team-reviewers might think about this case

lib/inspec/plugin/v2/registry.rb
```diff
@@ -50,6 +50,12 @@ def path_based_plugin?(name)
       known_plugin?(name.to_sym) && registry[name.to_sym].installation_type == :path
     end
 
+    def profile_bundled_plugin?(name)
+      return unless known_plugin?(name.to_sym)
+
+      registry[name.to_sym].installation_type == :profile_bundled || name.start_with?("profile")
```
clintoncwolfe, line 56: Unfortunately, the inspec codebase mixes strings and symbols freely. There are some test failures on ruby 2.4-2.6 from `name.start_with?("profile")` because `name` is a `Symbol`, and `Symbol#start_with?` is only defined in Ruby 2.7+. I suggest `name.to_s.start_with?`.

clintoncwolfe (commented):
I think this is an interesting idea. I'd like to see a functional test - add a profile that includes a profile-bundled plugin into `test/fixtures/profiles/` and one or more tests into `test/functional/plugins_test.rb` that exercise InSpec running the profile and somehow demonstrating that the plugin correctly loaded. Thanks!

docs/dev/plugins.md
```diff
@@ -30,6 +30,12 @@ The normal distribution and installation method is via gems, handled by the `ins
 
 For more on the `plugin` CLI command, run `inspec plugin help`.
 
+### Plugins can be bundled with a profile
+
+If plugin code is tightly coupled to one profile only, you can bundle it into the profile by placing it into the `libraries/` directory.
+
+The plugin name must be prefixed with "profile" to handle loading properly.
```
james-stocks, line 37: It would be good to give an example of what a plugin inside `libraries` looks like, that fixture could be linked to (or repeated) here.

We should agree on `^profile.*` meaning a profile bundled plugin (i.e. we cannot start the name with `profile` for a plugin not bundled with the profile, or for other code like a custom resource that is bundled in `libraries`)

clintoncwolfe:
Is there still interest in doing this? At this point I don't think we're looking to move in this direction - we're adding the ability for profiles to directly rely on gems which relieves some of the pressure here. 
If there is strong contributor interest in this, it could be brought up to releasable state and we'd accept it, but I don't think we'll adopt it otherwise. I'll check back in a few weeks and if there is no response, I'll close this PR.

thheinen:
From my point of view, having a way to somehow add plugins to a profile - be it in a bundled state or as external dependency - is the main need.

If you add a Gem dependency mechanism I am totally fine with that as a solution :+1:

thheinen:
Damn sorry, clicked the wrong button. didn't intend to close. Oops!

aaronlippold:
This actually kind of works already if you do the right tricks in the control file if you don't worry about putting your new functionality in a formal plugin. Those I agree with the ablity to package plugins with a profile. 

We actually have a core use case for the kubernetes benchmarks on high side and sensitive networks that could very much benifit from being able to do this.

I use a 'require_releative' and can include ramdom ruby libs in my controls - https://github.com/mitre/aws-s3-baseline/blob/master/controls/aws_s3_bucket_objects.rb#L1

However, I have some issues when running the profile from git or url vs a local profile from the runner. Works fine with the local run of a cloned profile but I get the above error.

I have not tried digging into why it isn't loading all the ruby in the libraries directory prior to execution of the profile - that was the orginal design. I would guess some things got broken when we hacked with lo ading so that the inspec shell could load dependant profiles and resource packs - I know that had some trickery involved in it.

The challenge with the 'put it in a gem' approach is that you then put the end user on the spot to build a gem, to have a place to put it, even a challenge if they don't want to publish to rubygems etc etc. 

If we could get the loading of all ruby in the library directory to be in scope at execution time of the profile I think that would give end users the most flexibility. 

My above example - even though it currently only works when cloned to the runner - is a good example of the power of being able to do this at the profile level. I was able to increase the efficency of auditing my s3 bucket object about 100x

PR #5211, opened 2020-08-17
Add ability to disable input caching for kitchen-inspec use

thheinen:
## Description

When trying to test with input plugins from `kitchen-inspec` the [caching of past values for an input](https://github.com/inspec/inspec/blob/c5ee0ea5970376e48f5eabe4ecddd71c02c52809/lib/inspec/input_registry.rb#L87-L88) will cause problems.

With a `kitchen.yml` like ...

```yaml
verifier:
  load_plugins: true

suites:
  - name: one
    run_list:
      - recipe[test_cookbook]
    attributes:
      value: one
    control:
      - cache_check
  - name: two
    run_list:
      - recipe[test_cookbook]
    attributes:
      value: two
    control:
      - cache_check
```
... and using the `inspec-chef` plugin via ...

```ruby
control 'cache_check' do
  describe file('/tmp/') do
    let(:value) { input('node:///attributes/value') }

    its('content') { should cmp value }
  end
end
```

... only the first suite gets the correct attribute value "one" and every other suite gets "one" as well as the input name is identical.

This turned into a blocker for a project where a cookbook is supposed to install different MS SQL versions/editions controlled by data bags as central site configuration. Having one `kitchen converge` will access the cached values due to use of `Singleton` in `InputRegistry`. Making a separate `kitchen` call for each suite/platform combination would work.

I added a setting `cache_inputs` (default: `true`) to the `InputRegistry` which is accessible via API and added as separate condition.

For `kitchen-inspec` related changes, see [kitchen-inspec PR #258](https://github.com/inspec/kitchen-inspec/pull/258)

## Related Issue

No filed issue

## Types of changes

## Checklist:

### DIFF
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -285,6 +285,20 @@ As of Chef InSpec 4.12, this mechanism has the following limitations:
   1. No [input options](#input-options-reference) may be set - only the name and value.
   2. Because the CLI is outside the scope of any individual profile and the inputs don't take options, the inputs are clumsily copied into every profile, effectively making the CLI mechanism global.
 
+## Setting Input values using Plugins
+
+Inputs can also be set by custom input plugins, which retrieve values from external sources like secret stores or databases. Please check Rubygems.Org for available InSpec input plugins.
+
+### Disabling Caching for Inputs
+
+Especially with plugins, it can be desirable to re-evaluate inputs every time and not cache them. By default, an existing input value is reused which can lead to problems if the retrieved values are expected to change. An example for this is using `kitchen-inspec` with input plugins to connect to a Vault server for password retrieval.
+
+To disable input caching, you can disable the cache from your Ruby code:
+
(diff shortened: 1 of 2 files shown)


clintoncwolfe (commented):
I like this change in general - especially since it solves a serious real-world problem. My only ask is that the feature be documented.  I think `docs/inputs.md`, in the section Advanced Topics, might be the best place.  Thanks!

thheinen:
I noticed the absence of an "Inputs from Plugins" section and took the freedom to make it a subheading there. Not sure if that's a good way or Advanced would've been better. What do you think?

Schwad (approved):
I like the currently planned heading, thanks for the PR and looks good! 👍 

I've unblocked the builds so they all run and we should be good once this is green

clintoncwolfe (approved):
Thanks!

PR #5219, opened 2020-08-24
Fix mysql_session stdout, stderr and exit_status parameters

### LINKED ISSUE #5218: mysql_session fails with undefined method `stdout'
I've confirmed that #5124 introduced the following bug:

```
  ×  mariadb_database: test creation and removal of databases (2 failed)
     ×  MySQL query: show databases stdout 
```

You can also see it in action [here](https://github.com/sous-chefs/mariadb/pull/313/checks?check_run_id=1010399944).

If I roll back to version 4.22.1 this problem goes away, but when I update to 4.22.2 and beyond, this issue is happening.

ramereth:
A feature introduced in #5124 changed how the mysql_session resource works to
where stdout, stderr, and exit_status no longer worked. The Lines class only
exposes output instead of stdout, stderr and exit_status.

As a work around, this adds backwards compatibilty for stdout, stderr and
exit_status parameters. In addition I updated the documentation to prefer output
over stdout/stderr.

This fixes #5218.

## Description

I wasn't sure if we wanted to deprecate ``stdout``and ``stderr`` or not so please let me know if we do.

## Related Issue
- #5218 

## Types of changes

## Checklist:

### DIFF
--- a/docs/resources/mysql_session.md.erb
+++ b/docs/resources/mysql_session.md.erb
@@ -24,14 +24,14 @@ This resource first became available in v1.0.0 of InSpec.
 A `mysql_session` resource block declares the username and password to use for the session, and then the command to be run:
 
     describe mysql_session('username', 'password').query('QUERY') do
-      its('stdout') { should match(/expected-result/) }
+      its('output') { should match(/expected-result/) }
     end
 
 where
 
 * `mysql_session` declares a username and password, connecting locally, with permission to run the query
 * `query('QUERY')` contains the query to be run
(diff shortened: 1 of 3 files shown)


fzipi:
Good catch!

rmoles (approved):
This seems like a reasonable change to make. Bonus for updating the documentation to prefer output.

Schwad (approved):
Thank you so much for this @ramereth - great catch! We'll get this going and in the next release to revert the issue

PR #5229, opened 2020-08-27
Allow license-acceptance 2.x gem dep

tas50:
This release bumps the tty deps so we can bring in newer versions of
those libs in chef and workstation.


Schwad (approved):
Thanks @tas50 - looks good

PR #5232, opened 2020-08-31
Run ps with an ulimited width on linux

### LINKED ISSUE #5226: command line returned by processes is truncated
On some systems the output of ps may be truncated, which may prevent profiles from working as intended.

#1173 intended to fix that issue but it doesn't cover what is a common behaviour. From ps' manpage:

> When specified last, this column will extend to the edge of the display.  If
> ps can not determine display width, as when output is redirected (piped) into
> a file or another command, the output width is undefined (it may be 80,
> unlimited, determined by the TERM variable, and so on).  The COLUMNS
> environment variable or --cols option may be used to exactly determine the
> width in this case.  The w or -w option may be also be used to adjust width.

(ps from procps-ng)

# Describe the problem
Example of a test that should be listing some firefox processes in my machine:

> inspec> processes(/omni/).entries
> => []

By locally passing a COLUMNS env var set to 9999 the problem is noticeable (COLUMNS=9999 inspec shell):
> inspec> processes(/omni/).entries.size
> => 3

## Possible Solution
Call ps with ww, or with a very large COLUMNS parameter. E.g.
```ruby
command = "ps wwaxo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command"
```
Resulting in:

> inspec> processes(/omni/).entries[0].command
> => "/usr/lib/firefox-esr/firefox-esr -contentproc -childID 1 -isForBrowser -prefsLen 1 -prefMapSize 205372 -parentBuildID 20200720181548 -greomni /usr/lib/firefox-esr/omni.ja -appomni /usr/lib/firefox-esr/browser/omni.ja -appdir /usr/lib/firefox-esr/browser 559903 true tab"

rgeissert:
## Description

On some systems the output of ps may be truncated, which may prevent
profiles from working as intended.

Some implementations specifically mention that the output width is
undefined "(it may be 80, unlimited, determined by the TERM variable,
and so on)".

## Related Issue

This fixes #5226 with the fix mentioned there.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/processes.rb
+++ b/lib/inspec/resources/processes.rb
@@ -138,7 +138,7 @@ module Inspec::Resources
           command: 8,
         }
       else
-        command = "ps axo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command"
+        command = "ps wwaxo label,pid,pcpu,pmem,vsz,rss,tty,stat,start,time,user:32,command"
         regex = /^(.+?)\s+(\d+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+([^ ]+)\s+(\w{3} \d{2}|\d{2}:\d{2}:\d{2})\s+([^ ]+)\s+([^ ]+)\s+(.*)$/
         field_map = {
           label: 1,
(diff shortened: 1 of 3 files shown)


Schwad:
I'm wondering if this may not be as straightforward as expected, looks like this change breaks some things down the line (re: tests)

Schwad:
Have you tried this out locally (`bundle exec inspec`) and if so what were the results?

rgeissert:
I force-pushed a change to the tests so that the command matches the one from processes.rb and so that the mock_loader correctly loads the mock's output. I think that should fix the failing tests.

Schwad:
Looks great @rgeissert thank you much. I think this is ready but am going to hold fire and request a quick look from @clintoncwolfe to make sure I'm not missing out on anything else edge-case-y here. 👍

clintoncwolfe (approved):
This seems well-guarded to be just linux, non-busybox - we want to avoid BSDs and solaris, etc. so I think this seems fine.  Thanks!

PR #5238, opened 2020-09-01
Use powershell-native regex escape method in package resource

### LINKED ISSUE #5235: Incorrect escaping of Windows package names
https://github.com/inspec/inspec/blob/d00706b9a6c01d4a345d70f082d612e06b5c60db/lib/inspec/resources/package.rb#L317

This uses `shellescape` to escape the string for the regex. This works fine for most things, but is designed to escape for BASH, not regex. In fact, if the string has a period in it, it is not escaped properly. Instead, we should use powershell's built in method for escaping for regex: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_regular_expressions?view=powershell-5.1#escaping-characters

lavahot:
Fixes #5235

Signed-off-by: Taylor Mansfield <lavahot@gmail.com>

## Description
shellescape() creates a string that is escaped properly for bash, but doesn't work in powershell. This uses the built-in method in the regex module, escape, to properly escape the string.

## Related Issue

#5235 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/package.rb
+++ b/lib/inspec/resources/package.rb
@@ -314,7 +314,7 @@ module Inspec::Resources
       # Find the package
       cmd = inspec.command <<-EOF.gsub(/^\s*/, "")
         Get-ItemProperty (@("#{search_paths.join('", "')}") | Where-Object { Test-Path $_ }) |
-        Where-Object { $_.DisplayName -match "^\s*#{package_name.shellescape}\.*" -or $_.PSChildName -match "^\s*#{package_name.shellescape}\.*" } |
+        Where-Object { $_.DisplayName -match "^\s*$([regex]::escape(#{package_name}))\.*" -or $_.PSChildName -match "^\s*$([regex]::escape(#{package_name}))\.*" } |
         Select-Object -Property DisplayName,DisplayVersion | ConvertTo-Json
       EOF
 


Schwad:
Thank you for this @lavahot - good to know!

It looks like it's currently failing the "verify windows package parsing" test in `package_test.rb:104`, I wonder if it's missing something 🤔

lavahot:
Yeah @Schwad, there's something not quite right there. I'll take a look tomorrow.

Schwad:
For anyone stumbling across this:

We have one block of three assertions testing windows package resource, here:

```ruby
it "verify windows package parsing" do
    resource = MockLoader.new(:windows).load_resource("package", "Chef Client v12.12.15")
    pkg = { name: "Chef Client v12.12.15 ", installed: true, version: "12.12.15.1", type: "windows" }
    _(resource.installed?).must_equal true
    _(resource.version).must_equal "12.12.15.1"
    _(resource.info).must_equal pkg
  end
```

And we're currently failing on the assertion ` _(resource.installed?).must_equal true`

ramereth:
@Schwad @lavahot if you make the following change to the mock data, it should pass now:
```diff
diff --git a/test/helpers/mock_loader.rb b/test/helpers/mock_loader.rb
index 7e1f9ff42..acda51190 100644
--- a/test/helpers/mock_loader.rb
+++ b/test/helpers/mock_loader.rb
@@ -269,7 +269,7 @@ class MockLoader
       "rmsock f0000000000000001 tcpcb" => cmd.call("rmsock-f0001"),
       "rmsock f0000000000000002 tcpcb" => cmd.call("rmsock-f0002"),
       # packages on windows
-      "f7718ece69188bb19cd458e2aeab0a8d968f3d40ac2f4199e21cc976f8db5ef6" => cmd.call("get-item-property-package"),
+      "6bd6b6b16f630095a1839c43421e23a78474c2829944da9a4f3ee7f85b8c3744" => cmd.call("get-item-property-package"),
       # service status upstart on ubuntu
       "initctl status ssh" => cmd.call("initctl-status-ssh"),
       # upstart version on ubuntu
```
This is a sha256sum of the new command that gets generated with the new code.

lavahot:
I apologize for not following up here sooner, been slammed at work.

So I actually figured out the specific character that was causing issues for me. I'm installing some Visual C++ Runtime, and the bash escaping escapes the '+' character incorrectly, interpreting it as a regex '+', not a literal '+'. And there is no way to coax it into being escaped correctly. I worked around this by using '.' in place of '+', which gets interpreted as a regex '.', which works perfectly fine for my purposes. 

I'll have some additional time to work on this some time soon. I'd like to build a test case that is specifically looking at characters that should be escaped when used in a regex.

ramereth:
@lavahot IMHO it might be better to let the user escape regexes instead of trying to figure out every use case. I was actually working on fixing the vcruntime cookbook and came across while testing it with the same problem you described. 

Would it be an issue to just remove the escape all together and add some documentation that this uses a regex and would need to be escaped when on Windows?

clintoncwolfe:
Because the approach works without modifying the input, we've chosen to go with the approach in #5323, so I'm going to close this - thanks for the discussion and diagnosis!

PR #5241, opened 2020-09-04
InSpec Cloud: Reporting by Resource

Schwad:
This is a (currently draft) PR being used to test, spike, and eventually implement the cloud workload of reporting by resource in conjunction with the automate team. 

Certain elements of the workload will be changes coming from the InSpec side, certain elements will come from the Automate side.


test/functional/inspec_exec_json_test.rb
```diff
@@ -188,6 +188,30 @@
...
+    describe 'results' do
+      let(:result) { ex1["results"][0] }
+      let(:result2) { ex2["results"][0] }
+
+      it "has a code_desc" do
+        _(result["code_desc"]).must_equal "File / is expected to be directory"
+      end
+
+      it "has a resource_name" do
+        _(result["resource_name"]).must_equal "file"
```
james-stocks: `file` would be the resource `class` (not `name`) in my opinion.

For a cloud example:

```
google_compute_instance(project: gcp_project_id, zone: zone_name, name: 'bitcoin_miner_2455')
```

...we need both the class `google_compute_instance` and the name `bitcoin_miner_2455`

A problem would be that resources don't have a common `name` property. I think the only thing they commonly implement currently is `to_s`.

I would like to double check with @kekaichinose what the ask here is

Schwad: I don't know if a resource can be composed without a `name` (as I believe that is also what is used to invoke the resource? and that's what automate were looking for) i.e. https://docs.chef.io/inspec/dsl_resource/#resource-structure.

We'd looked into `to_s` but automate wasn't sure as it can take many forms but they wanted the textual representation of the resource (i.e. `google_compute_instance`)

However even though it's called a `name` in the DSL that might be a clogged namespace. `class` I would find slightly muddling as a rubyist since it's what I'm pulling off of `name` 🤔 ... but calling it `resource_class` might be the way to go. 

I wouldn't even be opposed to a longer internal naming such as `name_of_the_resource` 😅

Schwad:
This PR is being held onto for its ideas but is currently on hold as we anticipate future planning/meetings with our automate friends. :)

test/functional/inspec_exec_json_test.rb
james-stocks: Taking a second look at this PR, it looks like the term `resource_name` is [already being used](https://github.com/inspec/inspec/pull/5241/files#diff-6fbf0fe2ad4e4d2f8fd70d7180517ef1bb9457a00fef17bcb0e0fc3f00cdb062R37) and debating the term used here might not be helpful.

I do feel it's a mistake to take the `name` from the DSL (which is concerned with defining classes) and calling it `name` in reporting (which is concerned with runtime _instances_ and not classes). It's like having an `employee` record where an employee has `name: 'employee'`

Schwad: Sorry, just saw this. I understand that we already use `resource_name` internally as our variable name which inspired keeping the name the same when we report out to automate. 

However I think it's okay to report this out to automate under any moniker that we deem fitting, so I'm making the change to `resource_class` and happy to further alter under review

james-stocks:
Looks OK to me except:
 * we should get @alexpop to +1 this maybe
 * commits need squashed
 * pry breakpoint needs removed
 * my previously stated gripe about field naming (OK to merge this as-is though)

Schwad:
cheers @james-stocks 👍 

gonna grab a coffee and see if this baby passes CI. There were some outstanding breaks I never got to. Also going to try and hop on a chat with @alexpop this morning if possible

Schwad:
^^ rebased in master with squash so hopefully addressed merge conflicts as well

Schwad:
There is one *weird* test. Saying that there's a diff on expectations and failing. It failed on me locally a few hours ago but no more. Don't know if it's a dependency thing where my bundle has something different from CI. 

By the way, the diff isn't even a diff. It's like

- hi
+ hi

Could be a whitespace thing but I can't reproduce at the moment. Approached the test from a different angle and seeing if that works on CI.

alexpop (approved):
Thank you lads!

PR #5250, opened 2020-09-15
WIP: Autoload instead of requires where possible

Schwad:
This should run faster for non-omnibus installs, if it doesn't break the world. (Check PR for
benchmark numbers)

Autoload only loads elements as needed and if they are not already
defined.

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-286

### DIFF
--- a/lib/bundles/inspec-supermarket/api.rb
+++ b/lib/bundles/inspec-supermarket/api.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require "net/http"
-require "addressable/uri"
+autoload :Net::HTTP
+autoload :Addressable::URI
 
 module Supermarket
   class API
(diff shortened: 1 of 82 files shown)


Schwad:
A related issue was recently closed, but for those coming here there's primary discussion and context around this generally at: https://github.com/inspec/inspec/pull/5199

The current tasklist for this PR is:

- [ ] get pipelines passing (autoload isn't completely a drop-in text replacement as @lamont-granquist pointed out)
- [ ] do manual benchmarking on local machine to record any performance improvements
- [ ] await slightly more robust test additions (possibly something like #5204) before merging due to this concern: https://github.com/inspec/inspec/pull/5199#pullrequestreview-498428763

Schwad:
This can be revisited in future but going to close this off for now -- as pointed out in a meeting today these are changes we'd love to just wing out and benchmark but require more robust testing to do this with confidence.

PR #5252, opened 2020-09-16
Accept arrays in parse_config's method_missing

rgeissert:
## Description
A single-element array allows Rspec's its behaviour to be worked around
and allow options containing dots to be tested using its.

This is already implemented by resources such as `json` and those based
on it (e.g. xml).

N.b. I feel like I'm missing something with regards to the documentation as I see it is in two directories. Is there some automagic tool I forgot to run?
Note as well that in the documentation I mention version `4.23.9` as the future version including this change, so feel free to correct it as needed.

## Related Issue
Related to issue inspec#875.

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/parse_config.md
+++ b/docs-chef-io/content/inspec/resources/parse_config.md
@@ -53,6 +53,10 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
+## Examples
+
+This resource is based on the `parse_config_file` resource. As such please refer to the examples of the [`parse_config_file`](/inspec/resources/parse_config_file) resource.
+
 ## Matchers
 
 For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
--- a/docs-chef-io/content/inspec/resources/parse_config.md
+++ b/docs-chef-io/content/inspec/resources/parse_config.md
@@ -53,13 +53,20 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
-## Examples
+## Options
 
-This resource is based on the `parse_config_file` resource. As such please refer to the examples of the [`parse_config_file`](/inspec/resources/parse_config_file) resource.
+This resource supports the following options for parsing configuration data. Use them in an `options` block stated outside of (and immediately before) the actual test:
 
-## Matchers
+    options = {
+        assignment_regex: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
+        multiple_values: true
+      }
 
-For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
+    output = command('some-command').stdout
+
+    describe parse_config(output,  options) do
+      its('setting') { should eq 1 }
+    end
 
 ### assignment_regex
 
@@ -118,3 +125,12 @@ Use `standalone_comments: false`, to parse the following:
 
     'key = value # comment'
     params['key'] = 'value'
+
+
+## Examples
+
+This resource is based on the `parse_config_file` resource. As such please refer to the examples of the [`parse_config_file`](/inspec/resources/parse_config_file) resource.
+
+## Matchers
+
+For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
--- a/docs-chef-io/content/inspec/resources/parse_config.md
+++ b/docs-chef-io/content/inspec/resources/parse_config.md
@@ -55,7 +55,7 @@ where each test
 
 ## Options
 
-This resource supports the following options for parsing configuration data. Use them in an `options` block stated outside of (and immediately before) the actual test:
+This resource supports multiple options to parse configuration data. Use the options in an `options` block stated outside of (and immediately before) the actual test. For example:
 
     options = {
         assignment_regex: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
@@ -129,7 +129,7 @@ Use `standalone_comments: false`, to parse the following:
 
 ## Examples
 
-This resource is based on the `parse_config_file` resource. As such please refer to the examples of the [`parse_config_file`](/inspec/resources/parse_config_file) resource.
+This resource is based on the `parse_config_file` resource. See the [`parse_config_file`](/inspec/resources/parse_config_file) resource for examples.
 
 ## Matchers
 
--- a/docs-chef-io/content/inspec/resources/parse_config_file.md
+++ b/docs-chef-io/content/inspec/resources/parse_config_file.md
@@ -50,6 +50,13 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
+### Option names containing dots
+
+A possibly unexpected behaviour may occur when attempting to access option names containing dots with `its()`. A known behaviour, there are two ways to work around it:
+
+* Access the option by using the `params` attribute of the returned resource object
+* Since 4.23.9, `its` can be used by passing the option name in a single-element array (cf. examples)
+
 ## Options
 
 This resource supports the following options for parsing configuration data. Use them in an `options` block stated outside of (and immediately before) the actual test:
@@ -82,6 +89,12 @@ The following examples show how to use this Chef InSpec audit resource.
 
     describe parse_config_file('/path/to/yum.conf') do
      its('main') { should include('gpgcheck' => '1') }
+    end
+
+### Test a configuration setting containing dots
+
+    describe parse_config_file('/etc/sysctl.conf') do
+     its(['kernel.domainname']) { should eq 'example.com' }
     end
 
 ## Matchers
--- a/docs-chef-io/content/inspec/resources/parse_config_file.md
+++ b/docs-chef-io/content/inspec/resources/parse_config_file.md
@@ -69,38 +69,6 @@ This resource supports the following options for parsing configuration data. Use
       its('setting') { should eq 1 }
     end
 
-## Examples
-
-The following examples show how to use this Chef InSpec audit resource.
-
-### Test a configuration setting
-
-    describe parse_config_file('/path/to/file.conf') do
-     its('PARAM_X') { should eq 'Y' }
-    end
-
-### Use options, and then test a configuration setting
-
-    describe parse_config_file('/path/to/file.conf', { multiple_values: true }) do
-     its('PARAM_X') { should include 'Y' }
-    end
-
-### Test a file with an ini-like structure (such as a yum.conf)
-
-    describe parse_config_file('/path/to/yum.conf') do
-     its('main') { should include('gpgcheck' => '1') }
-    end
-
-### Test a configuration setting containing dots
-
-    describe parse_config_file('/etc/sysctl.conf') do
-     its(['kernel.domainname']) { should eq 'example.com' }
-    end
-
-## Matchers
-
-For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
-
 ### assignment_regex
 
 Use `assignment_regex` to test a key value using a regular expression:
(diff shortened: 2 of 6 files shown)


test/unit/resources/parse_config_test.rb
```diff
@@ -24,4 +24,13 @@
     _(resource.params).must_equal result
     _(resource.send("kernel.domainname")).must_equal "example.com"
   end
+
+  it "parse_config resource accepts arrays" do
```
james-stocks: Maybe we should mention the rspec `its` case here. In future, we might read this test and wonder why we needed to support this case.

rgeissert: I modified it to say `parse_config resource accepts arrays due to rspec's its behavior`, does that look good enough to you?

james-stocks: Brilliant, thank you!

james-stocks (approved):
This is great, thank you @rgeissert !
I have added our docs team as reviewer.

docs-chef-io/content/inspec/resources/parse_config.md
```diff
@@ -53,9 +53,20 @@ where each test
...
 
-For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
+This resource supports the following options for parsing configuration data. Use them in an `options` block stated outside of (and immediately before) the actual test:
+
+    options = {
+        assignment_regex: /^\s*([^:]*?)\s*:\s*(.*?)\s*$/,
+        multiple_values: true
+      }
+
+    output = command('some-command').stdout
```
james-stocks, line 65: We should use this new example code; but bear in mind that we increasingly see issues with using Ruby variable assignment like this, because the code executes unconditionally even if the control is skipped/waived/excluded.

The only current solution is to write a custom resource that wraps up the code. In future we want InSpec to to properly skip code for excluded controls...

lib/inspec/resources/parse_config.rb
```diff
@@ -55,8 +55,11 @@ def initialize(content = nil, useropts = nil)
       read_params unless @content.nil?
     end
 
-    def method_missing(name)
-      read_params[name.to_s]
+    def method_missing(*name)
+      # catch bahavior of rspec its implementation
+      # @see https://github.com/rspec/rspec-its/blob/master/lib/rspec/its.rb#L110
```
clintoncwolfe: Unfortunately, this URL is already bad - consider replacing with a permalink

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+### Test a file with an ini-like structure (such as a yum.conf)
+
+    describe parse_config_file('/path/to/yum.conf') do
+     its('main') { should include('gpgcheck' => '1') }
+    end
+
+### Test a configuration setting containing dots
+
+    describe parse_config_file('/etc/sysctl.conf') do
+     its(['kernel.domainname']) { should eq 'example.com' }
```
clintoncwolfe, line 161: This is just opinion.

This just feels strange - I think this may be a case of implementation driving UX. If I were a user, and I wanted to do a query containing a dot, even assuming I knew it had to be escaped, the typical thing would be to use a slash, not to enclose the query in brackets.

I agree the need is valid, and here we wish we could hijack `rspec-its` dot-processing facility entirely. So why not do so? In method_missing, why not explicitly search for dots and transform the call as needed to fool `rspec-its`? Then there may be no need to alter the DSL syntax at all - you may be able to have:

```
its('kernel.domainname') { should eq 'example.com' }
```

simply work.

rgeissert: From my understanding of rspec-its' code our method_missing won't ever see the dot due to the code at:
https://github.com/rspec/rspec-its/blob/v1.2.0/lib/rspec/its.rb#L113

I drafted some code to return some magic object from method_missing so that the calls to `send` Just Work.
It fails, however, whenever the hash contains "foo.bar" and "foo.bar.baz" and one wants to get the latter.

If that sounds acceptable I can use that instead of accepting arrays (like other resources do already).

I don't know how it should behave with regards to_sth methods. E.g. only "foo.bar.baz" exists, but the caller wants 'foo.bar' only.
On the plus side, `its('foo.bar.baz.downcase')` just works.

clintoncwolfe: Ah, thank you, that code reference cleared up the situation for me. I think your existing approach using the array syntax is an acceptable compromise.

lib/inspec/resources/parse_config.rb
rgeissert: Fixed by making it point to tag v1.2.0, the version of rspec-its currently in use.

docs-chef-io/content/inspec/resources/parse_config_file.md
rgeissert: If it's okay for you to leave it as-is then I think this pull request is ready from my POV.

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
     'key = value # comment'
     params['key'] = 'value'
+
+## Matchers
+
+For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
+
+## Examples
+
+The following examples show how to use this Chef InSpec audit resource.
```
mjingle, line 136:
```suggestion
The following examples show how to use this Chef InSpec audit resource:
```

docs-chef-io/content/inspec/resources/parse_config.md
```diff
@@ -114,3 +125,12 @@ Use `standalone_comments: false`, to parse the following:
 
     'key = value # comment'
     params['key'] = 'value'
+
+
+## Examples
+
+This resource is based on the `parse_config_file` resource. As such please refer to the examples of the [`parse_config_file`](/inspec/resources/parse_config_file) resource.
```
mjingle:
```suggestion
This resource is based on the `parse_config_file` resource. See the [`parse_config_file`](/inspec/resources/parse_config_file) resource for examples.
```

docs-chef-io/content/inspec/resources/parse_config.md
```diff
@@ -53,9 +53,20 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
-## Matchers
+## Options
 
-For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
+This resource supports the following options for parsing configuration data. Use them in an `options` block stated outside of (and immediately before) the actual test:
```
mjingle:
```suggestion
This resource supports multiple options to parse configuration data. Use the options in an `options` block stated outside of (and immediately before) the actual test. For example:
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+     its('PARAM_X') { should include 'Y' }
+    end
+
+### Test a file with an ini-like structure (such as a yum.conf)
+
+    describe parse_config_file('/path/to/yum.conf') do
+     its('main') { should include('gpgcheck' => '1') }
+    end
+
+### Test a configuration setting containing dots
```
mjingle: @rgeissert When referring to "dots," are you talking about periods (`.`) within a URL, a specific technical file, or something else?

rgeissert: @mjingle It should read: "Test a configuration setting whose name contains dots" (or periods, or full stops, as you prefer :)). As in: `author.name = Raphael` in a configuration file.

mjingle: @rgeissert Awesome! Thank you for confirming. That will help inform my suggested edits to you, and then we'll get this docs part finished.

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -50,6 +50,13 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
+### Option names containing dots
```
mjingle:
```suggestion
### Option Names Containing Periods
```

Website formatting prefers to have the first letter of each word in a header capitalized.

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -50,6 +50,13 @@ where each test
...
+
+A possibly unexpected behaviour may occur when attempting to access option names containing dots with `its()`. A known behaviour, there are two ways to work around it:
```
mjingle:
```suggestion
A possible behavior may occur when attempting to access option names containing periods with `its()`. There are two ways to work around it:
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+
+## Matchers
+
+For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
+
+## Examples
+
+The following examples show how to use this Chef InSpec audit resource.
+
+### Test a configuration setting
```
mjingle:
```suggestion
### Test A Configuration Setting
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+
+    describe parse_config_file('/path/to/file.conf') do
+     its('PARAM_X') { should eq 'Y' }
+    end
+
+### Use options, and then test a configuration setting
```
mjingle:
```suggestion
### Use Options And Then Test A Configuration Setting
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+
+    describe parse_config_file('/path/to/file.conf', { multiple_values: true }) do
+     its('PARAM_X') { should include 'Y' }
+    end
+
+### Test a file with an ini-like structure (such as a yum.conf)
```
mjingle:
```suggestion
### Test A File With An INI File Structure
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -145,3 +126,35 @@ Use `standalone_comments: false`, to parse the following:
...
+
```
mjingle, line 153:
```suggestion
`yum.conf` is one example of an INI file structure type. 

```

docs-chef-io/content/inspec/resources/parse_config_file.md
mjingle:
```suggestion
### Test A Configuration Setting Containing Periods
```

docs-chef-io/content/inspec/resources/parse_config_file.md
```diff
@@ -50,6 +50,13 @@ where each test
 - May run a command to `stdout`, and then run the test against that output
 - May use options to define how configuration data is to be parsed
 
+### Option names containing dots
+
+A possibly unexpected behaviour may occur when attempting to access option names containing dots with `its()`. A known behaviour, there are two ways to work around it:
+
+* Access the option by using the `params` attribute of the returned resource object
+* Since 4.23.9, `its` can be used by passing the option name in a single-element array (cf. examples)
```
mjingle:
```suggestion
* Since 4.23.9, `its` can be used by passing the option name in a single-element array. See the `parse config file` examples.
```

mjingle: After this PR merges, I will insert an internal reference link to the `parse config file` Examples section as a part of a link clean-up project in the docs. So know that this is planned and won't be omitted in the long term.

mjingle (commented):
Okay, @rgeissert! I've suggested all of my suggested changes in the docs portion. After these are fixed, you should be good from a docs perspective. Thank you for your patience! :)

docs-chef-io/content/inspec/resources/parse_config_file.md
rgeissert: I bumped the version number but it should be fixed once the PR is actually merged.

mjingle (approved):
Good to go from a docs perspective. Thank you again!

clintoncwolfe (approved):
Thanks!

PR #5263, opened 2020-09-25
Add option for winrm-shell-type

catriona1:
Signed-off-by: Catriona Malone <cmalone@chef.io>

Adding option to use winrm-shell-type, in conjunction with https://github.com/inspec/train-winrm/pull/25

## Types of changes

- [ x] New content (non-breaking change)

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/docs-chef-io/content/inspec/cli.md
+++ b/docs-chef-io/content/inspec/cli.md
@@ -144,7 +144,9 @@ This subcommand has additional options:
     Whether to use disable sspi authentication, defaults to false (WinRM).
 * ``--winrm-transport=WINRM_TRANSPORT``
     Specify which transport to use, defaults to negotiate (WinRM).
-
+* ``--winrm-shell-type=WINRM_SHELL_TYPE``
+    Specify which shell type to use (powershell,elevated or cmd), defaults to powershell (WinRM).
+    
 ## env
 
 Output shell-appropriate completion configuration
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -118,6 +118,8 @@ module Inspec
         desc: "Disable SSL verification on select targets"
       option :target_id, type: :string,
         desc: "Provide a ID which will be included on reports"
+      option :winrm_shell_type, type: :string, default: "powershell",
+        desc: "Specify a shell type for winrm (eg. 'elevated' or 'powershell')"
     end
 
     def self.profile_options


lib/inspec/base_cli.rb
```diff
@@ -118,6 +118,8 @@ def self.target_options # rubocop:disable Metrics/MethodLength
         desc: "Disable SSL verification on select targets"
       option :target_id, type: :string,
         desc: "Provide a ID which will be included on reports"
+      option :winrm_shell_type, type: :string,
```
james-stocks: Is it OK not to have a default? Looking at https://github.com/inspec/inspec/blob/master/lib/inspec/config.rb#L144 I think the default case is that this option won't get passed as a config key through to train?

james-stocks:
@catriona1 this looks good to me. That's cool that you can just declare a CLI option and InSpec figures out what to pass to train!

I think this should also be added to https://github.com/inspec/inspec/blob/master/docs-chef-io/content/inspec/cli.md

It would be nice to see tests, but I don't see a precedent in the `test` folder for testing winrm transport. Can you at least confirm you've locally tested alternate cases like not passing the option at all, not passing a value for the option, or passing an unexpected value?

lib/inspec/base_cli.rb
catriona1: I've updated this now to add "powershell" by default

catriona1:
> @catriona1 this looks good to me. That's cool that you can just declare a CLI option and InSpec figures out what to pass to train!
> 
> I think this should also be added to https://github.com/inspec/inspec/blob/master/docs-chef-io/content/inspec/cli.md
> 
> It would be nice to see tests, but I don't see a precedent in the `test` folder for testing winrm transport. Can you at least confirm you've locally tested alternate cases like not passing the option at all, not passing a value for the option, or passing an unexpected value?

Updated cli.md in latest commit 👍 

I've added a PR here https://github.com/inspec/train-winrm/pull/28 to validate any input to winrm-shell-type and confirm it's a supported shell. These tests all pass now.

james-stocks (approved):

Schwad:
I think this should be ready to roll out - for sanity purposes should we promote `train-winrm` first so that this build will capture it? (If I've got the chain right... hopefully it wouldn't need to go train-winrm->train->inspec! 😅 )

james-stocks:
Good point @Schwad - train-winrm should get a gem release first.

catriona1:
> I think this should be ready to roll out - for sanity purposes should we promote `train-winrm` first so that this build will capture it? (If I've got the chain right... hopefully it wouldn't need to go train-winrm->train->inspec! 😅 )

There's a PR to add validation to the new option waiting to be merged: https://github.com/inspec/train-winrm/pull/28

clintoncwolfe (approved):

lib/inspec/base_cli.rb
```diff
@@ -118,6 +118,8 @@ def self.target_options # rubocop:disable Metrics/MethodLength
         desc: "Disable SSL verification on select targets"
       option :target_id, type: :string,
         desc: "Provide a ID which will be included on reports"
+      option :winrm_shell_type, type: :string, default: "powershell",
+        desc: "Specify a shell type for winrm (eg. 'elevated' or 'powershell')"
```
mjingle, line 122: @catriona1 Quick description to fix (in the suggestion below). We're trying not to add more Latin in the docs and descriptions.
```suggestion
        desc: "Specify a shell type for winrm, such as "elevated" or "cmd""
```

docs-chef-io/content/inspec/cli.md
```diff
@@ -144,7 +144,9 @@ This subcommand has additional options:
     Whether to use disable sspi authentication, defaults to false (WinRM).
 * ``--winrm-transport=WINRM_TRANSPORT``
     Specify which transport to use, defaults to negotiate (WinRM).
-
+* ``--winrm-shell-type=WINRM_SHELL_TYPE``
+    Specify which shell type to use (powershell,elevated or cmd), defaults to powershell (WinRM).
```
mjingle, line 148: @catriona1 Hi! Thank you for this work! Because this is a PR in a remote repo, I can't directly submit fixes to this PR. A fix to make:

```suggestion
    Specify which shell type to use: `powershell`, `elevated`, or `cmd`. Defaults to `powershell` (WinRM).
```

PR #5266, opened 2020-09-29
Support 'clear_cache'

Schwad:
Helps with Customer Bug 236 (see there for repro)

Currently, when we run `inspec compliance upload my_profile` it is
cached locally in inspec when run. If we update the version in the core
code and run another upload, `inspec compliance upload my_profile` again
it will run the old cached version instead of running a new copy
from automate.

The current workaround is to specify the desired version with
`inspec exec compliance://my_profile/admin#0.1.1`.

The caching happens before we have forward sight into the profile's
contents and only the target name. So the text used to generate the
cache would be `compliance://my_profile/admin` which does not change
version to version.

A fix here could simply identify when we are doing a local `inspec exec
compliance://` (hitting local profiles does not generate a cache) and
skips the cache if there's no version specified. That would eliminate
the unexpected behavior. However, it is a breaking change for customers
as some current caching taking place would no longer take place.

Instead, we have included a `clear_cache` cli method for InSpec,
which should assist the core team and other developers in the future
when debugging edge case issues in InSpec.

< edited 2020-11-24 >

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -395,6 +395,20 @@ class Inspec::InspecCLI < Inspec::BaseCLI
   end
   map %w{-v --version} => :version
 
+  desc "clear_cache", "clears the InSpec cache. Useful for debugging."
+  option :vendor_cache, type: :string,
+    desc: "Use the given path for caching dependencies. (default: ~/.inspec/cache)"
+  def clear_cache
+    o = config
+    configure_logger(o)
+    cache_path = o[:vendor_cache] || "~/.inspec/cache"
+    FileUtils.rm_r Dir.glob(File.expand_path(cache_path))
+
+    o[:logger] = Logger.new($stdout)
+    o[:logger].level = get_log_level(o[:log_level])
+    o[:logger].info "== InSpec cache cleared successfully =="
+  end
+
   private
 
   def run_command(opts)
--- a/test/functional/inspec_clear_cache_test.rb
+++ b/test/functional/inspec_clear_cache_test.rb
@@ -1,0 +1,27 @@
+require "functional/helper"
+require "securerandom"
+
+describe "inspec check" do
+  include FunctionalHelper
+
+  parallelize_me!
+
+  describe "inspec clear_cache" do
+    it "clears any existing cache" do
+      dirname = File.expand_path("~/.inspec/cache")
+      unless File.directory?(dirname)
+        FileUtils.mkdir_p(dirname)
+      end
+      newfile = "#{dirname}/#{SecureRandom.hex(10)}.txt"
+      File.write(newfile, SecureRandom.hex(100))
+
+      assert !Dir.glob(newfile).empty?
+
+      out = inspec("clear_cache")
+
+      assert_empty Dir.glob(newfile)
+      assert_exit_code 0, out
+      _(out.stdout).must_include "== InSpec cache cleared successfully ==\n"
+    end
+  end
+end
--- a/test/functional/inspec_clear_cache_test.rb
+++ b/test/functional/inspec_clear_cache_test.rb
@@ -8,7 +8,7 @@ describe "inspec check" do
 
   describe "inspec clear_cache" do
     it "clears any existing cache" do
-      dirname = File.expand_path("~/.inspec/cache")
+      dirname = File.expand_path("~/.inspec/#{SecureRandom.hex(10)}/alt-cache")
       unless File.directory?(dirname)
         FileUtils.mkdir_p(dirname)
       end
@@ -17,7 +17,7 @@ describe "inspec check" do
 
       assert !Dir.glob(newfile).empty?
 
-      out = inspec("clear_cache")
+      out = inspec("clear_cache --vendor-cache=#{dirname}")
 
       assert_empty Dir.glob(newfile)
       assert_exit_code 0, out
--- a/lib/inspec/runner_rspec.rb
+++ b/lib/inspec/runner_rspec.rb
@@ -5,7 +5,7 @@ require "matchers/matchers"
 require "inspec/rspec_extensions"
 
 # There be dragons!! Or borgs, or something...
-# This file and all its contents cannot be unit-tested. both test-suits
+# This file and all its contents cannot be unit-tested. both test-suites
 # collide and disable all unit tests that have been added.
 
 module Inspec


Schwad:
Checking CI builds here. Still need to implement tests for cache buster and ensuring our case is covered with the caching.

lib/inspec/cached_fetcher.rb
```diff
@@ -65,5 +68,18 @@ def assert_cache_sanity!
...
     end
+
+    private
+
+    ##
+    # This resolves a bug on compliance whereby uploading new profiles to
+    # automate (and not specifying a version number when running them) causes
+    # old, cached profiles to run. This removes caching when passing a generic
+    # profile name as a target.
+    def compliance_versioned
```
clintoncwolfe: Perhaps this method should end in a question mark, as it is a predicate...

lib/inspec/cached_fetcher.rb
```diff
@@ -65,5 +68,18 @@ def assert_cache_sanity!
...
+      return true if target.match(/^compliance:\/\/.*#(\d|\.)+$/)
+      return true if !target.include?("compliance")
```
clintoncwolfe: This second condition doesn't seem to make sense to me, or at least does not match the name of the method. Perhaps it should be renamed `should_use_cache?`

lib/inspec/cached_fetcher.rb
```diff
@@ -65,5 +68,20 @@ def assert_cache_sanity!
...
       raise exception_message if fetcher.resolved_source[:sha256] != target[:sha256]
     end
+
+    private
+
+    ##
+    # This resolves a bug on compliance whereby uploading new profiles to
+    # automate (and not specifying a version number when running them) causes
+    # old, cached profiles to run. This removes caching when passing a generic
+    # profile name as a target.
```
james-stocks: It might be easier to read a comment that's set in the present tense and describes current functionality, instead of needing people reading the code to have to think about bugs that existed in the past.
This comment might just need to describe that sometimes a cache will not be used.

lib/inspec/cli.rb
```diff
@@ -393,6 +393,13 @@ def version
   end
   map %w{-v --version} => :version
 
+  desc "clear_cache", "clears the InSpec cache at ~/.inspec/cache. Useful for\
+  debugging."
+  def clear_cache
+    FileUtils.rm_r Dir.glob(File.expand_path("~/.inspec/cache/*"))
```
james-stocks: This needs to work for Windows as well. 
I half remember the InSpec work dir (`~/.inspec`) being configurable to other locations as well.

lib/inspec/cli.rb
```diff
@@ -393,6 +393,13 @@ def version
...
+    puts "== InSpec cache cleared successfully =="
```
james-stocks: Can the logger be used here? Or at least puts to stderr to keep the convention that stdout is for result data.

test/unit/cached_fetcher_test.rb
```diff
@@ -25,8 +25,22 @@
          "inputs" => [],
          "latest_version" => "" }]
     end
+
     before do
       InspecPlugins::Compliance::Configuration.expects(:new).returns({ "token" => "123abc", "server" => "https://a2.instance.com" })
+
+      @stub_get =
+        stub_request(
```
james-stocks: Nice :)

james-stocks:
Big 👍 from me on making this change; but I think we should run it past @kekaichinose - some existing customers who are currently happy to run with a cached profile every time might perceive this change as a reduction in performance and increase in traffic. We'd just need to be ready to explain the improvement here.

lib/inspec/cli.rb
Schwad: I see we do configure cache on `--vendor-cache` option so added support for that as well. If there's anything else I'm missing let me know as that's all I could see! (Vendor cache always gives us the default unless something else is supplied. If you think I should investigate a way to support the functionality of vendor_cache but in a more generic name I can look into that too 👍 )

I believe this passes on CI on windows, but on a cursory look it appears that FileUtils.rm_rf can be risky in relation to symlinking on windows machines? My windows-foo is not great, and I might need some guidance on detecting OS
from the `cli.rb` level- but would we be looking at something sort of like this?

```ruby
if somehow_determine_windows?
  system "rmdir /s/q #{o[:vendor_cache]}"
else
  FileUtils.rm_rf File.expand_path(o[:vendor_cache])
end
```

Schwad:
This is currently blocked from the product side due to changing the caching experience for current users whilst solving the problem. I am currently looking into alternative ways to implement the fix without changing the experience. If I fail, then we'll have to simply ship the `clear_cache` as the option for users to debug this.

Schwad:
Updated the description to match commit description. It appears there's not a quick and easy way to back door the caching, and the first solution has not been accepted. Therefore I have let those that raised the issue know about the situation and have updated this PR just to implement `clear_cache` which will at least assist folks with rectifying this in the future.

test/functional/inspec_clear_cache_test.rb
```diff
@@ -0,0 +1,27 @@
...
+require "securerandom"
+
+describe "inspec check" do
+  include FunctionalHelper
+
+  parallelize_me!
+
+  describe "inspec clear_cache" do
+    it "clears any existing cache" do
+      dirname = File.expand_path("~/.inspec/cache")
```
clintoncwolfe: Parallelizing and using a global filesystem location during testing seems like asking for trouble. Any other test that caches may be writing to the cache at the same time, and may have it cleared during its testing.

Since the implementation supports using a `--vendor-cache` option for a custom cache location, it might be safer to use a temporary directory in a block and pass that as the cache location.

Schwad: Sorry I only just responded here - thanks for this point and I've updated the PR to hopefully catch this via the `vendor-cache` path. :)

clintoncwolfe (approved):

Schwad:
Ahoy! 👋 just dropping in to say I've had a look at this and it seems it's red because of two failures in the tests (likely something syntax-y). 

I'm hopeful I'll get a change to pull this up on my machine and get things green on this PR! 👍 

I can't see artifact and hab-build (rightfully so as they're blacklisted from community viewing) but I'm assuming those reds might be something a bit flaky or unrelated to the PR. 🤔

PR #5269, opened 2020-09-29
Added another example to illustrate use for only_if

unixengineer:
## Description
Added another example to illustrate use for only_if

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/dsl_inspec.md
+++ b/docs-chef-io/content/inspec/dsl_inspec.md
@@ -249,6 +249,21 @@ control 'nutcracker-connect-redis-001' do
 end
 ```
 
+Here is another simple example, which checks for, if certain pip packages are installed, only if /root/.aws exists
+```ruby
+control 'pip-packages-installed' do
+  title 'Check if essential pips are installed'
+  only_if('aws-cli config not created.') do
+    directory('/root/.aws').exist?
+  end
+  %w(aws-mfa PyYAML awscli).each do |aws_pip_deps|
+    describe pip(aws_pip_deps) do
+      it { should be_installed }
+    end
+  end
+end
+```
+
 Mixing this with other conditionals, such as checking existence of the files, can
 help to test different test paths using Chef InSpec. With this way, you can skip
 certain controls, which would 100% fail due to the way servers are prepared, but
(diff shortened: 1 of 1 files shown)


docs-chef-io/content/inspec/dsl_inspec.md
```diff
@@ -249,6 +249,21 @@ control 'nutcracker-connect-redis-001' do
 end
 ```
 
+Here is another simple example, which checks for, if certain pip packages are installed, only if /root/.aws exists
```
james-stocks: This is a valid example of `only_if` - I wanted to confirm, is there something about this case that technically differs from the previous example? It looks like (and your description for the example says) this is simply one more example, but I wanted to make sure. If there's something that doesn't work in the previous example that this example solves, we should highlight that.

james-stocks:
Thanks very much for this @unixengineer !

I have added our docs team as a reviewer.

Please note that this PR shows the [`DCO` check](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) is failing. If you're not familiar with that, it means that all commits must be signed by the contributor. And that means the last line of your commit message should be a sign-off string like

` Signed-off-by: Your Name <your_email_address@example.com>`

(git generates that for you if you use the `-s` option with `git commit`)

docs-chef-io/content/inspec/dsl_inspec.md
mjingle: @unixengineer Hello! Please update your PR to include the following wording. Thank you for adding this example!

```suggestion
This example checks for if certain pip packages are installed, but only if '/root/.aws' exists:
```

unixengineer: @james-stocks this additional example, I felt, explains better (better because, at first glance, the example was not explanatory enough to understand how it worked), and also provides an example to check directory exists as a condition.

unixengineer: Will update the PR and add a signed by too

unixengineer: The PR should have the updated changes now.

mjingle (approved):
Thank you @unixengineer! Looks good.

clintoncwolfe (approved):

PR #5273, opened 2020-10-01
Add command-timeout option

james-stocks:
TODO: 

- [ ] Add docs
- [ ] test one of our Automate profiles

Allow commands to have a timeout, set in minutes. This is considered an emergency mechanism to stop CI being halted indefinitely. If a test inherently needs to pass or fail based on a timeout, this should be scripted into the command string instead.

It can be set for a command resource, like:

```
  describe command('sleep 100', timeout: 1) do
  end
```

It can also be set as a new option `command-timeout`.
A timeout set on a `command` resource takes precedence over the `command-timeout` option.

If no timeout is set at all, a command will default to timing out after 1 hour

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -321,6 +321,9 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: "A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell"
   option :distinct_exit, type: :boolean, default: true,
     desc: "Exit with code 100 if any tests fail, and 101 if any are skipped but none failed (default).  If disabled, exit 0 on skips and 1 for failures."
+  option :command_timeout, type: :numeric, default: 60,
+      desc: "Maximum minutes to allow a command to run. Default 60.",
+      long_desc: "Maximum minutes to allow commands to run. Default 60. A timed out command is considered an error."
   option :inspect, type: :boolean, default: false, desc: "Use verbose/debugging output for resources."
   def shell_func
     o = config
--- a/lib/inspec/resources/command.rb
+++ b/lib/inspec/resources/command.rb
@@ -32,6 +32,15 @@ module Inspec::Resources
 
       @command = cmd
 
+      @timeout = options[:timeout]&.to_i || Inspec::Config.cached.final_options['command_timeout']&.to_i
+      if @timeout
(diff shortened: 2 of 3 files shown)


lib/inspec/cli.rb
```diff
@@ -321,6 +321,9 @@ def detect
     desc: "A space-delimited list of local folders containing profiles whose libraries and resources will be loaded into the new shell"
   option :distinct_exit, type: :boolean, default: true,
     desc: "Exit with code 100 if any tests fail, and 101 if any are skipped but none failed (default).  If disabled, exit 0 on skips and 1 for failures."
+  option :command_timeout, type: :numeric, default: 60,
```
james-stocks, line 324: I don't think this is needed. The important thing for shell is that users can paste the code that they would be running in `exec`. 
I'll remove this and test that a timeout set at the `command` resource still works in shell.

lib/inspec/resources/command.rb
```diff
@@ -32,6 +32,15 @@ def initialize(cmd, options = {})
 
       @command = cmd
 
+      @timeout = options[:timeout]&.to_i || Inspec::Config.cached.final_options['command_timeout']&.to_i
```
james-stocks, line 35: @clintoncwolfe this is the first time a resource has reached into `Inspec::Config`to look at settings. Any reason not to do this, or any other pattern I could use? Or does this seem reasonable?

clintoncwolfe: I don't see any issue with accessing Config from a Resource.

james-stocks:
This needs further testing and work - using an Ubuntu test VM I'm seeing a timed out command frequently put the connection in a bad state where all subsequent commands (i.e. all other tests in the profile) throw errors.

I've found this can be avoided by putting a sleep like so: 

```
    def result
      @result ||= inspec.backend.run_command(@command)
      @result ||= begin 
        inspec.backend.run_command(@command, timeout: @timeout)
      rescue Train::CommandTimeoutReached
        sleep 1
        raise Inspec::Exceptions::ResourceFailed,
              "Command `#{@command}` timed out after #{@timeout / 60} minute(s)"
      end
    end
```

However putting a sleep into InSpec isn't the wisest thing to do; I need to debug what's going wrong in `train` and fix it there.

PR #5280, opened 2020-10-15
Add http for windows 2

### LINKED ISSUE #3959: Add Windows support for the http resource
### 🎛 Description

**Choose one:** is this a 🐛 bug report or 🙋 feature request?

Feature request, though this feature request came from a Bug Report in #2822

Briefly describe the issue and the expected behavior. Also, please be aware of our [Code of Conduct](https://github.com/inspec/inspec/blob/master/CODE_OF_CONDUCT.md).

A user should be able to use the http resource on windows.

i.e. 

```
describe http('http://localhost:8080', enable_remote_worker: true) do
  its('body') { should match /Tomcat/ }
end
```

should run successfully on Windows, rather than returning an unsupported platform error.

Aha! Link: https://chef.aha.io/features/SH-2620

mbaitelman:
## Related Issue
Closes #3959

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-375


mbaitelman:
Using this to figure out the git part of the branching

test/unit/resources/http_test.rb
```diff
@@ -243,4 +243,23 @@
       _(headers.D).must_equal(4)
     end
   end
+
+  describe "Windows" do
+
+    # windows
+    it "simple HTTP request with no options" do
+      resource = MockLoader.new(:windows).load_resource("http", "https://www.example.com")
```
mbaitelman: The issue with the test is it seems to be defaulting to local while the code (currently) only works for remote. 
How can I change the test to use the correct backend?

test/unit/resources/http_test.rb
```diff
@@ -243,4 +243,19 @@
       _(headers.D).must_equal(4)
     end
   end
+
+  describe "Windows" do 
+    let(:backend)     { MockLoader.new.backend }
```
mbaitelman: The problem I'm running into now is passing this `inspec` object as being on Windows so that `inspec.os.windows?` is true

Schwad: Would the aim to be running the *same* test for HTTP - but running that test on a windows machine and ensuring it passes? Usually I'd run these tests actually from inside a windows machine to ensure they're working.

mbaitelman: From what I'm seeing it can be done in 2 ways
1) use the mock helper to match the generated command to a response
2) run this on windows (though it won't work yet as the local VS remote uses different classes. 
For now, I'd like to get remote working and then add local after.

clintoncwolfe: To get `inspec` to mock as windows, I modified line 248 to read:
```
    let(:backend)     { MockLoader.new(:windows).backend }
```
You'll also need to stub the detection of WebRequest, using something like this in the `it` block:
```
Inspec::Resources::Cmd.any_instance
          .stubs(:exist?)
          .returns(true)
```
After doing those two things, I still encountered a few minor errors regarding the assemble of the request_header_string, but those were easily corrected.

test/helpers/mock_loader.rb
```diff
@@ -519,6 +519,8 @@ def md.directory?
       "curl -i --head --connect-timeout 60 --max-time 120 'http://www.example.com'" => cmd.call("http-remote-head-request"),
       "curl -i -X OPTIONS --connect-timeout 60 --max-time 120 -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: origin, x-requested-with' -H 'Origin: http://www.example.com' 'http://www.example.com'" => cmd.call("http-remote-options-request"),
 
+      # http resource - windows
+      "Invoke-WebRequest -Method Get -TimeoutSec 120 https://www.example.com | Select-Object -Property * | ConvertTo-json" => cmd.call("http-windows-remote-no-options"),
```
clintoncwolfe: I found the invocation line needed to be adjusted to be:
```
Invoke-WebRequest -Method GET -TimeoutSec 120 'http://www.example.com'  | Select-Object -Property * | ConvertTo-json
```

test/fixtures/cmd/http-windows-remote-no-options
```diff
@@ -0,0 +1,519 @@
+{
```
clintoncwolfe, line 1: Sadly, this file appears to be invalid JSON. It looks like there is a mismatched curly bracket in the CSS.

n3mawashi:
Hi, Can I bump this PR? Having this functionality would be a great help for a project I'm working on.

PR #5281, opened 2020-10-16
Update Ruby to 2.7.2

tas50:
Now that Ruby 2.7.2 no longer throws deprecation warnings left and right
we should update InSpec to use the latest and greatest Ruby that we use
in Chef Infra Client and Workstation.

Aha! Link: https://chef.aha.io/features/SH-376

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -3,4 +3,4 @@
 # grab the current train release from rubygems.org
 train_stable = /^train \((.*)\)/.match(`gem list ^train$ --remote`)[1]
 override "train", version: "v#{train_stable}"
-override "ruby", version: "2.6.6"
+override "ruby", version: "2.7.2"


Schwad:
Worth noting for those coming here, the 2.7.1 -> 2.7.2 change __turns off deprecations__. They will still be there but not output unless you pass the `-w` flag. 

https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/

Schwad (approved):

james-stocks (approved):

btm:
Are we aware of any risk of Ruby changes between 2.6 and 2.7 causing issues for users profiles?

james-stocks:
The Ruby version for Habitat should also get bumped e.g. like https://github.com/inspec/inspec/pull/5287 bumps it to 2.6

kekaichinose:
@Schwad I wanted to bump @btm 's question. Are there any breaking implications for doing the upgrade?

tas50:
Not that I know of

Schwad:
@kekaichinose not that I'm aware of as well. If we were really paranoid I guess we could maybe have content (or someone) do a full run of a passing profile against this branch of inspec against a target to ensure it's green on both versions? 🤔

clintoncwolfe (approved):

kekaichinose:
@Schwad or @clintoncwolfe if there are no other objections, can we get this PR merged?

PR #5288, opened 2020-10-22
Add windows as a supported platform for ssh_config

rgeissert:
## Description
Openssh for windows is a thing, so allow the ssh_config and sshd_config
resources to be used on it.

## Related Issue
#4782

## Types of changes

## Checklist:

Aha! Link: https://chef.aha.io/features/SH-401

### DIFF
--- a/lib/inspec/resources/ssh_config.rb
+++ b/lib/inspec/resources/ssh_config.rb
@@ -20,7 +20,7 @@ module Inspec::Resources
     include FileReader
 
     def initialize(conf_path = nil, type = nil)
-      @conf_path = conf_path || "/etc/ssh/ssh_config"
+      @conf_path = conf_path || ssh_config_file("ssh_config")
       typename = (@conf_path.include?("sshd") ? "Server" : "Client")
       @type = type || "SSH #{typename} configuration #{conf_path}"
       read_content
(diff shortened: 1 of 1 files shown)


lib/inspec/resources/ssh_config.rb
```diff
@@ -7,6 +7,7 @@ module Inspec::Resources
   class SshConfig < Inspec.resource(1)
     name "ssh_config"
     supports platform: "unix"
+    supports platform: "windows"
```
james-stocks, line 10: This is a nice addition, thank you!

On lines 23 and 93 we refer to Posix filepaths. A quick read of Windows OpenSSH docs suggests the files will be in `%programdata%\ssh\`

It would be good to change

```
@conf_path = conf_path || "/etc/ssh/ssh_config"
```

to

```
@conf_path = conf_path || ssh_config_file("ssh_config")
```

where `ssh_config_file` is a new method that appends `%programdata%\ssh\` if Windows, else appends `/etc/ssh/`

rgeissert: `file` doesn't seem to handle `%foo%`, so I had to go add a few more lines to retrieve the variable.

rgeissert:
@james-stocks I believe the failed tests to be unrelated to this commit. Is there any way to re-run them?

PR #5311, opened 2020-11-23
Remove macOS 10.13 from the build matrix

tas50:
We support N-2 builds and now that Big Sur is out we can remove 10.13. I also cleaned up the supported platforms in the readme a bit.

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -20,8 +20,7 @@ builder-to-testers-map:
     - el-7-x86_64
     - el-8-x86_64
     - amazon-2-x86_64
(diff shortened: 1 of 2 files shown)


mjingle (commented):
@Schwad @clintoncwolfe @stevendanna I know we have a list of supported OS for Chef InSpec here: https://docs.chef.io/platforms/#chef-inspec . Do we need to make a distinction between the Platforms list and a different list of what Chef InSpec can scan but doesn't have packages for? (If so, can we consolidate this information somehow within the docs rather than the readme?)

clintoncwolfe (approved):

PR #5314, opened 2020-11-23
Return arrays for wmi properties

### LINKED ISSUE #5313: wmi resource can't handle select statements with multiple objects returned
```ruby
describe wmi({:namespace=>"root\\cimv2", :query=>"SELECT installstate FROM win32_optionalfeature"}) do
  its("installstate") { should include 2 }
end
```

The above example should work since an arbitrary WMI select statement can have 0 or multiple matches.

For the `installstate` property we only get a non-nil value when the WMI SELECT query has a single match. Behind the scene, this is the call sent by InSpec to the Windows target as well as the json output retrieved a parsed by the resource:

```shell
PS C:\> Get-WmiObject -namespace "root\cimv2" -Query "SELECT installstate FROM win32_optionalfeature WHERE name = 'SMB1Protocol'" | Aggregate | ConvertTo-Json
{
    "InstallState":  2
}
```

If the WMI SELECT query has more than one match, the json output retrieved from the target looks like this:

```shell
PS C:\> Get-WmiObject -namespace "root\cimv2" -Query "SELECT installstate FROM win32_optionalfeature" | Aggregate | ConvertTo-Json
[
    {
        "InstallState":  1
    },
    {
        "InstallState":  1
    },
    {
        "InstallState":  2
    }
]
```

The `params` function of the `wmi` resource is not expecting an array result back and all properties will receive nil values in describe tests.

## Possible Solution

We could achieve consistency in the data structure returned by avoiding the pipe to ConvertTo-Json like this:
```shell
PS C:\> ConvertTo-Json @(Get-WmiObject -namespace "root\cimv2" -Query "SELECT installstate FROM win32_optionalfeature WHERE name = 'SMB1Protocol'" | Aggregate)
[
    {
        "InstallState":  2
    }
]
PS C:\> ConvertTo-Json @(Get-WmiObject -namespace "root\cimv2" -Query "SELECT installstate FROM win32_optionalfeature" | Aggregate)
[
    {
        "InstallState":  1
    },
    {
        "InstallState":  1
    },
    {
        "InstallState":  2
    }
]
```

But then we would have to parse and convert this array into a hash to be compatible with the expectations of the **params** function. Best would be in this case for the `Aggregate` filter to return a map directly to InSpec.

alexpop:
## Related Issue
Fixes https://github.com/inspec/inspec/issues/5313

## Types of changes:

## Braking change:

Tests for the wmi resource that used the `eq` matcher without an array expected value, for example:
```ruby
describe wmi({:namespace=>"root\\cimv2", :query=>"SELECT installstate FROM win32_optionalfeature WHERE name = 'SMB1Protocol'"}) do
  its("installstate") { should eq 2 }
end
```

will need to upgrade to an expected array value, for example:
```ruby
  its("installstate") { should eq [2] }
```

or use the `cmp` matcher instead of `eq`
```ruby
  its("installstate") { should cmp 2 }
```

As a bonus, the resource is now compatible with a new `includes` matcher:
```ruby
  its("installstate") { should includes 2 }
```

## Checklist:

### DIFF
--- a/lib/inspec/resources/wmi.rb
+++ b/lib/inspec/resources/wmi.rb
@@ -16,7 +16,10 @@ module Inspec::Resources
         namespace: 'root\\rsop\\computer',
         filter: 'KeyName = \'MinimumPasswordAge\' And precedence=1'
       }) do
-         its('Setting') { should eq true }
+         its('Setting') { should cmp true }
+      end
+      describe wmi({namespace: "root\\cimv2", query: "SELECT installstate FROM win32_optionalfeature"}) do
+        its("installstate") { should include 2 }
       end
     EXAMPLE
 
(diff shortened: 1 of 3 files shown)


Schwad (approved):
This looks good to me, @alexpop. I see there's a spec failure we have. Might be worth having a chat with @kekaichinose to get his +1 on it with regards to the customer and possible breaking elements. thanks!

PR #5317, opened 2020-11-27
Fix issue Dir.home would break when HOME env absent

### LINKED ISSUE #4803: couldn't find login name -- expanding `~' (ArgumentError)
Inspec version - 4.18.51

I am trying to inspec target VM by ssh. 

Command used:
inspec exec /path/script.rb --host <host> --port <port> --backend ssh --user <username> --password <password> --reporter json

I am getting expected output when I execute this command with argument values from terminal(ubuntu). I am using python subprocess to call this and got error.

 /usr/local/lib/ruby/gems/2.6.0/gems/inspec-4.18.51/lib/inspec/globals.rb:3:in `home': couldn't find login name -- expanding `~' (ArgumentError)

How come the command is working in terminal and why it is not working in script?

Thanks

kvivek1115:
## Description

Fix the issue `Dir.home` would break when home directory lookup fail. when HOME even var is not set.
as `$HOME` is not set in systemd service `File.expand_path('~')` will not work here

## Related Issue

Fixes https://github.com/inspec/inspec/issues/4803, https://github.com/chef/chef/issues/10565

Steps to reproduce
- `unset HOME`
- `ruby -e 'print Dir.home'`

## Types of changes

## Checklist:

Signed-off-by: Vivek Singh <vivek.singh@msystechnologies.com>

### DIFF
--- a/lib/inspec/globals.rb
+++ b/lib/inspec/globals.rb
@@ -1,11 +1,20 @@
 require_relative "utils/install_context"
 
+require 'etc' unless defined?(Etc)
+
 module Inspec
 
   extend Inspec::InstallContextHelpers
-
+  
   def self.config_dir
-    ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(Dir.home, ".inspec")
+    begin
+  	  home = Dir.home
+  	rescue ArgumentError, NoMethodError
+  	  # $HOME is not set in systemd service File.expand_path('~') will not work here
+  	  home = Etc.getpwuid(Process.uid).dir
+  	end
+
+    ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(home, ".inspec")
   end
 
   def self.src_root
--- a/lib/inspec/globals.rb
+++ b/lib/inspec/globals.rb
@@ -14,7 +14,7 @@ module Inspec
   def self.home_path
     Dir.home
   rescue ArgumentError, NoMethodError
-    # $HOME is not set in systemd service File.expand_path('~') will not work here
+    # If ENV['HOME'] is not set, Dir.home will fail due to expanding the ~. Fallback to Etc.
     require "etc" unless defined?(Etc)
     Etc.getpwuid.dir
   end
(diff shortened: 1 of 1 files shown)


lib/inspec/globals.rb
```diff
@@ -1,11 +1,20 @@
...
   extend Inspec::InstallContextHelpers
-
+  
   def self.config_dir
-    ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(Dir.home, ".inspec")
+    begin
+  	  home = Dir.home
+  	rescue ArgumentError, NoMethodError
+  	  # $HOME is not set in systemd service File.expand_path('~') will not work here
+  	  home = Etc.getpwuid(Process.uid).dir
```
clintoncwolfe: It looks like you can simplify this to `Etc.getpwiud.dir`

lib/inspec/globals.rb
```diff
@@ -1,11 +1,20 @@
...
 
   extend Inspec::InstallContextHelpers
-
+  
   def self.config_dir
-    ENV["INSPEC_CONFIG_DIR"] ? ENV["INSPEC_CONFIG_DIR"] : File.join(Dir.home, ".inspec")
+    begin
+  	  home = Dir.home
+  	rescue ArgumentError, NoMethodError
+  	  # $HOME is not set in systemd service File.expand_path('~') will not work here
```
clintoncwolfe: Might want to make this comment more general, something like:
"If ENV['HOME'] is not set, Dir.home will fail due to expanding the ~. Fallback to Etc."

lib/inspec/globals.rb
```diff
@@ -1,11 +1,20 @@
 require_relative "utils/install_context"
 
+require 'etc' unless defined?(Etc)
```
clintoncwolfe: Could we move this into the rescue block, so we only require it if needed?

clintoncwolfe (commented):
Hi! I think the fallback to Etc is a good approach. A few small comments to address but nothing major.

In addition:
 * Please run the linter on your code - run `bundle exec chefstyle -a` to fix the code style issues.
 * Please rebase your code against master to pick up fixes for the Ruby 2.5 tests.

Thanks!

lib/inspec/globals.rb
kvivek1115: Make sense to me thanks!

clintoncwolfe (approved):

PR #5323, opened 2020-12-01
Fix escaping of Windows package names

### LINKED ISSUE #5235: Incorrect escaping of Windows package names

ramereth:
This resolves #5235 and supersedes #5238.

Instead of using `-match` use `-like` when querying the Windows registry. This completely removes the problem of trying to clean the string for a regex and just let's Powershell do that with `-like`.

I confirmed this works with the use case of having a package name such as `Microsoft Visual C++ 2015-2019 Redistributable (x86) - 14.28.29325` work as-is without any modification.

**NOTE: Someone with more PowerShell experience should let me know if this impacts backward compatibilty issues with older versions of Windows.**

### DIFF
--- a/lib/inspec/resources/package.rb
+++ b/lib/inspec/resources/package.rb
@@ -314,7 +314,7 @@ module Inspec::Resources
       # Find the package
       cmd = inspec.command <<-EOF.gsub(/^\s*/, "")
         Get-ItemProperty (@("#{search_paths.join('", "')}") | Where-Object { Test-Path $_ }) |
-        Where-Object { $_.DisplayName -match "^\s*#{package_name.shellescape}\.*" -or $_.PSChildName -match "^\s*#{package_name.shellescape}\.*" } |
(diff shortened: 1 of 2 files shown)


tas50:
@mwrock If you could take a look at this one I'm sure the Windows folks out there would appreciate it.

mwrock (approved):

clintoncwolfe (approved):

PR #5324, opened 2020-12-01
Remove Gemfile.local support to use Dependabot

tas50:
If you remove this then Dependabot can manage deps, which helps keep us
up to data without manual work.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,3 @@ end
 group :deploy do
   gem "inquirer"
 end
-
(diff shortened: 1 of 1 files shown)


Schwad (approved):
This makes sense to me but would also like @clintoncwolfe 's two cents in case I'm missing anything :)

clintoncwolfe (approved):
Yes please, this is how we run most other projects. It's a big win. We'll see a lot of simple automatic PRs.

PR #5325, opened 2020-12-01
Remove the dependency on json_schemer

tas50:
It looks like we're only using this in 3 test files and it conflicts with new RuboCop releases, which is blocking us from shipping that new release in Chef Infra Client and Workstation


Schwad (approved):
Thanks!

Issue #5225, opened 2020-08-25
Skip controls from a wrapper profile


szamfirov:
# Introduction
We are relying on InSpec as one of the tools to help us achieve a good security and compliance posture when leveraging Google Cloud Platform's services. We’ve been using InSpec with a “wrapper profile” which currently includes only the controls from the official [GCP CIS Benchmark 1.1 profile](https://github.com/GoogleCloudPlatform/inspec-gcp-cis-benchmark). Using the wrapper profile helps us manage which controls to use, what versions of the profile’s dependencies (in this case the GCP CIS Benchmark profile) to use etc. and maintain all of this in a single git repository which is being used across multiple pipelines managed by different teams.

# The problem
The wrapper profile is excluding some of the controls from the official GCP CIS Benchmarks (referred to as `inspec-gcp-cis-benchmarks`) based on how we use GCP's services. Another team would like to additionally skip a few controls extra and we’d noticed that the **“skip_control” functionality is not working** correctly. The diagram below shows how we're using this profile and what we’re trying to achieve in a nutshell:

![InspecWrapperProfile1 (1)](https://user-images.githubusercontent.com/4711683/91165167-86f4d600-e6d0-11ea-92aa-de04c7c33647.png)

Our workaround for this was to include the controls from the ‘inspec-gcp-cis-benchmarks’ in our ‘my-project’ profile (even though it’s not a direct dependency) in order to define which controls to additionally skip but that resulted in the controls **being executed twice**. This duplication of the controls being executed/evaluated results in a burst of API calls towards GCP which quite often gets rate limited by GCP's project quotas. The diagram below shows our attempt to work around this:

![InspecWrapperProfile2](https://user-images.githubusercontent.com/4711683/91165377-cf13f880-e6d0-11ea-8145-faead8012c19.png)

Here’s a partial output where you can see the duplicated execution of each of the controls:
```
  ↺  cis-gcp-6.1.1-db: [DB] Ensure that MySql database instance does not allow anyone to connect with administrative privileges. (2 skipped)
     ↺  This control is not scored
     ↺  This control is not scored
  ↺  cis-gcp-6.4-db: [DB] Ensure that Cloud SQL database instance requires all incoming connections to use SSL (2 skipped)
     ↺  [my-project] does not have CloudSQL instances.
     ↺  [my-project] does not have CloudSQL instances.
[...]

Issue #5215, opened 2020-08-19
InSpec CLI exec "--controls=foo" option is not working as expected


klose4711:
Using the [exec subcommand with option ](https://docs.chef.io/inspec/cli/#options-3)`--controls=foo` should only run the `foo` control, but it seems like it's still running all controls within the profile in background and only filtering the results output.
# Describe the problem
Have an InSpec profile with 3 controls: foo, bar, baz.

- Execute the InSpec profile without `--controls` option will run all three controls and will take e.g. 03:00 min.
- Execute the same InSpec profile with `--controls=foo` option should only run the one control, but it still takes 03:00 min. (Output is fine and only showing results for foo)
- Remove the other two controls (bar and baz) from the InSpec profile folder and then execute the profile will run only the one existing control and it will only take e.g. 00:25 min.

So it seems like that all controls in the profile folder affect the execution time, although only those selected with the option should be executed.

According to the option description only the selected controls should be run:
> `--controls=one two three` A list of control names to run, or a list of /regexes/ to match against control names. Ignore all other tests.

clintoncwolfe:
Thanks for the report, and the excellent description of the problem. Unfortunately, this reflects a deep-seated problem in inspec. InSpec [evaluates all controls](https://github.com/inspec/inspec/blob/40a81a580f1549bcc0ce276693709a817e354566/lib/inspec/profile.rb#L220) in order to access the metadata. Currently it does a full eval of the controls (but as I understand it, not the actual describe blocks), then [marks them for filtering](https://github.com/inspec/inspec/blob/40a81a580f1549bcc0ce276693709a817e354566/lib/inspec/profile.rb#L251) after the fact.

Schwad:
Hello @klose4711 -

Clinton hit the nail on the head there. Currently the code behaves as such:

```
control 'my control' do
  puts 'I will still be evaluated if waived or skipped! But not reported'
  sleep 10
  describe true do
    puts 'I will not run if waived or skipped!'
    sleep 10
    it { should eq true }
  end
end
```

The way to get the performance wins *today* is to leave as little time-sucking ruby code outside of the describe blocks as possible.

For example, you will really be hurt using `each` blocks to build your describe blocks, which is a pattern I've seen a lot:

```
google_storage_buckets.each do |bucket|
  describe bucket do
    ...
  end
end
```

In the longer term, this is a priority issue we are looking into as a team. I can't promise the fix tomorrow, but it is an issue product teams and engineering are aligned on resolving.

[...]

Issue #3760, opened 2019-01-28
skip_control may be confusing (it doesn't skip)


james-stocks:
### 🎛 Description

A customer wants to temporarily skip some controls. They use `skip_control` in a wrapper profile but find these controls have _no_ result rather than being skip results.

I believe this is as intended - `skip_control` and `skip` are different things. And `skip_control` would be more clearly named `exclude_control` ?

### Workaround

Use an `only_if`in the control with an appropriate condition.
Override the control in a wrapper profile and call `skip`.

james-stocks:
I believe this has been explained to the customer in question and they won't need an update on this; but I'll leave the issue open with the Question label - it might come up again.

jerryaldrichiii:
Yup, I think this should be `exclude_control` and then we can discuss if there should be a `skip_control`.

https://github.com/inspec/inspec/blob/a4362e7a0798af2c3eac3287fc083f6dc0529021/lib/inspec/control_eval_context.rb#L179

kevin-j-smith:
Hi, I have been looking for a possible solution to skipping a control inside of a control.  See the only_if solution, which your documentation states:
```txt
This shows how to allow skipping certain tests if conditions are not met, by using only_if. In this example the test will not be performed if redis-cli command does not exist. A optional message can say why it was skipped.
```

Does not actually report the control or mark it skip.  It just excludes it.   (inspec version 3.7.1).  As to the skip command with in a describe block:
```ruby
describe 'skip the rest of the tests...this does not' do
    skip 'The control still keeps executing' if some_check?
end
```

The only way for this to work is to wrap everything in a control structure but the logic could be too complex (ie looping through an array) and makes everything way less elegant.   The information in this issue around 'skip_control' would be nice to have as a escape mechanism within the control.  There for I could:
```ruby
control "V-X" do
    skip_control "User Account Administration attribute does not have packages defined: #{uaa}" if uaa[:packages].nil?
    skip_control "User Account Administration attribute does not have any packages defined: #{uaa}" if uaa[:packages].length == 0
    skip_control "User Account Administration attribute does not have config_files defined: #{uaa}" if uaa[:config_files].nil?
    skip_control "User Account Administration attribute does not have any config_files defined: #{uaa}" if uaa[:config_files].length == 0

    uaa[:packages].each do |serviceName|
      s = service(serviceName)
       skip_control "Service #{serviceName} is not installed"  if !s.installed?
       skip_control "Service #{serviceName} is not running" if !s.running?
[...]

PR #5327, opened 2020-12-04
InSpec Waivers: --filter-waived-controls

Schwad:
This PR is being adopted and brought up to a releasable state. Along the way we're probably going to drop the "ludicrous" term for something more formal :-), like `--filter-waived-controls`.

Major things include adding error checking, making the regexes less sensitive to accidental errors, and some refactoring/renaming.
@clintoncwolfe

----

This is an experimental PR to investigate alternative, optional flags for ensuring that waived controls are not evaluated at all. This could be considered __Ludicrous Speed__ for scaled customers.

ludicrous

## Background

**Links for background**

* https://github.com/inspec/inspec/issues/5068
* https://github.com/inspec/inspec/issues/3760
* https://github.com/inspec/inspec/issues/5225
* https://github.com/inspec/inspec/issues/5215
* https://github.com/inspec/inspec/issues/5215

TL;DR: Right now all control code outside describe blocks is silently evaluated, regardless of waiver. a `fail 'I broke this'` line or any other failure would break the control, but the most painful element is the performance hit. 

## This approach

This implementation will stop any waiver code from being executed. This will, as
a side effect, preclude all features that require skipped controls to be run. Those
features include (but are not limited to):

* Reporting number of skipped controls
* Output about results of skipped controls (sometimes we say if something skipped passed)
* `run: true` functionality no longer applies

It will, however, only run the correct controls.

This first implementation tests and focuses directly on the `waivers.yml` implementation; working on other paths is incidental and not tested in this MVP prototype.

# Current Behavior

Everything gets run behind the scenes. Even breaks in waived controllers!

```
bundle exec inspec exec test/fixtures/profiles/waivers/purely-broken-controls --waiver-file test/fixtures/profiles/waivers/purely-broken-controls/files/waivers.yml

#=>

Profile: purely-broken-controls
Version: 0.1.0
Target:  local://

  ×  01_my_broken_control: test/fixtures/profiles/waivers/purely-broken-controls/controls/purely_broken_controls.rb:1
     ×  Control Source Code Error test/fixtures/profiles/waivers/purely-broken-controls/controls/purely_broken_controls.rb:1 
     Everything is destroyed as this has been run
  ↺  02_my_working_but_waived_control: No-op
     ↺  Skipped control due to waiver condition: Sound reasoning
  ✔  03_my_working_control: true
     ✔  true is expected to eq true


Profile Summary: 1 successful control, 1 control failure, 1 control skipped
Test Summary: 1 successful, 1 failure, 1 skipped
```

# --ludicrous-speed
To test this passing, within the local directory run:

```
bundle exec inspec exec test/fixtures/profiles/waivers/purely-broken-controls --waiver-file test/fixtures/profiles/waivers/purely-broken-controls/files/waivers.yml

#=>
********************************************************************************

--ludicrous-speed is an experimental feature that stops waived controls altogether from being read into InSpec. Use with caution.

********************************************************************************
Profile: purely-broken-controls
Version: 0.1.0
Target:  local://

  ✔  03_my_working_control: true
     ✔  true is expected to eq true


```

`01` control should not exist at all or be skipped. Instead it fully evaluates and fails as seen by the failure. This is bad from an output perspective but also an issue from performance. That means a `sleep 100` (or some other time based issue) would stop everything for 100 seconds. This PR fixes that.

I have included a functional test that must be passed for this to work can be run as: `bundle exec m test/functional/waivers_test.rb:118`. This PR currently passes that test.

# --retain-waiver-data

This allows a second, alternative approach, where all prior existing functionality is retained. (i.e. we still have Skip counts, etc.) However we still don't evaluate any waived control code and it does not break. Run locally:

```
bundle exec inspec exec test/fixtures/profiles/waivers/purely-broken-controls --ludicrous-speed --retain-waiver-data --waiver-file test/fixtures/profiles/waivers/purely-broken-controls/files/waivers.yml

#=>

********************************************************************************

--ludicrous-speed is an experimental feature that stops waived controls altogether from being read into InSpec. Use with caution.

********************************************************************************
Profile: purely-broken-controls
Version: 0.1.0
Target:  local://

  ↺  01_my_broken_control: No-op
     ↺  Skipped control due to waiver condition: Sound reasoning
  ↺  02_my_working_but_waived_control: No-op
     ↺  Skipped control due to waiver condition: Sound reasoning
  ✔  03_my_working_control: true
     ✔  --- true is expected to eq true


Profile Summary: 1 successful control, 0 control failures, 2 controls skipped
Test Summary: 1 successful, 0 failures, 2 skipped
```

**TODO:**

I intend on moving this logic earlier in the code read and implement on the AST/AST-autocorrect level instead of on a pure ruby-parsing level. However this currently works as seen above with noted warnings and issues. (More discussion below)


lib/inspec/profile.rb
```diff
@@ -212,6 +212,36 @@ def collect_tests(include_list = @controls)
...
+        ui.red "*" * 80
+        ui.red "\n\n"
+        ui.red "--ludicrous-speed is an experimental feature that stops\
+ waived controls altogether from being read into InSpec. Use with caution."
+        ui.red "\n\n"
+        ui.red "*" * 80
+
+        ## Find the waivers file
+        collection_source_data_files = @runtime_profile.instance_variable_get(:@src).data_files
+        waiver_path = collection_source_data_files.keys.grep(/waivers.yml$/).first
```
clintoncwolfe: So, here's a still-gross-but-a-little-safer way to do this:
```
waiver_path = Inspec::Config.cached.instance_variable_get(:@cli_opts)["waiver_file"]&.first

## Pull together waiver
if waiver_path
  waived_controls = YAML.load_file(waiver_path).keys
end
```

That's better because it reaches less, and it also does not assume that the waiver file will be named `waivers.yml`.

It still has problems:

* We could expose cli_opts and avoid the instance_variable_get
* We are only accessing the first waiver file. What if there are are multiple?
* We don't handle misparsing the waiver file.

lib/inspec/profile.rb
```diff
@@ -212,6 +212,36 @@ def collect_tests(include_list = @controls)
 
         locked_dependencies.each(&:collect_tests)
 
+        # TODO: if --ludicrous-speed
```
clintoncwolfe:
```
if Inspec::Config.cached["ludicrous_speed"]
 #...
end
```

Schwad: Implemented, thanks! :)

lib/inspec/profile.rb
Schwad: Reformatted in this way, and manually referenced these issues in a comment above so we don't lose them

Schwad:
Updated PR and description to address comments and add additional `--retain-waiver-data` flag. This flag preserves prior functionality with maintaining skip counts while retaining performance boosts of ludicrous speed. (flags must be used together)

lib/inspec/profile.rb
```diff
@@ -213,6 +213,7 @@ def collect_tests(include_list = @controls)
         locked_dependencies.each(&:collect_tests)
 
         # Wipe out waived controls
+        require 'pry'; binding.pry
```
clintoncwolfe: Needs to be removed - debugger

lib/inspec/profile.rb
```diff
@@ -231,17 +230,33 @@ def collect_tests(include_list = @controls)
...
+            # waivers. This allows current waiver-data (e.g. skipis) to still
+            # be processed and rendered
+            tests.each do |key, value|
+              cleared_tests = value.split("control ").collect do |element|
+                next if element.blank?
+                splitlines = element.split("\n")
+                new_element = splitlines[0] + "\ndescribe '---' do\n" + splitlines[1..-1].join("\n") + "\nend\n"
+              end.join("control ")
+              purged_tests[key] = cleared_tests
+            end
```
clintoncwolfe: This needs to be looked at very closely, this seems.... fishy.

Schwad: If there's any way someone can find the lazy evaluation tactics used by RSpec in 'describe' and safely port that approach to 'control' we'd be in business.

I lost a few hours in a rabbit hole on that one.

Nothing inside a 'describe' block is evaluated when a controller is waived. (It can be tested with a 'raise' or 'sleep'). So until 'control' can be modified this would wrap all code in an unnamed 'describe' block at runtime.

I concur that the current approach is fishy- but should work in a desperate situation.

I hope I can find a few hours soon just to look at 'control' and the guts of 'describe' again.

If anyone cracks the 'control' case first let me know as I'd be very interested how it was done!

docs-chef-io/content/inspec/cli.md
```diff
@@ -272,6 +272,8 @@ This subcommand has additional options:
     Password for enable mode on Cisco IOS devices.
 * ``--filter-empty-profiles``, ``--no-filter-empty-profiles``
     Filter empty profiles (profiles without controls) from the report.
+* ``--filter-waived-controls``
+    Do not execute waived controls in InSpec at all. Must use with --waiver-file. Ignores `run` setting of waiver file.
```
IanMadd, line 301:
```suggestion
    Do not execute waived controls in InSpec at all. Must use with `--waiver-file`. Ignores `run` setting of waiver file.
```

PR #5338, opened 2020-12-09
Remove the kitchen.chef.yml file

tas50:
This is incredibly dated and doesn't look like it was ever setup correctly. It's mostly just logic copied from a kitchen.yml file in an omnibus install, but this has nothing to do with omnibus. Windows / mac tests with kitchen should be done in Azure Devops Pipelines similar to how we do it with chef to avoid the reliance on a developer locally running an hour of tests.


clintoncwolfe (approved):
Oh this is terrible.

PR #5362, opened 2020-12-18
Switch Docker build to be omnibus-based

nkierpiec:
This will inspec from packages.chef.io since it wont be on rubygems at the time of promotion
Signed-off-by: Nathaniel Kierpiec <nkierpiec@chef.io>

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.

PR #5377, opened 2021-01-07
Use ruby:2-alpine base image for Dockerfile

ramereth:
The ruby:alpine image has been updated to Ruby 3.x which InSpec is not currently ready for. Changing this to ruby:2-alpine to force using the latest 2.x release until InSpec is ready for Ruby 3.x.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:alpine
+FROM ruby:2-alpine
 LABEL maintainer="Chef Software, Inc. <docker@chef.io>"
 
 ARG EXPEDITOR_VERSION


ramereth:
This resolves the following [issue](https://gitlab.com/cinc-project/distribution/auditor/-/jobs/950074966#L127) we ran into with Cinc Auditor.

```
Step 9/15 : RUN gem install --no-document --source ${GEM_SOURCE} --version ${VERSION} inspec
---> Running in ee69eed837ec
ERROR:  While executing gem ... (NoMethodError)
  undefined method `request' for nil:NilClass
The command '/bin/sh -c gem install --no-document --source ${GEM_SOURCE} --version ${VERSION} inspec' returned a non-zero code: 1
```

clintoncwolfe:
We've decided to switch over to omnibus-based Docker builds - which makes it ubuntu based for now.

ramereth:
Ah, that would be #5362 I take it? Awesome, that works too!

PR #5414, opened 2021-02-21
Use first value of repeated settings in ssh_config and sshd_config

### LINKED ISSUE #5410: Resource sshd_config Not Handling Keyword Case Sensitivity Correctly
The `sshd_config` resource is not handling case sensitivity in regards to configuration keywords properly. Consequentially, the values that are given back for specified config keywords are potentially inaccurate.

# Describe the problem
The issue occurs when there are multiple instances of functionally identical keywords in the SSH config that have different capitalizations. Whichever capitalization occurs last will be returned by the resource as the keyword's configured value. It appears to be overwriting the values that came before it.

The sshd_config documentation states that **_"keywords are case-insensitive"_**. From this, alternate casings of configuration settings should be treated as identical. Additionally, that same documentation states, **_"For each keyword, the first obtained value will be used."_** whereas the`sshd_config` Inspec resource gives precedence to the last-occurring casing of a given keyword

* [Documentation for sshd_config](https://man.openbsd.org/sshd_config)

Here are a couple of examples of this issue occurring on a Linux node. Where "HostbasedAuthentication" and "HostBasedAuthentication" are competing.

#### Content From sshd_config
```
$ sudo cat /etc/ssh/sshd_config | grep -i hostbasedauthentication
HostBasedAuthentication yes
HostbasedAuthentication no
```
#### Inspec `sshd_config` Resource Value
```
inspec> sshd_config.hostbasedauthentication
=> "no"
```

#### Content From sshd_config
```
$ sudo cat /etc/ssh/sshd_config | grep -i hostbasedauthentication
HostBasedAuthentication yes
HostbasedAuthentication no
HostbasedAuthentication no
```
#### Inspec `sshd_config` Resource Value
```
inspec> sshd_config.hostbasedauthentication
=> ["no", "no"]
```

## Possible Solution
I think that this issue can be fixed by adjusting [this line of ssh_config](https://github.com/inspec/inspec/blob/master/lib/inspec/resources/ssh_config.rb#L42) and disabling `multiple_values` in the [SimpleConfig reader](https://github.com/inspec/inspec/blob/master/lib/inspec/resources/ssh_config.rb#L75).

The convert_hash method could be changed to assign a value to a key only if it does not already exist. Ex:
```ruby
    def convert_hash(hash)
      hash.each_with_object({}) do |(k, v), new_hash|
        new_hash[k.downcase] ||= v
      end
    end
```

With these two changes, the `sshd_config` Inspec resource will more accurately reflect the way that the sshd_config operates on a system. It would only accept the first-occuring setting for a given keyword, regardless of the alternative capitalizations for that keyword that may occur later in the file.

clintoncwolfe:
## Description

Alters the `ssh_config` and `sshd_config` resources to use the value from the first entry specified (rather than the last) when a setting is repeated, in accordance with the spec.

## Related Issue

Fixes #5410 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/ssh_config.rb
+++ b/lib/inspec/resources/ssh_config.rb
@@ -39,7 +39,7 @@ module Inspec::Resources
     def convert_hash(hash)
       new_hash = {}
       hash.each do |k, v|
-        new_hash[k.downcase] = v
+        new_hash[k.downcase] ||= v
       end
       new_hash
     end
--- a/test/fixtures/files/ssh_config
+++ b/test/fixtures/files/ssh_config
@@ -4,3 +4,6 @@ Host *
     SendEnv LANG LC_*
     HashKnownHosts yes
     GSSAPIAuthentication no
+    HostBasedAuthentication yes
+    HostbasedAuthentication no
+    HostbasedAuthentication no
--- a/test/unit/resources/ssh_conf_test.rb
+++ b/test/unit/resources/ssh_conf_test.rb
@@ -18,6 +18,11 @@ describe "Inspec::Resources::SshConfig" do
       _(resource.gssapiauthentication).must_equal "no"
       _(resource.GSSAPIAuthentication).must_equal "no"
     end
+
+    it "uses the first value encountered" do
+      resource = load_resource("ssh_config")
+      _(resource.HostBasedAuthentication).must_equal "yes"
+    end
   end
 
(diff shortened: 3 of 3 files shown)


CernerAssociate-ThomasDeranek (approved):
Thank you for getting this fix made quickly!

PR #5422, opened 2021-02-28
Expose conf_path, content, and params on auditd_conf

yarick:
In order to allow for dynamic tests, exposed these instance variables


aaronlippold:
Generally speaking, we should look at doing this for all the resources, aka if we already have this data in the resource, we should expose it as it just helps our end-users.

clintoncwolfe (approved):

clintoncwolfe:
Would you mind adding a few lines to the docs to indicate that the properties are now available?  Thanks!

https://github.com/inspec/inspec/blob/master/docs-chef-io/content/inspec/resources/auditd_conf.md

clintoncwolfe (approved):
Thanks!

PR #5426, opened 2021-03-05
Fix google_project_alert_policy Examples in the docs

wmetaw:
Signed-off-by: Ryo Takashima <lio114514@gmail.com>

Fix google_project_alert_policy Examples in the docs

## Description

- The loop is not executed because `.each` does not exist
- [ALERT_POLICY_ID] is correct for the argument of `name:`

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/google_project_alert_policy.md
+++ b/docs-chef-io/content/inspec/resources/google_project_alert_policy.md
@@ -32,15 +32,15 @@ end
 
 ### Test that a GCP alert policy is enabled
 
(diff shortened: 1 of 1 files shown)


clintoncwolfe (approved):

IanMadd:
Should this change also be applied to the same doc in inspec-gcp? I'm not sure how that content is updated or generated, but if this change is added here, it should also be there too.

dkumaras:
Pls check changes and let us know if we did meet the expectations.

dkumaras (approved):

sathish-progress (approved):
LGTM

PR #5432, opened 2021-03-16
Add m1 support to MacOS build list

### LINKED ISSUE #5429: MacOS m1 chip support
### 🎛 Description

We need to produce an omnibus package for MacOS on the ARM m1 chip, so that MacOS users can continue to use InSpec with a native package.

clintoncwolfe:

## Description

Adds m1 chip (ARM) support to our list of platforms for which we build packages.
As part of this, conditionally selects openssl version based on which platform we are building.

## Related Issue

Closes #5429 

## Types of changes

## Checklist:

### DIFF
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -4,3 +4,6 @@
 train_stable = /^train \((.*)\)/.match(`gem list ^train$ --remote`)[1]
 override "train", version: "v#{train_stable}"
 override "ruby", version: "2.7.2"
+
+# Mac m1
+override "openssl", version: "1.1.1j" if mac_os_x? && arm?
--- a/omnibus_overrides.rb
+++ b/omnibus_overrides.rb
@@ -6,4 +6,4 @@ override "train", version: "v#{train_stable}"
 override "ruby", version: "2.7.2"
 
 # Mac m1
-override "openssl", version: "1.1.1j" if mac_os_x? && arm?
+override "openssl", version: "1.1.1j" if mac_os_x?
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -26,6 +26,8 @@ builder-to-testers-map:
     - mac_os_x-10.14-x86_64
     - mac_os_x-10.15-x86_64
     - mac_os_x-11-x86_64
+  mac_os_x-11-arm64:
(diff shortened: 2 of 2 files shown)


tduffield (approved):

omnibus_overrides.rb
```diff
@@ -4,3 +4,6 @@
 train_stable = /^train \((.*)\)/.match(`gem list ^train$ --remote`)[1]
 override "train", version: "v#{train_stable}"
 override "ruby", version: "2.7.2"
+
+# Mac m1
+override "openssl", version: "1.1.1j" if mac_os_x? && arm?
```
tduffield: I feel like we would want to just update this for everything, no?

clintoncwolfe: InSpec has a FIPS build which requires 1.0.2. I'm not sure anyone is using the FIPS build, but it exists.

aaronlippold:
Yes. Folks are using the fips build

omnibus_overrides.rb
tas50: We do not support FIPS on macOS platforms so this should not be an issue for us on macOS.

btm: I want to add some clarity for anyone passing by. There may be a little confusion here talking about Inspec FIPS builds versus Inspec macOS FIPS builds.

For all the products that have OpenSSL FIPS builds, we build those on RHEL + Windows only, and it's always been that way. There are customers/users using this functionality, we need to keep it (on RHEL/Windows), and it's why we've stayed with OpenSSL 1.0.2 so long.

Other platforms like macOS or Ubuntu do not have the necessary pieces in the Omnibus builds. If anyone is using FIPS on those other platforms they aren't doing it using our builds.

We've stayed on OpenSSL 1.0.2 on the non-FIPS platforms to maintain consistency, but we need a newer OpenSSL for macOS ARM. It makes sense if we're moving to 1.1.1 for macOS ARM to also switch to 1.1.1 for macOS x64 to maintain that platform consistency. I don't think we've justified yet moving all the other non-FIPS platforms to 1.1.1 yet, mostly because we don't have a need to yet.

btm: tldr; I support dropping the `&& arm?` part, but we need to keep the `if mac_os_x?` part.

PR #5434, opened 2021-03-17
Fix for --controls option is not working as expected.

### LINKED ISSUE #5215: InSpec CLI exec "--controls=foo" option is not working as expected

Vasu1105:

## Description

As mentioned in issue #5215 the control block also gets evaluated and due to which the `--controls ` option is not working as expected.
The reason behind that the filtering of control here https://github.com/inspec/inspec/blob/master/lib/inspec/profile.rb#L228 is happening after evaluating the controls which are happening here https://github.com/inspec/inspec/blob/master/lib/inspec/profile.rb#L220

which calls this https://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/profile_context.rb#L154 and then this https://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/profile_context.rb#L162

which then calls this control definition https://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/control_eval_context.rb#L54
and describe definition  (descrhttps://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/control_eval_context.rb#L66

both these methods instantiate Inspec::rule https://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/rule.rb#L46
which basically causing the skipped control is getting evaluated (note not the describe block)

I have added a method in the control_eval_context.rb to filter the controls before evaluating it. I need to add the filter in describe method also because when we are having only describe block without control block the methods creates auto-generated control as described here https://github.com/inspec/inspec/blob/2f41c16ad444eaf69f7ec77f4ff681f35aad9210/lib/inspec/control_eval_context.rb#L63
so I am making sure those describe blocks are also get filtered when specifying --controls option.

## Related Issue

Fixes #5215
## Types of changes

## Checklist:

### DIFF
--- a/test/fixtures/profiles/controls-option-test/controls/example.rb
+++ b/test/fixtures/profiles/controls-option-test/controls/example.rb
@@ -31,3 +31,39 @@ control "11_pass2" do
     it { should cmp 'a thing' }
   end
 end
+
+describe 'a thing' do
+  it { should cmp 'a thing' }
+end
+
+describe.one do
+  describe 'ConfigurationA' do
+    it { should cmp 'a thing' }
+  end
+
+  describe 'ConfigurationB' do
+    it { should cmp 'a thing' }
+  end
+end
+
+title '/ profile'
+
+# you add controls here
+control 'tmp-1.0' do                                   # A unique ID for this control
+  impact 0.7                                           # The criticality, if this control fails.
+  title 'Create / directory'                        # A human-readable title
+  desc 'An optional description...'                    # Describe why this is needed
+  desc 'label', 'An optional description with a label' # Pair a part of the description with a label
+  tag data: 'temp data'                                # A tag allows you to associate key information
+  tag 'security'                                       # to the test
+  ref 'Document A-12', url: 'http://...'               # Additional references
+
+  describe file('/') do                             # The actual test
+    it { should be_directory }
+  end
+end
+
+# you can also use plain tests
+describe file('/') do
+  it { should be_directory }
+end
--- a/test/fixtures/profiles/controls-option-test/controls/example.rb
+++ b/test/fixtures/profiles/controls-option-test/controls/example.rb
@@ -1,5 +1,3 @@
-# copyright: 2018, The Authors
-
 control "foo" do
   describe 'a thing' do
     it { should cmp 'a thing' }
@@ -8,13 +6,6 @@ end
 
 control "bar" do
   puts 'bar'
-  describe 'a thing' do
-    it { should cmp 'a thing' }
-  end
-end
-
-control "baz" do
-  puts 'baz'
   describe 'a thing' do
     it { should cmp 'a thing' }
   end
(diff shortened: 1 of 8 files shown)


test/functional/inspec_exec_test.rb
```diff
@@ -178,6 +178,25 @@ def stderr
     assert_exit_code 100, out
   end
 
+  it "executes only specified controls when selecting the controls by literal names" do
+    inspec("exec " + File.join(profile_path, "filter_table") + " --no-create-lockfile --controls foo")
```
clintoncwolfe: Why did you choose the filter_table fixture? I don't think this problem has anything to do with FilterTable. I would suggest either finding a different profile, or creating a new one just for this purpose.

test/functional/inspec_exec_test.rb
```diff
@@ -178,6 +178,25 @@ def stderr
...
+    _(stdout).must_equal "
+Profile: InSpec Profile (filter_table)
+Version: 0.1.0
+Target:  local://
+
+  ✔  foo: a thing
```
clintoncwolfe: This is likely too brittle of a test - the check marks vary under Windows, for example. I suggest you write tests that look specifically for what you are looking for. For example, 

```
_(out.stdout).wont_include "bar"
_(out.stdout).wont_include "baz"
```

test/functional/inspec_exec_test.rb
Vasu1105: Yes after checking the impact decided to remove this from here. I thought since we are using --controls options in test with this profile that will be good place to include. But yes definitely this is not the right place to have this. Making the required changes.

test/functional/inspec_exec_test.rb
```diff
@@ -178,6 +178,27 @@ def stderr
...
+    inspec("exec " + File.join(profile_path, "controls-option-test") + " --no-create-lockfile --controls foo")
+    _(out.stdout).must_include "foo"
+    _(out.stdout).wont_include "bar"
+    _(out.stdout).wont_include "baz"
+    _(stderr).must_equal ""
+
+    assert_exit_code 0, out
+  end
+
+  it "executes only specified controls when selecting the controls by literal names" do
```
clintoncwolfe: This selects by regex, not by literal name

test/fixtures/profiles/controls-option-test/controls/example.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  describe file('/') do                             # The actual test
+    it { should be_directory }
+  end
+end
+
+# you can also use plain tests
+describe file('/') do
+  it { should be_directory }
+end
```
clintoncwolfe: This file has a lot of extra material in it that is not being tested. Either it should be removed or it should be tested - fixture files should be minimal. Thanks!

clintoncwolfe (approved):
Just a couple of minor notes about the test fixture file, nit-picking really. This works really well. I think users will be very happy with this improvement!

test/functional/inspec_exec_test.rb
Vasu1105: Updated.

test/fixtures/profiles/controls-option-test/controls/example.rb
Vasu1105: Thanks @clintoncwolfe .Did the required changes.

klose4711:
Thanks for fixing this! 🙂

PR #5435, opened 2021-03-17
Updates profile init for cloud platforms to use inputs

collinmcneese:
## Description

Currently when using `inspec init profile` with providers `aws` or `gcp`, templates which are generated use `attributes` instead of `inputs`.  This updates these templates to generate sample profiles using `inputs` instead of `attributes`

## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks for this!

Issue #5233, opened 2020-08-31
executing profile from remote repository expects default branch named 'master'


mjmenger:
# Describe the problem

when executing an inspec profile from a remote repository with a default branch not named *master* the following error is returned.
```
inspec exec https://github.com/org/profile.git

Profile git dependency failed - unable to resolve master to a specific git commit for https://github.com/org/profile.git
```
The expected behavior is that the profile should execute independent of the name of the default branch (e.g. master, primary, main, production) without having to designate the name of the branch.

Schwad:
Hey @mjmenger thanks for filing this issue. That is correct I believe it is expected behavior that whatever the `default` branch is if we exec directly against github we should be using that one. 

For anyone looking to pick this up in a PR - it would appear that these requisite lines are a good place to start for untangling this issue:

https://github.com/inspec/inspec/blob/d00706b9a6c01d4a345d70f082d612e06b5c60db/lib/inspec/fetcher/git.rb#L121-L131

deric4:
Would implementing this using the  output of `git remote show <repo url>`  and parsing the `HEAD branch:` line make sense?

```shell-session
# random repo with default branch set to 'main'
$ git remote show https://github.com/github/docs
* remote https://github.com/github/docs
  Fetch URL: https://github.com/github/docs
  Push  URL: https://github.com/github/docs
  HEAD branch: main
  ...

# repo with default still using 'master'
$ git remote show git@github.com:inspec/inspec.git
* remote git@github.com:inspec/inspec.git
  Fetch URL: git@github.com:inspec/inspec.git
  Push  URL: git@github.com:inspec/inspec.git
  HEAD branch: master
  ...
```

PR #5438, opened 2021-03-22
Fix for executing git profiles independent of the name of the default branch

Nik08:
Signed-off-by: Nikita Mathur @Nik08 
## Description

Solves the problem of executing GIT profiles independent of the name of the default branch (e.g. master, primary, main, production) without having to designate the name of the branch.

## Related Issue

Fixes [#5233 ](url)

## Types of changes

## Checklist:


lib/inspec/fetcher/git.rb
```diff
@@ -1,7 +1,7 @@
-require "tmpdir" unless defined?(Dir.mktmpdir)
-require "fileutils" unless defined?(FileUtils)
-require "mixlib/shellout" unless defined?(Mixlib::ShellOut)
-require "inspec/log"
+require 'tmpdir' unless defined?(Dir.mktmpdir)
+require 'fileutils' unless defined?(FileUtils)
+require 'mixlib/shellout' unless defined?(Mixlib::ShellOut)
+require 'inspec/log'
```
Vasu1105: Keep the double quotes that is convention that Chef Inspec follows as per my understanding. This change will fail you lint test.

clintoncwolfe: To run the linter, run `bundle exec chefstyle` and to autocorrect, run `bundle exec chefstyle -a`

test/functional/git_fetcher_test.rb
```diff
@@ -90,55 +109,70 @@ def assert_relative_fetch_works(profile_name, expected_profiles, expected_contro
...
+      assert_includes run_result.stderr, 'test/fixtures/profiles/git-fetcher/git-repo-01'
       assert_exit_code(1, run_result) # General user error
     end
   end
+
+  #------------ Happy Case for default branch GIT fetching -------------------#
+
+  describe 'running a remote GIT profile' do
+    it 'should use default HEAD branch' do
+      inspec("exec #{git_profiles}/git-repo-default-main")
```
clintoncwolfe: Here I would expect you to directly execute a git-based profile using a git URL, rather than using a local profile with a git dependency. Like this:
```
inspec("exec https://github.com/inspec/inspec-test-profile-default-main.git")
```

That directly tests the ability of InSpec to guess the default branch name.

test/fixtures/profiles/git-fetcher/git-repo-default-main/inspec.yml
```diff
@@ -0,0 +1,12 @@
...
+license: Apache-2.0
+summary: An inspec test profile which executes remote git profile.
+version: 0.1.0
+supports:
+  platform: os
+depends:
+- name: default-main
+  git: https://github.com/inspec/inspec-test-profile-default-main
+  branch:  main
+  version: 0.1.0
```
clintoncwolfe: I don't think you need this test fixture, you can just call `https://github.com/inspec/inspec-test-profile-default-main.git` directly from the CLI.

test/unit/fetchers/git_test.rb
```diff
@@ -54,82 +54,144 @@
...
+      expect_checkout(git_master_ref, 'fetchpath')
       result = fetcher.resolve({ git: git_dep_dir })
-      result.fetch("fetchpath")
-      _(result.archive_path).must_equal "fetchpath"
+      result.fetch('fetchpath')
+      _(result.archive_path).must_equal 'fetchpath'
+    end
+  end
+
+  describe 'when given a repository with default branch main' do
```
clintoncwolfe: Great tests!

lib/inspec/fetcher/git.rb
```diff
@@ -126,18 +125,26 @@ def resolved_ref
                         elsif @tag
                           resolve_ref(@tag)
                         else
-                          resolve_ref("master")
+                          resolve_ref(default_ref)
                         end
     end
 
+    def default_ref
+      shellout("git remote show #{@remote_url} | grep 'HEAD branch' | cut -d ':' -f 2").stdout&.strip
```
clintoncwolfe: I'm going to ask for two changes here:

First, the Windows test is failing because `grep` is not available on Windows. You'll need to do the searching and cutting in Ruby, which is easy enough. I found this worked, but you may be able to come up with something neater:

```
shellout("git remote show #{@remote_url}").stdout&.lines.detect { |l| l.include? "HEAD branch:" }.split(":").last.strip
```

Secondly, there is a very real chance the user may not have `git` installed on their computer. We need to handle that (and other error conditions) more gracefully. Add error handling like that used in `resolve_ref` - check the exit_status and raise an exception.

clintoncwolfe (changes requested):
Great first PR! 

Looks like you have your editor setup to transform quotes and do line wrapping. You'll want to disable that or alter it to match ChefStyle. We want our PRs to be as minimal as possible, introducing only relevant changes.

I'm submitting my review as "Changes Requested" because two changes need to be made to the CLI invocation in `default_ref` - one for Windows compatibility, and one for error handling when git is unavailable.

Overall, very good work, looking forward to your updates on the PR!

lib/inspec/fetcher/git.rb
Nik08: Got it! I will make the changes as per the chef style.

test/functional/git_fetcher_test.rb
Nik08: These points are insightful. Will work on the updates.

lib/inspec/fetcher/git.rb
Nik08: These points are insightful. Will work on the updates.

clintoncwolfe (approved):
Looks good - thanks!

PR #5446, opened 2021-03-31
Bug fix for loading hashmap inputs consistently

### LINKED ISSUE #4057: Hashes in inputs from inspec.yml and inputs file loaded differently
### 🎛 Description

This is a 🐛 bug report 

Attribute Hashes loaded from attributes.yml (note lack of : ):
{"us-east-1"=>
{:s3_bucket_name=>"s3_bucket_name_value", :sns_topic_arn=>"sns_topic_arn_value"}
}

Attribute Hashes loaded from inspec.yml (note : ):
{:"us-east-1"=>
{:s3_bucket_name=>"s3_bucket_name_value", :sns_topic_arn=>"sns_topic_arn_value"}
}

Effect: 

Identical hash key reference will fail for one of them and succeed on the other. 

test1['us-east-1'] or test1[:'us-east-1']

Seen in inspec -v 3.9.3 and 4.3.2

Nik08:

## Description

**Issues:** 
1. The inputs loaded from external file and metadata file were read in two different formats by InSpec, one hash with `symbol` keys and another hash with `string` as keys
2. Hash inputs loaded using runner API and profile DSL also, returns success only in one of the formats.

**Impact of solution:**

This approach will solve reading inputs irrespective of their formats, which enables using hashes inside InSpec controls in any way preferred.

## Related Issue

Fixes #4057 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/input_registry.rb
+++ b/lib/inspec/input_registry.rb
@@ -256,7 +256,7 @@ module Inspec
 
         data.inputs.each do |input_name, input_value|
           evt = Inspec::Input::Event.new(
-            value: input_value&.class.eql?(Hash) ? Thor::CoreExt::HashWithIndifferentAccess.new(input_value) : input_value,
+            value: input_value.is_a?(Hash) ? Thor::CoreExt::HashWithIndifferentAccess.new(input_value) : input_value,
             provider: :cli_files,
             priority: 40,
             file: path
@@ -307,7 +307,7 @@ module Inspec
     def handle_raw_input_from_metadata(input_orig, profile_name)
       input_options = input_orig.dup
       input_name = input_options.delete(:name)
-      input_options[:value] = Thor::CoreExt::HashWithIndifferentAccess.new(input_options[:value]) if input_options[:value]&.class.eql?(Hash)
+      input_options[:value] = Thor::CoreExt::HashWithIndifferentAccess.new(input_options[:value]) if input_options[:value].is_a?(Hash)
       input_options[:provider] = :profile_metadata
       input_options[:file] = File.join(profile_name, "inspec.yml")
       input_options[:priority] ||= 30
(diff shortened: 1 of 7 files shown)


clintoncwolfe:
@aaronlippold Does this approach - using HashWithIndifferentAccess - work for your use case?

lib/inspec/input_registry.rb
```diff
@@ -256,7 +256,7 @@ def bind_inputs_from_input_files(profile_name, file_list)
 
         data.inputs.each do |input_name, input_value|
           evt = Inspec::Input::Event.new(
-            value: input_value,
+            value: input_value.is_a?(Hash) ? Thor::CoreExt::HashWithIndifferentAccess.new(input_value) : input_value,
```
clintoncwolfe: For consistency, I wonder if we should make a similar change in `bind_inputs_from_runner_api`? Testing that would be difficult, however.

aaronlippold:
I'm not sure given that there isn't really a description of the change. Can you elaborate on what is being updated and remind me which use case we're talking about :-)

lib/inspec/input_registry.rb
Nik08: Yes for consistency that seems right. Let me look into the testing front for this.

Vasu1105: Can we add a method to handle this conversion `input_value.is_a?(Hash) ? Thor::CoreExt::HashWithIndifferentAccess.new(input_value) : input_value` so that we can just call that method wherever required and code will be maintainable ?

Nik08: Yes these checks could be re-used by having just a method :)

Nik08: @clintoncwolfe I observed this issue not only additionally happens for runner api input but also for other ways of input. Will try an alternate solution which could work for all the input types, rather than handling it in each method definition for different input types. Hope that would be a better solution.

clintoncwolfe (approved):
Great testing - very good work!

Issue #4342, opened 2019-07-30
group and groups resource: members does not work when the group has multiple members


steversmith:
### 🎛 Description

This is a 🐛 bug report

group and groups resource: members does not work when the group has multiple members

### 🌍 InSpec and Platform Version
Redhat Linux 6.6 running via the test kitchen:
inspec version: 3.9.3
kitchen version: 1.25.0
kitchen-inspec: 1.1.0

### 🤔 Replication Case

Using a simple cookbook recipe, I created a VM with two test groups:
test1 has one member, userA
test2 has two members, userA and userB
```
grep test /etc/group
test1:x:509:userA
test2:x:510:userA,userB
```

I setup two control groups for testing: both controls should pass, but Group test2 fails.
```
control 'Group test1' do
  describe group('test1') do
    its("members") { should include 'userA' }
    its("members") { should be_in ['userA'] }
  end

  describe groups.where { name == 'test1' } do
    its("members") { should include 'userA' }
    its("members") { should be_in ['userA'] }
  end
end

control 'Group test2' do
  describe group('test2') do
    its("members") { should include 'userA' }
    its("members") { should include 'userB' }
    its("members") { should be_in ['userA', 'userB'] }
  end

  describe groups.where { name == 'test2' } do
    its("members") { should include 'userA' }
    its("members") { should include 'userB' }
    its("members") { should be_in ['userA', 'userB'] }
  end
end
```

### 💁 Possible Solutions

I think that the root cause is that:
group.members returns a CSV formatted String of the members
groups.members returns a single element array with CSV formatted String of the members

### 💻 Stacktrace
The test output produces:
```
[PASS]  Group test1: Group test1
  [PASS]  Group test1 members should include "userA"
  [PASS]  Group test1 members should be in "userA"
  [PASS]  Groups with name == "test1" members should include "userA"
  [PASS]  Groups with name == "test1" members should be in "userA"
[FAIL]  Group test2: Group test2 (4 failed)
  [PASS]  Group test2 members should include "userA"
  [PASS]  Group test2 members should include "userB"
  [FAIL]  Group test2 members should be in "userA" and "userB"
  expected `userA,userB` to be in the list: `["userA", "userB"]`
  [FAIL]  Groups with name == "test2" members should include "userA"
  expected ["userA,userB"] to include "userA"
  [FAIL]  Groups with name == "test2" members should include "userB"
  expected ["userA,userB"] to include "userB"
[...]

PR #5479, opened 2021-04-23
Added new property `members_array` for group & groups resources.

Nik08:

## Description

1. Added new property `members_array` for group & groups resources.
2. Documented the nature of members for both the resources in Windows and Non-windows platform.

## Related Issue

Solutions #4342 
## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/group.md
+++ b/docs-chef-io/content/inspec/resources/group.md
@@ -62,6 +62,20 @@ The `members` property tests the members that belong to the group:
 
     its('members') { should include 'root' }
 
+where `members` returns
+- Array of group members for **Windows Platform**.
+Example: `["member1", "member2"]`
+- CSV formatted string of group members for **Non-Windows Platforms**.
+Example: `"member1,member2"`
+
+
+### members_array
+
+The `members_array` property tests the members that belong to a group just like `members` property.
+But the value returned by this property is always an array of group members.
+
+    its('members_array') { should include 'root' }
+
 ## Matchers
 
 For a full list of available matchers, please visit our [matchers page](/inspec/matchers/).
--- a/docs-chef-io/content/inspec/resources/group.md
+++ b/docs-chef-io/content/inspec/resources/group.md
@@ -62,17 +62,22 @@ The `members` property tests the members that belong to the group:
 
     its('members') { should include 'root' }
 
-where `members` returns
-- Array of group members for **Windows Platform**.
-Example: `["member1", "member2"]`
-- CSV formatted string of group members for **Non-Windows Platforms**.
-Example: `"member1,member2"`
+where `members` returns:
+
+- an array of group members for **Windows Platform**.
+
+    Example: `["member1", "member2"]`
+
+- a CSV formatted string of group members for **Non-Windows Platforms**.
+
+    Example: `"member1,member2"`
 
 
 ### members_array
 
-The `members_array` property tests the members that belong to a group just like `members` property.
-But the value returned by this property is always an array of group members.
+The `members_array` property tests the members that belong to a group just like the
+`members` property,
+but the value returned by this property is always an array of group members.
 
     its('members_array') { should include 'root' }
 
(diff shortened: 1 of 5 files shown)


docs-chef-io/content/inspec/resources/group.md
```diff
@@ -62,6 +62,20 @@ The `members` property tests the members that belong to the group:
 
     its('members') { should include 'root' }
 
+where `members` returns
+- Array of group members for **Windows Platform**.
+Example: `["member1", "member2"]`
+- CSV formatted string of group members for **Non-Windows Platforms**.
+Example: `"member1,member2"`
```
IanMadd:
```suggestion
where `members` returns:

- an array of group members for **Windows Platform**.

   Example: `["member1", "member2"]`

- a CSV formatted string of group members for **Non-Windows Platforms**.

   Example: `"member1,member2"`
```

docs-chef-io/content/inspec/resources/group.md
```diff
@@ -62,6 +62,20 @@ The `members` property tests the members that belong to the group:
...
+
+
+### members_array
+
+The `members_array` property tests the members that belong to a group just like `members` property.
+But the value returned by this property is always an array of group members.
```
IanMadd:
```suggestion
The `members_array` property tests the members that belong to a group just like the `members` property,
but the value returned by this property is always an array of group members.
```

docs-chef-io/content/inspec/resources/groups.md
```diff
@@ -74,6 +74,20 @@ The `members` property tests the members that belong to a group:
     its('members') { should include 'root' }
     its('members') { should include 'Administrator' }
 
+where `members` returns
+- Array of group members for **Windows Platform**.
+Example: `["member1", "member2"]`
+- Single element array with CSV formatted string of group members for **Non-Windows Platforms**.
+Example: `["member1,member2"]`
```
IanMadd:
```suggestion
where `members` returns:

- an array of group members for **Windows Platform**.

   Example: `["member1", "member2"]`

- a CSV formatted string of group members for **Non-Windows Platforms**.

   Example: `"member1,member2"`
```

docs-chef-io/content/inspec/resources/groups.md
```diff
@@ -74,6 +74,20 @@ The `members` property tests the members that belong to a group:
...
+
+### members_array
+
+The `members_array` property tests the group members just like `members` property.
+But the value returned by this property is always an array of group members.
```
IanMadd:
```suggestion
The `members_array` property tests the group members just like the `members` property,
but the value returned by this property is always an array of group members.
```

docs-chef-io/content/inspec/resources/groups.md
Nik08: @IanMadd Is there any reason behind the edit on line 83 and 85 because it goes against the fact that it returns a single element array of format `["member1,member2"]` ?
We should not let this information change I feel because it does return an array in groups resource on line 83 & 85.

Vasu1105: If I am not wrong we are introducing new property so that it returns the array of members regardless of any OS, so as per me we should return the result in the format of ["member1", "member2"] regardless of any OS and if they are different set of members and not the ["member1, member2"]..

Nik08: @Vasu1105 Yes this document change above is for already existing property `members` to clarify more on the way it is already working for different platforms. For `members_array` property the documents always states to return an array irrespective of the platform.

IanMadd: How about this instead:

`a single element array that contains a CSV string of group members for **Non-Windows Platforms**.`

clintoncwolfe (approved):
Looks good, just need to update the docs - thanks!

Vasu1105 (approved):
Looks Good. Thanks @Nik08

PR #5483, opened 2021-04-26
Group & Groups doc updated - about using local and etc groups

### LINKED ISSUE #5115: group and groups resources should be clearly documented as being for local groups
https://github.com/inspec/inspec/blob/master/lib/inspec/resources/groups.rb

The Unix group data comes from etc_group which only looks at the file `/etc/group` and does not see LDAP groups.
The Windows group data appears to query the local machine, I think this also means it only returns local groups?

The resource is documented as `Use the groups Chef InSpec audit resource to test multiple groups on the system.` which does somewhat hint that it is local groups; but this should be more clearly documented.

Aha! Link: https://chef.aha.io/features/SH-2380

Nik08:

Documents updation to mention clearly about the local and etc groups used in group and groups resources for different platforms.
## Related Issue

Fixes #5115 
## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/group.md
+++ b/docs-chef-io/content/inspec/resources/group.md
@@ -13,11 +13,11 @@ platform = "os"
 
 Use the `group` Chef InSpec audit resource to test a single group on the system.
 
-Following system group is used in the `group` resource:
+The `group` resource uses the following system groups:
 
-- in **Non-Windows** system, resource only works with a group listed in file `/etc/group` and not a local group of the system.
+- On **non-Windows** systems the group resource tests a local group defined in the`/etc/group` file.
 
-- in **Windows** system, resource only works with a local group of the system.
+- On **Windows** systems the group resource tests a local group defined by Local Users and Groups.
 
 
 ## Availability
--- a/docs-chef-io/content/inspec/resources/group.md
+++ b/docs-chef-io/content/inspec/resources/group.md
@@ -15,8 +15,9 @@ Use the `group` Chef InSpec audit resource to test a single group on the system.
 
 Following system group is used in the `group` resource:
 
-- in **Non-Windows** system, resource only works with a group listed in file `/etc/group` and not a local group.
-- in **Windows** system, resource only works with a local group.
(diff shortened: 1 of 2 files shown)


docs-chef-io/content/inspec/resources/group.md
```diff
@@ -13,6 +13,13 @@ platform = "os"
 
 Use the `group` Chef InSpec audit resource to test a single group on the system.
 
+Following system group is used in the `group` resource:
+
+- in **Non-Windows** system, resource only works with a group listed in file `/etc/group` and not a local group of the system.
+
+- in **Windows** system, resource only works with a local group of the system.
```
IanMadd: I'm not sure if this is correct, but it would be good to define where or how the group policy is specified on Windows 

```suggestion
The `group` resource uses the following system groups:

- On **non-Windows** systems the group resource tests a local group defined in the`/etc/group` file.
- On **Windows** systems the group resource tests a local group defined by Local Users and Groups.
```

docs-chef-io/content/inspec/resources/groups.md
```diff
@@ -13,6 +13,12 @@ platform = "os"
 
 Use the `groups` Chef InSpec audit resource to test multiple groups on the system.
 
+Following system groups are used in the `groups` resource:
+
+- in **Non-Windows** system, resource only works with groups listed in file `/etc/group` and not local groups of the system.
+
+- in **Windows** system, resource only works with local groups of the system.
```
IanMadd: Same as above
```suggestion
The `groups` resource uses the following system groups:

- On **non-Windows** systems the group resource tests local groups defined in the`/etc/group` file.
- On **Windows** systems the group resource tests local groups defined by Local Users and Groups.
```

docs-chef-io/content/inspec/resources/group.md
Nik08: @IanMadd I think it sounds correct. @clintoncwolfe @Vasu1105 Any thoughts?

clintoncwolfe: I believe Ian's suggestion is correct.

Issue #4318, opened 2019-07-18
Inheritance - exec with --controls filters out required dependent controls


kevin-j-smith:
# Describe the problem
I am trying to nest my controls for better orchestration of what is ran across development pipelines.  I also want to nest my controls into groups.   Take for instance
```
inheritance
|-- controls
   |-- dev.rb
|-- inspec.yml
|-- profile_group_1
   |-- controls
      |-- c1.rb
   |-- inspec.yml
```

```ruby
# dev.rb
require_controls 'group_1' do
  control 'c1'
end
```

```ruby
# inspec.yml
name: inheritance
version: 0.1.0
depends:
  - name: group_1
     path: ./profile_group_1
```

```ruby
# c1.rb
control 'c1' do
  title 'hello'
  desc 'world'
  impact 0.5
  describe 'HELLO' do
    it { should cmp 'HELLO' }
  end
end
```

When I run: 
```bash
inspec exec --no-create-lockfile -- .
```

I get what I expect to get, though I would accept it if No tests executed was not given.
```bash
Profile: itance (inher)
Version: 0.1.0
Target:  local://

     No tests executed.

Profile: world (hello)
Version: 0.1.0
Target:  local://

  ✔  c1: hello
     ✔  HELLO should cmp == "HELLO"


Profile Summary: 1 successful control, 0 control failures, 0 controls skipped
Test Summary: 1 successful, 0 failures, 0 skipped
```

Now if I expand my profile with stage.rb
inheritance
|-- controls
   |-- dev.rb
   |-- stage.rb
|-- inspec.yml
|-- profile_group_1
   |-- controls
      |-- c1.rb
   |-- inspec.yml

```ruby
# stage.rb
require_controls 'group_1' do
  control 'c1'
end
```

And run the same command:
```bash
Profile: itance (inher)
Version: 0.1.0
Target:  local://

     No tests executed.

Profile: world (hello)
Version: 0.1.0
Target:  local://

  ✔  c1: hello
     ✔  HELLO should cmp == "HELLO"
     ✔  HELLO should cmp == "HELLO"


Profile Summary: 1 successful control, 0 control failures, 0 controls skipped
Test Summary: 2 successful, 0 failures, 0 skipped
```

This is only somewhat expected.  And acceptable, that c1 is ran twice but with two results.  I could go either way: 
1. One result
2. Multiple results but the results are nested under the parent's results.   This would be better especially if one control uses the child control with no enhancements and the other uses:
  ```ruby
[...]

PR #5487, opened 2021-04-27
Fix  undefined method `+' for nil:NilClass\n\nProfile: - when using profile dependencies and require_controls

Vasu1105:

## Description

While using the dependencies if we add one more profile inside controls and use require_controls the InSpec scan breaks.

Example
```
inheritance
|-- controls
   |-- dev.rb
   |-- stage.rb
|-- inspec.yml
|-- profile_group_1
   |-- controls
      |-- c1.rb
   |-- inspec.yml
```
# dev.rb
```
require_controls 'group_1' do
  control 'c1'
end
```
# stage.rb
```
require_controls 'group_1' do
  control 'c1'
end
```
# inspec.yml
```
name: inheritance
version: 0.1.0
depends:
  - name: group_1
     path: ./profile_group_1
```
# c1.rb
```
control 'c1' do
  title 'hello'
  desc 'world'
  impact 0.5
  describe 'HELLO' do
    it { should cmp 'HELLO' }
  end
end
```

After running the below command 

```
inspec exec --no-create-lockfile
```

Following failure message is shown

```
[2021-04-27T15:11:06+05:30] ERROR: Failed to load profile inheritance: Failed to load source for controls/stage.rb: undefined method `+' for nil:NilClass

Profile:         InSpec Profile (inheritance)
Version:         0.1.0
Failure Message: Failed to load source for controls/stage.rb: undefined method `+' for nil:NilClass
Target:          local://

     No tests executed.
```
We can see the failing test here due to above error here https://buildkite.com/chef-oss/inspec-inspec-master-verify/builds/3010#b881f718-2a22-4cfd-965d-cf497306ced9/750-766

## Related Issue
This resolves small partt of #4318

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -91,7 +91,7 @@ module Inspec
     end
 
     def all_controls
-      ret = @rules.values.reject(&:nil?)
+      ret = @rules.values.compact
       ret += @control_subcontexts.map(&:all_rules).flatten
       ret
     end
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -91,7 +91,7 @@ module Inspec
     end
 
     def all_controls
-      ret = @rules.values
+      ret = @rules.values.reject(&:nil?)
       ret += @control_subcontexts.map(&:all_rules).flatten
       ret
     end
--- a/test/fixtures/profiles/dependencies/require_controls_test/controls/example1.rb
+++ b/test/fixtures/profiles/dependencies/require_controls_test/controls/example1.rb
@@ -1,0 +1,3 @@
+require_controls 'profile_b' do
+  control 'profileb-2'
+end


lib/inspec/profile_context.rb
```diff
@@ -91,7 +91,7 @@ def remove_rule(id)
     end
 
     def all_controls
-      ret = @rules.values
+      ret = @rules.values.reject(&:nil?)
```
Nik08: Seems like we could have also used `@rules.values.compact` here in this case? Even using reject will not have any implications, it's the same. It's just `compact` appears more neat. Just my opinion. What are your thoughts?

Nik08 (commented):
@Vasu1105 Is it also possible to add a test case for testing this profile dependency?

lib/inspec/profile_context.rb
Vasu1105: That can be done. Thanks @Nik08

clintoncwolfe (approved):

Vasu1105:
@Nik08 I added the additional profile for that and already existing test case is capturing that.

Nik08:
@Vasu1105 Oh because of the additional profile, I thought there will be some additional test cases. Thanks for the clarification :)

Issue #5157, opened 2020-07-21
alias "inspec compliance" with "inspec automate"


mattray:
Users looking for working with Automate for uploading/downloading/listing profiles have a hard time finding the `inspec compliance` command as the proper command. Conversely, there is now a(nother) Chef Compliance product further confusing the naming. I'd recommend simply providing the alias `inspec automate` for `inspec compliance` and deprecating `compliance` for removal in InSpec 5.

Aha! Link: https://chef.aha.io/features/SH-2656

james-stocks:
cc @kekaichinose 
We should also search the codebase + docs for references to "compliance" (the EOL Chef product), and remove them.

Vasu1105:
Fixed in #5490

PR #5490, opened 2021-04-30
Added alias command `automate` for `inspec compliance`

Nik08:

Added new command alias `automate` for InSpec compliance command 

## Description

Compliance command `automate` introduced for compliance commands, which works similarly to `inspec compliance`.
## Related Issue

Implements #5157 

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,8 @@ $ inspec --help
 Commands:
   inspec archive PATH                                       # archive a profile to tar.gz (default) ...
   inspec check PATH                                         # verify all tests at the specified PATH
-  inspec compliance SUBCOMMAND or automate SUBCOMMAND ...   # Chef Compliance commands
+  inspec automate SUBCOMMAND ...                            # Chef Automate commands
+  inspec compliance SUBCOMMAND ...                          # Chef Automate commands (backwards compatible alias)
   inspec detect                                             # detect the target OS
   inspec exec PATH(S)                                       # run all test files at the specified PATH.
   inspec help [COMMAND]                                     # Describe available commands or one spe...
--- a/README.md
+++ b/README.md
@@ -177,17 +177,17 @@ You should now be able to run:
 ```bash
 $ inspec --help
 Commands:
-  inspec archive PATH                # archive a profile to tar.gz (default) ...
-  inspec check PATH                  # verify all tests at the specified PATH
-  inspec compliance SUBCOMMAND ...   # Chef Compliance commands
-  inspec detect                      # detect the target OS
-  inspec exec PATH(S)                # run all test files at the specified PATH.
-  inspec help [COMMAND]              # Describe available commands or one spe...
-  inspec init TEMPLATE ...           # Scaffolds a new project
-  inspec json PATH                   # read all tests in PATH and generate a ...
-  inspec shell                       # open an interactive debugging shell
-  inspec supermarket SUBCOMMAND ...  # Supermarket commands
-  inspec version                     # prints the version of this tool
+  inspec archive PATH                                       # archive a profile to tar.gz (default) ...
+  inspec check PATH                                         # verify all tests at the specified PATH
+  inspec compliance SUBCOMMAND or automate SUBCOMMAND ...   # Chef Compliance commands
+  inspec detect                                             # detect the target OS
+  inspec exec PATH(S)                                       # run all test files at the specified PATH.
+  inspec help [COMMAND]                                     # Describe available commands or one spe...
+  inspec init TEMPLATE ...                                  # Scaffolds a new project
+  inspec json PATH                                          # read all tests in PATH and generate a ...
+  inspec shell                                              # open an interactive debugging shell
+  inspec supermarket SUBCOMMAND ...                         # Supermarket commands
+  inspec version                                            # prints the version of this tool
 
 Options:
   [--diagnose], [--no-diagnose]  # Show diagnostics (versions, configurations)
--- a/dev-docs/compliance.md
+++ b/dev-docs/compliance.md
@@ -2,13 +2,13 @@
 
 ## Purpose
 
-The `compliance` set of subcommands handle user-initiated communication with Chef Automate. The commands are provided so that a user can interact with an Automate installation.
+The `automate` set of subcommands handle user-initiated communication with Chef Automate. The commands are provided so that a user can interact with an Automate installation.
 
-`inspec compliance` is somewhat analogous to `knife` in that it can be used to upload, download, and manage profiles for distribution to other clients.
+`inspec automate` is somewhat analogous to `knife` in that it can be used to upload, download, and manage profiles for distribution to other clients.
 
-When Automate initiates scans, the `compliance` subcommand is not used.
+When Automate initiates scans, the `automate` subcommand is not used.
 
-`inspec automate` is an alias for `inspec compliance` and works the same way
+`inspec compliance` is a backwards compatible alias for `inspec automate` and works the same way
 
 ## Operational Notes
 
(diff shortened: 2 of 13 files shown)


docs-chef-io/content/inspec/cli.md
```diff
@@ -187,6 +187,10 @@ Chef Automate:
 inspec compliance login
 inspec exec compliance://username/linux-baseline
 ```
+An alternate command for login:
+```
+inspec automate login
+```
```
IanMadd:
````suggestion

`inspec automate` is an alias for `inspec compliance` and works the same way:

```
inspec automate login
```
````

File: docs-chef-io/content/inspec/plugin_kitchen_inspec.md [hunk 2]
```diff
@@ -40,6 +40,11 @@ inspec compliance login https://compliance.test --user admin --insecure --token
 
 where `--insecure` is required when using self-signed certificates.
 
+An alternate command for login:
+```bash
+inspec automate login https://compliance.test --user admin --insecure --token ''
+```
+
```
IanMadd: Someone correct me if I'm wrong, but the reference to Compliance Server starting on line 23 needs to be updated for Automate. I'm not sure if there's a problem with the yaml and bash examples.

````suggestion
`inspec automate` is an alias for `inspec compliance` and works the same way:

```bash
inspec automate login https://compliance.test --user admin --insecure --token ''
```

````

**Review by clintoncwolfe — COMMENTED** [review]
Offhand, the implementation looks OK. However, I am concerned that presenting the `automate` command as an alternative to the `compliance` command is confusing. I wonder if we should do a soft deprecation, presenting it as a replacement while still continuing to support the old command.

**Comment by Nik08** [comment]
@clintoncwolfe So by soft deprecation you do mean that we should show the deprecation warning, right? And yes we are supporting the old command anyway so that will continue if we go for soft deprecation too. 
I personally feel deprecating the old command would be a better experience. Nevertheless, we can take this up with the right team to decide this.

Though, if we present it as an alias rather than an alternative in the docs, does it still seem confusing? 

Also, one of my concern is with the error messages, that's where I need some inputs. Because 'compliance or automate', this part of the error messages does not seem right. You can refer this in target.rb & login.rb in this PR.

File: docs-chef-io/content/inspec/plugin_kitchen_inspec.md (hunk 2 above)
Comment (Nik08) [reply docs-chef-io/content/inspec/plugin_kitchen_inspec.md]: @IanMadd To me it seems all okay. @clintoncwolfe Can you validate this?

File: README.md [hunk 3]
```diff
@@ -177,17 +177,17 @@ You should now be able to run:
...
-  inspec exec PATH(S)                # run all test files at the specified PATH.
-  inspec help [COMMAND]              # Describe available commands or one spe...
-  inspec init TEMPLATE ...           # Scaffolds a new project
-  inspec json PATH                   # read all tests in PATH and generate a ...
-  inspec shell                       # open an interactive debugging shell
-  inspec supermarket SUBCOMMAND ...  # Supermarket commands
-  inspec version                     # prints the version of this tool
+  inspec archive PATH                                       # archive a profile to tar.gz (default) ...
+  inspec check PATH                                         # verify all tests at the specified PATH
+  inspec compliance SUBCOMMAND or automate SUBCOMMAND ...   # Chef Compliance commands
```
clintoncwolfe: List this as:
```
inspec automate SUBCOMMAND # Chef Automate commands
...
inspec compliance SUBCOMMAND # Chef Automate commands (backwards compatible alias)
```

dev-docs/compliance.md
```diff
@@ -65,9 +67,10 @@ There are several other minor commands not listed here - see `lib/cli.rb` for a
 
 ### login
 
-Saves a credentials file locally. Future invocations of `inspec compliance` use the credentials file to authenticate.
+Saves a credentials file locally. Future invocations of `inspec compliance` or `inspec automate` use the credentials file to authenticate.
```
clintoncwolfe: Throughout, list `inspec automate` first. Think of it as being the default command, and compliance as being the backwards compatible alias (even though that is not the implementation).

dev-docs/compliance.md
```diff
@@ -65,9 +67,10 @@ There are several other minor commands not listed here - see `lib/cli.rb` for a
...
 
-`be inspec compliance login --user=admin --token='1234567890asdfghjkl' --insecure https://chef-automate.test`
+`be inspec compliance login --user=admin --token='1234567890asdfghjkl' --insecure https://chef-automate.test` or
+`be inspec automate login --user=admin --token='1234567890asdfghjkl' --insecure https://chef-automate.test`
```
clintoncwolfe: List automate first.

docs-chef-io/content/inspec/cli.md
```diff
@@ -188,6 +188,12 @@ inspec compliance login
 inspec exec compliance://username/linux-baseline
 ```
 
+`inspec automate` is an alias for `inspec compliance` and works the same way:
```
clintoncwolfe: List automate before compliance, and present compliance as being a backwards-compatible alias.

docs-chef-io/content/inspec/plugin_kitchen_inspec.md
```diff
@@ -40,6 +40,12 @@ inspec compliance login https://compliance.test --user admin --insecure --token
 
 where `--insecure` is required when using self-signed certificates.
 
+`inspec automate` is an alias for `inspec compliance` and works the same way:
```
clintoncwolfe: List automate before compliance, and present compliance as being a backwards-compatible alias.

clintoncwolfe (commented):
I don't think we want to issue a deprecation warning at this point. Mainly I want the wording to be changed so that it appears that automate is the default, and compliance is merely a backwards-compatible alias.

dev-docs/compliance.md
Nik08: Got it. Will do so. Thanks!

lib/plugins/inspec-compliance/lib/inspec-compliance/target.rb
```diff
@@ -34,13 +34,13 @@ def self.check_compliance_token(uri, config)
         if config["token"].nil? && config["refresh_token"].nil?
           if config["server_type"] == "automate"
             server = "automate"
-            msg = "#{EXEC_NAME} compliance login https://your_automate_server --user USER --ent ENT --dctoken DCTOKEN or --token USERTOKEN"
+            msg = "#{EXEC_NAME} compliance or automate login https://your_automate_server --user USER --ent ENT --dctoken DCTOKEN or --token USERTOKEN"
```
clintoncwolfe: msg = "#{EXEC_NAME} [automate|compliance] login ...

IanMadd (approved):

lib/plugins/inspec-compliance/lib/inspec-compliance/cli.rb
```diff
@@ -6,8 +6,7 @@ module InspecPlugins
   module Compliance
     class CLI < Inspec.plugin(2, :cli_command)
       include Inspec::Dist
-
-      subcommand_desc "compliance SUBCOMMAND", "#{COMPLIANCE_PRODUCT_NAME} commands"
+      subcommand_desc "automate SUBCOMMAND or compliance SUBCOMMAND", "#{COMPLIANCE_PRODUCT_NAME} commands"
 
       # desc "login https://SERVER --insecure --user='USER' --ent='ENTERPRISE' --token='TOKEN'", 'Log in to a Chef Compliance/Chef Automate SERVER'
       desc "login", "Log in to a #{COMPLIANCE_PRODUCT_NAME}/#{AUTOMATE_PRODUCT_NAME} SERVER"
```
clintoncwolfe: When you run `inspec automate help`, you see the word `Compliance Server` a lot. I think that is fairly nonsensical now.  Those occurences should be replaced with AUTOMATE_PRODUCT_NAME.

Here, just eliminate the alternative:
`Log in to a #{AUTOMATE_PRODUCT_NAME} SERVER`

clintoncwolfe: But in each of the command `desc`, let's replace COMPLIANCE_PRODUCT_NAME with AUTOMATE_PRODUCT_NAME.

clintoncwolfe (commented):
I've done functional testing on this and it works well. My only concern is the help output from `inspec automate help` continuing to refer to Compliance Server.

PR #5506, opened 2021-05-05
Modified windows_feature to indicate enabled rather than just available

jwdean:
In the context of this resource, there are three conditions a windows feature should be tested for as represented by DISM:

1. Available in the OS and enabled
       dism returns 0 and dism output of state = 'Enabled'
2. Available in the OS but disabled
       dism returns 0 and dism output of state != 'Enabled'
3. Not available  
       dism returns != 0 

This change, in effect, modifies the meaning of installed == true to represent that the feature is both available and enabled. 
installed == false will now indicate the feature is either unavailable or disabled.
The prior implementation indicated the feature was available, but did not recognize the it might be disabled at that time. The most logical use for an InSpec evaluation of a Windows Feature seems to be determining if the feature is enabled or disabled rather than simply available. This updated implementation better represents that goal. 

This change also makes the behavior and meaning of "installed" in both the InSpec and Chef windows_feature resoruces consistent. In Chef's windows_feature resource (really windows_feature_dism), when the feature is available the :install action performs 'dism.exe /online /enable-feature' and the :remove action performs 'dism.exe /online /disable-feature'. The Chef resource throws an error if the feature is not available.

----------

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the
    best of my knowledge, is covered under an appropriate open
    source license and I have the right under that license to
    submit that work with modifications, whether created in whole
    or in part by me, under the same open source license (unless
    I am permitted to submit under a different license), as
    Indicated in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including
    all personal information I submit with it, including my
    sign-off) is maintained indefinitely and may be redistributed
    consistent with this project or the open source license(s)
    involved.

Signed-off-by: Jeff Dean <jwdean@scriptpro.com>

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/windows_feature.rb
+++ b/lib/inspec/resources/windows_feature.rb
@@ -79,10 +79,11 @@ module Inspec::Resources
         result = cmd.stdout
         feature_name_regex = /Feature Name : (.*)(\r\n|\n)/
         description_regex = /Description : (.*)(\r\n|\n)/
+        state_regex = /State : (.*)(\r\n|\n)/
         feature_info = {
           name: result.match(feature_name_regex).captures[0].chomp,
           description: result.match(description_regex).captures[0].chomp,
-          installed: true,
+          installed: result.match(state_regex).captures[0].chomp == 'Enabled',
         }
       end
 


clintoncwolfe (approved):
Thanks for this!

PR #5517, opened 2021-05-10
Added Common Errors page doc

### LINKED ISSUE #5477: Make a Common Errors docs page
Make a docs page that lists common error messages and how to deal with them.

Errors to take in count (initial list, so only one item at the moment. This will be updated as more of them are finalised before the work starts on this document)
1. #5475

Nik08:

## Description

Added Common Errors page doc ( with only 1 error for now to get started )
## Related Issue

Closes #5477 
## Types of changes

## Checklist:


docs-chef-io/content/inspec/common_errors.md
```diff
@@ -0,0 +1,27 @@
...
+
+* `Undefined local variable or method error for 'aws_eks_clusters' resource`
+
+    Syntax of using AWS resource in InSpec profile to test eks clusters:
+
+    ```bash
+        describe aws_eks_clusters.where( failed: true ) do
+            it { should_not exist }
+        end
+    ```
```
clintoncwolfe: Perhaps add a note that this error can occur with any of several AWS resources, not just aws_eks_clusters.

IanMadd: Can this happen with any of the resource packs, just AWS, or is it specific AWS resources?

Nik08: @IanMadd This can happen with any AWS resource. I am gonna mention that here now.

IanMadd: Couldn't I get the same error with GCP or Azure though if I didn't create a profile for those resource packs and then called one of their resources?

We should also have the instructions for creating a profile in docs.chef.io. There's instructions for GCP, but not the others for some reason. https://docs.chef.io/inspec/platforms/#create-an-inspec-profile-that-makes-use-of-inspec-gcp

Nik08: That makes sense. Let me think about someway I can include this fact for all the cloud resources because I was relying on an error to be mentioned on this page and then explain whats and whys around that.

And yes instructions page for other cloud resources is a good idea again. @clintoncwolfe should the documentations for profile creation using cloud resources be covered by cloud resource pack team or could we also pick this up in future?

IanMadd:
I made some edits. I moved the page from common errors to troubleshooting, formatted the error as a heading, and then listed all the resource pack readmes.

PR #5519, opened 2021-05-11
Changes returns nil on file non-existence through matcher `more_permissive_than`

### LINKED ISSUE #4891: File resource `more_permissive_than` matcher file existence check inconsistent
When the file does not exist and the `more_permissive_than` matcher is used, an exception and backtrace is generated and written to the reporter JSON.  This is inconsistent with all the other parts of the `file` resource, none of which also do this sort of check. This is taken care of by the resource itself. It caused some confusing behavior in my tools which check the JSON's backtrace attributes for exceptions and interprets it's existence as an error with the control. I don't think this check should exist.

https://github.com/inspec/inspec/blob/40966bc0393131ec4e7d0ec43f7648d66db5cf0f/lib/inspec/resources/file.rb#L139

Nik08:

## Description

Returns nil on file non-existence through matcher `more_permissive_than`
## Related Issue

Fixes #4891 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/file.rb
+++ b/lib/inspec/resources/file.rb
@@ -136,10 +136,10 @@ module Inspec::Resources
     alias sticky? sticky
 
     def more_permissive_than?(max_mode = nil)
-      raise Inspec::Exceptions::ResourceFailed, "The file" + file.path + "doesn't seem to exist" unless exist?
-      raise ArgumentError, "You must proivde a value for the `maximum allowable permission` for the file." if max_mode.nil?
-      raise ArgumentError, "You must proivde the `maximum permission target` as a `String`, you provided: " + max_mode.class.to_s unless max_mode.is_a?(String)
-      raise ArgumentError, "The value of the `maximum permission target` should be a valid file mode in 4-ditgit octal format: for example, `0644` or `0777`" unless /(0)?([0-7])([0-7])([0-7])/.match?(max_mode)
+      return nil unless exist?
+      raise ArgumentError, "You must provide a value for the `maximum allowable permission` for the file." if max_mode.nil?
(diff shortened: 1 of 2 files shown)


lib/inspec/resources/file.rb
```diff
@@ -136,10 +141,10 @@ def sticky
     alias sticky? sticky
 
     def more_permissive_than?(max_mode = nil)
-      raise Inspec::Exceptions::ResourceFailed, "The file" + file.path + "doesn't seem to exist" unless exist?
-      raise ArgumentError, "You must proivde a value for the `maximum allowable permission` for the file." if max_mode.nil?
-      raise ArgumentError, "You must proivde the `maximum permission target` as a `String`, you provided: " + max_mode.class.to_s unless max_mode.is_a?(String)
-      raise ArgumentError, "The value of the `maximum permission target` should be a valid file mode in 4-ditgit octal format: for example, `0644` or `0777`" unless /(0)?([0-7])([0-7])([0-7])/.match?(max_mode)
+      return false unless exist?
```
clintoncwolfe: I think this should return nil, not false.

lib/inspec/resources/file.rb
```diff
@@ -40,6 +40,11 @@ def initialize(path)
       # select permissions style
       @perms_provider = select_file_perms_style(inspec.os)
       @file = inspec.backend.file(path)
+      validate_file_existence
+    end
+
+    def validate_file_existence
+      raise Inspec::Exceptions::ResourceFailed, "The file" + file.path + " doesn't seem to exist" unless exist?
```
clintoncwolfe: This change will break a lot of things. Most tests in inspec are designed to check separately for existence. I would remove this change.

PR #5522, opened 2021-05-13
Fix for port resource performance: adding more specific search while using ss command

### LINKED ISSUE #4965: Make use of filter query where testing a single port with ss
Evaluating all [port](https://github.com/inspec/inspec/blob/master/lib/inspec/resources/port.rb data on the target host can take a long time - on some customer systems it might exceed 10 minutes.

Short term, we should at least be filtering the query made to the target host when a single port is being tested (`describe port(80)`) and where the underlying command is `ss`

i.e. 
`ss -tulpen "( dport = 80 or sport = 80 )"

Vasu1105:

## Description

Currently port InSpec resource is using `ss -tulpen` command to get the information about the listening ports which basically gives all ports result as output which can be reduce if we query on the specific port by using more filter option as mentioned in the issue #4965 

## Related Issue

Fix #4965
## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #5530, opened 2021-05-17
Bugfix for `inspec detect --no-color` to not return colourful output

### LINKED ISSUE #4505: `inspec detect --no-color` is too colorful
Subject says it all. It outputs color codes regardless.

Found while pairing with @miah.

Nik08:

## Description

Fix for `inspec detect --no-color` to not return colourful output
## Related Issue

Fixes #4505 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -305,7 +305,12 @@ class Inspec::InspecCLI < Inspec::BaseCLI
       puts res.to_json
     else
       ui.headline("Platform Details")
-      ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0, color: 36)
+
+      if ui.color?
+        ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0, color: 36)
+      else
+        ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0)
+      end
     end
   rescue ArgumentError, RuntimeError, Train::UserError => e
     $stderr.puts e.message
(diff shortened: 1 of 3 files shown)


test/functional/ui_test.rb
```diff
@@ -130,6 +139,15 @@ def show_spaces(str)
         assert_exit_code 0, run_result
       end
     end
+
+    describe "detect command" do
+      let(:result) { inspec("detect --no-color") }
+
+      it "has no color in the output" do
+        _(result.stdout).must_include("\e[1m\e[0m")
```
clintoncwolfe: I think the test should be that it does not include `"\e[1m` - in other words, ther should be no color control codes in the output.

lib/inspec/cli.rb
```diff
@@ -305,7 +305,9 @@ def detect
       puts res.to_json
     else
       ui.headline("Platform Details")
-      ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0, color: 36)
+
+      detect_cmd_ui_color = ui.color? ? 36 : 0
+      ui.plain Inspec::BaseCLI.format_platform_info(params: res, indent: 0, color: detect_cmd_ui_color)
```
clintoncwolfe: I think `Inspec::BaseCLI.format_platform_info` needs to be changed to avoid outputting color codes entirely when color == 0.

test/functional/ui_test.rb
Nik08: That's right. I confused no color with reset. Will make this change.

clintoncwolfe (approved):

PR #5536, opened 2021-05-21
Update inspec init plugin 

### LINKED ISSUE #5074: Update init plugin
# Describe the problem

The `inspec init plugin` templates could use some love.

1. They use single quotes. chefstyle demands double.
2. Template files should end in .erb
3. Should place the plugin def in lib/<loaderfile>.rb, not plugin.rb
4. Document init plugin in the plugin dev guide
5. Rename --hook to --activator, but leave an alias as --hook for backwards compatibility

Aha! Link: https://chef.aha.io/features/SH-2179

Vasu1105:

## Description

Updated inspec init plugin files 
1. To use double quotes as per chefstyle
2. Template files renamed to .erb
3. Rename --hook to --activators, but leave an alias as --hook for backwards compatibility

Note: There is not clarification about point 3 from the #5074 so not implemented that in this PR,
## Related Issue
Fix #5074

## Types of changes

## Checklist:


etc/deprecations.json
```diff
@@ -120,6 +120,11 @@
     "object_classes": {
       "action": "warn",
       "suffix": "These classes will be removed in InSpec 5.0."
+    },
+    "cli_option_hook":{
+      "action": "warn",
```
Nik08, line 125: Just a confusion, are we deprecating this option? Or just renaming it?

Vasu1105: We are renaming it and so deprecating it in the future. So this will warn user to use the renamed one as it will get removed.

Nik08:
LGTM!

Vasu1105:
> LGTM!

Thanks @Nik08

clintoncwolfe (approved):
I tested it and it works fine - thanks!

PR #5551, opened 2021-06-02
Fix mysql_session resource to raise exception if there is a error in connection or in query

### LINKED ISSUE #5309: SQL session resources should throw an error or warning if you try to execute a query but there is no valid session established
At the moment - if for some reason the sql session of the database resources do not make it clear that you are trying to query a nil sql session and it is cauing invalid results.

We should add some logic in the session and the query resources to protect aginst this case.

Vasu1105:

## Description

Fix mysql_session resource to handle the exceptions if the mysql session is not established. Currently it does not raise any exception due to which the InSpec test gives false result
## Related Issue

Fixes #5309 for mysql_session resource.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/mysql_session.rb
+++ b/lib/inspec/resources/mysql_session.rb
@@ -45,11 +45,13 @@ module Inspec::Resources
       @socket = socket
       init_fallback if user.nil? || pass.nil?
       raise Inspec::Exceptions::ResourceFailed, "Can't run MySQL SQL checks without authentication." if @user.nil? || @pass.nil?
+
       set_connection
     end
 
     def query(q, db = "")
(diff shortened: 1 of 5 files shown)


lib/inspec/resources/mysql_session.rb
```diff
@@ -68,6 +72,12 @@ def to_s
 
     private
 
+    # Querying on the database to make sure conneciton can be established. If not this will set the resource exception
+    # message which we raise before querying on the database using mysql_session object.
+    def set_connection
```
clintoncwolfe: I suggest we rename this to be "test connection" and use a query like "SELECT 1" - some users might not have permission to list databases.

clintoncwolfe (approved):
Looks good, waiting to merge until after tomorrow's release!

PR #5553, opened 2021-06-03
Fix postgres_session resource to raise exception if there is a error in connection or in query

### LINKED ISSUE #5309: SQL session resources should throw an error or warning if you try to execute a query but there is no valid session established

Vasu1105:

## Description

Fix postgres_session resource to raise exception if database connection is not established due to any reason or there is error with query. Currently it does not raise any exception due to which the InSpec test gives false result.

## Related Issue

Fixes #5309 for postgres_session resource
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/postgres_session.rb
+++ b/lib/inspec/resources/postgres_session.rb
@@ -51,7 +51,7 @@ module Inspec::Resources
     end
 
     def query(query, db = [])
-      raise Inspec::Exceptions::ResourceFailed, "#{resource_exception_message}" if self.resource_failed?
+      raise Inspec::Exceptions::ResourceFailed, "#{resource_exception_message}" if resource_failed?
 
       psql_cmd = create_psql_cmd(query, db)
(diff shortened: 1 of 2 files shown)


lib/inspec/resources/postgres_session.rb
```diff
@@ -45,21 +45,30 @@ def initialize(user, pass, host = nil, port = nil)
...
-        Lines.new(out, "PostgreSQL query with errors: #{query}")
+        raise Inspec::Exceptions::ResourceFailed, "PostgreSQL query with errors: #{out}"
       else
         Lines.new(cmd.stdout.strip, "PostgreSQL query: #{query}")
       end
     end
 
     private
 
+    def set_connection
```
clintoncwolfe: Let's rename this to be test_connection and use a more generic query like SELECT 1

clintoncwolfe (approved):
Looks good, will merge after release tomorrow.

PR #5562, opened 2021-06-14
Add support for mongodb_conf resource in InSpec

### LINKED ISSUE #5540: Add resources for Mongo database support
## Detailed Description
(i)Use the mongodb_session Chef InSpec audit resource to test json queries run against a Mongodb.

(ii)Use the mongodb_ conf Chef InSpec audit resource to test the contents of the configuration file for Mongodb, typically located at
/etc/mongod.conf
/etc/mongos.conf

## Context
Needed for CIS benchmark and customer needs.

Vasu1105:

## Description

Added support for `mongodb_conf` resource based on MongoDB docs and CIS benchmarks.
## Related Issue

Part Fix #5540
## Types of changes

## Checklist:


lib/inspec/resources/mongodb.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def version_from_dir(dir)
+      dirs = inspec.command("Get-ChildItem -Path \"#{dir}\" -Name").stdout
+      entries = dirs.lines.count
+      case entries
+      when 0
+        warn "Could not determine version of installed MongoDB by inspecting #{dir}"
+        nil
+      when 1
+        warn "Using #{dirs}: #{dir_to_version(dirs)}"
```
clintoncwolfe: This appears to be the happy case - we should not issue a warning here.

lib/inspec/resources/mongodb_conf.rb
```diff
@@ -0,0 +1,43 @@
...
+      if @conf_path.nil?
+        return skip_resource "MongoDB conf path is not set."
+      end
+
+      super(@conf_path)
+    end
+
+    def port
+      params["net"]["port"]
+    end
```
clintoncwolfe: Why make a special method for port?

Vasu1105: I thought that if anyone just wants to check port it will be easier. So added that. We can remove it if not needed. Should I remove that?

lib/inspec/resources/mongodb.rb
Vasu1105: So should we just return nil?

clintoncwolfe (approved):
Looks good, thanks!

PR #5567, opened 2021-06-16
Oracle Session Exception Handling

Nik08:

Exception handling in case the connection is not properly established with oracle database server
## Description

- Exception handling in case the connection is not properly established with oracle database server
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -87,11 +87,11 @@ module Inspec::Resources
       end
 
       if @db_role.nil?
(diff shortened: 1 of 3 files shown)


Nik08:
@clintoncwolfe @Vasu1105 I need your verification and inputs on this [commit](https://github.com/inspec/inspec/pull/5567/commits/88641c3fac0f0862a65c11d5eb3339799b8824af).
I was stuck on this for quite a while and it wasn't working successfully for windows because the syntax was wrong for power shell when used with escape `/` character. But my change made it work for both linux and windows. I have tested this on oracle 18c version and have verified this with oracle 11 doc too.. Just let me know your thoughts around this. I am hoping it does not crash on any other version of oracle.

clintoncwolfe (approved):

PR #5584, opened 2021-06-30
Remove default port for mssql_session, allowing named connections

### LINKED ISSUE #5548: MSSQL Named Instances doesn't work with current code
This issue is here for almost 2 years.
Named instances will not work without port. And MSSQL is capable of established that connection.

you can find the fix:
https://github.com/da3210/inspec/blob/dc/fixmssqlsession/lib/inspec/resources/mssql_session.rb

could you do a PR?

_Originally posted by @da3210 in https://github.com/inspec/inspec/issues/4319#issuecomment-852022224_

Nik08:

MSSQL resource fix for named instances (used with no port)
## Description

In current code, the default port was 1433 and so named instances had issue running without any port. So this change will enable using named instances without any port as well.
## Related Issue

Fixes #5548 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/mssql_session.rb
+++ b/lib/inspec/resources/mssql_session.rb
@@ -42,11 +42,7 @@ module Inspec::Resources
       @local_mode = opts[:local_mode]
       unless local_mode?
         @host = opts[:host] || "localhost"
-        if opts.key?(:port)
-          @port = opts[:port]
-        else
-          @port = "1433"
-        end
+        @port = opts[:port]
       end
       @instance = opts[:instance]
       @db_name = opts[:db_name]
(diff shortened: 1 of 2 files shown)


lib/inspec/resources/mssql_session.rb
```diff
@@ -42,11 +42,7 @@ def initialize(opts = {})
       @local_mode = opts[:local_mode]
       unless local_mode?
         @host = opts[:host] || "localhost"
-        if opts.key?(:port)
-          @port = opts[:port]
-        else
-          @port = "1433"
-        end
+        @port = opts[:port]
```
clintoncwolfe, line 45: What happens if someone neither specifies a port nor uses a named connection? In that case, shouldn't we default to a port?

Nik08: In that case also it works. It does not crash. MSSQL seems to establish connection even in that case. In my opinion it would work even without default port..but do you suggest we have the default port in case named instance is not used?

clintoncwolfe: OK, if you've tested it without a port, it seems fine to me.

PR #5596, opened 2021-07-09
Filter active controls in profile by tags

### LINKED ISSUE #5447: Filter Active Controls in Profile by Tag
## Detailed Description

Following up on inspec/inspec#5215, it would be particularly helpful for our use case not only to filter controls individual by their identifier, but groups of controls by tag. In short, enable similar functionality for `--control=foo`, but with `--tag=bar` and zero, one, or more controls with tag `bar` in its `tag` block(s) in that given profile will execute.

## Context

US federal agencies and detailed contractors must meet compliance objectives as defined by government compliance standards, particularly the NIST Risk Management Framework. Grouping one or more control(s) by `tag` block would be very useful and not require building logic outside of `inspec` to filter the tags "by hand." This will allow DevSecOps engineers to map security tests to high-level compliance objectives with less effort.

Example from MITRE:

https://github.com/mitre/redhat-enterprise-linux-7-stig-baseline/blob/master/controls/V-71849.rb#L66

## Possible Implementation

TBD, I will draft a PR, but likely follow the same implementation strategy as the `--controls=foo` sub-command pattern.

Nik08:

Filtering active controls in profile by tags.

## Description

Similar to the feature of filtering controls using `--controls`, a new option `--tags` was required which could be used for passing the tags names of controls and filtering them. 

### Solution
1. `--tags` option is introduced.
2. Fetched control tags using method [control_tags](https://github.com/inspec/inspec/blob/aeed833f2f85124a744ddee4d3848255b5edb00f/lib/inspec/control_eval_context.rb#L64).
3. Filtered controls by using tags that are passed through option `--tags` to match with control tags using method [tag_exist_in_control_tags?](https://github.com/inspec/inspec/blob/8e755063c5bf71b089e710e1a3cae466e6534a1b/lib/inspec/control_eval_context.rb#L229)

### Needs Suggestion (Code Refactoring)

The method introduced to [get controls tags](https://github.com/inspec/inspec/blob/aeed833f2f85124a744ddee4d3848255b5edb00f/lib/inspec/control_eval_context.rb#L64) is using block of control which is in string format. Tags are fetched from this string using regex and string manipulation. 

Though it works successfully on all the cases of ways and formats a tag can be used, as used in this [testing control](https://github.com/inspec/inspec/blob/8e755063c5bf71b089e710e1a3cae466e6534a1b/test/fixtures/profiles/control-tags/controls/example.rb). A better solution could be easier to understand.

## Related Issue

Closes #5447 
## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/cli.md
+++ b/docs-chef-io/content/inspec/cli.md
@@ -272,6 +272,8 @@ This subcommand has additional options:
     Read configuration from JSON file (`-` reads from stdin).
 * ``--controls=one two three``
     A list of control names to run, or a list of /regexes/ to match against control names. Ignore all other tests.
+* ``--tags=one two three``
+    A list of tags names that are part of controls to filter and run controls, or a list of /regexes/ to match against tags names of controls. Ignore all other tests. And when tag on the control is a hashmap, it only uses values of hashmap for filtering controls.
 * ``--create-lockfile``, ``--no-create-lockfile``
     Write out a lockfile based on this execution (unless one already exists)
 * ``--distinct-exit``, ``--no-distinct-exit``
@@ -379,6 +381,8 @@ This subcommand has additional options:
 
 * ``--controls=one two three``
     A list of controls to include. Ignore all other tests.
+* ``--tags=one two three``
+    A list of tags to filter controls and include only those. Ignore all other tests.
 * ``-o``, ``--output=OUTPUT``
     Save the created profile to a path
 * ``--profiles-path=PROFILES_PATH``
--- a/docs-chef-io/content/inspec/cli.md
+++ b/docs-chef-io/content/inspec/cli.md
@@ -272,8 +272,6 @@ This subcommand has additional options:
     Read configuration from JSON file (`-` reads from stdin).
 * ``--controls=one two three``
     A list of control names to run, or a list of /regexes/ to match against control names. Ignore all other tests.
-* ``--tags=one two three``
-    A list of tags names that are part of controls to filter and run controls, or a list of /regexes/ to match against tags names of controls. Ignore all other tests. And when tag on the control is a hashmap, it only uses values of hashmap for filtering controls.
 * ``--create-lockfile``, ``--no-create-lockfile``
     Write out a lockfile based on this execution (unless one already exists)
 * ``--distinct-exit``, ``--no-distinct-exit``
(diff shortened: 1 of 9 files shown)


IanMadd:
Can tags be put in alphabetical order with the other items.

docs-chef-io/content/inspec/cli.md
```diff
@@ -272,6 +272,8 @@ This subcommand has additional options:
     Read configuration from JSON file (`-` reads from stdin).
 * ``--controls=one two three``
     A list of control names to run, or a list of /regexes/ to match against control names. Ignore all other tests.
+* ``--tags=one two three``
+    A list of tags names that are part of controls to filter and run controls, or a list of /regexes/ to match against tags names of controls. Ignore all other tests. And when tag on the control is a hashmap, it only uses values of hashmap for filtering controls.
```
IanMadd:
```suggestion
    A list of tags or a list of regular expressions that match tags. `exec` will run controls referenced by the listed or matching tags.
```

docs-chef-io/content/inspec/cli.md
```diff
@@ -379,6 +381,8 @@ This subcommand has additional options:
 
 * ``--controls=one two three``
     A list of controls to include. Ignore all other tests.
+* ``--tags=one two three``
+    A list of tags to filter controls and include only those. Ignore all other tests.
```
IanMadd:
```suggestion
    A list of tags that reference certain controls. Other controls are ignored.
```

Nik08:
> Can tags be put in alphabetical order with the other items.

@IanMadd  I am a little confused can you elaborate this? What do you mean by putting it in alphabetical order?

docs-chef-io/content/inspec/cli.md
Nik08: @IanMadd I feel we should also include in doc that if tags are of type hash then it matches the value of an hash not a key to bring more clarity. I mentioned this sentence for that purpose "And when tag on the control is a hash map, it only uses values of hashmap for filtering controls." Can we put this in more clear way in doc ? What are your suggestions?

IanMadd:
On the cli page all the flags are in alphabetical order except the --tags that you're adding.

docs-chef-io/content/inspec/cli.md
IanMadd: Maybe something like this: 

A list of tags, a list of regular expressions that match tags, or a hash map where each value is a tag. 
`exec` will run controls referenced by the listed or matching tags.

IanMadd (approved):
LGTM

PR #5598, opened 2021-07-14
Fix range based filtering in filter tables

### LINKED ISSUE #5116: FilterTable where filtering with ranges or inequalities
https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md suggests that ranges in FilterTables where methods is possible. This doesn't appear to work as documented.

## Describe the problem
```ruby
  describe things.where(thing_id: (1..2)) do
    its('count') { should cmp 2 }
  end
```
actually returns
```
things.where(thing_id: (1..2))
=> Things with thing_id == 1..2
things.where(thing_id: (1..2)).entries
=> []
```
and does not return the expected two Things with ID 1 and 2. It returns no entries instead.

## Possible Solution
Either remove the example from FilterTable usage docs or fix/implement the behaviour.

Aha! Link: https://chef.aha.io/features/SH-2387

Nik08:

## Description

Fixed range based filtering in filter tables for issue explained below.

**Problem:** 

Issue with range based filtering : 
```
things.where(thing_id: (1..2)).entries
```
where data is 
```
[
   { thing_id: 1, color: :red },
   { thing_id: 2, color: :blue, tackiness: 'very' },
   { thing_id: 3, color: :red },
]
```
Returned [] entries while it should return first 2 entries.

## Related Issue

Fixes #5116 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/utils/filter.rb
+++ b/lib/inspec/utils/filter.rb
@@ -256,7 +256,7 @@ module FilterTable
     end
 
     def matches(x, y)
-      x === y # rubocop:disable Style/CaseEquality
+      y === x # rubocop:disable Style/CaseEquality
     end
 
     def filter_raw_data(current_raw_data, field, desired_value)
--- a/test/unit/utils/filter_table_test.rb
+++ b/test/unit/utils/filter_table_test.rb
@@ -35,6 +35,11 @@ describe FilterTable do
     _(resource.new(nil).where { false }.params).must_equal []
   end
 
+  it "supports range" do
+    factory.add_accessor(:where).connect(resource, :data)
+    _(instance.where({ foo: (3..5) }).params).must_equal [data[0]]
+  end
+
   it "retrieves the resource from all entries" do
     factory.add_accessor(:where)
       .add(:baz?) { |x| x.resource } # rubocop: disable Style/SymbolProc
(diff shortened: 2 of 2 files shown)


Vasu1105 (approved):
LGTM. Thanks @Nik08. Good work !!

Nik08:
Thanks @Vasu1105 :)

lib/inspec/utils/filter.rb
```diff
@@ -256,7 +256,7 @@ def matches_regex(x, y)
     end
 
     def matches(x, y)
-      x === y # rubocop:disable Style/CaseEquality
+      y === x # rubocop:disable Style/CaseEquality
```
clintoncwolfe, line 259: Would you mind explaining the fix?

Nik08: Yes sure. So this method is called when the filter table needs to match anything other than an Integer, float & regex. So this is the `else` condition. And since this one was called in case of **ranges**. It was breaking because I found it was checking initially for example `1 === 1..2` while my change checks for `1..2 === 1` and this returns true because this is how it should be used for this case. I referred [this](https://stackoverflow.com/questions/4467538/what-does-the-operator-do-in-ruby).

PR #5601, opened 2021-07-19
Fix apache_conf issue when Server Root is not present in configuration

### LINKED ISSUE #5583: apache_conf resource fails when ServerRoot is not included in configuration
As describe in the title, in configurations where the ServerRoot is not defined (which is not required by apache) the current way to get the conf_dir throws a SyntaxError. This is very inconvenient and blocks me currently. 

# Describe the problem
Create the following describe block: 

```
describe apache_conf do
  its('TimeOut') { should_not cmp nil }
end
```

This describe block (and any other similar describe block) fails when no ServerRoot is set explicility in the httpd/apache2.conf. This is not required by apache2/httpd. I can omit this value in my configuration and the server will still act normally. 

The failure message is the following: 

```
[2021-06-30T11:09:42+02:00] ERROR: Failed to load profile apache-test: Failed to load source for controls/example.rb: no implicit conversion of nil into String

Profile:         InSpec Profile (apache-test)
Version:         0.1.0
Failure Message: Failed to load source for controls/example.rb: no implicit conversion of nil into String
Target:          ssh://vagrant@10.211.55.124:22

     No tests executed.

Test Summary: 0 successful, 0 failures, 0 skipped

```

I would exepect that this does not prouce a "Source Code Error" and fails the execution of the InSpec profile. 

## Possible Solution

Add a safe default to the conf_dir method or handle the case where this method returns nil when constructing the included files path.

Nik08:
**Description**

### Problem:

Raised following error while running profile using apache_conf in case `ServerRoot` value is not present in the configuration file. 

`Failure Message: Failed to load source for controls/example.rb: no implicit conversion of nil into String`

It was caused because of `nil` value returned by `conf_dir` method which is derived from `ServerRoot` value of conf file.

### Solution

A simple fix of `nil` check solves this problem. But it should be noted that without setting this value in the configuration file, `Include` and `IncludeOptional` params are not accessible to the resource.

## Related Issue

Fixes #5583 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/apache_conf.rb
+++ b/lib/inspec/resources/apache_conf.rb
@@ -101,12 +101,14 @@ module Inspec::Resources
       include_files_optional = params["IncludeOptional"] || []
 
       includes = []
-      (include_files + include_files_optional).each do |f|
-        id    = Pathname.new(f).absolute? ? f : File.join(conf_dir, f)
-        files = find_files(id, depth: 1, type: "file")
-        files += find_files(id, depth: 1, type: "link")
+      unless conf_dir.nil?
+        (include_files + include_files_optional).each do |f|
+          id    = Pathname.new(f).absolute? ? f : File.join(conf_dir, f)
+          files = find_files(id, depth: 1, type: "file")
+          files += find_files(id, depth: 1, type: "link")
 
-        includes.push(files) if files
+          includes.push(files) if files
+        end
       end
 
       # [].flatten! == nil
--- a/test/helpers/mock_loader.rb
+++ b/test/helpers/mock_loader.rb
@@ -88,7 +88,7 @@ class MockLoader
       mockfile.call("emptyfile")
     }
 
-    mock.files = {
+    mock_files = {
       "/proc/net/bonding/bond0" => mockfile.call("bond0"),
       "/etc/ssh/ssh_config" => mockfile.call("ssh_config"),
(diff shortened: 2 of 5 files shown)


test/helpers/mock_loader.rb
```diff
@@ -175,6 +174,19 @@ def md.directory?
       "/etc/selinux/selinux_conf" => mockfile.call("selinux_conf"),
     }
 
+    if @platform[:name] == "ubuntu" && @platform[:release] == "14.04"
```
Vasu1105: I think we should remove test for 14.04 as we no more support it and include for latest one

lib/inspec/resources/apache_conf.rb
```diff
@@ -101,12 +101,14 @@ def include_files(params)
       include_files_optional = params["IncludeOptional"] || []
 
       includes = []
-      (include_files + include_files_optional).each do |f|
-        id    = Pathname.new(f).absolute? ? f : File.join(conf_dir, f)
-        files = find_files(id, depth: 1, type: "file")
-        files += find_files(id, depth: 1, type: "link")
+      unless conf_dir.nil?
```
Vasu1105, line 104: How we handle if conf_dir is nil ?

Nik08: Yes I need to discuss the possible solution for this on today's standup hence it is in WIP right now

PR #5604, opened 2021-07-22
Add rocky and almalinux to service resource

sspans-sbp:
Add support for rockylinux and almalinux to the service resource

## Description
Currently the service resource doesn't support rockylinux and almalinux.

```
  Service chef-client
     ↺  The `service` resource is not supported on your OS yet.
  Service sshd
     ↺  The `service` resource is not supported on your OS yet.
```

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -141,7 +141,7 @@ module Inspec::Resources
         elsif version > 0
           SysV.new(inspec, service_ctl || "/usr/sbin/service")
         end
-      when "redhat", "fedora", "centos", "oracle", "cloudlinux", "scientific"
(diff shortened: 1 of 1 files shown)


rjhornsby:
It looks like this is being held up because of a ruby version issue in the testing suite. Not sure how to fix that myself, but am hitting the `resource not supported` issue addressed by the PR. The change required to support Rocky and Almalinux is trivial.

Any chance we can get this looked at? Thanks!

sspans-sbp:
Not quite sure what the reason for the delay. This should be trivial to merge.

clintoncwolfe (approved):

PR #5606, opened 2021-07-27
Fix FreeBSD service enabled check substring edge-case

zofrex:
* Now services named a suffix of another service don't have false positives for being enabled
* Use new 'enabled' service command on FreeBSD 10+

## Description

Without this patch, inspec will falsely detect services as enabled on FreeBSD if their name is a suffix of another, enabled, service.

Example:

If `openntpd` is enabled, and `ntpd` is disabled, this check will fail when it shouldn't:

```
describe service('ntpd') do
  it { should_not be_enabled }
end
```

This patch fixes the bug by updating the regular expression used to search the list of enabled services, but it also makes use of the new `enabled` service command that was added in FreeBSD 10 when the detected FreeBSD version is 10 or greater.

When support for FreeBSD 9 is dropped, the old regex method could be removed entirely.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -504,7 +504,7 @@ module Inspec::Resources
 
       # search for the service
 
-      srv = %r{^.*/#{service_name}$)}.match(cmd.stdout)
+      srv = %r{^.*/(#{service_name}$)}.match(cmd.stdout)
       return nil if srv.nil? || srv[0].nil?
 
       enabled = true
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -504,7 +504,7 @@ module Inspec::Resources
 
       # search for the service
 
-      srv = /(^.*\/#{service_name}$)/.match(cmd.stdout)
+      srv = %r{^.*/#{service_name}$)}.match(cmd.stdout)
       return nil if srv.nil? || srv[0].nil?
 
       enabled = true
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -163,7 +163,12 @@ module Inspec::Resources
       when "mac_os_x", "darwin"
         LaunchCtl.new(inspec, service_ctl)
       when "freebsd"
-        BSDInit.new(inspec, service_ctl)
+        version = os[:release].to_f
+        if version < 10
+          BSDInit.new(inspec, service_ctl)
+        else
+          FreeBSD10Init.new(inspec, service_ctl)
+        end
       when "arch"
         Systemd.new(inspec, service_ctl)
       when "coreos"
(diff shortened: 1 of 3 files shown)


lib/inspec/resources/service.rb
```diff
@@ -510,6 +519,37 @@ def info(service_name)
     end
   end
 
+  # @see: https://www.freebsd.org/doc/en/articles/linux-users/startup.html
+  # @see: https://www.freebsd.org/cgi/man.cgi?query=rc.conf&sektion=5
+  # @see: https://www.freebsd.org/cgi/man.cgi?query=rc&apropos=0&sektion=8&manpath=FreeBSD+10.0-RELEASE&arch=default&format=html
+  class FreeBSD10Init < ServiceManager
```
zofrex, line 531: I have no opinion at all on whether it is better to name the new one `BSDInit` and rename the old one to `BSDInitOld` or `FreeBSD6Init`, or whatever.

lib/inspec/resources/service.rb
```diff
@@ -472,24 +477,28 @@ def info(service_name)
...
+      # /etc/rc.d/ip6addrctl
+      # /etc/rc.d/devd
+
       cmd = inspec.command("#{service_ctl} -e")
       return nil if cmd.exit_status != 0
 
       # search for the service
-      srv = /(^.*#{service_name}$)/.match(cmd.stdout)
+
+      srv = /(^.*\/#{service_name}$)/.match(cmd.stdout)
```
zofrex: The `\/` here is the actual bugfix

lib/inspec/resources/service.rb
```diff
@@ -776,7 +816,14 @@ class BSDService < Service
...
     def select_service_mgmt
-      BSDInit.new(inspec, service_ctl)
+      os = inspec.os
+      version = os[:release].to_f
+
+      if version >= 10
+        FreeBSD10Init.new(inspec, service_ctl)
+      else
+        BSDInit.new(inspec, service_ctl)
+      end
```
zofrex, line 832: I'm not 100% sure if it's the right approach to add an automatic selection inside the `BSDService` type, rather than add a second service type? The main thing that led me to this approach was trying to decide, if it was two service types, which one (FreeBSD10+ or the old one) would map to the existing `bsd_service` name. Most people presumably are on 10+ by now, but it would be backwards compatibility-breaking to default to a method that doesn't work on 9. In the end it seemed like this was the win/win approach, but I'm not married to it.

zofrex:
Happy to deal with the sonar issues (if issues in test code need dealing with), but I'll wait for feedback on the code before polishing it 🙂

lib/inspec/resources/service.rb
```diff
@@ -157,7 +157,12 @@ def select_service_mgmt # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCom
       when "mac_os_x", "darwin"
         LaunchCtl.new(inspec, service_ctl)
       when "freebsd"
-        BSDInit.new(inspec, service_ctl)
+        version = os[:release].to_f
+        if version < 10
```
tas50, line 167: Is there a reason to support FreeBSD 9 and earlier at this point given how terribly EOL those releases are?

zofrex: Well, that's really a question for you folks :) Wikipedia says that 10 came out in January 2014, and 9 was no longer supported since December 2016, so I think a good argument can be made that supporting 9 and earlier is no longer needed. However, right now the Inspec documentation says 9 is supported, so I made sure to maintain compatibility with it. Happy to go with whatever you think is best!

zofrex:
How does this look? From my side I'm happy with this and it seems ready to merge, unless there's any feedback/changes required?

PR #5629, opened 2021-08-26
Fix for security_policy resource does not return array for local groups

### LINKED ISSUE #5627: Inspec security_policy resource does not return array for local groups
# Describe the problem
Customer having some issues with the Inspec security policy resource. When the security right contains local groups (rather than SIDs), the inspec resource returns a comma-separated string, rather than an array.

For example, this code error
```
describe security_policy do
 its('SeServiceLogonRight') { should match_array ['SSRSServices' ,'SQLServices'] }
end
```
with the following message
```
 Security Policy
 [FAIL] SeServiceLogonRight is expected to contain exactly "SSRSServices" and "SQLServices"
 expected a collection that can be converted to an array with `#to_ary` or `#to_a`, but got "SQLServices,SSRSServices"
```
Other user rights assignments (which contain SIDs) work as expected
```
administrators = 'S-1-5-32-544'
power_users = 'S-1-5-32-547'
describe security_policy do
 its('SeInteractiveLogonRight') { should match_array ['SSRSUnattendExec', administrators, power_users] }
end
```
The User confirmed this issue occurs in multiple environments and occurs with both local and group policy.

Expected behavior
Not sure if it is expected for local groups, if not It should be working for local groups as well.

Customer bug : https://github.com/chef/customer-bugs/issues/460

Vasu1105:

## Description

security_policy resource was not returning values in array format when Privilege rights contain any local groups only. This is fixed in this PR.

So when any security policy for privilege rights containing only local group it does not contain the SID for example values are like below
`SeServiceLogonRight = DB2ADMNS,db2admin`

Where as if there are values other than local group then values are 
`SeServiceLogonRight = DB2ADMNS,db2admin,*S-1-5-80-0`

## Related Issue

Fix #5627 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/security_policy.rb
+++ b/lib/inspec/resources/security_policy.rb
@@ -147,7 +147,7 @@ module Inspec::Resources
 
     # extracts the values, this methods detects:
     # numbers and SIDs and optimizes them for further usage
-    def extract_value(val)
+    def extract_value(key, val)
       if val =~ /^\d+$/
         val.to_i
       # special handling for SID array
@@ -166,14 +166,15 @@ module Inspec::Resources
       elsif !(m = /^\"(.*)\"$/.match(val)).nil?
         m[1]
       else
-        val
+        # When there is Registry Values we are not spliting the value for backward compatibility
+        key.include?("\\") ? val : val.split(",")
       end
     end
 
     def convert_hash(hash)
       new_hash = {}
       hash.each do |k, v|
-        v.is_a?(Hash) ? value = convert_hash(v) : value = extract_value(v)
+        v.is_a?(Hash) ? value = convert_hash(v) : value = extract_value(k, v)
         new_hash[k.strip] = value
       end
       new_hash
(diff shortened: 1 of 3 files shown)


lib/inspec/resources/security_policy.rb
```diff
@@ -166,14 +166,15 @@ def extract_value(val)
       elsif !(m = /^\"(.*)\"$/.match(val)).nil?
         m[1]
       else
-        val
+        # When there is Registry Values we are not spliting the value for backward compatibility
+        key.include?("\\") ? val : val.split(",")
```
Nik08, line 170: Would you mind explaining this check `key.include?("\\")` ?

Vasu1105: I have added my comments but yes we get some values of Registry Path as `MACHINE\\Software\\Microsoft\\Windows NT\\CurrentVersion\\Setup\\RecoveryConsole\\SecurityLevel=4,0` which we are not going to split as there are chances that it will break if anyone is using string comparison. Check the existing test case where you will get a better idea.

Nik08:
Lgtm

PR #5638, opened 2021-08-30
Fix url fetcher when default git profile branch is not master

### LINKED ISSUE #5635: git fetchers need to assume default branch is now named "main"
# Describe the problem

The `git` fetcher currently assumes the default branch is named "master", but on August 24th, most github repos had their default branch renamed to be "main".  This breaks running exec from the command line against git repos.

Nik08:

## Description

Fix url fetcher when default git profile branch is not master

**Needs strict code review on this**
## Related Issue

Fixes #5635 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/fetcher/url.rb
+++ b/lib/inspec/fetcher/url.rb
@@ -74,15 +74,17 @@ module Inspec::Fetcher
     BITBUCKET_URL_REGEX = %r{^https?://(www\.)?bitbucket\.org/(?<user>[\w-]+)/(?<repo>[\w-]+)(\.git)?(/)?$}.freeze
     BITBUCKET_URL_BRANCH_REGEX = %r{^https?://(www\.)?bitbucket\.org/(?<user>[\w-]+)/(?<repo>[\w-]+)/branch/(?<branch>[\w\.]+)(/)?$}.freeze
     BITBUCKET_URL_COMMIT_REGEX = %r{^https?://(www\.)?bitbucket\.org/(?<user>[\w-]+)/(?<repo>[\w-]+)/commits/(?<commit>[\w\.]+)(/)?$}.freeze
+    GITHUB_URL = "https://github.com".freeze
+    BITBUCKET_URL = "https://bitbucket.org".freeze
 
     def self.transform(target)
       transformed_target = if m = GITHUB_URL_REGEX.match(target) # rubocop:disable Lint/AssignmentInCondition
-                             default_branch = default_ref(m)
+                             default_branch = default_ref(m, GITHUB_URL)
                              "https://github.com/#{m[:user]}/#{m[:repo]}/archive/#{default_branch}.tar.gz"
                            elsif m = GITHUB_URL_WITH_TREE_REGEX.match(target) # rubocop:disable Lint/AssignmentInCondition
                              "https://github.com/#{m[:user]}/#{m[:repo]}/archive/#{m[:commit]}.tar.gz"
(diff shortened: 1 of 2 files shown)


lib/inspec/fetcher/url.rb
```diff
@@ -71,11 +77,13 @@ def self.resolve_from_string(target, opts, username = nil, password = nil)
...
     def self.transform(target)
       transformed_target = if m = GITHUB_URL_REGEX.match(target) # rubocop:disable Lint/AssignmentInCondition
-                             "https://github.com/#{m[:user]}/#{m[:repo]}/archive/master.tar.gz"
+                             default_branch = default_ref(m)
+                             "https://github.com/#{m[:user]}/#{m[:repo]}/archive/#{default_branch}.tar.gz"
                            elsif m = GITHUB_URL_WITH_TREE_REGEX.match(target) # rubocop:disable Lint/AssignmentInCondition
                              "https://github.com/#{m[:user]}/#{m[:repo]}/archive/#{m[:commit]}.tar.gz"
                            elsif m = BITBUCKET_URL_REGEX.match(target) # rubocop:disable Lint/AssignmentInCondition
-                             "https://bitbucket.org/#{m[:user]}/#{m[:repo]}/get/master.tar.gz"
+                             default_branch = default_ref(m)
```
clintoncwolfe: default_ref assumes a github address, but here it is being used on a bitbucket repo - this will fail for bitbucket repos.

lib/inspec/fetcher/url.rb
```diff
@@ -120,6 +128,38 @@ def to_s
 
     private
 
+    class << self
+      def default_ref(match_data)
+        remote_url = "https://github.com/#{match_data[:user]}/#{match_data[:repo]}.git"
```
clintoncwolfe: Here is where default ref assumes a github address.

aaronlippold:
So this will work reguardless of the default branch name?

lib/inspec/fetcher/url.rb
aaronlippold: could we just make the whole URI part of the variable.. what if it is a git on gitlab ... like gitlab.mitre.org?

aaronlippold: we have the whole uri of the repo when we put it in the cli command right?

aaronlippold: and we put the whole uri in the `inpsec.yml` - basically every time we reference a git repo we do so via a full uri right?

clintoncwolfe: The existing code was applying transformations to fetch tarballs, if possible, by recognizing github and bitbucket URLs. We can't do that in a generic basis, as we don't know what transformation may need to be applied. There may be a transformation that works for gitlab, but the user can always just supply the correct tarball URL in the first place.

As this is a hotfix PR, to fix an urgent problem, we will not be expanding any functionality on this PR. If you'd like to explore expanding the transformation logic, feel free to do so, but we need to ship this fix.

clintoncwolfe (approved):
Thanks @Nik08, looks good to me.

lib/inspec/fetcher/url.rb
aaronlippold: Ok. I will open an issue, as I know we have customers using InSpec with other git platforms - GitLab mostly - which we should allow for. In theory, I would have guessed that given the user has to give the full git repo URI that we could use that.

PR #5639, opened 2021-08-30
Remove license note + update resource count in main docs page

tas50:
This is not the place to throw a weird licensing reminder and we have
500 resources not 80.


kagarmoe:
Each of the products has the note on  at least one central page. It is perhaps an artifact of the license model change, but it is also still important information.

tas50:
It should all be located on https://docs.chef.io/versions/ which has a list of community vs. commercial releases for each product.

kagarmoe (approved):

kagarmoe:
OK to merge once the buildkite tests are fixed.

PR #5656, opened 2021-09-07
Fix --controls option was not working for dependent profile

Vasu1105:
Signed-off-by: Vasu1105 <vasundhara.jagdale@chef.io>

## Description

- This fix the filtering not working for the dependent profile with `--controls` options
- Tested the same for `include_controls` and `require_controls`
- Tested the profile_a > profile_b > profile_c > profile_d >profile_e
## Related Issue

Fix #5647 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

PR #5657, opened 2021-09-07
Fix --tags filter for dependent profiles

### LINKED ISSUE #5653: --tags does not support dependent profiles
### 🎛 Description

Like #5647, but for --tags.

Nik08:

## Description

- Fix `--tags` filter for dependent profiles, it was not able to filter dependent profiles till now.
- Introduced filtering logic for both `include_controls` and `require_controls` cases.
- Tested manually till 4 levels, A->B->C->D
## Related Issue

Fixes #5653 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -104,7 +104,7 @@ module Inspec::DSL
     mock = Inspec::Backend.create(Inspec::Config.mock)
     include_ctx = Inspec::ProfileContext.for_profile(profile, mock)
     include_ctx.load(block) if block_given?
-    include_ctx.control_eval_context.instance_variable_set(:@conf, opts[:conf])
+    include_ctx.control_eval_context.conf = opts[:conf]
     control_eval_ctx = include_ctx.control_eval_context
     # remove all rules that were not registered
     context.all_rules.each do |r|
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -93,22 +93,29 @@ module Inspec::DSL
     context = dep_entry.profile.runner_context
     # if we don't want all the rules, then just make 1 pass to get all rule_IDs
     # that we want to keep from the original
-    filter_included_controls(context, dep_entry.profile, &block) unless opts[:include_all]
+    filter_included_controls(context, dep_entry.profile, opts, &block) if !opts[:include_all] || !(opts[:conf]["profile"].include_tags_list.empty?)
+
     # interpret the block and skip/modify as required
     context.load(block) if block_given?
     bind_context.add_subcontext(context)
   end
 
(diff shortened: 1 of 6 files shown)


aaronlippold:
Some good testing content is here:

https://saf.cms.gov/#/validate

this one is three levels of chain:

https://github.com/CMSgov/cms-ars-3.1-moderate-aws-rds-oracle-database-12c-stig-overlay

aaronlippold:
I woiuld suggest the unit test to do at least 4 to 5 levels of dep chain:

CORE -> ORG -> CLOUD -> TEAM -> REGION 

Is now a common pattern

aaronlippold:
Where region could be east or west, or secure and non-secure data enviroments etc

lib/inspec/dsl.rb
```diff
@@ -93,22 +93,29 @@ def self.load_spec_files_for_profile(bind_context, opts, &block)
...
     context.load(block) if block_given?
     bind_context.add_subcontext(context)
   end
 
-  def self.filter_included_controls(context, profile, &block)
+  def self.filter_included_controls(context, profile, opts, &block)
     mock = Inspec::Backend.create(Inspec::Config.mock)
     include_ctx = Inspec::ProfileContext.for_profile(profile, mock)
     include_ctx.load(block) if block_given?
+    include_ctx.control_eval_context.instance_variable_set(:@conf, opts[:conf])
```
clintoncwolfe: I had the same complaint on #5656 - we really should avoid instance_variable_Set on our own code whereever possible. we can make `conf` an attr_accessor on ControlEvalContext and set it properly.

Nik08: I understand now why you are suggesting to avoid using this. It goes against maintaining the class level encapsulations. Thanks for pointing this out. Have replaced this with `attr_accessor` usage.

aaronlippold:
Does it also makes sense, while we are in this part of the codebase, to also round out the #5327 PR?

Perhaps we could just update the waived implemenation to auto-tag the control waived and then the --tags filter could take care of it? 

Read waived file, make hash, on the load_control add a `if in tagged, add 'wavied' tag to control`, on the internal run logic, `unless tagged hash empty, apend --skip-tagged=waived` 

This would also give us `--skip-contorols` and --skip-tagged` which could be very useful in the general case. These should be just the inverser logic of their counterparts.

So the argument could also be just to use `--skip-control` internally to do waivers.

In my openion, waived is just a special case of tagging.

aaronlippold:
any blockers on this?

clintoncwolfe (approved):
Thanks!

clintoncwolfe:
We'll keep #5327 separate for now - it's a lot more complex.

PR #5665, opened 2021-09-16
Add csv without headers support in csv resource

### LINKED ISSUE #5450: csv resource should support csv files that don't have a header row
## Detailed Description

CSV resource only supports comma-separated files with a header row.

## Context

The csv resource currently only supports parsing a CSV file that has a header row defining what each of the columns in each row represent. In some cases, I have to manage and validate a CSV that doesn't have this header row. There is no way to use inspec and assert that each row is what I expect it to be.

## Possible Implementation

- I believe that the instantiation of the CSV object declares `headers: true` should be optional. The default value should be true, but give the option to set this to false, which would allow ruby to parse a csv file without a header row.
- https://github.com/inspec/inspec/blob/master/lib/inspec/resources/csv.rb#L31
- When a csv with no header row is parsed, matchers could refer to each field on each line as an array position

Vasu1105:

## Description

Add support for csv without headers in csv resource.
To use above options now user has to send the headers value as false in the csv resource as follows

```
describe csv('example.csv', false).params do
  its([0]) { should eq (['Product','Size','Color', 'Price']) }
end
```

## Related Issue

Fix #5450 
## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/csv.md
+++ b/docs-chef-io/content/inspec/resources/csv.md
@@ -31,7 +31,7 @@ A `csv` resource block declares the configuration data to be tested:
       its('name') { should cmp 'foo' }
     end
 
-Test ``csv`` without headers
+Test `csv` file without headers
 
     describe csv('file', false).params do
       its([0]) { should cmp 'name' }
@@ -40,17 +40,17 @@ Test ``csv`` without headers
 where
 
 - `'file'` is the path to a CSV file
-- 'true' is the value for headers. Default `true`. If `false` then it considers csv does not have headers.
+- `true` or `false` tests a CSV file with or without headers. Default value: `true`.
 - `name` is a configuration setting in a CSV file
 - `should eq 'foo'` tests a value of `name` as read from a CSV file versus the value declared in the test
-- `params` when headers are set to false use this method to fetch the data.
+- `params` is the method for fetching data from a CSV file without headers.
 - `[0]` is the array element position.
 
 ## Examples
 
 The following examples show how to use this Chef InSpec audit resource.
 
-### Test a csv file without headers
+### Test a CSV file without headers
 
     describe csv('some_file.csv', false).params do
       its([0]) { should eq ["name"] }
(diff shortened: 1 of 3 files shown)


docs-chef-io/content/inspec/resources/csv.md
```diff
@@ -27,20 +27,35 @@ This resource first became available in v1.0.0 of InSpec.
 
 A `csv` resource block declares the configuration data to be tested:
 
-    describe csv('file') do
+    describe csv('file', true) do
       its('name') { should cmp 'foo' }
     end
 
+Test ``csv`` without headers
```
IanMadd:
```suggestion
Test a CSV file without headers:
```

docs-chef-io/content/inspec/resources/csv.md
```diff
@@ -27,20 +27,35 @@ This resource first became available in v1.0.0 of InSpec.
...
+
+    describe csv('file', false).params do
+      its([0]) { should cmp 'name' }
+    end
+
 where
 
 - `'file'` is the path to a CSV file
+- 'true' is the value for headers. Default `true`. If `false` then it considers csv does not have headers.
```
IanMadd:
```suggestion
- `true` or `false` tests a CSV file with or without headers. Default value: `true`.
```

docs-chef-io/content/inspec/resources/csv.md
```diff
@@ -27,20 +27,35 @@ This resource first became available in v1.0.0 of InSpec.
...
 - `name` is a configuration setting in a CSV file
 - `should eq 'foo'` tests a value of `name` as read from a CSV file versus the value declared in the test
+- `params` when headers are set to false use this method to fetch the data.
```
IanMadd:
```suggestion
- `params` are the method for fetching data from a CSV file without headers.
```

docs-chef-io/content/inspec/resources/csv.md
```diff
@@ -27,20 +27,35 @@ This resource first became available in v1.0.0 of InSpec.
...
+- `[0]` is the array element position.
 
 ## Examples
 
 The following examples show how to use this Chef InSpec audit resource.
 
+### Test a csv file without headers
```
IanMadd:
```suggestion
### Test a CSV file without headers
```

clintoncwolfe (approved):
Thanks!

PR #5667, opened 2021-09-17
Group & User Resources - Resolve name case-sensitivity issue for windows

### LINKED ISSUE #5632: InSpec User and Group resources evaluate Windows Users and Group names case sensitively
Windows users and group names are case insensitive. 'bob_smith' and 'Bob_Smith' are the same user. 
The InSpec User and Group resource matchers fail if case does not match exactly. 

## Detailed Description
The following check will fail if 'AbC' is a member of Administrators, even though they are the same user but the user name only differs in case. 
```
describe group('Administrators') do
  ## User named AbC exists but this test fails
  its('members') { should_not include 'Abc' } 
end
```
The following check also fails because the Administrators group name differs in case. 
```
describe group('administrators') do
  ## This test fails because the group name is Administrators with a capitol A
  its('members') { should_not include 'Abc' }
end
```
It should be noted that Chef Infra Client's Group resource treat (at least) group and user names case insensitively such that this implementation is successful no matter the case of the user or group name. The user ABC is removed from Administrators.
```
  group 'Removing Abc from the Windows Administrators Group' do
    action                :modify
    append                true
    excluded_members      ['AbC']
    group_name            'administrators'
  end

```

## Context
Our brownfield systems include users that have been created and named manually, though with a common name. The manual creation of these users/groups has created a variety of username / case discrepancies. Being able to do case-insensitive comparisons would make this type of evaluation possible for us. 

## Possible Implementation
I guess there might be a way to implement this in the control rather than resource, however that exceeds my Ruby skills and acts against the desire for the code to be human readable.

Ideally the resource we be aware that it should do case insensitive matches on Windows operating systems.

Nik08:

## Description

- **Group name** will be evaluated case insensitively for windows group in **group** resource.
- **Group members** will be evaluated case insensitively using `include` matcher for windows group members in **group** resource.
- **User name** will be evaluated case insensitively for windows in **user** resource.
- **User groups** will be evaluated case insensitively using `include` matcher for windows user groups in **user** resource.

## Related Issue

Closes #5632 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/groups.rb
+++ b/lib/inspec/resources/groups.rb
@@ -150,7 +150,11 @@ module Inspec::Resources
     def group_info
       # we need a local copy for the block
       group = @group.dup
-      @groups_cache ||= inspec.groups.where { name == group }
+      if inspec.os.windows?
+        @groups_cache ||= inspec.groups.where { name.casecmp?(group) }
+      else
+        @groups_cache ||= inspec.groups.where { name == group }
+      end
     end
 
     def empty_value_for_members
--- a/lib/inspec/resources/groups.rb
+++ b/lib/inspec/resources/groups.rb
@@ -22,6 +22,18 @@ module Inspec::Resources
     end
   end
 
+  # Class defined to check for members without case-sensitivity
+  class Members < Array
+    def initialize(group_members)
+      @group_members = group_members
+      super
+    end
+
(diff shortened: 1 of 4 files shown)


clintoncwolfe (approved):
Thanks!

PR #5680, opened 2021-09-28
Apply cookstyle to the example profile

tas50:
Let's make sure people start off on the right foot.


clintoncwolfe (dismissed):
Thanks!

clintoncwolfe:
So looks like cookstyle and the inspec chefstyle conflict with their string literal settings - causing it to fail the linter.

clintoncwolfe (commented):
I resolved the conflict between chefstyle/cookstyle by temporarily disabling .rubocop.yml (which was excluding the control files) and then auto correcting the control files of the example profile. I then expanded the exclusion slightly to also cover the library files and plugins. The plugins and other example profiles are not cookstyle compliant; the plugins are code and likely should be chefstyle compliant.

This should pass the chefstyle CI linter now.

Issue #5644, opened 2021-08-31
Remove Ubuntu 14.04 and 15.04 unit tests which are still there as we don't support them.


Vasu1105:
# Describe the problem

https://github.com/inspec/inspec/blob/main/test/helpers/mock_loader.rb#L21-L22
Many resources like below using this
https://github.com/inspec/inspec/blob/7415fc11d000dae1813a14ba12d957851fa999b3/test/unit/resources/group_test.rb#L8
https://github.com/inspec/inspec/blob/7415fc11d000dae1813a14ba12d957851fa999b3/test/unit/resources/nginx_conf_test.rb#L10
https://github.com/inspec/inspec/blob/7415fc11d000dae1813a14ba12d957851fa999b3/test/unit/resources/ip6tables_test.rb#L9
https://github.com/inspec/inspec/search?q=Ubuntu1504

clintoncwolfe:
Is it possible to use a more modern version, like 20.04?

PR #5681, opened 2021-09-29
Update the unit test files to use latest versions of OS rather than the older. 

Vasu1105:

## Description

Many unit tests where we only do the mocking of the OS for testing (not doing any real testing on those OS) should be minimized. This PR addresses the same.
## Related Issue

#5644 
## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #5689, opened 2021-10-02
Add back Ubuntu 16.04 packages + testing

tas50:
Ubuntu 16.04 is no longer EOL. They've extended support for a full 10 years so it goes EOL in 2026 now.


clintoncwolfe (approved):
Thanks!

PR #5697, opened 2021-10-13
Add Windows support for http resource.

Vasu1105:
## Description

Adopted PR #5280
Done required changes to handle `post` request
## Related Issue

Fix #3969, Closes #5280
## Types of changes

## Checklist:


mbaitelman:
I think it should be documented that this only works with `enable_remote_worker: true`, which was one of the issues I had with my original PR.

Vasu1105:
> I think it should be documented that this only works with `enable_remote_worker: true`, which was one of the issues I had with my original PR.

@mbaitelman That option is no more required to be provided as the `http` resource remote worker is enabled by default for remote targets and cannot be disabled. Check this https://github.com/inspec/inspec/blob/main/lib/inspec/resources/http.rb#L32

mbaitelman:
@Vasu1105 what would happen if one ran Inspec from a windows host on itself? Will the resource work then?

Vasu1105:
@mbaitelman  Yes, that also works I tested that.

clintoncwolfe (approved):
Thank you for this!

kagarmoe (approved):

PR #5699, opened 2021-10-18
Removed the ServerSpec Reference from the Inspec Doc

dishanktiwari2501:
Signed-off-by: Dishank Tiwari <dtiwari@progress.com>

When we first release InSpec it was necessary to explain how it differed from Serverspec and why we didn't just further develop that tool. In 2021 that's not necessary and this document is just more confusing content to read. It should just go away.

## Related Issue

- https://github.com/chef/chef-web-docs/issues/3393

## Checklist:


lokeshk1987:
There can be community users who are the users of serverspec and migrated to Inspec. This can be an impact to such users. My take here is - to retain the existing serverspec doc.

kagarmoe:
What are your thoughts on @lokeshk1987's point @tas50 ?

tas50:
It's @lokeshk1987's call, but I think the days of ServerSpec in the Chef community are long long gone.

lokeshk1987:
@kagarmoe - Lets retain this for next 2 to 3 quarters. Then we can plan for removing it. Hope it should be fine.

kagarmoe:
Ok. Please let us know when you're ready to reopen it.

PR #5702, opened 2021-10-19
Oracle db session resource fix - when invoking query using os user and db role

Nik08:

## Description

1. Oracle db session resource fix - when invoking query using os user and db role
2. Handling of query with special character `$`
3. Raising `skip_resource` when in windows non supported `os user` is used with the resource.
## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #5502, opened 2021-05-04
Fix : windows_firewall_rule fails to validate more than 1 rule depending on how it's executed 

Vasu1105:
Signed-off-by: Vasu1105 <vasundhara.jagdale@chef.io>

## Description

windows_firewall_rule is using `Remove-TypeData System.Array` as a workaround for PS bug here: https://bit.ly/2SRMQ8M,
`Remove-TypeData` Deletes extended types from the current session.[Ref](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/remove-typedata?view=powershell-7.1)  So when we are executing more that one rules using different describe blocks as we have already deleted the extended type in the same session when next describe was getting called that command was giving error. due to which the cmd output was coming blank due to which the result was empty for next set of rules.

Following is the output of executing the Remove-TypeData in same Poweshell window(same session)
```
PS C:\Users\Vasundhara> Get-TypeData -TypeName System.Array

TypeName     Members
--------     -------
System.Array {[Count, System.Management.Automation.Runspaces.AliasPropertyData]}


PS C:\Users\Vasundhara> Remove-TypeData -TypeName System.Array
PS C:\Users\Vasundhara> Remove-TypeData -TypeName System.Array
Remove-TypeData : Error in TypeData "System.Array": The type "System.Array" was not found. The type name value must be
the full name of the type. Verify the type name and run the command again.
At line:1 char:1
+ Remove-TypeData -TypeName System.Array
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Remove-TypeData], RuntimeException
    + FullyQualifiedErrorId : TypesDynamicRemoveException,Microsoft.PowerShell.Commands.RemoveTypeDataCommand

PS C:\Users\Vasundhara>
```
So as a solution we are using `Get-TypeData` command to get extended type in current session and pipe it to the `Remove-TypeData` so if there is not System.Array in current session then it will not cause any issue. REF : https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/get-typedata?view=powershell-7.1

```
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\Vasundhara> Get-TypeData -TypeName System.Array

TypeName     Members
--------     -------
System.Array {[Count, System.Management.Automation.Runspaces.AliasPropertyData]}


[...]

PR #5704, opened 2021-10-21
Fix windows_firewall resource fails to validate more than 1 rule depending on how it's executed

### LINKED ISSUE #5703: Windows Firewall Inspec resource only returns values on first profile
When running Inspec tests using Windows Firewall resources only the first profile checked will get results, all subsequent profiles will not get any values and return false for enabled, even when they are enabled.

# Describe the problem

When running Inspec tests using Windows Firewall resource only the first profile checked will get results, all subsequent profiles will not get any values and return false for enabled, even when they are enabled.

```
control 'AC.X.XXXX-Firewall Enabled Domain' do
  impact 0.6
  title 'Firewall Enabled'
  desc 'Check if Windows Firewall is enabled.'

  describe windows_firewall('Domain') do
    it { should be_enabled }
    its("num_rules") { should be 541 }
  end
end

control 'AC.X.XXXX-Firewall Enabled' do
  impact 0.6
  title 'Firewall Enabled'
  desc 'Check if Windows Firewall is enabled.'

  describe windows_firewall('Public') do
    it { should be_enabled }
    its("num_rules") { should be 474 }
  end
  describe windows_firewall('Private') do
    it { should be_enabled }
    its("num_rules") { should be 521 }
  end
end
```
 
Here The Domain Profile is the first check and reports true for enabled and has a valid rule count. The Public and Private profiles seem to return nothing. 

When commenting out the Domain Profile control, the Public Profile will pass as it's now the first one that runs. Private Profile will still fail.

Any profile after the first doesn't return any data when the Inspec profile runs. If you swap the order of the profiles tested you should find that the first one will return values, and then the others won't.

## Possible Solution

https://github.com/inspec/inspec/pull/5502/files

Vasu1105:

## Description

This is a similar issue which we fixed for the windows_firewall_rule resource in #5502. The detailed description for the cause of the issue given on the #5502 

## Related Issue

Fix #5703 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/windows_firewall.rb
+++ b/lib/inspec/resources/windows_firewall.rb
@@ -77,7 +77,7 @@ module Inspec::Resources
 
     def load_firewall_profile(profile_name)
       <<-EOH
-        Remove-TypeData System.Array # workaround for PS bug here: https://bit.ly/2SRMQ8M
+        Get-TypeData -TypeName System.Array | Remove-TypeData # workaround for PS bug here: https://bit.ly/2SRMQ8M
         $profile = Get-NetFirewallProfile -Name "#{profile_name}"
         $count = @($profile | Get-NetFirewallRule).Count
         ([PSCustomObject]@{


clintoncwolfe (approved):
Thanks!

PR #5706, opened 2021-10-22
Fix Oracle db session resource issues 

Nik08:

## Description

Fix Oracle db session resource issues 
1. CSV delimiter issue when `,` present in the query data, Example: `(DROP,3)`
2. Super user usage with oracle, also returns special message on `sudo - sudo_user`. Query output extraction handled in this case. 

## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

PR #5716, opened 2021-11-09
Stop producing packages for macOS 10.14

tas50:
We support N-2 macOS releases which is now 12, 11, and 10.15


tas50:
Hold off merging this until we can get ad-hocs run

tas50:
@clintoncwolfe This can get merged now

PR #5722, opened 2021-11-13
Move cookstyle dep from inspec gemspec to inspec-core, add rake

clintoncwolfe:
The cookstyle dep should have been added to inspec-core, as it is part of inspec check (and not part of the compiled/cloud optional set of offerings in the superset). 

We also missed a new runtime dep on rake, which was breaking the habitat builds and possibly others.

## Types of changes

## Checklist:

### DIFF
--- a/inspec-core.gemspec
+++ b/inspec-core.gemspec
@@ -45,5 +45,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency "semverse",           "~> 3.0"
   spec.add_dependency "multipart-post",     "~> 2.0"
 
+  # cookstyle support for inspec check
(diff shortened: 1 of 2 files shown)


tas50:
@clintoncwolfe This change is going to be very problematic for Infra Client as it makes rubocop, cookstyle, and rake dependencies of every Infra Client install. The intention of inspec-core was that it would be InSpec as an engine/library for use in Infra Client without the full cloud deps. We need to ensure that InSpec core retains the lightweight design. Users should not be running anything CLI with InSpec core as it is intended for bundling only. Any CLI deps should stay in the full InSpec package.

Issue #5723, opened 2021-11-13
inspec check emits warning "Process.fork is not supported by this Ruby" on windows since 4.49.0


clintoncwolfe:
The introduction of the cookstyle integration (#5618) has caused a warning message to be emitted to STDERR on Windows whenever `inspec check` (or a command that calls `inspec check`, such as `archive`) is executed.

This breaks the Windows tests. - [example](https://buildkite.com/chef/inspec-inspec-main-artifact-habitat/builds/2871#1ed53b6e-04a8-47dd-8086-291f752f3685) 

It is unclear what the root cause is, but it may be a Rake task.

# Describe the problem

PR #5724, opened 2021-11-13
Disable CookStyle integration on Windows 

clintoncwolfe:
It appears that the current approach to CookStyle integration is broken under Windows (when InSpec is run from a Windows workstation, not necessarily on a Windows target). The feature fails to work, and a message is emitted to STDERR reading "Process.fork is not supported by this Ruby".  The root cause has not yet been identified but it appears to be in a Rake task.

For now, we are withdrawing the feature from Windows installations until such time as it can be fixed.

Refs #5723 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -122,8 +122,13 @@ class Inspec::InspecCLI < Inspec::BaseCLI
       end
       puts
 
+      enable_offenses = !Inspec.locally_windows? # See 5723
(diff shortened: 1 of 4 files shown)


Nik08:
This looks good to me Clinton! And I am curious to know, how to check this in future releases for any new feature so that we can cover this test.

Nik08 (approved):

PR #5757, opened 2021-12-03
Add proxy parameter to http resource

### LINKED ISSUE #4725: Add support for proxy parameter in http resource
Being able to test http connectivity through proxy only is a common use case for security purpose.
Currently the `http` resource doesn't support providing proxy details for the request

## Detailed Description
The current `http` implementation uses `curl` or `Faraday` to make http requests. Both tools support proxy through environment variables, however I'm not aware of an easy way to tell `http` to use or not use a proxy

## Context

My remote server has no internet access, however it can connect to internet through a private proxy.

A simple use case is:
Control 1:
I should not be able to request www.example.com if i'm not using the proxy
Control 2: 
I should be able to request www.example.com if i'm using the proxy

## Possible Implementation
Add a new option to do `http('url', proxy: 'user:password@proxy_host:proxy_port')` and `http('url', proxy: 'disable')`(to override http_proxy ENV)
Pass the option to Faraday: https://github.com/inspec/inspec/blob/master/lib/inspec/resources/http.rb#L149 => https://www.rubydoc.info/github/lostisland/faraday/Faraday/Connection#proxy-instance_method
Pass the option to curl https://github.com/inspec/inspec/blob/master/lib/inspec/resources/http.rb#L244 => `-x 

If `disabled` is passed, unset before to execute and set after

## Workaround

Is there a nice workaround to change ENV variable on a remote host but only for a specific control?

I'm currently configure my remote server with HTTP_PROXY... and use https://www.inspec.io/docs/reference/resources/host/ to test non-proxy traffic is not working.

Vasu1105:
## Description

Added `proxy` parameter to `http` resource: It accepts Proxy URI as values / Proxy hash as value or can set to `disable` to ignore proxy environment variable.
## Related Issue

Fixes #4725 
## Types of changes

## Checklist:


clintoncwolfe (approved):
Thanks!

docs-chef-io/content/inspec/resources/http.md
```diff
@@ -168,6 +168,36 @@ Beginning with Chef InSpec 1.41, you can enable the ability to have the HTTP tes
...
+      ...
+    end
+
+Note : For Windows remote target system proxy values needs to be provided in the hash format.
+
+`proxy` can be set to `disable` to ignore the proxy set in the environment variable.
+
+    describe http('http://localhost:8080/ping', proxy: 'disable') do
+      ...
+    end
```
IanMadd:
```suggestion
Specify a `proxy` to test by passing in the proxy URI or a hash of the proxy URI, a username, and password. Specify `disable` to ignore a proxy set as an environment variable.

You can include the username and password in the `proxy` parameter:

    describe http('http://localhost:8080/ping', proxy: "http://username:password@www.example.com:3128") do
      ...
    end

The `proxy` parameter also accepts proxy options in hash format:

    describe http('http://localhost:8080/ping', proxy: { uri: 'http://www.example.com:3128', user: 'username', password: 'proxypassword'}) do
      ...
    end

Use `disable` to ignore the proxy set in the environment variable:

    describe http('http://localhost:8080/ping', proxy: 'disable') do
      ...
    end

{{< note >}}

Windows remote targets do not accept username and password values in a string; use the hash format instead.

{{< /note >}}

{{< note >}}

Special characters in the URI must be converted to their UTF-8 equivalent when passed in to the `proxy` parameter as a string. For example, the string `http://username:bar@123@www.example.com:3128` must be passed in as `http://username:bar%40123@www.example.com:3128` instead.

Special characters may be passed into the hash format without conversion to UTF-8 characters.

{{< /note >}}
```

IanMadd (approved):

PR #5758, opened 2021-12-06
Added Timezone Resource

### LINKED ISSUE #4013: Support checking for server timezone settings
### 🎛 Description
🙋 feature request?

### 🌍 InSpec Version
Inspec version 3.9.0

Requesting a resource to check for timezone configuration.  Currently we're having to resort to this:

```
control "time-2" do
    impact 1.0
    title "Verify timezone is America/Chicago"
    describe bash('timedatectl status | grep Chicago') do
        its('stdout') { should include 'Time zone: America/Chicago (CDT, -0500)' }
        its('stderr') { should eq '' }
    end
end
```

It'd be nice if there was a native resource to check for various time settings including timezone.

Nik08:

## Description

- Added timezone resource to test timezone configuration values.
- Test cases and documentation covered.
- Supports Unix & Windows platform

## Related Issue

Closes #4013 
## Types of changes

## Checklist:


docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Availability
+
+### Installation
+
+This resource is distributed along with Chef InSpec itself. You can use it automatically.
+
+### Version
+
+This resource first became available in v1.0.0 of InSpec.
```
clintoncwolfe: Uh, current version is 4.53 ....

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,75 @@
...
+
+    describe timezone do
+      its('identifier') { should eq 'Asia/Kolkata' }
+      its('name') { should eq 'IST' }
+      its('time_offset') { should eq '+0530' }
+    end
+
+## Properties
+
+### identifier
```
clintoncwolfe, line 49: How about we call this `location` ?

docs-chef-io/content/inspec/resources/timezone.md
Nik08: Geez, my bad. Will change this. Thanks!

clintoncwolfe (approved):
Need to fix or remove the version available notice in the docs, but otherwise looks good - thanks!

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Availability
+
+### Installation
+
+This resource is distributed along with Chef InSpec itself. You can use it automatically.
+
+## Syntax
+
+A `timezone` resource fetches the timezone configurations of the system and compares the output with the test:
```
IanMadd:
```suggestion
A `timezone` resource block fetches the time zone configuration of a system and compares the output with the test:
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+    describe timezone do
+      its('property') { should eq 'expected value' }
+    end
+
+where
```
IanMadd, line 30:
```suggestion
where:
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Properties
+
+### identifier
+
+The `identifier` property returns the location identifier of the timezone.
+
+An example of checking the **identifier** for India Time :
+
+    its('identifier') { should eq 'Asia/Kolkata }
```
IanMadd:
```suggestion
    its('identifier') { should eq 'Asia/Kolkata' }
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+      its('time_offset') { should eq '+0530' }
+    end
+
+## Properties
+
+### identifier
+
+The `identifier` property returns the location identifier of the timezone.
+
+An example of checking the **identifier** for India Time :
```
IanMadd:
```suggestion
An example of checking the **identifier** for the Asia/Kolkata time zone name:
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+    its('identifier') { should eq 'Asia/Kolkata }
+
+### name
+
+The `name` property returns the name of the timezone.
```
IanMadd:
```suggestion
The `name` property returns the time zone abbreviation.
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+An example of checking the **name** of India Time :
```
IanMadd, line 57: I probably read more about time zones for this PR than I should have, but it turns out that these abbreviations aren't standardized and so there are [several ISTs and several CSTs](https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations). Anyway, it looks like this test would pass if a system was set in Ireland, Israel, or India, and the only way to resolve which time zone the system has is to use one of the other properties.

```suggestion
An example of checking the **name** for the IST time zone abbreviation:
```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+
+    its('name') { should eq 'IST' }
+
```
IanMadd, line 66:
```suggestion
    its('name') { should eq 'IST' }

{{< note >}}

Several time zones share the same time zone abbreviation. Use one of the other properties to verify a specific time zone with a common abbreviation.

{{< /note >}}

```

docs-chef-io/content/inspec/resources/timezone.md
```diff
@@ -0,0 +1,71 @@
...
+### time_offset
+
+The `time_offset` property returns the identifier of a time offset from UTC (Coordinated Universal Time).
+
+An example of checking the **time_offset** of India Time:
```
IanMadd:
```suggestion
An example of checking if the **time_offset** is UTC+05:30:
```

IanMadd (commented):
Docs feedback

Nik08:
@IanMadd For docs change, wherever you have made the suggestion to rename `name` as `abbreviation`, this will be an abbreviation in linux systems but for windows it returns the full name. So to reach the common ground I mentioned it as `name`.
Does this make sense? Or do you suggest we should mention how they appear for different platforms?

For example: 
In linux the output would be something like `Time zone: Asia/Kolkata (IST, +0530)`, whereas for windows it would be this
```
Id                         : India Standard Time
DisplayName                : (UTC+05:30) Chennai, Kolkata, Mumbai, New Delhi
StandardName               : India Standard Time
DaylightName               : India Daylight Time
BaseUtcOffset              : 05:30:00
SupportsDaylightSavingTime : False
```
Also I believe the term `India Time` in examples, is how India's time zone is represented. You have mentioned some changes around this term. Have some confusions around suggestions on line 49 & 57. 

- `the **identifier** for the Asia/Kolkata time zone name` , here `Asia/Kolkata` is itself the identifier so this sounds confusing.
- `checking the **name** for the IST time zone abbreviation` , here `IST` is itself the name so again the same confusion.

IanMadd:
The full proper name for India Time is India Standard Time, so we should use that term.

> wherever you have made the suggestion to rename name as abbreviation, this will be an abbreviation in linux systems but for windows it returns the full name

So on Windows, will this pass on a system in India:

```
   its('name') { should eq 'IST' }
```


The [tz database](https://www.iana.org/time-zones) (where terms like Asia/Kolkata come from) uses the term **name** for values like Asia/Kolkata, so I'm trying to be consistent with that.

And IST is an abbreviation, so that's how I'm describing it.

Nik08:
Okay it clears up things, thanks @IanMadd :)

And on windows the test would pick it like this, 
`its('name') { should eq 'India Standard Time' }`

IanMadd:
Ok @Nik08, I made some changes that should clarify things.

Nik08:
Awesome @IanMadd !! Thanks :)

IanMadd (approved):

clintoncwolfe (approved):
Approved, thanks! - ignoring test failures as we have already fixed the activesupport issue on main.

PR #5783, opened 2022-01-01
Remove support for EOL Ruby 2.5

tas50:
With the release of Ruby 3.1, Ruby 2.5 is now EOL. We should not be supporting an EOL release of Ruby.


clintoncwolfe (approved):
Thanks for this!

PR #5793, opened 2022-01-11
CFINSPEC-2 Added options to allow winrm SSL connection with certificates 

### LINKED ISSUE #5659: WinRM Connections don't allow validated SSL/TLS
WinRM connections should be able to connect to remote nodes using SSL and validate the hosts' certificates. By default the ca-trust-file is set to 'nil' in train-winrm/tansport.rb but it IS an accepted parameter. InSpec provides no command-line option to set this value for use in WinRM connections.

Furthermore, in order to connect to WinRM nodes with SSL at all one has to pass the '--self-signed' parameter. This is not intuitive since the problem is not that the certificates are self-signed it is that we are not allowed to tell InSpec how to trust them. There is another command-line parameter (--insecure) that would be much more relevant here. Regardless, users should be able to configure a CA trust location to make SSL/TLS connections be validated correctly.

Nik08:

## Description

Added options to allow winrm SSL connection with certificates and in a passwordless way

## Related Issue

Closes #5659 
## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/cli.md
+++ b/docs-chef-io/content/inspec/cli.md
@@ -105,6 +105,12 @@ This subcommand has the following additional options:
     Specifies the bastion port if applicable.
 * ``--bastion-user=BASTION_USER``
     Specifies the bastion user if applicable.
+* ``--ca_trust_file=PATH_TO_CA_TRUST_FILE``
+    Allow remote scans with CA certificates (WinRM).
+* ``--client_cert=PATH_TO_CLIENT_CERTIFICATE``
+    Allow remote scans with client certificates without a password (WinRM).
+* ``--client_key=PATH_TO_CLIENT_KEY``
+    Allow remote scans with client certificates where client key is mandatory (WinRM).
 * ``--config=CONFIG``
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
Thanks!

docs-chef-io/content/inspec/cli.md
```diff
@@ -105,6 +105,12 @@ This subcommand has the following additional options:
     Specifies the bastion port if applicable.
 * ``--bastion-user=BASTION_USER``
     Specifies the bastion user if applicable.
+* ``--ca_trust_file=PATH_TO_CA_TRUST_FILE``
```
clintoncwolfe: Are these correct? Are they really --ca-trust-file (note hyphens, not underscores)?

It would be better if the options used hyphens, not underscores.

Nik08: Okay! I was using these options on train's and inspec cli options end with `_`. I think inspec does this conversion of `-` from cli options into `_` when it is interpreted within the logic in inspec before sending options to train.
Let me check. Thanks for pointing it out.

PR #5799, opened 2022-01-17
Fix for dependent profiles to run with --insecure

### LINKED ISSUE #5781: --insecure doesn't seem to work with dependent profiles
When using --insecure with inspec exec, if the profile depends on another profile via a url path e.g., https://, the --insecure flag does not seem to extend to the dependent profile's path.

Nik08:

## Description

Since `--insecure` flag was not accessible at fetcher level for dependent profiles, the runner configurations values are passed using requirement class, which is a class covering logic for dependencies.

## Related Issue

Fixes #5781 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/dependencies/requirement.rb
+++ b/lib/inspec/dependencies/requirement.rb
@@ -102,7 +102,8 @@ module Inspec
     end
 
     def fetcher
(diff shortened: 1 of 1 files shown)


clintoncwolfe (approved):
Thanks!

Issue #4323, opened 2019-07-23
Allow to override legacy vs grub2 detection


KrisShannon:
Add a way to force grub legacy on platforms that currently default to grub2.

(Vice-versa would probably also be nice but I don't personally need that)

## Context
We have many systems running in a virtual environment which when they were installed only worked with legacy grub.

These happen to include many CentOS 7 VM's

I want to use the `grub_conf` resource on these VM's but it is trying to interpret the legacy grub config as grub2

KrisShannon:
For the moment I have copied the code for `grub_conf` into a `grub_legacy_conf` custom resource (_why isn't there a way to do resource inheritance 😞_) and edited it to remove the grub2 parts.

KrisShannon:
In Slack @aaronlippold suggested that another resource using inheritance is probably the right answer for this.

Although a custom resource can't currently inherit from a core resource (or more specifically you can't inherit across registries) it would be possible to do this in core - a la `etc_hosts_deny`.

I'm thinking of working on this by adding two new resources to [lib/inspec/resources/grub_conf.rb](https://github.com/inspec/inspec/blob/v4.18.113/lib/inspec/resources/grub_conf.rb) - `grub2_conf` and `grub_legacy_conf`

The `grub_conf` would stay as is with the OS detection.

Does this sound like a reasonable solution?

KrisShannon:
The workaround in #4328 of using `.extend Module` isn't helpful for this specific case so it's still an issue for us;  but it's become much less important as we now only have one profile left which needs this and only when running against one particular server.

If it isn't fixed by the next time the stale bot comes around I'm not going to bother responding and just let it close.

Issue #1613, opened 2017-03-31
Resource class inheritance error


4-20ma:
### Description

I'm having trouble with custom resource class inheritance.

ResourceC inherits correctly from ResourceB, but ResourceA does not properly inherit from ResourceB.

It appears the resources are loaded alphabetically; inheriting from a resource with filename alphabetically before the current file loads properly whereas an error is generated for a filename later in the alphabet.

I can't figure out how to properly load the parent resource class (require, require_relative, include do not seem to work).

### InSpec and Platform Version

ruby 2.3.1p112
inspec-1.18.0
chefdk-1.2.22 with version-pinned inspec-1.11.0
OS X 10.12.3
$ brew --version
Homebrew 1.1.11-203-g1ba401039
Homebrew/homebrew-core (git revision 8c0ba; last commit 2017-03-30)

### Replication Case

Refer to [inspec-resource-error](https://github.com/4-20ma/inspec-resource-error).
### Possible Solutions

My work-around is to strategically rename the file of the parent class to load before the child class.

### Stacktrace

Included in my [inspec-resource-error](https://github.com/4-20ma/inspec-resource-error) and below:

```
    libraries/resource_a.rb:14:in `load_with_context': uninitialized constant #<Class:#<#<Class:0x007f82167d6910>:0x007f82167d6758>>::ResourceB (NameError)
    	from /Users/me/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/inspec-1.18.0/lib/inspec/profile_context.rb:146:in `instance_eval'
    	from /Users/me/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/inspec-1.18.0/lib/inspec/profile_context.rb:146:in `load_with_context'
    	from /Users/me/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/inspec-1.18.0/lib/inspec/profile_context.rb:135:in `load_library_file'
    	from /Users/me/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/inspec-1.18.0/lib/inspec/profile_context.rb:124:in `block in load_libraries'
    	from /Users/me/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/inspec-1.18.0/lib/inspec/profile_context.rb:122:in `each'
[...]

PR #1355, opened 2016-12-09
WIP: allow subclass of standard inspec resource

jeremymv2:
This fixes https://github.com/chef/inspec/issues/1300

It allows a subclass of existing resource, for example:
```
# profile/libraries/node.rb
class NodeAttributes < JsonConfig
  name 'node'
  def initialize
    super('/tmp/node.json')
  end
end
```
and standard, brand new resource like this:

```
class GordonConfig < Inspec.resource(1)
  name 'gordon_config'
  ...
end
```
to work.

Signed-off-by: Jeremy J. Miller <jm@chef.io>

PR #5816, opened 2022-01-28
CFINSPEC-15 Allows inheritance of core resource into the custom resource.

Vasu1105:

## Description

This code change fixes 2 issues
1) uninitialized constant error
2) undefined local variable or method error

For core resource inheritance in custom resources to work along with these code changes, the user needs to require the core resource. E.g. If I want to inherit the `json` resource, I need to add this line `require inspec\resources\json` at the top of the Custom resource. Inheritance of Core resource will inherit the functionality but the user has to configure the attributes like supports in the Custom resource as that won't be inherited.

Ref
#1355
#1300
#1613
#4323

Thanks @jeremymv2  and @KrisShannon 

Below is the custom resource which I have tested

```
require "inspec/resources/json"

class NodeAttributes < JsonConfig
  name 'node'

  def initialize
    super('/tmp/node.json')
  end

  def to_s
    "Node Json"
  end
end
```
And

```
require "inspec/resources/grub_conf"

class GrubLegacyConfig < GrubConfig
  name 'grub_legacy_conf'
  
   supports platform: "unix"

  def initialize(path = nil, kernel = nil)
    @conf_path = path || '/boot/grub/grub.conf'
    @version = 'legacy'
    @content = read_file(@conf_path)
    @kernel = kernel || 'default'
  end

  def to_s
    "Grub Legacy Conf"
  end
end
```
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/library_eval_context.rb
+++ b/lib/inspec/library_eval_context.rb
@@ -30,6 +30,8 @@ module Inspec
 
       c3 = Class.new do
         include Inspec::DSL::RequireOverride
+        include Inspec::Resources
+
         def initialize(require_loader)
           @require_loader = require_loader
           @inspec_binding = nil
--- a/lib/inspec/profile_context.rb
+++ b/lib/inspec/profile_context.rb
@@ -68,6 +68,7 @@ module Inspec
     end
 
     def reload_dsl
+      @resource_registry.merge!(Inspec::Resource.new_registry)
       @control_eval_context = nil
     end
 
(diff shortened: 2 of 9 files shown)


lib/inspec/library_eval_context.rb
```diff
@@ -30,6 +30,8 @@ def self.create(registry, require_loader)
 
       c3 = Class.new do
         include Inspec::DSL::RequireOverride
+        include Inspec::Resources
```
clintoncwolfe, line 33: 🥇 for tracking that down

clintoncwolfe (approved):
I love a spooky two-line change.

Issue #5826, opened 2022-02-07
Amazon uses Grub2


btrieger:
Amazon uses Grub2 and not legacy

# Describe the problem
When running grub_conf against amazon linux 2 it should return the expected results.

## Possible Solution
Update the amazon version to grub2 from legacy and fix conditional to check os so that it reaches amazon check.

PR #5827, opened 2022-02-07
Update Amazon to grub2

btrieger:
Amazon was using legacy grub_conf and the conditional could never be reached

## Description
When running inspec grub_conf against Amazon it uses legacy grub which results in failures. Amazon Linux 2 uses grub 2 and also will return true to os.redhat? so the previous conditional would never be reached. 

## Related Issue
#5826

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/grub_conf.rb
+++ b/lib/inspec/resources/grub_conf.rb
@@ -37,7 +37,7 @@ module Inspec::Resources
 
     def config_for_platform(path)
       os = inspec.os
-      if os[:name] == "amazon" 
+      if os[:name] == "amazon"
         if os[:release] == "2"
           @conf_path = path || "/boot/grub2/grub.cfg"
           @defaults_path = GRUB2DEFAULTS
--- a/lib/inspec/resources/grub_conf.rb
+++ b/lib/inspec/resources/grub_conf.rb
@@ -34,7 +34,7 @@ module Inspec::Resources
 
     def config_for_platform(path)
       os = inspec.os
-      if (( os.redhat? || os[:name] == "fedora") && ! (os[:name] == "amazon" ))
+      if (( os.redhat? || os[:name] == "fedora") &&  (os[:name] != "amazon" ))
         config_for_redhatish(path)
       elsif os.debian?
         @conf_path = path || "/boot/grub/grub.cfg"
--- a/lib/inspec/resources/grub_conf.rb
+++ b/lib/inspec/resources/grub_conf.rb
@@ -22,8 +22,8 @@ module Inspec::Resources
 
     include FileReader
 
-    GRUB2DEFAULTS = "/etc/default/grub"
-    GRUB2ENV = "/boot/grub2/grubenv"
+    GRUB2DEFAULTS = "/etc/default/grub".freeze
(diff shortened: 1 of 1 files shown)


lib/inspec/resources/grub_conf.rb
```diff
@@ -34,16 +37,18 @@ def initialize(path = nil, kernel = nil)
...
     def config_for_platform(path)
       os = inspec.os
-      if os.redhat? || os[:name] == "fedora"
+      if os[:name] == "amazon"
+        @conf_path = path || "/boot/grub2/grub.cfg"
+        @defaults_path = GRUB2DEFAULTS
+        @grubenv_path = path || GRUB2ENV
+        @version = "grub2"
+      elsif os.redhat? || os[:name] == "fedora"
         config_for_redhatish(path)
```
deric4: I think this change would work for Amazon Linux 2, but wouldn't  the default`@conf_path` for Amazon Linux 1, which uses `/boot/grub/grub.cfg` still be broken ([if this aws supportarticle is still correct](https://aws.amazon.com/premiumsupport/knowledge-center/revert-stable-kernel-ec2-reboot/#Update_the_default_kernel_in_the_GRUB_bootloader))? 

>  Amazon Linux 2 uses grub 2 and also will return true to os.redhat? so the previous conditional would never be reached

I think you're correct here and guessing this has always been broken for both `amazon` distros because:

1. the default `@conf_path` of `/etc/grub.conf` is incorrect for **Amazon Linux 1**
1. since os.family would equal `red-hot`, the `os[:release]` for **Amazon Linux 1** would fail the current guard clause in `config_for_redhatish()` along with incorrectly capturing **Amazon Linux 2**
    https://github.com/inspec/inspec/blob/bc6e62aa4d1d9b16646557e9418925ed2ecbbfed/lib/inspec/resources/grub_conf.rb#L53
    https://github.com/inspec/inspec/blob/6ba272e81b9ec9df8cd9285ad3a8be9df29a1319/test/helpers/mock_loader.rb#L34-L35

deric4 (commented):
Hey @btrieger, 

I haven't actually been able to test this to verify my theories, but this issue/pr caught my eye since I help maintain the [dev-sec/cis-dil-benchmark profile](https://github.com/dev-sec/cis-dil-benchmark) and know there's some `grub_conf` checks in there but why this wouldn't have been failing in testing or been reported.

Looks like we have custom library for `grub_conf` defined in the profile itself since this was merged in https://github.com/dev-sec/cis-dil-benchmark/pull/76  which I guess _partly_ explains why it hasn't been "breaking" 🤔

lib/inspec/resources/grub_conf.rb
btrieger: Interesting so maybe I need to add in a 2nd check to see if it is amazon 1 or 2 somehow to fix this? Presumably look at the release I guess. I have been working on a CIS amazon linux 2 test which is how I caught this.

btrieger: I added in a version check and use /boot/grub/grub.cfg for amazon linux 1. Also curious why debian has /boot/grub/grub.cfg and not /boot/grub2/grub.cfg as it is grub2.

PR #5830, opened 2022-02-08
Execute HTTP Request Locally

### LINKED ISSUE #5825: Restore local HTTP functionality on remote targets
## Detailed Description

Prior to 2.0, it was possible to instruct inspec to execute the HTTP resource on either the local or remote host. It appears that since 2.0 it is no longer possible to run the check from the local host when running on a remote machine.

## Context

Some use cases require us to make calls to endpoints from the local host instead of the remote one.

For example:
* Confirming that configured services on the remote host are accessible from other hosts, not just the remote host itself
* Calls to other endpoints (e.g., databases, cloud resources, etc.) that are not accessible from the remote machine directly
  * Some remote hosts exist in a "vacuum" and don't have direct access to the internet
  * Some API endpoints are locked down to only a specific few hosts and shouldn't be opened up to every machine in the network
* Some checks are easier to write or already exist as microservices and can be referenced in the control instead of reinventing the wheel or adding unnecessary complexity to a compliance check.

## Possible Implementation

Have a flag similar to "enable_remote_worker" (such as what existed prior to 2.0) but switch the functionality to "enable_local_worker" to specify which host should execute the HTTP request.

bryaend:
Fixes #5825

## Description

Restored the ability to execute an HTTP request from the local machine instead of the target machine

## Related Issue

Fixes #5825 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/http.rb
+++ b/lib/inspec/resources/http.rb
@@ -41,7 +41,7 @@ module Inspec::Resources
       end
 
       # Run locally if InSpec is ran locally and remotely if ran remotely
-      if inspec.local_transport?
+      if inspec.local_transport? || opts.key?(:enable_local_worker)
         @worker = Worker::Local.new(http_method, url, opts)
       else
         @worker = Worker::Remote.new(inspec, http_method, url, opts)


bryaend:
Did not contribute DCO, will submit new PR with proper documentation.

PR #5835, opened 2022-02-10
CFINSPEC-38 Adds new release branch for InSpec 5 release.

Vasu1105:

## Description

This is to update the expeditor configuration to get ready for InSpec 5 release.
## Related Issue

CFINSPEC-38
## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -74,7 +74,7 @@ github:
 release_branches:
   - main:
       version_constraint: 5.*
-  - 4-stable:
+  - inspec-4:
      version_constraint: 4.*
   # We need to ensure all configs are in place to appropriately support this branch
   # - expeditor-development:
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -76,12 +76,6 @@ release_branches:
       version_constraint: 5.*
   - 4-stable:
      version_constraint: 4.*
-  - 3-stable:
-     version_constraint: 3.*
-  - 2-stable:
-     version_constraint: 2.*
-  - 1-stable:
-     version_constraint: 1.*
   # We need to ensure all configs are in place to appropriately support this branch
   # - expeditor-development:
   #    version_constraint: 4.*
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -73,13 +73,15 @@ github:
 
 release_branches:
   - main:
+      version_constraint: 5.*
+  - 4-stable:
      version_constraint: 4.*
-  - 1-stable:
-     version_constraint: 1.*
-  - 2-stable:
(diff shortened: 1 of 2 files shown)


.expeditor/config.yml
```diff
@@ -73,13 +73,15 @@ github:
...
+      version_constraint: 5.*
+  - 4-stable:
      version_constraint: 4.*
-  - 1-stable:
-     version_constraint: 1.*
-  - 2-stable:
-     version_constraint: 2.*
   - 3-stable:
      version_constraint: 3.*
+  - 2-stable:
```
tas50: Since these are never getting released they should just get removed

.expeditor/config.yml
```diff
@@ -73,13 +73,9 @@ github:
 
 release_branches:
   - main:
+      version_constraint: 5.*
+  - Inspec-4:
```
tas50:
```suggestion
  - 4-stable:
```

.github/dependabot.yml
```diff
@@ -1,12 +1,26 @@
...
\ No newline at end of file
+  open-pull-requests-limit: 10
+- package-ecosystem: bundler
+  target-branch: "4-stable"
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package-ecosystem: bundler
+  target-branch: "Inspec-4"
```
tas50:
```suggestion
  target-branch: "4-stable"
```

Vasu1105: Decided to keep Inspec-4 so did the change where I have missed updating that.

.expeditor/config.yml
Vasu1105: Decided to name it Inspec-4 so did the same changes all over where it's required.

Nik08: How about`inspec-4-stable`? That might also provide us some more clarity over the state of the branch. Example: It could be `stable`, `pre` etc too in future maybe?
Also I feel maybe the branch name should be lowercase, not capitalised as a good practice :)

Nik08: @Vasu1105 How about `inspec-4-stable` ? That might also provide us some more clarity over the state of the branch. Example: It could be stable, pre etc too in future maybe?
And I feel the branch name should be lowercase, not capitalised as a good practice :)

Vasu1105: @Nik08 I just followed what Chef is following https://github.com/chef/chef/blob/main/.expeditor/config.yml#L62. I can lower the case which is a good point. I am not sure about whether we will have something like stable or pre etc. but it's just a matter of branch name here which I feel can be created in future for such a scenario(pre, beta, etc.) if exist.

PR #5875, opened 2022-02-23
CFINSPEC-70 Added resource_id attribute for the custom resources in the base class

Nik08:

## Description

1. Added resource_id attribute for the custom resources in the base class
2. Test cases covered

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/dsl_resource.md
+++ b/docs-chef-io/content/inspec/dsl_resource.md
@@ -48,6 +48,7 @@ The following attributes can be configured:
 - desc - Description of the resource (optional)
 - example - Example usage of the resource (optional)
 - supports - (Chef InSpec 2.0+) Platform restrictions of the resource (optional)
+- resource_id - A unique identifier for each node of resource. Should specify the implementation to determine it (optional)
 
 The following methods are available to the resource:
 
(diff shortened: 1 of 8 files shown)


lib/inspec/resource.rb
```diff
@@ -34,6 +34,11 @@ def self.supports(criteria = nil)
       Inspec::Resource.support_registry[key].push(criteria)
     end
 
+    def resource_id(value = nil)
+      resource_id_value = value || ""
+      @resource_id ||= resource_id_value
```
clintoncwolfe: This logic makes the instance variable write-once. That seems like a subtle bug to me.
```
res.resource_id('a')
res.resource_id  => 'a'
res.resource_id('b')
res.resource_id  => 'a'  # this is surprising
```

docs-chef-io/content/inspec/dsl_resource.md
```diff
@@ -48,6 +48,7 @@ The following attributes can be configured:
 - desc - Description of the resource (optional)
 - example - Example usage of the resource (optional)
 - supports - (Chef InSpec 2.0+) Platform restrictions of the resource (optional)
+- resource_id - A unique identifier for each node of resource. Should specify the implementation to determine it (optional)
```
clintoncwolfe: I think this needs to be separated out from the rest of the DSL. It should be clearer that this is not a DSL method you can call on your own as a DSL method, but a method you have to call as an instance method in your resource.

clintoncwolfe (approved):

PR #5892, opened 2022-03-02
add ruby test 3.1 in verify pipeline

jayashrig158:
Signed-off-by: jayashri garud <jgarud@msystechnologies.com>

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,5 @@
 # Order is important. The last matching pattern has the most precedence.
 
-*                     @inspec/inspec-core-team @chef/chef-infra-reviewers @chef/chef-infra-approvers @chef/chef-infra-owners
+*                     @inspec/inspec-core-team
 docs-chef-io/**       @inspec/inspec-core-team @inspec/docs-team
-.expeditor/           @chef/infra-packages
-*.md                  @inspec/inspec-core-team @inspec/docs-team @chef/docs-team
+*.md                  @inspec/inspec-core-team @inspec/docs-team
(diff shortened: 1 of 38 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,6 @@
 # Order is important. The last matching pattern has the most precedence.
 
-*                     @inspec/inspec-core-team
+*                     @inspec/inspec-core-team @chef/chef-infra-reviewers @chef/chef-infra-approvers @chef/chef-infra-owners
```
Vasu1105: @jayashrig158 as per my discussion with @clintoncwolfe  The groups that are attempted to be added here can't be added because they are not in the same github org. The @inspec github org has no communication with the @chef github org.

Vasu1105 (changes requested):

Vasu1105:
@jayashrig158 Why do we need both ruby 3.0 and 3.1 here?

.github/CODEOWNERS
jayashrig158: @Vasu1105 sure, I will revert changes.Thanks!

jayashrig158:
> @jayashrig158 Why do we need both ruby 3.0 and 3.1 here?
We have a requirement to add ruby 3.0/3.1 to make sure we are running tests on ruby 3.0/3.1. inspec repo has been listed in this requirement here -https://chefio.atlassian.net/browse/IPACK-54. I will discuss further on this with @PrajaktaPurohit.

Vasu1105 (approved):
On 3.0 there is timeout. Otherwise 3.1 looks good to me.

Vasu1105:
@clintoncwolfe I think this is good to go if we are planning to merge this one.

PR #5916, opened 2022-03-10
RESOURCE-312 extend filter table to lazy loading for resource instances

sathish-progress:
Extend filter table to allow access to resource level methods 

## Description
provided another attribute `lazy_instance` to accept symbols(instance methods) and procs

## Types of changes

## Checklist:


clintoncwolfe:
I've added tests and docs, and in testing found several small bugs and fixed them. One major issue I couldn't fix was that lazy_instance, when called with a Symbol name of an instance method, seems to act on a different resource instance each time it is called. That's a serious problem, and needs to be fixed, but I'm not sure what the problem is.

You can see it demonstrated when you run:
```
bundle exec inspec exec test/fixtures/profiles/filter_table --controls 2370_where_method
```

PR #5924, opened 2022-03-13
Enhance progress bar

aaronlippold:
- accounted for profile errors at runtime and report them as [ERROR]s
- accounted for `impact 0.0` controls report them as [NA] - aka Not Applicable
- added the control title to the output for end-user context
- added the control status to help those who can't see color easily
- fixed control status to account for passed,skipped and passed,skipped,failed


aaronlippold:
I do have a couple questions about the implementation of this reporter, it seems that we are repeating a lot of code here that we could make a single function, is there any reason we need to do this?

aaronlippold:
I didn't Refactor given I don't know the reason for the original repetition.

aaronlippold:
Secondly, I wonder if we could use pass by reference with the Notification object to perhaps be more efficient about our using a lot of copy data that we don't really need to copy. I might be mistaken but it could be another efficiency we could bring to the reporter.

Nik08:
@aaronlippold Yes you can refactor those methods and reduce the code repetitions. And the enhancements looks like the UI changes to this reporter and @clintoncwolfe could take a call on this.

Whereas, 
- NA and skipped does seem to refer to the same thing, as far as I can think from a profile perspective. So having an NA could be like a repetition. Because most of the skipped controls have `impact 0.0`. 
- Control status in this reporter was used only with an indicator without the whole word "PASSED, SKIPPED, FAILED etc" to keep consistency with the UI of CLI reporter. Though adding the whole wording would not hurt either. Again UI preference. 
- Test cases are failing, do have a look. 

Thanks

lib/plugins/inspec-streaming-reporter-progress-bar/lib/inspec-streaming-reporter-progress-bar/streaming_reporter.rb
```diff
@@ -9,104 +10,144 @@ module InspecPlugins::StreamingReporterProgressBar
...
+                       elsif @status_mapping[control_id].uniq.include?('skipped') && @status_mapping[control_id].uniq.one?
+                         'skipped'
+                       elsif @status_mapping[control_id].uniq.any? { |val| /"passed"||"skipped"/ =~ val }
+                         'passed'
                        end
 
       indicator = INDICATORS[control_status]
-      message_to_format = ""
-      message_to_format += "#{indicator}  "
-      message_to_format += control_id.to_s.lstrip.force_encoding(Encoding::UTF_8)
```
Nik08, line 97: `force_encoding(Encoding::UTF_8)` we should probably keep this, this is a safe check.

aaronlippold:
> @aaronlippold Yes you can refactor those methods and reduce the code repetitions. And the enhancements looks like the UI changes to this reporter and @clintoncwolfe could take a call on this.

Whereas, 
- NA and skipped does seem to refer to the same thing, as far as I can think from a profile perspective. So having an NA could be like a repetition. Because most of the skipped controls have `impact 0.0`. 

In the Compliance and Validation space, SKIP and Not Applicable are very differnt things and they communicate differnt things. When I, as a Security Assessor, say a test or group of tests is outside the scope of your Compliance Evaluation I am saying, reguardless off the results of the tests, you will not be required to Remediate those findings - at least not at the moment. This is Not Appliciable, and this is why we usueally have a Justification sub-description attached to that case. In this case, I have a defined reason as to why, reguardless of the test results, I have made this Control - aka set of tests - NA. 

When I just skip tests, without attaching an updated Impact and Justification statement, that becomes a Not Reviewed in the formal A&A spacce. Aka we don't have a definititive reason one way or the other as to why the tests were not evaluated. This then triggers the need to either re-run the tests or Manually review those requirments to close the loop on their status for the accrediation.

IMO, we should add these states to the Scheme, JSON and CLI reporter. In real world application this is always the first post-proccessing I have to do on the results data.

This is much like the difference between Waivered vs Attested Tests. Waivered tests can be SKIP or NA - Attested tests can be PASS or FAILED. 

- Control status in this reporter was used only with an indicator without the whole word "PASSED, SKIPPED, FAILED etc" to keep consistency with the UI of CLI reporter. Though adding the whole wording would not hurt either. Again UI preference. 

I mainly did this becase it makes the result more clear for those who can't see the color or may just find it easier to read then look for a small symbol. Added benifit is you can grep easily for only results of a kind.

- Test cases are failing, do have a look. 

I oddly could not find them pior to pushing the update, but other unit tests were also failing and the mock target was also broken so I assumed that main just needed a bit of love and I would get back to updating the mock data and tests after the first look. Where is the test file for this reporter?

Thanks

aaronlippold:
> `force_encoding(Encoding::UTF_8)` we should probably keep this, this is a safe check. 

Agreed, but we have to account for the case when we have NIL data as the force_encoding was throwing an error on the NIL class. We could use a turnary statment. 

Specically this happens when you have a broken control that thows a ruby or runtime error and does not have results or title etc.

Suggestions?

clintoncwolfe:
I don't think we can accept treating impact 0.0 controls as "N/A" - that's not an InSpec outcome, and while it may be used widely in-house at MITRE, it's not standard for InSpec. Feel free to make your own streaming reporter plugin that includes  your own interpretation of results.

I like adding the word PASSED/FAILED/SKIPPED in addition to the symbol.
I like adding the control title.

I tried error handling:
  * using one profile with a divide by zero in the top-level code, which errored out immediately before it got to the reporter stage, which I think is fine.
  * using a profile with a divide by zero within a control. That silently appeared to complete, with no control title or indication of error. I think that needs some work.

aaronlippold:
"I tried error handling:

using one profile with a divide by zero in the top-level code, which errored out immediately before it got to the reporter stage, which I think is fine.
using a profile with a divide by zero within a control. That silently appeared to complete, with no control title or indication of error. I think that needs some work."

* If you can add those two test profiles to the test suite I will work on both of them to make sure they pass and the error handling catches them. I think or hope the current error handling I put in in the last couple of commits may catch this now, I hope.

"I don't think we can accept treating impact 0.0 controls as "N/A" - that's not an InSpec outcome, and while it may be used widely in-house at MITRE, it's not standard for InSpec. Feel free to make your own streaming reporter plugin that includes your own interpretation of results."

* I agree that the concept of NA - and even Waived, passthrough or Attested controls - is not a current 4.x standard at the core level, however, with the intoduction of these concepts of NA, Waivers and Attestation, I think the community is slowly coming around to these concepts. Given we are moving into the 5.x lifecycle, I feel like this is the right time to start putting concepts of assesment catigories into standard use in the project. For those that don't want to use them, they are free to continue as before of course, for those that need them to accomplish their business, we should enable them.

* These core Compliance concepts need to move into the core capability and - IMO - all reporters should report waived (skipped for timefarme), attested (passed/failed for timeframe), skipped ( aka Not Reviewed), passed and Not Applicable.

We have seen a huge adoption of the InSpec capability over the last few years and our ability to cover all the areas of Compliance is now more important then ever. Adoption of InSpec as a core technology by VMWare, Google, Amazon, GitHub, Microsoft and perhaps even CIS and Tenable pushes us to meet these challenges.

We are currently working on three core capabily updates at the moment in the community:
- Minimal Remote Ohai support for inventory and context -> leading to standardization of `.target` data all reporters can use (2018/2019 target from 3.x) - learning from the Chef Remote Target work.
- addition of `passthrough` to the command line for data inject at the cicd or run-time level
- migration of the 'attestation' plugin to a sister implemenation of Waiver to round out the Manual and Policy challenge
  - Waived Controls -> [skipped, not reviewed]
  - Attested Controls -> [passed, failed]

aaronlippold:
> > `force_encoding(Encoding::UTF_8)` we should probably keep this, this is a safe check.
> 
> Agreed, but we have to account for the case when we have NIL data as the force_encoding was throwing an error on the NIL class. We could use a turnary statment.
> 
> Specically this happens when you have a broken control that thows a ruby or runtime error and does not have results or title etc.
> 
> Suggestions?

Tracked down the real issue, which was calling `gsub!` on a frozen string. Fixed that so encoding wasn't the true issue.

ejaronne:
> while it may be used widely in-house at MITRE, it's not standard for InSpec. 

Actually, MITRE can't take credit for this notion of Not Applicable. It is a long-standing concept in DISA STIGS and the RMF assessment vernacular:

image

image


Also, many tools have the notion of Not Applicable == WARN == Informational:
e.g., Nessus:
https://docs.tenable.com/nessus/Content/Images/HostsScanPlugins.png

e.g, AWS Security Hub's "Not_appliable":
https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_Compliance.html

aaronlippold:
Eugene brings up a good point, all the concepts talked about above from waivers to manual attestation to not applicable all come from the core RMF and standard compliance community Of not only the United States but the international bodies as well. 

If you're going to work with these communities and be successful, all these concepts must be first class citizens.

I think this would be a great topic for a presentation at ChefConf.

Lastly, in order to actually support concepts like continuous authorization and ongoing compliance you need all of these concepts in place and articulate about by the language. There's no other way to achieve that goal.

aaronlippold:
And I think we could achieve both here, all we would have to do is add a legacy flag, and for those that want to use their original pass skip fail, they just throw that flag

lib/plugins/inspec-streaming-reporter-progress-bar/lib/inspec-streaming-reporter-progress-bar/streaming_reporter.rb
```diff
@@ -28,85 +32,128 @@ class StreamingReporter < Inspec.plugin(2, :streaming_reporter)
...
       indicator = INDICATORS[control_status]
       message_to_format = ""
       message_to_format += "#{indicator}  "
-      message_to_format += control_id.to_s.lstrip.force_encoding(Encoding::UTF_8)
+      message_to_format += "#{control_id.to_s.lstrip.dup.force_encoding(Encoding::UTF_8)}  "
+      message_to_format += "#{title.gsub(/\n*\s+/, " ").to_s.force_encoding(Encoding::UTF_8)}  " if title
+      message_to_format += "#{full_description.gsub(/\n*\s+/, " ").to_s.force_encoding(Encoding::UTF_8)}  " unless title
       format_with_color(control_status, message_to_format)
+    rescue StandardError => e
+      raise "Exception in show_progress: #{e}"
```
Nik08, line 153: I think we should have error messages that could be also understood by an inspec user. 
So putting `show_progress` and `format_with_color` does help a developer to debug. But maybe not so useful for the user. Maybe use a message that is useful to both user and a developer? Makes sense?

aaronlippold: Happy to, what kind of message should we provide do you think? Parse `e` to bubble up the issue or put a case statment in that captures the N common errors that can happen at this stage and 'decode' them into nature language for the end user.

aaronlippold: Do you have a few examples of other places in the code base where you have done this so I could use them as a reference?

ejaronne: > I think we should have error messages that could be also understood by an inspec user.

When it comes to InSpec profiles, we use the following short phrases to qualify what each outcome means to the user, so they are not confused as to what has happened and "whose fault is it?":

Passed: "All sub-tests for this control have passed or were skipped"
Failed: "At least one sub-test failed for this control"
Not Applicable: "System exception or absent component"
Not Reviewed: "Can only be tested manually at this time"
Profile Error: "Errors running test - check profile run privileges or check with the author of profile."

Could these somehow be added to the enhanced reporting engine?  Thanks!

Nik08: @aaronlippold Before you made this change what I had here was `Exception in Progress Bar streaming reporter`. Nothing fancy but somehow clear to the user where it went wrong? So maybe mix of this and the new change of message that you have made. Sounds good? 

And for examples, you can search with `exception` keyword in this repo and you will get an idea how we have framed messages in case of exceptions.

rlakey:
It would be nice to have more reported statuses to get a more accurate view of the results of a scan. For example with only skip it is hard to know without digging through the results if a skip was the result of needing to do a manual review for a control or if was truly just skipped for other reasons such as the package wasn't installed, or it didn't detect the right OS, or some other condition.

ejaronne:
> It would be nice to have more reported statuses to get a more accurate view of the results of a scan. For example with only skip it is hard to know without digging through the results if a skip was the result of needing to do a manual review for a control or if was truly just skipped for other reasons such as the package wasn't installed, or it didn't detect the right OS, or some other condition.

Hence why we distinctly do it this way:
Not Applicable: "System exception or absent component"
Not Reviewed: "Can only be tested manually at this time"
Profile Error: "Errors running test - check profile run privileges or check with the author of profile."

Try the sample "Triple Overlay Example" at https://heimdall-lite.mitre.org/, and look at controls:

- V-61409 (Manual Review)
- V-61451 (Not Applicable due to absent item)
- V-61749 (Not Applicable dues to a system exception)

clintoncwolfe:
OK. I see there is interest in finer-grained outcomes, and I think that's great. We can explore what MITRE has done, see what other customers want, and look into different approaches to possibly adopting something in InSpec 6 - possibly sooner on an optional or plugin basis in InSpec 5. 

However, new test outcomes is a big change, and we can't sneak it in a progress bar enhancement PR :-). It deserves its own discussion and PR(s).

ejaronne:
> However, new test outcomes is a big change, and we can't sneak it in a progress bar enhancement PR :-). It deserves its own discussion and PR(s).

To clarify, I don't think Aaron's intent is to change the output of tests (individual describe blocks), rather the output of the InSpec *control* status based on the one or more describe-block-test outcomes and impact level assigned/calculated to the control.

PR #5926, opened 2022-03-15
changed --diff flag to be 'false' by default

wdower:
Signed-off-by: Will Dower <wdower@mitre.org>

Changed --diff to be 'false' by default. The flag (and --no-diff) otherwise work the same.

## Description
Most InSpec devs do not want the diff comparison in their cli output -- too cluttered. Seems reasonable to make --no-diff the default behavior and only show the full diff output if specified.

## Types of changes

## Checklist:


clintoncwolfe:
Instead of changing the default for everyone, which changes behavior perhaps surprisingly, could you just add the `no_diff` option to your ~/.inspec/config.json cli options?

PR #5948, opened 2022-03-23
fix(service): ensure `installed: true` isn't always returned on FreeBSD

### LINKED ISSUE #5947: [Regression] FreeBSD service testing always returns `installed: true` even when the service is not available
# Describe the problem

A regression was introduced by #5606, FreeBSD service testing now always returns `installed: true`, which is a problem when checking for `it { should_not be_installed }`.

https://github.com/saltstack-formulas/vault-formula/blob/e9742b056dce98d91b17639851d8257de61c0f90/test/integration/install_binary/controls/vault_spec.rb#L16-L20

```ruby
describe service('vault') do
  it { should_not be_enabled }
  it { should_not be_running }
end
```

This was working fine some time back but failed in a recent CI run on both FreeBSD 13.0 and 12.3.

https://github.com/saltstack-formulas/vault-formula/actions/runs/2026994551

```ruby
  Service vault
     ×  is expected not to be installed
     expected Service vault not to be installed
     ✔  is expected not to be enabled
     ✔  is expected not to be running
```

## Possible Solution

As I mentioned in these comments:

* https://github.com/saltstack-formulas/vault-formula/pull/54#issuecomment-1076089784
* https://github.com/saltstack-formulas/vault-formula/pull/54#issuecomment-1076107888

We're using our own fork of this repo and I've already tested out this fix:

* https://gitlab.com/saltstack-formulas/infrastructure/inspec/-/commit/ba8ee70a7061b75d76e78d89bdf33ee5038ce587

It's working out fine:

* https://github.com/myii/vault-formula/runs/5666487007?check_suite_focus=true

I'd already started a PR here, where the notes stated that an issue needed to be created first.

myii:
Signed-off-by: Imran Iqbal <iqbalmy@hotmail.com>

A regression was introduced by #5606, FreeBSD service testing now always returns `installed: true`, which is a problem when checking for `it { should_not be_installed }`.

## Description

I've explained the problem in #5947.

Worked out the fix when looking through 41ea0da4a99f3b55b1be182c110a2442e7de8447.

https://github.com/inspec/inspec/blob/26e197a3a506f596152c0dc30d6a9a5ca6fdd811/lib/inspec/resources/service.rb#L165-L171

Did a diff of `BSDInit` and `FreeBSD10Init`, found this block missing from the latter:

https://github.com/inspec/inspec/blob/26e197a3a506f596152c0dc30d6a9a5ca6fdd811/lib/inspec/resources/service.rb#L496-L510

That's the portion which deals with ensuring `installed` resolves as `false`, particularly the last two lines.  Added that to our fork of InSpec and confirmed that our CI testing is working again (as explained at the bottom of #5947).  Note, that CI covers multiple suites, which test FreeBSD services for both `it { should_not be_installed }` and `it { should be_installed }`.

CC: @zofrex.

## Related Issue

* Fix regression introduced in 41ea0da4a99f3b55b1be182c110a2442e7de8447 (#5606).
* Close #5947.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/service.rb
+++ b/lib/inspec/resources/service.rb
@@ -537,9 +537,22 @@ module Inspec::Resources
     end
 
     def info(service_name)
+      # `service -l` lists all files in /etc/rc.d and the local startup directories
+      # % service -l
+      # accounting
+      # addswap
+      # adjkerntz
+      # apm
+      # archdep
+      cmd = inspec.command("#{service_ctl} -l")
+      return nil if cmd.exit_status != 0
+
+      # search for the service
+      srv = /^#{service_name}$/.match(cmd.stdout)
+      return nil if srv.nil? || srv[0].nil?
+
       # check if service is enabled
       cmd = inspec.command("#{service_ctl} #{service_name} enabled")
-
       enabled = cmd.exit_status == 0
 
       # check if the service is running


clintoncwolfe (approved):
Thanks for catching this and the thorough explanation and fix!

lib/inspec/resources/service.rb
```diff
@@ -537,6 +537,22 @@ def initialize(service_name, service_ctl = nil)
...
+      cmd = inspec.command("#{service_ctl} -e")
+      return nil if cmd.exit_status != 0
+
+      # search for the service
+
+      srv = %r{^.*/(#{service_name}$)}.match(cmd.stdout)
+      return nil if srv.nil? || srv[0].nil?
+
       # check if service is enabled
       cmd = inspec.command("#{service_ctl} #{service_name} enabled")
```
zofrex: This is where we'd bail to match the behaviour of `BSDInit` - instead of `enabled = cmd.exit_status == 0` this would be changed to:

```
return nil if cmd.exit_status != 0
```

zofrex:
Thanks for tagging me, @myii. I took a look into what's going on here, and I've discovered that this is actually broken in both FreeBSD service classes, but in different ways!

In the base BSDInit class (for versions 9 and below) it cannot detect that a service is installed if it is not enabled:

```
 ×  installed, not enabled or running service: Service watchdogd (1 failed)
     ×  Service watchdogd is expected to be installed
     expected that `Service watchdogd` is installed
     ✔  Service watchdogd is expected not to be enabled
     ✔  Service watchdogd is expected not to be running
```

In FreeBSD10Init it cannot detect that a service that is not installed (as you discovered):

```
  ×  not installed service: Service vault (1 failed)
     ×  Service vault is expected not to be installed
     expected Service vault not to be installed
     ✔  Service vault is expected not to be enabled
     ✔  Service vault is expected not to be running
```

Your fix is therefore unfortunately not 100% correct, as it copies code from BSDInit to FreeBSD10Init which is also wrong (just in the opposite way).

Specifically, `service -e` lists _enabled_ services, not _installed_ services, so the check is redundant (the later check does this anyway) and isn't checking for the right thing (if the service is installed).

If we want FreeBSD10Init to match the behaviour of BSDInit, this could be done more easily by adding a `return nil` if the `service <foo> enabled` command returns 1 (that is, if the service is not enabled, assume it is also not installed). I've added a comment inline to indicate where that would happen.

If we want to separately check if a service is installed, independently of whether it's enabled, I'm not 100% sure of the best way to do that. But it looks like `service -l` is one way to do that: this command lists all installed services (somewhat similar to `service -e`, except it doesn't prefix them with the path to the init script).

I think it would be nicer to use a command that interrogates whether a single service is installed rather than grepping a list of all services (similarly to `service foo enabled` vs `service -e`), if such a command exists, because that's simpler and less error prone. But if no such command exists, it does look like `service -l` would be suitable.

myii:
Thanks for the excellent feedback, @zofrex.  I've gone with using `service -l` as you've suggested and that I've found [elsewhere](https://github.com/saltstack/salt/blob/1b7611958342ba48a842363e3717f8c67439b2f8/salt/modules/freebsdservice.py#L398).  Its output also simplifies the matching:

```diff
-      srv = %r{^.*/(#{service_name}$)}.match(cmd.stdout)
+      srv = /^#{service_name}$/.match(cmd.stdout)
```

> If we want `FreeBSD10Init` to match the behaviour of `BSDInit`, this could be done more easily by adding a `return nil` if the `service <foo> enabled` command returns 1 (that is, if the service is not enabled, assume it is also not installed). I've added a comment inline to indicate where that would happen.

Worked through the various possibilities, to ensure the changes are working:

Installed | Enabled | Running | Situation | Outcome
:-: | :-: | :-: | --- | :-:
N | N | N | Possible | Done
N | N | Y | Not possible | -
N | Y | N | Not possible | -
N | Y | Y | Not possible | -
Y | N | N | Possible | Fixed
Y | N | Y | Possible | Done
Y | Y | N | Possible | Done
Y | Y | Y | Possible | Done

Since a service could be running, even if disabled, the rest of the block is doing the right thing -- no need to bail out after checking if the service is enabled.

I'd welcome any further feedback you've got -- thanks again for the review.

lib/inspec/resources/service.rb
myii: As mentioned in my [last comment](https://github.com/inspec/inspec/pull/5948#issuecomment-1078472882), we wouldn't want to bail out here, since the service should still be checked to see if it's running.

zofrex:
This looks great to me, @myii! My only note (and do bear in mind I am not a maintainer for this project or anything, this is just one random person on the internet's opinion) is that it would be nice to have tests for this.

clintoncwolfe (approved):

PR #5959, opened 2022-03-28
CFINSPEC-94: Enhance user resource

ahasunos:
Signed-off-by: Sonu Saha <sonu.saha@progress.com>

## Related Issue

**CFINSPEC-94: Enhance user resource**

## Description

Using the `user` resource
Given that the user has called the `user` resource
then the resource allows to test with new matchers and properties.

#### New Matchers:
- belong_to_group
- belong_to_primary_group
- have_uid
- have_home_directory
- have_login_shell
- have_authorized_key
 #### New Properties
- encrypted_password
- minimum_days_between_password_change
- maximum_days_between_password_change

## Types of changes

## Checklist:


lib/inspec/resources/users.rb
```diff
@@ -314,6 +336,59 @@ def credentials
...
+
+    # Helper method for encrypted_password property
+    def encrypted_password_info
+      # check if getent is available on the system
+      bin = find_getent_utility
+
+      # fetch details of the passwd file for the current user using getent
+      cmd = inspec.command("#{bin} shadow #{@username}")
+
+      raise Inspec::Exceptions::ResourceSkipped, "Cannot view encrypted_password information" if cmd.exit_status != 0
```
Vasu1105: Does all non-zero exit status refer to not having permission to view encrypted_password information? Is it possible to view the exact error message to the User?

lib/inspec/resources/users.rb
```diff
@@ -314,6 +336,59 @@ def credentials
 
       @cred_cache = @user_provider.credentials(@username) unless @user_provider.nil?
     end
+
+    # helper method for has_authorized_key matcher
+    # checks for key is present as part of authorized keys on the server
+    def authorized_key(key)
+      # cat is used in unix system to display content of file; similarly type is used for windows
+      inspec.os.windows? ? bin = "type" : bin = "cat"
```
Vasu1105: you can do something like this bin = inspec.os.windows? ? "type" : "cat"

lib/inspec/resources/users.rb
```diff
@@ -314,6 +336,59 @@ def credentials
 
       @cred_cache = @user_provider.credentials(@username) unless @user_provider.nil?
     end
+
+    # helper method for has_authorized_key matcher
+    # checks for key is present as part of authorized keys on the server
+    def authorized_key(key)
```
Vasu1105: shouldn't it be authorized_key?(key) as it returns true and false if I am not wrong here?

lib/inspec/resources/users.rb
```diff
@@ -258,36 +255,61 @@ def passwordage
...
-      Inspec.deprecate(:resource_user_serverspec_compat, "The user resource `has_login_shell?` matcher is deprecated. Please use `its('shell')`.")
       shell == compare_shell
     end
 
-    def has_authorized_key?(_compare_key)
-      Inspec.deprecate(:resource_user_serverspec_compat, "The user resource `has_authorized_key?` matcher is deprecated. There is no currently implemented alternative")
-      raise NotImplementedError
+    # has_authorized_key matcher: compatibility with serverspec
+    def has_authorized_key?(compare_key)
+      authorized_key(compare_key)
```
Vasu1105: According to me 
we don't require another method call here.

lib/inspec/resources/users.rb
ahasunos: Raised different message appending stderr

lib/inspec/resources/users.rb
ahasunos: Made the changes accordingly.

lib/inspec/resources/users.rb
ahasunos: changed `authorized_key(key)` to `get_authorized_keys` which would return the key/keys from the authorized_keys path.
Then the test for key is valid or not is done in `has_authorized_key?` method

lib/inspec/resources/users.rb
ahasunos: `authorized_key(compare_key)` changed to
```
# get_authorized_keys returns the list of key, check if given key is included.
get_authorized_keys.include?(compare_key)
```

lib/inspec/resources/users.rb
```diff
@@ -258,36 +255,64 @@ def passwordage
...
+    # belongs_to_group matcher: compatibility with serverspec
+    def belongs_to_group?(group_name)
+      groupsname = groups
+      is_valid_group = false
+      groupsname.each do |g|
+        if g == group_name
+          is_valid_group = true
+          break
+        end
+      end
```
clintoncwolfe: Please use groups.include?(group_name)

lib/inspec/resources/users.rb
```diff
@@ -314,6 +339,53 @@ def credentials
...
+      # auth_path gets assigned with the valid path for authorized_keys
+      auth_path = ""
+
+      # possible paths where authorized_keys are stored
+      %w{~/.ssh/authorized_keys ~/.ssh/authorized_keys2}.each do |path|
+        if inspec.command("#{bin} #{path}").exit_status == 0
+          auth_path = path
+          break
+        end
+      end
```
clintoncwolfe: No need for any of this. To simply read a file, please use `inspec.file(path).content` and `inspec.file(path).exists?`

lib/matchers/matchers.rb
```diff
@@ -340,3 +340,16 @@ def format_expectation(negate)
...
+  match do |user|
+    user.belongs_to_primary_group?(group)
+  end
+end
+
+RSpec::Matchers.define :belong_to_group do |group|
+  match do |user|
+    user.belongs_to_group?(group)
+  end
+end
```
clintoncwolfe: Why are these being defined here in the universal matchers file?

clintoncwolfe (changes requested):
Need to change the way that you are reading the auth keys file, but otherwise looks pretty good.

lib/inspec/resources/users.rb
ahasunos: Yep! Thanks for pointing this out.

lib/matchers/matchers.rb
ahasunos: In order to be able to support the `belong_to_group` and `belong_to_primary_group` matchers, I felt the need to define the custom matchers as `belong_to_` matcher didn't exist as part of InSpec or RSpec. 

Since other custom matchers were defined here, I thought this would be the right place to define them. Am I missing something here?

lib/inspec/resources/users.rb
ahasunos: inspec.command is used over inspec.file because inspec.file requires absolute path

lib/matchers/matchers.rb
ahasunos: Moved the custom matchers to `user.rb`

clintoncwolfe (approved):
Thanks!

dkumaras (commented):
@ahasunos - please check the suggested review changes and confirm.

ahasunos:
@dkumaras It seems you missed to sign-off the commit which is leading to the failure of DCO test.

dkumaras:
> ple

@ahasunos , @Vasu1105 , @Nik08 , @clintoncwolfe - I have introduced a new formatting for the parameters value description following a code block. Pls do check and share your thoughts.

docs-chef-io/content/inspec/resources/user.md
```diff
@@ -43,108 +44,126 @@ A `user` resource block declares a user name, and then one (or more) matchers:
...
-The `gid` property tests the group identifier:
+The `gid` property tests the group identifier.
 
     its('gid') { should eq 1234 }
 
-where `1234` represents the user identifier.
-The `gid` option is only available on Linux and will return `nil` for Windows os.
+> Where `1234` represents the user identifier.
+>
+> The `gid` option is only available on Linux and returns `nil` for Windows OS (Operating System).
```
clintoncwolfe: The indentation should not extend to this sentence.

docs-chef-io/content/inspec/resources/user.md
```diff
@@ -43,108 +44,126 @@ A `user` resource block declares a user name, and then one (or more) matchers:
...
-where `0` represents the maximum number of days.
+> where `0` represents the maximum number of days.
+
+### minimum_days_between_password_change
+
+The `minimum_days_between_password_change` property is equivalent to `mindays`.
+
+    its('minimum_days_between_password_change') { should eq 0 }
+
+> where `0` represents the maximum number of days.
```
ahasunos: Here's a typo: minimum instead of maximum.

> _where `0` represents the ~~maximum~~ minimum number of days_

docs-chef-io/content/inspec/resources/user.md
dkumaras: sure

dkumaras (commented):
Pls confirm the changes

PR #5961, opened 2022-03-30
CFINSPEC-80: Add mail alias resource

ahasunos:
✅ Signed-off-by: Sonu Saha <sonu.saha@progress.com> 

## Related Issue
**⚡️ CFINSPEC-80: Add `mail alias` resource**

## Description
Using the `mail_alias` resource 
Given that the user has called the `mail_alias` resource
and given that the system is unix-based
then the resource allows to test mail alias parameters

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/mail_alias.rb
+++ b/lib/inspec/resources/mail_alias.rb
@@ -32,16 +32,11 @@ module Inspec::Resources
       "mail_alias #{resource_id}"
     end
 
-    def is_aliased_to?(alias_value)
-      # positive or negative expectations specific to this resource instance
-      true
-    end
+    def aliased_to?(alias_value)
+      cmd = inspec.command("cat /etc/aliases | grep #{@alias_key}")
+      raise Inspec::Exceptions::ResourceFailed, "#{@alias_key} is not a valid key in the aliases" if cmd.exit_status.to_i != 0
 
-    private
-
-    # Methods to help the resource's public methods
-    def helper_method
-      # Add anything you need here
+      alias_value == cmd.stdout.split(":").map(&:strip)[1]
     end
   end
 end
--- a/lib/inspec/resources/mail_alias.rb
+++ b/lib/inspec/resources/mail_alias.rb
@@ -1,5 +1,4 @@
 require "inspec/resources/command"
-
 module Inspec::Resources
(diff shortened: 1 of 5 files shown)


lib/inspec/resources/mail_alias.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # resource appearance in test reports.
+    def to_s
+      "mail_alias #{resource_id}"
+    end
+
+    # aliased_to matcher checks if the given alias_value is set to the initialized alias_key
+    def aliased_to?(alias_value)
+      # /etc/aliases if the file where the alias and its value(s) are stored
+      cmd = inspec.command("cat /etc/aliases | grep #{@alias_key}")
```
clintoncwolfe: You can't just grep for the alias key, because you may catch entries that the alias_key is being aliased-to. You need to look for lines that start with the alias key and and with a colon.

clintoncwolfe (changes requested):
Need to tighten up the regex passed to grep, otherwise looks good!

lib/inspec/resources/mail_alias.rb
ahasunos: `cmd = inspec.command("cat /etc/aliases | grep '^#{@alias_key}:'")`
Does this look better? The grep matches for sentence starting with the alias followed by colon.

clintoncwolfe: I think you will need egrep, not just grep.

ahasunos: The regex meta characters which are required to be escaped while using grep are `'{', '}', '(', ')', '|', '+', '?'`, since none of them are used as part of `'^#{@alias_key}:'`, I think grep and egrep would work the same in this situation. If we would like to be more strict with the regex we could still go with `grep -e` option instead of `egrep` since _Direct invocation as either egrep or fgrep is deprecated_  Source: [gnu.org](https://www.gnu.org/software/grep/manual/grep.html#:~:text=Direct%20invocation%20as%20either%20egrep,on%20them%20to%20run%20unmodified.)

clintoncwolfe (approved):
Thanks!

dkumaras (commented):
Pls confirm the changes suggested.

PR #5981, opened 2022-04-12
CFINSPEC-93: Enhance `service` resource

ahasunos:
✅ Signed-off-by: Sonu Saha <sonu.saha@progress.com>

## Related Issue
**CFINSPEC-93: Enhance `service` resource with `be_monitored_by` and `have_start_mode` matchers**

## Description

- Using the `service` resource
Given that the user has called the `service` resource
then the resource allows to test with new matchers viz. `be_monitored_by` and `have_start_mode` matchers
- Syntax
  ```
  it { should be_monitored_by("god") }
  it { should have_start_mode("Manual") }
  ```
- The monitoring tools currently supported by this resource are: `monit` and `god`.

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/service.md
+++ b/docs-chef-io/content/inspec/resources/service.md
@@ -11,55 +11,62 @@ platform = "os"
     parent = "inspec/resources/os"
 +++
 
-Use the `service` Chef InSpec audit resource to test if the named service is installed, running and/or enabled.
+Use the `service` Chef InSpec audit resource to test whether the installation of the named service is successful and enabled.
 
-Under some circumstances, it may be necessary to specify the service manager by using one of the following service manager-specific resources: `bsd_service`, `launchd_service`, `runit_service`, `systemd_service`, `sysv_service`, or `upstart_service`. These resources are based on the `service` resource.
+It may be necessary to specify the service manager by using one of the following service manager-specific resources: `bsd_service`, `launchd_service`, `runit_service`, `systemd_service`, `sysv_service`, and `upstart_service`. These resources are based on the `service` resource.
 
 ## Availability
 
 ### Installation
 
-This resource is distributed along with Chef InSpec itself. You can use it automatically.
+The Chef InSpec distributes this resource.
 
 ### Version
 
-This resource first became available in v1.0.0 of InSpec.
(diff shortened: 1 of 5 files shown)


lib/inspec/resources/service.rb
```diff
@@ -893,4 +917,73 @@ def select_service_mgmt
...
+
+      monitoring_info = utility_cmd.stdout.strip
+      monitoring_info =~ /^#{service_name}: up/
+    end
+
+    private
+
+    def find_utility_or_error(utility_name)
+      super
+    end
```
clintoncwolfe: This definition isn't needed - you should automatically inherit the superclass definition.

lib/inspec/resources/service.rb
```diff
@@ -893,4 +917,73 @@ def select_service_mgmt
...
+        end
+      end
+      is_monitored
+    end
+
+    private
+
+    def find_utility_or_error(utility_name)
+      super
+    end
```
clintoncwolfe: This should not be needed.

clintoncwolfe (approved):
Just a minor quibble about not needing to re-implement an inherited method, otherwise looks good!

dkumaras (commented):
Please reply to my one query

docs-chef-io/content/inspec/resources/service.md
```diff
@@ -70,62 +77,93 @@ The following examples show how to use this Chef InSpec audit resource.
       it { should_not be_installed }
       it { should_not be_running }
     end
+```
 
-### Test Unix System V run levels
+### Test Unix SystemV run levels
 
 On targets that are using SystemV services, the existing run levels can also be checked:
```
dkumaras, line 84: Which are using SystemV services - run levels?

the existing run levels can also be checked: - how?

What does - on targets mean?

dkumaras: @ahasunos - pls check now

ahasunos: Here **"on target"** means the target instance (could be any system) which is running on an older version of Unix. The newer Unix system supports systemd. It can be checked as shown in the example following the header.

PR #5991, opened 2022-04-18
CFINSPEC-92: Enhance `processes` resource

ahasunos:

## Related Issue
**CFINSPEC-92: Enhance `processes` resource with `count` property and `be_running` matcher**

## Description
- Using the `processes` resource
Given that the user has called the `processes` resource
then the resource allows to test with new property & matcher
- Property: `count`
- Matcher: `be_running`
- Syntax:
  ```
  describe processes(Regexp.new("some_process")) do
      it { should be_running }
      its("count") { should eq 3 }
  end
  ```

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/processes.rb
+++ b/lib/inspec/resources/processes.rb
@@ -62,11 +62,13 @@ module Inspec::Resources
 
     # Matcher to check if the process is running
     def running?
-      # Check if Regex needs to be tightened.
-      # States value can be as:
-      # for Unix: R, R< or R+
-      # for Windows "True" or "False"
-      states.any? and !!(states[0] =~ /True/ || states[0] =~ /^R+/)
+      # A process is considered running if:
+      # unix: it is in running(R) state or either of sleep state(D: Uninterruptible or S: Interruptible)
+      # windows: it is responding i.e. state is True.
+
+      # Other codes like <(high priorty), N(low priority), +(foreground process group) etc. may appear after the state code in unix.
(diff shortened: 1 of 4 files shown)


lib/inspec/resources/processes.rb
```diff
@@ -60,6 +60,15 @@ def list
       @list
     end
 
+    # Matcher to check if the process is running
+    def running?
+      # Check if Regex needs to be tightened.
+      # States value can be as:
+      # for Unix: R, R< or R+
+      # for Windows "True" or "False"
+      states.any? and !!(states[0] =~ /True/ || states[0] =~ /^R+/)
```
clintoncwolfe: That's wrong. You probably don't want literally R for running - Sleep is probably OK too. Interpret this as "is the process alive, not dead"

clintoncwolfe (changes requested):
I think the logic is wrong on the be_running matcher.

lib/inspec/resources/processes.rb
ahasunos: Changed logic to consider sleep state (both uninterruptible and interruptible) to be running.

clintoncwolfe (approved):

dkumaras (commented):
Pls check and confirm

ahasunos:
Looks good to me. @dkumaras

dkumaras (approved):
Thanks Sonu

Vasu1105 (approved):

Issue #5809, opened 2022-01-21
Service checks failing on Amazon linux 2022


bot11:
# Service checks are failing on Amazon linux 2022
Inspec fails to check service on al2022-ami images.
Possibly because inspec is using initctl instead of systemd: https://github.com/inspec/inspec/blob/948b274d4c7ac312fea829117a88b3cd3f61a2fc/lib/inspec/resources/service.rb#L185

Tests as below:
* Sample ec2 vm with  Amazon linux 2022 with  sshd service is running:
```
[root@ip-172-31-12-188 ~]# uname -a
Linux ip-172-31-12-188.us-west-2.compute.internal 5.10.75-82.359.amzn2022.x86_64 #1 SMP Mon Nov 15 18:04:59 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
[root@ip-172-31-12-188 ~]# systemctl status sshd.service
● sshd.service - OpenSSH server daemon
     Loaded: loaded (/usr/lib/systemd/system/sshd.service; enabled; vendor preset: enabled)
     Active: active (running) since Fri 2022-01-21 11:15:48 UTC; 16min ago
       Docs: man:sshd(8)
             man:sshd_config(5)
```

*  Sample inspec test to check against sshd:
```ruby
[root@ip-172-31-12-188 ~]# cat service-test.rb
describe service('sshd') do
  it { should be_installed }
  it { should be_enabled }
  it { should be_running }
end
```

* Testing:
```bash
[root@ip-172-31-12-188 ~]# /opt/chef/bin/inspec --version
4.52.9
[root@ip-172-31-12-188 ~]# /opt/chef/bin/inspec exec service-test.rb

Profile: tests from service-test.rb (tests from service-test.rb)
Version: (not specified)
Target:  local://

  Service sshd
     ×  is expected to be installed
     expected that `Service sshd` is installed
     ×  is expected to be enabled
     expected that `Service sshd` is enabled
     ×  is expected to be running
     expected that `Service sshd` is running

Test Summary: 0 successful, 3 failures, 0 skipped
[root@ip-172-31-12-188 ~]#
```

PR #5998, opened 2022-04-20
Fix for issue 5809 Service checks failing on Amazon linux 2022

ahasunos:
✅  Signed-off-by: Sonu Saha <sonu.saha@progress.com>
## Related Issue
Issue #5809: Service checks failing on Amazon linux 2022

## Description
**CFINSPEC-186 - Issue 5809: Service checks are failing on Amazon linux 2022**
The `service` resource checks were failing on Amazon Linux 2 as InSpec was using `initctl` instead of `systemd` because of the following logic:

https://github.com/inspec/inspec/blob/948b274d4c7ac312fea829117a88b3cd3f61a2fc/lib/inspec/resources/service.rb#L185

The above logic fails because on `os[:release]` both Amazon Linux 1 and Amazon Linux 2 gives the output as their release years (eg: 2018.03 or 2022) thus failing to differentiate Amazon Linux 2022 as Amazon Linux 2 and triggering service with `initctl` instead of `systemd`.

**The patch for this is done by checking the availability of the utility on the system.**
``` 
  if inspec.command("initctl").exist? || inspec.command("/sbin/initctl").exist?
    Upstart.new(inspec, service_ctl)
  else
    Systemd.new(inspec, service_ctl)
  end
```

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/resources/service_test.rb
+++ b/test/unit/resources/service_test.rb
@@ -124,19 +124,24 @@ describe "Inspec::Resources::Service" do
     _(resource.params).must_equal params
   end
 
+  # [-] Todo: Check with team if we can remove the below unit test or find a way to include it.
+  # The below test is not required because we cannot differentiate between amazon and amazon2 during testing.
+  # After all, the differentiation is not based on the version but based on the utility available on the system.
+  # However, the service resource works perfectly fine for the actual instance of Amazon and Amazon2 Linux.
+
   # Amazon Linux
-  it "verify amazon linux service parsing" do
-    resource = MockLoader.new(:amazon).load_resource("service", "ssh")
-    params = Hashie::Mash.new({})
-    _(resource.type).must_equal "upstart"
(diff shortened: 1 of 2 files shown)


test/unit/resources/service_test.rb
```diff
@@ -124,19 +124,24 @@
     _(resource.params).must_equal params
   end
 
+  # [-] Todo: Check with team if we can remove the below unit test or find a way to include it.
```
ahasunos, line 127: Need comment on this.

_I have commented out the below unit this because imo_

> _The below test is not required because we cannot differentiate between amazon and amazon2 during testing. After all, the differentiation is not based on the version but based on the utility available on the system. However, the service resource works perfectly fine for the actual instance of Amazon and Amazon2 Linux._

ahasunos:
An attempt to fix the issue was done in PR #2901

clintoncwolfe (approved):

PR #6011, opened 2022-04-25
CFINSPEC-121 Parallel forking implementation

Nik08:
## Description

Second PR after #5975 to the parallel implementation which includes,

1. Added `-j` option to specify number of jobs/workers to run for parallel processing.
2. Implementation of forking in unix and reporting of child process status.
3. Implementation of spawning process in windows.
4. ERB templating feature for options file.
5. Options file can be used as a shell script or powershell script when options within the file needs to be dynamically loaded.
6. Child-status streaming reporter added.
7. Child status reporting mechanism added.

**NOTE** 
For code review purpose this PR is opened against last PR's branch `nm/inspec-parallel-cmd`, but before merging need to change it to the `main` branch.
## Types of changes

## Checklist:


lib/plugins/inspec-parallel/lib/inspec-parallel/command.rb
```diff
@@ -59,6 +60,30 @@ def read_options_file
...
+        if File.exist?(option_file)
+          unless [".sh", ".csh", ".ps1"].include? File.extname(option_file)
+            File.readlines(option_file)
+          else
+            if Inspec.locally_windows? && (File.extname(option_file) == ".ps1")
+              output = `powershell -File "#{option_file}"`
+              output.split("\n")
+            elsif [".sh", ".csh"].include? File.extname(option_file)
+              if system("bash #{option_file}", out: File::NULL)
+                output = `bash "#{option_file}"`
```
clintoncwolfe: I'm concerned here that we're executing the file twice. Is there anyway we could keep the stdout from the first execution, and if it worked, use it?

Nik08: Let me check a workaround for this.

PR #6019, opened 2022-05-01
CFINSPEC-143 Parallel - Thor-based validation; inject child-status reporter

clintoncwolfe:
This PR reworks the invocation validation code to make it Thor-based for consistency, renames some things, and then leverages the Thor options to inspect the reporter option to see if the user provided file outputs for their reporters, and injects a child-status reporter for each invocation.

If acceptable this will be merged up into the existing nm/inspec-parallel-forking branch.

## Types of changes

## Checklist:


lib/plugins/inspec-parallel/lib/inspec-parallel/validator.rb
```diff
@@ -42,39 +43,78 @@ def create_aliases_mapping
...
+
+        # LIMITATION: this approach is limited to having exactly one profile in the invocation
+        args = [invocation_words.shift] # That is, the profile path
+
+        # Here we're piggybacking on on a hook used by the start() method, and provides the
+        # specifics for the subcommand
+        config = { command_options: thor_options_for_sub_cmd }
+
+        # This performs the parse
+        thor = Inspec::InspecCLI.new(args, invocation_words, config)
```
Nik08, line 110: Wow, so much cleaner!

lib/plugins/inspec-parallel/lib/inspec-parallel/command.rb
```diff
@@ -4,45 +4,66 @@
...
         @sub_cmd = sub_cmd
         @logger  = Inspec::Log
-        @validation_passed = true
-        @option_lines = read_options_file
+        @invocations = read_options_file
       end
 
       def run
-        Runner.new(option_lines, options, sub_cmd).run
+        validate_invocations!
```
Nik08, line 23: Just curious, why are we validating invocation before run also now?

clintoncwolfe: Because validation needs to check (and possibly modify) the --reporter option to ensure that a child-status reporter is present.

Nik08:
Looks good to me. Just the conflict which needs to be taken care.

PR #6028, opened 2022-05-04
CFINSPEC-171 Background run option for inspec parallel

Nik08:

## Description

Added functionality to run parallel operations in background, with `--bg` option. Only supported for unix based platforms.

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/cli.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/cli.rb
@@ -19,6 +19,8 @@ module InspecPlugins::Parallelism
       desc: "Number of jobs to run parallely"
     option :ui, type: :string, default: "status",
       desc: "Which UI to use: status, text"
+    option :daemon, aliases: :d, type: :boolean,
+      desc: "Runs parallel processes in background"
     exec_options
     def exec(default_profile = nil)
       parallel_cmd = InspecPlugins::Parallelism::Command.new(options, default_profile)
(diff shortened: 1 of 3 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/cli.rb
```diff
@@ -17,6 +17,8 @@ class CLI < Inspec.plugin(2, :cli_command)
       desc: "Prints all thor options on dry run"
     option :jobs, aliases: :j, type: :numeric,
       desc: "Number of jobs to run parallely"
+    option :daemon, aliases: :d, type: :boolean,
```
clintoncwolfe: I think `daemon`  is a technical term that reflect how it is internally implemented.  To me, an option named `daemon` would mean "run as a server" - like inspec was going to run as a server and keep running, handling more jobs as we keep handing it new work. That's not what we're doing. I think a `--bg` is a better option name which better reflects what we are doing.

lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -28,6 +29,22 @@ def run
           cleanup_child_processes
           sleep 0.1
         end
+        cleanup_daemon_process if run_in_background
+      end
+
+      def initiate_background_run
+        if Inspec.locally_windows?
+          #TBD
```
clintoncwolfe: For now exit with a warnign message and a Inspec::UI.new.exit(:usage_error)

clintoncwolfe (commented):
Just a couple of minor changes, a request to rename the option and what to do if we are on Windows.

lib/plugins/inspec-parallel/lib/inspec-parallel/cli.rb
Nik08: Yeah that's true. I had the same opinion about it, but I also felt that option name as "bg" should be not creating confusion too because we are using it as a shortcut for "background". Alternatively, having a 2 character alias using thor seemed not possible.
But since you feel going with "bg" is clear enough with the option name, I will go for this change. Thanks!

PR #6044, opened 2022-05-06
CFINSPEC-188 CTL-C exit handling

Nik08:

## Description

To handle **CTL-C** exit gracefully.

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/super_reporter/base.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/super_reporter/base.rb
@@ -2,8 +2,6 @@ module InspecPlugins::Parallelism
   class SuperReporter
 
     def self.make(type, job_count, invocations)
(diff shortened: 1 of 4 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/super_reporter/base.rb
```diff
@@ -2,6 +2,8 @@ module InspecPlugins::Parallelism
   class SuperReporter
 
     def self.make(type, job_count, invocations)
+      puts "Press CTL+C to stop\n"
```
clintoncwolfe: Ah, we can't output anything here. We may have other super-reporters that are intentionally silent, or output JSON to STDOUT (in which case this would be a mis-parse).

Nik08: Got it!

clintoncwolfe (approved):

PR #6057, opened 2022-05-12
CFINSPEC-252: Fix profile gem dependency installation is failing when  gem version is not specified

### LINKED ISSUE #6053: requried_gems seems to fail when a `version` is not provided
When using the new required gems capability in the `inspec.yml`, the `version` property seems to be expected.

I would have expected if a version was not given, that the lastest version would be installed.

If we are going to require both, then we should document clearly that both are _required_ for things to work and this should be added to the `inspec check` capability.

Vasu1105:

## Description

* Profile gem dependency installation works now if the gem version is not specified in the inspec.yml
* Updated the User prompt text for the profile gem dependency installation to be self-explanatory.
## Related Issue

Fixes #6053
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -410,7 +410,7 @@ module Inspec
           else
             ui = Inspec::UI.new
             gem_dependencies.each { |gem_dependency| ui.list_item("#{gem_dependency[:name]} #{gem_dependency[:version]}") }
-            choice = ui.prompt.select("Would you like to install profile gem dependencies listed above?", %w{Yes No})
+            choice = ui.prompt.select("The above listed gem dependencies are required to run the profile. Would you like to install them ?", %w{Yes No})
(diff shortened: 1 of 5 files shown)


lib/inspec/profile.rb
```diff
@@ -410,7 +410,7 @@ def load_gem_dependencies
           else
             ui = Inspec::UI.new
             gem_dependencies.each { |gem_dependency| ui.list_item("#{gem_dependency[:name]} #{gem_dependency[:version]}") }
-            choice = ui.prompt.select("Would you like to install profile gem dependencies listed above?", %w{Yes No})
+            choice = ui.prompt.select("The above listed gem dependencies are required to run profile. Would you like to install them ?", %w{Yes No})
```
clintoncwolfe: "to run the profile"

PR #6058, opened 2022-05-12
CFINSPEC-189 Added logging for inspec parallel

Nik08:
## Description

1. To create runner logs and errors log using child process id
2. Feature to use ERB escape `<%= pid %>` instead of reporter file name within options file, which will be then replaced by child process id.
3. By default logs are created in current directory. Option `--log-path` added to provide custom log path where logs should be created.
## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/command.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/command.rb
@@ -42,7 +42,7 @@ module InspecPlugins
           exit Inspec::UI::EXIT_TERMINATED_BY_CTL_C
         end
       end
-      
+
       def validate_thor_options
         # only log path validation needed for now
         validate_log_path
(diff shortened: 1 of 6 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/command.rb
```diff
@@ -77,6 +91,7 @@ def read_options_file
         end
         content.each.with_index(1) do |str, index|
           data_hash = { line_no: index }
+          str = "<% pid = 'CHILD_PID' %>" + str if str.include? "pid"
```
clintoncwolfe: You must actually evaluate the variable pid with the value CHILD_PID. You cannot predict what the string is going to contain.

lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -61,10 +62,17 @@ def spawn_another_process
...
         # Construct command-line invocation
-        cmd = "#{$0} #{sub_cmd} #{invocation}"
-        child_pid = Process.spawn(cmd, out: parent_writer)
-        @child_tracker[child_pid] = { io: child_reader }
-        @ui.child_spawned(child_pid, invocation)
+        child_pid = nil
+        begin
+          cmd = "#{$0} #{sub_cmd} #{invocation}"
+          log_msg = "#{Time.now.iso8601} Start Time: #{Time.now}\n#{Time.now.iso8601} Arguments: #{invocation}\n"
+          child_pid = Process.spawn(cmd, out: parent_writer)
```
clintoncwolfe: You need to redirect stderr here too for the log method

clintoncwolfe (changes requested):
Need to redirect stderr in spawn, and need to actually eval vars in the ERB eval.

PR #6074, opened 2022-05-20
CFINSPEC-249 Fix for dependent profiles with same name but different version

### LINKED ISSUE #5978: inspec cli output unclear when same resource pack is used in different versions by dependencies
# Describe the problem

I'm using inspec-aws as a dependency within the "base" control in e.g. version 1.80.
The "base" controls also depends on another (git-hosted) my2nd-profile which again depends on inspec-aws however in this case a different version, e.g. 1.30.
The output of inspec executes all controls (base and my2nd) however the output only shows inspec-aws profile in version 1.30.
Inspec.lock file includes both versions of inspec-aws.
So I'm assuming that inspec pulls in the respective version when needed however the output only lists one version ( the last in sequence).

## Possible Solution

Ooutput both profile versions even if the profile (resource pack) name is not unique but the same.

Nik08:

## Description

Fix for dependent profiles with same name but different version, while using aws profiles

## Related Issue

Fixes #5978
## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,7 @@ group :test do
   gem "json_schemer", ">= 0.2.1", "< 0.2.19"
   gem "m"
   gem "minitest-sprint", "~> 1.0"
-  gem "minitest", "~> 5.5"
+  gem "minitest", "5.15.0"
   gem "mocha", "~> 1.1"
   gem "nokogiri", "~> 1.9"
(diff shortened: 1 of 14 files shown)


clintoncwolfe (approved):
This was team-reviewed and approved a couple of days ago.

PR #6107, opened 2022-06-02
CFINSPEC-269 Adds resource_id group 8

Vasu1105:

## Description

This adds resource_id for the following resources.
* npm
* ntp_conf
* oneget
* opa_api
* opa_cli

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/opa_api.rb
+++ b/lib/inspec/resources/opa_api.rb
@@ -6,6 +6,12 @@ module Inspec::Resources
     supports platform: "unix"
     supports platform: "windows"
 
+    example <<~EXAMPLE
+      describe opa_api(url: "localhost:8181/v1/data/example/violation", data: "input.json") do
+        its(["result"]) { should eq 'value' }
+      end
+    EXAMPLE
+
     def initialize(opts = {})
       @url = opts[:url] || nil
       @data = opts[:data] || nil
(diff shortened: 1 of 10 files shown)


lib/inspec/resources/opa_api.rb
```diff
@@ -18,6 +24,10 @@ def allow
       @content["result"]
     end
 
+    def resource_id
+      @url
```
ahasunos: This might lead to nil error since in line 16 @url is initialized as `@url = opts[:url] || nil`.
Suggested fix: `@nil || "OPA api"

lib/inspec/resources/opa_cli.rb
```diff
@@ -22,6 +28,10 @@ def allow
       @content["result"][0]["expressions"][0]["value"] if @content["result"][0]["expressions"][0]["text"].include?("allow")
     end
 
+    def resource_id
+      "#{@policy}:#{@query}"
```
ahasunos: This might lead to nil error too, since in line 17 `@policy = opts[:policy] || nil`

lib/inspec/resources/opa_api.rb
Vasu1105: Yes but there is validation also below that which does not allow URL as nil.

lib/inspec/resources/opa_cli.rb
Vasu1105: Same here there is validation.

clintoncwolfe (approved):

PR #6110, opened 2022-06-02
CFINSPEC-272: resource_ids group 11

ahasunos:

## Description
**CFINSPEC-272: resource_ids group 11**
This PR adds resource_id methods to the following resources:
- ssh_config
- sshd_config
- sybase_conf
- sybase_session
- sys_info
- timezone

When a user runs inspec exec [profile] --reporter json
then when they look at the resource_id field for the core controls
then they would be able to see an identifier for each control

## Related Issue
**CFINSPEC-272: resource_ids group 11**

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/sybase_session.rb
+++ b/lib/inspec/resources/sybase_session.rb
@@ -64,6 +64,10 @@ module Inspec::Resources
       DatabaseHelper::SQLQueryResult.new(isql_cmd, parse_csv_result(isql_cmd.stdout))
     end
 
+    def resource_id
+      "Sybase Session"
+    end
+
     def to_s
       "Sybase Session"
     end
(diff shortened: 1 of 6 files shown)


lib/inspec/resources/sybase_session.rb
```diff
@@ -64,6 +64,10 @@ def query(sql)
       DatabaseHelper::SQLQueryResult.new(isql_cmd, parse_csv_result(isql_cmd.stdout))
     end
 
+    def resource_id
+      "Sybase Session"
```
Nik08: Maybe we can use username, db or some other combination here right? These are defined in initialiser.

ahasunos: Those values can be sensitive information

Nik08: Except password, user name or db name can be used I feel. Those might not be that sensitive for creating resource id.

ahasunos: Changed accordingly to consider the database.

clintoncwolfe (approved):

PR #6153, opened 2022-06-20
CFINSPEC-353 Suggest- Basic tool for packaging

clintoncwolfe:
Part of InSpec Suggest.

Basic tool to do packaging of suggestion criteria. Simplest thing that could work. Assumes source profiles are already cloned into tmp/ under inspec source directory, and take a list of globs from the config file. Looks for all files in glob/suggest/*.rb and concatenates them into etc/suggest/SETNAME/controls/criteria.rb .

Should probably do more, but that will be on another ticket, this is just to unblock further progress.

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,4 +37,5 @@ results/
 terraform.tfstate*
 terraform.tfstate.backup
 terraform.tfvars
-test/**/*.lock
+test/**/*.locktmp
+tmp/*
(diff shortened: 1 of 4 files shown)


clintoncwolfe:
This is now blocking work on the command branch - I'm going to go ahead and merge it as it is just a POC and it works well enough as a POC.

PR #6156, opened 2022-06-21
CFINSPEC-231 Suggest - Basic Reporters

clintoncwolfe:
A set of reporters for InSpec suggest.

First, a diagnostic reporter, suggest-debug.
Next, a streaming reporter, suggest-progress. Simple progress bar, no text output.
Modifies the core reporter validation to allow both streaming reporter and one-shot reporter output to STDOUT.
Finally, a simple reporter that outputs the results of the scan contest, suggest-text.

Intended to be run in normal use with both suggest-progress and suggest-text enabled.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/config.rb
+++ b/lib/inspec/config.rb
@@ -384,12 +384,21 @@ module Inspec
       end
 
       # check to make sure we are only reporting one type to stdout
-      stdout_reporters = 0
-      reporters.each_value do |reporter_config|
-        stdout_reporters += 1 if reporter_config["stdout"] == true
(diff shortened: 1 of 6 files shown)


aaronlippold:
The text output should be `grep` and `cut` strcutured output in my openion

clintoncwolfe:
I agree that the output should be machine-parseable - tabular at least. We'll take that up on a later ticket, I think we'll be making some other improvements too regarding handling the versions.

ahasunos (approved):

PR #6175, opened 2022-06-30
CFINSPEC-229 Suggest: suggtool fetch

clintoncwolfe:
Git fetch support for the suggtool inspec support tool. Fetches profiles from git repos into the working directory specified on the CLI or in the config file.

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/fetch.rb
+++ b/lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/fetch.rb
@@ -1,0 +1,23 @@
+require_relative "base"
+module InspecPlugins::Suggest
+  class SuggToolCommand < Inspec.plugin(2, :cli_command)
+
+    desc "fetch SETNAME", "Fetches sets of source profiles for later processing."
+    def fetch(*requested_sets)
+      setup(requested_sets)
+
+      with_requested_sets do |set_name, set_cfg|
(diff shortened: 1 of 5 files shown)


Vasu1105:
Docs for overall suggest tool is good to have or the link if it's already there.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/fetch.rb
```diff
@@ -7,16 +10,26 @@ def fetch(*requested_sets)
...
+            ui.info "fetch: updated"
+          else
+            git = Git.clone(fetch_cfg["repo"], checkout_dir)
+            ui.info "fetch: cloned from #{fetch_cfg["repo"]}"
+          end
+
+          ref = fetch_cfg["ref"] || "main"
+          git.checkout(ref)
+          ui.info "fetch: checked out #{ref}"
+        end
```
Vasu1105, line 32: Do we require any error handling for git connection or error while fetching the repo from git? Or we are ignoring that since this tool will be used internally?

clintoncwolfe: I agree, there is no error checking. BUt since it will be used internally and will spew errors as backtraces to STDERR I think we can tolerate it.

Vasu1105 (approved):

PR #6192, opened 2022-07-07
CFINSPEC-361: Add `podman_image` resource

ahasunos:

## Related Issue
**CFINSPEC-361: Add `podman_image` resource**

## Description
- This PR adds a new resource: `podman_image`
- Using the`podman_image` resource
Given that the user has called the `podman_image` resource
then the resource allows to test with the container image properties.
- Syntax:
```ruby
  describe podman_image("docker.io/library/busybox") do
      it { should exist }
      its("resource_id") { should eq "sha256:3c19bafed22355e11a608c4b613d87d06b9cdd37d378e6e0176cbc8e7144d5c6" }
  end
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/podman_image.rb
+++ b/lib/inspec/resources/podman_image.rb
@@ -40,49 +40,49 @@ module Inspec::Resources
     end
 
     def exist?
-      inspect_info.key?("id")
+      get_value("id") != nil
     end
 
     def repo_tags
-      inspect_info["repo_tags"]
+      get_value("repo_tags")
     end
 
     def size
       # TODO: Convert bytes to KB or MB; if required
-      inspect_info["size"]
+      get_value("size")
     end
 
     def digest
-      inspect_info["digest"]
+      get_value("digest")
     end
 
     def created_at
-      inspect_info["created_at"]
+      get_value("created_at")
     end
 
     def version
-      inspect_info["version"]
+      get_value("version")
     end
 
     def names_history
-      inspect_info["names_history"]
+      get_value("names_history")
     end
 
     def repo_digests
-      inspect_info["repo_digests"]
+      get_value("repo_digests")
     end
 
     def architecture
-      inspect_info["architecture"]
+      get_value("architecture")
     end
 
     def os
-      inspect_info["os"]
+      get_value("os")
     end
 
     def virtual_size
       # TODO: Convert bytes to KB or MB
-      inspect_info["virtual_size"]
+      get_value("virtual_size")
     end
 
     def resource_id
@@ -129,6 +129,12 @@ module Inspec::Resources
 
       require "json" unless defined?(JSON)
       JSON.parse(podman_inspect_cmd.stdout)
+    end
+
+    def get_value(key)
+      return nil if inspect_info.nil?
+
+      inspect_info[key]
     end
   end
 end
--- a/lib/inspec/resources/podman_image.rb
+++ b/lib/inspec/resources/podman_image.rb
@@ -43,6 +43,10 @@ module Inspec::Resources
       get_value("id") != nil
     end
 
+    def id
+      get_value("id")
+    end
+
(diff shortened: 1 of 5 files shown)


lib/inspec/resources/podman_image.rb
```diff
@@ -0,0 +1,116 @@
...
+      o = { image: opts } if opts.is_a?(String)
+      @opts = sanitize_options(o)
+    end
+
+    def image
+      "#{repo}:#{tag}" if object_info.entries.size == 1
+    end
+
+    def repo
+      object_info.repositories[0] if object_info.entries.size == 1
```
clintoncwolfe: What happens if object_info.entries.size != 1 ?

lib/inspec/resources/podman_image.rb
```diff
@@ -0,0 +1,116 @@
...
+      opts
+    end
+
+    def object_info
+      return @info if defined?(@info)
+
+      opts = @opts
+      @info = inspec.podman.images.where do
+        (repository == opts[:repo] && tag == opts[:tag]) || (!id.nil? && !opts[:id].nil? && (id == opts[:id] || id.start_with?(opts[:id])))
+      end
```
clintoncwolfe: Check for multiple entries here

lib/inspec/resources/podman_image.rb
```diff
@@ -0,0 +1,144 @@
...
+    EXAMPLE
+
+    attr_reader :opts, :inspect_info
+
+    def initialize(opts)
+      skip_resource "The `podman_image` resource is not yet available on your OS." unless inspec.os.unix?
+      opts = { image: opts } if opts.is_a?(String)
+      @opts = sanitize_options(opts)
+      @inspect_info = get_inspect_info
```
Vasu1105: In `get_inspec_info` method, we actually getting the image data using the `podman image inspect` command, so naming that method to something specific might be helpful. and maybe also the `@inspect_info` to `@image_info`

ahasunos: Any suggestion for the method name?

ahasunos: get_image_inspect_info?

Vasu1105 (approved):
Small suggestion else all looks good. Good work!!

lib/inspec/resources/podman_image.rb
ahasunos: Renamed the attribute and method name to `image_info` and `get_image_info` as discussed.

Nik08 (approved):
Looks good to me too!

Vasu1105:
@ahasunos Can we have a test that will check if the given image name does not exist?

Vasu1105:
> @ahasunos Can we have a test that will check if the given image name does not exist?

Ignore it seems you already have it.

dkumaras (commented):
@ahasunos - Pls chk n cnfrm.

Vasu1105:
@ahasunos I get below result for the non existing image

```
describe podman_image("test") do
  it { should exist }
end

```
```
Profile:   InSpec Profile (podman_test)
Version:   0.1.0
Target:    local://
Target ID: 0744915e-9ac4-5993-a769-a7cf9db10a72

  podman_image test:latest
     ×  Unable to retrieve podman image info for test:latest.
     Error message: Error: inspecting object: unable to inspect "test:latest": failed to find image test:latest: test:latest: image not known
```

I should get false if the image does not exist right? But it's failing with an error.

ahasunos:
@Vasu1105 

Just made the changes. It was failing with an error because I was raising the exception right at the initialisation while finding inspect information for the image.

Vasu1105:
@ahasunos we merged the PR but that fix for the non-existing image will not raise an error in case of Podman connection failures. Please update that in when you are working on the other Podman resources.

PR #6194, opened 2022-07-08
CFINSPEC-370: Enhance `suggtool` with `bump` capability

ahasunos:

## Related Issue
**CFINSPEC-370: Enhance `suggtool` with `bump` capability**

## Description
- Extends `suggtool` utility with `bump` capability.
- Command: `inspec suggtool bump SET_NAME [OPTIONS]`
- Given a `tmp` directory with a `profiles/SET_NAME` profile
and given an `inspec.yml` file present in `profiles/SET_NAME` with version 0.1.0
then `inspec suggtool bump SET_NAME` increases the version number to 0.1.1
- `suggtool bump` supports 3 bumps (major, minor and patch) that can be passed as options (`--level minor`, `--level major` or `--level patch`). Default bump is `patch`.

## Types of changes

## Checklist:


Nik08 (approved):
Looks good!

etc/suggest.yaml
```diff
@@ -14,3 +14,6 @@ sets:
     # check:
     #   disable-checks:
     #     - trailing-newline
+  chef-compliance-2:
+    paths:
+      - profiles/chef-compliance
```
clintoncwolfe: This doesn't appear to be used - is it needed?

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/base.rb
```diff
@@ -55,6 +55,14 @@ def with_matching_files(set_cfg, &block)
         end
       end
 
+      def with_matching_yaml_files(set_cfg, &block)
```
clintoncwolfe: I would suggest renaming this to `with_set_metadata_files`

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/base.rb
```diff
@@ -55,6 +55,14 @@ def with_matching_files(set_cfg, &block)
...
+        set_cfg["paths"].each do |path|
+          Dir.glob(File.join(@working_dir, path, "inspec.yml")).each do |control_file_path|
+            yield(control_file_path)
```
clintoncwolfe: This is not iterating on control_files. Please rename the iterator variable to be something like "metadata_file"

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,66 @@
...
+
+module InspecPlugins::Suggest
+  class SuggToolCommand < Inspec.plugin(2, :cli_command)
+
+    option :patch, type: :boolean, default: true,
+      desc: "Increase third digit of the version number"
+    option :minor, type: :boolean, default: false,
+      desc: "Increase second digit of the version number"
+    option :major, type: :boolean, default: false,
+      desc: "Increase first digit of the version number"
```
clintoncwolfe: Having three different options creates the possibility of running with incoherent options - like `inspec suggtool bump --major --minor`.  I would suggest having one option that takes a parameter, like `--level [patch|minor|major]` default patch.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    desc "bump SETNAME", "Increase version number of SETNAME in inspec.yml."
+    def bump(*requested_sets)
+      setup(requested_sets)
+
+      with_requested_sets do |set_name, set_cfg|
+        with_matching_yaml_files(set_cfg) do |control_file_path|
```
clintoncwolfe: again, it's not a control_file_path, it's a metadata file.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,66 @@
...
+          # TODO: We may not need the full path here.
+          file_info = {
+            full_path: control_file_path,
+            rel_path: control_file_path.sub(Inspec.src_root + "/", ""),
+          }
+          bump_version(file_info[:rel_path])
```
clintoncwolfe: This calling pattern is clearly copied from `inspec suggtool check` - it doesn't make sense here. The only thing you need is the path.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,66 @@
...
+    option :major, type: :boolean, default: false,
+      desc: "Increase first digit of the version number"
+
+    desc "bump SETNAME", "Increase version number of SETNAME in inspec.yml."
+    def bump(*requested_sets)
+      setup(requested_sets)
+
+      with_requested_sets do |set_name, set_cfg|
+        with_matching_yaml_files(set_cfg) do |control_file_path|
+          # TODO: We may not need the full path here.
```
clintoncwolfe: You do need the full path. Under testing, you may be in a temporary directory.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,66 @@
...
+        yaml_data["version"] = new_version
+
+        File.open(file_path, "w") do |file|
+          file.write(yaml_data.to_yaml)
+        end
+        ui.info "Bumped:#{bump_label} version of #{file_path} to #{new_version}"
+      end
+
+      def increase_version(version)
+        Gem::Version.new(version).bump.to_s
```
clintoncwolfe: I don't think you need to call Gem::Version.bump here - that operates on dotted triples, whereas you're passing a single digit. You can just use `version.succ`.

lib/plugins/inspec-suggest/test/functional/suggtool_bump_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:fixture_path) { File.join(repo_path, "lib", "plugins", "inspec-suggest", "test", "fixtures") }
+  let(:config) { File.join(fixture_path, "sugg-conf", "suggest-bump.yaml") }
+
+  describe "when packaging the basic test fixture set" do
+    it "bump:patch version default" do
+      result = run_inspec_process("suggtool bump --suggest-config #{config}")
+      _(result.stderr).must_equal ""
+      _(result.exit_status).must_equal 0
+      _(result.stdout).must_include "Bumped:Patch version of"
+    end
```
clintoncwolfe: Ideally, we need to figure out a way to do this in a temporary directory, having called `package` on the set, which will create the metadata yml file. We don't want to modify the metadata yaml file in the fixture directory, since multiple tests will be accessing it in parallel.

lib/plugins/inspec-suggest/test/functional/suggtool_bump_test.rb
```diff
@@ -0,0 +1,32 @@
...
+require "tmpdir" unless defined?(Dir.mktmpdir)
+
+describe "inspec suggtool bump" do
+  include CorePluginFunctionalHelper
+
+  let(:fixture_path) { File.join(repo_path, "lib", "plugins", "inspec-suggest", "test", "fixtures") }
+  let(:config) { File.join(fixture_path, "sugg-conf", "suggest-bump.yaml") }
+
+  describe "when packaging the basic test fixture set" do
+    it "bump:patch version default" do
```
clintoncwolfe: Need to have a separate test for explicitly setting patch.

clintoncwolfe (changes requested):
Looks like a good start but several small changes and a problem with testing indicate to me that this needs some further work.

etc/suggest.yaml
ahasunos: I was testing on my local. This needs to be removed.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/base.rb
ahasunos: Renamed!

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
ahasunos: Right, changed accordingly.

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
ahasunos: Used `.succ` to eliminate misuse of `bump`

lib/plugins/inspec-suggest/test/functional/suggtool_bump_test.rb
ahasunos: Done!

lib/plugins/inspec-suggest/test/functional/suggtool_bump_test.rb
ahasunos: Added separated test explicitly for patch

lib/plugins/inspec-suggest/lib/inspec-suggest/suggtool/bump.rb
```diff
@@ -0,0 +1,53 @@
...
+    option :level, type: :string, default: "patch",
+      desc: "Level of the version number to bump. One of: patch, minor, major"
+
+    desc "bump SETNAME", "Increase version number of SETNAME in inspec.yml."
+    def bump(*requested_sets)
+      setup(requested_sets)
+
+      with_requested_sets do |set_name, set_cfg|
+        with_set_metadata_files(set_cfg) do |metadata_file_path|
+          rel_path = metadata_file_path.sub(Inspec.src_root + "/", "")
```
clintoncwolfe: I wouldn't calculate the relative path at all - just use the path as given. Especially under testing, you'll want to use the full path.

lib/plugins/inspec-suggest/test/fixtures/profiles/chef-compliance/inspec.yml
```diff
@@ -0,0 +1,12 @@
+---
```
clintoncwolfe: You should not need this file anymore - if you're using suggtool package in the tests, it will automatically expand the metadata from the test-fixture.iaf file.

lib/plugins/inspec-suggest/test/fixtures/sugg-conf/suggest-bump.yaml
```diff
@@ -0,0 +1,6 @@
+working_directory: lib/plugins/inspec-suggest/test/fixtures
+
+sets:
+  chef-compliance:
```
clintoncwolfe: Better to use test-fixture here. We might not use chef-compliance as the permanent name.

clintoncwolfe (commented):
Getting there! just a few minor adjustments!

clintoncwolfe (approved):
Looks great - good work!

PR #6196, opened 2022-07-11
CFINSPEC-371 Suggest: suggtool sign

clintoncwolfe:
The last of the 5 low-level tools for supporting InSpec Suggest.

Takes a suggestion set profile directory (as created by suggtool package) which may have been version bumped (by suggtool bump) and signs it with a key (provided on the CLI) by internally calling `inspec sign`. Places the IAF file in etc/suggest (may be overridden by the --out-dir option for testing). Names the IAF file without a version so that `inspec suggest` can find it easily, and we don't have to have a bunch of filenames in our version control.

## Types of changes

## Checklist:


ahasunos (approved):
Functional test and manual test is working as expected 💯

PR #6204, opened 2022-07-15
CFINSPEC-351: Add `podman_volume` and `podman_pod` resources

ahasunos:
Signed-off-by: Sonu Saha [sonu.saha@progress.com](mailto:sonu.saha@progress.com)

## Related Issue
**CFINSPEC-351: Add `podman_volume` and `podman_pod` resources**

## Description
- Extend support for podman in InSpec core with `podman_volume` and `podman_pod` resources.
- Supported matcher (for both the resource): `exist`
- Supported properties (for both the resource): all keys obtained in `podman [volume|pod] inspect` standard output
- The preferred way to review this PR is to walk through the commits

## Types of changes

## Checklist:


dkumaras (commented):
Pls check and confirm.

ahasunos:
> Pls check and confirm.

Looks good to me @dkumaras 
Thanks!

dkumaras (approved):

Vasu1105 (approved):
Looks good to me.

test/unit/resources/podman_volume_test.rb
```diff
@@ -0,0 +1,40 @@
+require "inspec/globals"
+require "#{Inspec.src_root}/test/helper"
+require_relative "../../../lib/inspec/resources/podman_volume"
+
+describe Inspec::Resources::PodmanVolume do
+  it "checks podman volume parameter and works correctly" do
+    resource = MockLoader.new("unix".to_sym).load_resource("podman_volume", "my_volume")
```
ahasunos, line 7: Change directly to :unix instead of "unix".to_sym

test/unit/resources/podman_pod_test.rb
```diff
@@ -0,0 +1,50 @@
+require "inspec/globals"
+require "#{Inspec.src_root}/test/helper"
+require_relative "../../../lib/inspec/resources/podman_pod"
+
+describe Inspec::Resources::PodmanPod do
+  it "checks podman pod parameter and works correctly" do
+    resource = MockLoader.new("unix".to_sym).load_resource("podman_pod", "nginx-frontend")
```
Nik08, line 7: Change directly to :unix instead of "unix".to_sym

test/fixtures/cmd/podman-errors
```diff
@@ -1,2 +1,6 @@
 Error: inspecting object: network min: network not found
-Error: inspecting object: unable to inspect \"abc\": failed to find image abc: abc: image not known
\ No newline at end of file
+Error: inspecting object: unable to inspect \"abc\": failed to find image abc: abc: image not known
```
Nik08, line 2: We should maybe keep these error files differently since it will create more clarity over what each resources are expecting in terms of errors.

test/fixtures/cmd/podman-errors
```diff
@@ -1,2 +1,6 @@
...
+Error: no pod with name or ID non_existing_pod found: no such pod
+[]
+error inspecting object: no such object: "non_existing_volume"
+Error: inspecting object: no such object: "volume"
```
sathish-progress, line 6: better to split this into different fixtures

Vasu1105: This is a fixture and can be treated as a list of errors. As this is a unit test it's not going to make difference as we are not doing any functional testing. This is just to verify whether the string match is working or not. If the change in the error message happens on the API this unit test as it's not going to help anyway so reducing the number of fixture file make sense for me here. We can add comments in the fixture file itself about what that file is all about.

PR #6207, opened 2022-07-19
CFINSPEC-382: Create `suggtool build` high-level command

ahasunos:
Signed-off-by: Sonu Saha [sonu.saha@progress.com](mailto:sonu.saha@progress.com)

## Related Issue
**CFINSPEC-382: Create `suggtool build` high-level command**

## Description
This PR introduces the `suggtool build` feature which performs all the suggtool activity.
Suggested way to review this PR is to walk through the commits.

## Types of changes

## Checklist:


lib/plugins/inspec-suggest/test/functional/suggtool_build_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    Dir.mktmpdir do |workdir|
+      cmd = "suggtool build bad-fixture-branch"
+      cmd += " --suggest-config #{config}"
+      cmd += " --work-dir #{workdir}"
+      run_suggtool_build = run_inspec_process(cmd)
+
+      _(run_suggtool_build.exit_status).must_equal 1
+      _(run_suggtool_build.stdout).must_include "fetch: working on set bad-fixture-branch"
+      _(run_suggtool_build.stdout).must_include "fetch: checked out ss/profiles-for-build"
+      _(run_suggtool_build.stdout).must_include "check: Saw 2 error(s), 2 warning(s)"
```
Nik08, line 96: Is it a good idea to just check for `check:` instead of the whole string after? Because what if it does not match the exact error message and it throws some other kind of message in return. 
This change holds true for all the other test cases as well.

Nik08:
Looks good, except for one minor thing in test cases.

lib/plugins/inspec-suggest/test/functional/suggtool_build_test.rb
ahasunos: > It holds true for all the other test cases as well

Do you mean the strings
`"fetch: working on set bad-fixture-branch"` should be changed to `"fetch:"`
`"bump: Bumped patch version of"` should be changed to `"bump:"`
`"package: Extracted inspec.yml from"` should be changed to `"package:"`
and so on?

ahasunos: Done.

clintoncwolfe (approved):

PR #6222, opened 2022-08-02
CFINSPEC-246/CFINSPEC-247 Attestation changes for N/R outcomes

Nik08:
## Description

Attestation feature which enables attesting  N/R controls and marking them `pass` or `fail` manually using a file option `--attestation-file`

## Types of changes

## Checklist:


lib/inspec/attestations.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    def self.determine_expiry(attestation_data, control_id)
+      if attestation_data["expiration_date"]
+        attestation_data["expiration_date"]
+      elsif !attestation_data["updated"].blank? && !attestation_data["frequency"].blank?
+        # logic to find expiration date using frequency and updated date.
+        updated_last = attestation_data["updated"]
+        begin
+          if updated_last.is_a?(Date) || (updated_last.is_a?(String) && Date.parse(updated_last).year != 0)
```
Vasu1105: Maybe you can write this whole bunch of if and else in another method to make it cleaner.

dkumaras (approved):

lib/inspec/attestations.rb
Nik08: Sure, let me think about how I can logically separate those lines since I think they are part of the logic used for fetching the expiration date.

lib/inspec/attestations.rb
```diff
@@ -106,47 +106,50 @@ def self.determine_expiry(attestation_data, control_id)
...
         rescue => e
           ui = Inspec::UI.new
-          ui.error("Unable to parse attestation updated date '#{updated_last}' for control #{control_id}. Error: #{e.message}")
+          ui.error("Unable to parse attestation updated date '#{attestation_data["updated"]}' for control #{control_id}. Error: #{e.message}")
           ui.exit(:usage_error)
         end
       end
     end
+
+    def self.fetch_expiry_using_frequency_and_updated_date(updated_date, frequency, control_id)
```
Vasu1105: Maybe this method name can be reduced to fetch_expiry_date as its arguments are self-explanatory in that they need, `updated_date` and `frequency`. You can add code comments.

Nik08: I think the method name should also explain its usage. So it seems fine to me.

Vasu1105: No problem I felt it's too long :)

clintoncwolfe (approved):
Reviewed and discussed in meetings - thanks for the good work!

Vasu1105 (approved):
Other than the long method name all looks good. Thanks!!

PR #6229, opened 2022-08-05
CFINSPEC-374 DSL keyword `only_applicable_if` added

Nik08:
## Description

DSL keyword `only_applicable_if` will help set the impact value to zero conditionally when the condition inside the block is not true. It also invokes a failure test in the result containing the reason for why it is not applicable.

**Usage:**

```
control "test-1" do
  only_applicable_if ("Some reason for N/A") { false }
  impact 1.0
  describe ...
end
```
**Output:**

```
×  test-1: No-op
     ×  No-op 
     N/A control due to only_applicable_if condition: Some reason for N/A
```

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/inspec_exec_test.rb
+++ b/test/functional/inspec_exec_test.rb
@@ -1347,12 +1347,12 @@ EOT
 
     it "should show enhanced_outcomes for controls with impact 0" do
       _(run_result.stdout).must_include "5 skipped"
-      _(run_result.stdout).must_include "3 controls not applicable"
+      _(run_result.stdout).must_include "4 controls not applicable"
       _(run_result.stdout).must_include "N/A"
     end
 
     it "should show enhanced_outcomes for controls with errors" do
-      _(run_result.stdout).must_include "3 failures"
+      _(run_result.stdout).must_include "4 failures"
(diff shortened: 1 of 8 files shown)


dkumaras (approved):

clintoncwolfe (approved):
Thanks!

test/functional/inspec_exec_test.rb
```diff
@@ -1347,12 +1347,12 @@ def stderr
 
     it "should show enhanced_outcomes for controls with impact 0" do
       _(run_result.stdout).must_include "5 skipped"
-      _(run_result.stdout).must_include "3 controls not applicable"
+      _(run_result.stdout).must_include "4 controls not applicable"
```
Vasu1105, line 1350: While going through the test I felt that we could have added enhanced outcomes and attestation tests in different files as we did for the streaming progress bar test, as this inspec exec test file is getting bigger and bigger. This is not something as a blocker for this PR just a suggestion.

Vasu1105 (approved):

Vasu1105:
@Nik08 Verify pipeline is failing.

PR #6237, opened 2022-08-12
CFINSPEC-451 Inspec Parallel options reading using stdin

Nik08:
## Description

Option to read Inspec Parallel options reading using stdin, to use inspec parallel list.

Command to run:

`inspec parallel list -r aws_ecr_images -p "repository_name: 'test-inspec'" -t aws:// |  inspec parallel exec profiles/test-aws -o -
`

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
@@ -86,7 +86,8 @@ module InspecPlugins
         content = []
         begin
           if cli_options_to_parallel_cmd[:option_file] == "-"
-            content = $stdin.read&.split("\n") || []
+            content = $stdin.readlines if !$stdin.tty? && $stdin.stat.pipe?
+            @logger.error("Standard input options are empty") if content.empty?
           else
             content = content_from_file(cli_options_to_parallel_cmd[:option_file])
           end
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
@@ -83,8 +83,13 @@ module InspecPlugins
 
       def read_options_file
         opts = []
+        content = []
         begin
-          content = content_from_file(cli_options_to_parallel_cmd[:option_file])
+          if cli_options_to_parallel_cmd[:option_file] == "-"
+            content = $stdin.read&.split("\n") || []
+          else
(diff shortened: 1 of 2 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/list_reporter.rb
```diff
@@ -20,7 +20,7 @@ def example_passed(notification)
         @resource_formatter = InspecPlugins::Parallelism::ListReporter::ResourceFormatter::Base.formatter_for(resource_name, parameters: parameters, query: query)
       end
 
-      puts "-t " + resource_formatter.format_id_for_list(raw_id)
+      puts "-t " + resource_formatter.format_id_for_list(raw_id) + " --reporter json:<%= pid%>.json"
```
clintoncwolfe: Good catch. I'm not sure PID is the right thing - maybe something based on raw_id - but something is needed for reporter.

lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
```diff
@@ -83,8 +83,13 @@ def dry_run_commands
 
       def read_options_file
         opts = []
+        content = []
         begin
-          content = content_from_file(cli_options_to_parallel_cmd[:option_file])
+          if cli_options_to_parallel_cmd[:option_file] == "-"
+            content = $stdin.read&.split("\n") || []
```
clintoncwolfe: Are we obliged to slurp the entire STDIN at once? That will limit things that want to pipe chains of things. Is there any way we can read one line at a time?

clintoncwolfe (commented):
Concerned about how we are reading the input all at once.

lib/plugins/inspec-parallel/lib/inspec-parallel/list_reporter.rb
Nik08: You mean to say to use raw_id for the output file?

lib/plugins/inspec-parallel/lib/inspec-parallel/exec_command.rb
Nik08: Sure let me look into that.

Nik08: So you are suggesting a scenario where inspec parallel list is called multiple times and then the output for each pipe should get read properly. Correct me if I am wrong.

clintoncwolfe (approved):

PR #6337, opened 2022-12-05
CFINSPEC-533: Fix undefined method 'summary' for Gem::SourceFetchProblem (NoMethodError) when airgapped 

Vasu1105:

## Description
This fixes the undefined method 'summary' for #<Gem::SourceFetchProblem:0x000000000139dc70> (NoMethodError) while invoking any inspec command if the custom plugin is installed and if inspec commands are run in an air gap environment.
This fix removes an earlier call for getting gem specification from rubygems.org which is the cause for the above error and replaces it with fetching the summary from a locally saved gemspec file. This will work in both airgap and non-airgap environments. 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/loader.rb
+++ b/lib/inspec/plugin/v2/loader.rb
@@ -290,19 +290,18 @@ module Inspec::Plugin::V2
         when :user_gem
           status.entry_point = status.name.to_s
           status.version = plugin_entry[:version]
-          status.description = fetch_plugin_specs(status.name.to_s)&.summary
+          # Fetch the summary of the gem from local gemspec file instead of remote call using Gem::SpecFetcher.fetcher.
+          unless plugin_entry[:version].nil? # safe check very rare case.
+            version_string = plugin_entry[:version].gsub(/[=,~,>,<]/, "").strip
+            plugin_name_with_version = "#{status.name}-#{version_string}"
+            status.description = fetch_gemspec(File.join(plugin_gem_path, "gems", plugin_name_with_version, "/", status.name.to_s + ".gemspec"))&.summary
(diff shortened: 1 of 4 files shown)


lib/inspec/plugin/v2/loader.rb
```diff
@@ -290,7 +290,12 @@ def read_conf_file_into_registry
         when :user_gem
           status.entry_point = status.name.to_s
           status.version = plugin_entry[:version]
-          status.description = fetch_plugin_specs(status.name.to_s)&.summary
+          # Fetch the summary of the gem from local gemspec file instead of remote call using Gem::SpecFetcher.fetcher.
+          unless plugin_entry[:version].nil? #safe check very rare case.
```
Nik08: Hey can you also remove `fetch_plugin_specs` method then? And I think this method is used within the installer too. Can you verify if the use case is similar in that too? If yes, then maybe that could be fixed too. Because yes this would be failing in air-gapped env.

Vasu1105: Removed it from the `loader.rb`. Inside `installer.rb` it is getting used for `inspec plugin search ` functionality and it's not at least breaking the code right now in the air gap environment, it just returns an empty search results. Maybe we can discuss and create different PR for the search if that needs to be worked in an air gap environment. Currently, it seems the search needs to look out in rubygems.org to find the given plugin information.

Nik08: Yeah true. Cool makes sense!

Nik08 (approved):
LGTM!

clintoncwolfe (approved):

PR #6342, opened 2022-12-07
Fix env smoke test by updating ERB.new in `inspec env`; add additional test

clintoncwolfe:
## Description

Under Ruby 3.x, the calling pattern for ERB.new now insists you use named params, and issues warnings to stderr if you don't. The `env` command uses ERB to template the shell completion functions based on the current list of inspec commands. We test for warnings to stderr in our post-install smoke tests for env, but strangely had no pre-install tests for env.

So, shift that test left, and check for env to work correctly at CI verify time, like we do with all the other commands.

Also, make the simple change needed to fix the call to ERB new, fixing the smoke test.
Verified the fix works on 3.1.2 and 2.7.7 .

This PR blocks #6341 .

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/env_printer.rb
+++ b/lib/inspec/env_printer.rb
@@ -35,7 +35,7 @@ module Inspec
     private
 
     def print_completion_for_shell
-      erb = ERB.new(File.read(completion_template_path), nil, "-")
+      erb = ERB.new(File.read(completion_template_path), trim_mode: "-")
       puts erb.result(TemplateContext.new(@command_class).get_bindings)
     end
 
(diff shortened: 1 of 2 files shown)


lib/inspec/env_printer.rb
```diff
@@ -35,7 +35,7 @@ def print_and_exit!
     private
 
     def print_completion_for_shell
-      erb = ERB.new(File.read(completion_template_path), nil, "-")
+      erb = ERB.new(File.read(completion_template_path), trim_mode: "-")
```
clintoncwolfe, line 38: Some explanation about the change. The old way of calling used positional arguments to indicate using a default safe level (nil), then went on to indicate the trim mode "-" (which is one of the modes you can select).

The new way of saying the same thing using named arguments would be to omit the nil and to say `trim_mode: "-"`.

Vasu1105 (approved):

PR #6351, opened 2022-12-09
CFINSPEC-528 Upgraded mock Ubuntu fixture to latest LTS version.

Nik08:

## Description

Upgraded mock Ubuntu fixture to the latest LTS version to keep the unit tests updated. 
## Types of changes

## Checklist:

### DIFF
--- a/test/helpers/mock_loader.rb
+++ b/test/helpers/mock_loader.rb
@@ -20,7 +20,7 @@ class MockLoader
     macos10_10: { name: "mac_os_x", family: "darwin", release: "10.10.4", arch: nil },
     macos10_16: { name: "darwin", family: "darwin", release: "10.16", arch: nil },
     ubuntu1404: { name: "ubuntu", family: "debian", release: "14.04", arch: "x86_64" },
-    ubuntu: { name: "ubuntu", family: "debian", release: "20.04", arch: "x86_64" },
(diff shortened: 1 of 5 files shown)


test/helpers/mock_loader.rb
```diff
@@ -20,7 +20,7 @@ class MockLoader
     macos10_10: { name: "mac_os_x", family: "darwin", release: "10.10.4", arch: nil },
     macos10_16: { name: "darwin", family: "darwin", release: "10.16", arch: nil },
     ubuntu1404: { name: "ubuntu", family: "debian", release: "14.04", arch: "x86_64" },
-    ubuntu: { name: "ubuntu", family: "debian", release: "20.04", arch: "x86_64" },
+    ubuntu: { name: "ubuntu", family: "debian", release: "22.04", arch: "x86_64" },
```
Vasu1105, line 23: We want to keep 20.04 and 22.04 both and remove the Versions which is EOL like Ubuntu 14.04. Could you please confirm this?

Nik08: I think you are right. That makes sense. Thanks I will make the changes.

Vasu1105 (approved):

ahasunos (approved):

PR #6377, opened 2023-01-10
CFINSPEC-542 Bug fix for profiles with dependent profiles

Nik08:
## Description

This was reported by an Inspec user.

**Following is the excerpt from that:**

I have a new issue with automate compliance that came up recently. As of a few weeks ago, compliance runs all fail due to a dependency error. I have a compliance profile called fedramp-centos7-level1-server that includes a copy of cis-centos7-level1-server that we host on our git server. In overlay.rb:

`include_controls 'cis-centos7-level1-server' do`

and in inspec.yml:

```
depends:
  - name: 'cis-centos7-level1-server'
    git: 'ssh://git@path-to-our-git-server/comply/cis-centos7-level1-server.git'
    branch: 'master'
```

However, recently now whenever Chef runs I get this error:

```
Failed to load profile fedramp-centos7-level1-server: Failed to load source for controls/overlay.rb: Cannot load 'cis-centos7-level1-server' since it is not listed as a dependency of fedramp-centos7-level1-server.

Dependencies available from this context are:
    cis-centos7-level1-server-2.2.0-13

```
It seems to be expecting the version appended to the name? However, if I just try to add that version number to the end of the include_controls line in overlay.rb, inspec archive won't build the archive because the name of the dependency mismatches.How do I solve this dependency issue?We're still running Automate version 2, 20220329091442

**Cause of the issue:**

https://github.com/inspec/inspec/blob/17d7822365e24fdd8d3f1b58c78446d7dff67935/lib/inspec/dsl.rb#L93 This else condition needs a fix.

The above issue is happening because the dependency also has 2.2.0-13" "-" in this

**Possible ways to fix:**

1. Either try catching version and removing it from key on line 96 in dsl.rb, since this could break with versions which does not follow A.B.C pattern # Best solution
2. Or try to put condition that it does not append version when in source_version “-“ is present - Inside dependency_set.rb in line 29 and 45 # This is not a good solution
3. Either document how include controls can work by appending version. # New change documentation.

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/profiles.md
+++ b/docs-chef-io/content/inspec/profiles.md
@@ -404,6 +404,44 @@ As with the prior example, only `baseline-2` and `baseline-4` are executed, but
 if `baseline-2` fails, it will report with an impact of `0.5` instead of the
 originally-intended `1.0` impact.
 
+## Including or Selecting controls from a profile with same name and different version.
+
+When an inspec profile has dependency on another profile to it's specific version, then the controls can be included or selected by using profile name with version separated by `-`.
+
+Here, the Profile - A has following dependency:
+
+```yaml
+name: profile-a
+depends:
+  - name: ssh
+    git: https://github.com/dev-sec/ssh-baseline.git
+    tag: 2.6.0
+```
+
+And Profile - B has following dependency:
+
+```yaml
+name: profile-b
+depends:
+  - name: ssh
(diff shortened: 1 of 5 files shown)


clintoncwolfe (approved):
Thanks - looks good to me. Thanks for improving the regex comment.

PR #6378, opened 2023-01-11
CFINSPEC-553: Fix failing verify pipeline builds on ruby 3.0

Vasu1105:

## Description

*  Changes included in `verify.sh` to `bundle install` without kitchen group from Gemfile. Reason is verify pipeline is breaking for ruby 3.0 for Chef 18. Chef 18 works with ruby >= 3.1. In the future, we will be removing the support for ruby 3.0 as Chef 18 was never released with ruby 3.0. After discussing with @clintoncwolfe instead of updating the kitchen group currently, we preferred to exclude the kitchen group while bundling the gems.

https://buildkite.com/chef-oss/inspec-inspec-main-verify/builds/5197#01859c63-1d3d-4527-adc8-f6727097a90d/651-658

*  Removes all excluded groups from the bundle config set which was removed from the Gemfile.

## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/buildkite/verify.ps1
+++ b/.expeditor/buildkite/verify.ps1
@@ -5,7 +5,7 @@ ruby -v
 bundle --version
 
 echo "--- bundle install"
-bundle config set --local without tools maintenance deploy
+bundle config set --local without deploy
 bundle install --jobs=7 --retry=3
 
 echo "+++ bundle exec rake test:parallel"
--- a/.expeditor/buildkite/verify.ps1
+++ b/.expeditor/buildkite/verify.ps1
@@ -5,7 +5,7 @@ ruby -v
 bundle --version
 
 echo "--- bundle install"
-bundle config set --local without deploy
+bundle config set --local without deploy kitchen
 bundle install --jobs=7 --retry=3
 
 echo "+++ bundle exec rake test:parallel"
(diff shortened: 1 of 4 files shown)


omnibus/config/software/inspec.rb
```diff
@@ -35,7 +35,7 @@
 
   # We bundle install to ensure the versions of gems we are going to
   # appbundle-lock to are definitely installed
-  bundle "config set --local without test integration tools maintenance", env: env
+  bundle "config set --local without test", env: env
```
Vasu1105: @clintoncwolfe  Do you think we should add a group kitchen here too as the omnibus build does not have any dependency on the kitchen right?

.expeditor/buildkite/verify.ps1
```diff
@@ -5,7 +5,7 @@ ruby -v
 bundle --version
 
 echo "--- bundle install"
-bundle config set --local without tools maintenance deploy
+bundle config set --local without deploy
```
clintoncwolfe: I think we want to exclude kitchen here too. Why would we want to exclude it on the unix-based tests but not on the Windows based tests?

clintoncwolfe (commented):
I think `without kitchen` needs to be in place in verify.ps1 as well.

.expeditor/buildkite/verify.ps1
Vasu1105: Yes, good question. I choose not to include it here because we don't have the ruby 3.0 test running for Windows in our verify pipeline.

PR #6381, opened 2023-01-16
CFINSPEC-555 Ruby 3.1 support: Adds ucrt platform to resolve windows omnibus build.

Vasu1105:
## Description

Windows omnibus build fails with `cannot load such file -- 3.1/ffi_c (LoadError)` https://buildkite.com/chef/inspec-inspec-main-omnibus-adhoc/builds/343#01859bcb-7c3c-4851-a0a3-4eb677a5de04/6-6092. The reason is for ruby 3.1 64 bit Windows builds needs to be switched to use `UCRT64` toolchain instead of `MINGW64` which is the default.

Also after adding `UCRT64` it further started breaking where it was unable to download some gems like `ffi-1.15.5-x64-mingw-ucrt.gem` from artifactory. We need to verify whether adding the environment variable `IGNORE_ARTIFACTORY_RUBY_PROXY: true` to the omnibus release pipeline resolves the issue as manually adding this resolved the issue on the ad-hoc build.

## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -23,6 +23,7 @@ pipelines:
      # The git cache is corrupt more often than not. This always purges the cache.
      # https://chefio.atlassian.net/wiki/spaces/RELENGKB/pages/2204336129/Resolving+git+cache+build+errors+in+Omnibus
      - EXPIRE_CACHE: 1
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
Thanks!

PR #6403, opened 2023-02-03
CFINSPEC-550 Add RHEL 9 support

Vasu1105:

## Description

Adds Rhel9 platform to build matrix.
Here is the adhoc build reference for this change which is tested - https://buildkite.com/chef/inspec-inspec-main-omnibus-adhoc/builds/372
Also did basic manual testing of the generated packages on RHEL 9 aarch64 and x86_64 machines.
## Types of changes

## Checklist:

### DIFF
--- a/.expeditor/release.omnibus.yml
+++ b/.expeditor/release.omnibus.yml
@@ -28,6 +28,10 @@ builder-to-testers-map:
     - el-8-aarch64
   el-8-x86_64:
     - el-8-x86_64
+  el-9-aarch64:
+    - el-9-aarch64
+  el-9-x86_64:
+    - el-9-x86_64
   mac_os_x-10.15-x86_64:
     - mac_os_x-10.15-x86_64
     - mac_os_x-11-x86_64


Nik08:
We might need to document this support in README

Vasu1105:
> We might need to document this support in README

Once we release it we will document it. Currently, it's not released so doesn't make sense to me to update it right now.

Nik08:
> > We might need to document this support in README
> 
> Once we release it we will document it. Currently, it's not released so doesn't make sense to me to update it right now.

So are we waiting for some more changes to be done for this support? Or is this the only change for RHEL9 support? Because doc changes would be merged and released together with this right?

Vasu1105:
> > > We might need to document this support in README
> > 
> > 
> > Once we release it we will document it. Currently, it's not released so doesn't make sense to me to update it right now.
> 
> So are we waiting for some more changes to be done for this support? Or is this the only change for RHEL9 support? Because doc changes would be released together with this right?


I don't think any more changes are required. README doc changes reflect on the repo as soon as we merge the PR where as the RHEL9 package will be out only when we will release it. So for me, it is not making sense to mention it in the README right now. If there are any other docs that have mentions of platform support that goes with the release I need to check that or let me know if anyone else know that.

ahasunos (approved):
[Build](https://buildkite.com/chef/inspec-inspec-main-omnibus-adhoc/builds/372) is green ✅

Nik08 (approved):

PR #6408, opened 2023-02-07
CFINSPEC-576 : Fix profile gem dependency loading issue when dependent gem is required inside profile libraries.

Vasu1105:

## Description

This fixes the profile gem dependency loading issue when the gem is required inside profile libraries.
An earlier call to load_libraries to get the gem dependencies was resulting in actually loading the libraries before installing the gem due to which it was unable to find the required gem. Found that we can fetch the gem dependencies using locked_dependencies without calling the method load_libraries
 Here is the failing test https://buildkite.com/chef-oss/inspec-inspec-main-verify/builds/5273#01862a94-0632-4bc7-9e6e-9094102e73f7/1212-1215

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -383,24 +383,22 @@ module Inspec
       @runner_context
     end
 
-    def collect_gem_dependencies(profile_context)
+    def collect_gem_dependencies
       gem_dependencies = []
-      all_profiles = []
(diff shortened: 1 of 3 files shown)


Nik08 (approved):
The description was really useful in understanding this. Based on that, this change looks good to me.

clintoncwolfe (approved):

PR #6410, opened 2023-02-09
CFINSPEC-580 Dependent profiles fix for any scheme of version used in profiles

### LINKED ISSUE #6409: Dependent profiles are not found when profile version does not follow semver
In InSpec 5.21.29 dependent profiles specified by path that do not have version failed to run with the following error

```
ERROR: Failed to load profile vmware-vcsa-7.0-stig-baseline: Failed to load source for controls/photon.rb: Cannot load 'photon' since it is not listed as a dependency of vmware-vcsa-7.0-stig-baseline.

Dependencies available from this context are:
    eam-1.0.4
    lookup-1.0.4
    perfcharts-1.0.4
    photon-3.0 Release 1 Version 5
    postgres-1.0.4
    rhttpproxy-1.0.4
    sts-1.0.4
    vami-1.0.4
    vsphere-ui-1.0.4
```

# Describe the problem
Profiles with dependent profiles should run those profiles without having to specify a version.

inspec.yml depends block for the profile generating the error that worked in prior versions without issue.

```
depends:
  - name: eam
    path: eam
  - name: lookup
    path: lookup
  - name: perfcharts
    path: perfcharts
  - name: photon
    path: ../vmware-photon-3.0-stig-inspec-baseline
  - name: postgres
    path: postgresql
  - name: rhttpproxy
    path: rhttpproxy
  - name: sts
    path: sts
  - name: vami
    path: vami
  - name: vsphere-ui
    path: ui
```

Possibly related to https://github.com/inspec/inspec/pull/6377

Nik08:

## Description

Fix to load dependent profiles which follow any scheme for versioning profiles. Before this, it was only restricted to semver versioning.

## Related Issue

Fixes #6409 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -95,8 +95,10 @@ module Inspec::DSL
         # 1. Fetching VERSION from a profile dependency name which is in a format NAME-VERSION.
         # 2. Matching original profile dependency name with profile name used with include or require control DSL.
         source_version = value.source_version
-        profile_id_key = key.split("-#{source_version}")[0]
-        new_profile_id = key if profile_id_key == profile_id
+        unless source_version.blank?
+          profile_id_key = key.split("-#{source_version}")[0]
+          new_profile_id = key if profile_id_key == profile_id
+        end
       end
     end
     # If dep profile does not contain a source version, key does not contain a version as well. In that case new_profile_id will be always nil and instead profile_id would be used to fetch profile from dependency list.
--- a/lib/inspec/dsl.rb
+++ b/lib/inspec/dsl.rb
@@ -91,16 +91,12 @@ module Inspec::DSL
     if profile_version
       new_profile_id = "#{profile_id}-#{profile_version}"
     else
-      # This scary regex is used to match version following semantic Versioning (SemVer). Thanks to https://ihateregex.io/expr/semver/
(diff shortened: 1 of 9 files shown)


lib/inspec/dsl.rb
```diff
@@ -91,16 +91,12 @@ def self.load_spec_files_for_profile(bind_context, opts, &block)
...
-      dependencies.list.keys.each do |key|
+      dependencies.list.each do |key, value|
         # 1. Fetching VERSION from a profile dependency name which is in a format NAME-VERSION.
         # 2. Matching original profile dependency name with profile name used with include or require control DSL.
-        fetching_semver = key.match(regex_for_semver).to_s
-        unless fetching_semver.nil? || fetching_semver.empty?
-          profile_id_key = key.split("-#{fetching_semver}")[0]
-          new_profile_id = key if profile_id_key == profile_id
-        end
+        source_version = value.source_version
```
Vasu1105, line 97: in case value.source_version is empty or nil. Are we covering that case here?

Nik08: I will cover it up.

Vasu1105 (approved):
I only had a doubt about using blank but that got clarified by @Nik08. So this looks good to me.

bdwyertech:
This introduces a dependency on ActiveSupport to use the blank method.  I am receiving errors and had to revert to 5.21.29 to resolve.

@Nik08 @Vasu1105 -- can we please get rid of this or add active_support to the gem dependencies?

bdwyertech:
> I only had a doubt about using blank but that got clarified by @Nik08. So this looks good to me.

You were correct in questioning usage of blank.

Nik08:
@bdwyertech Inspec uses train which has a listed gem dependency on `active-support` https://github.com/inspec/train/blob/main/train.gemspec#L34.
And this has been used in inspec for over last 2 years now. There is the usage of `blank?` in a lot of instances within InSpec.

Maybe resolving your dependency on [train](https://github.com/inspec/train), would help with this.

PR #6483, opened 2023-03-10
Fix Syntax

bryaend:
Proper syntax for `its('size')` is `{ should be <|>|<=|>= <size>}`

Updated syntax

## Description

Code written as currently documented fails due to a syntax error.

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/file.md
+++ b/docs-chef-io/content/inspec/resources/file.md
@@ -299,13 +299,13 @@ The `size` property tests if a file's size matches, is greater than, or is less 
 Greater than:
 
 ```ruby
-    its('size') { should > 64 }
+    its('size') { should be > 64 }
 ```
 
 Less than:
 
 ```ruby
-    its('size') { should < 10240 }
+    its('size') { should be < 10240 }
 ```
 
 ### type


Vasu1105:
Both the syntax seems to be working. Could you please attach a screenshot of the error that you get?

Screenshot 2023-03-16 at 11 22 03 AM

bryaend:
I think I might see what led us astray here. When we try to test for a file that doesn't exist, we get the `undefined method` error, but if we point it at a file that we know exists it behaves as expected:

```
  ×  test-file-size: Testing different file size methods (3 failed)
     ✔  File /var/log/messages is expected to exist
     ✔  File /var/log/messages size is expected to > 0
     ✔  File /var/log/messages size is expected to be > 0
     ×  File /tmp/imnotreal is expected to exist
     expected File /tmp/imnotreal to exist
     ×  File /tmp/imnotreal size is expected to > 0
     undefined method `>' for nil:NilClass
     ×  File /tmp/imnotreal size is expected to be > 0
     expected: > 0
          got:   nil
```

So I guess the edge case is if your control is "Make sure this file size is greater/less than `x`" then you'll get this output on your size check if the file doesn't exist.

Is it worth documenting that this will throw an exception if the file doesn't exist or is the right approach to write a bug fix to catch that edge case at runtime?

IanMadd:
We've got the [matchers page](https://docs.chef.io/inspec/matchers/), if we're telling users they can do `{ should > x }`, can we add that to the matchers page and document how that's different from the other matchers. Assuming we're keeping this in the file.md doc.

Vasu1105:
> We've got the [matchers page](https://docs.chef.io/inspec/matchers/), if we're telling users they can do `{ should > x }`, can we add that to the matchers page and document how that's different from the other matchers. Assuming we're keeping this in the file.md doc.

Yes, we should. @clintoncwolfe what do you think?

Vasu1105:
What is the version of inspec you are using @bryaend? I am still not able to produce the exact issue.

bryaend:
> What is the version of inspec you are using @bryaend? I am still not able to produce the exact issue.

I'm running `5.21.29`

clintoncwolfe:
I think the "be" syntax is both safer (less likely to throw an execption) and more readable. Let's document that.

PR #6484, opened 2023-03-11
fix: ensure Invoke-WebRequest headers can be configured

amlodzianowski:
Fix for configuring headers on a `http` resource on Windows using remote transport

## Description
We encountered an issue for the `http` resource on Windows when running InSpec via Test Kitchen and trying to configure headers. The code in question that we use:
```
http('http://169.254.169.254/latest/api/token', method: 'PUT', headers: { 'X-aws-ec2-metadata-token-ttl-seconds': '600' })
```
throws the following error:
```
undefined method `<<' for nil:NilClass
```

We narrowed the issue down to the following block which tries to append an array of strings to `nil` in a `.each` loop:
```
          request_header_string = nil
          request_headers.each do |k, v|
            request_header_string << " #{k} = #{v}"
          end
```

Not sure if Ruby ever supported such behavior, but 2.7 and 3.1 do not:
```
irb(main):001:0> request_header_string = nil
=> nil
irb(main):002:0> request_header_string << " test"
(irb):2:in `<main>': undefined method `<<' for nil:NilClass (NoMethodError)
        from /usr/local/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        from /usr/local/bin/irb:25:in `load'
        from /usr/local/bin/irb:25:in `<main>'
irb(main):003:0> 
```

Also added a fix to ensure that keys and values for `-Headers` are quoted, otherwise PowerShell assumes we are referring to variables. End result should look like:
```
"-Headers @{ 'X-aws-ec2-metadata-token-ttl-seconds' = '600'; 'test' = 'yy'}"
```
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.3#example-9-skipping-header-validation

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ x] I have read the **CONTRIBUTING** document.


Vasu1105 (approved):
LGTM. @amlodzianowski Could you please check if you can add the test for this?

amlodzianowski:
Thanks @Vasu1105 ! I added a unit test for this case, looks like the pipeline just passed. Please let me know if there is anything else I should add to this PR.

Vasu1105:
That looks good to me @amlodzianowski. Thanks!!

PR #6489, opened 2023-03-16
CHEF-1458 Multiple values changes in SimpleConfig library

Nik08:
## Description

Multiple values changes in SimpleConfig library.

**Existing behavior:**
Returns `["baz boz bop abc def bhb hbjb"]` for key value `biz`

**New change:**
Returns `["baz", "boz", "bop", "abc", "def", "bhb", "hbjb"]`

**Usage**:

New default option for Simple Config library:
 `multiple_values_regex` which will be helpful in splitting values by any specified delimiter regex.

`SimpleConfig.new(content_from_config, multiple_values: true, multiple_value_regex: /\s+/)`

This will enable testing config with values
```
foo:  bar
biz:  baz boz bop
biz:  abc def bhb hbjb
foz:  foo [!deny] bar
```

With a test like this:

`its('biz') { should cmp %w(baz boz bop abc def bhb hbjb) }`

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/utils/simpleconfig.rb
+++ b/lib/inspec/utils/simpleconfig.rb
@@ -61,8 +61,8 @@ class SimpleConfig
 
     if opts[:multiple_values]
       @vals[m[1]] ||= []
-      value_to_array = values.split(" ")
-      if value_to_array.count > 1
+      if opts[:multiple_value_regex] # can be used only if multiple values is set as true
+        value_to_array = values.split(opts[:multiple_value_regex])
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
This look fine to me, if a bit obscure - is this what the original requestor wanted?

Nik08:
> This look fine to me, if a bit obscure - is this what the original requestor wanted?

Yes, confirmed with the requester.

PR #6497, opened 2023-03-27
Fix inspec-4 CI pipeline

ahasunos:
## Description
This PR removes the html-proofer dependency as no test is dependent on the gem.

## Types of changes

## Checklist:


ahasunos:
Taking the #6496 approach because we intend to support inspec-4 with Ruby version 3.0 and above.

PR #6511, opened 2023-05-03
CHEF-1631 Clarify that command timeout default was withdrawn

clintoncwolfe:
When we withdrew the command timeout default on #5472 we forgot to update the docs on the command resource page. A customer found those docs and was misled.

## Types of changes

## Checklist:

### DIFF
--- a/docs-chef-io/content/inspec/resources/command.md
+++ b/docs-chef-io/content/inspec/resources/command.md
@@ -147,7 +147,7 @@ end
 ```
 
 This example would run the `find` command for up to 300 seconds, then give up and fail the control if it exceeded that time.
-On supported target platforms, the default timeout is 3600 seconds (one hour).
+On supported target platforms, briefly from versions 4.31.0 to when it was withdrawn in 4.33.1, a default of 3600 seconds was provided, but this proved problematic and was removed.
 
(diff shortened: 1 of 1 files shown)


Nik08 (approved):

ahasunos (approved):

docs-chef-io/content/inspec/resources/command.md
```diff
@@ -147,7 +147,7 @@ end
 ```
 
 This example would run the `find` command for up to 300 seconds, then give up and fail the control if it exceeded that time.
-On supported target platforms, the default timeout is 3600 seconds (one hour).
+On supported target platforms, briefly from versions 4.31.0 to when it was withdrawn in 4.33.1, a default of 3600 seconds was provided, but this proved problematic and was removed.
```
IanMadd: We should probably just yank this whole line. It was only a few versions. It won't hurt anyone if this goes away and users of those versions think they have to specify a value.

```suggestion
```

On line 141 it says 

> command resource takes an optional `timeout:` parameter

Should that be changed so that it says it's required?

clintoncwolfe: No, the parameter is optional - if not provided, the command will not timeout and will run indefinitely (subject to other constraints)

IanMadd: Ok. Yeah, I'd just delete the line. There's no sense in keeping a line that applies to only 2 minor versions.

IanMadd (approved):

PR #6514, opened 2023-05-08
add train-kubernetes as a dependency

sathish-progress:
add `train-kubernetes` as a dependency to inspec

## Description

installing `train-kubernetes` as plugin has a series of issues and behaves differently for every platform. Hence the most consistent way to load the `train-kubernetes` plugin is to make it available as dependency for inspec so every customer has a seamless experience.

## Types of changes

## Checklist:


sathish-progress:
already exists https://github.com/inspec/inspec/pull/6512/files

PR #6526, opened 2023-06-06
Update Docker base image to be ubuntu 22.04

### LINKED ISSUE #6521: Update the outdated ubuntu image in Dockerfile
### 🎛 Description

The base image `ubuntu:18.04` is quite old. Security scanning of the image raises warnings because of the openssl 1.x version bundled. Also Ubuntu 18.04 LTS has hit its eol since May 31st 2023. While `inspec` is normally used as as a security scanning tool, lower down the security level of the tool image is not quite accetable.

### 💁 Possible Solutions

Update the base image defined in [Dockerfile](https://github.com/inspec/inspec/blob/main/Dockerfile#L1) to use more recent version, e.g. ubuntu 22.10.

Thanks.

clintoncwolfe:
Updates the Dockerfile to use Ubuntu 22.04, as 18.04 is now EOL. I did a local docker build as a test.

## Related Issue

Fixes #6521

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:22.04
 LABEL maintainer="Chef Software, Inc. <docker@chef.io>"
 
 ARG VERSION=5.22.3


ahasunos (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:22.04
```
ahasunos, line 1: The docker image fails to run on an M1 system with an error as: `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`

The fix to make it run was to change it to `FROM --platform=linux/amd64 ubuntu:22.04`:
```
diff --git a/Dockerfile b/Dockerfile
index 981fc4dcd..d267cf59d 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM --platform=linux/amd64 ubuntu:22.04
 LABEL maintainer="Chef Software, Inc. <docker@chef.io>"
 
 ARG VERSION=5.22.3
```

PR #6528, opened 2023-06-08
Remove Undocumented Dependency on ActiveSupport

### LINKED ISSUE #6527: Undocumented Dependency on ActiveSupport
# Describe the problem
https://github.com/inspec/inspec/pull/6410 introduced an undocumented dependency on ActiveSupport due to its usage of the .blank? method on string values.  This function is provided by ActiveSupport

## Possible Solution
Use something other than .blank? e.g. .empty? instead

bdwyertech:
Resolves https://github.com/inspec/inspec/issues/6527 

https://github.com/inspec/inspec/pull/6410 introduced an undocumented dependency on ActiveSupport due to its usage of the .blank? method on string values. This function is provided by ActiveSupport

I consume inspec-core-bin in some custom tooling and #6410 introduced a stack trace.

### DIFF
--- a/lib/inspec/dependencies/dependency_set.rb
+++ b/lib/inspec/dependencies/dependency_set.rb
@@ -26,7 +26,7 @@ module Inspec
       dep_list = {}
       dependencies.each do |d|
         # if depedent profile does not have a source version then only name is used in dependency hash
-        key_name = (d.source_version.blank? ? "#{d.name}" : "#{d.name}-#{d.source_version}") rescue "#{d.name}"
+        key_name = (d.source_version.empty? ? "#{d.name}" : "#{d.name}-#{d.source_version}") rescue "#{d.name}"
         dep_list[key_name] = d
       end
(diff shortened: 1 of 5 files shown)


lib/inspec/dependencies/dependency_set.rb
```diff
@@ -26,7 +26,7 @@ def self.from_array(dependencies, cwd, cache, backend)
       dep_list = {}
       dependencies.each do |d|
         # if depedent profile does not have a source version then only name is used in dependency hash
-        key_name = (d.source_version.blank? ? "#{d.name}" : "#{d.name}-#{d.source_version}") rescue "#{d.name}"
```
Nik08, line 29: Having check both for `empty` and `nil` should work, since `blank?` checks for both.

Nik08:
This could be closed, since this fix will be covered and tested in PR https://github.com/inspec/inspec/pull/6914
Thanks @bdwyertech for this submission.

PR #6539, opened 2023-06-23
[W.I.P] CHEF-3759 inspec parallel: Error logging changes to fix renaming file error in windows

Nik08:
## Description

Simplifies error log handling under Windows, resolving a critical issue in which inspec parallel jobs would immediately die. Jobs now work on this branch, but we are currently looking at an issue regarding the cleanup of empty error log files.

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
@@ -74,14 +74,11 @@ module InspecPlugins
 
         # Construct command-line invocation
         child_pid = nil
-        error_log_file_name = "#{Time.now.nsec}.err"
 
         begin
           cmd = "#{$0} #{sub_cmd} #{invocation}"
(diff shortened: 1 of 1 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -34,7 +34,7 @@ def run
           cleanup_child_processes
           sleep 0.1
         end
-        cleanup_empty_error_log_files
+        cleanup_empty_error_log_files unless Inspec.locally_windows?
```
clintoncwolfe, line 37: While InSpec Parallel exec now works on Windows, it generates spurious error logs regarding deleting empty log files. I don't think it is so important to delete empty log files that we should cause an error. I think - at least temporarily - on windows we should not delete empty error log files.

clintoncwolfe:
We have a different approach we're trying first.

PR #6540, opened 2023-06-24
CHEF-3759 inspec parallel: Error logging changes to fix renaming file error in windows - Final changes

Nik08:
## Description

Issue:
Simplifies error log handling under Windows, resolving a critical issue in which inspec parallel jobs would immediately die. 

This covers:
1. Solution to renaming and deletion of error logs
2. Proper logging
3. Renaming error log file with child pid on control-c interrupt.

The PR https://github.com/inspec/inspec/pull/6539 could be closed, after the changes here are verified in code review.
## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
@@ -71,6 +71,24 @@ module InspecPlugins
       def kill_child_processes
         @child_tracker.each do |pid, info|
           Process.kill("SIGKILL", pid)
+        rescue Exception => e
+          $stderr.puts "Error while shutting down process #{pid}: #{e.message}"
+        end
+        # Waiting for child processes to die after they have been killed
+        wait_for_child_processes_to_die
+      end
+
+      def wait_for_child_processes_to_die
+        until @child_tracker.empty?
+          begin
+            exited_pid = Process.waitpid(-1, Process::WNOHANG)
+            @child_tracker.delete exited_pid if exited_pid && exited_pid > 0
+            sleep 1
+          rescue Errno::ECHILD
+            Inspec::Log.info "Processes shutdown complete!"
+          rescue Exception => e
+            Inspec::Log.debug "Error while waiting for child processes to shutdown: #{e.message}"
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


Vasu1105 (approved):
This looks good to me.

lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -34,6 +35,10 @@ def run
           cleanup_child_processes
           sleep 0.1
         end
+
+        # Requires renaming operations on windows only
+        # Do Rename and delete operations after all child processes have exited successfully
```
clintoncwolfe, line 40: There is an unfortunate effect that if you interrupt (Ctl-C) the parallel process controller, the files do not get renamed, so they end up being named with the timestamp. That would be very frustrating. Could we improve that? Can we call the rename hook from the control-c interrupt?

Nik08: Done, but the way I have handled, it also captures interrupt exceptions

lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -72,24 +83,40 @@ def spawn_another_process
...
           log_msg = "#{Time.now.iso8601} Start Time: #{Time.now}\n#{Time.now.iso8601} Arguments: #{invocation}\n"
-          child_pid = Process.spawn(cmd, out: parent_writer, err: error_log_file_name)
-          # Rename error log file if exist
-          rename_error_log(error_log_file_name, child_pid) if File.exist?(error_log_file_name)
+          child_pid = Process.spawn(cmd, out: parent_writer, err: error_log_file.path)
+
+          # Trap Control-c Interrupts
+          trap("SIGINT") do
+            # Kill child process instantly
+            Process.kill("SIGINT", child_pid)
```
Vasu1105: This fix does not seem to be working for me. When I interrupt the process using ctrl+c, it hangs on there.

clintoncwolfe:
OK, Nikita and I have stared at this problem for several days, and we've cornered the bug to an edge case: occasionally, on Windows, when you hit Ctl-C, there is a race condition that occasionally results in a failure to rename the error logs, which we report with an informative error message. We're still looking into addressing that issue, but the major blocker - not being able to run inspec parallel on Windows at all - is addressed.

clintoncwolfe:
This branch has been merged in the integration repo.

PR #6545, opened 2023-06-27
[W.I.P] Fix inspec parallel hangs on windows for certain CIS profile

Vasu1105:
## Description

This PR provides a fix for an issue in which InSpec parallel would hang on certain profiles. It was eventually traced to the EOF detection code in the child-status reporter, which is used to communicate between the child processes and the master process. On Windows, we need to send an explicit EOF marker, which means detecting when we are done; previously this was done by watching the control count, but it was found that certain profiles (those containing several zero-test controls, which is common in large CIS profiles) cause the control count to never be accurate. While that issue will be resolved another day, in the meantime, this PR fixes the hanging issue by moving the end--of-report detection to a rSpec Formatter close() handler, which is more reliable.

Tested on Windows 2019 and MacOS.

## Types of changes

## Checklist:


clintoncwolfe:
This change was accepted in to InSpec 6, but was pulled into the downstream integration repo which is not yet public. We are thus closing this PR.

PR #6568, opened 2023-07-14
CHEF-4115 Added ability to merge reporter configurations from both CLI and config

Nik08:
## Description

Added ability to merge reporter configurations from both CLI and config

**User story details:**

If you provide reporter options on the CLI using --reporter, those options entirely clobber whatever is in the config file specified by --config.

It would be a better UX if the options merged.

## Types of changes

## Checklist:


test/functional/inspec_exec_test.rb
```diff
@@ -1002,6 +1002,30 @@ def stderr
       end
     end
 
+    describe "when using the --config option and --reporter option to configure reporter in a run in a correct manner" do
+      outpath = Dir.tmpdir
+      let(:cli_args) { "--config " + File.join(config_dir_path, "json-config", "reporter-json-config.json") + " --reporter json:#{outpath}/test-json-report.json" }
```
clintoncwolfe: So, this isn't a very clear test to me. I want a test that says, clearly, the --config will create THIS file, and I will check for it; and the --reporter option will create THAT file, and I will check for it. 

If both files exist, then the --reporter cannot have clobbered the --config, and the bug is clearly fixed.

Why are we expecting anything at all on STDOUT?

docs-chef-io/content/inspec/reporters.md
```diff
@@ -15,7 +15,7 @@ A `reporter` is a facility for formatting and delivering the results of a Chef I
 
 Chef InSpec allows you to output your test results to one or more reporters.
 
-Configure the reporter(s) using either the `--reporter` option or as part of the general configuration file using the `--config` (or `--json-config`, prior to v3.6) option. While you can configure multiple reporters to write to different files, only one reporter can output to the screen(stdout).
+Configure the reporter(s) using the `--reporter` option or as part of the general configuration file using the `--config` (or `--json-config`, prior to v3.6) option. Or it could also be configured by using both the options together. While you can configure multiple reporters to write to different files, only one reporter can output to the screen(stdout).
```
clintoncwolfe: How about "Both the --reporter and --config options may be used, in which case the options are merged."
As is, it reads like you are suggesting both --config and --json-config may be used.

test/functional/inspec_exec_test.rb
Nik08: I have forgotten to commit and push the config file that I created for this test, so that would have cleared up why I am testing `stdout`.
So the above tests intend to test the merging of options passed using reporter CLI and using config.
So if the config file says STDOUT: true, then it should obey that.
Yes, the best test would be with STDOUT false and the file path, but testing that by providing a dynamic path in config file seemed difficult. Let me try modifying these tests.

test/functional/inspec_exec_test.rb
```diff
@@ -1085,6 +1085,34 @@ def stderr
...
+      _(stdout).must_include "0 controls skipped"
+      _(stderr).must_be_empty
+      assert_exit_code 0, out
+    end
+  end
+
+  describe "When specifying a config file and --reporter option to configure reporter with stdout true from both the options" do
+    let(:cli_args) { "--config " + File.join(config_dir_path, "json-config", "reporter-json-config.json") + " --reporter json html2" }
+    let(:run_result) { run_inspec_process("exec " + File.join(profile_path, "basic_profile") + " " + cli_args) }
+    it "should raise error that only single reporter can have output to stdout" do
```
Vasu1105, line 1110: Instead of raising an error can we show the warning and give precedence to any of them? That way we won't be breaking the flow of generating a report.

Nik08: This is the current way of how if you add multiple reporter which have `stdout: true`, then it give user this message to change the way they are using reporter.
This happens in all the scenarios:
1. While using `--reporter cli json` - No file mentioned, but multiple reporters to output
2. While using `stdout: true` in config file, for multiple reporter options
3. While using `stdout: true` in config file in reporter options and also using `--reporter REPORTER_OPTION_WITHOUT_FILE` ( Replicated same behavior for this newly added scenario now )

So this is the current behaviour. Not bringing a new change with this implementation.

clintoncwolfe (approved):
Looks good - good demo!

PR #6578, opened 2023-07-25
CHEF-4559 Extended support to inspec parallel for reporters using config

Nik08:
## Description

Extended support to inspec parallel for reporters using config, through config file or **STDIN**

Following scenarios using config are covered :

1. `inspec parallel exec test/fixtures/profiles/complete-metadata -o simple-options.txt -j 3 --config test-config.json`
2. `echo '{"reporter":{"json":{"stdout": false, file: "test-path/test-json.json"}}}' | inspec parallel exec test/fixtures/profiles/complete-metadata -o simple-options.txt --config=-`
3. Using a default `config.json` file, if configured.

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/validator.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/validator.rb
@@ -127,14 +127,7 @@ module InspecPlugins
         end
 
         reporter_config = JSON.parse(config_content)["reporter"] unless config_content.nil? || config_content.empty?
-        if reporter_config
-          reporter_config.each do |reporter, config|
-            unless config["stdout"] == false
-              invocation_data[:validation_errors] << "Config should have reporter option with stdout `false` for each invocation which is not using --reporter option in options file"
-              break
-            end
-          end
-        else
+        unless reporter_config
           invocation_data[:validation_errors] << "Config should have reporter option specified for each invocation which is not using --reporter option in options file"
         end
         @config_content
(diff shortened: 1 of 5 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/validator.rb
```diff
@@ -101,6 +106,47 @@ def check_for_reporter_options(invocation_data)
...
+            invocation_data[:validation_errors] << "Could not read configuration file at #{config_opts}"
+            return
+          end
+          @config_content = File.open(config_opts).read
+        end
+
+        reporter_config = JSON.parse(config_content)["reporter"] unless config_content.nil? || config_content.empty?
+        if reporter_config
+          reporter_config.each do |reporter, config|
+            unless config["stdout"] == false
```
clintoncwolfe: Why do we insist on this painful requirement? Can't we just assume all reporters that have a filename are not outputting to stdout?

clintoncwolfe: Also, why is this logic duplicated way over here in InSpec Parallel?

Nik08: > Why do we insist on this painful requirement? Can't we just assume all reporters that have a filename are not outputting to stdout?

It is possible for configurations to not have a file path, in case we are configuring it with automate. But stdout false is basic for every configuration which is either redirecting it to a file or to automate.

Nik08: > Also, why is this logic duplicated way over here in InSpec Parallel?

Yes, it is a logic fetched from `config.rb` to here to make this work, because it could not be reused from the config class itself. Config class expects inputs in different formats and so do the functions defined in that class. While here in the parallel validator class, we only have a config file path.

PR #6593, opened 2023-08-11
[W.I.P] Fix: Waived control executes describe block

Vasu1105:
## Description

Waived control should not be executed and skipped without execution of describe block.

This change helps in eliminating the efforts to use options `--filter-waived-controls` and `--retain-waiver-data`.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -46,12 +46,11 @@ module Inspec
       return unless block_given?
 
       begin
-        instance_eval(&block)
-
-        # By applying waivers *after* the instance eval, we assure that
-        # waivers have higher precedence than only_if.
         __apply_waivers
 
+        unless waiver_is_valid?
+          instance_eval(&block)
+        end
       rescue SystemStackError, StandardError => e
         # We've encountered an exception while trying to eval the code inside the
         # control block. We need to prevent the exception from bubbling up, and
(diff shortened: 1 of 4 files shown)


lib/inspec/rule.rb
```diff
@@ -46,12 +46,11 @@ def initialize(id, profile_id, resource_dsl, opts, &block)
       return unless block_given?
 
       begin
-        instance_eval(&block)
-
-        # By applying waivers *after* the instance eval, we assure that
-        # waivers have higher precedence than only_if.
```
Vasu1105, line 52: Not saw any impact on the precedence even if _apply_waivers is called before instance_eval. Calling _apply_waivers here is required to get all the waiver data so that only in the valid condition we will evaluate the describe block, for example, if the run is set to true and the if the waiver is expired.

lib/inspec/secrets/yaml.rb
```diff
@@ -25,8 +25,8 @@ def initialize(target)
       end
 
       if @inputs == false || !@inputs.is_a?(Hash)
-        Inspec::Log.warn("#{self.class} unable to parse #{target}: invalid YAML or contents is not a Hash")
         @inputs = nil
+        raise("#{self.class} unable to parse #{target}: invalid YAML or contents is not a Hash")
```
Vasu1105, line 29: We should raise the error if the yaml is corrupted or it's empty.

clintoncwolfe: Does this help address CHEF-3849 ? - abort immediately if waiver file is corrupt?

Vasu1105: Yes.

Vasu1105:
Closing this in respect of #6807 as the PR was open against inspec. And fix need to be go against main.

PR #6604, opened 2023-08-17
Expeditor Config Changes - Keep the verify pipeline alive

ahasunos:
## Description

This PR adds a dummy verify pipeline in order to keep the pipeline alive to preserve the history of builds in the same.

## Types of changes

## Checklist:


sean-sype-simmons (approved):
lgtm, you keep the same pipeline name; which should NOT delete it.

PR #6610, opened 2023-08-17
Move chef-licensing gem dependency to inspec gemspec file

Vasu1105:
## Description

Moves chef-licensing gem dependency to the inspec gemspec file .

Continues and formalizes the practice of fetching Chef-owned gems, like chef-licensing, from internal Chef artifactory.

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,8 @@
+# For Chef internal builds, allows preview versions of gems if available.
+if ENV["ARTIFACTORY_BASE_URL"]
+  source ENV["ARTIFACTORY_BASE_URL"] + "/artifactory/api/gems/omnibus-gems-local/"
+end
+
 source "https://rubygems.org"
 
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -47,7 +47,3 @@ end
 group :deploy do
   gem "inquirer"
 end
-
-source "https://artifactory-internal.ps.chef.co/artifactory/api/gems/omnibus-gems-local/" do
```
clintoncwolfe, line 51: So, we want our builds to source from Artifactory. To continue to make this effect happen, we need to leave this stanza in place. However, to make it consumable by the public (ie not break CINC auditor), we need to conditionalize this under an environment variable. I'm going to propose CHEF_USE_INTERNAL_GEMSOURCES. 

I'll push a change to that effect.

PR #6627, opened 2023-08-28
Explicitly list which gems to pull from Artifactory

clintoncwolfe:
## Description

Try explicit list of gems to pull from artifactory to avoid issue on CI.

Starts with just chef-licensing. Should likely list more.

The issue appears to be corrupted metadata, but we can't repro it locally, which makes us think it might be a connectivity issue?

https://buildkite.com/chef/inspec-inspec-main-verify-private/builds/40#018a2d7e-98ed-436d-814f-43a3b6c40394/459-464

## Types of changes

## Checklist:


clintoncwolfe:
I'm going to go ahead and merge this, as it gets the main verify green and we need to rebase an important PR.

PR #6644, opened 2023-09-01
CHEF-3849: InSpec should exit quickly and clearly if waiver file is malformed/corrupt

Vasu1105:
## Description

1) If you provide a malformed waiver file - one that fails YAML lint, in other words - inspec will continue to run, showing an error as the first control failing with a control source code error.
This buries the error message(Warning) deeply in the InSpec reporting output, which on a large profile may be hundred of kilobytes of text.
It also wastes the user’s time, since on a large profile, we know waiver file is bad immediately, but we doggedly execute the whole profile (which may take hours) before finally returning unusable (incorrect, unwaivered) results.

2) In case of invalid parameters in the waiver file currently it shows a warning message whereas it should raise the error in case of required parameters are not present in the waiver file.

This fix includes
 * Detect a malformed waiver file and exit immediately. 
 * Raise error in case of required parameter is not present in the waiver file.
 * Introduced a new exception class to handle invalid formatting of waiver file.
 * Updated code to validate the required parameter for yaml and json file for each waived control.
 
In case of an empty waiver file, extra parameters, or nil header (in CSV formatted) waiver file it will show a warning message that there will not be any behavioral changes.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/secrets/yaml.rb
+++ b/lib/inspec/secrets/yaml.rb
@@ -25,7 +25,7 @@ module Secrets
       end
 
       # In case of empty yaml file raise the warning else raise the parsing error.
-      if @inputs.nil? || @inputs == false
+      if !@inputs || @inputs.empty?
         Inspec::Log.warn("Unable to parse #{target}: YAML file is empty.")
         @inputs = nil if @inputs == false
       elsif !@inputs.is_a?(Hash)
--- a/lib/inspec/secrets/yaml.rb
+++ b/lib/inspec/secrets/yaml.rb
@@ -26,7 +26,7 @@ module Secrets
 
       # In case of empty yaml file raise the warning else raise the parsing error.
       if @inputs.nil? || @inputs == false
-        Inspec::Log.warn("#{self.class} unable to parse #{target}: Yaml file is empty.")
+        Inspec::Log.warn("Unable to parse #{target}: YAML file is empty.")
         @inputs = nil if @inputs == false
       elsif !@inputs.is_a?(Hash)
         # Exits with usage error.
--- a/lib/inspec/secrets/yaml.rb
+++ b/lib/inspec/secrets/yaml.rb
@@ -24,11 +24,17 @@ module Secrets
         @inputs = ::YAML.load_file(target)
       end
 
-      if @inputs == false || !@inputs.is_a?(Hash)
-        Inspec::Log.warn("#{self.class} unable to parse #{target}: invalid YAML or contents is not a Hash")
-        @inputs = nil
+      # In case of empty yaml file raise the warning else raise the parsing error.
+      if @inputs.nil? || @inputs == false
+        Inspec::Log.warn("#{self.class} unable to parse #{target}: Yaml file is empty.")
+        @inputs = nil if @inputs == false
+      elsif !@inputs.is_a?(Hash)
+        # Exits with usage error.
+        Inspec::Log.error("Unable to parse #{target}: invalid YAML or contents is not a Hash")
+        Inspec::UI.new.exit(:usage_error)
       end
     rescue => e
+      # Any other error related to Yaml parsing will be raised here.
       raise "Error reading InSpec inputs: #{e}"
     end
   end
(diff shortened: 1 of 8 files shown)


lib/inspec/secrets/yaml.rb
```diff
@@ -24,11 +24,17 @@ def initialize(target)
         @inputs = ::YAML.load_file(target)
       end
 
-      if @inputs == false || !@inputs.is_a?(Hash)
-        Inspec::Log.warn("#{self.class} unable to parse #{target}: invalid YAML or contents is not a Hash")
-        @inputs = nil
+      # In case of empty yaml file raise the warning else raise the parsing error.
+      if @inputs.nil? || @inputs == false
```
sathish-progress: why would `@inputs` be false?

sathish-progress: For this case, shouldn't be `.empty?`

@inputs.empty? returns `(irb):5:in <main>: undefined method 'empty?' for nil:NilClass (NoMethodError)` as @inputs returns nil in case of file is empty.

Vasu1105: > why would `@inputs` be false?

As per my knowledge, YAML load_file used to return false in case of empty files so I thought to keep both the checks in place. Currently, it returns nil. ( this is the only reference I found https://github.com/ruby/psych/issues/149)

Vasu1105: in ruby 2.7.0 it returns false
```
2.7.0 :002 > inputs = ::YAML.load_file('/workspace/profiles/ssh-baseline/files/empty-waivers1.yaml')
2.7.0 :003 > inputs.nil?
 => false 
2.7.0 :004 > inputs.class
 => FalseClass 
```

sathish-progress:
```suggestion
      if !@inputs || @inputs.empty?
```

sathish-progress (approved):
LGTM

lib/inspec/waiver_file_reader.rb
```diff
@@ -20,9 +20,12 @@ def self.read_waivers_from_file(profile_id, files)
 
         if data.nil?
           raise Inspec::Exceptions::WaiversFileNotReadable,
-              "Cannot find parser for waivers file '#{file_path}'. " \
+              "Cannot find parser for waivers file." \
```
Nik08, line 23: Why are we removing `file_path` from this message? Any reason?

lib/inspec/secrets/yaml.rb
```diff
@@ -24,12 +24,18 @@ def initialize(target)
         @inputs = ::YAML.load_file(target)
       end
 
-      if @inputs == false || !@inputs.is_a?(Hash)
-        Inspec::Log.warn("#{self.class} unable to parse #{target}: invalid YAML or contents is not a Hash")
-        @inputs = nil
+      # In case of empty yaml file raise the warning else raise the parsing error.
+      if !@inputs || @inputs.empty?
+        Inspec::Log.warn("Unable to parse #{target}: YAML file is empty.")
+        @inputs = nil if @inputs == false
```
Nik08: Small refactoring suggestion, instead of `if @inputs == false` we can use `unless @inputs`. Just the way you have refactored in `if` condition

Vasu1105: This is only in case of input is false cause input will be already nil if its nil but I will add similar checks here too. Thanks!!

Nik08 (approved):
Functionally it looks good to me. Just have some small refactoring suggestions. Thanks

lib/inspec/waiver_file_reader.rb
Vasu1105: That will be taken care here now https://github.com/inspec/inspec/pull/6644/files#diff-840ef8435a8ecac3b53d637518402eec237eaf326ca7e89c80f1c596b7d24148R27

lib/inspec/secrets/yaml.rb
Vasu1105: That check itself is not required. Removing it.

lib/inspec/waiver_file_reader.rb
Nik08: Yeah I see now, you did that to handle both `Inspec::Exceptions::WaiversFileNotReadable` and `Inspec::Exceptions::WaiversFileInvalidFormatting` errors in one place.

clintoncwolfe (approved):

PR #6661, opened 2023-09-20
Add --legacy-export option to inspec export

sathish-progress:
Ability to run export in the legacy way.

## Description

This PR allows a new option `--legacy-export` to allow users to choose between the new export and old export options. Default is the new export.

## Types of changes

- [X] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

### DIFF
--- a/lib/inspec/cli.rb
+++ b/lib/inspec/cli.rb
@@ -94,6 +94,8 @@ class Inspec::InspecCLI < Inspec::BaseCLI
     desc: "For --what=profile, a list of controls to include. Ignore all other tests."
   option :tags, type: :array,
     desc: "For --what=profile, a list of tags to filter controls and include only those. Ignore all other tests."
+  option :"legacy-export", type: :boolean, default: false,
+         desc: "Run with legacy export."
   profile_options
   def export(target, as_json = false)
     Inspec.with_feature("inspec-cli-export") {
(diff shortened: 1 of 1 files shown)


lib/inspec/cli.rb
```diff
@@ -94,6 +94,8 @@ def json(target)
     desc: "For --what=profile, a list of controls to include. Ignore all other tests."
   option :tags, type: :array,
     desc: "For --what=profile, a list of tags to filter controls and include only those. Ignore all other tests."
+  option :"legacy-export", type: :boolean, default: false,
+         desc: "Run with legacy export."
```
clintoncwolfe, line 98: You'll also need a conditional to choose between calling info() and info_from_parse() around line 140 (or you can refoctor that more sensbly)

aaronlippold: Good call. Can we add this to the json function as well.

clintoncwolfe: json is actually just an alias to export, since InSpec 5. https://github.com/inspec/inspec/blob/7c46d14f3b47dd74aea3a0da32c49facb1dbb06f/lib/inspec/cli.rb#L82

aaronlippold:
Roger that.

lib/inspec/cli.rb
sathish-progress: this is done

PR #6673, opened 2023-09-25
Run inspec check using output info_from_parse

sathish-progress:
Running validity checks using the data obtained from `info_from_parse`

## Description

`metadata.params` evaluates profiles and we don't want to evaluate profiles while doing `inspec check`. Hence this change runs the validity checks on the data obtained from `info_from_parse`. This info data is obtained without evaluating the profiles. 
- This change also build checks params 

## Types of changes

## Checklist:


lib/inspec/profile.rb
```diff
@@ -709,7 +709,8 @@ def check # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metri
     def validity_check # rubocop:disable Metrics/AbcSize
       errors = []
       warnings = []
-      info = info_from_parse
+      yml_metadata = Metadata.from_file(@source_reader.target.abs_path(@source_reader.metadata.ref), nil) || {}
```
clintoncwolfe: Is this really needed? Can't you just call self.metadata.params here? Why are we duplicating the logic about how to locate and load a metadata file in the middle of the check logic?

lib/inspec/utils/profile_ast_helpers.rb
```diff
@@ -251,6 +251,7 @@ def on_block(block_node)
               impact: 0.5,
               refs: [],
               tags: {},
+              checks: Set.new
```
clintoncwolfe: This (eventually) gets turned into JSON, so introducing more complex objects than hashes and arrays and simple scalars is problematic. Is this absolutely needed?

sathish-progress: this was done to avoid duplicates .. let me check with array

docs-chef-io/content/inspec/cli.md
```diff
@@ -116,6 +116,10 @@ This subcommand has the following additional options:
 `--no-with-cookstyle`
 : Enable or disable cookstyle checks.
 
+`--legacy-check`
+`--no-legacy-check`
+: For running checks on profiles that needs train connections, run `inspec check --legacy-check`. Default: do not check.
```
clintoncwolfe: "Run check in legacy mode, which examines the profile in a different way. Default: use current parser-based method."

clintoncwolfe (approved):
LGTM - thanks!

PR #6686, opened 2023-09-26
Improve ControlIDCollector and other fields of export data

ahasunos:
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -586,6 +586,11 @@ module Inspec
         # For each control ID
         #  Look for per-control metadata
         # Filter controls by --controls, list of controls to include is available in include_controls_list
+
+        # NOTE: This is a hack to duplicate refs.
+        # TODO: Fix this in the ref collector or the way we traverse the AST
+        @info_from_parse[:controls].each { |control| control[:refs].uniq! }
+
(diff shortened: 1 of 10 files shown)


lib/inspec/profile.rb
```diff
@@ -525,23 +525,29 @@ def info_from_parse
...
-      # PASS 2: Using the control IDs, deterimine the extents - 
-      # line locations - of the coontrol IDs in each file, and 
+      # PASS 2: Using the control IDs, deterimine the extents -
+      # line locations - of the coontrol IDs in each file, and
       # then extract each source code block. Use this to populate the source code
       # locations and 'code' properties.
 
       # @source_reader.tests contains a hash mapping control filenames to control file contents
-      @source_reader.tests.each do |_control_filename, control_file_source|
+      @source_reader.tests.each_with_index do |(control_filename, control_file_source), index|
```
clintoncwolfe: This index is not what you want. It is not the index of the controls in the profile. It is just the number of the file. If There are three files in the profile, this index will range 0..2. 

You will want to pass the control filename as a parameter to Inspec::Profile::AstHelper::ControlIDCollector.new.

ahasunos: I see your point!

PR #6705, opened 2023-09-28
CHEF-6439 Mandatory Profile Signing (Preview)

Nik08:
## Description

Now using signed profiles with Chef InSpec commands can be enabled as a mandatory default behaviour. 

Please read the docs edit for details about this behavior - scary name, but it is just saying that profiles have to be signed in order for InSpec to execute them. This is optional in InSpec 6, and will be enabled by default sometime in the future on a major version.

For now, enable by setting a feature flag `CHEF_PREVIEW_MANDATORY_PROFILE_SIGNING` in the environment. 
To bypass this behaviour, you can either use the `--allow-unsigned-profile` flag or set the `CHEF_ALLOW_UNSIGNED_PROFILE` environment variable.

## Types of changes

## Checklist:


lib/inspec/profile.rb
```diff
@@ -180,6 +181,26 @@ def failed?
...
+      end
+    end
+
+    def signed?
+      # Signed profiles have .iaf extension
+      (@source_reader.target.parent.class == Inspec::IafProvider)
+    end
+
+    def verify_signed_profile
+      InspecPlugins::Sign::Base.profile_verify(target, true)
```
clintoncwolfe: You may need to ensure the plugin has been loaded at this stage. When run from non-CLI -drivenenvironments - such as kitchen-inspec or compliance-phase - InSpec does not load all Command plugins by default. So this class may not be available.

lib/plugins/inspec-sign/lib/inspec-sign/base.rb
```diff
@@ -92,12 +92,16 @@ def self.profile_sign(profile_path, options)
         Inspec::UI.new.exit(:usage_error)
       end
 
-      def self.profile_verify(signed_profile_path)
+      def self.profile_verify(signed_profile_path, runner_call = false)
```
clintoncwolfe: How about "silent = false"? The flag seems to control whether the check outputs messages to the UI.

lib/inspec/profile.rb
Nik08: Will test this out

lib/inspec/base_cli.rb
```diff
@@ -229,6 +231,8 @@ def self.exec_options
         desc: "Include full source code of controls in the CLI report"
       option :enhanced_outcomes, type: :boolean,
         desc: "Show enhanced outcomes in output"
+      option :allow_unsigned, type: :boolean, default: false,
```
Vasu1105: since `profile_options` already contains the `allow_unsigned` option and `profile_options` are part of `exec_options` I think we can remove this one

ahasunos (approved):

PR #6772, opened 2023-10-09
Updated Git actions workflow to label PR with InSpec major versions

Nik08:
## Description

Updated Git actions workflow to label PR based on the base branch. 
The only base branches that are allowed are: `main`, `inspec-4` and `inspec-5`
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -24,13 +24,16 @@ jobs:
           BASE_BRANCH="${{ github.event.pull_request.base.ref }}"
           if [ "$BASE_BRANCH" == "main" ]; then
             echo "::set-output name=base_branch_name::inspec-6"
-          else
+          elif [ "$BASE_BRANCH" == "inspec-4" ] || [ "$BASE_BRANCH" == "inspec-5" ]; then
             echo "::set-output name=base_branch_name::${BASE_BRANCH}"
+          else
+            echo "Unknown base branch for labelling : ${BASE_BRANCH}"
           fi
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -12,6 +13,28 @@ jobs:
...
+        uses: actions/labeler@main
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Fetch pull request base branch
+        id: pr_info
+        run: |
+          BASE_BRANCH="${{ github.event.pull_request.base.ref }}"
+          if [ "$BASE_BRANCH" == "main" ]; then
+            echo "::set-output name=base_branch_name::inspec-6"
```
clintoncwolfe, line 30: This will not age well - we'll have to update this every time we have a new major version. I'm not sure how to get out of that, though. At least adding a comment would be good.

Nik08: Maybe I can read the VERSION file content, and fetch the major version from it and append "inspec-#{MAJOR VERSION}", when the base branch is main.

And for the rest, If the base branch starts with `inspec-` then just directly pick the base branch. This way it gets generic but it takes away our safe checks `[ "$BASE_BRANCH" == "inspec-4" ] || [ "$BASE_BRANCH" == "inspec-5" ]`. 
By this I mean, if by any chance someone creates a dummy branch with `inspec-` then it will create a label for that. It could be `inspec-random-branch-name` or `inspec-100` etc..

Nik08: Will have to check though if reading file content is possible or not. I know matching is possible, will play around with reading content though.

PR #6807, opened 2023-10-19
[W.I.P] Fix: Waived control executes describe block

Vasu1105:
## Description

Evaluates the describe block only if the waived control run is set to true or the waiver is expired. This ensures the describe block of the waived control should not be executed.

Waived control should not be executed and skipped without execution of describe block.

This change helps in eliminating the efforts to use options --filter-waived-controls and --retain-waiver-data.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -46,12 +46,11 @@ module Inspec
       return unless block_given?
 
       begin
-        instance_eval(&block)
-
-        # By applying waivers *after* the instance eval, we assure that
-        # waivers have higher precedence than only_if.
         __apply_waivers
 
+        unless waiver_is_valid?
+          instance_eval(&block)
+        end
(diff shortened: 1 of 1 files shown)


lib/inspec/rule.rb
```diff
@@ -46,12 +46,11 @@ def initialize(id, profile_id, resource_dsl, opts, &block)
       return unless block_given?
 
       begin
-        instance_eval(&block)
-
-        # By applying waivers *after* the instance eval, we assure that
-        # waivers have higher precedence than only_if.
```
Vasu1105, line 52: Not seen any impact on the precedence even if _apply_waivers is called before instance_eval. Calling _apply_waivers here is required to get all the waiver data so that only in the valid condition we will evaluate the describe block, for example, if the run is set to true and if waiver is expired.

PR #6815, opened 2023-10-20
Pin Thor to < 1.3.0 to fix test failures related to aliases

Vasu1105:
## Description

In Thor's recent [release](https://rubygems.org/gems/thor) version 1.3.0 they introduced [changes](https://github.com/rails/thor/releases/tag/v1.3.0) related to aliases here https://github.com/rails/thor/pull/800
that started breaking the inspec test for aliases.  This PR pins the version of Thor to be < 1.3.0 and then we can identify the impact of the Thor changes in InSpec and then remove the pinning.

## Types of changes

## Checklist:


ahasunos (approved):
This is one of the potential fix which fixes the CI test. Thor 1.3.0 released recently breaks our current test.

If we are sure that fixing only the test is not going to affect our other functionality, we can go ahead and fix the test and remove the constraint to use thor below 1.3.0

ahasunos:
This will require a backport to inspec-5 and inspec-4

PR #6816, opened 2023-10-23
CHEF-6437: Implement different version of `inspec export`

ahasunos:
## Description

This PR implements a different version of `inspec export` where we parse a profile manually to reproduce equivalent data as the legacy export.

Building on that, we then take that parser-based implementation and re-implement `inspec check` to similarly read a profile.

## Related Issue

CHEF-6437: Modify inspec export to be parser-based

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -611,6 +611,18 @@ module Inspec
       controls
     end
 
+    def include_group_data?(group_data)
+      unless include_controls_list.empty?
+        # {:id=>"controls/example-tmp.rb", :title=>"/ profile", :controls=>["tmp-1.0"]}
+        # Check if the group should be included based on the controls it contains
+        group_data[:controls].any? do |control_id|
+          include_controls_list.any? { |id| id.match?(control_id) }
+        end
+      else
+        true
+      end
+    end
+
     def update_groups_from(control_filename, src)
       group_data = {
         id: control_filename,
@@ -622,7 +634,7 @@ module Inspec
       group_controls = @info_from_parse[:controls].select { |control| control[:source_location][:ref] == source_location_ref }
       group_data[:controls] = group_controls.map { |control| control[:id] }
 
-      @info_from_parse[:groups].push(group_data)
+      @info_from_parse[:groups].push(group_data) if include_group_data?(group_data)
     end
 
     def cookstyle_linting_check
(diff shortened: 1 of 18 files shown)


lib/inspec/profile.rb
```diff
@@ -514,6 +515,116 @@ def info(res = params.dup) # rubocop:disable Metrics/CyclomaticComplexity, Metri
...
+          input_collector.process(n)
+        }
+
+        # For each control ID
+        #  Look for per-control metadata
+        # Filter controls by --controls, list of controls to include is available in include_controls_list
+
+        # NOTE: This is a hack to duplicate refs.
+        # TODO: Fix this in the ref collector or the way we traverse the AST
+        @info_from_parse[:controls].each { |control| control[:refs].uniq! }
```
sathish-progress, line 589: by any chance these duplicate control keys have different values ? because `uniq!` is going to keep the last occurrence of that key.

ahasunos: I tested the legacy and current behaviour for a profile with tags as below:

```
 tag severity: nil
 tag data: "tag2"
 tag data: "tag3"
```

The output for legacy and latest export is same which is:

```
"tags": {
        "severity": null,
        "data": "tag3"
      }
```

aaronlippold:
This is always been the case as the value of data gets overwritten. 

If the implementation goal is that data will becoming an array of strings that would be a change. 

At the moment the union of these does not happen by default.

It could be argued that this would be a welcome addition logically But it would be a change of behavior

aaronlippold:
It may be also very useful to allow export to spit out the AST as well some people like myself, might find it useful

clintoncwolfe:
> It may be also very useful to allow export to spit out the AST as well some people like myself, might find it useful

Actually, InSpec doesn't need to dump the AST, since there is a tool that already does that. You want the ruby-parse tool, which comes with the `parser` gem. It does exactly that. 

```
ruby-parse test/fixtures/profiles/old-examples/profile/controls/example-tmp.rb
(begin
  (send nil :title
    (str "/ profile"))
  (block
    (send nil :control
      (str "tmp-1.0"))
    (args)
    (begin
    (..... snip .... )
```

clintoncwolfe (approved):
Long journey, but the compatibility is pretty close at this point.

PR #6829, opened 2023-10-25
CHEF-6493: Support `--legacy-export` option in `inspec archive`

ahasunos:
## Description

This PR adds support for `--legacy-export` flag for including the **inspec.json** file as part of the archive using the information from the legacy export procedure i.e. the results of running `inspec export --legacy-export`.

## Related Issue

CHEF-6493: Modify inspec archive to use `--legacy-export` option

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -809,9 +809,11 @@ module Inspec
       # TODO ignore all .files, but add the files to debug output
 
       # Generate temporary inspec.json for archive
-      if opts[:export]
+      export_opt_enabled = opts[:export] || opts[:legacy_export]
+      if export_opt_enabled
+        info_for_profile_summary = opts[:export] ? info_from_parse : info
         Inspec::Utils::JsonProfileSummary.produce_json(
-          info: info, # TODO: conditionalize and call info_from_parse
+          info: info_for_profile_summary,
           write_path: "#{root_path}inspec.json",
           suppress_output: true
         )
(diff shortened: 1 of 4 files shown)


lib/inspec/profile.rb
```diff
@@ -808,9 +808,11 @@ def archive(opts)
       # TODO ignore all .files, but add the files to debug output
 
       # Generate temporary inspec.json for archive
-      if opts[:export]
+      export_opt_enabled = opts[:export] || opts[:legacy_export]
+      if export_opt_enabled
+        info_for_profile_summary = opts[:export] ? info_from_parse : info
```
clintoncwolfe: This is wrong - the conditional should be:
```
opts[:legacy_export] ? info : info_from_parse
```

ahasunos: I believe it is not wrong.

https://github.com/inspec/inspec/blob/532e573671ccc1c57a83f2201932cdee2705f9fb/lib/inspec/profile.rb#L813


The above line is checked only if `export_opt_enabled` is true i.e. either the flag `--export` or `--legacy-export` is passed in the cli and is part of the `opts`.

This means the conditions - 
```
opts[:export] ? info_from_parse : info
```
and 
```
opts[:legacy_export] ? info : info_from_parse
```
are both going to equivalent.

ahasunos: Let me know if you mean something else here, which I maybe missing

ahasunos: Addressed!

PR #6833, opened 2023-10-26
CHEF-7670 Ability to set feature preview flag using ENV

Nik08:
## Description

Ability to set feature preview flag using ENV. This extends the functionality of the existing Inspec feature subsystem if some feature execution needs to be conditionalised.

**Details**:
This subsystem uses `env_preview: true or env_preview: false`, set in the config file `etc/features.yaml` to determine if a feature is previewable or not.

- If it is previewable, meaning `env_preview: true`, then the feature block is only executed if the ENV value `CHEF_PREVIEW_#{FEATURE_NAME}` is set in the environment. 
- If it is not previewable, meaning `env_preview: false`, then the feature block is not executed.

In all the other scenarios, the behaviour is unaltered, i.e. it always executes block.

**Note**
Dev docs for the feature subsystem can be covered separately since a doc on this subsystem does not exist now.
Usage of a feature flag from a user perspective can be updated on a per-feature basis. The first flag will be introduced with PR #6705 

## Types of changes

## Checklist:


etc/features.yaml
```diff
@@ -85,4 +85,6 @@
     inspec-reporter-progress-bar:
       description: Use progress bar streaming reporter
     inspec-reporter-child-status:
-      description: Child status reporter used in inspec parallel reporting.
\ No newline at end of file
+      description: Child status reporter used in inspec parallel reporting.
+    inspec-profile-mandatory-signing:
+      flag: CHEF_PREVIEW_MANDATORY_PROFILE_SIGNING
```
clintoncwolfe: Please try to keep your work on separate PRs.

I'd like you to add a feature-neutral test flag. Say, CHEF_PREVIEW_TEXT_FIXTURE_FLAG. That way, you have a feature you can test with on this PR, and it has nothing to do with the other work going on elsewhere.

Then we block the mandatory signing pr until this one merges (should be fast).

Then, on the mandatory signing PR, you should introduce a new feature named for mandatory profile signing. That way all the impact of mandatory signing stays on the other PR.

clintoncwolfe: I'm wondering about the word "flag". Can we do this instead:
```
inspec-my-new-feature:
  # This will cause us to be sensitive to CHEF_PREVIEW_MY_NEW_FEATURE
  env_preview: true
  ```
  
 Then later you can do something like:
 ```
   if feat.previewable? do
     yield
    end
 ```
 
 and then in Feature.previewable?
 check to see if feat.env_preview is enabled (if false, we're not in preview, so return true)
 if the feature does support env preview, then generate the associated ENV var name and check for it, and return true if it is set.

clintoncwolfe (approved):
Just want a comment example about the transformation between a feature name and an env var, but other wise looks great!

PR #6838, opened 2023-10-31
CHEF-7694 (CHEF-8271) : Fix security_policy resource returns array for non comma separated values

Vasu1105:
## Description

Security Policy which non-comma separated string for example for LockDuration the assigned value is -1 but currently it returns an array ["-1"] which fails the control.  This is a regression introduced in #5629 and this PR fixes that and adds additional data to test.

Example:

```
    describe security_policy do
      its("LockoutDuration") { should be >= "-1" }
    end
  end


Security Policy LockoutDuration is expected to be >= -1
expected: >= "-1"
     got:    ["-1"]
```

Failing Test before fix: https://buildkite.com/chef/inspec-inspec-main-verify-private/builds/377#018b878c-b850-4777-87f2-797bb9a30d7b/1038-1324

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/security_policy.rb
+++ b/lib/inspec/resources/security_policy.rb
@@ -169,9 +169,11 @@ module Inspec::Resources
       # special handling for string values with "
       elsif !(m = /^\"(.*)\"$/.match(val)).nil?
         m[1]
+      # When there is Registry Values value is not seprated with comma we are not spliting the value for backward compatibility
+      elsif !key.include?("\\") && val.match(/,/)
+        val.split(",")
       else
-        # When there is Registry Values we are not spliting the value for backward compatibility
-        key.include?("\\") ? val : val.split(",")
+        val
       end
     end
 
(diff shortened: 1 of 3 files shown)


ahasunos:
As mentioned in the example by you:

```
describe security_policy do
  its("LockoutDuration") { should be >= "-1" }
end
```

Should the value of lockout duration be of type string as mentioned above `"-1"` or a numeric type `-1`?

Vasu1105:
> As mentioned in the example by you:
> 
> ```
> describe security_policy do
>   its("LockoutDuration") { should be >= "-1" }
> end
> ```
> 
> Should the value of lockout duration be of type string as mentioned above `"-1"` or a numeric type `-1`?

It's a string

lib/inspec/resources/security_policy.rb
```diff
@@ -169,9 +169,11 @@ def extract_value(key, val)
       # special handling for string values with "
       elsif !(m = /^\"(.*)\"$/.match(val)).nil?
         m[1]
+      # When there is Registry Values value is not seprated with comma we are not spliting the value for backward compatibility
```
Nik08: Can you explain this comment in more detail with an example? This comment has a few typos as well.

Nik08 (approved):
LGTM

PR #6847, opened 2023-11-07
Remove checks for export

sathish-progress:
The `checks` subfield should not be available part of the `controls` data structure for `inspec export`.

## Description

`check` are basically list of tests per control. This is used to check for availability of tests during `inspec check` where as it is not needed for `export` action. 
In the legacy mode, we [delete the checks ](https://github.com/inspec/inspec/blob/ee490412e811de57023002384089fcb8545dde71/lib/inspec/profile.rb#L462) while running export. With the recent changes, since it is a costly operation to populate these values at first place this change by default does not populate it and provides an option to populate checks when needed. In this case, the only action that needs it are `inspec check`. 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/profile.rb
+++ b/lib/inspec/profile.rb
@@ -818,6 +818,9 @@ module Inspec
         offenses: [],
       }
 
+      # memoize `info_from_parse` with tests
+      info_from_parse(include_tests: true)
+
       entry = lambda { |file, line, column, control, msg|
         {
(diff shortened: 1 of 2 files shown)


clintoncwolfe (approved):
First of many, I'm sure. We'll eventually want info_from_parse to have all sort of conditionals, I imagine.

PR #6873, opened 2023-11-22
CHEF-8609 Reduce version pinning

Vasu1105:
## Description

Removed pinning for a few of the gems from Gemfile 

verify pipeline is green https://buildkite.com/chef/inspec-inspec-main-verify-private/builds/489
adhoc build is green https://buildkite.com/chef/inspec-inspec-main-omnibus-adhoc/builds/490#_
## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,11 +46,6 @@ group :test do
   gem "simplecov"
   gem "simplecov_json_formatter"
   gem "webmock"
-
-  if Gem.ruby_version >= Gem::Version.new("3.0.0")
-    # html-proofer has a dep on io-event, which is ruby-3 only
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -32,20 +32,20 @@ group :omnibus do
...
-  gem "simplecov", "~> 0.21"
+  gem "pry"
+  gem "rake"
+  gem "simplecov"
   gem "simplecov_json_formatter"
-  gem "webmock", "~> 3.0"
+  gem "webmock"
 
   if Gem.ruby_version >= Gem::Version.new("3.0.0")
     # html-proofer has a dep on io-event, which is ruby-3 only
```
clintoncwolfe: I"m willing to even be more drastic and REMOVE the test that relies on html-proofer - its only one test, that verifies the output of the html2 reporter. And for that, we carry this very expensive dependency. I think we should drop the test. We know our HTML is valid.

ahasunos (approved):

PR #6894, opened 2023-11-29
Update to auditd resource to check and warn of unloaded rules

aaronlippold:
- This also helps clarify FT errors when looking for rules that have been written to disk but not loaded into memory

The challenge was that the FT errors being reported were not really clear to the end user that it couldn't find the 'rule' or 'file path of the watch file' that the resource was asking for. This usually happens when a user changes the rule but does not reload them on the sytem or restart the auditd system. 

This doesn't fix the issues with FT communicating better when it can't find what its looking for ( this should be another PR) but does at least make sure the user isn't in the ' I forgot to reload case' which is pretty common.


aaronlippold:
Feel free edit, add or modify this as you see fit.

PR #6919, opened 2023-12-21
CHEF-4191 Waivers backward compatibility support with Chef Client

Nik08:
## Description

Waivers backward compatibility support with Chef Client
Chef Client passes waiver information to InSpec via inputs and hence there needs to be a support to read inputs for waivers.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/rule.rb
+++ b/lib/inspec/rule.rb
@@ -375,19 +375,24 @@ module Inspec
     # only_if mechanism)
     # Double underscore: not intended to be called as part of the DSL
     def __apply_waivers
+      @__waiver_data = nil
       control_id = @__rule_id # TODO: control ID slugging
+
       waiver_files = Inspec::Config.cached.final_options["waiver_file"] if Inspec::Config.cached.respond_to?(:final_options)
+      unless waiver_files.nil? || waiver_files.empty?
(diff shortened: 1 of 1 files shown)


Vasu1105:
All looks good to me, only good to have test coverage for data coming through inputs hash would be +1

lib/inspec/rule.rb
```diff
@@ -375,19 +375,24 @@ def __add_check(describe_or_expect, values, block)
     # only_if mechanism)
     # Double underscore: not intended to be called as part of the DSL
     def __apply_waivers
+      @__waiver_data = nil
       control_id = @__rule_id # TODO: control ID slugging
-      waiver_files = Inspec::Config.cached.final_options["waiver_file"] if Inspec::Config.cached.respond_to?(:final_options)
 
-      waiver_data_by_profile = Inspec::WaiverFileReader.fetch_waivers_by_profile(__profile_id, waiver_files) unless waiver_files.nil?
+      waiver_files = Inspec::Config.cached.final_options["waiver_file"] if Inspec::Config.cached.respond_to?(:final_options)
+      unless waiver_files.nil? || waiver_files.empty?
```
Vasu1105: We can use if... else instead of unless.... else here.

ahasunos (approved):

Vasu1105 (approved):

PR #6923, opened 2024-01-09
CHEF-7265 Telemetry opt-in for CINC users

Nik08:
## Description

By default, telemetry is opted-in by Inspec.
And most users will not be able to disable telemetry. However, CINC users can.

So using the CLI option `--no-enable-telemetry` by CINC users will disable telemetry. 
But if a user is not a CINC user, we need to issue a warning since it is not permissible to disable telemetry for them.

## Types of changes

## Checklist:


Nik08:
Closing this, since this change is part of telemetry and not LDC client integration. New PR for telemetry opt-in https://github.com/inspec/inspec/pull/6966

PR #6947, opened 2024-01-22
CHEF-6412: Fix for virutalization resource undefined method file_read

Vasu1105:
## Description

This PR is to fix the error undefined method file_read

When the detect_container method is called  it checks for the condition  if /proc/1/environ file has an entry for Podman container. For reading the contents of the file file_read method is used, which does not exist throughout the code and due to which the virtualization resource sets the platform values as nil and gives the wrong result.

This PR removes file_read command and uses file method of the inspec backend to read the contents. It also adds the unit test for Podman container detection.

## Types of changes

## Checklist:


ahasunos (approved):
lgtm

PR #6966, opened 2024-02-01
CHEF-7265 Telemetry opt-in for CINC users

Nik08:
## Description

By default, telemetry is opted-in by Inspec.
And most users will not be able to disable telemetry. However, CINC users can.

So using the CLI option --no-enable-telemetry by CINC users will disable telemetry.
But if a user is not a CINC user, we need to issue a warning since it is not permissible to disable telemetry for them.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/utils/telemetry.rb
+++ b/lib/inspec/utils/telemetry.rb
@@ -25,7 +25,6 @@ module Inspec
 
       # Telemetry opt-in/out enabled for other Inspec distros
       return Inspec::Telemetry::Null if Inspec::Dist::EXEC_NAME != "inspec" && telemetry_disabled?
-      # TODO: ask chef-telemetry gem if we should be enabled or not
 
       # Don't perform telemetry action if running under Automate - Automate does LDC tracking for us
       return Inspec::Telemetry::Null if Inspec::Telemetry::RunContextProbe.under_automate?
--- a/lib/inspec/utils/telemetry.rb
+++ b/lib/inspec/utils/telemetry.rb
@@ -8,6 +8,7 @@ module Inspec
   class Telemetry
 
     @@instance = nil
+    @@config = nil
 
     def self.instance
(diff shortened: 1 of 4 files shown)


ahasunos (approved):
looks good, except if we can introduce test for the same; that would be great.

Nik08:
> looks good, except if we can introduce test for the same; that would be great.

It just logs in the case of InSpec.Only a different behaviour for Non-inspec distros. Let me see what I can do. Thanks!

Vasu1105 (approved):

lib/inspec/utils/telemetry.rb
```diff
@@ -17,6 +18,15 @@ def self.determine_backend_class
...
       return Inspec::Telemetry::Null unless license&.license_type == "free"
 
+      if Inspec::Dist::EXEC_NAME == "inspec" && telemetry_disabled?
+        # Issue a warning if an InSpec user is explicitly trying to opt out of telemetry using cli option
+        Inspec::Log.warn "Telemetry opt-out is not permissible."
+      end
+
+      # Telemetry opt-in/out enabled for other Inspec distros
+      return Inspec::Telemetry::Null if Inspec::Dist::EXEC_NAME != "inspec" && telemetry_disabled?
+      # TODO: ask chef-telemetry gem if we should be enabled or not
```
Vasu1105: Not understtood what is this TODO is for.

lib/inspec/utils/telemetry.rb
```diff
@@ -34,15 +44,25 @@ def self.license
     # These class methods make it convenient to call from anywhere within the InSpec codebase.
     ######
```
Vasu1105, line 45: May be we can remove this

lib/inspec/utils/telemetry.rb
Nik08: Removed

PR #6993, opened 2024-03-26
CHEF-3833 Added delay for InSpec parallel status reporter

Nik08:
## Description

Added delay for InSpec parallel status reporter for clear reporting.
Additionally, the output of chef licensing is redirected to STDERR in between parallel runs.

**Why this approach?**
- The recurring STDOUT from chef-licensing was creating a blinking effect for `inspec parallel` status reporter.
- **Con** of this approach:  If the license is expired by the end of the 50th invocation of a long-running list in `inspec parallel` then the output will be redirected to STDERR.

**Effect of silencing chef-licensing output in between parallel runs**
Chef licensing output is only printed after the `inspec parallel` command invocation and not between parallel runs.

parallel-licensing

## Types of changes

## Checklist:

### DIFF
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
@@ -22,6 +22,7 @@ module InspecPlugins
 
       def run
         initiate_background_run if run_in_background # running a process as daemon changes parent process pid
+        original_stdout_stream = ChefLicensing::Config.output
         until invocations.empty? && @child_tracker.empty?
           # Changing output to STDERR to avoid the output interruption between runs
           ChefLicensing::Config.output = STDERR
@@ -37,6 +38,8 @@ module InspecPlugins
           cleanup_child_processes
           sleep 0.1
         end
+        # Reset output to the original STDOUT stream as a safe measure.
+        ChefLicensing::Config.output = original_stdout_stream
 
         # Requires renaming operations on windows only
         # Do Rename and delete operations after all child processes have exited successfully
--- a/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
+++ b/lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
@@ -23,9 +23,9 @@ module InspecPlugins
       def run
         initiate_background_run if run_in_background # running a process as daemon changes parent process pid
         until invocations.empty? && @child_tracker.empty?
+          # Changing output to STDERR to avoid the output interruption between runs
+          ChefLicensing::Config.output = STDERR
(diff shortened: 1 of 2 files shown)


lib/plugins/inspec-parallel/lib/inspec-parallel/runner.rb
```diff
@@ -24,6 +24,8 @@ def run
         initiate_background_run if run_in_background # running a process as daemon changes parent process pid
         until invocations.empty? && @child_tracker.empty?
           while should_start_more_jobs?
+            # Changing output to STDERR to avoid the output interruption between runs
+            ChefLicensing::Config.output = STDERR
```
ahasunos: Could we reset back to it's original output stream after this `while` logic is completed?

Ignore if such a case doesn't arise where there's execution of code post this `while` block.

ahasunos: And it could be set before the `while` loop or the `until` loop starts. Thoughts?
Something like
```
original_stdout_stream = ChefLicensing::Config.output
ChefLicensing::Config.output = STDERR

while do
# something
end

ChefLicensing::Config.output = original_stdout_stream

```

Nik08: > Could we reset back to it's original output stream after this `while` logic is completed?
> 
> Ignore if such a case doesn't arise where there's execution of code post this `while` block.

It seems safe to do so. In case anyone runs `inspec parallel` parallelly using a script or something. However, that would be a bad use of `inspec parallel` command. (Running parallel command parallelly)
Otherwise, for a single invocation, I feel we do not need to reset as such. We can discuss this.

Nik08: > And it could be set before the `while` loop or the `until` loop starts. Thoughts? Something like
> 
> ```
> original_stdout_stream = ChefLicensing::Config.output
> ChefLicensing::Config.output = STDERR
> 
> while do
> # something
> end
> 
> ChefLicensing::Config.output = original_stdout_stream
> ```

Yes, this makes sense. Initially, I intended to have a check that this should be set only the first time any process is forked or spawned. But later removed that check. 
Makes more sense to put it before the while condition.

Nik08: > Could we reset back to it's original output stream after this `while` logic is completed?
> 
> Ignore if such a case doesn't arise where there's execution of code post this `while` block.

Also I would not want to reset it after every while loop. I would rather do it after until, if at all.

Vasu1105 (approved):
LGTM

ahasunos (approved):

Issue #6064, opened 2022-05-16
inspec git and url fetcher require the $pwd to be a git repo - with inspec exec or inspec supermarket


aaronlippold:
In my plain-old homedir ...

```shell
 ~ inspec supermarket exec dev-sec/mysql-baseline -t mock://
Traceback (most recent call last):
	36: from /Users/alippold/.rvm/gems/ruby-2.7.2/bin/ruby_executable_hooks:22:in `<main>'
	35: from /Users/alippold/.rvm/gems/ruby-2.7.2/bin/ruby_executable_hooks:22:in `eval'
	34: from /Users/alippold/.rvm/gems/ruby-2.7.2/bin/inspec:23:in `<main>'
	33: from /Users/alippold/.rvm/gems/ruby-2.7.2/bin/inspec:23:in `load'
	32: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/inspec-bin-5.14.0/bin/inspec:11:in `<top (required)>'
	31: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/inspec-core-5.14.0/lib/inspec/base_cli.rb:35:in `start'
	30: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor/base.rb:485:in `start'
	29: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
	28: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
	27: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor/command.rb:27:in `run'
	26: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor.rb:243:in `block in subcommand'
	25: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor/invocation.rb:116:in `invoke'
	24: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
	23: from /Users/alippold/.rvm/gems/ruby-2.7.2/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
[...]

PR #7023, opened 2024-04-17
CHEF-6000: Fix URL fetcher failure to work when current directory is not a Git repository

ahasunos:
## Description

This PR fixes an issue where the URL fetcher fails to execute properly when the current directory is not a Git repository. 

This behaviour was observed to result in failures when attempting to execute profiles using url fetcher in directories outside the scope of a Git.

## Related Issue

CHEF-6000: URL fetcher fails if current directory is not a .git directory
#6064 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/fetcher/url.rb
+++ b/lib/inspec/fetcher/url.rb
@@ -2,6 +2,7 @@ require "uri" unless defined?(URI)
 require "openssl" unless defined?(OpenSSL)
 require "tempfile" unless defined?(Tempfile)
 require "open-uri" unless defined?(OpenURI)
+require "git" unless defined?(Git)
 
 module Inspec::Fetcher
   class Url < Inspec.fetcher(1)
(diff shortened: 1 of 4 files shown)


Nik08:
Seems like a good usage of the `git` gem. 
Just an idea, should we also revamp the git fetcher logic to start using this gem instead of doing a shellout?

ahasunos:
> Seems like a good usage of the `git` gem. Just an idea, should we also revamp the git fetcher logic to start using this gem instead of doing a shellout?

Yes, I agree on that.

lib/inspec/fetcher/url.rb
```diff
@@ -133,33 +134,28 @@ def to_s
...
+          # output: \"ssh: connect to host github.com port 22: Undefined error: 0\\r\\nfatal: Could not read from remote repository.\\n\\n
+          # Please make sure you have the correct access rights\\nand the repository exists.\\n\""
+          # We aim to provide a more readable and clear error message to the user.
+
+          # Extract the output message from the error message
+          output_message = e.message[/output:\s+"(.+?)"/, 1]
+
+          # Replace literal "\r" and "\n" with actual newline characters and strip whitespace
+          output_message = output_message.gsub(/\\r|\\n/, "\n").strip
+          raise Inspec::FetcherFailure, "Profile git dependency failed with default reference - #{remote_url} - error running 'git remote show #{remote_url}': #{output_message}"
```
Vasu1105: which line is running git remote show command am I missing something here?

ahasunos: this line `repo = ::Git.ls_remote(remote_url)` is equivalent of running `git remote show`

Vasu1105: This seems confusing we are not running any git remote show command here. We should change the error message to reflect why it failied.

Vasu1105: git remote show and git ls-remote <url> is two different commands altogether show if we look at the error message anybody will think that the error while running git remote show command.

Vasu1105: I have verified that Git library requires git to be installed on the system. So as per my suggestion we should not add this dependency and use our build in shell to run the git ls-remote command which Git ruby library internals are doing.

ahasunos: It's sad that the `git` gem, though it offers great error handling and data manipulation features, essentially wraps the git tool. Considering our circumstances, opting for a InSpec's shellout command might be the straightforward path, just as you suggested.

On a system without git it gives the following error
```
Error: git '-c' 'core.quotePath=true' '-c' 'color.ui=false' 'version'  2>&1
status: pid 83 exit 127
output: "sh: 1: git: not found\n"
```

Vasu1105 (approved):
LGTM

PR #7045, opened 2024-05-23
CHEF-10392 load default telemetry url conditionally

sathish-progress:
Load the default telemetry URL 

## Description

- Loads valid production URL as default URL.
- Allows loading a custom acceptance URL in case the Licensing Server is pointed towards Acceptance.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/utils/telemetry/http.rb
+++ b/lib/inspec/utils/telemetry/http.rb
@@ -7,7 +7,7 @@ module Inspec
     class HTTP < Base
       TELEMETRY_JOBS_PATH = "v1/job"
       TELEMETRY_URL = if ChefLicensing::Config.license_server_url&.match?("acceptance")
-                        ENV["TELEMETRY_URL"]
+                        ENV["CHEF_TELEMETRY_URL"]
                       else
                         "https://services.chef.io/telemetry/"
                       end
--- a/lib/inspec/utils/telemetry/http.rb
+++ b/lib/inspec/utils/telemetry/http.rb
@@ -1,11 +1,16 @@
 # frozen_string_literal: true
 require_relative "base"
 require "faraday" unless defined?(Faraday)
+require "inspec/utils/licensing_config"
 module Inspec
   class Telemetry
     class HTTP < Base
       TELEMETRY_JOBS_PATH = "v1/job"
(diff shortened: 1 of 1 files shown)


Nik08 (approved):
LGTM. But I feel we should probably hold merging this till our testing with product env is done. Thanks!

lib/inspec/utils/telemetry/http.rb
```diff
@@ -1,11 +1,16 @@
...
 require_relative "base"
 require "faraday" unless defined?(Faraday)
+require "inspec/utils/licensing_config"
 module Inspec
   class Telemetry
     class HTTP < Base
       TELEMETRY_JOBS_PATH = "v1/job"
-      TELEMETRY_URL = ENV["TELEMETRY_URL"] # TODO Set to production telemetry URL before release
+      TELEMETRY_URL = if ChefLicensing::Config.license_server_url&.match?("acceptance")
+                        ENV["TELEMETRY_URL"]
```
clintoncwolfe: Could we change this ENV var name to have a CHEF_ prefix, or would that break the world?

sathish-progress: done

lib/inspec/utils/telemetry/http.rb
```diff
@@ -1,11 +1,16 @@
...
 module Inspec
   class Telemetry
     class HTTP < Base
       TELEMETRY_JOBS_PATH = "v1/job"
-      TELEMETRY_URL = ENV["TELEMETRY_URL"] # TODO Set to production telemetry URL before release
+      TELEMETRY_URL = if ChefLicensing::Config.license_server_url&.match?("acceptance")
+                        ENV["CHEF_TELEMETRY_URL"]
+                      else
+                        "https://services.chef.io/telemetry/"
+                      end
```
ahasunos, line 13: Maybe we can refactor this to: `ENV["CHEF_TELEMETRY_URL"] || "https://services.chef.io/telemetry/"` ?

Nik08: No, we do not want the user to override the production URL with any other dummy URL. So this OR condition is not enough.

PR #7069, opened 2024-06-25
CHEF-12024 Adds C Shell support to sybase_session resource

Vasu1105:
## Description

This PR fixes the `sybase_session` InSpec resource, which when the user's default shell was set to `/bin/csh`, would lead to the error:
 
```
isql exited with code 1 and stderr 'LANG=en_US.UTF-8: Command not found.\n', stdout ''","exception":"RuntimeError"
```

Changes include identifying the current shell and using the right syntax for setting env variables as per current shell.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/sybase_session.rb
+++ b/lib/inspec/resources/sybase_session.rb
@@ -44,10 +44,18 @@ module Inspec::Resources
       # try to get a temp path
       sql_file_path = upload_sql_file(sql)
 
-      # isql reuires that we have a matching locale set, but does not support C.UTF-8. en_US.UTF-8 is the least evil.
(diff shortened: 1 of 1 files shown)


lib/inspec/resources/sybase_session.rb
```diff
@@ -44,10 +44,18 @@ def query(sql)
       # try to get a temp path
       sql_file_path = upload_sql_file(sql)
 
+      current_shell = inspec.command("echo $SHELL")
```
clintoncwolfe, line 48: Just for posterity, I'll lament that we don't have a better way, or a better place, to detect this. But I think we can let this go for now.

clintoncwolfe (approved):

lib/inspec/resources/sybase_session.rb
Vasu1105: Yes. I will add TODO so that we can track this and improve.

Nik08 (approved):

Vasu1105:
Just waiting on omnibus release pipeline to get fixed.

PR #7091, opened 2024-07-18
InSpec 7 Major Release

Vasu1105:
## Description
Development branch for InSpec 7

### Long Term Support

InSpec 7 will be a full LTS release, with an anticipated 3-year cycle between breaking changes. In between LTS releases, InSpec may experience Innovation releases, in which we introduce exciting new features for customers to try, with a shorter support horizon. 

### Gem-Based Resource Packs

Support for packaging InSpec Resources as Ruby Gems (rather than tarball'd profile directories) provides more distribution options, more usage information, better metadata, and improved installation experiences.  Under the hood, it's a new Gem Fetcher and a new Resource Pack Plugin Type.

### Modularizing the Core Resources

Many groups of core resources - the Docker group, for example - are difficult to justify being in the core, and limit the flexibility of InSpec. Some customers want all the pieces; some customers want just a few.  Some pieces are iterated on quickly, while others are iterated on slowly. We're breaking out groups of core resources into separate resource packs over the life of InSpec 7. Some core resources are planned be exported at the time of initial release, while others are planned be exported over the course of InSpec 7's life. Don't worry - we plan to have a graceful mechanism to automatically detect when a newly exported resource is referenced, and then dynamically install the needed gem resource pack. Provisions have been made for airgapped deployments.

### Cloud Resource Packs as Gems

Since we can now have Resource Packs as Gems, we're going to convert the Cloud Resource Packs into gems to enjoy the benefits.

### Habitat-Based Packaging

We are excited (if not a bit relieved) to announce the migration to Habitat-based packaging from Omnibus, which should allow faster and more reliable releases.


Vasu1105:
Note: This change https://github.com/inspec/inspec/pull/7092 needs to be reverted afterwords.

Vasu1105:
I am closing this PR right now as this will be required only when we are merging the changes in the main branch from the inspec-7

PR #7137, opened 2024-08-16
BACKPORT: Oracle db session resource fixes

Nik08:
## Description

Backports #7136 
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -57,7 +57,7 @@ module Inspec::Resources
       inspec_cmd = inspec.command(command)
       out = inspec_cmd.stdout + "\n" + inspec_cmd.stderr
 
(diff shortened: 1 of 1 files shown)


lib/inspec/resources/oracledb_session.rb
```diff
@@ -96,7 +96,8 @@ def command_builder(format_options, query)
       if @db_role.nil? || @su_user.nil?
         verified_query = verify_query(query)
       else
-        escaped_query = escape_query(query)
+        escaped_query = query.gsub(/\\\\/, "\\").gsub(/"/, '\\"')
```
sathish-progress: could this be moved into the same `escape_query` method?

Nik08: Yes, can keep the old refactoring change on this. Will do.

sathish-progress: Hey this is still not addressed

sathish-progress (approved):

PR #7154, opened 2024-08-27
Handled postgres_session resource password authentication error when fatal

gokulks97:
## Description

This PR addresses an issue with postgres_session resource where execution of query method (postgres_session:query) is not throwing error as
```
     PostgreSQL connection error: 
     psql: FATAL:  password authentication failed for user "postgres"
```
but gets passed with
```
query.output = ""
```
as "fatal" check is not available.

## Types of changes

## Checklist:


Nik08 (approved):
Thanks for the changes!

ahasunos (approved):

PR #7198, opened 2024-10-28
Improved regex to load readme from inspec profiles

Nik08:
## Description

Improved regex to load readme from inspec profiles. It was unable to load readme files without `.md` extensions in the file.
## Types of changes

## Checklist:

### DIFF
--- a/lib/source_readers/inspec.rb
+++ b/lib/source_readers/inspec.rb
@@ -66,7 +66,7 @@ module SourceReaders
     end
 
     def load_readme
-      load_all(/README.md/)
+      load_all(/README(\.md)?$/)
     end
   end
 end


ahasunos (approved):
Looks good to me.

Tried out the below example to test the difference.

```ruby
# Define some test strings
string1 = "path/to/README.md"
string2 = "path/to/README"
string3 = "README.md/extra"

# First regex: /README.md/
puts "/README.md/ matches:"
puts string1.match?(/README.md/)  # => true (contains README.md)
puts string2.match?(/README.md/)  # => false (no .md at the end)
puts string3.match?(/README.md/)  # => true (README.md appears in the string)

# Second regex: /README(\.md)?$/
puts "\n/README(\\.md)?$/ matches:"
puts string1.match?(/README(\.md)?$/)  # => true (ends with README.md)
puts string2.match?(/README(\.md)?$/)  # => true (ends with README)
puts string3.match?(/README(\.md)?$/)  # => false (README.md is not at the end)
```

Output:
```bash
/README.md/ matches:
true
false
true

/README(\.md)?$/ matches:
true
true
false
```

PR #7224, opened 2024-12-04
CHEF-16631 Implement fallback logic when gem caching is manually deleted

Nik08:
## Description

The following changes are covered:
- A fallback logic when the gem cache is manually deleted or is deleted due to some unknown misconfiguration. Since this was breaking `inspec exec` a fallback logic to clear gem's specification and enable gem fetcher logic to run step by step again is implemented.
- The unimplemented Gemprovider logic is extended now. Scenario when the gem is not locally referenced and it uses gem provider then it loads gem from the path and returns it.
- Null values error handling

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/cached_fetcher.rb
+++ b/lib/inspec/cached_fetcher.rb
@@ -57,12 +57,12 @@ module Inspec
           [cache_value, false]
         else
           Inspec::Log.debug "Dependency does not exist in the cache for target #{target}"
-          cache_key_name = fetcher.cache_key
+          cache_key_name = cache_key
           if cache_key_name.start_with?("gem:")
             # When cache for gem - meaning gemspec exists but gem does not exists then clearing up gemspec is required
             # This logic enables the gem fetcher logic to work step by step again
             Inspec::Log.debug "Clearing cached gemspec to fix dependency issue and enable fresh download."
-            FileUtils.rm_rf(cache.gemspec_path_for(fetcher.cache_key))
+            FileUtils.rm_rf(cache.gemspec_path_for(cache_key))
           end
         end
       else
--- a/lib/inspec/cached_fetcher.rb
+++ b/lib/inspec/cached_fetcher.rb
@@ -57,6 +57,13 @@ module Inspec
           [cache_value, false]
         else
           Inspec::Log.debug "Dependency does not exist in the cache for target #{target}"
+          cache_key_name = fetcher.cache_key
+          if cache_key_name.start_with?("gem:")
+            # When cache for gem - meaning gemspec exists but gem does not exists then clearing up gemspec is required
+            # This logic enables the gem fetcher logic to work step by step again
+            Inspec::Log.debug "Clearing cached gemspec to fix dependency issue and enable fresh download."
+            FileUtils.rm_rf(cache.gemspec_path_for(fetcher.cache_key))
+          end
         end
       else
         begin
--- a/lib/inspec/cached_fetcher.rb
+++ b/lib/inspec/cached_fetcher.rb
@@ -52,7 +52,12 @@ module Inspec
         fetch
       elsif cache.exists?(cache_key) && !cache.locked?(cache_key)
         Inspec::Log.debug "Using cached dependency for #{target}"
-        [cache.prefered_entry_for(cache_key), false]
+        cache_value = cache.prefered_entry_for(cache_key)
+        if cache_value
+          [cache_value, false]
+        else
+          Inspec::Log.debug "Dependency does not exist in the cache for target #{target}"
+        end
       else
         begin
           Inspec::Log.debug "Dependency does not exist in the cache #{target}"
--- a/lib/inspec/file_provider.rb
+++ b/lib/inspec/file_provider.rb
@@ -111,16 +111,24 @@ module Inspec
       # Determine a path to the gem installation directory
       gem_name = gem_info[:gem]
       bootstrap_file = gem_info[:path]
+      gem_version = gem_info[:version]
       if bootstrap_file
         # We were given an explicit path - go two directories up
         parts = Pathname(gem_info[:path]).each_filename.to_a
         @gem_lib_root = File.join(parts.slice(0, parts.length - 2))
+        @files = Dir[File.join(Shellwords.shellescape(@gem_lib_root), "**", "*")]
       else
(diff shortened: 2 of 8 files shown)


lib/inspec/file_provider.rb
```diff
@@ -7,6 +7,7 @@
 module Inspec
   class FileProvider
     def self.for_path(path)
+      raise "Profile or dependency path not resolved." if path.nil?
```
sathish-progress, line 10: why do we have to raise this error ?

lib/inspec/file_provider.rb
```diff
@@ -107,16 +108,24 @@ def initialize(gem_info)
...
         # Look up where the gem is installed, respecting version
-        raise NotImplementedError, "TODO: Gem FileProvider when gem #{gem_name} is not a local reference"
+        loader = Inspec::Plugin::V2::Loader.new
+        gem_path = loader.find_gem_directory(gem_name, gem_version)
+        if File.exist?(gem_path)
+          gem = DirProvider.new(gem_path)
+          @files = gem.files
+          gem
+        else
+          raise "Dependency does not exist #{gem_name}"
```
sathish-progress: We should not raise error here.. instead return nil or a custom exception which should be captured by GemFetcher to try installing the gem ( can be handled in a different PR )

lib/inspec/file_provider.rb
Nik08: Because if the path is nil, it throws a stack trace with a ruby error. So this is just a nil check handling. Something must be wrong because of which the dependency is not resolved and hence the path is nil.

lib/inspec/file_provider.rb
Nik08: I feel the job of gem provider is to just return the gem if it exists else just return the error. The logic to retry the installation and everything is handled in cache_fetcher, which anyway calls gem fetcher to do that. So GemProvider library is just doing a dummy job. That is what I think looking into different providers logic.

lib/inspec/plugin/v2/loader.rb
```diff
@@ -170,6 +170,23 @@ def self.find_gem_directory(gem_name, version = nil)
...
+      matching_gem_versions = list_managed_gems.filter { |g| g.name == gem_name }.sort(&:version)
+      selected_gemspec = nil
+      if version
+        selected_gemspec = matching_gem_versions.find { |g| g.version == version }
+      else
+        # Use latest
+        selected_gemspec = matching_gem_versions.last
+      end
+
+      selected_gemspec && selected_gemspec.loaded_from
```
sathish-progress:
```suggestion
      def self.find_gemspec_directory(gem_name, version = nil)
  selected_gemspec = list_managed_gems
                       .select { |g| g.name == gem_name }
                       .sort_by(&:version)
                       .yield_self { |gems| version ? gems.find { |g| g.version == version } : gems.last }
  selected_gemspec&.loaded_from
end

```

lib/inspec/file_provider.rb
sathish-progress: this method has a bulk of if-else which eventually has a else block which should actually be run when path is nil. What error are you facing ? may be it is not even related to this method..

lib/inspec/file_provider.rb
sathish-progress: I disagree with the scope of Gem Provider. But yeah with the current plugin system we don't follow SRP principle so okay with the scope of current gem provider, but all am asking is instead of raising a StandardError, could you raise a custom error object so we can capture it at the calling function and fallback to installation.

lib/inspec/plugin/v2/loader.rb
```diff
@@ -170,6 +170,23 @@ def self.find_gem_directory(gem_name, version = nil)
       selected_gemspec && selected_gemspec.full_gem_path
     end
 
+    def self.find_gemspec_directory(gem_name, version = nil)
+      matching_gem_versions = list_managed_gems.filter { |g| g.name == gem_name }.sort(&:version)
```
sathish-progress: okay so this clearly doesn't capture all installed gems. `list_managed_gems` only lists gems for which the gemspec is available in `~/.inspec/gems/specifications`. But the resource packs itself can be installed in the environment and need not be in the plugin dir.

lib/inspec/file_provider.rb
Nik08: Okay, let me check. If you already know, can you tell me which custom error class that is? Or are you suggesting to create a new one?

lib/inspec/plugin/v2/loader.rb
Nik08: Oh is it? I did not see that part of logic where the resource packs are installed in the environment. Is that via manual intervention?

lib/inspec/file_provider.rb
Nik08: @sathish-progress Made the changes for this.

lib/inspec/cached_fetcher.rb
```diff
@@ -52,7 +52,19 @@ def fetch
         fetch
       elsif cache.exists?(cache_key) && !cache.locked?(cache_key)
         Inspec::Log.debug "Using cached dependency for #{target}"
-        [cache.prefered_entry_for(cache_key), false]
+        cache_value = cache.prefered_entry_for(cache_key)
+        if cache_value
+          [cache_value, false]
+        else
+          Inspec::Log.debug "Dependency does not exist in the cache for target #{target}"
+          cache_key_name = fetcher.cache_key
```
ahasunos: I think the value for `fetcher.cache_key` and `cache_key` is going to be the same. So, any reason behind using `fetcher.cache_key`? 



```
(byebug) cache_key
"gem:inspec-mongodb-resources:"
(byebug) fetcher.cache_key
"gem:inspec-mongodb-resources:"
```

Although , the value for `cache_key` maybe different based on the `k`'s value in the method `cache_key`
https://github.com/inspec/inspec/blob/b896dfc3405006d86e32f6e1af02d5c6c18670a9/lib/inspec/cached_fetcher.rb#L29

sathish-progress (approved):

lib/inspec/cached_fetcher.rb
Nik08: Nice catch, will change this.

ahasunos (approved):

PR #7239, opened 2024-12-17
CHEF-14246 Implement version handling of resource pack gem via gem fetcher

Nik08:
Implement version handling of resource pack gem via gem fetcher

## Description

Implemented version handling of resource pack gem by introducing working usage of `version` key in metadata file under dependency section.
```
depends:
  - name: inspec-ibmdb2-resources
    gem: inspec-ibmdb2-resources
    version: 0.1.1
```
When not used the `version` key in the metadata file, it picks the last activated version of the resource pack gem.
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -217,7 +217,7 @@ module Inspec::Plugin::V2
         if opts.key?(:version) && plugin_version_installed?(plugin_name, opts[:version])
           raise InstallError, "#{plugin_name} version #{opts[:version]} is already installed."
         else
-          # Do not redirect to plugin update when using resource packs plugin
+          # Do not redirect to plugin update when using gem based plugin
           unless is_a_gem_based_plugin?(plugin_name)
             raise InstallError, "#{plugin_name} is already installed. Use 'inspec plugin update' to change version."
           end
@@ -346,14 +346,14 @@ module Inspec::Plugin::V2
       # Activate all current plugins before trying to activate the new one
       loader.list_managed_gems.each do |spec|
         # Skip in case of update mode
-        # Skip in case using a resource pack plugin
+        # Skip in case using a gem based plugin
         next if spec.name == new_plugin_dependency.name && (update_mode || is_a_gem_based_plugin?(new_plugin_dependency.name))
         spec.activate
       end
 
       # Make sure we remove any previously loaded gem on update
       # Make sure we remove any previously loaded gem when trying to use resource pack gem
-      # Resource pack gems when updated need to deactivate older version of gem
+      # Gem based plugin when updated need to deactivate older version of gem
       Gem.loaded_specs.delete(new_plugin_dependency.name) if update_mode || is_a_gem_based_plugin?(new_plugin_dependency.name)
 
       # Test activating the solution. This makes sure we do not try to load two different versions
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -218,7 +218,7 @@ module Inspec::Plugin::V2
           raise InstallError, "#{plugin_name} version #{opts[:version]} is already installed."
         else
           # Do not redirect to plugin update when using gem based plugin
-          unless is_a_gem_based_plugin?(plugin_name)
+          unless gem_based_plugin?(opts)
             raise InstallError, "#{plugin_name} is already installed. Use 'inspec plugin update' to change version."
           end
         end
@@ -314,7 +314,7 @@ module Inspec::Plugin::V2
                 end
 
       begin
-        install_gem_to_plugins_dir(plugin_dependency, [sources], opts[:update_mode])
+        install_gem_to_plugins_dir(plugin_dependency, [sources], opts[:update_mode], opts)
       rescue Gem::RemoteFetcher::FetchError => gem_ex
         # TODO: Give a hint if the host was not resolvable or a 404 occured
         ex = Inspec::Plugin::V2::InstallError.new(gem_ex.message)
@@ -325,7 +325,7 @@ module Inspec::Plugin::V2
 
     def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics/AbcSize
       extra_request_sets = [],
-      update_mode = false)
+      update_mode = false, opts = {})
 
       # Get a list of all the gems available to us.
       gem_to_force_update = update_mode ? new_plugin_dependency.name : nil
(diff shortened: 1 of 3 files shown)


lib/inspec/plugin/v2/installer.rb
```diff
@@ -340,15 +343,21 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
         raise ex
       end
 
+      it_is_resource_pack_gem = new_plugin_dependency.name =~ /^(inspec)-/ && new_plugin_dependency.name =~ /-resources/
```
ahasunos: Cosmetic change - Do you think this could be moved to a method like 

```
def is_resource_pack_gem?
  new_plugin_dependency.name =~ /^(inspec)-/ && new_plugin_dependency.name =~ /-resources/
end
```

and then invoke the method wherever `it_is_resource_pack_gem` variable is used.

lib/inspec/plugin/v2/installer.rb
```diff
@@ -217,7 +217,10 @@ def validate_installation_opts(plugin_name, opts)
         if opts.key?(:version) && plugin_version_installed?(plugin_name, opts[:version])
           raise InstallError, "#{plugin_name} version #{opts[:version]} is already installed."
         else
-          raise InstallError, "#{plugin_name} is already installed. Use 'inspec plugin update' to change version."
+          # Do not redirect to plugin update when using resource packs plugin
+          unless plugin_name =~ /^(inspec)-/ && plugin_name =~ /-resources/
```
ahasunos: same method `is_resource_pack_gem?` could be invoked here as well.

ahasunos (approved):

lib/inspec/plugin/v2/installer.rb
```diff
@@ -342,13 +345,17 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
 
       # Activate all current plugins before trying to activate the new one
       loader.list_managed_gems.each do |spec|
-        next if spec.name == new_plugin_dependency.name && update_mode
+        # Skip in case of update mode
+        # Skip in case using a resource pack plugin
+        next if spec.name == new_plugin_dependency.name && (update_mode || is_resource_pack_gem?(new_plugin_dependency.name))
```
sathish-progress: can we rather check for spec name and version here?

lib/inspec/plugin/v2/installer.rb
```diff
@@ -342,13 +345,17 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
...
 
         spec.activate
       end
 
       # Make sure we remove any previously loaded gem on update
-      Gem.loaded_specs.delete(new_plugin_dependency.name) if update_mode
+      # Make sure we remove any previously loaded gem when trying to use resource pack gem
+      # Resource pack gems when updated need to deactivate older version of gem
+      Gem.loaded_specs.delete(new_plugin_dependency.name) if update_mode || is_resource_pack_gem?(new_plugin_dependency.name)
```
sathish-progress: again here we can check in loaded specs if there are multiple versions then deactivate the ones we don't want rather than doing a regex based check

lib/inspec/plugin/v2/loader.rb
```diff
@@ -158,7 +158,7 @@ def list_managed_gems
     end
 
     def self.find_gem_directory(gem_name, version = nil)
-      matching_gem_versions = list_managed_gems.filter { |g| g.name == gem_name }.sort(&:version)
+      matching_gem_versions = list_managed_gems.filter { |g| g.name == gem_name }.sort_by(&:version)
```
sathish-progress, line 161: this is the ideal logic am looking for activation as well.. is it possible to bring similar logic as well while deleting different version specs?

lib/inspec/plugin/v2/installer.rb
Nik08: Regex check is only for checking if a dependency is a resource pack dependency or not. And this is only called while someone uses a different version which is not already installed in the system, else this is not even called here.

lib/inspec/plugin/v2/loader.rb
Nik08: I think having this version check makes sense if it is absolutely necessary to delete a specific version. In this scenario the only need is to remove old cached activated resource pack gems.

lib/inspec/plugin/v2/installer.rb
Nik08: So the new plugin dependency version is going to be an updated version and it will never match the version of old cached dependencies. 
We can clear all old cache safely because in this scenario a user has made a decision to either upgrade and downgrade and the old cache is not required anyway. There is no reason to have a version check. Do you have any specific reasoning behind having a version check?

sathish-progress: okay it is not the case that the user decided to upgrade or downgrade. There is a scenario where the user simply wants to run with different versions.

lib/inspec/plugin/v2/loader.rb
sathish-progress: we do need this to activate a particular version too

lib/inspec/plugin/v2/installer.rb
sathish-progress: ehh this will be called anytime a `install_gem_to_plugins_dir` is run

lib/inspec/plugin/v2/installer.rb
```diff
@@ -217,7 +217,10 @@ def validate_installation_opts(plugin_name, opts)
         if opts.key?(:version) && plugin_version_installed?(plugin_name, opts[:version])
           raise InstallError, "#{plugin_name} version #{opts[:version]} is already installed."
         else
-          raise InstallError, "#{plugin_name} is already installed. Use 'inspec plugin update' to change version."
+          # Do not redirect to plugin update when using gem based plugin
+          unless is_a_gem_based_plugin?(plugin_name)
```
sathish-progress: as discussed offline, this will be renamed to `gem_based_plugin?` and its implementation will accept `opts` and look for `:gem` keyword inside opts to determine if it is gem based plugin

Nik08: Making the change, yes!

lib/inspec/plugin/v2/installer.rb
```diff
@@ -342,13 +345,17 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
...
       loader.list_managed_gems.each do |spec|
-        next if spec.name == new_plugin_dependency.name && update_mode
+        # Skip in case of update mode
+        # Skip in case using a gem based plugin
+        next if spec.name == new_plugin_dependency.name && (update_mode || gem_based_plugin?(opts))
 
         spec.activate
       end
 
       # Make sure we remove any previously loaded gem on update
```
sathish-progress, line 355: duplicate comments..

lib/inspec/plugin/v2/installer.rb
```diff
@@ -342,13 +345,17 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
 
       # Activate all current plugins before trying to activate the new one
       loader.list_managed_gems.each do |spec|
-        next if spec.name == new_plugin_dependency.name && update_mode
+        # Skip in case of update mode
```
sathish-progress, line 348: duplicate comments

sathish-progress (approved):

PR #7244, opened 2024-12-20
CHEF-18182 Implement sources handling for gem dependencies

sathish-progress:
## Description

Make gem dependencies work with `source` option in metadata file. When the gem fetcher runs, we know that the requested plugin is going to be needed, so it needs be installed - we may check now to see if the gem is installed and if not install and activate it now.

For example, when we see this in an `inspec.yml`:
```
depends:
 - name: inspec-gcp-resources
    gem: inspec-gcp-resources  
    source: public-gems.progress.com
```
## Types of changes

## Checklist:


lib/inspec/plugin/v2/installer.rb
```diff
@@ -304,11 +304,14 @@ def install_from_remote_gems(requested_plugin_name, opts)
       plugin_dependency = Gem::Dependency.new(requested_plugin_name, opts[:version] || "> 0")
 
       # BestSet is rubygems.org API + indexing, APISet is for custom sources
```
Nik08: We can modify this comment. It is still mentioning about APIset. And since we are now removing "api/v1/dependencies" - should we mention something about this in the comments? Just a thought.

sathish-progress: oh yeah nice catch..

sathish-progress: so I checked if there is any known usage of this custom source with @clintoncwolfe and he confirmed there is nothing that he is aware of.. but yes we can notify this in the changelog

Nik08 (approved):

balasubramanian-s (approved):

PR #7245, opened 2024-12-23
CHEF-18207 Removed SSH resources and tests

sa-progress:
## Description
This pull request removes SSH resources and associated tests from the InSpec core framework. These resources, along with their tests, will now be available in the newly introduced `inspec-ssh-resources` gem. This modular approach allows for better separation of concerns and easier maintenance while keeping the core lightweight. Users can now include the `inspec-ssh-resources` gem as a dependency in their `inspec.yml` to continue using SSH functionalities.

## Related Issue

This change aligns with the ongoing effort to modularize InSpec resources.
## Types of changes

## Checklist:

- [x] I have updated relevant documentation or comments to reflect this change.
- [x] I have verified this change locally.


Nik08:
You will also need to remove the usage of ssh resources from resources.rb file, mock loader and test fixtures for these. Which might also solve the failing pipeline.
Also is there going to be a separate PR for removing docs?

PR #7258, opened 2025-01-08
CHEF-16379 auto-resolve gem conflicts

sathish-progress:
Transitive dependencies fail with conflicts when pointed to rubygems.org source

## Description

Override the required dependencies by choosing the most recent version of already loaded gem vs requested gem and deactivating already older gem version and activating the recent most version from resource packs.

## Related Issue

https://progresssoftware.atlassian.net/browse/CHEF-16379

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -373,9 +373,14 @@ module Inspec::Plugin::V2
 
           gem_name = requested_gemspec.name
           loaded_gem = Gem.loaded_specs[gem_name]
-          # TODO: give priority to more recent version
-          Gem.loaded_specs.delete(gem_name) if loaded_gem && loaded_gem.version != requested_gemspec.version
-          activation_request.full_spec.activate
+          if loaded_gem
+            if requested_gemspec.version > loaded_gem.version
+              Gem.loaded_specs.delete(gem_name)
+            else
+              next # don't activate requested gemspec
+            end
+          end
+          requested_gemspec.activate
         end
       rescue Gem::LoadError => gem_ex
         ex = Inspec::Plugin::V2::InstallError.new(gem_ex.message)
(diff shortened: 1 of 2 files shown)


lib/inspec/plugin/v2/installer.rb
```diff
@@ -368,9 +368,19 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
       # of the same gem on the stack or a malformed dependency.
       begin
         solution.each do |activation_request|
-          unless activation_request.full_spec.activated?
-            activation_request.full_spec.activate
+          requested_gemspec = activation_request.full_spec
```
Nik08, line 371: Can you add some comments over this block around it's functionality? Or the use cases this is handling? I know this might be old code.. but since we have got into it now..Will be good for future ref to understand code later.

sathish-progress: added comments as part of this https://github.com/inspec/inspec/pull/7259

lib/inspec/plugin/v2/installer.rb
```diff
@@ -368,9 +368,19 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
...
+          next if requested_gemspec.activated?
+
+          gem_name = requested_gemspec.name
+          loaded_gem = Gem.loaded_specs[gem_name]
+          if loaded_gem
+            if requested_gemspec.version > loaded_gem.version
+              Gem.loaded_specs.delete(gem_name)
+            else
+              next # don't activate requested gemspec
+            end
```
ahasunos, line 381: Not an important change, but to improve readability, we could consider refactoring something like:
```ruby
if loaded_gem && requested_gemspec.version <= loaded_gem.version
  next # Skip activation if the requested version is not newer
end

Gem.loaded_specs.delete(gem_name) if loaded_gem

```

ahasunos (approved):

lib/inspec/plugin/v2/installer.rb
sathish-progress: I guess you would like the refactoring done here https://github.com/inspec/inspec/pull/7259

PR #7259, opened 2025-01-12
CHEF-18510 resolve gem conflicts installed plugins

sathish-progress:
We have the following conflicts for installed plugins during the second Inspec run.
```
bundle exec inspec exec /Users/sbabu/workspace/inspec-elasticsearch-resources/test/integration/inspec-profile-with-gem-dependency  --log-level=debug --no-create-lockfile
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository: 'ext/ffi_c/libffi/.git'
fatal: not a git repository: 'ext/ffi_c/libffi/.git'
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
[2025-01-08T21:03:27+05:30] ERROR: Could not load plugin inspec-elasticsearch-resources: can't activate concurrent-ruby-1.1.7, already activated concurrent-ruby-1.3.4
[2025-01-08T21:03:28+05:30] ERROR: Errors were encountered while loading plugins...
[2025-01-08T21:03:28+05:30] ERROR: Plugin name: inspec-elasticsearch-resources
[2025-01-08T21:03:28+05:30] ERROR: Error: can't activate concurrent-ruby-1.1.7, already activated concurrent-ruby-1.3.4
[2025-01-08T21:03:28+05:30] ERROR: Run again with --debug for a stacktrace.
```

## Description

- make loader to skip activation of already activated plugins
- skips Dependency activation of plugins when different versions are attempted to install

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -373,9 +373,14 @@ module Inspec::Plugin::V2
 
           gem_name = requested_gemspec.name
           loaded_gem = Gem.loaded_specs[gem_name]
-          # TODO: give priority to more recent version
-          Gem.loaded_specs.delete(gem_name) if loaded_gem && loaded_gem.version != requested_gemspec.version
-          activation_request.full_spec.activate
+          if loaded_gem
+            if requested_gemspec.version > loaded_gem.version
+              Gem.loaded_specs.delete(gem_name)
+            else
+              next # don't activate requested gemspec
+            end
+          end
+          requested_gemspec.activate
         end
       rescue Gem::LoadError => gem_ex
         ex = Inspec::Plugin::V2::InstallError.new(gem_ex.message)
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -12,6 +12,7 @@ require "rubygems/uninstaller"
 require "rubygems/remote_fetcher"
 
 require "inspec/plugin/v2/filter"
+require "inspec/plugin/v2/concerns/gem_spec_helper"
 
 module Inspec::Plugin::V2
   # Handles all actions modifying the user's plugin set:
@@ -23,6 +24,7 @@ module Inspec::Plugin::V2
   class Installer
     include Singleton
     extend Forwardable
+    include Inspec::Plugin::V2::GemSpecHelper
 
     Gem.configuration["verbose"] = false
 
--- a/lib/inspec/plugin/v2/installer.rb
+++ b/lib/inspec/plugin/v2/installer.rb
@@ -355,9 +355,16 @@ module Inspec::Plugin::V2
         # Skip in case of update mode
         # Skip in case using a gem based plugin
         next if spec.name == new_plugin_dependency.name && (update_mode || gem_based_plugin?(opts))
-        # give priority to already loaded gems
-        next if Gem.loaded_specs[spec.name]
 
+        # give priority to already loaded gems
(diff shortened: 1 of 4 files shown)


lib/inspec/plugin/v2/installer.rb
```diff
@@ -368,9 +377,19 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
...
         solution.each do |activation_request|
-          unless activation_request.full_spec.activated?
-            activation_request.full_spec.activate
+          requested_gemspec = activation_request.full_spec
+          next if requested_gemspec.activated?
+
+          gem_name = requested_gemspec.name
+          loaded_gem = Gem.loaded_specs[gem_name]
+          if loaded_gem
+            if requested_gemspec.version > loaded_gem.version
```
Nik08: I am having confusion understanding how does `requested_gemspec.version` and `loaded_gem.version` will be different? Because loaded gem is using the same spec of requested_spec...How would it load a different spec version internally? I know in this scenarios 2 specs exists in the system.. but which one does `Gem.loaded_specs[gem_name]` picks?

lib/inspec/plugin/v2/installer.rb
```diff
@@ -368,9 +377,19 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
       # of the same gem on the stack or a malformed dependency.
       begin
         solution.each do |activation_request|
-          unless activation_request.full_spec.activated?
-            activation_request.full_spec.activate
+          requested_gemspec = activation_request.full_spec
```
Nik08, line 374: Can you add some comments over this block around it's functionality? Or the use cases this is handling? I know this might be old code.. but since we have got into it now..Will be good for future ref to understand code later.

lib/inspec/plugin/v2/installer.rb
```diff
@@ -356,6 +356,15 @@ def install_gem_to_plugins_dir(new_plugin_dependency, # rubocop: disable Metrics
         # Skip in case using a gem based plugin
         next if spec.name == new_plugin_dependency.name && (update_mode || gem_based_plugin?(opts))
 
+        # give priority to already loaded gems
```
Nik08: Can you add some detailed comments over this block around it's functionality? Or the use cases this is handling? I know this might be old code.. but since we have got into it now..Will be good for future ref to understand code later.

lib/inspec/plugin/v2/installer.rb
sathish-progress: No both are different.. loaded gemspec is loaded via bundler when Inspec is booted while requested gemspec is loaded via Gem::RequesSet object

lib/inspec/plugin/v2/installer.rb
sathish-progress: I mean if somebody has to understand this code they have to dig into this either way .. but to make this process easier I shall add some comments

lib/inspec/plugin/v2/installer.rb
sathish-progress: if somebody has to understand this code they have to dig into this either way .. but to make this process easier I shall add some comments

lib/inspec/plugin/v2/installer.rb
Nik08: Thanks!

Nik08 (approved):

ahasunos (approved):

balasubramanian-s (approved):

PR #7261, opened 2025-01-12
CHEF-18596 Fix code scanning alert no. 63: Use of a broken or weak cryptographic hashing algorithm on sensitive data

sathish-progress:
Fixes [https://github.com/inspec/inspec/security/code-scanning/63](https://github.com/inspec/inspec/security/code-scanning/63)

To fix the problem, we need to replace the use of the SHA-1 hashing algorithm with a stronger and more secure hash function. In this case, we can use SHA-256, which is part of the SHA-2 family and is considered secure for cryptographic purposes.

- Replace the `Digest::SHA1` with `Digest::SHA256` in the `uuid_from_string` method.
- Ensure that the rest of the code remains unchanged to maintain existing functionality.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/lib/inspec/reporters/automate.rb
+++ b/lib/inspec/reporters/automate.rb
@@ -66,9 +66,9 @@ module Inspec::Reporters
     # Then it downgrades the 160bit SHA1 to a 128bit
     # then we format it as a valid UUIDv5.
     def uuid_from_string(string)
-      hash = Digest::SHA1.new
+      hash = Digest::SHA256.new
       hash.update(string)
-      ary = hash.digest.unpack("NnnnnN")
+      ary = hash.digest[0, 16].unpack("NnnnnN")
       ary[2] = (ary[2] & 0x0FFF) | (5 << 12)
       ary[3] = (ary[3] & 0x3FFF) | 0x8000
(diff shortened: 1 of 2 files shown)


ahasunos (changes requested):
The verify pipeline appears to be failing due to a broken unit test for Automate. It seems the issue might be caused by a difference in the hash digest generated by `Digest::SHA1.new` and `Digest::SHA256.new`. Updating the hash digest in the Automate unit test could resolve this.

Here's the relevant test that may need attention:
https://github.com/inspec/inspec/blob/6d5bce2b088baa1c7b1c301694d9702bd1174c6a/test/unit/reporters/automate_test.rb#L46

```
describe "#uuid_from_string" do
  it "converts a string to a uuid" do
    end_time = "2018-03-28T14:10:50Z"
    node_uuid = "22ad2f99-f84f-5456-95a0-7e91b4b66690"
    assert = "e8293cb7-f18f-5a59-bf64-5ec9f13517b5"
    _(report.send(:uuid_from_string, end_time + node_uuid)).must_equal assert
  end
end
```

ahasunos (approved):

PR #7262, opened 2025-01-12
CHEF-18596 Fix code scanning alert no. 56: Incomplete regular expression for hostnames

sathish-progress:
Fixes [https://github.com/inspec/inspec/security/code-scanning/56](https://github.com/inspec/inspec/security/code-scanning/56)

To fix the problem, we need to escape the `.` character in the regular expression to ensure it matches only a literal `.` and not any character. This can be done by replacing `.` with `\.` in the relevant strings. This change will ensure that the regular expression matches only the intended hostnames and does not allow unintended matches.

The changes need to be made in the file `test/unit/fetchers/url_test.rb` on line 72, where the URLs are defined. Specifically, we need to update the URLs to escape the `.` characters.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/test/unit/fetchers/url_test.rb
+++ b/test/unit/fetchers/url_test.rb
@@ -64,12 +64,12 @@ describe Inspec::Fetcher::Url do
       _(Inspec::Fetcher::Url.resolve(__FILE__)).must_be_nil
     end
 
-    %w{https://github.com/chef/inspec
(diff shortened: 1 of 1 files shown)


ahasunos (changes requested):
The unit test appears to be failing, and addressing this issue would help get the verify pipeline back to green.

PR #7275, opened 2025-01-21
CHEF-14322 private rubygem server integ

sathish-progress:
All private resource packs will now be hosted in private Chef rubygems server. The plugin system will look into it to fetch the private resource packs.

## Description

- Introduced a new singleton Source Manager object ( Plugin Installer is already a singleton instance )
- check for duplicates 
- Construct Chef rubygems server using available license keys
- custom method to add chef rubygems server 
- refactor the add custom sources 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/gem_source_manager.rb
+++ b/lib/inspec/plugin/v2/gem_source_manager.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 require "singleton" unless defined?(Singleton)
-require 'forwardable' unless defined?(Forwardable)
+require "forwardable" unless defined?(Forwardable)
 require "chef-licensing"
 
 module Inspec::Plugin::V2
@@ -32,7 +32,7 @@ module Inspec::Plugin::V2
     end
 
     def licenses_string
-      ChefLicensing.license_keys.join(',')
+      ChefLicensing.license_keys.join(",")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/inspec/plugin/v2/gem_source_manager.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require "singleton" unless defined?(Singleton)
+require "forwardable" unless defined?(Forwardable)
```
Nik08, line 3: Do we need to use `forwardable` here? Does not seem like we are using it currently here.

sathish-progress: yes thanks for highlighting.. modified to use delegate

lib/inspec/plugin/v2/gem_source_manager.rb
```diff
@@ -0,0 +1,43 @@
...
+require "chef-licensing"
+
+module Inspec::Plugin::V2
+  class GemSourceManager
+    include Singleton
+    extend Forwardable
+
+    DEFAULT_CHEF_RUBY_GEMS_SERVER = "rubygems.chef.io"
+    DEFAULT_USERNAME = "v1"
```
ahasunos, line 12: Not a major concern, but is `DEFAULT_USERNAME` the right name to use in this context?

ahasunos (approved):

lib/inspec/plugin/v2/gem_source_manager.rb
sathish-progress: it is because that is the default username and in future this could become v2 may be but it is out of scope currently

Nik08 (approved):

PR #7276, opened 2025-01-21
CHEF-16676-Postgres session resource fix for special characters in password(InSpec-5)

balasubramanian-s:
Postgres session resource does not function correctly when the password contains special characters.
## Description

Encoded the password variable to ensure it is properly formatted 
and does not break when included in the connection string.
## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,8 +63,9 @@ end
 gem "zeitwerk", "~> 2.6.0", "< 2.7"
 
 # Pinning dry-core,dry-core to < 1.1.0 as it is breaking the build because 1.1.0 is incompatible with the current version, ruby 3.0.x on CI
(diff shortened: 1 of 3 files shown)


ahasunos (approved):

Nik08:
Verify pipeline is breaking.

Gemfile
```diff
@@ -37,7 +37,8 @@ group :test do
   gem "minitest-sprint", "~> 1.0"
   gem "minitest", "5.15.0"
   gem "mocha"
-  gem "nokogiri"
+  # Pinning this version as it breaking for ruby 3.1.0
+  gem "nokogiri", "< 1.17.2"
```
sathish-progress, line 41: why not then pin it specific to ruby 3.1?

Nik08 (approved):

PR #7284, opened 2025-01-29
CHEF-15083: Update groups resource to use `getent` utility to fetch groups info

ahasunos:
## Description

This PR updates the `groups` resource to retrieve group information using the `getent` utility, instead of directly reading it from `/etc/group` file. 

This resolves the limitation where groups added by other methods were not detected, causing failures even if the group existed.

## Related Issue

CHEF-15083: InSpec for group resource using /etc/group instead of using getent

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/groups.rb
+++ b/lib/inspec/resources/groups.rb
@@ -206,32 +206,53 @@ module Inspec::Resources
     private
 
     def get_group_info
-      # check if getent is available on the system
+      # First, try to fetch group info using getent
+      group_info = fetch_group_info_using_getent
+
+      return group_info unless group_info.empty?
+
+      # If getent fails, fallback to reading group info from /etc/group using inspec.etc_group.entries
+      Inspec::Log.debug("Falling back to reading group info from /etc/group as getent is unavailable or failed.")
+      inspec.etc_group.entries
+    end
+
+    # Fetches group information using the getent utility
+    def fetch_group_info_using_getent
+      # Find getent utility on the system
       bin = find_getent_utility
 
-      # fetch details of the groups using getent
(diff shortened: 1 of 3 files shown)


lib/inspec/resources/groups.rb
```diff
@@ -200,7 +200,38 @@ def groups
...
+      end
+    end
+
+    # check if getent exist in the system
+    def find_getent_utility
+      %w{/usr/bin/getent /bin/getent getent}.each do |cmd|
+        return cmd if inspec.command(cmd).exist?
+      end
+
+      raise Inspec::Exceptions::ResourceFailed, "Could not find `getent` on your system."
```
Nik08: If the `getent` utility is not available should we switch back to the old approach of using `inspec.etc_group.entries`?  And then log an error about the unavailability of `getent`?

Similarly, if there is an error encountered while using `getent` utility on line 213, should we again fallback to old approach and give it a try?

ahasunos: Personally, I wouldn't mind having the fallback to fetch group information from `inspec.etc_group.entries`. Do you think we should implement this fallback?

ahasunos: We could log the same exception message, appending an extension that indicates reading group information from `/etc/group` at the debug level, without raising an exception

Nik08: I feel that currently there might be some percentage of users already using this resource without `getent` utility. So it should not break for them if they upgrade. Or there might be a scenario that this one does not cover.

Nik08 (approved):

sathish-progress (approved):

PR #7311, opened 2025-02-25
[Do not merge] CHEF-14268-Update inspec init template with gem dependency

balasubramanian-s:
Update `inspec init profile` with new gem-based resource dependency template 

## Description

inspec init profile --resource-dependency / inspec init profile -r will generate a profile template depending on the inspec gem-based resource

## Types of changes

## Checklist:


lib/plugins/inspec-init/lib/inspec-init/cli_profile.rb
```diff
@@ -24,14 +24,24 @@ def valid_profile_platforms
              desc: "Which platform to generate a profile for: choose from #{valid_profile_platforms.join(", ")}"
       option :overwrite, type: :boolean, default: false,
              desc: "Overwrites existing directory"
+      option :resource_dependency, default: nil, type: :string, aliases: [:r],
```
Nik08, line 27: Are we supporting passing multiple resource pack dependencies in this? If not then simply `aliases: :r` should be used here.

lib/plugins/inspec-init/lib/inspec-init/cli_profile.rb
```diff
@@ -24,14 +24,24 @@ def valid_profile_platforms
...
           unless valid_profile_platforms.include?(options[:platform])
             ui.error "Unable to generate profile: No template available for platform '#{options[:platform]}' (expected one of: #{valid_profile_platforms.join(", ")})"
             ui.exit(:usage_error)
           end
-          template_path = File.join("profiles", options[:platform])
+          # Get the index of "--resource-dependency" in ARGV
+          resource_index = ARGV.index("--resource-dependency")
 
+          # Check if "--resource-dependency" exists but has no argument
+          if resource_index && (resource_index + 1 == ARGV.size || ARGV[resource_index + 1].start_with?("--"))
```
Nik08, line 39: Instead of using ARGV, can you instead try taking the approach of using `options[:resource_dependency]` for `nil` handling?

Nik08: Also check, that even though this new key has a default value of `nil` it seems to be returning `"resource_dependency"=>"resource_dependency"` in thor options when nothing is passed in resource dependency. Please have a check on this behaviour.

PR #7332, opened 2025-03-17
CHEF-20111: Fix `inspec check` for a profile with only comments or empty file

ahasunos:
## Description

This PR introduces a fix to parsing using the safe navigation operator to handle cases where `src.ast` is nil. This ensures that `inspec check` can parse the controls correctly without encountering a runtime exception.

## Related Issue

CHEF-20111: InSpec check fails with new parser when controls file empty or only has comments

## Types of changes

## Checklist:


test/fixtures/profiles/profile-with-comments-and-empty-file/controls/example_with_only_comments.rb
```diff
@@ -0,0 +1,13 @@
+# # copyright: 2018, The Authors
+
+# title "sample section"
```
Nik08: Is it possible for you to separate these 3 different test case scenarios into 3 different test profiles? You might actually not need `test/fixtures/profiles/profile-with-comments-and-empty-file/controls/example.rb` this then since you can use existing ones. So only 2 new profiles then and you can use an existing ones for testing the basic scenario. Usually, we also recommend having different test scenarios in different profiles.

ahasunos: My thought process was that the profile contains multiple files in the controls directory - of which some are empty, some have a basic control (like example.rb in my case) and some only have comments. Because of this, I decided to keep it as a single profile instead of splitting it into multiple profiles.

Nik08: Got it. But if we keep it separate it will be good for us to even reuse these profiles in future. I understand a certain customer used it this way. But maintaining our profile code standards, which is usually keeping different scenarios separately is better for us in my opinion. And internally the control tests are executed in a same manner only. So we should be good even if we separate it.

Nik08 (approved):

PR #7339, opened 2025-03-19
CHEF-18044 Waiver error handling fixes

Nik08:
## Description

This PR includes error-handling improvements for the following scenarios:
- When data is `nil`
- When an unsupported file extension is used
- When the waiver file is empty - only throw a warning and do not fail execution

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/waiver_file_reader.rb
+++ b/lib/inspec/waiver_file_reader.rb
@@ -5,6 +5,8 @@ require "inspec/utils/waivers/json_file_reader"
 module Inspec
   class WaiverFileReader
 
+    SUPPORTED_FILE_EXTENSION = %w{.yaml .yml .csv .json}.freeze
+
     def self.fetch_waivers_by_profile(profile_id, files)
       read_waivers_from_file(profile_id, files) if @waivers_data.nil? || @waivers_data[profile_id].nil?
       @waivers_data[profile_id]
@@ -16,6 +18,13 @@ module Inspec
 
       files.each do |file_path|
         file_extension = File.extname(file_path)
+
+        unless SUPPORTED_FILE_EXTENSION.include?(file_extension)
+          raise Inspec::Exceptions::WaiversFileNotReadable,
+          "Cannot find parser for waivers file '#{file_path}'. " \
+          "Check to make sure file has the appropriate extension. Supported file extensions are: #{SUPPORTED_FILE_EXTENSION.join(", ")}"
+        end
+
         data = nil
         if [".yaml", ".yml"].include? file_extension
           data = Secrets::YAML.resolve(file_path)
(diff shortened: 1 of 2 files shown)


ahasunos (approved):

lib/inspec/waiver_file_reader.rb
```diff
@@ -53,8 +71,10 @@ def self.validate_headers(headers, json_yaml = false)
     end
 
     def self.validate_json_yaml(data)
+      return if data.nil?
+
       headers = []
-      data.each_value do |value|
+      data&.each_value do |value|
```
sathish-progress: since you already check for nil, you don't need this safe check here

lib/inspec/waiver_file_reader.rb
```diff
@@ -15,34 +17,50 @@ def self.read_waivers_from_file(profile_id, files)
...
+        output.merge!(data) if data.is_a?(Hash)
       end
 
       @waivers_data[profile_id] = output
     end
 
+    def self.valid_waiver_file?(file_path)
+      # Check if the file is readable
+      file_extension = File.extname(file_path)
+      unless SUPPORTED_FILE_EXTENSION.include?(file_extension)
```
sathish-progress, line 32: functionally okay. But when the file doesn't exist this condition will raise UnsupportedFileExtension while we should raise as file doesn't exist.

lib/inspec/waiver_file_reader.rb
```diff
@@ -15,34 +17,50 @@ def self.read_waivers_from_file(profile_id, files)
...
+        data = parse_waiver_file(file_path)
+        output.merge!(data) if data.is_a?(Hash)
       end
 
       @waivers_data[profile_id] = output
     end
 
+    def self.valid_waiver_file?(file_path)
+      # Check if the file is readable
+      file_extension = File.extname(file_path)
```
sathish-progress: I think we should also downcase the extension, I have seen people storing extensions in caps for example, `.YAML, .CSV` and OS supports that

sathish-progress (changes requested):

lib/inspec/waiver_file_reader.rb
Nik08: Just a safe check

lib/inspec/waiver_file_reader.rb
Nik08: When a file does not exist, it raises that error via another library since that is not part of the waiver file reader's job. This is only when a file exists and a format is unsupported.

lib/inspec/waiver_file_reader.rb
Nik08: Good point. Will do.

sathish-progress (approved):

PR #7345, opened 2025-03-27
Backports: CHEF-3849: InSpec should exit quickly and clearly if waiver file is malformed/corrupt

Nik08:
Backports #6644 

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/waiver_file_reader.rb
+++ b/lib/inspec/waiver_file_reader.rb
@@ -21,6 +21,9 @@ module Inspec
 
         data = parse_waiver_file(file_path)
         output.merge!(data) if data.is_a?(Hash)
+      rescue Inspec::Exceptions::WaiversFileNotReadable, Inspec::Exceptions::WaiversFileInvalidFormatting => e
+        Inspec::Log.error "Error reading waivers file #{file_path}. #{e.message}"
+        Inspec::UI.new.exit(:usage_error)
       end
 
       @waivers_data[profile_id] = output
@@ -52,7 +55,7 @@ module Inspec
         validate_json_yaml(data)
       when ".csv"
         data = Waivers::CSVFileReader.resolve(file_path)
-        validate_headers(Waivers::CSVFileReader.headers)
+        validate_csv_headers(Waivers::CSVFileReader.headers)
       when ".json"
         data = Waivers::JSONFileReader.resolve(file_path)
         validate_json_yaml(data)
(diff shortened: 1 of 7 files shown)


lib/inspec/waiver_file_reader.rb
```diff
@@ -61,23 +64,58 @@ def self.parse_waiver_file(file_path)
...
-      data.each_value do |value|
-        headers.push value.keys
+      missing_required_field = false
+      data.each do |key, value|
+        # In case of yaml or json we need to validate headers/parametes for each value
+        invalid_headers_info = fetch_invalid_headers_info(value.keys, true)
+        # WARN in case of extra parameters found in each waived control
+        Inspec::Log.warn "Control ID #{key}: extra parameter/s #{invalid_headers_info[:extra_headers]}" unless invalid_headers_info[:extra_headers].empty?
+        unless invalid_headers_info[:missing_required_fields].empty?
+          missing_required_field = true
```
ahasunos, line 108: Just a thought – we could maybe raise the error 

```ruby
raise Inspec::Exceptions::WaiversFileInvalidFormatting,
             "Missing required parameter [justification]. Fix parameters in file to proceed."
```

as soon as we hit the first missing required field instead of looping through the rest of the data. That way we fail fast and avoid doing the checks for the rest of the data. What do you think?

lib/inspec/waiver_file_reader.rb
```diff
@@ -61,23 +64,58 @@ def self.parse_waiver_file(file_path)
...
+          # Log error for each waived control
+          Inspec::Log.error "Control ID #{key}: missing required parameter/s #{invalid_headers_info[:missing_required_fields]}"
+        end
+      end
+
+      # Raise error if any of the waived control has missing required filed
+      if missing_required_field
+        raise Inspec::Exceptions::WaiversFileInvalidFormatting,
+             "Missing required parameter [justification]. Fix parameters in file to proceed."
```
ahasunos, line 117: Raising this exception here may not be needed if we we decide to fail early as mentioned above - keeps things a bit simpler too; thoughts?

lib/inspec/waiver_file_reader.rb
Nik08: We wanted to raise the issues against each Control ID, which is why the iteration is going on. That is the intent of the behaviour.

Nik08: We want the user to know all the missing details against each control ID so that they are not hindered with issues again and again. And they can know about all the places they have to change to fix this at once.

ahasunos (approved):

PR #7366, opened 2025-05-12
Load time improvements for InSpec commands

Nik08:
## Description

This PR includes some initial basic improvements to improve load time for commands like : 
- `inspec version`
- `inspec schema`
- `inspec env` 
- `inspec clear_cache`. 

These basic commands do not do any heavy lifting and should be quick with their results. 
Also, InSpec does not use a plugin-based system for these commands, and hence, the load time, which was added due to plugin loading on each InSpec execution of these commands, could be conditionalised. 

The logic can be made more dynamic and improved. For now, conditionalising the plugin load on each run has drastically improved the load time.

Proof of considerable load time difference in `inspec version` command, **before** vs **after**

Screenshot 2025-05-12 at 5 28 12 PM

## Types of changes

## Checklist:


Vasu1105:
@Nik08, as you are working on these improvements may be able to cherry-pick these changes in this same PR https://github.com/inspec/inspec/pull/7361/commits/52e1a8e098d1730c5e5aec1acdba0a46ce149e35 and then may be we can close other PR where I have added improvements for version command which you already taking care of

PR #7377, opened 2025-05-27
CHEF-21376 Disable licensing and telemetry from InSpec 7

Nik08:
## Description

Disabling licensing and telemetry from InSpec 7
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/plugin/v2/gem_source_manager.rb
+++ b/lib/inspec/plugin/v2/gem_source_manager.rb
@@ -17,9 +17,10 @@ module Inspec::Plugin::V2
 
     def_delegator :@sources, :sources
 
-    def add_chef_rubygems_server
-      register_source(chef_rubygems_server)
-    end
+    ## Disabled licensing - undo this change once ChefLicensing is enabled
+    # def add_chef_rubygems_server
+    #   register_source(chef_rubygems_server)
+    # end
 
     def add(sources)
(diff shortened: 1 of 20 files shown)


lib/inspec/plugin/v2/gem_source_manager.rb
```diff
@@ -37,7 +37,9 @@ def register_source(source)
     end
 
     def licenses_string
```
sathish-progress: since you have disabled it here https://github.com/inspec/inspec/pull/7378 , we shall simply comment this function entirely `def chef_rubygems_server`

Nik08: We can do that, but I kept it as is, just in case we change the implementation of it going forward. Maybe we will still get `chef_rubygem_server` in future, but probably without licensing? And since it won't be called, there is no loss at this point.

Vasu1105 (approved):
LGTM

lib/inspec/plugin/v2/gem_source_manager.rb
sathish-progress: CHEF rubygem server cannot exit without licensing, it defeats its purpose and even if in worst case we want to bring it back there will still be modifications on the URL as it is tied to licensing today. So I would still stand by my recommendation here.

sathish-progress (approved):

PR #7398, opened 2025-06-23
CHEF-18030- Fix GitFetcher to Clear Empty Cache Directory on Fetch

balasubramanian-s:
In case of a Git fetch failure, the created cache directory remains empty, causing the Test Kitchen suite to fail in subsequent executions across multiple platforms.

## Description

FIXES
- [x] Added logic to detect and remove empty cache directories when a Git fetch fails.
- [x] Ensured cleanup is performed before retrying or exiting the fetch process.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/fetcher/git.rb
+++ b/lib/inspec/fetcher/git.rb
@@ -73,7 +73,9 @@ module Inspec::Fetcher
             # this means the git repository did not contain any files (or the checkout failed).
             # In this case, remove the destination directory to avoid
             # leaving an empty or invalid profile directory.
-            FileUtils.rm_r(destination_path)
+            if Dir.exist?(destination_path)
+              FileUtils.rm_rf(destination_path)
+            end
             raise Inspec::FetcherFailure, "Profile git dependency failed for #{@remote_url} - no files found in the repository."
           end
           if @relative_path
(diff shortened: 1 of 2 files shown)


lib/inspec/fetcher/git.rb
```diff
@@ -68,11 +68,18 @@ def fetch(destination_path)
       else
         Dir.mktmpdir do |working_dir|
           checkout(working_dir)
-          if @relative_path
+          if Dir.empty?(working_dir)
+            # If the temporary working directory is empty after checkout,
+            # this means the git repository did not contain any files (or the checkout failed).
+            # In this case, remove the destination directory to avoid
+            # leaving an empty or invalid profile directory.
+            Inspec::Log.debug("Remove #{destination_path} directory because temp directory {working_dir} it is empty")
```
SathishAchilles: #{working_dir}

lib/inspec/fetcher/git.rb
```diff
@@ -68,11 +68,18 @@ def fetch(destination_path)
...
+            FileUtils.rm_r(destination_path)
+          elsif @relative_path
             perform_relative_path_fetch(destination_path, working_dir)
```
SathishAchilles, line 82: What if it fails here ?

sathish-progress (approved):

Nik08 (approved):

SathishAchilles (approved):

PR #7402, opened 2025-06-30
FOREPORT: CHEF-23231 Upgrading to latest version of train and train-winrm - InSpec 7

Nik08:
## Description

- Pinning InSpec to the latest versions of train and train-winrm
- Solving verify pipeline issues caused due to warnings - circular dependency warning, redefined variable warning, unused variable warning.

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/backend.rb
+++ b/lib/inspec/backend.rb
@@ -17,7 +17,11 @@ module Inspec
 
     def self.add_resource_methods # TODO: remove
       Inspec::Resource.registry.each do |id, r|
-        define_method id.to_sym do |*args|
+        method_name = id.to_sym
+        # Skip if method is already defined (e.g. built-in Ruby methods or methods from gems)
+        next if instance_methods.include?(method_name)
+
+        define_method method_name do |*args|
           r.new(self, id.to_s, *args)
         end
       end
--- a/lib/inspec/control_eval_context.rb
+++ b/lib/inspec/control_eval_context.rb
@@ -142,7 +142,7 @@ module Inspec
         ::Inspec::Rule.set_skip_rule(control, true, msg)
       end
 
-      profile_context_owner.register_rule(control, &block) unless control.nil?
+      profile_context_owner.register_rule(control) unless control.nil?
     end
 
     def input(input_name, options = {})
--- a/lib/inspec/resources/file.rb
+++ b/lib/inspec/resources/file.rb
@@ -46,9 +46,9 @@ module Inspec::Resources
     %w{
       type exist? file? block_device? character_device? socket? directory?
       symlink? pipe? mode mode? owner owned_by? group grouped_into?
-      link_path shallow_link_path linked_to? mtime size selinux_label immutable?
+      link_path shallow_link_path linked_to? mtime size selinux_label
       product_version file_version version? md5sum sha256sum
-      path basename source source_path uid gid
+      basename source source_path uid gid
     }.each do |m|
       define_method m do |*args|
         file.send(m, *args)
--- a/lib/inspec/utils/telemetry.rb
+++ b/lib/inspec/utils/telemetry.rb
@@ -2,7 +2,6 @@ require "time" unless defined?(Time.zone_offset)
 # require "chef-licensing" ## Disabled licensing
 require_relative "telemetry/null"
 require_relative "telemetry/http"
-require_relative "telemetry/run_context_probe"
 
 module Inspec
   class Telemetry
@@ -18,6 +17,9 @@ module Inspec
       # Don't perform telemetry action for other InSpec distros
       # Don't perform telemetry action if running under Automate - Automate does LDC tracking for us
       # Don't perform telemetry action if license is a commercial license
+
+      # Require the run_context_probe only when needed to avoid circular dependency
+      require_relative "telemetry/run_context_probe"
 
       if Inspec::Dist::EXEC_NAME != "inspec" ||
           Inspec::Telemetry::RunContextProbe.under_automate? ||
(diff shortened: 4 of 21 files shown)


lib/inspec/backend.rb
```diff
@@ -17,7 +17,11 @@ class Backend
 
     def self.add_resource_methods # TODO: remove
       Inspec::Resource.registry.each do |id, r|
-        define_method id.to_sym do |*args|
+        method_name = id.to_sym
+        # Skip if method is already defined (e.g. built-in Ruby methods or methods from gems)
+        next if instance_methods.include?(method_name)
```
sathish-progress, line 22: but what if we really want to override the default methods in the resource this would not allow it to be picked right ?

lib/inspec/control_eval_context.rb
```diff
@@ -142,7 +142,7 @@ def register_control(control, &block)
         ::Inspec::Rule.set_skip_rule(control, true, msg)
       end
 
-      profile_context_owner.register_rule(control, &block) unless control.nil?
+      profile_context_owner.register_rule(control) unless control.nil?
```
sathish-progress, line 145: why remove block argument?

lib/inspec/resources/file.rb
```diff
@@ -46,9 +46,9 @@ def initialize(path)
     %w{
       type exist? file? block_device? character_device? socket? directory?
       symlink? pipe? mode mode? owner owned_by? group grouped_into?
-      link_path shallow_link_path linked_to? mtime size selinux_label immutable?
+      link_path shallow_link_path linked_to? mtime size selinux_label
```
sathish-progress, line 49: why remove `immutable?` and `path`?

lib/inspec/utils/telemetry.rb
```diff
@@ -19,6 +18,9 @@ def self.determine_backend_class
       # Don't perform telemetry action if running under Automate - Automate does LDC tracking for us
       # Don't perform telemetry action if license is a commercial license
 
+      # Require the run_context_probe only when needed to avoid circular dependency
+      require_relative "telemetry/run_context_probe"
```
sathish-progress: may be we can do?
```suggestion
      require_relative "telemetry/run_context_probe" unless defined?(Inspec::Telemetry::RunContextProbe)
```

lib/inspec/utils/telemetry/base.rb
```diff
@@ -20,6 +20,9 @@ def fetch_license_ids
       end
 
       def create_wrapper
+        # Require the run_context_probe only when needed to avoid circular dependency
+        require_relative "run_context_probe"
```
sathish-progress: same here 
```suggestion
        require_relative "run_context_probe" unless defined?(Inspec::Telemetry::RunContextProbe)
```

lib/plugins/inspec-init/lib/inspec-init/renderer.rb
```diff
@@ -82,9 +82,14 @@ def render_with_values(template_subdir_path, template_type, template_values = {}
...
+          method_names = Set.new
           hash.each do |key, value|
-            define_method key.to_sym do
-              value
+            method_name = key.to_sym
+            unless method_names.include?(method_name)
+              method_names << method_name
+              define_method method_name do
+                value
+              end
```
sathish-progress, line 92: am really curious how will there be duplicate keys in a hash?

lib/plugins/inspec-reporter-html2/lib/inspec-reporter-html2/reporter.rb
```diff
@@ -13,6 +13,10 @@ def render
       js_path = cfg[:alternate_js_file] || (template_path + "/default.js")
       css_path = cfg[:alternate_css_file] || (template_path + "/default.css")
 
+      # Mark variables as used (they're used in the ERB template via binding)
```
sathish-progress, line 16: why are we marking these variables as used? are we being flagged somewhere?

lib/inspec/control_eval_context.rb
Nik08: It was not used anywhere, and there were warnings for the unused block. It is not used within the logic of the function as well.

lib/inspec/resources/file.rb
Nik08: So these methods are already defined in this resource, and hence there was a warning for this. Something around redefinition.

lib/inspec/utils/telemetry.rb
Nik08: I will try and see if this removes the warning we were getting.

lib/plugins/inspec-reporter-html2/lib/inspec-reporter-html2/reporter.rb
Nik08: Yes, the changes in these code are specific to solving the warnings we have been receiving in CI and with the upgrade of train based gems because of low level design issues in our code.

sathish-progress: got it .. do we really use these variables then dynamically maybe? should we get rid of it ?

Nik08: Yeah we use these variables

lib/inspec/utils/telemetry.rb
Nik08: This did not work, but i found the alternate solution to this..

lib/inspec/utils/telemetry/base.rb
Nik08: Found alternate solution to this, and this did not work either

Nik08:
Closing this - will raise a new PR once the issue is fixed in `chef-winrm-fs`

PR #7459, opened 2025-07-24
Externalize HAB Env from Plan files

balasubramanian-s:
Externalize HAB Env from Plan files
## Description

Change the way we set HAB Env's so that we can replicate the same pipelines for different channels
## Types of changes

## Checklist:


balasubramanian-s:
Acheived through github actions
https://github.com/inspec/inspec/pull/7466

PR #7495, opened 2025-08-11
Revert "CHEF-21376 Disable licensing and telemetry from InSpec 7"

balasubramanian-s:
Reverts inspec/inspec#7377


Nik08 (approved):
Will create a new PR with the additional check for telemetry - to only run when the `make_licensing_optional` flag is false.

PR #7501, opened 2025-08-15
Require CSV gem to avoid warnings on modern Ruby

tas50:
InSpec is making a lot of warning noise in any project that has it in the dep tree. This avoids that by requiring the csv gem that is used by the project.

Example:

```
 /home/runner/work/kitchen-dokken/kitchen-dokken/vendor/bundle/ruby/3.3.0/gems/inspec-core-6.8.24/lib/inspec/waiver_file_reader.rb:2: warning: csv was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
You can add csv to your Gemfile or gemspec to silence this warning.
```

While I was here I set the ruby version in cookstyle to match the gemspec

## Types of changes

## Checklist:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 AllCops:
+  TargetRubyVersion: 3.1 # this should match the required ruby in the gemspec
   Exclude:
     - 'test/kitchen/**/*'
     - 'test/integration/**/controls/**/*.rb'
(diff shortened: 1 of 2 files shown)


tas50:
Closing as there doesn't seem to be an interest in fixing this issue

PR #7506, opened 2025-08-21
CHEF-26507 Adds copilot instructions file  - InSpec 6

Vasu1105:
## Description
This PR adds GitHub Copilot instructions for the InSpec project to improve AI-assisted development workflows.
- Provides consistent guidance for AI-assisted development
- Reduces onboarding time for new contributors
- Ensures code quality and adherence to InSpec patterns
- Prevents common mistakes in documentation and code changes
- Streamlines workflow from issue analysis to PR creation
## Types of changes

## Checklist:

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -562,15 +562,31 @@ All tasks should follow this prompt-based workflow:
    - Identify files to be modified/created
    - Plan test strategy
 3. **Provide Summary**: Summarize understanding and ask for confirmation
-4. **Prompt**: **Analysis Complete**. Next step: Start implementation. Other steps remaining: Implementation → Testing → PR Creation. Do you want to continue with the implementation?"
+4. **Prompt**: **Analysis Complete**. Next step: Start implementation. Other steps remaining: Implementation → Testing → Committing the code → PR Creation. Do you want to continue with the implementation?"
 
 #### Step 2: Implementation
 1. **Code Implementation**:
-   - Follow existing code patterns and styles
-   - Implement changes incrementally
-   - Follow Ruby/InSpec best practices
+   - **Read InSpec Context First**: Review relevant sections from this instructions file based on your task:
+     - **Documentation work**: Follow the **Documentation Structure and Guidelines** section
+     - **CLI changes**: Reference **CLI Option Management** patterns and **Base CLI Options**
+     - **Resource development**: Use **Resource Development** patterns and check `lib/inspec/resources/`
+     - **Transport/connection work**: Review **Transport Layer Integration** and **Remote Connection Development**
+     - **Testing framework**: Follow **Testing Patterns** section (Unit Tests, Functional Tests, Integration Tests)
+     - **Licensing features**: Reference **Licensing System** integration patterns
+     - **Plugin development**: Check **Plugin system files** and existing plugins in `lib/plugins/`
+     - **Packaging/distribution**: Review **Packaging and Distribution** section for Habitat/Omnibus changes
+   - **Follow Coding Standards**:
+     - Use **Code Style Guidelines** for Ruby version compatibility and string literals
+     - Follow **Error Handling Patterns** for InSpec-specific exceptions
+     - Check **Key Development Patterns** for architecture-specific guidance
+   - **Implementation Approach**:
+     - Follow existing code patterns and styles in similar files
+     - Implement changes incrementally
+     - Follow Ruby/InSpec best practices as documented above
+     - Check **Common File Locations** to understand where different types of code should go
+     - Review **Files and Areas to Avoid Modifying** before making changes
 2. **Progress Updates**: After each significant change, provide summary
-3. **Prompt**: **Implementation Complete**. Next step: Create unit tests. Other steps remaining: Testing → PR Creation. Do you want to continue with creating tests?"
+3. **Prompt**: **Implementation Complete**. Next step: Create unit tests. Other steps remaining: Testing → Committing the code → PR Creation. Do you want to continue with creating tests?"
 
 #### Step 3: Testing
 1. **Create Unit Tests**:
@@ -579,16 +595,20 @@ All tasks should follow this prompt-based workflow:
    - Use Minitest framework with describe/it syntax as documented
    - Follow existing test patterns in `test/unit/` and use `MockLoader` for resource testing
    - Ensure good test coverage
-2. **Run Tests**: Execute tests to verify functionality
-3. **Prompt**: **Tests Complete**. Next step: Create PR with GitHub CLI. Other steps remaining: PR Creation. Do you want to continue with PR creation?"
+   - Skip the test cases if the change is a documentation change.
+2. **Run Tests**: Execute tests to verify functionality if test cases were added
+3. **Prompt**: **Tests Complete**. Next step: Committing the code. Other steps remaining: Committing the code → PR Creation. Do you want to continue with committing the code?"
 
-#### Step 4: PR Creation
+#### Step 4: Committing the code
 1. **Create Branch**: Use Jira ID as branch name (if provided)
-2. **Commit Changes**: Stage and commit all changes
-3. **Push Branch**: Push to remote repository
-4. **Create PR**: Use GitHub CLI to create pull request
-5. **PR Description**: Include HTML-formatted summary of changes
-6. **Final Prompt**: **PR Created Successfully**. All steps completed!"
+2. **Commit Changes**: Stage and commit all changes. Use meaningful commit messages. Create separate commits for different types of changes (code, tests, docs).
+6. **Prompt**: **Code committed Successfully**. Next step: Create PR with GitHub CLI. Other steps remaining: PR Creation. Do you want to continue with PR creation?"
+
+#### Step 5: PR Creation
+1. **Push Branch**: Push to remote repository
+2. **Create PR**: Use GitHub CLI to create pull request
+3. **PR Description**: Include HTML-formatted summary of changes
+4. **Final Prompt**: **PR Created Successfully**. All steps completed!"
 
 ### 3. Prompt-Based Execution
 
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,255 @@
...
+## Code Style Guidelines
+
+### Ruby Version Compatibility
+
+The project supports multiple Ruby versions with conditional dependencies:
+
+```ruby
+# In Gemfile - version-specific constraints
+if RUBY_VERSION < "3.1.0"
+  gem "dry-types", "<= 1.7.2"
```
Nik08, line 190: Looks like this was generated using copilot - this section could take some work

.github/copilot-instructions.md
```diff
@@ -0,0 +1,255 @@
...
+
+CLI options are defined in `lib/inspec/base_cli.rb` using Thor framework:
+
+```ruby
+def self.target_options
+  option :target, aliases: :t, type: :string,
+    desc: "Simple targeting option using URIs"
+  option :backend, aliases: :b, type: :string,
+    desc: "Choose a backend: local, ssh, winrm, docker"
+  # WinRM-specific options
```
Nik08: I think this is incomplete as well. Since we do not only have cli options only via `target_options`

.github/copilot-instructions.md
Vasu1105: We can keep updating it.

.github/copilot-instructions.md
Vasu1105: Same we can keep updating it.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+- **Bug Fixes**: Important bug resolutions, critical fixes, or corrections to existing functionality
+- **CVEs**: Any security vulnerabilities addressed, security patches, or security-related fixes
+- **Dependency Updates**: Updates to key dependencies, Ruby version changes, or major library upgrades
+
+**Content Structure Guidelines:**
+- **New Features**: Major functionality additions, new resources, new CLI commands, or significant new capabilities
+- **Improvements**: Performance enhancements, usability improvements, code optimizations, or enhanced existing features
+- **Bug Fixes**: Important bug resolutions, critical fixes, or corrections to existing functionality
+- **CVEs**: Any security vulnerabilities addressed, security patches, or security-related fixes
+- **Dependency Updates**: Updates to key dependencies, Ruby version changes, or major library upgrades
```
sathish-progress: This section is duplicated

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+
+**Prohibited Files/Areas**:
+- `VERSION` file (managed by release process)
+- `.expeditor/` configuration (CI/CD automation)
+- `habitat/` packaging files (unless specifically required)
+- `omnibus/` packaging (unless specifically required)
+- `Gemfile.lock` (generated file)
+- Any files marked as auto-generated
+
+**Restricted Areas** (modify with extreme caution):
```
sathish-progress: Why are these flagged as restricted? this would only confuse the models to find an alternate viable solution when changes really have to be made within these.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+
+### 6. Files and Areas to Avoid Modifying
+
+**Prohibited Files/Areas**:
+- `VERSION` file (managed by release process)
+- `.expeditor/` configuration (CI/CD automation)
+- `habitat/` packaging files (unless specifically required)
+- `omnibus/` packaging (unless specifically required)
+- `Gemfile.lock` (generated file)
+- Any files marked as auto-generated
```
sathish-progress: https://docs.github.com/en/copilot/how-tos/configure-content-exclusion/exclude-content-from-copilot is the recommended way to exclude indexing in copilot. This is way too ambigous.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+  - [ ] I have read the **CONTRIBUTING** document.
+
+### 5. MCP Server Integration (atlassian-mcp-server)
+
+For Jira interactions, use the following MCP server functions:
+
+- **Get Issue**: `mcp_atlassian-mcp_getJiraIssue` - Fetch issue details
+- **Search Issues**: `mcp_atlassian-mcp_searchJiraIssuesUsingJql` - Search for related issues
+- **Add Comments**: `mcp_atlassian-mcp_addCommentToJiraIssue` - Add progress comments
+- **Transition Issue**: `mcp_atlassian-mcp_transitionJiraIssue` - Move issue status
```
sathish-progress, line 716: This assumes that every environment with the repo and copilot has mcp server settings which is not the case.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
+# GitHub Copilot Instructions for InSpec
```
sathish-progress, line 1: in general we have clubbed context, agents and instructions all in 1 file. We should be following https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=webui#creating-repository-wide-custom-instructions-1

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+
+- **Thor** - CLI framework
+- **Train** - Transport abstraction layer
+- **RSpec** - Testing framework (for some tests)
+- **Minitest** - Primary testing framework  
+- **Chef ecosystem gems** - Various Chef/InSpec utilities
+
+## When Writing Code
+
+1. **Follow existing patterns** in similar files
```
sathish-progress: This is ambiguous. We should tell it specifically what existing patterns.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+- **Handle network timeouts** gracefully
+- **Escape shell commands** properly
+
+This project emphasizes **security**, **compliance testing**, and **cross-platform compatibility**, so always consider these aspects when contributing code.
+
+## Task Implementation Workflow
+
+### 1. Jira Integration with MCP Server
+
+When a Jira ID is provided:
```
sathish-progress, line 545: I think we can add 
```
If MCP server is unavailable, Copilot must request the developer to paste Jira story details manually.
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+   - Use Minitest framework with describe/it syntax as documented
+   - Follow existing test patterns in `test/unit/` and use `MockLoader` for resource testing
+   - Ensure good test coverage
+   - Skip the test cases if the change is a documentation change.
+2. **Run Tests**: Execute tests to verify functionality if test cases were added
+3. **Prompt**: **Tests Complete**. Next step: Committing the code. Other steps remaining: Committing the code → PR Creation. Do you want to continue with committing the code?"
+
+#### Step 4: Committing the code
+1. **Create Branch**: Use Jira ID as branch name (if provided)
+2. **Commit Changes**: Stage and commit all changes. Use meaningful commit messages. Create separate commits for different types of changes (code, tests, docs).
```
sathish-progress: > Stage and commit all changes
This is not correct. There can be files on the file system that should be exempt or need not be committed. We can ask it to prompt the dev before staging all the changes.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+6. **Prompt**: **Code committed Successfully**. Next step: Create PR with GitHub CLI. Other steps remaining: PR Creation. Do you want to continue with PR creation?"
+
+#### Step 5: PR Creation
+1. **Push Branch**: Push to remote repository
+2. **Create PR**: Use GitHub CLI to create pull request
+3. **PR Description**: Include HTML-formatted summary of changes
```
sathish-progress: it should respect the existing PR template

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+1. **Authentication**: Use GitHub CLI authentication (no .profile references)
+   ```bash
+   gh auth status
+   # If not authenticated:
+   gh auth login
+   ```
+
+2. **Branch Creation**: Use Jira ID as branch name
+   ```bash
+   git checkout -b {JIRA_ID}
```
sathish-progress, line 648: I personally like `<JIRA_ID>_<summary_of_feature>` so we don't have to look into JIRA to understand what these changes are

.github/copilot-instructions.md
```diff
@@ -0,0 +1,741 @@
...
+   ```
+
+3. **PR Creation**: Use GitHub CLI with HTML-formatted description
+   ```bash
+   gh pr create --title "feat: {JIRA_ID} - Brief description" --body "{HTML_DESCRIPTION}"
```
sathish-progress: it should be clear that  `feat` should be used for features only and as I said above the PR body should respect existing template

sathish-progress (changes requested):

.github/copilot-instructions.md
Nik08: @sathish-progress No every repo that we will use and every copilot will needs to be setup with mcp. That is the requirement going forward.

.github/copilot-instructions.md
Nik08: template is mentioned below - which is using existing template

.github/copilot-instructions.md
Nik08: We can modify this but a lot of this has been followed by instructions setup internally for projects.

.github/copilot-instructions.md
Nik08: So the way this instruction file has been generated is by using prompting with copilot itself - these are not hand written for the sake of model to understand this more clearly.

Nik08: But will read more on this. The part of this task was to collab with copilot to create instructions which can be understood by copilot itself.

.github/copilot-instructions.md
Nik08: We can test this out. And make this more accurate. But last time I tried - it only picked the changes it made.

.github/copilot-instructions.md
Nik08: I dont think we need this.... because we are making usage of mcp mandatory for our internal use cases

PR #7513, opened 2025-09-01
CHEF-23442 Added support to use trace level with log level option

Nik08:
## Description

Added support to use trace level with log level option
## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/base_cli.rb
+++ b/lib/inspec/base_cli.rb
@@ -370,7 +370,7 @@ module Inspec
     # get the log level
     # DEBUG < INFO < WARN < ERROR < FATAL < UNKNOWN
     def get_log_level(level)
-      valid = %w{debug info warn error fatal}
+      valid = %w{trace debug info warn error fatal}
 
       if valid.include?(level)
         l = level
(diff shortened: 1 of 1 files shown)


lib/inspec/base_cli.rb
```diff
@@ -370,15 +370,15 @@ def config
...
+      valid = %w{trace debug info warn error fatal}
 
       if valid.include?(level)
         l = level
       else
         l = "info"
       end
 
-      Logger.const_get(l.upcase)
+      Mixlib::Log.const_get(l.upcase)
```
sathish-progress, line 381: here we change the logger library itself .. is this necessary ? are we sure it has not caused any formatting changes or any changes to output and available methods across Inspec?

Nik08: No this will not cause any formatting changes - This is only used to access the value - Logger library itself does not support `trace`. In mixlib log there is a monkey patch done on logger for extending trace

Nik08: This method only does get operation of constant

sathish-progress: ohh right .. thanks for clarifying.

sathish-progress (approved):

PR #7603, opened 2025-10-15
CHEF-27163-Add github release pipeline

balasubramanian-s:
Add github release pipeline

Tested these changes in private repo and it creates release as below
image

This was tested using current `inspec-7`'s changelog. And since the last release was beta release - it create release tag for that version during testing.

## Description

Fetching the version from changelog file and using it for creating releases

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/create-github-release.yml
+++ b/.github/workflows/create-github-release.yml
@@ -1,0 +1,38 @@
+name: Create GitHub Release
+
+on: workflow_dispatch
+
+jobs:
+  create_release:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write  # Needed to create a release
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Read version from VERSION file
+        id: get_version
+        run: |
+          VERSION=$(cat VERSION | tr -d ' \n')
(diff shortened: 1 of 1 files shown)


Nik08 (approved):

.github/workflows/create-github-release.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write  # Needed to create a release
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: inspec-7
```
sathish-progress, line 16: this can be parameterised , so we can use for other branches..

sathish-progress: something like this?

```ref: ${{ github.event.inputs.branch }}```
and we define an input block to accept this in wf

sathish-progress (approved):
some enhancements recommended..

.github/workflows/create-github-release.yml
Nik08: @sathish-progress  We don't need this for other branches; they are automated via expeditor. This was only not going to work going forward for InSpec 7, because we have started using a new workload in InSpec 7, `project_promoted` for promotion. And this workload does not support creating GitHub release tags via Expeditor itself.

PR #7616, opened 2025-11-07
Exclude git folder in the cache

balasubramanian-s:
Fix for Windows Permission Error in GitFetcher
## Problem

When using InSpec with Git-based profile dependencies on Windows (especially in Test Kitchen), users encounter permission errors like:

```
Permission denied @ rb_sysopen - C:/windows/system32/config/systemprofile/.inspec/cache/<hash>/./.git/objects/pack/pack-<hash>.idx
```

## Root Cause

The `GitFetcher` class was copying the **entire** Git repository (including the `.git` directory) to the InSpec cache. The `.git` directory contains Git's internal files with specific permission requirements that cause issues on Windows, especially in system profile directories.

### Why This Happens

1. Git's internal files (in `.git/objects/pack/`) have specific permission requirements
2. On Windows, especially under system accounts, these permissions cannot be properly replicated
3. The `.git` directory is not needed for InSpec profile execution - only the profile content is required
4. Copying `.git` also unnecessarily bloats the cache
## Related Issue

https://github.com/inspec/inspec/pull/7398 fixed a different but related issue - it addressed **empty** Git repositories leaving invalid cache directories. This fix addresses **permission issues with `.git` directories** in the cache.

Both fixes improve the reliability of Git-based profile dependencies, but target different failure modes:
- CHEF-18030: Empty repos leaving invalid cache entries
- This fix: Permission errors from copying `.git` to cache

## Types of changes

## Checklist:


balasubramanian-s:
Found Alternative fixes
https://github.com/inspec/inspec/pull/7658

PR #7656, opened 2025-12-04
CHEF-28672  Updated resolv dependency

Nik08:
## Description

Updated version to `0.2.3`
## Types of changes

## Checklist:

### DIFF
--- a/post-bundle-install.rb
+++ b/post-bundle-install.rb
@@ -1,0 +1,27 @@
+#!/usr/bin/env ruby
+
+gem_home = Gem.paths.home
+
+puts "fixing bundle installed gems in #{gem_home}"
+
+# Handle resolv gem conflict with default ruby gem
+puts "Checking resolv gem installation..."
+resolv_info = `gem info resolv`
+
+if resolv_info.include?("Installed at (default):") && resolv_info.include?("resolv (0.2.1)")
+  # Extract the default gem path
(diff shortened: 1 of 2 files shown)


Nik08:
Adhoc build https://buildkite.com/chef/inspec-inspec-inspec-5-omnibus-adhoc/builds/222

post-bundle-install.rb
```diff
@@ -0,0 +1,27 @@
...
+  default_path = resolv_info.match(/Installed at \(default\): (.+)$/)[1]
+
+  if default_path
+    gemspec_path = File.join(default_path.strip, "specifications", "default", "resolv-0.2.1.gemspec")
+
+    if File.exist?(gemspec_path)
+      puts "Removing default resolv gemspec: #{gemspec_path}"
+      File.delete(gemspec_path)
+    end
+  end
```
sathish-progress, line 22: why not check using rubygems?

```
spec = Gem::Specification.find_by_name('resolv')
pec.version == Gem::Version.new('0.2.1') && spec.default_gem?
```

Nik08: The suggestion is pretty neat. But this is not working - and breaks when there are multiple versions of the same gem present. Using `gem info` seems safer in this scenario. On testing, I found - `Gem::Specification.find_by_name('resolv')`, this is not always returning the default gem, but the latest upgraded one.

sathish-progress (approved):

PR #7684, opened 2025-12-15
CHEF-28019: Enable TCPS Connection to Oracle DB

sa-progress:
- Add tns_alias parameter for TNS-based connections (required for TCPS/SSL)
- Add env parameter to pass environment variables (TNS_ADMIN, LD_LIBRARY_PATH, etc.)
- Add debug parameter to output masked command for troubleshooting
- Support both TNS alias and traditional host:port/service connection methods
- Maintain full backward compatibility with existing code
- Enable Oracle TCPS (SSL) connections with proper certificate DN validation

Tested with Oracle XE 21c using TCPS protocol with wallet-based SSL configuration. All acceptance tests pass (11/11) including:
- TCPS connectivity with TNS alias
- Environment variable support
- Certificate DN validation
- Debug mode functionality
- Backward compatibility with TCP and host:port/service

## Types of changes

## Checklist:

### DIFF
--- a/lib/inspec/resources/oracledb_session.rb
+++ b/lib/inspec/resources/oracledb_session.rb
@@ -52,11 +52,11 @@ module Inspec::Resources
       @db_role = opts[:as_db_role]
       @sqlcl_bin = opts[:sqlcl_bin] || nil
       @sqlplus_bin = opts[:sqlplus_bin] || "sqlplus"
-      
+
       # CHEF-28019: Support for TNS alias and environment variables
       @tns_alias = opts[:tns_alias]
       @env_vars = opts[:env] || {}
-      
+
       skip_resource "Option 'as_os_user' not available in Windows" if inspec.os.windows? && su_user
       fail_resource "Can't run Oracle checks without authentication" unless su_user || (user || password)
     end
@@ -182,14 +182,14 @@ module Inspec::Resources
       env_prefix = build_env_prefix
       connect_string = build_connect_string
       heredoc_content = "connect #{connect_string}\n#{format_options}\n#{verified_query}\nEXIT"
-      
+
       if @su_user
         cmd = %{su - #{@su_user} -c "#{oracle_echo_str} #{env_prefix} #{@bin} -s /nolog <<'INSPECSQL'\n#{heredoc_content}\nINSPECSQL"}
       else
         cmd = %{#{oracle_echo_str}#{bin} -s /nolog <<'INSPECSQL'\n#{heredoc_content}\nINSPECSQL}
         cmd = env_prefix.empty? ? cmd : "#{env_prefix} #{cmd}"
       end
-      
+
       cmd
     end
 
@@ -204,7 +204,7 @@ module Inspec::Resources
     def build_env_prefix
       return "" if @env_vars.nil? || @env_vars.empty?
 
-      @env_vars.map { |k, v| "#{k}='#{v}'" }.join(' ')
+      @env_vars.map { |k, v| "#{k}='#{v}'" }.join(" ")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


test/unit/resources/oracledb_session_test.rb
```diff
@@ -194,4 +194,155 @@
...
+    resource = quick_resource(:oracledb_session, :linux,
+      user: "USER",
+      password: "password",
+      tns_alias: "XEPDB1_TCPS",
+      sqlplus_bin: "/bin/sqlplus")
+
+    # When using TNS alias, resource_id should show the alias
+    # Note: Current implementation shows localhost-1521-USER
+    # This test documents the current behavior
+    _(resource.resource_id).must_equal "localhost-1521-USER"
```
sathish-progress: rather why not use the TNS alias as resource_id in this PR itself?

lib/inspec/resources/oracledb_session.rb
```diff
@@ -153,5 +176,35 @@ def parse_csv_result(stdout)
...
+    def build_tns_command(format_options, verified_query, oracle_echo_str)
+      env_prefix = build_env_prefix
+      connect_string = build_connect_string
+      heredoc_content = "connect #{connect_string}\n#{format_options}\n#{verified_query}\nEXIT"
+
+      if @su_user
+        cmd = %{su - #{@su_user} -c "#{oracle_echo_str} #{env_prefix} #{@bin} -s /nolog <<'INSPECSQL'\n#{heredoc_content}\nINSPECSQL"}
+      else
+        cmd = %{#{oracle_echo_str}#{bin} -s /nolog <<'INSPECSQL'\n#{heredoc_content}\nINSPECSQL}
+        cmd = env_prefix.empty? ? cmd : "#{env_prefix} #{cmd}"
```
sathish-progress:
```suggestion
        cmd = "#{env_prefix} #{cmd}" unless env_prefix.empty?
```

test/unit/resources/oracledb_session_test.rb
```diff
@@ -194,4 +194,155 @@
...
+      password: "password",
+      tns_alias: "XEPDB1_TCPS",
+      env: {
+        "TNS_ADMIN" => "/opt/oracle/network/admin",
+        "LD_LIBRARY_PATH" => "/opt/oracle/lib",
+      },
+      sqlplus_bin: "/bin/sqlplus") do |cmd|
+        cmd.strip!
+        case cmd
+        when "TNS_ADMIN='/opt/oracle/network/admin' LD_LIBRARY_PATH='/opt/oracle/lib' echo 'oracle_query_string';/bin/sqlplus -S -s /nolog <<'INSPECSQL'\nconnect USER/password@XEPDB1_TCPS\nSET PAGESIZE 32000\nSET FEEDBACK OFF\nSET UNDERLINE OFF\nSELECT NAME AS VALUE FROM v$database;\nEXIT\nINSPECSQL" then
```
sathish-progress: This is non deterministic .. even though the build_env_prefix uses `.map` it is still building a hash where in ruby 2.7 + the order is preserved but going by the laws of Hash we shouldn't expect the hash to preserve the order of keys at the application level, hence I would recommend to modify this test to test against proper env vars

test/unit/resources/oracledb_session_test.rb
```diff
@@ -194,4 +194,155 @@
       _(resource.resource_id).must_equal "localhost-1527-USER"
     end
   end
```
sathish-progress, line 196: could you add some more tests to cover the following, 

```
TNS + db_role + su_user - to test the build_connect_string logic where db_role is excluded when su_user is present
TNS + db_role + env 
Special characters in password
```

sathish-progress (changes requested):

test/unit/resources/oracledb_session_test.rb
sa-progress: Updating it , good catch.

sathish-progress (approved):

PR #7685, opened 2025-12-16
CHEF-29378-Remove faraday-middleware for Faraday 2.0 compactability

balasubramanian-s:
## Description

Update `inspec.gemspec` file to address compatibility with Faraday 2.x. The dependency on `faraday_middleware` has been commented out, along with a note explaining that `faraday_middleware` 1.x is incompatible with Faraday 2.x because it requires Faraday ~> 1.0.

Updates the InSpec HTTP resource to be fully compatible with Faraday v2+ while maintaining backward compatibility with Faraday v1.

## Types of changes

## Checklist:


lib/inspec/resources/http.rb
```diff
@@ -149,9 +149,15 @@ def response_headers
         def response
           return @response if @response
 
-          Faraday.ignore_env_proxy = true if proxy == "disable"
```
balasubramanian-s, line 152: The Faraday.ignore_env_proxy property was deprecated and removed in Faraday v2.0. This change removes the dependency on this legacy API while preserving the same functionality through the existing proxy configuration logic.

PR #7751, opened 2026-01-21
Remove gemfilelock from deps

sathish-progress:
## Description

This pull request adds a cleanup step to the build process that removes unnecessary `Gemfile.lock` files from installed gems. This helps ensure that only runtime dependencies are included in the final package, reducing clutter from development artifacts.

Build artifact cleanup:

* Added a block to `remove-old-gems.rb` that finds and deletes all `Gemfile.lock` files from the installed gems directory, as these are development artifacts and not needed for runtime.
## Types of changes

## Checklist:

### DIFF
--- a/omnibus/config/software/remove-old-gems.rb
+++ b/omnibus/config/software/remove-old-gems.rb
@@ -31,16 +31,23 @@ build do
     command "#{gembin} uninstall net-imap -v \"<0.2.5\" -a -I", env: env
   end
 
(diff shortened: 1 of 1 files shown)


Nik08 (approved):

Nik08:
Adhoc omnibus build https://buildkite.com/chef/inspec-inspec-inspec-5-omnibus-adhoc/builds/271 is green

PR #7870, opened 2026-04-20
fix: CHEF-32686 - Pre-check waivers before control block evaluation

chandra-progress:
## Description

CIS Automate profiles use eager resource evaluation where commands like 'find /' are executed directly in the control block. Previously, waivers were only applied after instance_eval, meaning expensive commands ran even for controls waived with run: false.

Move waiver check before instance_eval so controls waived with run: false skip block evaluation entirely. Re-apply waivers after instance_eval for non-waived controls to maintain precedence over only_if.

Foreport #7832 
## Types of changes

## Checklist:

### DIFF
--- a/test/fixtures/profiles/waivers/eager-eval/controls/eager_eval.rb
+++ b/test/fixtures/profiles/waivers/eager-eval/controls/eager_eval.rb
@@ -1,0 +1,31 @@
+# This control uses eager evaluation pattern found in CIS profiles.
+# When waived with run: false, the command should NOT be executed.
+control "01_eager_waived_not_ran" do
+  result = command("echo eager_eval_marker_01").stdout.strip
+  describe result do
+    it { should eq "eager_eval_marker_01" }
+  end
+end
+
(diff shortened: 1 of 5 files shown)


test/fixtures/profiles/waivers/eager-eval/controls/eager_eval.rb
```diff
@@ -0,0 +1,31 @@
+# This control uses eager evaluation pattern found in CIS profiles.
+# When waived with run: false, the command should NOT be executed.
+control "01_eager_waived_not_ran" do
+  result = command("echo eager_eval_marker_01").stdout.strip
```
james-stocks, line 4: This was never supposed to work, to be honest. All code execution should be encapsulated by the resource, otherwise it _will_ be evaluated while InSpec is collecting the controls in the profile.

Nik08: @james-stocks can you please elaborate on this? We want to understand this better.

Nik08 (approved):

PR #7930, opened 2026-06-01
fix: CHEF-35015 — upgrade zlib 3.0.1 and erb 4.0.3.1 gems 

chandra-progress:
## Description
Upgrade `zlib` to `3.0.1` and `erb` to `4.0.3.1` gems at the package level in Habitat install hooks for both Linux and Windows.

There is currently no fix from the Hab team for these core dependencies of Ruby 3.1. This change removes the default gems and installs the upgraded versions as a temporary fix.

## Related Issue
/CHEF-35015

## Types of changes

## Checklist:


habitat/x86_64-linux/plan.sh
```diff
@@ -63,6 +63,11 @@ do_install() {
   # for omnibus we also install this as part of the package
   gem install ed25519 bcrypt_pbkdf --no-document
 
+  # Pre-install native gems that require compilation (zlib, erb) at build time
+  # so the install hook (which runs without make/gcc) can skip reinstalling them.
+  gem install zlib -v 3.0.1 --no-document
```
Nik08, line 68: @chandra-progress Why do we need this change if the upgrade is also managed by hooks?

chandra-progress: Because to install Zlib and erb, It required **make** which is not available at runtime. That is why we install it at build time.
@Nik08

chandra-progress:
@copilot resolve the merge conflicts in this pull request

PR #7940, opened 2026-06-10
fix: pin bigdecimal < 4.0 for Ruby 3.1 compatibility

chandra-progress:
## Description
Pin `bigdecimal` to `< 4.0` as an explicit runtime dependency in `inspec-core.gemspec`.

`bigdecimal` 4.x changed its native extension internals and fails to compile against the embedded Ruby 3.1 environment used by InSpec omnibus packages (`/opt/inspec/embedded`). Adding the constraint directly in the gemspec ensures it applies to all consumers — `bundle install` (local dev + omnibus builds) and direct `gem install inspec` users alike.

## Related Issue
N/A

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -14,6 +14,7 @@ PATH
       train-winrm (~> 0.4.0)
     inspec-core (5.24.15)
       addressable (~> 2.9)
+      bigdecimal (< 4.0)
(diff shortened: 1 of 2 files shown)


Nik08:
Closing this - since this should be on the train's end.

PR #7943, opened 2026-06-15
feat: CHEF-34404 - Add aarch64-linux habitat package support

mohan-progress:
## Description

Add support for building a Habitat package for Linux ARM (`aarch64-linux`) by reusing the existing `x86_64-linux` plan file. The new `habitat/aarch64-linux/plan.sh` sources `habitat/x86_64-linux/plan.sh` directly, ensuring full consistency across platforms without duplicating any logic.

- Created `habitat/aarch64-linux/plan.sh` which sources the shared x86_64-linux plan

The `aarch64-linux` package has been successfully built and installed on an Ubuntu ARM VM (AWS Graviton), and `inspec version` was verified to work correctly.

This work was completed with AI assistance following Progress AI policies.

## Related Issue

CHEF-34404

## Types of changes

## Checklist:

### DIFF
--- a/habitat/aarch64-linux/plan.sh
+++ b/habitat/aarch64-linux/plan.sh
@@ -1,0 +1,2 @@
+# Reuse the x86_64-linux plan for Linux ARM builds.
+source "$(dirname "${BASH_SOURCE[0]}")/../x86_64-linux/plan.sh"
--- a/.expeditor/artifact.habitat.yml
+++ b/.expeditor/artifact.habitat.yml
@@ -23,11 +23,13 @@ steps:
   - label: ":linux: Arm64 Validate Habitat Builds of Chef InSpec"
     commands:
       - .expeditor/buildkite/artifact.habitat.test.sh
-    expeditor:
-      executor:
-        docker:
-          image: ruby:3.4.2-bullseye
+    agents:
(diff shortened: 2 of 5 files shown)


habitat/aarch64-linux/plan.sh
```diff
@@ -0,0 +1,2 @@
+# Reuse the x86_64-linux plan for Linux ARM builds.
+source "$(dirname "${BASH_SOURCE[0]}")/../x86_64-linux/plan.sh"
```
Nik08: @mohan-progress can we not just copy and paste the file instead? Because my concern is whether there could be some changes made for x86_64 Linux that could break the AArch64 Linux plan? Otherwise, the working demo for this task looks great. Thanks!

Vasu1105: There could be 2 possibilites either maintain the seperate file now (which currently is code duplication) or keep the same file unless there is anything specific that needs to be implemented differently for this architecture

mohan-progress: yes @Nik08  I'm completely agree with you,  currently everything is same so I want to avoid code duplication.
In future if something specific to Linux-x86 OR aarch  then we can make it separate. 

Thank you

PR #7950, opened 2026-06-16
fix(source_reader): only claim profile as gem if gemspec is at the root

### LINKED ISSUE #7934: 7.x: GemReader claims any directory profile containing a *.gemspec (e.g. a vendored bundle), reporting zero controls
## Summary

In InSpec 7.x, `SourceReaders::GemReader` claims **any directory profile whose file tree contains a `*.gemspec` anywhere** (not just the profile's own gemspec), and because it has a higher priority than `InspecReader`, the profile is read as a gem resource pack — which exposes **zero controls**. A profile that vendors gems under its own directory (e.g. `bundle install --path vendor/bundle`, or a CI `bundler-cache`) therefore silently runs **no controls**.

## Versions

- **Affected:** inspec-core 7.0.x and 7.1.x (reproduced with `cincproject/auditor` 7.0.107 and 7.1.7).
- **Not affected:** 5.x / 6.x (e.g. 5.24.7, 6.8.24) — they have no `gem` source reader, so `InspecReader` always wins.

## How we hit it

Our profile lives at the repository root (top-level `inspec.yml` + `controls/` + `libraries/`). CI uses the `ruby/setup-ruby` action with `bundler-cache: true` (working directory `test/`), which runs `bundle install` and vendors the test gems into `test/vendor/bundle/ruby/<ver>/`, including `.../specifications/*.gemspec`. When cinc-auditor evaluated the repo root, those vendored gemspecs caused 7.x to discover zero controls — every control reported `status: error`. The same happens after a plain local `bundle install` under `test/`. It was invisible on 5.x/6.x, so the breakage appeared only after `cincproject/auditor:latest` rolled from a 6.x image to 7.1.7.

## Reproduction

```sh
mkdir -p p/controls
cat > p/inspec.yml <<'YAML'
name: demo
version: 0.1.0
supports:
  - platform: linux
YAML
cat > p/controls/example.rb <<'RB'
control 'demo-1' do
  describe true do
    it { should eq true }
  end
end
RB

# Baseline: works
cinc-auditor exec p --reporter json | jq '.profiles[0].controls | length'   # => 1

# Drop a single, unrelated *.gemspec anywhere under the profile dir
touch p/controls/unrelated.gemspec        # or p/vendor/bundle/.../specifications/foo.gemspec
cinc-auditor exec p --reporter json | jq '.profiles[0].controls | length'   # => 0
cinc-auditor check p                                                        # WARN: No controls or tests were defined.
```

## Expected vs. actual

- **Expected:** a directory profile with a valid top-level `inspec.yml` and `controls/` is read by `InspecReader`; unrelated vendored `*.gemspec` files (e.g. under `vendor/`) are ignored.
- **Actual:** `GemReader` claims the profile, `@tests = {}`, and the run reports zero controls and exits 0 — no error, no warning beyond `check`'s "No controls or tests were defined".

## Root cause

`lib/source_readers/gem.rb`:

```ruby
class GemReader < Inspec.source_reader(1)
  name "gem"
  priority 20
  def self.resolve(target)
    return new(target) unless target.files.grep(/gemspec/).empty?   # matches ANY path containing "gemspec"
    nil
  end
  def initialize(target)
    @tests = {}   # "As of July 2024 only resource packs, not controls, may be packaged as gems"
    ...
  end
end
```

The source-reader registry (`Inspec::Plugin::V1::Registry#resolve`) sorts readers by priority, highest first, and selects the first whose `resolve` is truthy. Priorities are `gem: 20`, `inspec: 10`, `flat: 5`. So whenever the file list contains a path matching `/gemspec/`, `GemReader` (20) is tried and claims the profile before `InspecReader` (10) — even though a perfectly valid top-level `inspec.yml` + `controls/` are present.

`GemReader.resolve` matches `/gemspec/` as an unanchored substring against *every* file path in the tree, so `vendor/.../specifications/x.gemspec` — or even a file literally named `unrelated.gemspec` — claims a profile that also has a valid root `inspec.yml`.

## Impact

Any profile that vendors its test/tooling gems inside the profile directory (a common pattern with `bundle config set --local path vendor/bundle`, or `ruby/setup-ruby` with `bundler-cache: true` and the profile at the repo root) will, on 7.x, evaluate to zero controls with no clear error. Depending on the harness, this is easy to misread as a pass or as unrelated errors.

Mallikarjunadevops:
Fixes #7934

## Description

The `GemReader` in InSpec 7.x (`lib/source_readers/gem.rb`) was claiming any directory profile whose file tree contains a `*.gemspec` anywhere (e.g., inside a vendored `vendor/bundle/` path). As a result, the profile was improperly treated as a gem resource pack rather than being read by `InspecReader`.

This PR updates `GemReader.resolve` to only match `.gemspec` files present in the root of the target directory by using a more exact regex `%r{^[^/\\]+\.gemspec$}`.

## Testing

I added unit tests in `test/unit/source_readers/gem_test.rb` simulating `Inspec::FileProvider` output for different target configurations. The tests verify that `GemReader.resolve` successfully identifies root `.gemspec` files while properly rejecting nested `.gemspec` files or missing ones.

Signed-off-by: Mallikarjunadevops

### DIFF
--- a/lib/source_readers/gem.rb
+++ b/lib/source_readers/gem.rb
@@ -7,7 +7,7 @@ module SourceReaders
     priority 20
 
     def self.resolve(target)
-      return new(target) unless target.files.grep(/gemspec/).empty?
+      return new(target) unless target.files.grep(%r{^[^/\\]+\.gemspec$}).empty?
 
       nil
     end
--- a/test/unit/source_readers/gem_test.rb
+++ b/test/unit/source_readers/gem_test.rb
@@ -1,0 +1,48 @@
+require "helper"
+require "inspec/source_reader"
+require "source_readers/gem"
+
+describe SourceReaders::GemReader do
+  let(:reader) { SourceReaders::GemReader }
+
+  it "registers with the source readers registry" do
+    reg = Inspec::SourceReader.registry
+    _(reg["gem"]).must_equal reader
+  end
+
+  describe "resolve" do
+    class MockTarget
+      attr_reader :files
+      def initialize(files)
+        @files = files
+      end
+
+      def read(file)
+        if file == "inspec.yml"
+          "name: my_gem\nversion: 0.1.0\n"
+        else
+          "dummy content"
+        end
+      end
+    end
+
+    it "resolves the target when gemspec is in root" do
+      target = MockTarget.new(["inspec.yml", "my_gem.gemspec", "lib/my_gem.rb"])
+      res = SourceReaders::GemReader.resolve(target)
+      _(res).wont_be_nil
+      _(res).must_be_kind_of reader
+    end
+
+    it "does not resolve the target when gemspec is in a subfolder" do
+      target = MockTarget.new(["inspec.yml", "vendor/bundle/ruby/3.1.0/specifications/other.gemspec"])
+      res = SourceReaders::GemReader.resolve(target)
+      _(res).must_be_nil
+    end
+
+    it "does not resolve the target when there is no gemspec" do
+      target = MockTarget.new(["inspec.yml", "controls/example.rb"])
+      res = SourceReaders::GemReader.resolve(target)
+      _(res).must_be_nil
+    end
+  end
+end


Nik08:
@Mallikarjunadevops I think the fix here https://github.com/inspec/inspec/pull/7962 covers the same kind of fix. If yes, we can close this PR.

Mallikarjunadevops:
Hi @Nik08, yes you are absolutely correct! PR #7962 addresses the exact same root cause (preventing `GemReader` from improperly claiming profiles with vendored `.gemspec` files). 

Our implementations differ slightly (I used a regex to strictly match root-level `.gemspec` files, whereas #7962 explicitly rejects the `vendor/` prefix), but the end result successfully fixes the problem in both cases.

I will go ahead and close this PR in favor of #7962. Thanks!

PR #7970, opened 2026-06-24
feat: CHEF-36276 - Add deprecation warning for --overwrite flag in inspec compliance plugin

Nik08:
## Description
Adds a deprecation warning when the `--overwrite` flag is used with `inspec compliance upload`. The `--overwrite` option does not work as expected while uploading to Chef Automate and will be removed going forward.

When a user passes `--overwrite`, the following warning is now displayed:
```
DEPRECATION: This command does not work as expected while uploading to Automate. This option will be removed going forward.
```

**Changes:**
- Added `cli_option_compliance_overwrite` deprecation group to `etc/deprecations.json` with `action: "warn"`
- Added `Inspec.deprecate` call at the start of the `upload` method in the compliance plugin CLI when `--overwrite` is set
- Added a functional test to verify the deprecation message is shown

## Related Issue
CHEF-36276

## Types of changes

## Checklist:


Nik08:
Closing this because pr got created against main

PR #7977, opened 2026-06-30
fix: CHEF-36279 - Make gcc and make available at runtime for native gem compilation

Nik08:
## Description

When running InSpec (as a Habitat package) on a fresh RHEL VM and executing a profile that installs gems with native C extensions (e.g., the MongoDB resource pack → `mongo` → `bson`), the build fails because build toolchain binaries are not available at runtime.

**Root Cause (discovered in stages):**

1. `core/gcc` and `core/make` were only in `pkg_build_deps` — not available at runtime when InSpec compiles native gem extensions.
2. `core/gcc` in the `base-2025` channel is a **meta-package with no binaries**. The actual compiler lives in `core/gcc-base`. All earlier attempts referencing `core/gcc` binary paths were silently failing.
3. Even with `core/gcc-base` in PATH, `gcc.real` (the internal compiler binary) failed with `cannot execute 'as': posix_spawnp: No such file or directory`. GCC is a driver that orchestrates separate tools — the assembler (`as`) and linker (`ld`) come from `core/binutils`, which was also missing at runtime.

**Full compilation pipeline that native gem builds require:**
```
source.c → cpp (preprocessor) → cc1 (compiler) → as (assembler) → ld (linker)
             [gcc-base]            [gcc-base]       [binutils]      [binutils]
```

**Fix — added to `pkg_deps` (runtime dependencies):**
| Package | Provides |
|---|---|
| `core/gcc-base` | `gcc` wrapper + `cc1` C compiler (replaces meta-package `core/gcc`) |
| `core/gcc-libs` | `libstdc++.so`, `libgcc_s.so` — GCC runtime libraries needed at link time |
| `core/binutils` | `as` (assembler) + `ld` (linker) — required by `gcc.real` internally |
| `core/make` | `make` — runs the generated `Makefile` during gem native extension builds |

**Additional wrapper script changes:**
- Prepend `gcc-base/bin`, `binutils/bin`, and `make/bin` to `PATH` at runtime
- Set `CC=$(pkg_path_for core/gcc-base)/bin/gcc` so `mkmf.rb` uses the Hab-bundled compiler (it reads `RbConfig::CONFIG['CC']` by default, which is hardcoded at Ruby compile time and may point to a non-existent path)
- Set `MAKE=$(pkg_path_for core/make)/bin/make` for the same reason

**Verified on EC2:** The full mkmf compile test (with mkmf's `LD_LIBRARY_PATH` restriction) passes successfully with this toolchain in place. The `bson` gem native extension now compiles successfully on a fresh RHEL VM with no system gcc/make.

This work was completed with AI assistance following Progress AI policies.

## Related Issue
CHEF-36279

## Types of changes

## Checklist:

### DIFF
--- a/habitat/x86_64-linux/plan.sh
+++ b/habitat/x86_64-linux/plan.sh
@@ -14,11 +14,12 @@ pkg_deps=(
   core/git
   core/ruby3_4
   core/bash
-  core/gcc
+  core/gcc-base
+  core/gcc-libs
   core/make
 )
 pkg_build_deps=(
-  core/gcc
+  core/gcc-base
   core/make
   core/readline
   core/sed
@@ -66,7 +67,6 @@ do_install() {
     gem install inspec-bin*.gem --no-document
   popd
 
-  wrap_compiler_bins
   wrap_inspec_bin
 
   # Copy NOTICE.TXT to the package directory
@@ -89,40 +89,6 @@ do_install() {
   find "$GEM_HOME" -xdev -perm -0002 -type f -print 2>/dev/null | xargs -I '{}' chmod go-w '{}'
 }
 
-# Create wrapper scripts for gcc/cc/g++/c++ so that native gem compilation
-# works at runtime. mkmf.rb overrides LD_LIBRARY_PATH when running compile
-# tests (to only include Ruby's lib dir), which prevents Hab's gcc from
-# loading its own shared libraries (libstdc++.so.6, libgcc_s.so.1).
-# A bash wrapper script avoids this problem: it sets LD_LIBRARY_PATH from
-# within the forked process itself, AFTER mkmf's override, so the Hab gcc
-# binary can always find its own shared libraries.
-wrap_compiler_bins() {
-  local gcc_bin_dir="$(pkg_path_for core/gcc)/bin"
-  local gcc_lib="$(pkg_path_for core/gcc)/lib"
-  local gcc_lib64="$(pkg_path_for core/gcc)/lib64"
-  local bash_bin="$(pkg_path_for core/bash)/bin/bash"
-
-  for name in gcc cc g++ c++; do
-    local wrapper="$pkg_prefix/bin/$name"
-    local real_bin
-    case "$name" in
-      cc|gcc)  real_bin="$gcc_bin_dir/gcc" ;;
-      c++|g++) real_bin="$gcc_bin_dir/g++" ;;
-    esac
-
-    build_line "Creating compiler wrapper $wrapper -> $real_bin"
-    cat <<EOF > "$wrapper"
-#!$bash_bin
-# Ensure Hab gcc's own shared libraries (libstdc++, libgcc_s) are available.
-# This is set inside the wrapper so it takes effect even when mkmf.rb
-# overrides the outer LD_LIBRARY_PATH during native extension compilation.
-export LD_LIBRARY_PATH="${gcc_lib}:${gcc_lib64}:\$LD_LIBRARY_PATH"
-exec "$real_bin" "\$@"
-EOF
-    chmod -v 755 "$wrapper"
-  done
-}
-
 # Need to wrap the InSpec binary to ensure paths are correct
 wrap_inspec_bin() {
   local bin="$pkg_prefix/bin/$pkg_name"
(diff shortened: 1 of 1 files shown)


habitat/x86_64-linux/plan.sh
```diff
@@ -14,6 +14,10 @@ pkg_deps=(
   core/git
   core/ruby3_4
   core/bash
+  core/gcc-base
+  core/gcc-libs
```
neha-p6, line 18: Are these packages maintained and kept upto date? Also these are getting added as run time dependencies which means Inspec hab package will now ship with those C libraries so we'll have to make sure they are audited and maintained for security

habitat/x86_64-linux/plan.sh
```diff
@@ -95,8 +99,17 @@ wrap_inspec_bin() {
...
+# core/gcc-base provides a gcc wrapper script (via hab-cc-wrapper) that
+# already handles its own LD_LIBRARY_PATH and environment setup internally,
+# so it works correctly even when mkmf.rb overrides the outer LD_LIBRARY_PATH
+# during native gem extension compilation (e.g. bson for MongoDB resource pack).
+export PATH="$(pkg_path_for core/gcc-base)/bin:$(pkg_path_for core/binutils)/bin:$(pkg_path_for core/make)/bin:/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:\$PATH"
+
+# Explicitly set CC/MAKE so mkmf uses the Hab-bundled toolchain even when
+# RbConfig::CONFIG['CC'] points to a different compiler path.
+export CC="$(pkg_path_for core/gcc-base)/bin/gcc"
+export MAKE="$(pkg_path_for core/make)/bin/make"
```
neha-p6, line 112: Setting these vars make them globally overriden, so what happens on systems where gcc and make is already present assuming this bug is specifically for cases where a system does not have native gcc? Have you tested that use case if that breaks anything? Instead overriding them globally we could make it so that during native gem installations on systems where gcc, make are not present the packaged ones are used

habitat/x86_64-linux/plan.sh
```diff
@@ -95,8 +99,17 @@ wrap_inspec_bin() {
...
 set -e
 
-# Set binary path that allows InSpec to use non-Hab pkg binaries
-export PATH="/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:\$PATH"
+# Set binary path that allows InSpec to use non-Hab pkg binaries.
+# core/gcc-base provides a gcc wrapper script (via hab-cc-wrapper) that
+# already handles its own LD_LIBRARY_PATH and environment setup internally,
+# so it works correctly even when mkmf.rb overrides the outer LD_LIBRARY_PATH
+# during native gem extension compilation (e.g. bson for MongoDB resource pack).
+export PATH="$(pkg_path_for core/gcc-base)/bin:$(pkg_path_for core/binutils)/bin:$(pkg_path_for core/make)/bin:/sbin:/usr/sbin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:\$PATH"
```
mwrock, line 107: hab should be adding these to the path automatically.

habitat/x86_64-linux/plan.sh
mwrock, line 112: keep in mind that these variables are set at build time of the hab package. They are not set at runtime for users of inspec. I'm not sure if that was understood.

habitat/x86_64-linux/plan.sh
Nik08: @neha-p6 I see the last update on this was in 2025. And yes, that could pose security maintenance, but do you have some suggestions for doing this in an alternate way? I thought this could be the only way of doing this via a plan file, since the dependencies could be fulfilled by using the hab components themselves.

habitat/x86_64-linux/plan.sh
mwrock: Actually I just noticed this is all inside of a wrapper. So it would be set when the wrapper is invoked.

habitat/x86_64-linux/plan.sh
mwrock: oh never mind. Just noticed this is inside of a wrapper. So setting these is correct.

mwrock (approved):

habitat/x86_64-linux/plan.sh
Nik08: @neha-p6 Thanks for this pointer. It works even if the system already has GCC present in it. Verified.

PR #7997, opened 2026-08-12
Fix oracledb_session returning nil for multi-column query results

pburkholder:
Hello! It's been nine (!) years since my last issue/PR, but I came across this case, so here we go:

parse_csv_result could only handle single-column results. Two defects
combined to make `.column(name)` return nil for every column of any query
that selected more than one column:

1. `gsub(",", "comma_query_sub")` ran on the entire output before CSV.parse,
   which still used the default comma delimiter. This replaced the structural
   delimiters too, so a row like `NAME,VALUE\nDROP,3` parsed as a single
   column whose header was the literal string "name,value".

2. `row.entries.flatten` turned `[[h1,v1],[h2,v2]]` into `[h1,v1,h2,v2]`, and
   `[revised_row].to_h` cannot represent more than one column (it raises for
   >=2 columns, and degenerates to a single bogus key for exactly 2).

The net effect was a silent wrong answer (nil), which is especially dangerous
for compliance profiles where a nil can flip a control's pass/fail.

sqlplus (SET MARKUP CSV) and sqlcl (set sqlformat csv) both emit standard
RFC 4180 CSV, so parse it directly with CSV.parse and build each row via
row.to_h. This keeps every column addressable and correctly preserves commas
inside quoted fields, while leaving single-column behavior (and the existing
nil/empty-result handling) unchanged.

Add a multi-column unit test and fixture; existing single-column tests still
pass.

Signed-off-by: Peter Burkholder <peter.burkholder@gsa.gov>


imdadnitm (approved):
Hi @pburkholder  The changes looks good to me. Thanks for your contributions.

Nik08 (approved):

PR #8013, opened 2026-09-22
Fix Debian 11 EOL apt 404 errors in CI cache dependency install

chandra-progress:
## Summary

Debian 11 (bullseye) reached EOL, so its `deb.debian.org` / `security.debian.org` repos no longer publish fresh `Release` files, and their package indexes have drifted out of sync with each other. This caused `apt-get install` in `install_cache_deps()` to fail intermittently with 404 errors (e.g. `sudo`) on `ruby:3.1-bullseye` CI pipeline steps:

```
Err:1 http://deb.debian.org/debian-security/main amd64 sudo amd64 1.9.5p2-3+deb11u4
  404  Not Found
E: Unable to fetch some archives
```

## Fix

Added `fix_debian_11_eol_repos()` to `.expeditor/buildkite/cache_support.sh`:

- Detects Debian 11 via `/etc/os-release` (`ID=debian`, `VERSION_ID=11`).
- Disables any live `deb.debian.org` / `security.debian.org` repo entries, in both classic `/etc/apt/sources.list` and deb822 `sources.list.d/*.sources` formats, regardless of location.
- Writes an explicit, pinned `snapshot.debian.org` mirror instead, so `main`/`security`/`updates` stay internally consistent with each other.

`install_cache_deps()` now calls this helper before `apt-get update`, and only relaxes `Acquire::Check-Valid-Until` (needed because the pinned snapshot's own `Release` file is intentionally "expired") when the Debian 11 workaround is actually active. Other apt-based platforms (e.g. Ubuntu) are unaffected and keep the normal freshness check.

This mirrors the equivalent, already-verified fix in chef/knife#206.

## Testing

Verified locally in a real `ruby:3.1-bullseye` container (same image used in CI):

```bash
docker run --rm -it \
  -v $(pwd):/workdir -w /workdir \
  -e BUILDKITE_PIPELINE_SLUG=test -e BUILDKITE_LABEL=test \
  ruby:3.1-bullseye bash
source .expeditor/buildkite/cache_support.sh
install_cache_deps sudo
```

- `apt-get update` succeeds (no "Release file is expired" error) using the pinned `snapshot.debian.org` mirror.
- `apt-get install sudo -y` — the exact package that 404'd in CI — installs successfully.
- `bash -n` syntax check passes.
- Confirmed non-Debian-11 platforms are unaffected (workaround branch only triggers when `/etc/os-release` matches `debian`/`11`).


Nik08 (approved):

chandra-progress:
Closing this PR as we have diff and simple solution to it in another PR
https://github.com/inspec/inspec/pull/8014

